PR #3, opened 2021-10-24
Fixed generic distinct conversion proc calls

beef331:


compiler/jsgen.nim
```diff
@@ -1288,75 +1288,89 @@ template isIndirect(x: PSym): bool =
...
-proc genAddr(p: PProc, n: PNode, r: var TCompRes) =
-  case n[0].kind
-  of nkSym:
-    let s = n[0].sym
-    if s.loc.r == nil: internalError(p.config, n.info, "genAddr: 3")
-    case s.kind
-    of skParam:
-      r.res = s.loc.r
-      r.address = nil
+proc genSymAddr(p: PProc, n: PNode, typ: PType, r: var TCompRes) =
```
saem, line 1291: Can haz doc comments for the uninitiated?

compiler/jsgen.nim
```diff
@@ -1288,75 +1288,89 @@ template isIndirect(x: PSym): bool =
...
+    r.typ = etyNone
+  of skVar, skLet, skResult:
+    r.kind = resExpr
+    let jsType = mapType(p):
+      if typ.isNil:
+        n.typ
+      else:
+        typ
+    if jsType == etyObject:
+      # make addr() a no-op:
```
saem, line 1310: It's that because they're already refs in JS land?

compiler/jsgen.nim
```diff
@@ -1288,75 +1288,89 @@ template isIndirect(x: PSym): bool =
...
-      # namely toOpenArray(a, 1, 3)
+      gen(p, n, r)
+      #internalError(p.config, n.info, "genAddr: 4 " & renderTree(n))
+  else: internalError(p.config, n.info, $("genAddr: 2", s.kind))
+
+proc genAddr(p: PProc, n: PNode, r: var TCompRes) =
+  if n.kind == nkSym:
+    genSymAddr(p, n, nil, r)
+  else:
+    case n[0].kind
```
saem, line 1333: Why do we look at the first child?

compiler/types.nim
```diff
@@ -1121,15 +1121,15 @@ proc sameTypeAux(x, y: PType, c: var TSameTypeClosure): bool =
     case c.cmp
     of dcEq: return false
     of dcEqIgnoreDistinct:
-      while a.kind == tyDistinct: a = a[0]
-      while b.kind == tyDistinct: b = b[0]
+      a = a.skipTypes({tyDistinct, tyGenericInst})
```
saem: Skip types is much better!

Just wondering if we're missing any others? 🤔

tests/converter/texplicit_conversion.nim
```diff
@@ -11,3 +11,9 @@ converter toInt(s: string): int =
 
 let x = (int)"234"
 echo x
+
+block: # Test for nkconv
```
saem: A few words describing what the spec should be/is would help teach people the spec and/or build it out as we go in an incremental fashion.

tests/distinct/tdistinct.nim
```diff
@@ -140,6 +141,22 @@ block tRequiresInit:
     let s = "test"
     doAssert s == "test"
 
+block: #17322
```
saem: I'd prefer we keep nim issue references only for deep dive, but being in information into nimskull.

saem (changes requested):

compiler/types.nim
beef331: Added a `skipDistincts` to make it easier to handle all cases when wanting to skip distincts, still need to replace any uses of skipTypes with it.

saem (approved):

PR #4, opened 2021-10-24
Fixed borrowing procs with typeclasses

beef331:


tests/borrow/typeclassborrow.nim
```diff
@@ -0,0 +1,32 @@
+type
```
saem, line 1: A few one liners saying the intention of the various cases is the only thing that takes some guess work to figure out.

PR #5, opened 2021-10-24
Fixed generic inheritance dispatch

beef331:
Using inherited generics does not dispatch properly and just takes the generic's base. This properly dispatches on the generic types so procs can overide based of inherited parameters. Making the following work:
```nim
type
  A[T] {.inheritable.} = object
  B = object of A[int]
  C = object of A[float]
proc doStuff(a: A[int]) = discard
proc doStuff(a: A[float) = discard
doStuff(B())
doStuff(C())
```

### DIFF
--- a/compiler/sigmatch.nim
+++ b/compiler/sigmatch.nim
@@ -1412,6 +1412,10 @@ proc typeRel(c: var TCandidate, f, aOrig: PType,
     let origF = f
     var f = if prev == nil: f else: prev
 
+    if a.len > 0:
+      let skippedA = a.skipTypesOrNil({tyObject}) # Skips if it's inheriting from a generic
+      if skippedA != nil:
+        a = skippedA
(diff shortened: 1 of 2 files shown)


compiler/sigmatch.nim
```diff
@@ -1412,6 +1412,10 @@ proc typeRel(c: var TCandidate, f, aOrig: PType,
     let origF = f
     var f = if prev == nil: f else: prev
 
+    if a.len > 0:
+      let skippedA = a.skipTypesOrNil({tyObject}) # Skips if it's inheriting from a generic
```
saem, line 1416: One day when we have options in the compiler life will be grand.

saem (approved):

saem:
Could you add a few summary remarks as to what/why in the PR comment?

Hoping to have a useful for commit history over time.

PR #6, opened 2021-10-25
Showing doc comments as valid statements in language spec

beef331:


tests/lang/s01_basics/s01_comments/t1_comments.nim
```diff
@@ -87,4 +87,34 @@ yeaaaaaaah
   ]##
 echo baz
 
-echo 13
\ No newline at end of file
+echo 13
+
+proc doStuff = 
+  ## Doc comments are valid ast so procs with them are valid!
```
saem, line 93: So is this implicitly a discard?

saem (changes requested):

tests/lang/s01_basics/s01_comments/t1_comments.nim
beef331: Discard isnt unique any ast inside a block will make it valid, discard is another no op.

saem: I mean they're having the same effect here?

In which case a comment would help get the point across that these are equivalent to a no op/discard.

saem (approved):

PR #7, opened 2021-10-26
`InstantiationInfo` => `instantiationInfo`

ringabout:
There is no `InstantiationInfo` in system


ringabout:
Well, there is no CI now.

saem:
Is the same true for the assertion defect?

saem:
I'm guessing this isn't important since there was no reply.

PR #11, opened 2021-10-28
disable custom pragmas tests for let/var

ringabout:
ref https://github.com/nim-works/nimskull/commit/a3b142887df719079bdc183a732b45b3e4a2eca7


ringabout:
^ @alaviss

ringabout:
It seems to be blocked at megatest
https://github.com/xflywind/nimskull/runs/4033530518?check_suite_focus=true
image

PR #21, opened 2021-10-28
enable squash commits for bors

alaviss:
Known issues:

https://github.com/bors-ng/bors-ng/issues/1217

https://github.com/bors-ng/bors-ng/issues/987

We should consider trying Github's merge queue, which seems to work like bors, but its in limited beta atm: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/using-a-merge-queue

PR #23, opened 2021-10-29
add workflow to enforce one-commit PRs

alaviss:
An alternative for #21 where we require contributors to squash their commits.

Potential issue https://github.com/bors-ng/bors-ng/issues/390#issuecomment-398923658, but the test is short enough that it should never be a problem in practice.

### DIFF
--- a/.github/workflows/pr-req.yml
+++ b/.github/workflows/pr-req.yml
@@ -1,0 +1,23 @@
+name: Requirements for merging PR
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


haxscramper:
Documentation or contributor guide should explain why this is needed and how I'm expected to achieve this. For example - if I have a PR, then got review saying I need to fix something - what should I do. Force push the same commit, or use some special git magic to do everything properly.

PR #24, opened 2021-10-29
Removed marshal ops from VM

beef331:
Removing marshal ops makes the VM procedure call generation faster and no one used it anyway, in macro heavy compilation can see around 10% faster compilation.


disruptek:
bors +r

disruptek:
No, silly goose.
bors +r
bors +r
bors +r

PR #27, opened 2021-10-30
Rework qualified lookup with nkerror

saem:


saem:
Need to make the tests passed

alaviss:
Let's test this out a bit:

bors try

saem:
Thanks to @beef331 got unblocked on the compiles issue with nkError, it's actually a really tidy fix.

Now onto pragmas, which is the next hill, see the comment here: https://github.com/nim-works/nimskull/pull/27/commits/d2368efc799c3f9e81ab1caba1608f1fcd6ef8dd#diff-1d1109c3f7bc16cf4e770fe792e9073bff67591721f3c72b96cb4b4a3b76bd88R806-R817

PR #28, opened 2021-10-30
Specification tests PR1

haxscramper:
Full spec todo - https://github.com/nim-works/nimskull/issues/10

Issue #10, opened 2021-10-27
Language specification in form of tests


haxscramper:
Similarly to the https://github.com/nim-works/nimskull/tree/devel/tests/lang - implement language specification in form of proper tests. There are a lot of things to be checked, but a rough list includes at least (suggestions for more bullet points to test are welcome).

Example implementation PRs

- https://github.com/nim-works/nimskull/pull/28
- https://github.com/nim-works/nimskull/pull/43

krux02:
Nice comment on macros.

haxscramper:
For tests that involve types also check everything with distinct. 

> So it's good to put the 3 permutations of them in different places
> A distinct to a type, a generic distinct, and a distinct of a generic distinct
> Basically anywhere you *should* be able to put a type ensure you put those distincts

```nim
type
  A = distinct int
  B[T] = distinct A
  C = distinct B[string]
```

haxscramper:
I would much rather prefer to have some built-in language construct to assert compilation errors, instead of testament hack

```nim
let failure = getFailedNkErrorNodes:
  SomeCodeThatMakesNoSenseSemantically
doAssert failure.description.contains "illegal recursion in type 'Object'"
```


```nim
discard """
  cmd: "nim check $file"
  nimout: '''teffects1.nim(22, 28) template/generic instantiation from here
teffects1.nim(23, 13) Error: can raise an unlisted exception: ref IOError
teffects1.nim(22, 29) Hint: 'lier' cannot raise 'IO2Error' [XCannotRaiseY]
teffects1.nim(38, 21) Error: type mismatch: got <proc (x: int): string{.noSideEffect, gcsafe, locks: 0.}> but expected 'MyProcType = proc (x: int): string{.closure.}'
.raise effects differ'''
"""
{.push warningAsError[Effect]: on.}
type
  TObj {.pure, inheritable.} = object
  TObjB = object of TObj
    a, b, c: string

  IO2Error = ref object of IOError

proc forw: int {. .}

proc lier(): int {.raises: [IO2Error].} =
  #[tt.Hint                 ^ 'lier' cannot raise 'IO2Error' [XCannotRaiseY] ]#
  writeLine stdout, "arg" #[tt.Error
            ^  can raise an unlisted exception: ref IOError
  ]#

proc forw: int =
  raise newException(IOError, "arg")

{.push raises: [Defect].}

type
  MyProcType* = proc(x: int): string #{.raises: [ValueError, Defect].}

[...]

Issue #8, opened 2021-10-26
Will `nimskull` embrace breaking changes?


ringabout:
For instance, some features I want to see:

- identifier sensitive and `_` can be used among identifier names for C interoperability 
- implement https://github.com/nim-lang/Nim/issues/18422 (breaking change)
- remove some nasty or unused features (remove concepts or not; remove method (at least generic based method) and implement vtable instead)

Or `nimskull` will be compatible with Nim as far as possible?

saem:
Please see the readme specifically the [direction](https://github.com/nim-works/nimskull#direction) section.

You'll see we're willing to make breaking changes; however, breaking for the sake of breaking or minor convenience isn't something we have the capacity to work through just yet.

If they're on mission then all good, otherwise they're presently a distraction. Specific areas where breaking changes are in fact very welcome are the currently on mission items. Chiefly, nkError related work.

We can consider further ones that aren't explicitly part of the current mission as long as they show excellent taste in reducing the over size of the language, standard library, tool chain, and/or bring about greater consistency in both the language and implementation.

The first two changes are to areas that aren't actively being thought about but I'm curious which unused features you'd consider removing?

ringabout:
Allowing identifiers starts with underscore was discussed: 
https://github.com/nim-lang/RFCs/issues/428

> remove some nasty or unused features

such as `drnim`,
something like `proc find*(buf: cstring, pattern: Regex, matches: var openArray[string], start = 0, bufSize: int)`,
replace generic threads procs with low level threads procs wrapper.
dispatching by procs parameter names (`proc hello(a: int)` and `proc hello(b: int)`),
remove `owned`

haxscramper:
`nimsuggest` could also be deprecated in the future in favor of **much** more widely used LSP protocol that is adopted by virtually any modern text editor there is.

Probably can kill TRM macros since there is hardly any use for them, even if they weren't that buggy. 

Maybe move some stdlib modules out into https://github.com/disruptek/dist or deprecate them entirely.

disruptek:
Just my opinion, but the problem with term-rewriting macros is that they are just way too powerful and become essentially magical in use.  It's a tough call.

saem:
> Just my opinion, but the problem with term-rewriting macros is that they are just way too powerful and become essentially magical in use. It's a tough call.

[...]

PR #32, opened 2021-10-31
update readme formatting

shayanhabibi:
Some formatting suggestions to make the readme cleaner and less choc full of text

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,4 +1,26 @@
-# <img src="https://raw.githubusercontent.com/nim-lang/assets/master/Art/logo-crown.png" height="28px"/> Nimskull
+<div id="top"></div>
+
+<br />
+
+<div align="center">
+  <a href="https://github.com/nim-works/nimskull">
+    <img src="https://raw.githubusercontent.com/nim-lang/assets/master/Art/logo-crown.png" height="80px"/>
+  </a>
+
+  <h3 align="center">NimSkull</h3>
+  <p align="center">
+    The Nim-Works compiler, stdlib, tools, and documentation repository.
+    <br />
+    <br />
+    <!-- <a href="https://github.com/nim-works/nimskull/issues">Report Bug</a>
+    ·
+    <a href="https://github.com/nim-works/nimskull/issues">Request Feature</a> -->
+  </p>
+</div>
+
+<br />
+
+## About the Project
 
 The Nim-Works compiler, stdlib, tools, and documentation repository. Nim-Works
 is presently a derivative of [Nim][nim-site]. Its intention best described by:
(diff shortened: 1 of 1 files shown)


haxscramper:
I don't think it is necessary to "Directions", "Near-term development" portions of the readme in spoilers, they are important enough to keep them open.

shayanhabibi:
I would disagree; I think that having graphics of the near-term development instantly viewable is a good idea, but the directions should be slimmed into a short stunner that contains the meat of the project and then the rest hidden within a spoiler.

Until that happens I think it should be kept in a spoiler (as its a huge block of text)

The idea is to create a cleaner and easier to navigate readme.

shayanhabibi:
Actually I would say there are a couple short stunners already in the direction, I just don't know which one saem would want to be forerunner

saem:
> Actually I would say there are a couple short stunners already in the direction, I just don't know which one saem would want to be forerunner

Curse of knowledge, I don't either.

The main thing this needs to do is:
- show new folks the what, how, and why of contribution
- remember things for current contributors so they can refer back

Practically speaking, @shayanhabibi I think you're representative of a major part of the audience. Eager and able, now need to understand enough to turn that into action.

haxscramper:
Maybe some sort of FAQ? 

From discussion in `#main` - I guess these questions worth mentioning/answering. Probably something else as well, but I just went over and noted questions that seemed meaningful enough


-------------

> what's the rational?

We got tired of the way Nim is going, especially development process
> What are you going to do now?

https://github.com/nim-works/nimskull/projects
> Will this break my code?

Maybe https://github.com/nim-works/nimskull/issues/8 https://github.com/nim-works/nimskull#direction

> Will you plan to merge changes back?

Generally speaking no, but there might be some exceptions
> Can I help somehow?

Presently we're more interested in people contributing than growing a
community, so if you want to see it sooner, get on helping hax with language spec tests
> Any chat room on matrix/irc/discord?

Not now, we will add it sometimes later
> So the aim isn't just for a different compiler, but a new language? Similar
> but different? So tutorials / books will be written eventually?

Just more well-defined spec through excessive testing, which might lead to slight
language changes to make it more correct. At least that's my take.

---------------

haxscramper:
In addition to that (elaborate on the "direction where community is going")

> failing commit rates, lack of future vision from the community leadership, half a dozen different directions the language is trying to evolve while having only a couple of fully-time contributors, lack of transparency when it comes to discussion of the new features, complete unwillingness to break anything even if it means sinking deeper and deeper in backwards compat hacks while having very few core contributors, dependence on status basically tying araq's hands and shoving their targeted features as they see fit
> we added 15 new stdlib modules
> 2021 was supposed to be the year of tooling
> do you know what is the 1.8 target features?
> do you know what is the main target objective right now?
> do you know the roadmap for the commuity in the next quarter at least?
> It is not that I don't I feel bad about the language, not in the slightest (well, maybe there are some stuff in the stdlib that I don't like, but that is not as fundamental),
> I like the language and where it is heading
> despite all of these shortcomings
> but man it is starting harder to figure out where are we heading

^ (from discussion in  `#offtopic, I did rephrase some things to avoid directly naming anyone)

> if you mean whether we are willing to do breakages - then yes, but we are not aiming to just destroy everything for the sake of some abstract understanding of what constitutes better
> But in the end right now this is not something that we are really inviting people to contribute and discuss, mostly because there are a lot of things that need to be figured out
> So we are afraid of taking too much different options at once, most of which we won't be able to meaningfully process

^ That one is for telling people about our plan for managing community

shayanhabibi:
Dunno if I want to include too much of the negatives from that last post. IMO the read-me should instill an air of positivity around this project; I'd rather avoid the rest and leave that for perhaps a separate doc that can be linked from the readme..

Anyway, have a looksy @haxscramper and see what you think.

readme.md
```diff
@@ -1,191 +1,303 @@
...
+</details>
+
+<details>
+<summary class="blue">Will you plan to merge Nim upstream changes?</summary>
+Generally? No. There will likely be exceptions to this. Our current focus remains on solidifying the foundations before moving forward.
+</details>
+
+<details>
+<summary class="blue">Can I help somehow?</summary>
+Presently we're very interested in people contributing than to growing our community; so if you would like to contribute, a good start is to help haxscramper with the language spec tests
```
haxscramper: > a good start is to help haxscramper with the language spec tests 

Don't redirect to me, redirect to #10 for to-do, or #28 for examples on how this should be done.

haxscramper:
Aside from the "help haxscramper" (I'm just a contributor here, people should be thinking about helping the project itself), I think it looks good.

I added in negative parts as well mostly for the sake of reference, it does not have to be included directly in the README (though "we believe that an element of the vision was lost" can be elaborated a little more)

shayanhabibi:
Hahaha yeah I probably should have looked at the convo on matrix before copy pasting the FAQ answers. I thought you were actually suggesting those as answers

saem:
Are we g2g in this?

haxscramper:
>  > a good start is to help haxscramper with the language spec tests 

> Don't redirect to me, redirect to #10 for to-do, or #28 for examples on how this should be done.

https://github.com/nim-works/nimskull/pull/32#discussion_r740924218 still not fixed, otherwise it looks good to me

readme.md
saem:
```suggestion
Presently we're very interested in people contributing than to growing our community; so if you would like to contribute, a good start is to help the [language spec as tests effort which is being led be @haxscramper](https://github.com/nim-works/nimskull/issues/10). If you're willing to drive deeper into the compiler then see the [nkError refactor to make the compiler approachable](https://github.com/nim-works/nimskull/projects) project.
```

saem:
> > > a good start is to help haxscramper with the language spec tests
> 
> > Don't redirect to me, redirect to #10 for to-do, or #28 for examples on how this should be done.
> 
> [#32 (comment)](https://github.com/nim-works/nimskull/pull/32#discussion_r740924218) still not fixed, otherwise it looks good to me

I used the suggest feature, how does the suggestion look?

haxscramper:
good, I think we can leave it at that

readme.md
saem: I'm on my phone so going to shy away from hitting commit as I can't really ensure the formatting and everything checkout.

So @shayanhabibi can you haz a look and accept if acceptable or adjust as you'd like?

PR #35, opened 2021-11-02
initial work for an action that squash PR into one commit

alaviss:
Goals:
- [x] Retain all co-authors
- [x] Generate commit message based on
  - [x] Title
  - [x] Body
  - [ ] Shortlog of all commits


alaviss:
It works, but the mechanism is too complicated and it doesn't support any form of access control.

Will replace with a script for the user later.

PR #43, opened 2021-11-07
Specification tests PR2

haxscramper:


tests/lang/s01_basics/s00_atoms/t00_builtins.nim
```diff
@@ -0,0 +1,50 @@
...
+  doAssert value == 1
+
+  ## Decrement value using `dec` command
+  dec value
+
+  ## Value of the variable returned to 0
+  doAssert value == 0
+
+block assert_type_of_the_expression:
+  ## Nim is a statically typed programming language, which means every
```
saem: I'd prefer saying Nimskull over Nim.

tests/lang/README.md
```diff
@@ -53,3 +53,11 @@ block failing_example:
 # Example of the failing code
     
 ```
+
```
saem, line 56: Thank you, this will make it easier to explain.

tests/lang/s01_basics/s00_atoms/t01_statement.nim
```diff
@@ -0,0 +1,169 @@
...
+  ## Statement 3 - asserts value of the variable is equal to 1
+  doAssert value == 1
+
+block noreturn_statement:
+  ## noreturn statements allow to affect control flow of the program in
+  ## different ways, depending on the context. Each specific type of the control
+  ## flow-altering statement is more thoroughly discussed in the respective
+  ## sections, this one only provides a higher-level overview of the subject.
+
+  block break_statement:
```
saem, line 33: Not a deal breaker, but I haven't pieced together why break before loops?

tests/lang/s01_basics/s00_atoms/t03_compiletime.nim
```diff
@@ -0,0 +1,45 @@
...
+  echo "echo inside of static"
+
+## This code will be executed only when compiled binary is run.
+echo "echo outside of static"
+
+static:
+  ## This code will also be executed during compilation.
+  echo "second echo inside of static"
+
+## It is possible to define compile-time variable using `{.compiletime.}` pragma
```
saem, line 27: I thought it was trickier than that and it could be referenced by runtime code?

tests/lang/s01_basics/s00_atoms/t01_statement.nim
haxscramper: Because break does not require loops, it can work with `block` as well, and block is a simpler structure, so probably better explained first.

tests/lang/s01_basics/s00_atoms/t00_builtins.nim
haxscramper: I think it does not really make sense to mention either of them and "Nim is a statically typed programming language, which means" is also not needed. We just state the fact (specification) - not try to reason about why this made the way it is now.

tests/lang/s01_basics/s00_atoms/t03_compiletime.nim
haxscramper:
```nim
var globalCompiletimeInt {.compiletime.}: int = 12

macro compiletimeGetter() =
  echo "current value of the compiletime: ", globalCompiletimeInt

compiletimeGetter()

static:
  globalCompiletimeInt = 12

compiletimeGetter()

## Data can be moved from `compiletime` variables into `const`

const canAccessAtRuntime = globalCompiletimeInt

doAssert canAccessAtRuntime == 12
doAssert globalCompiletimeInt == 12

## Note that subsequent modifications of the compiletime variable will not
## affect value of the constant after it has been declared.

static:
  globalCompiletimeInt = 24

const newConst = globalCompiletimeInt

doAssert newConst == 24
doAssert canAccessAtRuntime == 12
doAssert globalCompiletimeInt == 12

```

well, it seems like it is *possible* to reference compiletime variable by the code, but the behavior does not really make a lot of sense

haxscramper: I added more complex tests, but I don't know how to explain it, because this behavior does not make a lot of sense to me as it stands now.

saem: Yeah, I think we can take a sunset of the current behavior and say that's what makes sense and what were supporting.

It's currently ridiculous

tests/lang/s05_pragmas/s01_interop/t02_cxx_header.hpp
```diff
@@ -10,11 +10,39 @@ struct __cxx_internal{
   CxxForward* forward;
 };
 
-template <typename T1, typename T2, typename T3 = int>
+struct CxxComplete {
```
saem: For a future improvement, we should split out the backend code or platform specific pragmas into their own thing -- it's not very basic or simple and they should come fairly late.

Moreover, as these are the ones that are most likely to be poorly thought out in the first place we want the fewest things depending upon them, even in testing.

This just needs a good place to be noted down so we don't ossify this decision.

tests/lang/s01_basics/s00_atoms/t01_statement.nim
saem: From layering the language perspective. Blocks, loops, etc... are sorta independent, then break applies to all of them, so it's "on top".

This won't make a difference for a while, but for things like say slowly implementing CPS within the compiler, then the above logical flow makes a pretty big difference for an implementer in terms of working through language support in a nice sequential fashion. Likely other benefits too.

shayanhabibi:
Please review impacted files from my edits/proofing.

tests/lang/s01_basics/s00_atoms/t01_statement.nim
```diff
@@ -58,47 +81,60 @@ block noreturn_statement:
...
         ## First statement in the loop is executed.
         cond = false
+        ## Setting this variable to false will mean the next iteration of the
+        ## cyclic path will not proceed since `cond` is no longer true.
 
-        ## Upon reaching `continue` control flow is transferred to the next
+        ## Upon reaching `continue`, control flow is transferred to the next
         ## iteration.
         continue
+        # The control flow will return to the beginning of the cyclic path;
```
haxscramper: Must be a documentation comment

tests/lang/s01_basics/s00_atoms/t01_statement.nim
```diff
@@ -58,47 +81,60 @@ block noreturn_statement:
...
+    ## of execution and returns to the beginning of the cyclic path. This is different
+    ## to `break` which ends the cyclic path altogether and instead returns to
+    ## its parent path.
 
     block continue_in_while:
       var value = 0
       var cond = true
 
-      while cond:
+      while cond: # This cyclic path can only be entered if `cond` is true.
```
haxscramper: Move to documentation comment, this should be accessible for the current code parser if we ever decided to generate `manual.rst` from this specification.

tests/lang/s01_basics/s00_atoms/t01_statement.nim
```diff
@@ -107,34 +143,83 @@ block noreturn_statement:
...
+      ## postContinue = 9
+      ## 
+      ## i = 3
+      ## preContinue = preContinue + i
+      ## postContinue = 9
+      ## 
+      ## i = 4
+      ## preContinue = preContinue + i
+      ## postContinue = 9
+      ## ```
```
haxscramper: Overly verbose, either reduce number of iterations to the minimal possible value or simplify explanation.

tests/lang/s01_basics/s00_atoms/t01_statement.nim
```diff
@@ -146,24 +231,29 @@ block noreturn_statement:
...
+      ## When exception is raised it propagates through all function calls (if
+      ## any) until it reaches the topmost level (main module/path from which the first
       ## function was called) or a `try` statement. For particular details on
       ## how exceptions affect control flow see respective part of the
       ## specification. In this case `try` was needed so subsequent `doAssert`
       ## would be executed correctly.
       raiseInside()
 
     except:
+      # `try`, `except`, `finally`, will be explored in their respective section.
```
haxscramper, line 230: Link concrete specification session

haxscramper (commented):
All explanation comments in code should be a documentation comments. General explanation edit seem fine to me.

tests/lang/s01_basics/s00_atoms/t01_statement.nim
haxscramper: Actually it is not written yet, so add `FIXME` for the time being

shayanhabibi: tbd - specification session still not written afaik

tests/lang/s05_pragmas/s02_misc/t06_object.nim
```diff
@@ -0,0 +1,24 @@
...
+  ## The union pragma can be applied to any object type. It means all of the object's 
+  ## fields are overlaid in memory.This produces a union instead of a struct in the 
+  ## generated C/C++ code.
+  type
+    Union {.union.} = object
+      field1: int
+      field2: array[12, int]
+      field3: tuple[idx1, idx2: int]
+
+  doAssert sizeof(Union) == max(sizeof(int), sizeof(array[12, int]))
```
saem, line 11: So we prove the size works in earlier tests?

Likely need to more todos about this for backend/platform spec tests.

saem: We should max all three

tests/lang/s05_pragmas/s02_misc/t05_procedure.nim
```diff
@@ -0,0 +1,7 @@
+proc getGlobal(): int = 
+  var state {.global.}: int
```
saem, line 4: Scoped locally but stored globally?

If so that can be the comment by for now

tests/lang/s05_pragmas/s02_misc/t03_hint_pragmas.nim
```diff
@@ -0,0 +1,42 @@
...
+
+{.hint: "User-provided hint message".}
+
+
+proc gen1[T](arg: T) =
+  {.hint: "gen1 size of the argument is " & $sizeof(arg).}
+
+gen1[int](1)
+
+when false:
```
saem: Can we make a const with a name indicating, skip because broken, and when on that, this way we know we're skipping it and it's searchable?

tests/lang/s05_pragmas/s02_misc/t03_hint_pragmas.nim
```diff
@@ -0,0 +1,42 @@
...
+  proc gen2[T](arg: T) =
+    {.line: instantiationInfo().}:
+      {.hint: "gen2 size of the argument is " & $sizeof(arg).}
+
+  gen2[int](1)
+
+  # REVIEW - this does not work. It should work or not?
+  {.line: ("override-file.nim", 999, 999).}:
```
saem: I don't think these use the same data, the former will effect stack spaces because they change the generated code, while emitting the hint uses the AST and its location info

tests/lang/s05_pragmas/s02_misc/t02_flag_pragmas.nim
```diff
@@ -0,0 +1,67 @@
...
+block enable_bound_checking:
+  {.push boundChecks:on.}
+  proc impl() = 
+    var gotDefect = false
+    try:
+      var a: array[10, int]
+      let idx = 9
+      assert cast[ptr array[9, int]](a.addr)[idx] == 0
+
+    except IndexDefect:
```
saem, line 19: Wait, you can catch defects?

tests/lang/s05_pragmas/s01_interop/t03_emit.nim
```diff
@@ -0,0 +1,26 @@
...
+
+  var struct = CStruct()
+  struct.field = 12
+
+block interpolate_variables:
+  proc impl() = 
+    var nimValue: cint = 0
+
+    doAssert nimValue == 0
+    {.emit: [nimValue, " += 2;"].}
```
saem, line 19: Cute

tests/lang/s05_pragmas/s01_interop/t02_cxx_interop.nim.hpp
```diff
@@ -0,0 +1,62 @@
+#ifndef __CXX_HEADER_H_
```
saem, line 1: It burns, it burns!

tests/lang/s05_pragmas/s01_interop/t01_c_interop.nim
```diff
@@ -0,0 +1,59 @@
...
+description: '''
+Pragmas for wrapping C code in nim.
+'''
+targets: "c"
+cmd: "nim c -r $options $file"
+
+joinable: false
+"""
+
+{.compile("t01_c_interop.nim.c", "-DNIM_DEFINED_FLAG").}
```
saem, line 11: Future thing to test, error out when the wrong backend

tests/lang/s01_basics/s09_control_flow/t03_case.nim
```diff
@@ -0,0 +1,132 @@
...
+      doAssert value == 12
+
+    casewrap()
+
+block case_for_strings:
+  ## In addition to handling ordinal values, `case` also supports matching
+  ## against strings.
+
+  var value = 0
+  case "string":
```
saem, line 114: Didn't want to test case without colon after the expression?

tests/lang/s01_basics/s09_control_flow/t03_case.nim
```diff
@@ -0,0 +1,132 @@
...
+      ## This assert will never be exectude because previous statement executed
+      ## `break name_2` statement.
+      doAssert false
+
+  block case_return:
+    proc casewrap() =
+      let value = case small1:
+                    of small1: 12
+                    of small2: 13
+                    else: return
```
saem, line 103: A case can elif, I thought

saem (changes requested):

tests/lang/s01_basics/s00_atoms/t01_statement.nim
```diff
@@ -0,0 +1,239 @@
+discard """
+description: '''
+This test covers statements, sequential evaluation and control flow.
+'''
+"""
```
alaviss, line 7:
```suggestion
joinable: false
"""
```

Since this test uses `quit`, it cannot be joined by megatest (if joined, it will terminate the test early)

tests/lang/s01_basics/s00_atoms/t03_compiletime.nim
```diff
@@ -0,0 +1,50 @@
+# Given this is a specification I don't think it is an appropriate place to
```
alaviss, line 11:
```suggestion
discard """
output: '''
echo outside of static
'''
"""

# Given this is a specification I don't think it is an appropriate place to
```

This one needs an output specifier or megatest will complain about extra output.

alaviss (commented):
These two changes will get CI to pass for you.

tests/lang/s05_pragmas/s02_misc/t02_flag_pragmas.nim
haxscramper: Yes, it is possible to catch defects just "not recommended" because of the `--panics:on` that someone *might* turn on.

saem:
Todo, remove panics flag?

saem:
Just had another quick read, it's looking good.

A more natural architecture is starting to show up, the organization at the language concept level is helping clarify in my head some of the structuring to be done in the compiler.

saem (approved):

PR #45, opened 2021-11-12
remove coro from std

ringabout:


tests/coroutines/tgc.nim
```diff
@@ -1,5 +1,6 @@
 discard """
   matrix: "--gc:refc; --gc:arc; --gc:orc"
+  disabled: true
```
saem: Could you add a readme for this directory stating that we disabled it, and the replacement for it will serve similar purposes but will be CPS based.

Or just kill this all together.

saem:
One little bit of cleanup, but this is great.

tests/coroutines/tgc.nim
ringabout: Make sense, I will remove the directory.

ringabout: done

saem:
Bors r+

disruptek:
I'm in favor of removing stuff, but I think this code isn't garbage and should be moved into an external library.  That said, we can always create such libraries from mainline sources, so :+1: but we need to keep track of things like this that represent legitimately valuable software which simply shouldn't be part of the compiler.

PR #64, opened 2021-11-16
fixing regression in vm.nim... maybe?

zacharycarter:
I have no idea if this actually fixes the problem - I know it gets me passed the compiler error I described [here](https://github.com/alaviss/union/issues/16)

As far as what the implications / side effects of calling `skipTypes` vs `skipTypesOrNil` are, I have no clue... I have no experience really hacking on the Nim compiler and barely any with the VM. 

I know I was getting a sigsev and based on the stacktrace it was pretty easy to track down to this line. I looked at what `skipTypes` did and figured that if the `PNode`'s `type` field is assigned a nil value, it would bomb out, so I thought this might be the change. Lo and behold, it worked - or at least allowed my code to compile. 

I provided a simple example harness in that repo but you can probably just take the one nim file I provided `profile.nim` and run it against nimskull `HEAD` and get the error described in the issue.

### DIFF
--- a/compiler/types.nim
+++ b/compiler/types.nim
@@ -1127,7 +1127,12 @@ proc sameTypeAux(x, y: PType, c: var TSameTypeClosure): bool =
     of dcEqOrDistinctOf:
       a = a.skipDistincts()
       if a.kind != b.kind: return false
-
+  
+  #[
+    The following code should not run in the case either side is an generic alias,
+    but it's not presently possible to distinguish the genericinsts from aliases of
+    objects ie `type A[T] = SomeObject`
+  ]#
   # this is required by tunique_type but makes no sense really:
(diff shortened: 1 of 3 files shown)


beef331:
Instead of this it probably should be what I eventually PR'd to mainline nim as it's more sensible and less prone to bugs. https://github.com/nim-lang/Nim/blob/devel/compiler/vm.nim#L1045-L1048 is more proper. There is still a bug that generic aliases do not resolve properly correctly on both sides `sameType(A, B) != sameType(B, A)` in that case. Really this entire commit should be brought over, now that I think about it. Forgot to PR it here since I was trying to resolve the overall bug but never did it. https://github.com/nim-lang/Nim/commit/cc984217a98438de5a36447ea6574eb2c5da4ae4

PR #65, opened 2021-11-16
fixing opcSameNodeType check

zacharycarter:


saem:
Huh, why'd the ci bits get cancelled?

Anyways, let's see what the @beef331 says.

beef331:
Should be fine, and the explanation seems fine. `type A[T] = int` will make a `genericInst -> int` so it's just not possible to properly handle genericAliases.

Issue #61, opened 2021-11-15
testament add `knownIssue` and simplify `disabled`


saem:
Testament presently treats `disabled` to mean two things:
1. this tests isn't working for some reason (issue, CI woes, etc) don't run it
2. disable it in certain parts of the test execution matrix. 

The idea here is to:
- create a `knownIssue` (to be finalized) field in the testament spec that allows either:
  - a truth-y value, including a reference to an issue
  - a false-y value, which is also the default
- this field is interpreted as disable due to a broader issue
- while the existing `disabled` field will now no longer accept booleans, and instead only carry build matrix related disabling

See original discussion link for full context below.

----
### How will this be used?

This is follow-on work, but the intention is to enable:
* CI can then run `knownIssue` tests and if they start working as expected it can take actions to update issues/PRs referencing it or other such nice to haves
* this also allows for a fast path of issue to tests case, where manual correlation is less burdensome; instead effort can be spent on refining the analysis (likely multi-factor) and development of associated specifications and test cases

### How do `knownIssue` and `disabled` fit together with everything else?

This handy diagram helps explain much of this:
![](https://user-images.githubusercontent.com/20562256/141661199-faab6bad-b414-4fe2-9b96-3ff3fe00bbc3.png)

----
## Links and References
_Originally posted by @haxscramper in https://github.com/nim-works/nimskull/discussions/60#discussioncomment-1639893_

saem:
Ongoing:

### Testament Related Discoveries/Notes:
1. multiple occurrences - `disabled` can be specified multiple times in a test, so can `ccodecheck`
2. matrix - these are `;` separated command line parameters, where the test definition results in `matrix X targets` distinct runs
3. disabling platforms/runtimes - are done via platform name detection + a `when`, eg: `when defined(posix): ...`, this

Initial conclusions:
* multiple occurences - for `knownIssue` makes sense, but it should likely preclude false-y values

saem:
Not sure how big of an issue it is, but if a test has `knownIssue: 'nimsulk/72'` or whatever, then it'll mark the whole tests for all platforms as a known issue.

Current stats:

|files|instances|platform specific|
|-----|-------------|----------------------|
|85|104|~50%|

saem:
`knownIssue` has been fully implemented, along with testing for changes to results in CI, considering this complete.

PR #30, opened 2021-10-31
disable CI on draft pull request

alaviss:
As requested by @saem

PR #68, opened 2021-11-20
Added knownIssue as key for testament spec

canelhasmateus:
This references issue [61]( https://github.com/nim-works/nimskull/issues/61 )

- added 'knownIssue' as key for spec.
- 'disabled' is now interpreted as a semicolon separated list of (platform names or truthy values).

KnowIssues is allowed to be mentioned multiple times;
The proposal of no longer accepting boolean values seems to be more intricate, since various tests rely on this option.
Still not handling the case of knownIssue for specific platforms.

### DIFF
--- a/testament/specs.nim
+++ b/testament/specs.nim
@@ -104,6 +104,7 @@ type
                       # but don't rely on much precision
     inlineErrors*: seq[InlineError] # line information to error message
     debugInfo*: string # debug info to give more context
+    knownIssues*: seq[string] ## known issues to be fixed
 
 proc getCmd*(s: TSpec): string =
   if s.cmd.len == 0:
@@ -275,7 +276,7 @@ proc parseSpec*(filename: string): TSpec =
     case e.kind
     of cfgKeyValuePair:
       let key = e.key.normalize
-      const allowMultipleOccurences = ["disabled", "ccodecheck"]
+      const allowMultipleOccurences = ["disabled", "ccodecheck" , "knownissue"]
         ## list of flags that are correctly handled when passed multiple times
         ## (instead of being overwritten)
       if key notin allowMultipleOccurences:
@@ -385,7 +386,7 @@ proc parseSpec*(filename: string): TSpec =
         of "netbsd":
           when defined(netbsd): result.err = reDisabled
         else:
-          result.parseErrors.addLine "cannot interpret as a bool: ", e.value
+          result.parseErrors.addLine "cannot interpret as a bool or platform name: ", e.value
       of "cmd":
         if e.value.startsWith("nim "):
           result.cmd = compilerPrefix & e.value[3..^1]
@@ -408,6 +409,12 @@ proc parseSpec*(filename: string): TSpec =
       of "matrix":
         for v in e.value.split(';'):
           result.matrix.add(v.strip)
+      of "knownissue":
+        case e.value.normalize
+        of "n", "no", "false", "0": discard
+        else: 
+            result.knownIssues.add e.value
+            result.err = reDisabled
       else:
         result.parseErrors.addLine "invalid key for test spec: ", e.key
 
--- a/tests/concepts/t3330.nim
+++ b/tests/concepts/t3330.nim
@@ -1,5 +1,5 @@
 discard """
-disabled: "true"
+knownIssue: "https://github.com/nim-works/nimskull/pull/27"
 description: "explain reporting isn't working during the transition to nkError; revise and ressurrect once fixed"
 errormsg: "type mismatch: got <Bar[system.int]>"
 nimout: '''
--- a/tests/concepts/texplain.nim
+++ b/tests/concepts/texplain.nim
@@ -1,6 +1,6 @@
 discard """
   cmd: "nim c --verbosity:0 --colors:off $file"
-  disabled: "true"
+  knownIssue: "https://github.com/nim-works/nimskull/pull/27"
   description: "explain reporting isn't working during the transition to nkError; revise and ressurrect once fixed"
   nimout: '''
 texplain.nim(144, 10) Hint: Non-matching candidates for e(y)
--- a/tests/concepts/twrapconcept.nim
+++ b/tests/concepts/twrapconcept.nim
@@ -1,5 +1,5 @@
 discard """
-  disabled: "true"
+  knownIssue: "https://github.com/nim-works/nimskull/pull/27"
   description: "concept error reporting isn't working during the transition to nkError; revise and ressurrect once fixed"
   errormsg: "type mismatch: got <string>"
   nimout: "twrapconcept.nim(11, 5) Foo: concept predicate failed"
--- a/tests/errmsgs/tundeclared_field.nim
+++ b/tests/errmsgs/tundeclared_field.nim
@@ -1,5 +1,5 @@
 discard """
-disabled: "true"
+knownIssue: "https://github.com/nim-works/nimskull/pull/27"
 description: '''
 disabled this test because the ast is being screwed up and we get gensyms when we shouldn't
 
--- a/tests/misc/tnoinst.nim
+++ b/tests/misc/tnoinst.nim
@@ -1,7 +1,8 @@
 discard """
   errormsg: "instantiate 'notConcrete' explicitly"
   line: 12
-  disabled: "true"
+  knownIssue: "https://github.com/nim-lang/Nim/issues/1708"
+  knownIssue: "https://github.com/nim-lang/Nim/issues/871"
 """
 
 proc wrap[T]() =
--- a/tests/proc/t17157.nim
+++ b/tests/proc/t17157.nim
@@ -1,6 +1,7 @@
 discard """
   errormsg: "'untyped' is only allowed in templates and macros or magic procs"
-  disabled: true
+  knownIssue: "https://github.com/nim-lang/Nim/issues/18113"
+  knownIssue: "https://github.com/nim-lang/Nim/issues/18124"
 """
 
 template something(op: proc (v: untyped): void): void =
--- a/tests/concepts/tswizzle.nim
+++ b/tests/concepts/tswizzle.nim
@@ -3,7 +3,7 @@ discard """
 [1, 3]
 [2, 1, 2]
 '''
-  disabled: "true"
+  knownIssue: "https://github.com/nim-lang/Nim/issues/2346"
 """
 
 import macros, strutils
--- a/tests/cpp/tterminate_handler.nim
+++ b/tests/cpp/tterminate_handler.nim
@@ -2,7 +2,7 @@ discard """
   targets: "cpp"
   outputsub: "Error: unhandled unknown cpp exception"
   exitcode: 1
-  disabled: true
+  knownIssue: "https://github.com/nim-lang/Nim/issues/13695"
 """
 type Crap {.importcpp: "int".} = object
 
--- a/tests/cpp/tvectorseq.nim
+++ b/tests/cpp/tvectorseq.nim
@@ -2,7 +2,7 @@ discard """
   targets: "cpp"
   output: '''(x: 1.0)
 (x: 0.0)'''
-  disabled: "true"
+  knownIssue: "https://github.com/nim-lang/Nim/issues/2536"
 """
 
 # This cannot work yet because we omit type information for importcpp'ed types.
--- a/tests/dll/nimhcr_integration.nim
+++ b/tests/dll/nimhcr_integration.nim
@@ -1,5 +1,7 @@
 discard """
-  disabled: "true"
+  knownIssue: "https://github.com/nim-lang/Nim/pull/12105"
+  knownIssue: "https://github.com/nim-lang/Nim/pull/17311"
+  knownIssue: "https://github.com/nim-lang/Nim/pull/15458"
   output: '''
 main: HELLO!
 main: hasAnyModuleChanged? true
--- a/tests/fragmentation/tfragment_alloc.nim
+++ b/tests/fragmentation/tfragment_alloc.nim
@@ -2,7 +2,9 @@
 discard """
   output: '''occupied ok: true
 total ok: true'''
-  disabled: "true"
+knownIssue: "https://github.com/nim-lang/Nim/issues/8509"
+knownIssue: "https://github.com/nim-lang/Nim/issues/7120"
+knownIssue: "https://github.com/nim-lang/Nim/issues/9421"
 """
 
 import strutils, data
--- a/tests/fragmentation/tfragment_gc.nim
+++ b/tests/fragmentation/tfragment_gc.nim
@@ -1,7 +1,9 @@
 discard """
   output: '''occupied ok: true
 total ok: true'''
-  disabled: "true"
+knownIssue: "https://github.com/nim-lang/Nim/issues/8509"
+knownIssue: "https://github.com/nim-lang/Nim/issues/7120"
+knownIssue: "https://github.com/nim-lang/Nim/issues/9421"
 """
 
 import strutils, data
--- a/tests/iter/tchainediterators.nim
+++ b/tests/iter/tchainediterators.nim
@@ -6,7 +6,7 @@ discard """
 128
 192
 '''
-  disabled: "true"
+  knownIssue: "https://github.com/nim-lang/Nim/issues/1838"
 """
 
 # This all relies on non-documented and questionable features.
--- a/tests/lang/s01_basics/s05_data_types/t4_char.nim
+++ b/tests/lang/s01_basics/s05_data_types/t4_char.nim
@@ -1,6 +1,0 @@
-discard """
-disabled: true
-description: '''
-Placeholder file just in case, covered elsehwere
-'''
-"""
--- a/tests/macros/tmacro7.nim
+++ b/tests/macros/tmacro7.nim
@@ -3,7 +3,7 @@ output: '''
 calling!stuff
 calling!stuff
 '''
-disabled: true
+knownIssue: "https://github.com/nim-lang/Nim/issues/11883"
 """
 
 # this test modifies an already semchecked ast (bad things happen)
--- a/tests/objvariant/tfloatrangeobj.nim
+++ b/tests/objvariant/tfloatrangeobj.nim
@@ -2,7 +2,8 @@ discard """
   output: '''(kind: 2.0, twoStr: "TWO STR")
 (kind: 1.0)
 '''
-disabled: "true"
+knownIssue: "https://github.com/nim-lang/Nim/issues/12379"
+knownIssue: "https://github.com/nim-lang/Nim/issues/12591"
 """
 type
   FloatRange = range[1.0..3.0]


testament/specs.nim
```diff
@@ -350,42 +351,43 @@ proc parseSpec*(filename: string): TSpec =
...
-          when defined(arm64): result.err = reDisabled
-        of "i386":
-          when defined(i386): result.err = reDisabled
-        of "openbsd":
-          when defined(openbsd): result.err = reDisabled
-        of "netbsd":
-          when defined(netbsd): result.err = reDisabled
-        else:
-          result.parseErrors.addLine "cannot interpret as a bool: ", e.value
+        for v in e.value.normalize.split(";"):
```
haxscramper: `disabled` should remain the same in terms of basic syntax - i.e. `windows; linux` *might* be useful, but better addressed separately.

testament/specs.nim
```diff
@@ -350,42 +351,43 @@ proc parseSpec*(filename: string): TSpec =
...
+          case v
+          of "y", "yes", "true", "1", "on": result.err = reDisabled
+          of "n", "no", "false", "0", "off": discard
```
haxscramper: `disabled` should no longer accept simple boolean values, since it is no longer used to just turn tests on and off. It will cause issues with [existing](https://github.com/nim-works/nimskull/search?q=disabled%3A+true) tests, so ideally those should be fixed as a part of the PR.

testament/specs.nim
```diff
@@ -104,6 +104,7 @@ type
                       # but don't rely on much precision
     inlineErrors*: seq[InlineError] # line information to error message
     debugInfo*: string # debug info to give more context
+    knownIssues*: seq[string] # known issues to be fixed
```
haxscramper: Consider converting regular comments into documentation (`##`)

haxscramper (commented):
Thanks for working on this one. Aside from things I commented on, it looks good to me.

haxscramper:
A minor note: please use human-readable PR and issue names, `- issues/61` is hard to make sense of, and your commit has a much better-fit description `added 'knownissue' as key for spec.`. The same applies to the commit message titles.

haxscramper:
A known issue for a specific platform is a tricky problem, because fundamentally, it can be seen as a specific entry (or set of entries) in the build matrix - os, gc, backend and so on. So "disable" and "matrix", "target" provide a specific set of options that narrow down the run matrix entry, and then we mark the set of resulting runs as problematic.

I'm not sure if it really makes sense, but for issue like https://github.com/nim-lang/Nim/issues/17552 expected test would look like

```nim
discard """
knownIssue: "https://github.com/nim-lang/Nim/issues/17552"
matrix: "--gc:orc"
```

If all runs of a known issue pass, it means that issue is fully solved, but individual successful runs should also be noted.

At the same time, we have known issues that are not related to specific matrix configuration (general compilation failure like "Accepts invalid", or "rejects valid"), and those are marked as `knownIssues` without any matrix configurations.

haxscramper:
So to summarize what I wrote above - I'm not sure if there is even a  "case of knownIssue for specific platforms." that need to be handled explicitly. Of course if you had an idea what this should be I would be happy to discuss.

testament/specs.nim
canelhasmateus: How should these be fixed? i'm inclined to add a "disabled" entry for each platform.

canelhasmateus:
There you go.
The syntax for disabled is no longer changed, 
and the other tests, which were `disabled : true ` 
now contain a  `disabled` entry for each platform.

testament/specs.nim
saem: disabled: One per platform makes sense, this states that the test is unsupported on that platform and never will be. So boolean value don't belong anymore. Semicolon separators aren't bad, but a change that we didn't think through and we've got a much bigger overhaul to do down the road. So the existing multiple occurrences works.

known issue: one per reason (usually link) and marks the whole test as not expected to work. And can combine all the mentioned reasons as one big ol' reason on output.

This isn't a perfect design, but the parseCfg parser has limits, and it's workable for now. What we haven't explored with parseCfg is whether we can use sections inside testament. That's ok to come back to afterwards because that entails a more significant overhaul of how we declare tests.

saem: > How should these be fixed? i'm inclined to add a "disabled" entry for each platform.

About half the tests are disabled, for those I would take the following approach:
1. If it's specific to a platform, then nothing to do here
2. If it's true, then note the reason out of git history, eg: I disabled a few because of nkError changes, so these are known issues
3. otherwise if true, then delete because they're silly or flag for discussion by using the known issue field "discuss before merging" 😁

tests/concepts/t3330.nim
```diff
@@ -1,5 +1,17 @@
 discard """
-disabled: "true"
+disabled: "windows"
```
saem: This should be a known issue, broke because of nkError changes.

tests/closure/texplicit_dummy_closure.nim
```diff
@@ -1,5 +1,17 @@
 discard """
-  disabled: true
+  disabled: "windows"
```
saem: Should be deleted, missed as part of async removal.

tests/bind/tdatabind.nim
```diff
@@ -1,5 +1,17 @@
 discard """
-  disabled: true
+  disabled: "windows"
```
saem: Delete test, it's a deprecated module

tests/assign/toverload_asgn2.nim
```diff
@@ -1,7 +1,19 @@
 discard """
   output: '''i value 88
 2aa'''
-  disabled: "true"
+  disabled: "windows"
+  disabled: "linux"
```
saem: Curious if this test works now?

tests/concepts/texplain.nim
```diff
@@ -1,6 +1,18 @@
 discard """
   cmd: "nim c --verbosity:0 --colors:off $file"
-  disabled: "true"
+  disabled: "windows"
```
saem: This should be a known issue, broke because of nkError changes.

tests/concepts/twrapconcept.nim
```diff
@@ -1,5 +1,17 @@
 discard """
-  disabled: "true"
+  disabled: "windows"
```
saem: This should be a known issue, broke because of nkError changes.

tests/errmsgs/tundeclared_field.nim
```diff
@@ -1,5 +1,17 @@
 discard """
-disabled: "true"
+disabled: "windows"
```
saem: This should be a known issue, broke because of nkError changes.

tests/navigator/tincludefile.nim
```diff
@@ -1,5 +1,17 @@
 discard """
-  disabled: true # this test is flakey with testament all
+  disabled: "windows" # this test is flakey with testament all
```
saem: Known issue, using the comment by the disabled true

haxscramper:
> changed boolean 'disabled' keys to reference all platform names.

No, `disabled: true` earlier meant that test is disabled because it is totally broken and/or a known issue. Moving these to reference all platform names is not correct, since it converts the test from "needed to track progress at addressing existing issues" to "disabled on certain platforms due to implementation reasons". 

Ideally `disabled: true` should've been transitioned to the issue link, or textual description of the issue, or `knownIssue: true` at the very least if it is not possible, or hard to track down a specific reason for the test to be disabled.

tests/proc/t17157.nim
```diff
@@ -1,6 +1,18 @@
 discard """
   errormsg: "'untyped' is only allowed in templates and macros or magic procs"
-  disabled: true
+  disabled: "windows"
```
saem: Known issue, should work but not sure how to fix.

tests/stdlib/tjsontestsuite.nim
```diff
@@ -19,19 +31,19 @@ let parsing_testdata = {
   "i_string_incomplete_surrogates_escape_valid": """["\uD800\uD800\n"]""",
   "i_string_inverted_surrogates_U+1D11E": """["\uDd1e\uD834"]""",
   "i_string_lone_second_surrogate": """["\uDFAA"]""",
-  "i_string_not_in_unicode_range": """["�"]""",
-  "i_string_truncated-utf-8": """["��"]""",
+  "i_string_not_in_unicode_range": """["����"]""",
```
saem: Other characters in this file were changed, that shouldn't happen.

saem (changes requested):

saem:
> > changed boolean 'disabled' keys to reference all platform names.
> 
> No, `disabled: true` earlier meant that test is disabled because it is totally broken and/or a known issue. Moving these to reference all platform names is not correct, since it converts the test from "needed to track progress at addressing existing issues" to "disabled on certain platforms due to implementation reasons".
> 
> Ideally `disabled: true` should've been transitioned to the issue link, or textual description of the issue, or `knownIssue: true` at the very least if it is not possible, or hard to track down a specific reason for the test to be disabled.

Have to agree after reviewing, I see two options:
1. Keep disabled true as allowed for now to reduce the number of files changed
2. Convert all disabled true to known issues, but also delete tests that cause problems, eg: there is a test in there mentioning deadlock issues

Even though it's less clean in the long run, I think options 1 will be an easier transition. 😕

testament/specs.nim
saem: This is now out of date, see conversation below.

tests/concepts/texplain.nim
canelhasmateus: does nkError have a github issue?  i found [this one ]( https://github.com/nim-lang/Nim/pull/16915 ) , but is there a original?

canelhasmateus:
i'm down for the second one.  
I believe i'll be able to find the issue for most of these, and will post a update. 
We can them discuss what to do about the rest, if any.

saem:
> i'm down for the second one. I believe i'll be able to find the issue for most of these, and will post a update. We can them discuss what to do about the rest, if any.

If you can do that, awesome!

Otherwise, this would put us in a good logical progression point:
- getting known issue in (done)
- converting the tests that should be know issue instead of disabled that we've noted so far (in comments)
- leave disabled bool in for now (revert that bit of the change)
- don't touch tests we don't understand (revert those few)

tests/concepts/texplain.nim
saem: This is the project: https://github.com/nim-works/nimskull/projects/1#column-16654467

The done column has the pr that broke it. Looking at projects tab https://github.com/nim-works/nimskull/projects you can see a big ol' description of nkError if you scroll down.

canelhasmateus:
Conducted a more in-depth investigation for every test that had a "disabled: true".
Instead of only the ones i didn't find knownIssues for, i decided to jolt down a brief context and the action being done for each one of them, so that we can decide the best course of actions.
Dont mind the formatting, will probably edit it later, if need be.

Mind that some of the issues referenced below are not actually the reason the tests were disabled, 
but issues that the tests is trying to fix / prevent regression.
This pattern seems common enough to warrant reification.

assign

	\toverload_asgn2.nim 
		Disabled just for CI's sake around 2018-04. 
		at the timeline, very near fixes to issues #7637, #6393, altough doesn't seem related to any of them.
		This test doesn't work, but seems fixable. 
		Running it gives the log:
		Expected:
			i value 88
			2aa

			Gotten:
			i value 88
			0aa

		Action: Keeping `disabled: true`

concepts
	
	\t3330.nim
	
		Disabled due to nkError issue.  Due to lack of original issue , i'm using https://github.com/nim-works/nimskull/pull/27.
		Before this, it seemed to be disabled in x32 platforms around 2019-07.
		Last time it was fully functional was around 2019-07.
		First time it was disabled was aroud 2018-10.

		Action: added knownIssue

	\texplain.nim
	
		Disabled due to nkError issue.  Due to lack of original issue , i'm using https://github.com/nim-works/nimskull/pull/27.
		First disabled around 2019-02, only at 32bit platforms, due to flaky message ordering

		Action: added knownIssue

	\twsizzle.nim
		Disabled due to https://github.com/nim-lang/Nim/issues/2346 , around 2015-03
		not sure what this is actually supposed to test.

		Action: keeping `disabled: true`, but looks like a fine removal candidate. 

	\twrapconcept.nim
		Disabled due to nkError issue.  Due to lack of original issue , i'm using https://github.com/nim-works/nimskull/pull/27.
		References https://github.com/nim-lang/Nim/issues/5127

		Action: added knownIssue

cpp
	
	\tdont_init_instantiation.nim

		References  https://github.com/nim-lang/Nim/issues/5140
		Disabled around 2016-12, for CI sake, it seems.


		Action: keeping disabled true.

	\tterminate_handler.nim
		Disabled in 2020-03 , in a commit that references 13695
		References 10765

		Action: added knownIssue 13695

	\tvectorseq.nim
		created disabled, around 2015-04, in a commit that references 2536

		running it results in the following log:
		Expected:
			(x: 1.0)
			(x: 0.0)

			Gotten:
			(x: 1.0)
			(x: 1.0)

		Action: keeping disabled true.

dll

	\nimhcr_integration.nim

		created around 2019-02 in a commit that references 10729
		disabled on macosx around 2019-10 due to flakiness
		disabled on openbsd around 2020-04 in a commit that references 12105
		disabled on netbsd  around 2020-10 in a commit that references 15458
		disabled fully around 2021-03 in a commit that references 17311

		Action: 
effects

	\toutparams.nim
		created disabled around 2020-06 in a commit that references 14521

		the only error in this test is the variable name - possibly an easy fix.
		Expected:
		use explicit initialization of 'x' for clarity [Uninit]

		Gotten:
		use explicit initialization of 'tmp' for clarity [Uninit]

errmsgs

	\tundeclaredfield.nim
		Disabled due to nkError issue.  Due to lack of original issue , i'm using https://github.com/nim-works/nimskull/pull/27.

		Action: Added knownIssue


compiler

	\tdebugutils.nim 
		was born disabled. I guess it is used only for demonstration purposes? 
		from my current understanding, these type of tests should be coalesced into the specs?



fragmentation
	
	\tfragment_alloc.nim and \fragmentation\tfragment_gc.nim
		Seemed to always present flakiness in their lifetimes, being changed several times,  referencing #8509, #7120, #8510 
		Disabled on windows around 2018-03
		Disabled fully around 2018-10, in a commit that references #9421
		
		These two actually passed sucessfully in my pc ( Windows) , but i guess that it is flaky. 
		
		Action: added issues. Maybe re-enable if shown to be not-flaky.


generics

	\toverloading_typedesc.nim
	
		Crreated around 2015-06 in a commit that references #3022
		Disabled around 2017-11 for CI sake?
		Working in my pc.

		Action: Keeping disabled: true . Maybe re-enable.

global
	
	\tglobal.nim
		Disabled around 2015-03, for CI sake?
		Working on my pc.
		Action: Keeping disabled: true.


iter

	\tchainediterators.nim
		Disabled around 2015-01 in a commit that references #1838
		Action: Added known issue.

	\titerable.nim
		Created in 2014-03 in a commit that references #966
		Disabled around 2015-11 citing 'lambda lifting support for iterToProc plugin'		
		Action: keeping disabled: true
	\tscheduler.nim
		Disabled around 2015-07, citing dubiousness ( flakiness? )
		Working in my pc.
		Action: keeping disabled: true


macros

	\tmacro7.nim
		Created around 2019-05 in a commit that references #7792
		Disabled around 2019-08 in a commit that references #11883
		Also, references #7120 and #4547

		Action: Added knownIssue. But probably delete, since it seems to use no-longer-allowed behaviour.


method

	\tmultim.nim and method\tmultimjs.nim 
		Disabled iaround 2020-03 in a commit that references #13803, for CI sake?
		Working in my pc.
		Action: keeping disabled

misc

	\tdlvar.nim
		Created around 2014-04 in a commit that references #1075
		Disabled around 2018-11 for CI sake?
		Action keeping disabled.

	\tgenconstraints.nim
		Created disabled around 2014-04 in a commit that references #1075
		Running gives the output:
			Expected:
			cannot instantiate T2

			Gotten:
			undeclared identifier: 'TNumber'
		Action: keeping disabled. Possibly fixable.
		
	\tmissingnilcheck.nim
		Created disabled around 2014-04 in a commit that references #1075

	\tnoinst.nim
		Created around 2014-04 in a commit that references #1075
		Disabled around 2014-12 in a commit that references #1708, #871
		Running gives the output:
			Expected:
			instantiate 'notConcrete' explicitly

			Gotten:
			internal error: proc has no result symbol

	 \tunsignedcomp.nim
	 	Created around 2015-01
	 	Disabled around 2015-01, not sure why.
	 	Action: keeping disabled.

modules

	\tnotuniquename2.nim
		Disabled around 2019-04 in a commit that references #11064
		Seems to be testing removed behaviour.

		Action: keeping disabled. Probably should be removed.

	\trecmod.nim
		not sure?

navigator

	\tincludefile.nim
		Created around 2021-04 in a commit that references #17784
		Disabled around 2021-07 due to flakiness.

		Action keeping disabled. Probably should be removed.

objects

	\tobjpragma.nim
		Disabled around 2015-03 for CI sake
		There seems to be a ton of issues referencing this:
		https://github.com/nim-lang/Nim/issues/15752
		https://github.com/nim-lang/Nim/issues/9668
		https://github.com/nim-lang/Nim/issues/10082
		https://github.com/nim-lang/Nim/pull/10227
		https://github.com/nim-lang/Nim/issues/4763

		seems to have been fixed on https://github.com/nim-lang/Nim/pull/10227
		
		working on my pc

		Action: keeping disabled. Probably re-enable

objvariant

	\tfloatarangeobj.nim
		Created around 2019-05 in a commit that references #11264
		Disabled around 2019-11 in a commit that references #12379 , #12591
		Action: Added knownIssue

parallel

	\treadafterwrite.nim
		Created disabled around 2014-12 in a commit that references #1597
		Action: keeping disabled.


	\t5000.nim
		Created around 2014-11 in a commit that references #1646
		Disabled around 2014-12.
		Working on my pc, but probably flaky
		what is this even testing, though
		Action: deleted

pragmas

	\tpragmas_misc.nim

		Created around 2020-05 in a commit that references #14443
		Changed in a commit that references #15920
		Disabled around 2021-10 for CI sake

		Action: keeping disabled.

proc

	\t17157.nim
		Created around 2021-03 in a commit that references #17157 and #17531
		Disabled in a commit that references #18113 and #18124
		Action: Added known issues.

stdlib

	\tparsopt.nim

		Created around 2008.
		Disabled around 2018-11 for CI sake
		Action: Keeping disabled. Probably delete.


system

	\t7894.nim
    	        Created around 2018-08 in a commit that references #7894 and #8496
    	        Disabled around 2019-02 for CI sake - references #10608
    	        Seems to run fine
    	        Action: keeping disabled.

	\tio.nim
		Disabled around 2011-12 because of infinite blocking, reenabled shortly after
		Disabled around 2018-11 due to 'just causing problems'
		Action: keeping disabled.


template

	\tmodulealias.nim
		Disabled around 2011-05 in a commit that references #30 and #26.		
		Action: keeping disabled. Probably delete.


typerel

	\tno_gcmem_in_shared.nim
		Disabled around  2014, back when 'shared' was a keyword.
		Action: keeping disabled. Probably delete.

	\trectuple.nim
		Disabled around 2014 in a commit that 'tuple field names are ignored'
		Action: keeping disabled. Probably delete.

	\trectype.nim
		Disabled around 2016-12.
		This seems to correctly validate errors in recursive types, just that the message changed over time.
		Running it gives:
			Expected:
			internal error: cannot generate C type for: PA

			Gotten:
			illegal recursion in type 'PA'

		Action: keeping disabled. Probably fixable.

types

	\tisop.nim
		Disabled around 2014 in a commit that 'tuple field names are ignored'
		Action: keeping disabled. Probably delete.

tests/misc/tgenconstraints.nim
```diff
@@ -2,7 +2,7 @@ discard """
   errormsg: "cannot instantiate T2"
   file: "tgenconstraints.nim"
   line: 25
-  disabled: true
+  #disabled: "true"
```
saem: does this work now? if so we can drop this line all together?

tests/misc/tnoinst.nim
```diff
@@ -1,7 +1,8 @@
 discard """
   errormsg: "instantiate 'notConcrete' explicitly"
   line: 12
-  disabled: "true"
+  #knownIssue: "https://github.com/nim-lang/Nim/issues/1708"
```
saem: why the `#`?

tests/navigator/tincludefile.nim
```diff
@@ -1,5 +1,5 @@
 discard """
-  disabled: true # this test is flakey with testament all
+  disabled: true
```
saem: Let's leave the comment in -- can leave the file untouched

tests/parallel/treadafterwrite.nim
```diff
@@ -1,7 +1,9 @@
 discard """
   errormsg: "'foo' not disjoint from 'foo'"
   line: 23
-  disabled: "true"
+
```
saem: need to be careful with new lines introduced here, it'll throw the line number above off.

saem (changes requested):

canelhasmateus:
For now, i'm keeping the ability to use boolean values in the disabled key.
I changed the disabled to reference an issue, for tests that i could directly tie the disabling commit to an issue.
Mind you that i did not validate that the issue is correct. 

I'll be opening  a second PR for the remaining 'boolean disabled' tests, so that we can deal with them.

saem:
Heck yeah!

Please rebase the commits and include your excellent analysis on a test by test basis as part of it.

saem (approved):

saem:
@haxscramper I'm g2g for this PR, how ya feeling about it?

canelhasmateus:
re-structured the commit message. This should be good to go.

haxscramper:
It looks good to me, I think I will PR follow-up documentation update to reiterate on the explanations from this PR, and some other thoughts that I have (from #59 where I started creating new tests with `knownIssue`)

saem:
Bors r+

PR #69, opened 2021-11-20
make cancelled CI a failure

alaviss:
#65 managed to skip through bors by having its CI cancelled. This is not optimal and has to be fixed.


saem:
nice

PR #70, opened 2021-11-21
nkerror remove legacy lookups and concept fixes

saem:
replace the legacy `qualifiedLookUp`  with the implementation from `qualifiedLookUp2`.

Also fixed some issues with sigmatch/semcall such that concept error reporting is almost back to normal.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -406,137 +406,7 @@ proc presentFailedCandidates(
     conf: ConfigRef,
     n: PNode,
     errors: seq[SemCallMismatch]
-  ): (TPreferedDesc, string) =
-  ## Format failed candidates for call overload resolution failure
-
-  var prefer = preferName
-  # to avoid confusing errors like:
-  #   got (SslPtr, SocketHandle)
-  #   but expected one of:
-  #   openssl.SSL_set_fd(ssl: SslPtr, fd: SocketHandle): cint
-  # we do a pre-analysis. If all types produce the same string, we will add
-  # module information.
-  let proto = describeArgs(conf, n, 1, preferName)
-  for err in errors:
-    var errProto = ""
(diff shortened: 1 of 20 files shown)


compiler/front/cli_reporter.nim
```diff
@@ -477,7 +477,8 @@ proc presentFailedCandidates(
         maybeWrongSpace = true
 
     for diag in err.diagnostics:
-      candidates.add reportFull(conf, diag) & "\n" # XXX: Tests fail because this is "Error: ..." and not "<Typename>: ..."
+      # Replace the "Error: " in the error messages with the name of the concept that the diagnostics concern
+      candidates.add reportFull(conf, diag).replace("Error:", err.diagnosticsTarget.name.s & ':') & '\n'
```
haxscramper: Use `reportBody` and add missing parts of the message (via `conf.toStr(r.location.get())`) instead of replacing parts of the fully generated message report.

compiler/sem/semexprs.nim
```diff
@@ -266,8 +266,16 @@ proc maybeLiftType(t: var PType, c: PContext, info: TLineInfo) =
   if lifted != nil: t = lifted
 
 proc isOwnedSym(c: PContext; n: PNode): bool =
+  ## checks to see if the `n`ode is the "owned" sym node
   let s = qualifiedLookUp(c, n, {})
-  result = s != nil and sfSystemModule in s.owner.flags and s.name.s == "owned"
+  result =
+    if s.isError:
+      # XXX: move to propagating nkError, skError, and tyError
+      for e in walkErrors(c.config, s.ast):
```
saem: Pretty sure local report does a traversal now, if that's the case we can simply drop the loop from all these places.

saem (commented):
LGTM

Bors r+

PR #73, opened 2021-11-22
[DOC] Clean up speech tics

haxscramper:
Removed the "ditto" speech tics from some comments.

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -3034,7 +3034,7 @@ Example:
     echo "permission denied"
   of "go-for-a-walk":     echo "please yourself"
   elif line.len == 0:     echo "empty" # optional, must come after `of` branches
-  else:                   echo "unknown command" # ditto
+  else:                   echo "unknown command" # optional, must come after `of` and `elif` branches
 
   # indentation of the branches is also allowed; and so is an optional colon
   # after the selecting expression:
--- a/lib/pure/random.nim
+++ b/lib/pure/random.nim
@@ -145,7 +145,7 @@ proc next*(r: var Rand): uint64 =
   runnableExamples("-r:off"):
     var r = initRand(2019)
     assert r.next() == 13223559681708962501'u64 # implementation defined
-    assert r.next() == 7229677234260823147'u64 # ditto
+    assert r.next() == 7229677234260823147'u64 # implementation defined
 
   let s0 = r.a0
   var s1 = r.a1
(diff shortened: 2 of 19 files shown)


lib/pure/random.nim
```diff
@@ -145,7 +145,7 @@ proc next*(r: var Rand): uint64 =
   runnableExamples("-r:off"):
     var r = initRand(2019)
     assert r.next() == 13223559681708962501'u64 # implementation defined
-    assert r.next() == 7229677234260823147'u64 # ditto
+    assert r.next() == 7229677234260823147'u64 # implementation defined
```
haxscramper, line 148: Found a funny "implementation defined" annotation that is probably going to be a huge pain to define properly.

haxscramper:
Remote the "ditto" speech tics from some comments.

lib/pure/random.nim
saem: LoL, well that's fun.

doc/manual.rst
```diff
@@ -3034,7 +3034,7 @@ Example:
     echo "permission denied"
   of "go-for-a-walk":     echo "please yourself"
   elif line.len == 0:     echo "empty" # optional, must come after `of` branches
-  else:                   echo "unknown command" # ditto
+  else:                   echo "unknown command"
```
```suggestion
  else:                   echo "unknown command" # optional, must come after `of` and `elif` branches
```

tests/ccgbugs/tsamename3.nim
```diff
@@ -8,7 +8,7 @@ block: # bug #15526
       x2: int
     let f2 = Foo(x2: 2)
 
-block: # ditto
+block:
```
saem:
```suggestion
block: # bug #15526
```

It's this perhaps what was meant?

saem (commented):
I think the ditto remarks might need some replication, with adjustment, some comments.

nimdoc/testproject/expected/subdir/subdir_b/utils.html
```diff
@@ -193,10 +193,10 @@ <h1 class="title">subdir/subdir_b/utils</h1>
 </ul>
 
   </div>
-  
+
```
saem:
```suggestion
  
```

nimdoc/testproject/expected/subdir/subdir_b/utils.html
```diff
@@ -193,10 +193,10 @@ <h1 class="title">subdir/subdir_b/utils</h1>
...
   <div class="nine columns" id="content">
   <div id="tocRoot"></div>
-  
+
```
saem:
```suggestion
  
```

saem (changes requested):
re-added some spaces because these stupid tests are sensitive to this crap.

we can figure out how to solve that some other day, hopefully the suggestions should fix it for now.

PR #74, opened 2021-11-23
Nonref object inheritance hints and converts to base in varargs a…

beef331:
…nd on assignment

This should make it so non ref objects can be used as one expected. `varargs[Base]` can accept `[Base, Derived, Base, Derived2]` and so forth. It also makes it so any object conversion up or down(Though down conversion is explicit so no clue why it checks) emits a hint to inform about loss of possible information. `var a: Base = Derived(field2: 300)` emits `'Derived(field2: 300)' implicitly converted to 'Base' [/home/jason/nimskull/bin/abuse.nim(26, 22)] [ImplicitObjConv]`


saem:
Waiting for #59 in order check against the spec semantics.

saem:
@beef331 mind rebasing this, I'm with you on the hint. They're all value types.

haxscramper:
Ideally this should also include spec-like test for https://github.com/nim-works/nimskull/pull/59#issuecomment-976260222 that checks that `f(a, b, c)` works and does implicit conversion (for different usage scenarios), and `f([a, b, c])` does not work, because we construct an array literal, and those should not do the implicit type conversion on construction.

Latter part can be revised later (check array literal differently and/or perfrom implicit conversions), but for now this should be made into compiler error (if it isn't already).

Same applies to the `f(@[a, b, c])`, although technically this does not need to be tested, since `@[]` is already a function application on the array literal, but seq construction is magic, so better check this one as well.

tests/types/tinheritance_arrays.nim
```diff
@@ -0,0 +1,53 @@
+discard """
```
saem, line 1: @beef331 is the intention with these tests to fold into the spec?

beef331: It probably should be moved to the spec I guess since it's for regression assurance.

saem: OK, if you know how to do it now then go for it, otherwise ask @haxscramper when he's about and do it later?

(we can merge this in the meantime)

PR #78, opened 2021-11-25
runs PR requirement CI on conversion to/from draft

alaviss:
Otherwise it would be possible to bypass this requirement by starting
off as a draft then switch to "ready for review".


alaviss:
Look like it works

bors r+

PR #85, opened 2021-11-29
Improve koch reliability and bootstrap

alaviss:
This PR focuses on dealing with friction and reliability issues related to `koch` and its subcommands.

One of the main change is the introduction of a new concept: bootstrapping compiler.

The bootstrapping compiler job is to build `koch` and the first iteration of the bootstrap process. This compiler will be pinned so that the build process always have a constant starting point, no matter how much you messed up the main compiler in prior steps.

Right now this compiler is `csources`, but this might be changed in the future should `csources` become no longer feasible to maintain.

Implementing this concept is the new `koch.py` script. This serves as the launcher to bootstrap `koch.nim` automatically as well as managing the bootstrapping compiler. The script is in Python due to its excellent cross-platform support as well as widespread availability.

---

Status:

- [x] Reduces friction
  - [x] Make sure `koch` can be run in as little steps as possible: `koch.py` is introduced to handle bootstrapping of the native version.
  - [x] Eliminate reliance on the built compiler in `PATH` by `koch.nim`.

- [x] Increase reliability
  - [x] Make sure bootstrapping is reproducible: The bootstrapping compiler is versioned and tracked by `koch.py`, ensuring a consistent path to bootstrapping.
  - [x] Make sure that the native `koch.nim` is always up-to-date: This is done with `koch.py`.

- [x] Clean up
  - [x] Consolidate `koch` sources into one location.


koch.sh
```diff
@@ -0,0 +1,145 @@
...
+bootstrapSource="$baseDirectory/build/csources"
+
+# The nimcache folder.
+nimcache="$baseDirectory/nimcache"
+
+# Build config
+buildConfig="$baseDirectory/config/build_config.txt"
+
+# Get number of processing cores. Returns 1 if this number can not be retrieved.
+get_num_cores() {
```
saem: I thought the typical recommendation was for `min(RAM/2GB, 2 x CPU-threads)`? With the former winning over the latter if lazy.

koch.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/sh
```
saem: Not saying is a good idea, but the only way I see not having this file without being driven mad is have Koch check something on startup, then either rebuild the real Koch or use the real Koch to do the work... so a proxy in which case we're basically back to this script, except in nimskull, and only difference is because we don't need a batch file for Windows?

I'm not sure that's all that great an idea, I'd lean script because that can come later of it becomes a pain.

koch.sh
alaviss: CPU threads is several order of magnitude easier to get (in a cross-platform manner) :P

It keeps things simple and that's what most builder uses nowadays.

saem:
I won't lie the thought of injecting python in the middle of this is off putting. I absolutely despise dealing with python, it's versions, libraries, and whatever other crap. I will go into a full rage if I even have to consider a pip install, dnf anything, or whatever to make it work.

If it works I'm not going to complain, I use Fedora and system python moves fast so that's the only thing that's remotely concerning, so as long as that won't be an issue I'm totally ok with Python in here.

As for the direction, I think it makes sense. I've been bitten by an older koch, forgetting to test bootstrap, and generally speaking the whole virtuous cycle of bootstrapping needs to be maintained. This is a mechanism that does exactly that.

koch.py
```diff
@@ -0,0 +1,234 @@
...
+import os
+import shutil
+import subprocess
+import sys
+
+# Nim source directory, which is the directory this script resides in
+NimSource: Final = Path(__file__).resolve(strict=True).parent
+
+# The nimcache folder
+Nimcache: Final = NimSource / "nimcache"
```
saem: Not a deal beaker; I think people will want to configure this. I don't think we need to do anything yet.

Mostly mentioning this because I'm trying to think where the line is in terms of how much we put in here vs not.

alaviss:
> I won't lie the thought of injecting python in the middle of this is off putting. I absolutely despise dealing with python, it's versions, libraries, and whatever other crap. I will go into a full rage if I even have to consider a pip install, dnf anything, or whatever to make it work.

Don't worry, I'll be sure to make it run with just the stdlib. Though I might have to scale back the python version of this if its become an issue (mainly removing type annotations). So far CI is happy, though.

koch.py
alaviss: I think the line is whatever that we need before `koch.nim` can be run.

shayanhabibi:
I'll be able to test on windows on Monday (when im home). If anyone has a windows machine and is able to test it before then that'd be appreciated.

disruptek:
Seems fine to me.  It doesn't have to be the last word in bootstrapping; merely the first.  Good job.

disruptek (approved):

saem (changes requested):
I think we need to update the docs:
- [x] koch section in the [readme](https://github.com/nim-works/nimskull#koch)
- [x] [internals](https://nim-works.github.io/nimskull/intern.html)
- [x] [contributing](https://nim-works.github.io/nimskull/contributing.html)

saem (approved):
yay, the docs are becoming better too!

I'm cleaning them up generally right now so this all works together nicely.

PR #86, opened 2021-11-30
remove unused "cycle.h" header.

alaviss:
This seems to be very old, but there are no references within Nim to this
header, so I suspect that this has long been obsolete.

This commit remove this header file.


alaviss:
Tests passed, so I don't think it's much harm to merge this

bors r+

PR #88, opened 2021-12-01
[WIP] Make Arc decrefs and increfs atomic

shayanhabibi:
As per the discussion on shared refs and arc, this fix will prevent any undefined behavior occuring.

The idea is to:

- [x] Have all atomic operations on ref count changes when threads:on

- [ ] Make atomic operations opt-out using a Owned[T] type

- [ ] Include a one file test that demonstrates the behavior before/after


alaviss (commented):
I think we should not go with relaxed atomics until we can verify that everything works as expected.

shayanhabibi:
The dec and incs must be ordered correctly; the load can be relaxed as it will not be able to pass any dec/incs due to their stricter ordering. The subsequent atomicdec being ordered prevents the load moving forward. Since the load is used as the condition for entering the block, should be aiight. Else I'll change it to acqrel

alaviss:
> The subsequent atomicdec being ordered prevents the load moving forward.

I guess you haven't noticed this:

https://github.com/nim-works/nimskull/blob/989adc65731d01b29ec34520d17ffaa233e907f0/lib/system/atomics.nim#L228-L233

shayanhabibi:
F*** 🤣 I assumed seqcst is the default!

Hahaha I made an ass out of myself. Thank you bb ❤️ I'll fix it up

lib/system.nim
```diff
@@ -1805,13 +1805,14 @@ when not declared(sysFatal):
...
-    discardable, benign.}
-    ## Atomic decrement of `memLoc`. Returns the value after the operation.
+  
+  # proc atomicInc*(memLoc: var int, x: int = 1; order: AtomMemModel = ATOMIC_RELAXED): int {.inline,
+  #   discardable, benign.}
+  #   ## Atomic increment of `memLoc`. Returns the value after the operation.
+
+  # proc atomicDec*(memLoc: var int, x: int = 1; order: AtomMemModel = ATOMIC_RELAXED): int {.inline,
+  #   discardable, benign.}
+  #   ## Atomic decrement of `memLoc`. Returns the value after the operation.
```
disruptek, line 1815: Prefer `when false:` and document why these are removed.

lib/system/arc.nim
```diff
@@ -116,15 +116,30 @@ proc nimNewObjUninit(size, alignment: int): pointer {.compilerRtl.} =
     cprintf("[Allocated] %p result: %p\n", result -! sizeof(RefHeader), result)
 
 proc nimDecWeakRef(p: pointer) {.compilerRtl, inl.} =
-  dec head(p).rc, rcIncrement
+  # We want to use atomic operations when threading is on
```
disruptek, line 119: Add a period.

lib/system/arc.nim
```diff
@@ -116,15 +116,30 @@ proc nimNewObjUninit(size, alignment: int): pointer {.compilerRtl.} =
...
+  # We will only use non-atomic operations if user 'owns' the ref - which implies
+  # that no other thread is handling the ref anymore.
+  template decOpr(x,y: untyped): untyped =
+    when hasThreadSupport:
+      atomicDec(x, y, ATOMIC_ACQ_REL)
+    else:
+      dec(x, y)
```
disruptek, line 126: Move it out and reuse it below.  Add a comment or more descriptive name.  Why not use `typed`?

lib/system/arc.nim
```diff
@@ -187,20 +202,31 @@ when defined(gcOrc):
...
+    when hasThreadSupport:
+      atomicLoadN(x.addr, ATOMIC_ACQUIRE)
+    else:
+      x
+  template decOpr(x, y: untyped): untyped =
+    when hasThreadSupport:
+      atomicDec(x, y, ATOMIC_RELEASE)
+    else:
+      dec(x, y)
+
```
disruptek, line 215: Comment or rename to be more descriptive.

disruptek (changes requested):
If you're going to touch every line in `atomics` why not improve it -- do we need to be exporting a distinct cint for AtomMemModel or can we make the API idiomatic, with an enum?  These variables and constants are code smell; you can hide them in here with a converter if you're lazy.  Also not a fan of the `barrier` template.  Yikes.

disruptek:
Also, I would favor a single `when hasThreadSupport:` where you define two sets of templates; you can document the differences there and explain why they exist instead of scattering the conditionals in templates throughout the code.

shayanhabibi:
> If you're going to touch every line in `atomics` why not improve it -- do we need to be exporting a distinct cint for AtomMemModel or can we make the API idiomatic, with an enum? These variables and constants are code smell; you can hide them in here with a converter if you're lazy. Also not a fan of the `barrier` template. Yikes.

Bloody CLRF.

PR #90, opened 2021-12-02
bors: cut PR description at the first thematic break

alaviss:
After the thematic break, the information typically stops being related
to the PR itself (for example, dependabot command lists). Thus cut the
body there to avoid introducing junk into the merge commit.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -7,3 +7,7 @@ status = [
 pr_status = [
   "PR has only one commit"
 ]
+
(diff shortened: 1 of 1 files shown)


saem (approved):
Should add a tip to the pr template?

Can do it separately, so I left it as approved.

disruptek:
Seems like it only really applies to the bot PRs at this point, so whatever...

disruptek (approved):

PR #94, opened 2021-12-04
Structured reports

haxscramper:
Full rework of the compiler message handling pipeline. Remove old-style message generation that was based purely on strings that were formatted in-place, and instead implement structured logging where main compiler code only instantiates objects with required information.

This would allow to
- decouple error presentation logic from the semantic checking and other parts of the compiler. Added bonus: user-facing error messages can be improved much faster, since changes don't have to be propagated through the test suite each time.
- allow compilation reports to be printed out as a stream of S-expression or JSON lines
- Added bonus: user-facing error messages can be improved much faster, since changes don't have to be propagated through the test suite each time.

In addition to architectural improvements this PR also reorganizes and documents various error kinds. Whole thing has to be rewritten from the ground up, so RFCs like 
https://github.com/nim-lang/RFCs/issues/323 https://github.com/nim-lang/RFCs/issues/324 https://github.com/nim-lang/RFCs/issues/325 will be implemented as a collateral

Decoupling of the data and presentation also allows to properly test compilation errors, warnings and hints generated by the compiler and finally untie compiler unit tests from the hardcoded string formatting in the semantic checking phase.

This PR is an orthogonal to the `nkError` project, and is only concerned with a *content* of the error report, while `nkError` is all about *presence* of errors during compilation.


compiler/reports.nim
```diff
@@ -0,0 +1,108 @@
+## This module provides type definitions for all structured report entries
+## that compiler can provide.
+##
+## Note that this module specifically does not import anything else from
```
haxscramper, line 4: There are two ways this can be handled - either by getting all the necessary information when we need it and populating various string fields (for type names, arguments and so on), effectively replicating a section of the compiler data types, or by importing ast.nim and adding it to the reports. First approach would make it easier to write out full report (in json), and generate pretty-printed version of it (no need to pull in ModuleGraph and ConfigRef to be able to print things out), but it will require some duplication.

haxscramper: I will try to see if I can find a good example of how one report can be generated with these two approaches and then figure out which one is more promising.

compiler/reports.nim
```diff
@@ -0,0 +1,108 @@
...
+    repCmd ## Report related to execution of the external command - start
+           ## of the command, execution failure, succes and so on.
+
+    repDebug ## Side channel for the compiler debug report. Sem expansion
+             ## traces and other helper messages designed specifically to
+             ## aid development of the compiler
+
+
+type
+  ReportLineInfo* = object
```
saem, line 966: Just because of reasons we're going to bias towards report points, line and column pair, but do we want to support the option to have a report span, so we have a start and end line and column pairs?

This would be in a variant because sometimes points still make sense.

haxscramper: I see no problem supporting report span as well, although right now I can't recall any existing compiler output that could require span.

compiler/reports.nim
haxscramper: Main use case for the import-ast solution is reporting of the overload resolution errors - "but expression has type", known proc signatures.

compiler/reports.nim
haxscramper: Maybe "but expression has type" section for the overload resolution reports?

compiler/reports.nim
haxscramper: `msgs.writeContex()` (and some other procedures) report information in form of `"XXX of <name of the procedure/template/macro>`, maybe with `defined in <location>`. Report might contain type like 

```nim
SemReportEntry = object
  name*: string
  kind*: enum = sreProc, sreMacro, sreField, sreXXX ...
  definedIn*: ReportLineInfo
  sigDescription*: <object that encodes whole description of the used entry signature>
                   <for example for procs it would be an arg/return pragmas and so on>
```

If ast is used we would need to reconstruct the whole "sig description" report using procs that traverse the AST and get necessary information for pretty printing, and this kind of thing would be very annoying to serialize properly.

haxscramper: I think it is better to decouple these sections of the compiler at the cost of some repetition, since we would still have to do *exactly the same* thing for pretty-printing or serialization, so why not prepare the data fully beforehand.

haxscramper: `m*: MsgConfig` from `ConfigRef` is populated via procs like 

```nim
proc pushInfoContext*(conf: ConfigRef; info: TLineInfo; detail: string = "") =
  conf.m.msgContext.add((info, detail))
```

which in turn are populated via `sym.name` (sorry, `template detailedInfo*(sym: PSym): string = sym.name.s`, so we would need to push syms into message context for the config anyway, or implement proc that converts syms to

haxscramper:
```nim
  SemReportEntryKind* = enum
    srekProc
    srekTemplate
    srekMacro
    srekMethod
    srekConverter

    srekEnum
    srekObject
    srekAlias
    srekTypeclass

  SemReportEntry* = object
    ## Entry mentioned in the sem report - type or procedure definition,
    ## macros, template or any similar construct. Used in reports like
    ## overload failures, "template/generic instantiation of" and such, to
    ## store information about entry referred to.
    declaredIn*: ReportLineInfo ## Location of the entry declaration
    kind*: SemReportEntryKind
```

Pushed in the b395e02

haxscramper: Is it safe (as in - syms are not invalidated at some point?) to push syms into configRef `m*: MsgConfig` (`msgContext*: seq[tuple[info: TLineInfo, detail: string]]`) and then later unpack things from there as needed and convert to the `SemReportEntry` instead of doing expensive conversion each time?

compiler/reports.nim
haxscramper: Made it into a combination of a separate types for point and range that are combined in a single variant - this way cases where explicit range or point is necessary can still be handled correctly (without creating two different alternatives).

haxscramper: The only report event I can recall right now, one that would actually require spans, is a `"but expression <XXX> has a type"` (ideally the report should contain both expanded an unexpanded form, so people could make sense of this mess - `mapIt().filterIt()` generates absolutely unreadable garbage mountain when used)

compiler/reports.nim
saem: Modularity and data oriented go hand in hand I find. Basically the nkError stuff bright back a whole bunch of old lessons for me confined with a bunch of other old lessons from seemingly unrelated experience. I'll speak generally first, then try and go more specific.

----

### general

Generally speaking I think module decomposition should work like so:
1. Write a module that covers a while topic or responsibility and call it good

Once the dependencies start arising out they're there from the get go, because existing code the fuller decomposition of modules is more like this:
- data module: fundamental data for the topic, optimized for write or reporting of that's much more expensive
- command or write module: contains the constructors, necessary or minimal read functionality for queries required for write decision making, traversal, etc; all the sort of create commands
- effects or reporting or read module: all the heavy reads and consequent actions, uses the above module to traverse and decide where necessary; might produce secondary or additional data structures related to its bookkeeping

The various dependencies force a separation like this anyhow because otherwise you end up with recursive import issues.

So now back to this, the current AST isn't data oriented, but also not entirely the worst, and this has a mild cascade effect that it as some chonky boys like ConfigRef and friends to come along for the ride. The abstractions to work around this aren't great either.

----

### specific

I'm always more partial to fundamental fixes, which I think would begin with wrapping all node construction with a config aware constructor. That would let us use the config as a context which will enable making the whole AST data oriented. Locating all the construction sites might not be too bad, if we can leverage error or deprecation pragma or some other trickery.

If the above feels like a mountain then I would head the data duplication route, but make this data oriented so we can more easily converge later, somehow.

haxscramper: I can't exactly go into fundamental fixes there, because it would create this crooked intermediate state where I spend just as much time making work continiously instead of burning this whole contraption to the ground. Example - "template/generic instantiation of" is built on top of the construction that essentially appears to be a stack of pairs that are push-pop edited via `msgs.pushInfoContext` and `msgs.popInfoContext`. This stack contains strings, or "detail" information, which is generated immediately in various places. So I need to get rid of this string as well, otherwise I'm unable to even construct `SemContext` object meaningfully. I could've just made it into a string and passed it along, but this feels like an absolute hack that I would have to remove anyway.

haxscramper: > but make this data oriented so we can more easily converge later, somehow. 

Can you elaborate on this one specifically - does my approach seem like a move in a right direction for that, or I need to change something? Because right now I just want to go in the "Create an object that thoroughly describes the report" route. Later this object is simply converter to a string (for pretty-printing) or json (for structured reporting)

saem: > Is it safe (as in - syms are not invalidated at some point?) to push syms into configRef

Yes, I believe this is safe, the only time things get a bit weird is that we could be in metaprogramming land, like a compilation inside a compilation, and so the inner stuff might be throw away. If we hang onto a ref of the symbols from the inner speculative compilation then we will have potential space leak (that the right term?).

saem: > Can you elaborate on this one specifically - does my approach seem like a move in a right direction for that, or I need to change something? Because right now I just want to go in the "Create an object that thoroughly describes the report" route. Later this object is simply converter to a string (for pretty-printing) or json (for structured reporting)

From a strict performance perspective it isn't. I don't care about that and it totally misses the primary benefit of data oriented design.

I'm looking at the data and seeing about normalizing it, mostly. You've got no refs which is a good sign. The only thing that stands out is the inheritance, which isn't always a non-starter, and I'd consider the look aside arrays/tables approach, but I'm not a purist. I see it potentially needing to change but I'll admit this might be easier to read.

For me I'd say yes for now, but watch out if you feel the urge to add refs in the middle of the data.

haxscramper:
I'm not exactly sure if separation of report kinds by topics is the right approach - it makes creation and processing more meaningful, but that has to deal with setting/unsetting various hints will be somewhat more ugly. Previously, absolutely all compilation messages were stored in the `TMsgKinds`, including things that should probably be a configuration options (like `hintSource`). `errGenerated` is a complete disaster - it has to be split in ~80 separate enum values that actually make report contextually valuable. With current approach, each compiler subsystem has its own `XXXReport` type - `SemReport`, `LexerReport`, `BackendReport` and so on., which makes it easier to add more specific fields, especially compared to a single object with single enum type that ranges over all pipeline elements at once. But if I want to set the hint from the string (via `--hint[]` flag) I don't really have a distinct mapping - `"HintLineTooLong"` belongs to a `LexerReportKind`, but `"HintPattern"` is a sem layer (`hlo.nim` pass when TRM macros are expanded).




----

UPD: this problem was solved by using a single enum and defining object variants for each category. Then adding helper asserts for the valid ranges of each report category. Ideally this could've been fixed by using `range[CatMin .. CatHigh]` as a discriminant, but this is currently not allowed because discriminant must have a low() equal to zero.

haxscramper:
As per our earlier discussion, I think that `nkError` node should store a `Report` object, although not itself, but rather index into the table of reports, since doing otherwise would either needlessly increase the size of the `PNode`, or use `ref Report` which is not in the "middle" of the data, is on the very top. It still seems like `.sons` field must be accessible in the `nkError`, so maybe something like makes sense?

```nim
    else:
      report*: ref Report ## Report associated with a node (can (in most cases is) be nil)
      sons*: TNodeSeq
```

or

```nim
    else:
      report*: ReportId ## Id of the report, the real list is in the `.msgContext` field in
                        ## the config.
      sons*: TNodeSeq
```

compiler/reports.nim
```diff
@@ -115,9 +287,7 @@ func severity*(rep: LexerReport): ReportSeverity =
     else: rsevTrace
 
 type
-  ParserReportKind* = enum
-    rparName = "Name" ## Linter report about used identifier
-
+  ParserReportKind* = range[rparInvalidIndentation .. rparName]
   ParserReport* = object of ReportBase
     kind*: ParserReportKind
```
haxscramper: This requires disabling check for the nonzero discriminant value. In short, I want to try and use this pattern for reports. 

```nim
type
  E = enum a, b, c, d
  E2 = range[c .. d]
  Ob = object
    case k*: E2
      of c: discard
      of d: discard
```

Having multiple enums for different kinds of reports is not sustainable, too much friction. Having a single enum that is split into ranges makes things a lot easier. 

An alternative would be to merge all report kinds into a single variant object, but then I'm afraid we will simply run out of meaningful field names, or will end up with another `TSym`-like object that contains absolutely everything mixed into one pot, with no separation whatsoever.

haxscramper: Fixing this requires

```nim
  if firstOrd(c.config, typ) != 0:
    localError(c.config, n.info, "low(" & $a[0].sym.name.s &
                                     ") must be 0 for discriminant")
```

Check to be removed and subsequently allows for negative values to be used as discriminant as well. I personally would much rather prefer to have a proper `default()` implementation and temporarily weaken the check before we fix things for real.

haxscramper: https://github.com/nim-works/nimskull/pull/99

haxscramper:
<div align="center">

image

</div>

Brought to you by

```nim

const
  errInvalidCmdLineOption = "invalid command line option: '$1'"
  errOnOrOffExpectedButXFound = "'on' or 'off' expected, but '$1' found"
  errOnOffOrListExpectedButXFound = "'on', 'off' or 'list' expected, but '$1' found"
  errOffHintsError = "'off', 'hint' or 'error' expected, but '$1' found"

const
  errNoneBoehmRefcExpectedButXFound = "'arc', 'orc', 'markAndSweep', 'boehm', 'go', 'none', 'regions', or 'refc' expected, but '$1' found"
  errNoneSpeedOrSizeExpectedButXFound = "'none', 'speed' or 'size' expected, but '$1' found"
  errGuiConsoleOrLibExpectedButXFound = "'gui', 'console' or 'lib' expected, but '$1' found"
  errInvalidExceptionSystem = "'goto', 'setjump', 'cpp' or 'quirky' expected, but '$1' found"
```

haxscramper:
This would also allow ensuring that the compiler is exhaustively tested for all possible user-facing reports, previously it would've been a test to one's sanity (and continuous maintenance burden that would make changing error messages much harder)

haxscramper:
And ensure everything is documented, what tf is `"union type may not have an object header"` is even responsible for? It was added here, https://github.com/nim-works/nimskull/commit/21fc8b4d4d82637fab831cffed0e5b995ec7f16c of course ctrlf for this commit includes *one* occurence, so my assumption is that this **supposedly user-facing** (emitted via `localError`, not `internalError`) might potentially bubble up for users to see (although right now I'm not even sure what "object header" is exactly - I know `object`, I know `{.header:` - is that `object` **with** header? Or some 'header' part of an object type?). 

------

UPD: yes, "object header" is not a `{.header`, it is an implicit chunk of data that is added from the inheritance (maybe in some other contexts as well)

```nim
type
  A {.union.} = object of RootObj
    field: int
    field2: float
```

Crude search for tests against this error `rg 'object\sheader'` shows it is *not* tested (with current test architecture, I would've found `errormsg: "union type may not have an object header"` or something similar.

```nim
lib/system/alloc.nim
968:    sysAssert(cast[ptr FreeCell](x).zeroField == 1, "dealloc: object header corrupted")

lib/system/arc.nim
13:at offset 0 then. The ``ref`` object header is independent from the
51:    rc: int # the object header is now a single RC field.

compiler/sizealignoffsetimpl.nim
379:          localError(conf, info, "union type may not have an object header")

tests/misc/tsizeof.nim
326:      # only there the object header is smaller than the first member
```

haxscramper:
When I stated this PR, I thought of it as a detour that I might take before returning to the https://github.com/nim-works/nimskull/projects/2 PR, but now I understand it is a hard requirement to get any meaningful progress on the part of the specification that are concerned with errors, warnings and hints.

haxscramper:
I'm not sure if I have the capacity to clean up the reporting *and* think about global implications of various errors, so this part need to be revised separately as well - I'm adding tons of enums for different errors and diagnostics, but I'm pretty sure at least some of them can be generalized after careful inspection. Current implementation should still be considered a proof-of-concept wrt. to general *design* of the reporting system.

I will return to this when I get to the actual formatting part, which hopefully will happen soon enough. There are ~160 files in the `compiler/` folder, I've progressed through ~30 of them, so I'm around 20% through with basic changes on this PR (file-wise, not SLOC-wise, that part will be much smaller - looking at the current trend of code edits, I would say that most of the remaining stuff is dumb edits like

```diff
-            localError(g.config, s.info,
-              "{.intdefine.} const was set to an invalid integer: '" &
-                g.config.symbols[s.name.s] & "'")
+            g.config.localError SemReport(
+              kind: rsemInvalidIntdefine, expressionStr: g.config.symbols[s.name.s])
```

saem:
Caught up on this PR after a while, it's definitely headed in the right direction. The approach of focusing on the necessary mechanical cleanup of the callsites is looking quite promising.

juancarlospaco:
> 
> ```nim
> const
>   errInvalidCmdLineOption = "invalid command line option: '$1'"
>   errOnOrOffExpectedButXFound = "'on' or 'off' expected, but '$1' found"
>   errOnOffOrListExpectedButXFound = "'on', 'off' or 'list' expected, but '$1' found"
>   errOffHintsError = "'off', 'hint' or 'error' expected, but '$1' found"
> ```

I think if you gonna do it this way, for single formatted arguments,
work with the wording in a way that allows you to skip the formatting string ops at all, kinda like:

```nim
const someErr = "Invalid Foo in your Bar: "

echo someErr & wrongArgument
```

haxscramper:
@juancarlospaco more like this. 

```nim
proc report(rep: ExternallReport) = 
  case rep.kind:
    of rextUnexpectedValue:
      echo(
        "Unexpected value for ",
        $rep.cmdlineSwitch,
        " expected one of ",
        $rep.cmdlineAllowed,
        ", but found ",
        $rep.cmdlineProvided
      )

    of ...:
     ...
```

or this

```nim
proc report(rep: ExternallReport) = 
  case rep.kind:
    of rextUnexpectedValue:
      echo "Unexpected value for $1. Expected one of $2, but found $3 (did you mean $4)" % [
        $rep.cmdlineSwitch,
        $rep.cmdlineAllowed,
        $rep.cmdlineProvided,
        suggestCorrections(rep.cmdlineProvided, rep.cmdlineAllowed)
      ]

    of ...:
     ...
```

I see absolutely no value in moving things into `const` strings for templating, if the report is done a single function anyway. The current compiler probably used it because the same message could be generated in dozens of places, and the only option was to copy-paste the message (which was used pretty often), or move into `const` strings.

haxscramper:
Structured reports would also allow us to actually implement a *real* typo suggestion - not just for missing identifiers, but also for fields, named arguments, `obj.call()` and so on. Command-line options.

haxscramper:
For many errors, I have **absolutely no idea** whether it is a user-facing error, or it is a "internal assert of some kind" that just somehow ended up using `localError` or `globalError`

```nim
    if result.kind notin {kind, skTemp}:
      localError(c.config, n.info, "cannot use symbol of kind '$1' as a '$2'" %
        [result.kind.toHumanStr, kind.toHumanStr])
```

```nim
    localError(c.config, n.info,
      "VM: immediate value does not fit into an int8")
```


```nim
  if d >= 0:
    globalError(c.config, n.info, "VM problem: dest register is set")
```

and so on

haxscramper:
Current count: `328 different kinds of diagnostics`

haxscramper:
Current count: `373 different kinds of diagnostics`

haxscramper:
This PR can be applied in several steps - I did a nuclear refactor of the `msgs.nim`, which was mandatory to weed out *all* the string-based report handling, but in the end changes can be split into several steps:

1. Reorganize compiler files and add report type definitions - test suite fully passes as it is simply file shuffling
2. Replace string-based errors with strings wrapped in the Report nodes - test suite fully passes, as we just wrapped string into a special type.
3. Replace all `localError` calls etc. with Reports (major part of this PR, and most time-consuming one). Test suite is completely broken due to changed error reporting strategies
4. Automatically patch found error messages in the tests and examine for changes
	  1. Tests must be automatically rewritable - thankfully *that* part of the testament is relatively separated. I need to replace `discard """errmsg: ''"""` with a new one
	  2. New compiler output is structured, but for the time being it can be compared as a literal string. Implementing line-by-line diffing (for better test failure reporting) is not a hard prerequisite
	  3. Manually examine all changed tests - previous compiler output did not have any structure, so the best thing we can do is to rely on the human checks

Change is fully applied, and nim compiler is transitioned to the structured output

haxscramper:
I think steps 3 and 4 can be applied iteratively, although I don't see any real value in this, especially considering that the test suite itself is a mix of everything at once in every subdirectory.

haxscramper:
There must be some kind of error organization as well. Right now, compiler reports (generally speaking) things that are:

- Expected (`rsemExpectedObjectForOf`, `rsemExpectedTypeOrValue`)
- Disallowed (`rsemDisallowedNilDeref`)
- Unexpected (`rsemUnexpectedYield`)
- Invalid (`rsemInvalidTupleSubscript), Illegal (`rsemIllegalNimvmContext`, `rsemIllegalConversion`)
- Cannot (`rsemCannotBeConvertedTo`)
- Requires (`rsemRequiresDeepCopyEnabled`, `rsemEnableNotNilExperimental`) - ideally dozens of separate (and often broken) language dialects need to be phased out, but for a transition period warnings/errors are necessary.
- General errors like `rsemExprHasNoAddress`, `rsemIndexOutOfBounds`

At least naming-wise those need to be stabilized somehow

Fixup also created several hints and diagnostics that were never configurable, and a couple existing ones have really weird names like `    rsemUnusedRaises           = "Effect"`. For now, I used old string mappings, but in future those should be reconsidered as well

haxscramper:
Current count: `493 different kinds of diagnostics`

haxscramper:
Current count: `524 different kinds of diagnostics`

haxscramper:
All report refactoring is in, total count of the compiler diagnostics is:

<div align="center">

**`557`**

</div>

Implementation is not finished (I can't run the compiled compiler since I haven't implemented the reporting hook itself yet), but I think I will manage to implement proof-of-concept form today as well.

haxscramper:
Quick & dirty test for different modes of output (I omitted printing of the symbols/types/nodes as they require manual writing - `PSym` by itself contains ~15 different state fields, none of which are necessary for the end user)

S-expression (~75 lines to implement) 

```clojure
(Sem-XDeclaredButNotUsed :expression nil :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind XDeclaredButNotUsed :context nil :location (:isRange nil :lpoint (:file "/mnt/workspace/github/nimskull/lib/system/mmdisp.nim" :line 17 :col 2)) :reportInst (:file "/mnt/workspace/github/nimskull/compiler/lookups.nim" :line 360 :col 12))
(Sem-XDeclaredButNotUsed :expression nil :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind XDeclaredButNotUsed :context nil :location (:isRange nil :lpoint (:file "/mnt/workspace/github/nimskull/lib/system/mmdisp.nim" :line 38 :col 2)) :reportInst (:file "/mnt/workspace/github/nimskull/compiler/lookups.nim" :line 360 :col 12))
(Sem-XDeclaredButNotUsed :expression nil :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind XDeclaredButNotUsed :context nil :location (:isRange nil :lpoint (:file "/mnt/workspace/github/nimskull/lib/system/mmdisp.nim" :line 39 :col 2)) :reportInst (:file "/mnt/workspace/github/nimskull/compiler/lookups.nim" :line 360 :col 12))
(Sem-XDeclaredButNotUsed :expression nil :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind XDeclaredButNotUsed :context nil :location (:isRange nil :lpoint (:file "/mnt/workspace/github/nimskull/lib/system/osalloc.nim" :line 27 :col 6)) :reportInst (:file "/mnt/workspace/github/nimskull/compiler/lookups.nim" :line 360 :col 12))
(Sem-XDeclaredButNotUsed :expression nil :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind XDeclaredButNotUsed :context nil :location (:isRange nil :lpoint (:file "/mnt/workspace/github/nimskull/lib/system/repr.nim" :line 145 :col 7)) :reportInst (:file "/mnt/workspace/github/nimskull/compiler/lookups.nim" :line 360 :col 12))
(Sem-XDeclaredButNotUsed :expression nil :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind XDeclaredButNotUsed :context nil :location (:isRange nil :lpoint (:file "/mnt/workspace/github/nimskull/lib/system/seqs_v2_reimpl.nim" :line 19 :col 9)) :reportInst (:file "/mnt/workspace/github/nimskull/compiler/lookups.nim" :line 360 :col 12))
(Sem-XDeclaredButNotUsed :expression nil :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind XDeclaredButNotUsed :context nil :location (:isRange nil :lpoint (:file "/mnt/workspace/github/nimskull/lib/system/sysstr.nim" :line 118 :col 5)) :reportInst (:file "/mnt/workspace/github/nimskull/compiler/lookups.nim" :line 360 :col 12))
(Sem-Processing :expression nil :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind Processing :processing (:isNimscript nil :importStackLen 1 :fromModule "" :isToplevel t :moduleStatus "import" :path "/mnt/workspace/github/nimskull/file.nim") :context nil :location nil :reportInst (:file "/mnt/workspace/github/nimskull/compiler/modulegraphs.nim" :line 607 :col 8))
(Sem-ProcessingStmt :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind ProcessingStmt :context nil :location (:isRange nil :lpoint (:file "/mnt/workspace/github/nimskull/file.nim" :line 2 :col 2)) :reportInst (:file "/mnt/workspace/github/nimskull/compiler/passaux.nim" :line 29 :col 13))
(Sem-rsemUndeclaredField :expression (Ident field12) :expressionStr "" :rtype <type> :psym <sym> :msg "" :kind rsemUndeclaredField :context nil :location (:isRange nil :lpoint (:file "/mnt/workspace/github/nimskull/file.nim" :line 5 :col 6)) :reportInst (:file "/mnt/workspace/github/nimskull/compiler/semcall.nim" :line 379 :col 27))
```

json (for testing I reused serializer from hmisc, so it only took 4 lines to add)

```json
{"category": "repSem", "semReport": {"kind": "XDeclaredButNotUsed", "expression": "[TODO write PNode]", "expressionStr": "", "rtype": "[TODO write PType]", "psym": "[TODO write PSym]", "msg": "", "context": [], "location": {"isRange": false, "lpoint": {"file": "/mnt/workspace/github/nimskull/lib/system/osalloc.nim", "line": 27, "col": 6}}, "reportInst": {"file": "/mnt/workspace/github/nimskull/compiler/lookups.nim", "line": 360, "col": 12}}}
{"category": "repSem", "semReport": {"kind": "XDeclaredButNotUsed", "expression": "[TODO write PNode]", "expressionStr": "", "rtype": "[TODO write PType]", "psym": "[TODO write PSym]", "msg": "", "context": [], "location": {"isRange": false, "lpoint": {"file": "/mnt/workspace/github/nimskull/lib/system/repr.nim", "line": 145, "col": 7}}, "reportInst": {"file": "/mnt/workspace/github/nimskull/compiler/lookups.nim", "line": 360, "col": 12}}}
{"category": "repSem", "semReport": {"kind": "XDeclaredButNotUsed", "expression": "[TODO write PNode]", "expressionStr": "", "rtype": "[TODO write PType]", "psym": "[TODO write PSym]", "msg": "", "context": [], "location": {"isRange": false, "lpoint": {"file": "/mnt/workspace/github/nimskull/lib/system/seqs_v2_reimpl.nim", "line": 19, "col": 9}}, "reportInst": {"file": "/mnt/workspace/github/nimskull/compiler/lookups.nim", "line": 360, "col": 12}}}
{"category": "repSem", "semReport": {"kind": "XDeclaredButNotUsed", "expression": "[TODO write PNode]", "expressionStr": "", "rtype": "[TODO write PType]", "psym": "[TODO write PSym]", "msg": "", "context": [], "location": {"isRange": false, "lpoint": {"file": "/mnt/workspace/github/nimskull/lib/system/sysstr.nim", "line": 118, "col": 5}}, "reportInst": {"file": "/mnt/workspace/github/nimskull/compiler/lookups.nim", "line": 360, "col": 12}}}
{"category": "repSem", "semReport": {"kind": "Processing", "expression": "[TODO write PNode]", "expressionStr": "", "rtype": "[TODO write PType]", "psym": "[TODO write PSym]", "msg": "", "processing": {"isNimscript": false, "importStackLen": 1, "fromModule": "", "isToplevel": true, "moduleStatus": "import", "path": "/mnt/workspace/github/nimskull/file.nim"}, "context": [], "location": null, "reportInst": {"file": "/mnt/workspace/github/nimskull/compiler/modulegraphs.nim", "line": 607, "col": 8}}}
{"category": "repCmd", "cmdReport": {"kind": "rcmdExecuting", "cmd": "gcc   -o /mnt/workspace/github/nimskull/file  /tmp/nimcache/stdlib_digitsutils.nim.c.o /tmp/nimcache/stdlib_assertions.nim.c.o /tmp/nimcache/stdlib_dollars.nim.c.o /tmp/nimcache/stdlib_system.nim.c.o /tmp/nimcache/@mfile.nim.c.o    -ldl", "msg": "", "code": 0, "context": [], "location": null, "reportInst": {"file": "/mnt/workspace/github/nimskull/compiler/extccomp.nim", "line": 405, "col": 6}}}
{"category": "repInternal", "internalReport": {"kind": "rintSuccessX", "msg": "", "buildParams": {"isCompilation": true, "project": "/mnt/workspace/github/nimskull/file.nim", "output": "/mnt/workspace/github/nimskull/file", "mem": 69410816, "isMaxMem": true, "sec": 1.179445743560791, "threads": false, "backend": "", "buildMode": "debug", "optimize": "debug", "gc": "refc"}, "context": [], "location": null, "reportInst": {"file": "/mnt/workspace/github/nimskull/compiler/msgs.nim", "line": 799, "col": 6}}}
```

haxscramper:
> Change is fully applied, and nim compiler is transitioned to the structured output

After spending some more time in the error generation, I think this approach is a little too idealistic. If we had a proper test suite that allowed us to safely do the compiler refactoring, this would've been a non-issue, but considering we test *at most* sem/lexer/parser, and not every possible report at that, I don't want to take any changes and go with "smart" plans. For now, I will reroll old error output as close as I possibly can - even with this, benefits of structured reporting would be accessible, since **ALL** pretty-printing will still happen in the *single* file and *single* callback that we can iterate on (for better error formatting), or swap out at moment's notice (json and S-expression reports would be readily available at all times).

haxscramper:
Rebased on the latest devel, starting final fixes and test runs, for now expect everything to fail

haxscramper:
This is more of a todo/cleanup note for myself. I was trying to fix bug where `vmgen.genRdVar` failed with index defect. Of course this took a lot more time than I'd like to admit, and the fix itself was pretty trivial, but the amount of monkey-code that I've encountered while trying to dig up this is absolutely insane.

- when list of possible candidates in `call(<arg>)` is constructed, the compiler also considers ones that have `static[]` parameters, and (obviously) it needs to know whether the expression can be evaluated at compile-time, so it calls into vm code. `sem.tryConstExpr` calls into `vm.evalConstExprAux`
- `evalConstexprAux` constructs new/cleans up existing VM context using `setupGlobalContext`, which removes all registers (sets to empty list). That's where the main part of the misunderstanding started - `genRdVar` requires registers to be properly set, but global context setup wipes them out clean, and there was *not a single proc in execution path that actually filled in these registers*. 
- as it turns out, the registers were not filled in because:
- prior to calling `genRdVar()` (which generates vm code for `variable/global/const` etc. read)
- compiler calls `checkCanEval`, which supposedly checks whether AST can be evaluated at compile-time
- after cleaning up a couple of layers of weird heuristics, I find this lovely condition
  ```nim
  s.kind in {skVar, skTemp, skLet, skParam, skResult} and
	not s.isOwnedBy(c.prc.sym) and s.owner != c.module and c.mode != emRepl
  ```
    - where `s` is a symbol we are trying to check for compile-time readability
    - if it is a var/temp/let/param/result it obviously can't be evaluated at compiletime
    - not sure about  `isOwnedBy` part
- then it checks for
  ```nim
  if sfGenSym in s.flags and c.prc.sym == nil: discard
  ```

because of https://github.com/nim-lang/Nim/issues/12612

- and then calls `globalError(c.config, n.info, "cannot evaluate at compile time: "`

Continuing trail of insanity, `globalError`, despite being described as
```nim
  ## `local` means compilation keeps going until errorMax is reached (via `doNothing`),
  ## `global` means it stops.
```

is actually used for exception-based control flow here, it seems, since it raises an exception which is supposedly caught somewhere up the chain, specifically
```nim
  try:
    result = evalConstExpr(c.module, c.idgen, c.graph, e)
    if result == nil or result.kind == nkEmpty:
      result = nil
    else:
      result = fixupTypeAfterEval(c, result, e)
  except ERecoverableError:
    result = nil
```

So, to recap - instead of returning true/false `nil/PNode` for the "can evaluate as a constant expression", compile uses *the error reporting system* to raise an exception that is then bounced through **TWO** compiler subsystems (VM and sem) and then finally caught in sem layer. Oh, and all of that happens when we consider *possible* candidates for procedure overloading, and just so happened to have one candidate that has `static` argument.

haxscramper:
Cleanup/todo notes

- Compiler has a lot of 'fail fast' checks for the `compiles()` environment error reporting - I think this should be made into runtime configuration value - debugging spontaneous `compiles()` failures have been one of the most frustrating experiences to me
- Found multiple smaller (aka "good first issue") refactoring spots, these need to be documented as issues.
    - VM `rawExecute` uses an implicitly injected `return` statement that is added by `template stackTrace()`. This makes code much harder to read, since you can't properly see where the loop is terminated. I've managed to trip myself on this multiple times. 
    - `globalError()` monkey-code described above

saem:
initial notes of findings when running tests, will use this as a sort of todo list:
* generally I think --hints:off is being ignored
* `tests/effects/tdiagnostic_messages.nim`: is a suprious failure locally due to the stacktrace from the debug compiler
* `tests/errmsgs/t10735.nim`:
  * shows `error` in the message; need to turn that off in renderer
  * shouldn't repeat output as much -- preferred, non-blocker
  * shouldn't crash (good)
* `tests/objvariant/temptycaseobj.nim`: fix things
* `tests/tools/tlinter.nim`: upgraded mismatched casing from hint to warning
* `tests/strictnotnil/tnilcheck.nim`: is a crash
* `tests/osproc/treadlines.nim`: seems like hints are being passed through or just a stacktrace in the way
* `tests/parser/twrongcmdsyntax.nim`: very different error message
* `tests/errmsgs/t5167_5.nim`: totally different error messages
* `tests/effects/tgcsafe.nim`: line number mismatch
* `tests/generics/tlateboundgenericparams.nim` assertion failure/crash
* missing data in message:
  * `tests/effects/tstrict_effects.nim`
  * `tests/effects/tstrict_effects_sort.nim`
* NPE in reports.nim:
  * `tests/notnil/tmust_compile.nim`
  * `tests/collections/ttablesthreads.nim`
  * `tests/stdlib/tsharedtable.nim`
* `tests/errmsgs/t9768.nim`: hints and stacktrace message screwing things up
* `tests/types/tinheritance_conversion.nim`: line info missing from message
* `tests/parallel/tgc_unsafe2.nim`: no message
* `tests/errmsgs/tsigmatch2.nim`: repeated message
* `tests/errmsgs/tundeclared_routine.nim`: message format change
* `tests/errmsgs/twrongcolon.nim`: totally different error message
* `tests/lang/s05_pragmas/s02_misc/t04_warning.nim`: message is different and doesn't contain the same amount of info -- hints not suppressed
* `tests/misc/trunner.nim`: don't even understand the test
* `tests/stdlib/tosprocterminate.nim`: code gen error?
* `tests/stdlib/tvmutils.nim`: missing source output

----

reviewing the list, I think the first thing to fix is the hints handling which doesn't seem to be ignored -- also the colouring.

compiler/cli_reporter.nim
```diff
@@ -0,0 +1,3409 @@
...
+proc reportFull*(conf: ConfigRef, r: ExternalReport): string =
+  assertKind r
+  reportBody(conf, r)
+
+proc reportShort*(conf: ConfigRef, r: ExternalReport): string {.inline.} =
+  # mostly created for nimsuggest
+  reportBody(conf, r)
+
+const
+  dropTraceExt = on
```
saem:
```suggestion
  dropTraceExt = off
```

it's useless to me without this, I can't click through to go to definition.

compiler/msgs.nim
```diff
@@ -478,175 +477,246 @@ proc sourceLine*(conf: ConfigRef; i: TLineInfo): string =
...
   else:
     result = conf.m.fileInfos[i.fileIndex.int32].quotedName
 
-template listMsg(title, r) =
-  msgWriteln(conf, title, {msgNoUnitSep})
-  for a in r: msgWriteln(conf, "  [$1] $2" % [if a in conf.notes: "x" else: " ", $a], {msgNoUnitSep})
+proc listWarnings*(conf: ConfigRef) =
+  conf.localReport(InternalReport(
+    kind: rintListWarnings,
+    enabledOptions: repWarningKinds * conf.notes))
```
saem, line 702: I don't think it's here, but somewhere along the way the `computeNotesVerbosity` related information that gets populated into `mainPackageNotes` and then used to determine whether to show the message origins got lost along the way.

That needs a pain as remember `--hint:MsgOrigin=on` or whatever it is is difficult to remember and really should be a CLI option.

saem (changes requested):

haxscramper:
commit message wip

```
Make compiler report structured

Full rework of the compiler message handling pipeline. Remove old-style message generation that was
based purely on strings that were formatted in-place, and instead implement structured logging where
main compiler code only instantiates objects with required information.

Explanation of changes for this commit will be split into several sections, matching number of edit
iterations I had to make in order for this to work properly.

* File reorganization

In addition to the architectural changes this PR requires some type definition movements.

- PNode, PSym and PType definitions (and all associated types and enums) were moved to the
  ast_types.nim file which is then reexported in the ast.nim
- Enum defininitions for the nilability checking were moved to nilcheck_enums.nim and then
  reexported
- Enum definitions for the VM were moved to to vm_enums.nim

* New files

- Main definition of the report types is provided in the reports.nim file, together with minor
  helper procedures and definition of the ReportList type. This file is imported by options, msgs
  and other parts of the compiler.
- Implementation of the default error reporting is provided in the cli_reporter.nim - all
  disguisting hardcoded string hacks were moved to this single file. Now you can really witness the
  "error messages are good enough for me" thinking that prevailed in the compiler UX since it's
  inception.

* Changed files

Most of the changes follow very simple pattern - old-style hardcoded hacks are replaced with
structural report that is constructed in-place and then converted to string /when necessary/. I'm
pretty sure this change already puts me close to getting CS PHD - it was *unimaginably hard* to
figure out that hardcoding text formatting in place is not the best architecture. Damn, I can
probably apply to the nobel prize right now after I figure that out.

** Changes in message reporting pipeline

Old error messages where reportined via random combinations of the following:

- Direct calls to `msgs.liMessage` proc - it was mostly used in the helper templates like
  `lintReport`
- `message`
- `rawMessage`
- `fatal`
- `globalError` - template for reporting global errors. Misused to the point where name completely
  lost it's meaning and documentation does not make any sense whatsoever. [fn:global-err]
- `localError` - template for reporting necessary error location, wrapper around `liMessage`
- `warningDeprecated` - used two times in the whole compiler in order to print out error message
  about deprecated command switches.

Full pipeline of the error processing was:

- Message was generated in-place, without any colored formatting (was added in `liMessage`)
  - There were several ways message could be generated - all of them were used interchangeably,
    without any clear system.
    1. Some file had a collection of constant strings, such as `errCannotInferStaticParam = "cannot
       infer the value of the static param '$1'"` that were used in the `localReport` message
       formatting immediately. Some constants were used pretty often, some were used only once.
    2. Warning and hint messages were categorized to some extent, so and the enum was used to
       provide message formatting via `array[TMsgKind, string]` where `string` was a `std/strutils`
       formatting string.
    3. In a lot of cases error message was generated using hardcoded (and repeatedly copy-pasted)
       string
- It was passed down to the `liMessage` (removed now) procedure, that dispatched based on the global
  configuration state (whether `ignoreMsgBecauseOfIdeTools` holds for example)
- Then message, together with necessary bits of formatting (such as `Hint` prefix with coloring) was
  passed down to the `styledMsgWriteln(args: varargs[typed])` template, whcih did the final dispatch
  into
- One of the two different /macros/ for writing text out - `callStyledWriteLineStderr` and
  `callIgnoringStyle`.
  - Dispatch could happen into stdout/stderr or message writer hook if it was non-nil
- When message was written out it went into `writeLnHook` callback (misused for
  `{.explain.}` [fn:writeln-explain]) (hacked for `compiles()` [fn:writeln-compiles]) and was
  written out to the stdout/stderr.

It is now replaced with:

- `Report` object is instantiated at some point of a compilation process (it might be an immediate
  report via `localError` or delayed via `nkError` and written out later)
- `structuredReportHook` converts it to string internally. All decitions for formatting, coloring
  etc. happen inside of the structured report hook. Where to write message, which format and so on.
- `writeLnHook` /can/ be used by the `structuredReportHook` if needed - right now it is turned into
  simple convenience callback. In future this could even be replaced with simple helper proc, for
  now I left it as it is now because I'm not 100% sure that I can safely remove this.

** Changes in the warning and hint filtering

Report filtering is done in the particular *implementation* of the error hook -
`options.isEnabled()` provides a default predicate to check whether specific report can be written
out, but it must still be called manually. This allows to still see all the reports if needed. For
example, `cli_reporter.reportHook()` uses additional checks to force write some reports (such as
debug trace in `compiles()` context).

Most of the hint and warning were already categorized, altough some reports had to be split into
several (such as `--hint=Performance` that actually controlled reports for `CopiesToSink` and
`CannotMakeSink`, `--hint=Name` that controlled three reports).

None of the errors were categorized (reports from the `nkError` progress was incorporated, but in
I'm mostly describing changes wrt. to old reporting system) and all were generated in-place


** Minor related changes

- List of allowed reports is now stored in the `noteSets: array[ConfNoteSet, ReportKinds]` field of
  the `ConfigRef`, instead of *seven* separate feilds. Access is now done via getter/setter procs,
  which allows to track changes in the current configuration state.
- Renamed list of local options to `localOptions` - added accessors to track changes in the state.
- Move all default report filter configuration to `lineinfos.computeNotesVerbosity()` - previously
  it was scattered in two different places: `NotesVerbosity` for `--verbosity:N` was calculated in
  `lineinfos`, foreignPackageNotesDefault was constructed in `options`
- Changed formatting of the `internalAssert` and `internalError` messages
- Removed lots of string formatting procs from the various `sem*` modules - ideally they should
  *all* be moved to the `cli_reporter` and related.



-------

Additional notes

[fn:global-err] As I said earlier, `globalError` was misused - it is still not possible to fully get
rid of this sickening hack, since it is used for /control flow/ (you read this correct - it is an
error reporting template, and it is used for control flow. Wonderful design right?).

So, in short - `globalError` can raise `ERecoverableError` that can be caught in some other layer
(for example `sem.tryConstExpr` abuses that in order to bail out (pretty sure it was done as an
'optimization' of some sort, just like 'compiles') when expression fails to evaluate at
compile-time [fn:except])

[fn:except] https://github.com/nim-works/nimskull/pull/94#issuecomment-1006927599

[fn:writeln-explain] Of course you can't have a proper error reporting in the nim compiler, so this
hook was also misused to the point of complete nonsense. Most notable clusterfuck where you could
spot this little shit is implementation of `{.explain.}` pragma for concepts. It was implemented via
really 'smart' (aka welcome to hell) solution in

https://github.com/nim-works/nimskull/commit/74a80988d9289e8147a791c4b0939d4287baaff3 (sigmatch
~704) and then further "improved" in
https://github.com/nim-lang/Nim/commit/fe48dd1cbec500298f7edeb75f1d6fef8490346c by slicing out parts
of the error message with `let msg = s.replace("Error:", errorPrefix)`

For now I had to reuse similar hack - I *do* override error reporting hook in order to collect all
the missing report information. In the future it would be unnecessary - when concept is matched it's
body will be evaluated to `nkError` with reports that are written out later.

[fn:writeln-compiles] when `compiles()` check is executed, all error output is temporarily disabled
(both stderr and stdout) via setting allowed output flags to `{}` (by default it is set to

```

.gitignore
```diff
@@ -114,3 +114,7 @@ nimdoc.out.css
 # For anyone who wants to make a workspace out of nimskull repo
 *.code-workspace
 .vscode
+
+file.nim
+build.sh
+**/haxdoc/*
```
alaviss: What are these?

compiler/ast.nim
```diff
@@ -1042,6 +60,12 @@ const
   OverloadableSyms* = {skProc, skFunc, skMethod, skIterator,
     skConverter, skModule, skTemplate, skMacro, skEnumField}
 
+  skipForDiscardable* = {nkIfStmt, nkIfExpr, nkCaseStmt, nkOfBranch,
+    nkElse, nkStmtListExpr, nkTryStmt, nkFinally, nkExceptBranch,
+    nkElifBranch, nkElifExpr, nkElseExpr, nkBlockStmt, nkBlockExpr,
+    nkHiddenStdConv, nkHiddenDeref}
```
alaviss, line 66: Need docs for this

compiler/ast_types.nim
```diff
@@ -0,0 +1,1173 @@
...
+
+  PNode* = ref TNode
+  TNodeSeq* = seq[PNode]
+  PType* = ref TType
+  PSym* = ref TSym
+  ReportId* = distinct uint32 ## Id of the report in the report list.
+  ## Report itself is defined in the `reports.nim` which imports `ast.nim`,
+  ## so id type is 'forward'-declared here to avoid cyclic dependencies.
+  ## Main type definitions are in the `reports.nim`, storage of the reports
+  ## list (that report id indexes into)
```
alaviss, line 824: I don't remember if this is needed, but usually you need to add an indent for docs to get attached to the definition.

```suggestion
   ## Report itself is defined in the `reports.nim` which imports `ast.nim`,
   ## so id type is 'forward'-declared here to avoid cyclic dependencies.
   ## Main type definitions are in the `reports.nim`, storage of the reports
   ## list (that report id indexes into)
```

compiler/cli_reporter.nim
```diff
@@ -0,0 +1,3405 @@
...
+
+  result.add str
+  result.add "\e[0m"
+
+func wrap*(
+    conf: ConfigRef,
+    str: string,
+    color: ForegroundColor,
+    style: set[Style] = {}
+  ): string =
```
alaviss, line 61: Need docs for just about every procs in this module

compiler/cli_reporter.nim
```diff
@@ -0,0 +1,3405 @@
+import reports,
```
alaviss, line 12: Need at least a summary of what this module does.

compiler/nim.cfg
```diff
@@ -30,3 +30,5 @@ define:useStdoutAsStdmsg
   experimental:strictEffects
   warningAsError:Effect:on
 @end
+
+errormax:4
```
alaviss: Why do we have this here?

compiler/options.nim
```diff
@@ -389,16 +400,250 @@ type
...
+    # let n = rintMsgOrigin
+    # if (n in before) != (n in after):
+    #   if n notin after:
+    #     writeStackTrace()
+    #     echo "changed conf $# -> $#" % [$(n in before), $(n in after)]
+
+  else:
+    body
+
+proc incl*(conf: ConfigRef, nset: ConfNoteSet, note: ReportKind) =
```
alaviss, line 436: Need docs for the new stuff.

alaviss (commented):
This is what I got out from a quick skim

compiler/ast.nim
haxscramper: no idea what this was, moved from sem

compiler/ast_types.nim
haxscramper: not needed

compiler/cli_reporter.nim
haxscramper: provided in `reports` module itself

compiler/cli_reporter.nim
```diff
@@ -0,0 +1,3405 @@
...
+  for candidate in candidates:
+    result.add "\n ($1, $2): '$3'" % [
+      $candidate.dist,
+      $candidate.depth,
+      $candidate.sym.name.s
+    ]
+
+    result.addDeclaredLoc(conf, candidate.sym)
+
+proc reportBody*(conf: ConfigRef, r: SemReport): string =
```
saem, line 491: I would put this proc as late as possible in the file because it's a beast, public API, and drowns out what follows.

saem: In fact I would use prototypes to shove the big ol' report body procs way at the bottom.

saem (approved):

saem:
Bors r+

compiler/cli_reporter.nim
```diff
@@ -0,0 +1,3405 @@
...
+  if path in conf.m.filenameToIndexTbl:
+    let id = conf.m.filenameToIndexTbl[path]
+    result = toFilenameOption(conf, id, conf.filenameOption)
+
+  else:
+    const compilerRoot = currentSourcePath().parentDir()
+    # Path not registered in the filename table - most likely an
+    # instantiation info report location
+    if conf.filenameOption == foCanonical and
+       path.startsWith(compilerRoot):
```
alaviss: Will need a bit more explaination (code comments) for this code path.

Also, the whole full source path query should be gated behind excessiveStackTrace as discussed in chat

PR #96, opened 2021-12-06
Testament - implement --retry and fix --failing

shayanhabibi:
- [x] Documentation
- [ ] Make easier to maintain and better file structure
- [x] Make --failing also filter stdout

### DIFF
--- a/testament/backend.nim
+++ b/testament/backend.nim
@@ -37,28 +37,22 @@ proc getCommit(): CommitId =
   result = CommitId(hash)
 
 var
-  results: File
+  results: JsonNode
   currentCategory: string
-  entries: int
 
 proc writeTestResult*(name, category, target, action, result, expected, given: string) =
   createDir("testresults")
   if currentCategory != category:
     if currentCategory.len > 0:
-      results.writeLine("]")
-      close(results)
+      var resFile = open("testresults" / category.addFileExt"json", fmWrite)
+      resFile.write pretty(results)
+      close resFile
     currentCategory = category
-    results = open("testresults" / category.addFileExt"json", fmWrite)
-    results.writeLine("[")
-    entries = 0
+    results = newJArray()
 
-  let jentry = %*{"name": name, "category": category, "target": target,
+  results.add %*{"name": name, "category": category, "target": target,
     "action": action, "result": result, "expected": expected, "given": given,
     "machine": thisMachine.string, "commit": thisCommit.string, "branch": thisBranch}
-  if entries > 0:
-    results.writeLine(",")
-  results.write($jentry)
-  inc entries
 
 proc open*() =
   thisMachine = getMachine()
@@ -66,5 +60,95 @@ proc open*() =
 
 proc close*() =
   if currentCategory.len > 0:
-    results.writeLine("]")
-    close(results)
+    var resFile = open("testresults" / currentCategory.addFileExt"json", fmWrite)
+    resFile.write pretty(results)
+    close resFile
+
+proc cacheResults*() =
+  ## Traverses the testresults directory and extracts any failed json entries
+  ## and inserts them into a new file within the cacheresults directory.
+  createDir("testresults/cacheresults")
+  # We will ignore any entries that have this as their result
+  const passResults = ["reJoined", "reSuccess", "reDisabled", ""] # "" is defaulted to if result field not found
+  let searchPattern = "testresults" / "*.json"
+  # Prepare json array which will be written to our new cache file
+  var fresults = newJArray()
+  # Walk files in testresults that have a json ext
+  for f in walkFiles(searchPattern):
+    var jdata: JsonNode
+    try:
+      # Parse the json file into the json node
+        jdata = f.parseFile()
+    except IOError:
+      echo getCurrentExceptionMsg()
+      echo "Failed to open/parse the json file in /testresults/. It is likely the file",
+            " is corrupted or locked by another process. Try deleting all files in /testresults/ ",
+            "and then running testament without '--retry'."
+      jdata = newJArray()
+    except JsonParsingError:
+      echo getCurrentExceptionMsg()
+      echo "Failed to parse the cached results. It is likely the file was not ",
+            "written to correctly, or was interrupted while doing so. Try ",
+            "running testament without '--retry' first."
+      jdata = newJArray()
+    # The testresults files should be made up of arrays of json objects representing
+    # an entry each. Traverse the array and check the result field of each object.
+    # If the result is not one of the "passResults", then we add it to our cache
+    # node 'fresults'
+    for node in jdata:
+      let noderesult = node{"result"}.getStr()
+      if not passResults.contains(noderesult):
+        fresults.add(node)
+  
+  var results = open("testresults" / "cacheresults" / "result".addFileExt"json", fmWrite)
+  results.write(fresults.pretty())
+  close(results)
+
+proc getRetries*(): (seq[string], seq[(string, string)]) =
+  ## Returns the stored categories and tuples of test names/targets within the
+  ## cache. This is typically combined with the RetryContainer object to set
+  ## a global accessible object with the information.
+  ## Result = (Categories that had fails, Name and target combination of failed test)
+  # Gets the files that failed in the previous run from the cache
+  var
+    cats: seq[string]
+    file_targs: seq[(string, string)]
+    jdata: JsonNode
+  # Directory of cache
+  const cacheDir = "testresults" / "cacheresults"
+  createDir(cacheDir) # does nothing if dir already exists
+  const cacheFile = cacheDir / "result".addFileExt"json"
+
+  jdata = newJArray()
+  try:
+    # Parse the json file into the json node
+    jdata = cacheFile.parseFile()
+  except IOError:
+    echo getCurrentExceptionMsg()
+    echo "Failed to open/parse the cached results. It is likely the file does",
+          " not exist yet. Try running testament without '--retry' first."
+  except JsonParsingError:
+    echo getCurrentExceptionMsg()
+    echo "Failed to parse the cached results. It is likely the file was not ",
+          "written to correctly, or was interrupted while doing so. Try ",
+          "running testament without '--retry' first."
+
+  let cacheArray = cacheFile.parseFile()
+  # Traverse the array of entries, add the category to cats if it has not been seen,
+  # either way add the "name" field to file_targs
+  for entry in cacheArray:
+    let
+      cat = entry["category"].getStr()
+      file_targ = entry["name"].getStr()
+      # Split target from file name. RSplit with maxsplit of 2 is used to
+      # avoid corrupting test names of tests with spaces in dir
+      name_target = file_targ.rsplit(maxsplit = 2)
+    # Do not want duplicates; only add if not already
+    if cat notin cats:
+      cats.add cat
+    file_targs.add:
+      if name_target.len > 1: (name_target[0], name_target[1])
+      # In the case that rsplit did not separate the name into a target and
+      # test name then we still add it
+      else: (name_target[0], "")
+  result = (cats, file_targs)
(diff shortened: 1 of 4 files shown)


testament/testament.nim
```diff
@@ -30,10 +30,30 @@ var backendLogging = true
...
+## Blanket method to encaptsulate all echos while testament is detangled.
+## Using this means echo cannot be called with separation of args and must instead
+## pass a single concatenated string so that optional paramters can be
+## included
+type
+  TEchoDetail = enum
+    Undefined,
+    Progress,
+    ProcessCmdCall
+proc techo(str: string, lvl: TEchoDetail = Undefined) =
```
alaviss: Maybe change this to something like:

```nim
proc msg(msgType: MessageType, parts: varargs[string, `$`]) =
  <logic to not print here>

  # This is basically what `echo` is, without the magical `varargs[typed]`
  stdout.writeLine parts
  flushFile stdout
```

testament/categories.nim
```diff
@@ -553,6 +553,8 @@ proc processCategory(r: var TResults, cat: Category,
                      options, testsDir: string,
                      runJoinableTests: bool) =
   let cat2 = cat.string.normalize
+  if cat2 in retryContainer.cats:
```
saem: Should this be not in?

testament/specs.nim
```diff
@@ -265,6 +273,10 @@ proc isCurrentBatch*(testamentData: TestamentData; filename: string): bool =
 
 proc parseSpec*(filename: string): TSpec =
   result.file = filename
+  if retryContainer.retry:
+    if retryContainer.names.anyIt(it in result.file):
+      result.err = reDisabled
```
saem: You're not returning early or anything, so it's still doing the rest of the work.

saem (changes requested):
Suggested a few things.

saem:
Your code is so much better. 😭

Ok so, is shrink it to get your retry bits in and maybe have a draft for the additional changes. Imma take a crack at "testament2" this weekend. Basically spec parsing only, shrink the category madness, and ditch megatest.

Should give a much cleaner base to work with... seriously.

testament/specs.nim
```diff
@@ -265,6 +273,10 @@ proc isCurrentBatch*(testamentData: TestamentData; filename: string): bool =
 
 proc parseSpec*(filename: string): TSpec =
   result.file = filename
+  if retryContainer.retry and retryContainer.names.anyIt(it in result.file):
```
saem: So here is an idea, that you'll love.... Ok maybe love is a strong word, but if it's possible to store the full log, we could see which tests are newly discovered and need to be run, or only store successes?

So some failures, add tests, retry runs failure and new.

testament/testament.nim
```diff
@@ -30,10 +30,32 @@ var backendLogging = true
...
+    Undefined,
+    Progress,
+    ProcessCmdCall
+
+proc msg(msgType: MessageType; parts: varargs[string, `$`]) =
+  block:
+    if optFailing and not optVerbose:
+      case msgType
+      of ProcessCmdCall:
+        break
```
saem: Can't the whole block break thing go away so a single if?

saem (commented):
Non-blockers, just noticed something.

testament/testament.nim
shayanhabibi: It can; but thats not as extensible. I'll reduce it for now.

shayanhabibi:
> So here is an idea, that you'll love.... Ok maybe love is a strong word, but if it's possible to store the full log, we could see which tests are newly discovered and need to be run, or only store successes?

I like this. Can add this after; I just really want to be able to use the failing/retry when working with other branches from devel.

shayanhabibi:
I am happy with the current implementation of --failing and --retry.

It's still not neat enough and there are things to add as saem has suggested, however I think improving compiler tooling is pretty important.

In essence:

- `--failing` will now filter stdout and only print failed tests
- `--retry` will only run previously failed tests from last `all` run

testament/testament.nim
```diff
@@ -637,156 +658,194 @@ proc loadSkipFrom(name: string): seq[string] =
...
 
-  # now figure out what we're going to do
   case action
-  of "all":
+  of "all": # Run all tests
+    var cats: seq[string]
+    var cmds: seq[string]
+    ## def qol procedure
+    proc progressStatus(idx: int) =
+      msg Progress: `%`("progress[all]: $1/$2 starting: cat: $3", [$idx, $cats.len, cats[idx]])
```
alaviss:
```suggestion
      msg Progress:
        "progress[all]: $1/$2 starting: cat: $3" % [$idx, $cats.len, cats[idx]]
```

testament/backend.nim
```diff
@@ -1,70 +1,121 @@
...
+  # Prepare json array which will be written to our new cache file
+  var fresults = newJArray()
+  # Walk files in testresults that have a json ext
+  for f in walkFiles(searchPattern):
+    var jdata: JsonNode
+    try:
+      # Parse the json file into the json node
+        jdata = f.parseFile()
+    except IOError, JsonParsingError:
+      echo getCurrentExceptionMsg()
```
alaviss: Don't you want to use `msg` here?

And you should write more than just the exception message, it provides pretty much no substance for this case.

testament/backend.nim
```diff
@@ -1,70 +1,121 @@
...
+    resFile.write pretty(results)
+    close resFile
+
+proc cacheResults*() =
+  ## Traverses the testresults directory and extracts any failed json entries
+  ## and inserts them into a new file within the cacheresults directory.
+  createDir("testresults/cacheresults")
+  # We will ignore any entries that have this as their result
+  const passResults = ["reJoined", "reSuccess", "reDisabled", ""] # "" is defaulted to if result field not found
+  let searchPattern = "testresults" & DirSep & "*.json"
```
alaviss:
```suggestion
  let searchPattern = "testresults" / "*.json"
```

testament/backend.nim
```diff
@@ -1,70 +1,121 @@
...
+    # The testresults files should be made up of arrays of json objects representing
+    # an entry each. Traverse the array and check the result field of each object.
+    # If the result is not one of the "passResults", then we add it to our cache
+    # node 'fresults'
+    for node in jdata:
+      let noderesult = node{"result"}.getStr()
+      if not passResults.contains(noderesult):
+        fresults.add(node)
+  
+  # if len(fresults) > 0: # We want to overwrite the cache if we have a succesful
```
alaviss:
```suggestion
  # We want to overwrite the cache if we have a successful
```

The commented `if` doesn't add anything here.

testament/backend.nim
```diff
@@ -1,70 +1,121 @@
...
+  # run, therefore we should write to the file even if the cache node is empty
+  var results = open("testresults" / "cacheresults" / "result".addFileExt"json", fmWrite)
+  results.write(fresults.pretty())
+  close(results)
+
+proc getRetries*(): (seq[string], seq[(string, string)]) = # Returns all the categories of concern and then the "name" entry list
```
alaviss: What does this trailing comment do?

testament/backend.nim
```diff
@@ -1,70 +1,121 @@
...
+  ## Result = (Categories that had fails, Name and target combination of failed test)
```
alaviss: Please actually write out what this function does.

testament/backend.nim
```diff
@@ -1,70 +1,121 @@
...
+  # Gets the files that failed in the previous run from the cache
+  var cats: seq[string]
+  var file_targs: seq[(string, string)]
```
alaviss: Merge these please.

testament/backend.nim
```diff
@@ -1,70 +1,121 @@
...
+  const cacheDir = "testresults" / "cacheresults"
+  createDir(cacheDir)
+  const cacheFile = cacheDir / "result".addFileExt"json"
+  let cacheArray = cacheFile.parseFile()
+  # Traverse the array of entries, add the category to cats if it has not been seen,
+  # either way add the "name" field to file_targs
+  for entry in cacheArray:
+    let cat = entry["category"].getStr()
+    let file_targ = entry["name"].getStr()
+    let name_target = file_targ.split()
```
alaviss: Are you trying to split `tests/something/foo.nim C` into path and target? This kind of splitting will not work with the `dir with spaces` tests. Probably an `rsplit` with `maxsplit` of 2 would work.

testament/specs.nim
```diff
@@ -106,6 +107,13 @@ type
     debugInfo*: string # debug info to give more context
     knownIssues*: seq[string] ## known issues to be fixed
 
+  RetryContainer* = object
```
alaviss, line 110: Needs documentation

testament/testament.nim
```diff
@@ -637,156 +658,194 @@ proc loadSkipFrom(name: string): seq[string] =
...
       skips = loadSkipFrom(skipFrom)
       for i, cati in cats:
         progressStatus(i)
         processCategory(r, Category(cati), p.cmdLineRest, testsDir, runJoinableTests = false)
     else:
       addExitProc azure.finalize
-      quit osproc.execProcesses(cmds, {poEchoCmd, poStdErrToStdOut, poUsePath, poParentStreams}, beforeRunEvent = progressStatus)
-  of "c", "cat", "category":
+      let processOpts =
+        if optFailing and not optVerbose:
```
alaviss, line 804: I'm not sure how `optFailing` can override `optVerbose`?

alaviss (changes requested):

testament/backend.nim
shayanhabibi: Circular import unfortunately

testament/testament.nim
shayanhabibi: optVerbose implies more information. optFailing hides everything to not do with progress and failures; however, combined with optVerbose, you will still have information printed such as the command that was passed to the process to run that test (as per Saems request).

saem (approved):

saem:
This is better, I can make a bunch of other improvements in the weekend, we should be in good shape.

testament/backend.nim
```diff
@@ -37,34 +37,109 @@ proc getCommit(): CommitId =
...
+  # Prepare json array which will be written to our new cache file
+  var fresults = newJArray()
+  # Walk files in testresults that have a json ext
+  for f in walkFiles(searchPattern):
+    var jdata: JsonNode
+    try:
+      # Parse the json file into the json node
+        jdata = f.parseFile()
+    except IOError, JsonParsingError:
+      echo getCurrentExceptionMsg()
```
alaviss, line 83: You need to do better than just printing the exception message. Don't print an error and pretend like nothing happened.

testament/testament.nim
```diff
@@ -637,156 +658,195 @@ proc loadSkipFrom(name: string): seq[string] =
...
-      else:
-        quit Usage
-    of "backendlogging":
-      case p.val:
-      of "on":
-        backendLogging = true
-      of "off":
-        backendLogging = false
+  ## Define CLI Options/Args Parsing loops
+  template parseOptions: untyped {.dirty.} =
```
alaviss: Not sure about templates here, wouldn't named blocks do the trick?

shayanhabibi: I prefer being able to fold what should have probably been independent code blocks/procedures from the beginning at the head of the procedure since the case action is imo the most significant code block.

testament/backend.nim
shayanhabibi: Yep thats my bad; had changed the other occurrence to have a better message printed but didn't copy it over

testament/testament.nim
alaviss: Then it should take a parser argument.

alaviss (approved):
`[TESTAMENT]` -> `testament:` is the only nit I have left.

Great work!

testament/testament.nim
```diff
@@ -637,156 +658,195 @@ proc loadSkipFrom(name: string): seq[string] =
...
-      else:
-        quit Usage
-    of "backendlogging":
-      case p.val:
-      of "on":
-        backendLogging = true
-      of "off":
-        backendLogging = false
+  ## Define CLI Options/Args Parsing loops
+  template parseOptions(p: var OptParser): untyped {.dirty.} =
```
alaviss: I lied, here is an another nit: Don't use `{.dirty.}` unless you have to. I don't see you declaring any variables that have to be accessible outside of the template (nor should you), so it should lose the `{.dirty.}` qualifier

alaviss (approved):

PR #98, opened 2021-12-06
remove unused or weird modules

ringabout:
## Remove unused or weird modules

- remove compiler/debuginfo.nim and lib/system/sysspawn.nim which are not used.
- remove lib/pure/collections/chains.nim which is a weird module.

Reasonale:

- It is dead code but still rendered unnecessary changes: https://github.com/nim-lang/Nim/commits/devel/lib/system/sysspawn.nim
- People cannot know whether it is used or not at first sight, it could waste contributors' time caused by looking into/playing with the module


alaviss (commented):
Is `sysspawn` not used for `threadpools`?

ringabout:
yep, it has been barely used since it was created. https://github.com/nim-lang/Nim/commit/6195dbe491ccd864c5dcb59f87826291ac1f1ff4#diff-dc783edd06384eee05375a8462c79a57bd8ef8d1510e82e8ca3868164734ea84 removed the use of sysspawn.

alaviss (commented):
LGTM, though please specify the modules you are removing in the commit message body as well.

saem:
@xflywind I was thinking and I could be wrong, that we'd want to keep debuginfo.

I didn't know about it for a long time and prior to that I did a whole bunch of really annoying things in the Python gdb pretty printer that would likely be dramatically simplified if this was included.

Would you mind taking a look and sharing your thoughts here? Thanks.

ringabout:
I'm fine with it being kept. It needs to be reworked though (the module even cannot compile due to the changes of `SigHash`).

saem (approved):
Just read chains, it's a hack for destructive lists from a quick glance.

I'm glad it's going. Thanks!

Bors r+

alaviss:
Looks like the server got a stroke, lets try again.

bors r+

alaviss:
Maybe it'll work now

bors r+

PR #99, opened 2021-12-06
[FIX] Allow low(k) != discriminants

haxscramper:
Allow types with nonzero `low()` to be used in variant objects. PR should fail only for the error message test.


saem:
Can we put those in draft/help wanted?

I'd like to get some tests against this. I saw a bunch of the discussion around it, so o get the motivation, but I think it'll cause a bunch of bugs with the way things are now. :(

haxscramper:
That does indeed cause bugs, but only if you want to use a type with nonzero-low, which is a rare case even with this PR. I'm not sure how this can be tested, because the whole point is to *not* check anything, and accept the fact this flexibility might sometimes cause an *unwanted, though totally consistent with everything else* behavior.

alaviss (changes requested):
It's a no from me, Nim depends too much on structures being zero-initialized right now, and this change will have a lot of implications on destructors, of which assumes a zero-initialized structure to be valid, and who knows what other parts does this as well.

Also, I don't think it is appropriate to introduce even more unwanted behaviors, we have enough of them to clean up right now.

haxscramper:
In that case we probably need to put this on hold and return to it when spec for this part is done. Right now it is too broken and we can't reason about implications of this change.

Maybe schedule this after default(T) works properly.

saem:
Thanks, I agree that we should wait to spec it out, lots of weirdness.

haxscramper:
I think this PR is certainly not something that we can think about in the close future/(maybe ever), so I'm closing it, since it already started accumulating merge conflicts, despite being a +3/-19 edit.

Issue #89, opened 2021-12-02
Streamline development and build processes for CI and users


alaviss:
There are a few issues with the compiler development/build processes:

- Requires `PATH` modification to run all tests.
- Requires a bootstrap compiler, but further build invocations will replace it, hiding bootstrap bugs until they hit CI. This might also frustrate developers who start from an old tree cluttered with build artifacts.
- It is a PITA to build the bootstrap compiler.
- Build artifacts are littered all over the tree.

This issue tracks the efforts in reducing friction and increasing build reliability.

- [x] #85 
- [x] Remove bootstrapping (implicit) dependency on users configuration
- [x] Remove main test suite dependency on `nim` in `PATH`
- [ ] #126
- [x] Easy generation of a compiler source archive
- [x] Easy generation of a release compiler binary archive

alaviss:
Per discussion with @haxscramper, we might want to disable the use of user configuration altogether in compiler bootstrap.

Is this desirable and should we take a step further and disable it for the compiler source tree altogether?

@saem @disruptek

saem:
Very much agree.

I think configuration during bootstrap can't be a free for all. We should ratchet it down to zero as a start.

The sinking feeling I have is that were going to need some version over time crap one fine day, because mistakes will have been made. So the lock down now reduces that chance, or at least reduces the complexity.

alaviss:
With the recently merged PRs this is pretty much done.

Consolidating build artifacts will be a rather complicated endeavor, given that a lot of our tools rely on being exactly where they are in the source tree. Work on this portion will need a lot of discussion, so it should be left for later.

PR #100, opened 2021-12-07
koch.nim: ignore user configs for bootstrapping

alaviss:
This is a wildcard that can easily make or break a compiler build. It is
best to leave it out.

In the future we might expand this to tool builds.

Ref https://github.com/nim-works/nimskull/issues/89#issuecomment-985013197


alaviss:
I'm not expecting this to be too much of a change

bors r+

PR #103, opened 2021-12-07
nimpretty/tester: always use built nimpretty

alaviss:
There are little reasons to use a nimpretty in PATH. The test suite
would only work for the nimpretty built from the same source.

This also removes the "feature" that automatically builds nimpretty
(which also depends on the nim in PATH, which might not be able to build
this at all). In the future this might return via koch.

Ref #89

### DIFF
--- a/nimpretty/tester.nim
+++ b/nimpretty/tester.nim
@@ -10,12 +10,7 @@ const
 var
   failures = 0
 
-when defined(develop):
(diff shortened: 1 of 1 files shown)


saem (approved):
"feature" indeed.

Hmm I think I'm starting to see the prerequisite/build dag inspiration.

alaviss:
Hopefully this works now

bors r+

PR #104, opened 2021-12-07
nimsuggest: prefer stdlib in the same distribution

alaviss:
This is still a horrible hack, but now it will be a tiny bit more robust
and likely to pair with the stdlib it was built for.

This also allow nimsuggest tester to run without `nim` in PATH.

Ref #89


saem (approved):
I feel less bad about this than the alternative. 😅

Bors r+

alaviss:
Server crashed earlier, retrying

bors r+

PR #105, opened 2021-12-07
tosproc: completely remove compiler execution

alaviss:
Previously this test would depend on the compiler being available to execute to recompile itself.

Not only is this hugely inefficient, but it also produce a dependency on the compiler being in `PATH`.

Instead this pull request:

- Fold external runners into the same test binary. This removes the need of recompilation altogether.
- Replaced `nim r` in tests with much simpler "consume all" runner that allow testing the same thing. This prevents compiler bugs from tripping this test up.

Done as part of #89 to remove `PATH` dependency from test suite.

### DIFF
--- a/tests/stdlib/tosproc.nim
+++ b/tests/stdlib/tosproc.nim
@@ -4,16 +4,19 @@ joinable: false
 
 #[
 joinable: false
-because it'd need cleanup up stdout
+because it has to execute itself
 
 see also: tests/osproc/*.nim; consider merging those into a single test here
 (easier to factor and test more things as a single self contained test)
 ]#
 
-when defined(case_testfile): # compiled test file for child process
-  from posix import exitnow
+import stdtest/[specialpaths, unittest_light]
+import std/[os, osproc, streams, strtabs, strutils]
+from posix import exitnow
+
+proc exitTest(args: openArray[string]) =
+  ## Test various process exit methods?
   proc c_exit2(code: c_int): void {.importc: "_exit", header: "<unistd.h>".}
-  import os
   var a = 0
   proc fun(b = 0) =
     a.inc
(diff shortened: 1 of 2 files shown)


tests/stdlib/tosproc.nim
```diff
@@ -4,60 +4,60 @@ joinable: false
...
+    if true: quit(0)
+  of "exitnow_139":
+    if true: exitnow(139)
+  of "c_exit2_139":
+    if true: c_exit2(139)
+  of "quit_139":
+    # `exitStatusLikeShell` doesn't distinguish between a process that
+    # exit(139) and a process that gets killed with `SIGSEGV` because
+    # 139 = 11 + 128 = SIGSEGV + 128.
+    # However, as #10249 shows, this leads to bad debugging experience
```
saem: Should we try to add repo qualifiers for the bug references?

tests/stdlib/tosproc.nim
```diff
@@ -4,60 +4,60 @@ joinable: false
 
 #[
 joinable: false
-because it'd need cleanup up stdout
+because it has to execute itself
```
saem, line 7: I think this would be better if the runner supported an array argument sets to pass a test so it could call it repeatedly instead. Although in this case we'd need more for the whole input output bit? Also scope creep.

saem (approved):
Minor comments, I think I get it. It's better than before, that's for sure.

tests/stdlib/tosproc.nim
alaviss: This test tests `osproc`, and it executes itself with `osproc.*` to run its subprogram to verify behaviors. It has nothing to do with having to run with different arguments :P

saem: Wait, the sub-programs are itself, right? Or did I read that wrong?

Anyhow, I was considering it for each of the permutation that it would run, would still need to have other processes and all that. Just that the reporting could be a little bit more fine grained.

alaviss: Ah yea, we can certainly add a runner <-> test IPC in the future.

PR #110, opened 2021-12-08
introduce compiler build reproducibility test to CI

alaviss:
This check makes sure that the compiler can be built reproducibly on the same environment.

To save time, this check is only run on merge via bors.

### DIFF
--- a/.github/workflows/reproducible.yml
+++ b/.github/workflows/reproducible.yml
@@ -1,0 +1,73 @@
+name: Test compiler build reproducibility
+
+on:
+  push:
+    # Only run on bors branches
+    branches:
+      - staging
+      - trying
+
+jobs:
+  reprotest:
+    # This job is meant for testing whether the compiler can be built
+    # reproducibly given the same build environment.
+    #
+    # There are two tools used for this test:
+    #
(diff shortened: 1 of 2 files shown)


alaviss:
bors try

alaviss:
bors try

saem:
For the dummies (me) who might be as of a sudden containted with trying to make sense of this it might help to gave a few comments about what diffoscope and reprotest are and with a rather narrow focus on how we use them than be overly general.

Otherwise lgtm.

alaviss:
> it might help to gave a few comments about what diffoscope and reprotest are and with a rather narrow focus on how we use them than be overly general.

How would you want me to do this? Like document them in the workflow file or?

saem:
> > it might help to gave a few comments about what diffoscope and reprotest are and with a rather narrow focus on how we use them than be overly general.
> 
> How would you want me to do this? Like document them in the workflow file or?

Yeah, just inline. It's where people will encounter it, so don't have to go anywhere, just to start getting them oriented.

alaviss:
bors try

alaviss:
bors try-
bors try

saem (approved):
Bors r+

PR #111, opened 2021-12-08
Support `:key value`  for sexp parser

haxscramper:
- CHANGED ::
  - Allow symbols to actually contain any characters except for `()` -
    current parser cannot even properly handle `kebab-case`.
  - "space" is now parsed more robustly - instead of skipping single
    whitespace character and erroring out on newlines it uses `while`
  - Improved tests a little
- ADDED ::
  - Support `(Fun :key value)` syntax for the s-expression parser
  - Add necessary constructor procs, `convertSexp()` now supports `key =
    value` syntax for construction.

Regardless of whether we will use S-expressions in tests (https://github.com/nim-works/nimskull/discussions/60#discussioncomment-1766166) S-expression parser could be improved to actually use symbol as defined in most lisps (not only "ident characters + underscore"). Changed error messages to be more useful (maybe guidelines should include something like "invalid" __**MUST**__ always be followed by an explanation of why something is invalid `"expected X"` -> `"expected X but found Y"`)

### DIFF
--- a/nimsuggest/sexp.nim
+++ b/nimsuggest/sexp.nim
@@ -15,11 +15,13 @@ import
 import std/private/decode_helpers
 
 type
-  SexpEventKind* = enum  ## enumeration of all events that may occur when parsing
+  SexpEventKind* = enum  ## enumeration of all events that may occur when
+                         ## parsing
     sexpError,           ## an error occurred during parsing
     sexpEof,             ## end of file reached
     sexpString,          ## a string literal
     sexpSymbol,          ## a symbol
+    sexpKeyword,         ## `:keyword` event
     sexpInt,             ## an integer literal
     sexpFloat,           ## a float literal
     sexpNil,             ## the value ``nil``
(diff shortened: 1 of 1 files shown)


saem:
QQ: is there a sexp spec in the broader works that you based this on? I know lispers of the world would have us believe it's all roses, but curious if there is a variant you referenced?

haxscramper:
> QQ: is there a sexp spec in the broader works that you based this on? I know lispers of the world would have us believe it's all roses, but curious if there is a variant you referenced?

Due to reader macros etc. I don't think there is a fixed EBNF grammar for lisp that I can refer to, but http://www.lispworks.com/documentation/HyperSpec/Body/02_cd.htm defines symbol as 

>  Any token that is not a potential number, does not contain a package marker, and does not consist entirely of dots will always be interpreted as a symbol. Any token that is a potential number but does not fit the number syntax is a reserved token and has an implementation-dependent interpretation. In all other cases, the token is construed to be the name of a symbol.

saem:
Ok, can you pop a link to it as reference in the code if not present; tune up the commit message based on discussion in chat, no "feature"; then I'm good with it.

saem (approved):
See previous comment about minor changes, otherwise good to go.

PR #114, opened 2021-12-09
move metadata retrivial from compiler source to koch

alaviss:
This metadata is only correct in the context of the compiler source. If the compiler is used as a library in an another location, then this information would be incorrect.

koch is a better location for this as it is guaranteed to run in the compiler source.

### DIFF
--- a/compiler/commands.nim
+++ b/compiler/commands.nim
@@ -87,10 +87,11 @@ proc writeFullhelp(conf: ConfigRef; pass: TCmdLinePass) =
 
 proc getNimSourceData(): tuple[hash, date: string] {.compileTime.} =
   ## Retrieve metadata about the compiler source code.
-  let hashCall = gorgeEx("git rev-parse --verify HEAD")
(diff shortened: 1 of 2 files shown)


compiler/commands.nim
```diff
@@ -87,10 +87,11 @@ proc writeFullhelp(conf: ConfigRef; pass: TCmdLinePass) =
 
 proc getNimSourceData(): tuple[hash, date: string] {.compileTime.} =
   ## Retrieve metadata about the compiler source code.
-  let hashCall = gorgeEx("git rev-parse --verify HEAD")
-  let dateCall = gorgeEx("git log -1 --format=%cs HEAD")
-  if hashCall.exitCode == 0 and dateCall.exitCode == 0:
-    result = (hashCall.output.strip(), dateCall.output.strip())
+  const
```
saem, line 90: Things like these really make me wish modules were more first class, so we could pass parameters and such, this is correct, the module doesn't know it's usage so it shouldn't try and figure it out.

saem (approved):
Bors r+

PR #115, opened 2021-12-09
niminst: let the nim compiler be specified by koch

alaviss:
By default `niminst` uses the compiler in `PATH`, which might not be compatible with the source it is compiling. This causes csources generation to fail for nimskull as the compiler was incompatible.

This commit enable the exact compiler to be specified by koch, solving this issue.

### DIFF
--- a/tools/niminst/niminst.nim
+++ b/tools/niminst/niminst.nim
@@ -60,6 +60,7 @@ type
     explicitPlatforms: bool
     vars: StringTableRef
     app: AppType
+    nimExe: string
     nimArgs: string
     debOpts: TDebOptions
     nimblePkgName: string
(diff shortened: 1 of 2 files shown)


alaviss:
One step closer to nightlies

bors r+

tools/niminst/niminst.nim
```diff
@@ -193,10 +195,14 @@ proc parseCmdLine(c: var ConfigData) =
...
         if idx < 0: quit("invalid command line")
         c.vars[substr(val, 0, idx-1)] = substr(val, idx+1)
+      of "nim":
+        c.nimExe = val
       else: quit(Usage)
     of cmdEnd: break
   if c.infile.len == 0: quit(Usage)
   if c.mainfile.len == 0: c.mainfile = changeFileExt(c.infile, "nim")
+  # Defaults to `nim` in `PATH` if nothing is specified
+  if c.nimExe.len == 0: c.nimExe = "nim"
```
saem, line 205: Does this have to takeOS into account and add the exe for widows in the else branch?

alaviss: nope, `cmd.exe` handles that part during `PATH` search

PR #118, opened 2021-12-10
rework release archive creation

alaviss:
This is a big change, but most of it circles around the model of archive
creation embedded in niminst.

Previously niminst have two modes of operations: either it creates a
binary zip archive for Windows, or it creates a source tar.xz archive
for Unix.

This commit refactored that into three new pieces:

- `archive`: The main command to create a release archive. By default
             this archive contains only source code, and the source
             archive can be used to create binary archives.
- `--format`: A switch allowing selection of the release archive format.
- `--binaries`: A switch that set archive build mode to "binary
                archive", where either windows or unix (or both)
                binaries will be included in the archive. In this mode,
                files not relevant to the target platforms will also be
                omitted.

With these, koch is now equipped with some "new" commands:

- `archive`: Create a source archive
- `winrelease`: Create a binary zip archive for Windows
- `unixrelease`: Create a binary tarball for Unix-like

Some are removed:

- `testinstall`: Automated test to see if Unix binaries can be built
                 from sources. Currently have too many assumptions on
                 the way the old architecture work, so it is axed for
                 now.
- `zip`: superseded by `winrelease`
- `xz`: superseded by `archive`

And finally, `installer.ini` is updated to synchronize the binary list
with what is actually built.

Depends on #117

### DIFF
--- a/compiler/installer.ini
+++ b/compiler/installer.ini
@@ -36,6 +36,7 @@ App: Console
 License: "copying.txt"
 
 [Config]
+Files: "config/build_config.txt"
 Files: "config/*.cfg"
 Files: "config/config.nims"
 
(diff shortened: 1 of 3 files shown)


saem (approved):
Bors r+

PR #124, opened 2021-12-11
store extra metadata in release archives

alaviss:
This allows koch to fill in commit/commit date data when compiling the compiler from a source archive.

A prerequisite for nightlies.

We are getting deep into the `niminst` is compiler-only territory with this. I hope that in the future I can make #123 happen and consolidate everything into koch.

### DIFF
--- a/compiler/installer.ini
+++ b/compiler/installer.ini
@@ -4,6 +4,8 @@
 [Project]
 Name: "Nim"
 Version: "$version"
+Commit: "$commit"
+CommitDate: "$commitdate"
 Platforms: """
   windows: i386;amd64
   linux: i386;hppa;ia64;alpha;amd64;powerpc64;arm;sparc;sparc64;m68k;mips;mipsel;mips64;mips64el;powerpc;powerpc64el;arm64;riscv32;riscv64
--- a/tools/koch/koch.nim
+++ b/tools/koch/koch.nim
@@ -24,7 +24,7 @@ when defined(amd64) and defined(windows) and defined(vcc):
 when defined(i386) and defined(windows) and defined(vcc):
   {.link: "icons/koch-i386-windows-vcc.res".}
 
-import std/[os, strutils, parseopt, osproc]
+import std/[json, os, strutils, parseopt, osproc]
   # Using `std/os` instead of `os` to fail early if config isn't set up properly.
   # If this fails with: `Error: cannot open file: std/os`, see
   # https://github.com/nim-lang/Nim/pull/14291 for explanation + how to fix.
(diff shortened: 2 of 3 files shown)


compiler/installer.ini
```diff
@@ -4,6 +4,8 @@
 [Project]
 Name: "Nim"
 Version: "$version"
+Commit: "$commit"
```
saem, line 7: worth adding repository here?

saem: Change the author below?

compiler/installer.ini
```diff
@@ -4,6 +4,8 @@
 [Project]
```
saem, line 4: s/Nim/NimSkull

tools/koch/koch.nim
```diff
@@ -219,8 +225,9 @@ proc archive(args: string) =
   ensureCleanGit()
   nimexec("cc -r $2 --var:version=$1 --var:mingw=none --main:compiler/nim.nim scripts compiler/installer.ini" %
        [VersionAsString, compileNimInst])
-  exec("$# --var:version=$# --var:mingw=none --main:compiler/nim.nim --format:tar.xz $# archive compiler/installer.ini" %
-       ["tools" / "niminst" / "niminst".exe, VersionAsString, args])
+  let (commit, date) = getSourceMetadata()
+  exec("$# --var:version=$# --var:mingw=none --var:commit=$# --var:commitdate=$# --main:compiler/nim.nim --format:tar.xz $# archive compiler/installer.ini" %
```
saem, line 229: long line, but I'll live.

tools/koch/koch.nim
```diff
@@ -134,15 +134,21 @@ proc getSourceMetadata(): tuple[hash, date: string] =
...
+      # obtaining the data from git instead
+      let hashCall = execCmdEx("git -C " & quoteShell(nimSource) & " rev-parse --verify HEAD")
+      if hashCall.exitCode == 0:
+        hash = hashCall.output.strip()
+
+      let dateCall = execCmdEx("git -C " & quoteShell(nimSource) & " log -1 --format=%cs HEAD")
+      if dateCall.exitCode == 0:
+        date = dateCall.output.strip()
 
   result = (hash, date)
```
saem, line 153: so if these two aren't set for whatever reason we're still ok with that?

saem (approved):
nothing I would block over, but have a look if you want to make any adjustments.

tools/koch/koch.nim
alaviss: Yea, other than missing information (that we only display in `--version`), I don't think there is much to worry about

compiler/installer.ini
alaviss: I'm just patching things here and there to get stuff going. In the future I'd get rid of this file and niminst both. Hence I try not to edit too much.

alaviss: An another approach would be to turn these into special niminst parameters instead of this weird "fields in a config" approach, if you think that one would be better.

alaviss:
yea no, I can do this stuff in a different way

bors r-

alaviss:
It might be a hack but it works.

As the builder matures, this hack and niminst should go away.

bors r+

PR #128, opened 2021-12-21
Remove Atlas

juancarlospaco:
### Remove Atlas

* Remove Atlas.


saem (changes requested):
@juancarlospaco thanks for creating a PR. Could you please follow the pr template and commit message guidelines.

Specifically this is removing Atlas and moreover I'd like to know your reasoning behind it.

saem:
Everyone is saying remove Atlas, I'm guessing it's bad. I haven't had a chance to dig in, but the glimpse I got suggested it was possibly on to something.

For posterity's sake I'd like to memorialize the reasoning in the commit message and pr -- so others looking to learn or is trying to remember why can look back and understand. A few sentences as to why we removed it and I think it's good to go.

Thank you for doing this and being patient.

juancarlospaco:
- I think it should be unbundled because is not required for a compiler to work.
- No one is actually using it in production.
- It wont attract any contributors, has exotic design, offers no API, undocumented.
- It is not supported by Koch, you have to compile manually.
- No relevant commits in a long time.
- Less code to maintain.
- If you want to add a new module to stdlib, just copy the file!, KISS.

alaviss (changes requested):
I don't disagree with removing `atlas`, but the commit message has to be changed.

Your commit message has to explain:

- What is Atlas and what was it used (or not used) for?
- One or two sentence about why is it being removed
  - Examples
    - It does not work, and there are no effort in continuation of its development, nor is there any users.
    - Something else already does the job better (please tell what is it)

Please also refer to the commit guidelines [here](https://github.com/nim-works/nimskull/blob/devel/doc/contributing.rst#general-commit-rules)

juancarlospaco:
I do not agree with the unnecessarily convoluted commit formats prerequisites, 
whats the difference between https://github.com/nim-works/nimskull/commit/2c9c6c00491dc2486a168c8073f0e23b4307b59d and https://github.com/nim-works/nimskull/pull/128/commits/4d07a28881895b33fe686bf67380d14dde67f729 ?.
😐

alaviss:
The difference is that https://github.com/nim-works/nimskull/commit/2c9c6c00491dc2486a168c8073f0e23b4307b59d has a reason attached in the commit itself, an another example would be https://github.com/nim-works/nimskull/commit/080e0d1f13dc2c7607f9f212b55175513eae1a49. The reason does not have to be complex, the ones I provided are merely suggestions. You can use the same reasoning you wrote in https://github.com/nim-works/nimskull/pull/128#issuecomment-999640073, but it has to be in the commit message.

I'd recommend including a few words about what Atlas is (this is something that https://github.com/nim-works/nimskull/commit/2c9c6c00491dc2486a168c8073f0e23b4307b59d did not have in the commit but only in the [PR](https://github.com/nim-works/nimskull/pull/98#issue-1072243775), but going forward I believe it is better to keep that crucial data in the commit itself)

We want commits to explain themselves for these reasons:

- It attaches context to the commit, helping people who follows the `git blame` chain or studying changes in the project to understand why a decision was made without jumping between PRs and issues to find out.
- It helps us reviewers understand *your* reasons for the changes.
- It provides structure for people who are trying to understand a big change, especially if that commit was squashed. (This point is not relevant to this PR in particular)

I do understand that it takes effort to write comprehensive commit messages, but what is written here will become the history of the project itself (literally). Keeping relevant contexts will help you and others who come after understand why a change was made at that time.

saem:
The commit guidelines aren't up for debate in this PR, this project's core principle is sustainability.

A history not written for future contributors and maintainer of this project means that we're encumbering them with puzzles of the past. Not to mention increasing the barrier to entry for those attempting to understand and learn.

Most of all, it must be a norm in how we conduct ourselves as it sets an example for others. Please look at the commit history that we've recently established. It's not simply everyone does whatever they want, this is a shared space and our actions impact each other now and those in the future.

As you can see this is very important to us as two of us have written at length in order to explain it in full.

I've rewritten the commit message, admittedly with my own wording, but it should illustrate key elements in terms of legibility and tone. While still being simple bullet points so as to not over burden an author. Feel free to use it as is, edit it, or revise your own.

I'm not expecting everyone to have equal comfort with the English language, and I don't want that to be a barrier. For folks in that boat I and others are happy to help crafting these things if the content is there. 

```
Remove Atlas 

Atlas is meant to pull in dependencies by cloning repositories.
It even works with the nimble file format.

Why remove it?
- it doesn't seem to be used, at all
- it's not required for the compiler, so we can resurrect elsewhere if needed
- the design has quirks and the API isn't documented
- Koch doesn't build it, extra reason why it shouldn't be part of the repo
- no recent commits
- less stuff to maintain
```

saem (approved):

alaviss (approved):
bors r+

PR #130, opened 2021-12-22
Remove DrNim

juancarlospaco:
# Remove DrNim

- Remove DrNim


alaviss (commented):
Right now I disagree with removing Dr. Nim, its code could be useful when we want to make nimskull provable.

juancarlospaco:
I disagree with your disagreement.

saem:
I've looked at drnim code and I see the promise... and the problems.

We do need something to head in the proving direction, so the real question is are we getting rid of dead code or do a few of the tests run and anchor one or two guarantees.

Why even one or two guarantees? Because even something small can anchor the idea and fundamental concepts of the proving direction.

haxscramper:
There are also some error message diagnostics that were added with drnim https://github.com/nim-lang/Nim/blob/7a5314c571ec2c370f63d5fab0afd336e04ec161/compiler/sempass2.nim#L759 and other changes to sempass https://github.com/nim-lang/Nim/commit/3a2697dd731cb8fcfd0d279bb856090eca5028ee also need to be accounted for.

juancarlospaco:
> I've looked at drnim code and I see the promise... and the problems.

But it is empty promises and problems for no real concrete practical reason.

It would be better to just have a proper DbC module in StdLib without hacks in the compiler like D lang has.

haxscramper:
For commit message, you can use your explanation from the https://github.com/nim-lang/RFCs/issues/431

It can be added that correctness prover is an important tool, but preferably it should be implemented as an external application that uses stabilized compiler API rather than hacked-in support into the main compiler.

Right now we don't have necessary resources to support development or maintain drnim, so I also think this removal is justified.

haxscramper:
While I have no objections to the PR itself, commit messages must conform to the minimal requirements as [explained](https://nim-works.github.io/nimskull/contributing.html#pr-and-commit-message-guidelines-commit-message-guidelines) in the contribution guidelines.

In this specific instance, "why" part of the commit message is unclear - as indicated above, this can be solved by simply reusing the explanation from the RFC.

compiler/modulegraphs.nim
```diff
@@ -419,15 +419,15 @@ proc registerModule*(g: ModuleGraph; m: PSym) =
...
-  
+
   initStrTables(g, m)
 
 proc registerModuleById*(g: ModuleGraph; m: FileIndex) =
   registerModule(g, g.packed[int m].module)
 
 proc initOperators*(g: ModuleGraph): Operators =
   # These are safe for IC.
-  # Public because it's used by DrNim.
```
krux02, line 430: when DrNim is removed that means this can become private again?

PR #133, opened 2021-12-27
koch: introduce builder

alaviss:
This is the first implementation of https://github.com/nim-works/nimskull/discussions/123

Builder is a new component enabling build actions to be declared as
opposed to be executed imperatively. This allow for side-effect less
declaration of build pipeline and enable them to be parallelized.

This library is heavily influenced by Meson with a touch of zig build.

This first implementation is incomplete and have a few issues, posted here so everyone can keep track and make suggestions:

- Docs might be out-of-date
- Cannot do bootstrapping
- No CLI interface yet
- Does not silence nim compiler calls (which makes it spammy in parallel
  runs)
- Artifacts are still being placed outside of the build directory
- Target description is not supported yet
- koch integration is kind of hacky
- No tests
- Plenty of dead code
- No `clean` function
- No `install` function

### DIFF
--- a/tools/koch/builder.nim
+++ b/tools/koch/builder.nim
@@ -25,7 +25,7 @@ type
     ## An operand
 
   Instruction {.pure.} = enum
-    ## How to run the builder
+    ## Instruction for a target recipe step, each value describes operand interpretation.
     Noop ## Don't do anything
 
     Build ##\
@@ -43,14 +43,15 @@ type
     ## number of arguments
 
   TargetRecipe = object
-    ## The recipe for the target
+    ## Every target is built via a recipe, which forms a small build language.
     instr: Instruction ## The instruction of the target, how operands are
                        ## interpreted depends on this
-    a: Operand ## First operand
-    b: Operand ## Second operand
+    a: Operand ## First operand, see `Instruction`
+    b: Operand ## Second operand, see `Instruction`
 
   Builder* = object
-    ## The builder
+    ## The builder holds all the necessary state for building a compiler managing source,
+    ## targets, dependencies, inputes, and the recipies that combine them.
     sourceRoot: string ## The root directory of the source code
     buildDir: string ## The directory to store build artifacts
 
@@ -150,7 +151,7 @@ func addInputs(b: var Builder, inputs: varargs[string]): Input =
 
 func addExtraOperands(b: var Builder, operands: varargs[Operand]): Operand =
   ## Add `operands` to the list of extra operands and return the index of the
-  ## first operand added as an operand
+  ## first operand added
   if b.extraOperands.high + operands.len > high(Operand).int:
     raise newException(ValueError):
       "Adding operands will exceed the maximum capacity of " & $high(Operand)
--- a/tools/koch/builder.nim
+++ b/tools/koch/builder.nim
@@ -1,0 +1,438 @@
+#
+#
+#               Builder library for Nim
+#   Copyright (c) 2021 Leorize <leorize+oss@disroot.org>
+#
+# See the file "copying.txt" included in this distribution
+# for details about the copyright.
+
+import std/[os, osproc, parseopt, sequtils, setutils, tables]
+
+type
+  Target* = distinct uint16
+    ## A build target
+
+  Executable* = distinct Target
+    ## A build target producing a Nim executable
+
+  Run* = distinct Target
+    ## A target that runs a command
+
+  Input = uint16
+    ## An input
+
+  Operand = uint16
+    ## An operand
+
+  Instruction {.pure.} = enum
+    ## How to run the builder
+    Noop ## Don't do anything
+
+    Build ##\
+    ## Build a Nim executable, operand `a` is the source input id
+    BuildWithArgs ##\
+    ## Build a Nim executable, operand `a` is the source input id, operand `b`
+    ## is the number of arguments after the source id
+
+    RunBuilt ##\
+    ## Run a binary built by another target, operand `a` is the target id
+    RunBuiltWithArgs ##\
+    ## Run a binary built by another target with extra arguments, operand `a`
+    ## is the target id, `b` is the extra operand id where the entry at `b` is
+    ## the starting argument id in input list and the entry at `b + 1` is the
+    ## number of arguments
+
+  TargetRecipe = object
+    ## The recipe for the target
+    instr: Instruction ## The instruction of the target, how operands are
+                       ## interpreted depends on this
+    a: Operand ## First operand
+    b: Operand ## Second operand
+
+  Builder* = object
+    ## The builder
+    sourceRoot: string ## The root directory of the source code
+    buildDir: string ## The directory to store build artifacts
+
+    compiler: string ## The Nim compiler to be used
+    defaultCompilerArgs: seq[string] ## Default compiler arguments
+
+    names: seq[string] ## Target to name mapping. Its usage depends on the target itself
+
+    recipes: seq[TargetRecipe] ## Recipe for each target
+    dependencies: Table[Target, set[Target]] ## Mapping of target dependencies
+    inputs: seq[string] ## List of string inputs
+
+    extraOperands: seq[Operand] ## List of extra operands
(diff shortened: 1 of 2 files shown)


tools/koch/builder.nim
```diff
@@ -0,0 +1,438 @@
...
+      for newDep in Target(0) ..< target:
+        let dep = sortedOrder[newDep]
+        # If dep was a dependency of the target
+        if dep in deps:
+          # Re-link it to the new ID
+          newDependencies[target].incl newDep
+
+  b.names = newNames
+  b.recipes = newRecipes
+  b.dependencies = newDependencies
```
alaviss, line 234: Dead code

tools/koch/builder.nim
```diff
@@ -0,0 +1,438 @@
...
+    ## number of arguments
+
+  TargetRecipe = object
+    ## The recipe for the target
+    instr: Instruction ## The instruction of the target, how operands are
+                       ## interpreted depends on this
+    a: Operand ## First operand
+    b: Operand ## Second operand
+
+  Builder* = object
```
saem, line 52: I would lead with this, since type sections rearrange -- this gives a nice map of the concepts/data.

tools/koch/builder.nim
```diff
@@ -0,0 +1,438 @@
...
+    ## The builder
```
saem:
```suggestion
    ## The builder holds all the necessary state for building a compiler managing source,
    ## targets, dependencies, inputes, and the recipies that combine them.
```

tools/koch/builder.nim
```diff
@@ -0,0 +1,438 @@
...
+    RunBuilt ##\
+    ## Run a binary built by another target, operand `a` is the target id
+    RunBuiltWithArgs ##\
+    ## Run a binary built by another target with extra arguments, operand `a`
+    ## is the target id, `b` is the extra operand id where the entry at `b` is
+    ## the starting argument id in input list and the entry at `b + 1` is the
+    ## number of arguments
+
+  TargetRecipe = object
+    ## The recipe for the target
```
saem:
```suggestion
  TargetRecipe = object
    ## Every target is built via a recipe, which forms a small build language.
```

I'd put this second.

tools/koch/builder.nim
```diff
@@ -0,0 +1,438 @@
...
+    ## A target that runs a command
+
+  Input = uint16
+    ## An input
+
+  Operand = uint16
+    ## An operand
+
+  Instruction {.pure.} = enum
+    ## How to run the builder
```
saem:
```suggestion
  Instruction {.pure.} = enum
    ## Instruction for a target recipe step, each value describes operand interpretation.
```

tools/koch/builder.nim
```diff
@@ -0,0 +1,438 @@
...
+    ## is the target id, `b` is the extra operand id where the entry at `b` is
+    ## the starting argument id in input list and the entry at `b + 1` is the
+    ## number of arguments
+
+  TargetRecipe = object
+    ## The recipe for the target
+    instr: Instruction ## The instruction of the target, how operands are
+                       ## interpreted depends on this
+    a: Operand ## First operand
+    b: Operand ## Second operand
```
saem:
```suggestion
    a: Operand ## First operand, see `Instruction`
    b: Operand ## Second operand, see `Instruction`
```

tools/koch/builder.nim
```diff
@@ -0,0 +1,438 @@
...
+#               Builder library for Nim
+#   Copyright (c) 2021 Leorize <leorize+oss@disroot.org>
+#
+# See the file "copying.txt" included in this distribution
+# for details about the copyright.
+
+import std/[os, osproc, parseopt, sequtils, setutils, tables]
+
+type
+  Target* = distinct uint16
```
saem, line 12: General remark for DOD: Part of me would consider reserving `Target` and such names for 'view objects' that provide an API over the internals; but the builder isn't really a library right now so I wouldn't worry about it yet.

tools/koch/builder.nim
```diff
@@ -0,0 +1,438 @@
...
+  if b.inputs.high + inputs.len > high(Input).int:
+    raise newException(ValueError):
+      "Adding inputs will exceed the maximum capacity of " & $high(Input)
+
+  result = Input b.inputs.len
+  b.inputs.add inputs
+
+func addExtraOperands(b: var Builder, operands: varargs[Operand]): Operand =
+  ## Add `operands` to the list of extra operands and return the index of the
+  ## first operand added as an operand
```
saem:
```suggestion
  ## first operand added
```

tools/koch/builder.nim
```diff
@@ -0,0 +1,438 @@
...
+        # If dep was a dependency of the target
+        if dep in deps:
+          # Re-link it to the new ID
+          newDependencies[target].incl newDep
+
+  b.names = newNames
+  b.recipes = newRecipes
+  b.dependencies = newDependencies
+
+func name*(b: Builder, target: Target): string =
```
saem, line 236: This is where it might be nice to have `Target` be `TargetId` and `Target` be:
```nim
type
  Target = object
    builder: Builder
    id: TargetId
```

at least from the outside people are dealing with a friendlier concept of a holistic target, while internally the memory layout and data are nice and orderly.

tools/koch/koch.nim
```diff
@@ -243,6 +240,15 @@ proc buildTools(args: string = "") =
       "--opt:speed --stacktrace -d:debug --stacktraceMsgs -d:nimCompilerStacktraceHints --excessiveStackTrace:off " & defineSourceMetadata() & " " & args,
       outputName = "nim_dbg")
 
+proc addTools(b: var Builder) =
+  let tools = [
```
saem, line 244: maybe a `b.buildGroup` call that takes a name and a bunch of `addxxx` calls, returning the name as an alias?

tools/koch/koch.nim
```diff
@@ -287,6 +293,24 @@ proc binArchive(target: BinArchiveTarget, args: string) =
...
+    defaultArgs = [
+      "--var:version=" & VersionAsString,
+      "--var:mingw=none",
+      "--main:" & (nimSource / "compiler" / "nim.nim")
+    ]
+
+  let compilerManifest = nimSource / "compiler" / "installer.ini"
+
+  let buildScript: Target = b.addRun("installscripts", niminst, @defaultArgs & @["scripts", compilerManifest])
+  discard b.addRun("archive", niminst, @defaultArgs & @["--format:tar.xz", "archive", compilerManifest], dependsOn = {buildScript})
```
saem, line 310: few cases where I think discardable makes sense, I don't think it catches errors here; because the dependency passing will force that.

saem (commented):
API looks really good, though I think `addRun` might benefit from a tweak.

tools/koch/builder.nim
alaviss: It will be at one point, but having builder as a `ref` is not something I'd like now due to how easy it is to hide mutations.

saem: Well it could only happen inside module defined procs and if you wanted extra safety the builder could be a distinct to reduce the surface further.

tools/koch/builder.nim
```diff
@@ -0,0 +1,439 @@
...
+func builtExe*(b: Builder, exe: Executable): string =
+  ## Returns the path to the built executable
+  # TODO: Move this to the build directory
+  b.sourceRoot / "bin" / b.name(exe).addFileExt(ExeExt)
+
+func cacheDirectory(b: Builder, exe: Executable): string =
+  ## Returns the path to the nimcache of the executable
+  b.buildDir / "nimcache" / b.name(exe)
+
+func addRun*(b: var Builder, name: string, exe: Executable,
```
saem, line 272: To get transactional behavior:
1. Immediately make a copy of b
2. Wrap the current body in a try
3. manipulate the copy instead of b in the try
4. in the last line of the try copy back the changes over from the copy into b
5. profit ... and by that I mean you probably want to make that a template

PR #135, opened 2021-12-27
Testament: test process batching and internals rework

saem:
Started as introducing knownIssue and disabled, but reworking internals heavily and minor consequences to the interface.

Refactor Consequences:
- introduces a more data oriented design
- simplifies categories
  - remove dead/unnecessary ones (IC/Navigator)
  - remove the megatest category hack
  - likely other changes before finishing
- makes js a first class target
  - now a default execution target
  - it should be explicitly excluded (likely need a feature)

break changes forecast:
- this will necessitate some test/spec clean-up
- might run a fair bit slower without megatest
- although in some areas it's more parallelizeable (always runs in parallel)
- removed ic and navigator categories, need to fix spec parsing 

reviewable things:
- ExecutionState is getting pretty stable
- The batching is somewhat final, but making a bit lazier can avoid IO
  - can't quite add that to the stack of things to change yet
- all, category, and single test all behave the same now
- should we bother with pattern?

Known issues:
- need to get an end to end execution going
- runTests is a beast
- lol docs
- tests for testament

Lessons learned:
Many, biggest one so far is orchestration of dependencies and testing
are separate things and the existing testament conflates these.


testament/testament.nim
```diff
@@ -851,4 +884,437 @@ proc main() =
...
+    parseSuccess       ## successfully parsed cli params
+    parseQuitWithUsage ## parsing failed, quit with usage message
+
+const
+  testResultsDir = "testresults"
+  cacheResultsDir = testResultsDir / "cacheresults"
+  noTargetsSpecified: TestTargets = {}
+  defaultExecFlags = {outputColour}
+  defaultBatchSize = 10
+  noMatrixEntry: EntryId = -1
```
saem: @alaviss as you can see I learned similar lessons with DOD. 😆

alaviss: You really should turn those `int` into `int32`, no one do more than 2M tests yet

saem: Done locally.

haxscramper:
@saem aside from `xxx` todos, maybe failing tests and rebase, is there anything else that needs to be done for this PR? I want to continue this after https://github.com/nim-works/nimskull/pull/476 is merged and initial test directory cleanup is finished.

haxscramper:
Many changes from this PR are already implemented so this can be closed and PRed in a simplified manner to avoid dealing with all merge conflicts.

PR #136, opened 2021-12-28
testament: remove stdlib isMainModule testing

saem:
Presently testament will parse all modules under `pure/lib` and `lib/packages/docutils` then:
1. it would read the contents, looking for a string match of `when isMainModule`
  1. resulting in a compile action instead of run
2. then include these as "tests" within its run

There are only two modules that are making use of this:
* `lib/packages/docutils/highlite.nim`: which was piggybacking a build check
* `lib/pure/xmlparser.nim`: had a basic test case

For the former the build check was integrated into `tools/koch/kochDocs` and for the latter
a quick test was created with `tests/stdlib`. After which point the `lib/packages/docutils` was
removed from testament's stdlib category and now content parsing for `when isMainModule`
has been removed, this should speed up execution for slower disks. Also these are now
used as simple compiles tests instead of attempting a run.

After review I'll squash. The background motivation for this is it simplifies the testament
rework that I'm presently undertaking in: https://github.com/nim-works/nimskull/pull/135/files

### DIFF
--- a/lib/packages/docutils/highlite.nim
+++ b/lib/packages/docutils/highlite.nim
@@ -97,7 +97,7 @@ const
     "Other"]
 
   # The following list comes from doc/keywords.txt, make sure it is
-  # synchronized with this array by running the module itself as a test case.
+  # synchronized with this array -- a check exists within `koch/kochdocs`.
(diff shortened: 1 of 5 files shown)


beef331 (approved):

saem:
Bors r+

PR #137, opened 2021-12-29
niminst: create deterministic archives

alaviss:
This pull makes niminst creates reproducible archives following the
recommendations at https://reproducible-builds.org/docs/archives/

In particular, these are the changes made:

- Make niminst outputs the commands being run and exit on failure if
  any of them fail to run.

- Procedures that walk directories will produce a sorted list to
  minimize undeterminism from OS ordering.

- Archived files have their timestamp set to the commit date if
  available or epoch 0 otherwise to minimize the effect of time on the
  build. The timezone used is always UTC.

- Archived files are stored in the archive in their alphabetical order,
  allowing them to be stored without regards to their ordering in the
  file system.

- Switched from 7-Zip to Info-Zip for creating zip files, as Info-Zip
  allow us to omit extra metadata from the archive that depends on
  undeterministic values.

- For tar archives, the user/group is set to 0 and use the GNU format to
  eliminate non-deterministic metadata. Since the options to perform
  this task is not standard, the `tar` utility support has been reduced
  to libarchive and GNU tar, the two most popular tar utilities.

- Modified reproducibility test to test source and release archive
  generation.

### DIFF
--- a/tools/niminst/niminst.nim
+++ b/tools/niminst/niminst.nim
@@ -7,8 +7,8 @@
 #    distribution, for details about the copyright.
 
 import
-  os, osproc, strutils, parseopt, parsecfg, strtabs, streams, debcreation,
-  std / sha1, json
+  algorithm, os, osproc, strutils, parseopt, parsecfg, strtabs, streams,
+  debcreation, std / sha1, json, times, sequtils
 
 const
   maxOS = 20 # max number of OSes
@@ -284,6 +284,10 @@ proc walkDirRecursively(s: var seq[string], root, explicit: string,
         walkDirRecursively(s, f, explicit, allowHtml)
       of pcLinkToDir: discard
 
+  # Sort alphabetically to make sure that the file system ordering has no
+  # bearing on our generators
+  s.sort()
+
 proc addFiles(s: var seq[string], patterns: seq[string]) =
   for p in items(patterns):
     if dirExists(p):
@@ -297,6 +301,10 @@ proc addFiles(s: var seq[string], patterns: seq[string]) =
           add(s, unixToNativePath(f))
           inc(i)
       if i == 0: echo("[Warning] No file found that matches: " & p)
+
+  # Sort alphabetically to make sure that the file system ordering has no
+  # bearing on our generators
+  s.sort()
 
 proc pathFlags(p: var CfgParser, k, v: string,
                t: var tuple[path, flags: string]) =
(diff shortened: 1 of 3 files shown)


alaviss:
bors try

alaviss:
bors try

alaviss:
bors try-

bors try

alaviss:
bors try

alaviss:
I modified the configuration so have to restart the test

bors try-

bors try

alaviss:
bors try

alaviss:
bors try

alaviss:
bors try

alaviss:
bors try

alaviss:
bors try

alaviss:
bors try

alaviss:
bors try

alaviss:
bors try

alaviss:
bors try

alaviss:
bors try-

bors try

alaviss:
bors try

tools/niminst/niminst.nim
```diff
@@ -749,22 +845,107 @@ proc archiveDist(c: var ConfigData) =
...
+      DefaultPermission = {fpUserRead, fpUserWrite, fpGroupRead, fpOthersRead}
+        ## Permissions applied to files
+      
+      ExecutablePermission = DefaultPermission + {fpUserExec, fpGroupExec, fpOthersExec}
+        ## Permissions applied to executables/directories
+
+    for path in walkDirRec(proj, {pcFile, pcDir}, checkDir = true):
+      if AvoidFragments.allIt(it notin path):
+        # Normalize permissions so that a reasonably stricter/laxer umask won't
+        # cause undeterminism
```
saem: s/undeterminism/nondeterminism

tools/niminst/niminst.nim
```diff
@@ -749,22 +845,107 @@ proc archiveDist(c: var ConfigData) =
...
+        if fpUserExec in getFilePermissions(path):
```
saem: I would prefer to invert this such that only the part of the code that varies is covered by the if, while the rest remains as is.

tools/niminst/niminst.nim
```diff
@@ -672,6 +680,94 @@ proc setupDist2(c: var ConfigData) =
...
+proc verboseExec(args: varargs[string]): int =
+  ## Print and execute command formed by `args`
+  if args.len == 0:
+    raise newException(ValueError):
+      "No command was provided for execution"
+
+  const ProcessOpts = {poUsePath, poParentStreams, poEchoCmd}
+
+  let p =
+    if args.len > 1:
```
saem, line 728: Are there really any savings to be had, can we not simply pass an empty array if there are no args instead?

tools/niminst/niminst.nim
```diff
@@ -749,22 +845,107 @@ proc archiveDist(c: var ConfigData) =
...
+          setFilePermissions(path, DefaultPermission)
+
+        # Set the time of every file/directories to be archived to the
+        # deterministic timestamp
+        setLastModificationTime(path, timestamp)
+
+    # Collect all files to be archived
+    var paths = collect(newSeq):
+      for path in walkDirRec(proj, checkDir = true):
+        if AvoidFragments.allIt(it in path):
```
saem: Ajay have sequtils, why not use filterait?

saem (approved):
Thanks!

Nothing major, just noticed a few things.

tools/niminst/niminst.nim
alaviss: I'm not sure if `toOpenArray` works with empty list, hence the `if`.

tools/niminst/niminst.nim
alaviss: Zipped this one to the loop above it. It was originally an external loop since I wanted to avoid having directories in the list, which would be recursed by the archivers. The switch in archivers let me disable recursive travel so there is no need to filter directories any more.

PR #144, opened 2021-12-31
Experiement: Data Oriented Design AST Sketch

saem:
First pass attempt at drafting a DOD AST based on this discussion:
https://github.com/nim-works/nimskull/discussions/139

Breaks/Issues:
- `ast.transitionX` procs now return a new node
- no more nil means have to use a sentinel value

signature change for transition procs mean we're creating new node
this could well break assumptions that pervade the compiler. Worst case
will need to supprot the mutability for now to get the first port, then
figure out how to start changing up the compiler based on
fork/join/mutation observations.

Blocker(s):
The VM takes address of nodes, but now transitions might mean those are
no longer valid. Even if we rely on id as address, this is likely not
the right semantics, as we want to point to the AST "position", for
example when we're doing constant expression evaluation incrementally

Reviewers:
- meant as an example and discussion point for now in the discussion thread
- It doesn't compile because of VM breakages elaborated above
- ast.nim contains all the real content 
  - other changes are to satisfy the compiler
- **do not merge**

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -34,7 +34,7 @@ type Gconfig = object
   ## we put comments in a side channel to avoid increasing `sizeof(TNode)`,
   ## which reduces memory usage given that `PNode` is the most allocated
   ## type by far.
-  comments: Table[int, string] # nodeId => comment
+  comments: Table[NodeId, string] # nodeId => comment
   useIc*: bool
 
 var gconfig {.threadvar.}: Gconfig
@@ -319,6 +319,7 @@ proc discardSons*(father: PNode)
 type Indexable = PNode | PType
 
 proc len*(n: Indexable): int {.inline.} =
+  ## number of children, unsafe if called on leaf nodes, see `safeLen`
   result = n.sons.len
 
 proc safeLen*(n: PNode): int {.inline.} =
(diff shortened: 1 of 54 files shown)


compiler/ast.nim
```diff
@@ -761,7 +761,62 @@ type
...
+
+  State = ref object
+    nodeList: NodeList        ## each node, NodeId is their sequence index
+    nodeFlag: seq[TNodeFlags] ## flags for each node, rarely accessed bloat
+    nodeInf: seq[TLineInfo]   ## info on a per node basis
+
+    astData: AstTree          ## actual tree structure for the various AST
+
+    # sparse data, not all nodes have these
+    nodeSym: OrderedTableRef[NodeId, PSym]         ## symbols
```
haxscramper: Why use table ref in addition to the state ref? We would be doing triple indirection in order to get the necessary data - (1) unref state, (2) unref table, (3) go to the table data store.

compiler/ast.nim
```diff
@@ -1002,12 +1057,152 @@ type
...
+
+#------------------------------------------------------------------------------
+# DOD AST Draft - End
+#------------------------------------------------------------------------------
 
 type Gconfig = object
   # we put comments in a side channel to avoid increasing `sizeof(TNode)`, which
   # reduces memory usage given that `PNode` is the most allocated type by far.
-  comments: Table[int, string] # nodeId => comment
+  comments: Table[NodeId, string] # nodeId => comment
```
haxscramper: Pretty sure you have already though about this, but we can put comments in  the AST state as well, and make all comment accessor functions side-effect free as well

haxscramper (commented):
Not a real review, just had a couple questions while looking through the code

compiler/ast.nim
saem: Was bashing at something that wasn't working at the time, didn't revert.

compiler/ast.nim
saem: I did, just wanted to get things working first, but the VM is being a jerk.

saem:
resurrected, this was far less painful, it's getting very close to allow for some discussion, just a few clean-up items but that can be somewhat concurrent.

Immediate Items:

- [ ] test fixes
  - [ ] tcompilerapi
  - [ ] tcan_compile_nim
  - [ ] results
  - [ ] ic category
- [ ] it's pretty slow right now
- [ ] clean-up type and report storage, tables are super slow

----

Big thanks to @zerbina their VM rework has been a huge unblocker.

PR #151, opened 2022-01-11
make toSeq not use compiles feature

krux02:
If you think `when compiles(...)` is bad, this should be an improvement.

This code is from my own fork of Nim. But I had to adapt it to "modern" nim. Here `lent` caused problems and I had to include special casing to get rid of it. The rest if as in my old Nim fork.

### DIFF
--- a/lib/pure/collections/sequtils.nim
+++ b/lib/pure/collections/sequtils.nim
@@ -457,6 +457,15 @@ iterator filter*[T](s: openArray[T], pred: proc(x: T): bool {.closure.}): T {.ef
     if pred(s[i]):
       yield s[i]
 
+iterator filter*[T](s: seq[T], pred: proc(x: T): bool {.closure.}): T {.effectsOf: pred.} =
+  # this overload of `filter` is written to work around this bug:
+  # https://github.com/nim-lang/Nim/issues/13417
+  var i = 0
+  while i < len(s):
+    if pred(s[i]):
+      yield s[i]
+    inc(i)
+
 proc filter*[T](s: openArray[T], pred: proc(x: T): bool {.closure.}): seq[T]
                                                                   {.inline, effectsOf: pred.} =
   ## Returns a new sequence with all the items of `s` that fulfill the
@@ -774,48 +783,87 @@ template anyIt*(s, pred: untyped): bool =
       break
   result
 
-template toSeq1(s: not iterator): untyped =
-  # overload for typed but not iterator
-  type OutType = typeof(items(s))
-  when compiles(s.len):
-    block:
-      evalOnceAs(s2, s, compiles((let _ = s)))
-      var i = 0
-      var result = newSeq[OutType](s2.len)
-      for it in s2:
-        result[i] = it
-        i += 1
-      result
-  else:
-    var result: seq[OutType] = @[]
-    for it in s:
-      result.add(it)
-    result
+template toSeqBuiltin[T](arg: openArray[T]): seq[T] = @arg
 
-template toSeq2(iter: iterator): untyped =
-  # overload for iterator
-  evalOnceAs(iter2, iter(), false)
-  when compiles(iter2.len):
-    var i = 0
-    var result = newSeq[typeof(iter2)](iter2.len)
-    for x in iter2:
-      result[i] = x
-      inc i
-    result
-  else:
-    type OutType = typeof(iter2())
-    var result: seq[OutType] = @[]
-    when compiles(iter2()):
-      evalOnceAs(iter4, iter, false)
-      let iter3 = iter4()
-      for x in iter3():
-        result.add(x)
-    else:
-      for x in iter2():
-        result.add(x)
-    result
+proc toSeqBuiltin[T](arg: set[T]): seq[T] {.inline.} =
+  for it in arg:
+    result.add it
 
-template toSeq*(iter: untyped): untyped =
+proc toSeqBuiltin(a: cstring): seq[char] {.inline.} =
+  let len = a.len
+  result.setLen len
+  var i = 0
+  while i < len:
+    result[i] = a[i]
+    inc i
+
+proc toSeqBuiltin[E: enum](t: typedesc[E]): seq[E] =
+  for it in t:
+    result.add it
+
+macro toSeqBuiltin(arg: iterator): untyped =
+  let typ = arg.getTypeInst
+  typ.expectKind nnkProcTy # only for closure iterators
+  let formalParams = typ[0]
+  formalParams.expectKind nnkFormalParams
+  formalParams.expectMinLen 1
+  let resultTyp = nnkBracketExpr.newTree(bindSym"seq", formalParams[0])
+  let tmpSym = genSym(nskVar)
+  result = nnkStmtListExpr.newTree
+  result.add nnkVarSection.newTree(nnkIdentDefs.newTree(tmpSym, resultTyp, newNimNode(nnkEmpty)))
+  result.add quote do:
+    for it in `arg`:
+      `tmpSym`.add it
+  result.add tmpSym
+
+template toSeqBuiltin[T: not proc](arg: T): untyped =
+  # fallback for ``IntSet`` and ``HashSet``. Explicit overload would
+  # be better, but this would require an additional import here.
+  var tmp: seq[typeof(items(arg), typeOfIter)]
+  for it in arg:
+    tmp.add it
+  tmp
+
+proc isSystemItems(arg: NimNode) : bool =
+  ## Return true if `arg` is the `items` iterator symbol defined in stdlib.
+  if arg.kind == nnkSym and arg.symKind == nskIterator:
+    if arg.eqIdent("items"):
+      if arg.owner.owner.eqIdent("stdlib"):
+        return true
+  return false
+
+macro toSeqImpl(arg: typed): untyped =
+  let iteratorCall = arg[1]
+  if isSystemItems(iteratorCall[0]):
+    let iteratorArg = iteratorCall[1]
+    result = newCall(bindSym"toSeqBuiltin", iteratorArg)
+  else:
+    var typ = iteratorCall.getTypeInst()
+    # `lent` is broken feature. and this special handling to get rid of lent is why it is broken.
+    if typ.kind == nnkBracketExpr and typ.len == 2 and typ[0].eqIdent("lent"):
+      # TODO this is not guaranteed to be the acutal system `lent` symbol.
+      typ = typ[1]
+
+    result = quote do:
+      var tmp: seq[`typ`]
+      for it in `iteratorCall`:
+        tmp.add it
+      tmp
+
+proc genCallToSeqImpl(arg: NimNode): NimNode =
+  newCall(bindSym"toSeqImpl",
+    nnkForStmt.newTree(
+      ident"x", arg, newStmtList()))
+
+macro toSeqImplSingleSym(arg: typed): untyped =
+  if arg.kind != nnkSym:
+    error "identifier not unique", arg # arg is symChoice
+  elif arg.symKind == nskIterator:
+    result = genCallToSeqImpl(newCall(arg))
+  else:
+    result = genCallToSeqImpl(arg)
+
+macro toSeq*(arg: untyped): untyped =
   ## Transforms any iterable (anything that can be iterated over, e.g. with
   ## a for-loop) into a sequence.
   ##
@@ -832,26 +880,10 @@ template toSeq*(iter: untyped): untyped =
     assert mySeq1 == @[1, 2, 3, 4, 5]
     assert mySeq2 == @[1'i8, 3, 5]
 
-  when compiles(toSeq1(iter)):
-    toSeq1(iter)
-  elif compiles(toSeq2(iter)):
-    toSeq2(iter)
+  if arg.kind in {nnkSym, nnkIdent, nnkOpenSymChoice, nnkClosedSymChoice}:
+    result = newCall(bindSym"toSeqImplSingleSym", arg)
   else:
-    # overload for untyped, e.g.: `toSeq(myInlineIterator(3))`
-    when compiles(iter.len):
-      block:
-        evalOnceAs(iter2, iter, true)
-        var result = newSeq[typeof(iter)](iter2.len)
-        var i = 0
-        for x in iter2:
-          result[i] = x
-          inc i
-        result
-    else:
-      var result: seq[typeof(iter)] = @[]
-      for x in iter:
-        result.add(x)
-      result
+    result = genCallToSeqImpl(arg)
 
 template foldl*(sequence, operation: untyped): untyped =
   ## Template to fold a sequence from left to right, returning the accumulation.


lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +774,86 @@ template anyIt*(s, pred: untyped): bool =
...
+    result[i] = a[i]
+    inc i
+
+proc toSeqBuiltin[E: enum](t: typedesc[E]): seq[E] =
+  for it in t:
+    result.add it
+
+macro toSeqBuiltin(arg: iterator): untyped =
+  let typ = arg.getTypeInst
+  typ.expectKind nnkProcTy # only for closure iterators
```
saem, line 806: Should this be guarded in the JS case?

Just thinking about closure iterators being unsupported there.

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +774,86 @@ template anyIt*(s, pred: untyped): bool =
...
+  result.add nnkVarSection.newTree(nnkIdentDefs.newTree(tmpSym, resultTyp, newNimNode(nnkEmpty)))
+  result.add quote do:
+    for it in `arg`:
+      `tmpSym`.add it
+  result.add tmpSym
+
+template toSeqBuiltin[T: not proc](arg: T): untyped =
+  # fallback for ``IntSet`` and ``HashSet``. Explicit overload would
+  # be better, but this would require an additional import here.
+  var tmp: seq[typeof(items(arg), typeOfIter)]
```
saem, line 822: Seeing as we're in sequtils, wouldn't it be appropriate to map this? 😁

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +774,86 @@ template anyIt*(s, pred: untyped): bool =
...
+  for it in arg:
+    tmp.add it
+  tmp
+
+macro toSeqImpl(arg: typed): untyped =
+  let iteratorCall = arg[1]
+  let iteratorSym = iteratorCall[0]
+  if iteratorCall.len == 2 and iteratorSym.eqIdent("items") and iteratorSym.owner.owner.eqIdent("stdlib"):
```
saem: Minor: having a let for iteratorGrandParent or something, would trim the length if clause and each term would read at the same conceptual level.

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +774,86 @@ template anyIt*(s, pred: untyped): bool =
...
-      for x in iter3():
-        result.add(x)
-    else:
-      for x in iter2():
-        result.add(x)
-    result
+    var typ = iteratorCall.getTypeInst()
+    # `lent` is broken feature. and this special handling to get rid of lent is why it is broken.
+    if typ.kind == nnkBracketExpr and typ.len == 2 and typ[0].eqIdent("lent"):
+      # TODO this is not guaranteed to be acutally be the system `lent` symbol.
```
saem:
```suggestion
      # TODO this is not guaranteed to be the actual system `lent` symbol.
```

saem (changes requested):
Mostly the typo in the todo, but it's easier to follow than before.

saem:
For the commit message could you update it?

Commit message guide is here: https://nim-works.github.io/nimskull/contributing.html#pr-and-commit-message-guidelines-commit-message-guidelines

tl;dr:

```
where: what? 50 chars or less

*why*? format to 72 chars width

Also consider writing about "how?" and/or more elaborate explanation of
"what?"
```

These are the sorts of audience/scenario we're thinking about:
- changelog technical writer
- early contributors that are learning from commits
- dev trying to debug an issue
- sometime trying to figure out why some decision got made

That sort of thing.

lib/pure/collections/sequtils.nim
krux02: you mean literally use `sequtils.map`? Well maybe, why not. I'll see if it causes problems.

lib/pure/collections/sequtils.nim
krux02: I think I would rather have a separate proc like `isStdlib` and then use that to shorten the line.

lib/pure/collections/sequtils.nim
saem: Sure, I'm also just being a bit cute, so feel free to not do it.

lib/pure/collections/sequtils.nim
```diff
@@ -457,6 +457,13 @@ iterator filter*[T](s: openArray[T], pred: proc(x: T): bool {.closure.}): T {.ef
     if pred(s[i]):
       yield s[i]
 
+iterator filter*[T](s: seq[T], pred: proc(x: T): bool {.closure.}): T {.effectsOf: pred.} =
```
alaviss, line 460: Need docs for this one.

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +781,86 @@ template anyIt*(s, pred: untyped): bool =
...
+macro toSeqBuiltin(arg: iterator): untyped =
+  let typ = arg.getTypeInst
+  typ.expectKind nnkProcTy # only for closure iterators
+  let formalParams = typ[0]
+  formalParams.expectKind nnkFormalParams
+  formalParams.expectMinLen 1
+  let resultTyp = nnkBracketExpr.newTree(bindSym"seq", formalParams[0])
+  let tmpSym = genSym(nskVar)
+  result = nnkStmtListExpr.newTree
+  result.add nnkVarSection.newTree(nnkIdentDefs.newTree(tmpSym, resultTyp, newNimNode(nnkEmpty)))
```
alaviss, line 813: Some styling to add clarity to the code.

```suggestion
  result.add:
    nnkVarSection.newTree:
      nnkIdentDefs.newTree(tmpSym, resultTyp, newEmptyNode())
```

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +781,86 @@ template anyIt*(s, pred: untyped): bool =
...
+  for it in t:
+    result.add it
+
+macro toSeqBuiltin(arg: iterator): untyped =
+  let typ = arg.getTypeInst
+  typ.expectKind nnkProcTy # only for closure iterators
+  let formalParams = typ[0]
+  formalParams.expectKind nnkFormalParams
+  formalParams.expectMinLen 1
+  let resultTyp = nnkBracketExpr.newTree(bindSym"seq", formalParams[0])
```
alaviss, line 810: Wouldn't this explode spectacularly if generics are used?

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +781,86 @@ template anyIt*(s, pred: untyped): bool =
...
+  let tmpSym = genSym(nskVar)
+  result = nnkStmtListExpr.newTree
+  result.add nnkVarSection.newTree(nnkIdentDefs.newTree(tmpSym, resultTyp, newNimNode(nnkEmpty)))
+  result.add quote do:
+    for it in `arg`:
+      `tmpSym`.add it
```
alaviss, line 816: `:` can be very ambiguous sometimes, so it's better to split it away from command call.

```suggestion
  result.add:
    quote do:
      for it in `arg`:
        `tmpSym`.add it
```

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +781,86 @@ template anyIt*(s, pred: untyped): bool =
...
+    # `lent` is broken feature. and this special handling to get rid of lent is why it is broken.
+    if typ.kind == nnkBracketExpr and typ.len == 2 and typ[0].eqIdent("lent"):
+      # TODO this is not guaranteed to be the acutal system `lent` symbol.
+      typ = typ[1]
+
+    result = quote do:
+      var tmp: seq[`typ`]
+      for it in `iteratorCall`:
+        tmp.add it
+      tmp
```
alaviss, line 851: It might be better to rope `sugar.collect` in instead, its impl is not perfect but it can do all this type inference properly.

alaviss (changes requested):

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +781,86 @@ template anyIt*(s, pred: untyped): bool =
...
-        result.add(x)
-    else:
-      for x in iter2():
-        result.add(x)
-    result
+    var typ = iteratorCall.getTypeInst()
+    # `lent` is broken feature. and this special handling to get rid of lent is why it is broken.
+    if typ.kind == nnkBracketExpr and typ.len == 2 and typ[0].eqIdent("lent"):
+      # TODO this is not guaranteed to be the acutal system `lent` symbol.
+      typ = typ[1]
```
alaviss, line 845: You can use this to get the return type of any iterable expression:

```nim
template iterRetTyp(u: untyped): untyped =
  proc inferer(): auto {.genSym.} =
    for it in u:
      return it

  typeof(inferer())

iterator test[T](): T =
  yield default(T)

echo iterRetTyp(test[int]())
echo iterRetTyp(['a', 'b', 'c'].items)
```

With this you basically hand the job to the compiler, so it's always correct.

lib/pure/collections/sequtils.nim
Vindaar: Damn, I keep forgetting this is valid syntax 👀

lib/pure/collections/sequtils.nim
beef331: Should this not be `openArray[T]`?

lib/pure/collections/sequtils.nim
krux02: not sure, can you provide an example that could cause problems?

lib/pure/collections/sequtils.nim
krux02: an overload for `filter` with `openArray` already exists. It also already has the documentation needed. Not sure if Nimworks is different here, but IIRC sets of overloads that do the same thing don't all need a copy paste documentation string. Or do they now?

But yes, documentation needs to be added for this one. Not for what it does (obvious), but for why it is even necessary and why the `openArray` alone doesn't cut it. This is the situation:

```nim
for it in myiterator(expr()):
  discard
```

When myiterator takes an ``openArray`` argument, ``expr`` is reevaluated on each loop cycle. There are test cases that ensure on ``toSeq`` to not have side effects. In essence I am working around a Nim compiler bug here.

If you have suggestions on what I should write, I am open for suggestions. I don't know how do document these things.

lib/pure/collections/sequtils.nim
krux02: Actually I've written a lot of code in this concise way (command call). I think it is very readable. Especially this exact pattern `result.add quote do:` is a pattern I've used a lot in the past. I think my opengl sandbox project is full of it.

Long story short, I'd prefer not not split the line here.

lib/pure/collections/sequtils.nim
krux02: can you provide an example? I am asking, because also for me it was quite some time age since I originally wrote this rewrite of `toSeq`, it is just recent that I contributed it to nimskull.

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +781,86 @@ template anyIt*(s, pred: untyped): bool =
...
+    result = quote do:
+      var tmp: seq[`typ`]
+      for it in `iteratorCall`:
+        tmp.add it
+      tmp
+
+proc genCallToSeqImpl(arg: NimNode): NimNode =
+  newCall(bindSym"toSeqImpl",
+    nnkForStmt.newTree(
+      ident"x", arg, newStmtList()))
```
alaviss, line 856: Some structuring to improve clarity:

```suggestion
  newCall(bindSym"toSeqImpl"):
    nnkForStmt.newTree(ident"x", arg):
      newStmtList()
```

lib/pure/collections/sequtils.nim
alaviss: It's more about how the compiler sometimes bind the `do:` to the wrong callee, which I tripped on a few times while developing CPS.

It's fine if you left it as-is in this case.

lib/pure/collections/sequtils.nim
alaviss: I couldn't get that part to trip up, which is great, but I caused some other code to fail, though.

lib/pure/collections/sequtils.nim
alaviss: I'd say it's not a concern.

lib/pure/collections/sequtils.nim
alaviss: I think you can use `typeKind` to identify the type of `lent` and/or `var` to skip as well.

lib/pure/collections/sequtils.nim
alaviss: You should document that it is a workaround for a compiler bug (please also file an issue for that bug) and hyperlink it to the other iterator, ie.

> Overload of \<link this\>`filter(openArray[T]) iterator` to workaround a compiler bug...

alaviss: With that said, it seems like you are not using this new iterator in this PR. Can you split this part out to a separate PR instead?

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +781,86 @@ template anyIt*(s, pred: untyped): bool =
...
-      for it in s2:
-        result[i] = it
-        i += 1
-      result
-  else:
-    var result: seq[OutType] = @[]
-    for it in s:
-      result.add(it)
-    result
+template toSeqBuiltin[IX, T](arg: array[IX, T]): seq[T] = @arg
```
alaviss: Wouldn't the `openArray` version perform the job of this one?

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +781,86 @@ template anyIt*(s, pred: untyped): bool =
...
-    var result = newSeq[typeof(iter2)](iter2.len)
-    for x in iter2:
-      result[i] = x
-      inc i
-    result
+template toSeqBuiltin[T](arg: openArray[T]): seq[T] = @arg
+
+template toSeqBuiltin(arg: string): seq[char] = @arg
+
+template toSeqBuiltin[T](arg: seq[T]): seq[T] = arg
```
alaviss: Should be covered by the `openArray` version, no?

lib/pure/collections/sequtils.nim
```diff
@@ -774,48 +781,86 @@ template anyIt*(s, pred: untyped): bool =
...
+      tmp
+
+proc genCallToSeqImpl(arg: NimNode): NimNode =
+  newCall(bindSym"toSeqImpl",
+    nnkForStmt.newTree(
+      ident"x", arg, newStmtList()))
+
+macro toSeqImplSingleSym(arg: typed): untyped =
+  if arg.kind != nnkSym:
+    error "identifier not unique", arg # arg is symChoice
```
alaviss, line 860: I think it's better to generate a call for it and have the compiler ambiguous call handling trigger instead, which would give a better error message.

alaviss (changes requested):
A test case that fails with this new implementation:

```nim
import std/sequtils

iterator test[T](): T =
  yield default(T)

echo toSeq(test[int])
```

Error:

```
test.nim(6, 12) Error: type mismatch: got <iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}>
but expected one of:
iterator items(a: cstring): char
  required type for a: cstring
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
iterator items(a: string): char
  required type for a: string
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
iterator items[IX, T](a: array[IX, T]): T
  required type for a: array[IX, T]
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
iterator items[T: Ordinal](s: Slice[T]): T
  required type for s: Slice[items.T]
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
iterator items[T: char](a: openArray[T]): T
  required type for a: openArray[T: char]
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
iterator items[T: enum and Ordinal](E: typedesc[T]): T
  required type for E: typedesc[T: enum and Ordinal]
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
iterator items[T: not char](a: openArray[T]): lent2 T
  required type for a: openArray[T: not char]
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
iterator items[T](a: seq[T]): lent2 T
  required type for a: seq[T]
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
iterator items[T](a: set[T]): T
  required type for a: set[T]
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
iterator items[T](xs: iterator (): T): T
  required type for xs: iterator (): T{.closure.}
  but expression 'test' is of type: iterator (): int{.inline, noSideEffect, gcsafe, locks: 0.}
  Calling convention mismatch: got '{.inline.}', but expected '{.closure.}'.

expression: items(test)
```

lib/pure/collections/sequtils.nim
krux02: well that would inject another (big?) import  into `sequtils`. Not sure if that is a good idea.

krux02:
@alaviss 
> ```nim
> test[int]
> ```

well you get the problem because `test[int]` is actually not a call to an itearator in the first place. It isn't a call at all. You should write `toSeq(test[int]())` at least.

krux02:
@alaviss 
> With that said, it seems like you are not using this new iterator in this PR. Can you split this part out to a separate PR instead?

That iterator is used in the tests. I added it to make the tests not fail. As I said. I am working around a compiler bug.

lib/pure/collections/sequtils.nim
krux02: I've added a comment. I also added elsewhere why this part can't be put into another PR. I hope this answers your questions.

lib/pure/collections/sequtils.nim
krux02: I've written extensive macros. I've written Nim to glsl compilers like in OpenGL sandbox. I've never ever used `typeKind` for anything, and would like to keep it that way.

lib/pure/collections/sequtils.nim
krux02: I know where you come from. Is there a style guide that enforces this for the compiler? I am asking, because I am also a lisp programmer, I am really really used to the style that I've been put in here. In other words, I'd rather not do it, if that is allowed.

lib/pure/collections/sequtils.nim
krux02: Sorry I don't know what you mean, or if it is possible. Again it is really long ago since I wrote it, but knowing myself, I certainly would have done it that way, if I would have seen it feasible.

lib/pure/collections/sequtils.nim
alaviss: You are going into `typed` land and the best way to be correct is to ask the compiler. It's one of the parts in typed macro to be solid-ish and trustable.

lib/pure/collections/sequtils.nim
alaviss: We haven't formally adopted a style yet, but this is not Lisp and I'd prefer that we use something more idiomatic for Nim, that is, embrace spacing and use block structures where appropriate.

lib/pure/collections/sequtils.nim
alaviss: I mean to just `genCallToSeqImpl(newCall(arg))` it, which passes the burden downstream for the compiler to handle.

lib/pure/collections/sequtils.nim
```diff
@@ -457,6 +457,15 @@ iterator filter*[T](s: openArray[T], pred: proc(x: T): bool {.closure.}): T {.ef
     if pred(s[i]):
       yield s[i]
 
+iterator filter*[T](s: seq[T], pred: proc(x: T): bool {.closure.}): T {.effectsOf: pred.} =
+  # this overload of `fileter` is written to work around this bug:
```
alaviss:
```suggestion
  # this overload of `filter` is written to work around this bug:
```

lib/pure/collections/sequtils.nim
krux02: It is not about typed vs untyped. It is about the fact that `typeKind` feels 100% redundant to `getTypeInst.kind`. It is not obvious why a separate proc with its own enum type is necessary here, when all of this information is already present in the ast.

Maybe you are right in terms of getting rid of the `lent` flag (And I will change my PR accordingly). But that doesn't change my opinion on the redundancy of this feature. After all this information is contained in the type ast, therefore it could be eleminated.

lib/pure/collections/sequtils.nim
saem: No style guide yet, but I have to admit after CPS I grew to have a significant appreciation for the approach @alaviss is suggesting. Admittedly, it was really off putting at first, I don't have a strong opinion. I just fiddle with indentation and alignment until it feels like the visual cues are drawing attention to the right things.

lib/pure/collections/sequtils.nim
krux02: if you look closely, the `seq` and the `array` implementation do slightly differ. On `seq` I don't generate an `@` because the type is already correct (I know `@` on `seq` isn't supposed to to actually have an effect as well).

lib/pure/collections/sequtils.nim
krux02: I just tested `typeKind`, it returns `ntyUnused2`. In other words, no `typeKind` can't be used here.

krux02: @alaviss regarding `iterRetTyp` I've tested it. and it works. But I'd rather not use it here. The major reason for it is, I am in macro land, I should be able to just ask the compiler "gimme the return type of this iterator", I should not need to rely of crutches like you suggested to get my type. And I do it, I ask the compiler to get the type, but the compiler also puts in garbage into my answer and that is a problem that needs to be fixed. In order for problems to be fixed, they need to be visible. That is what I am doing here, I make a problem visible. I just can't fix it within this PR.

lib/pure/collections/sequtils.nim
alaviss: The non-seq versions should all be covered by the `openArray` overload so you should be able to trim most of the fat there.

lib/pure/collections/sequtils.nim
alaviss: Can you check if comparing directly with `bindSym"lent"` works then? That'd be a lot more robust.

alaviss (approved):
Looks mostly fine to me!

Thanks for the contribution @krux02 🎉

bors r+

saem:
Yay, congrats on the first PR, @krux02.

You helped unblock me on my early compiler commits in Nim and I haven't forgotten that. Super happy that you are contributing here. 🎉❤️

krux02:
> Yay, congrats on the first PR, @krux02.
> 
> You helped unblock me on my early compiler commits in Nim and I haven't forgotten that. Super happy that you are contributing here. tadaheart

Well but I forgot that I helped you. Honestly I can't remember. But thank you for reminding me.

PR #153, opened 2022-01-13
Generic type constraints can now apply to statics in type defs

beef331:
Previously `type A[T; Y: static T] = object` was not capable of being instantiated. This PR resolves it enabling things like generic Modulo types. The changed code happens in the `tyStatic` branch as doing it later results in loss of information which means we cannot resolve the type is equal.


alaviss (approved):

saem (approved):
Bors r+

PR #154, opened 2022-01-13
Static object fields can now be accessed in generics

beef331:
Accessing static object fields  in generics such as `type DoThing[T: static Hslice[int]] = distinct array[T.a..T.b, int]` now compiles. Previously `tyStatic` was not handled during field access, this fixes that.


alaviss (approved):
:ship: it

bors r+

PR #156, opened 2022-01-16
LLDB Pretty Printing

canelhasmateus:
Inside .vscode there is lldb.py. 
It contains a nim lldb pretty printer ( in the making ) 
It is necessary because GDB doesn't play well with windows, leaving LLDB as the remaining interactive debugger, which doesn't receive much love today. 
Although it does have some ( non - oficial ) support, the instructions and printers are scattered all through the internet. 

Currentely the script contains a full lldb api type hinting for python ( the first 13k or so lines ) , followed by a badly coded pretty printer. In the upcoming days, i'll hopefully code this into a more useful state. 

Most of the current implementation comes from a random reddit thread , which should ( somehow ) be given credit **https://www.reddit.com/r/nim/comments/lhaaa6/debugging_support_formatters_for_lldb_in_vscode/**


saem:
This is cool, @canelhasmateus.

If it helps maybe take a peek at the gdb stuff in the tools directory, it's got tests and you might be able to rig them to also work with this.

We'll have to figure out CI, but one step at a time. At least when I was working tests made it much easier. It was too easy to fix one thing and break another otherwise.

There are also some hints on the gdb one in order to call routines exposed in the compiled binary under debug, which is how we could expose a bunch of reusable (gdb, lldb, and others) work.

Not requirements, just some encouragement and inspiration. 🚀

bin/nim-gdb
```diff
@@ -1,25 +0,0 @@
...
-# Find out where the pretty printer Python module is
-GDB_PYTHON_MODULE_PATH="$NIM_SYSROOT/tools/nim-gdb.py"
-
-# Run GDB with the additional arguments that load the pretty printers
-# Set the environment variable `NIM_GDB` to overwrite the call to a
-# different/specific command (defaults to `gdb`).
-NIM_GDB="${NIM_GDB:-gdb}"
-# exec replaces the new process of bash with gdb. It is always good to
-# have fewer processes.
-exec "${NIM_GDB}" -eval-command="source $GDB_PYTHON_MODULE_PATH" "$@"
```
krux02, line 25: Why is this file marked to be removed?

canelhasmateus: this was done pretty rushed-ly - i commited whatever was in my staging area, in case my pc caught fire.
 don't worry about it too much haha

krux02: Well I do worry about it. This I wrote this file and it was attacked in the past from left and right. Hand to restore its functionality multiple times. Couldn't find a way to automatically test for this file to work in CI.

canelhasmateus: Understandable. In its final form, this pr will restore your file to its former glory

PR #157, opened 2022-01-16
use improved range check of 128 for float64

### LINKED ISSUE #93: 1e100 is considered convertible to int64 and the result is invalid
### Example

This was extracted from [`tcompiletime_range_checks.nim:45`](https://github.com/nim-works/nimskull/blob/f2a9b0c18ff02266257b985e154e199c2464266f/tests/range/tcompiletime_range_checks.nim#L45), where the declaration should have produced an error but didn't.

```nim
echo int64(1e100)
```

### Current Output
```
0
```

### Expected Output

A compile-time error

### Possible Solution

There are two parts to this issue:

First is this conditional:

https://github.com/nim-works/nimskull/blob/f2a9b0c18ff02266257b985e154e199c2464266f/compiler/semexprs.nim#L183-L186

Since this value won't fit in `int64`, converting become `low(int64)`, which of course, is a valid `int64` value (though do note that in C standard this is undefined). I think either we do the comparison in `float`s, or `int128` it, which leads to this second portion:

https://github.com/nim-works/nimskull/blob/f2a9b0c18ff02266257b985e154e199c2464266f/compiler/semfold.nim#L400-L401

`toInt128(1e100)` produced `0` here, which is invalid, and also means we can't use `int128` to perform our checks.

Tagging @krux02 here, as he is the author of this code. I'm not sure what is the best to do here, though I think we should code in a range check for numbers that won't fit in `int128` (as a fallback in case the check prior fails).

### Additional Information

* Found while testing nimskull on ARM machines

```
$ nim -v
Nim Compiler Version 1.6.0 [Linux: amd64]
Copyright (c) 2006-2021 by Andreas Rumpf

Source hash: ffb7dc7223ff0eb6768add1e065e406af3397ed6
Source date: 2021-12-03

active boot switches: -d:release
```

krux02:
* Add a range check function to test if a float64 is in rage of the `int128` type.
 * Use that range check function for `semexpr.checkConvertible`
 * fixes #93

While this fixes the original issue mentioned above, it does not fix or implement the discussions in that issue. They need to be ported to their own issue.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -182,10 +182,11 @@ proc checkConvertible(c: PContext, targetTyp: PType, src: PNode): TConvStatus =
       if src.kind in nkCharLit..nkUInt64Lit and
           src.getInt notin firstOrd(c.config, targetTyp)..lastOrd(c.config, targetTyp):
         result = convNotInRange
-      elif src.kind in nkFloatLit..nkFloat64Lit and
-          (classify(src.floatVal) in {fcNan, fcNegInf, fcInf} or
-            src.floatVal.int64 notin firstOrd(c.config, targetTyp)..lastOrd(c.config, targetTyp)):
-        result = convNotInRange
+      elif src.kind in nkFloatLit..nkFloat64Lit:
+        if not src.floatVal.inInt128Range:
+          result = convNotInRange
+        elif src.floatVal.toInt128 notin firstOrd(c.config, targetTyp)..lastOrd(c.config, targetTyp):
+          result = convNotInRange
     elif targetBaseTyp.kind in tyFloat..tyFloat64:
       if src.kind in nkFloatLit..nkFloat64Lit and
           not floatRangeCheck(src.floatVal, targetTyp):
--- a/compiler/utils/int128.nim
+++ b/compiler/utils/int128.nim
@@ -3,7 +3,7 @@
 ## hold all from `low(BiggestInt)` to `high(BiggestUInt)`, This
 ## type is for that purpose.
 
-from math import trunc
+from math import trunc, pow, isNaN
 
 type
   Int128* = object
@@ -20,8 +20,8 @@ const
   Zero* = Int128(udata: [0'u32, 0, 0, 0])
   One* = Int128(udata: [1'u32, 0, 0, 0])
   Ten* = Int128(udata: [10'u32, 0, 0, 0])
-  Min = Int128(udata: [0'u32, 0, 0, 0x80000000'u32])
-  Max = Int128(udata: [high(uint32), high(uint32), high(uint32), uint32(high(int32))])
+  Min* = Int128(udata: [0'u32, 0, 0, 0x80000000'u32])
+  Max* = Int128(udata: [high(uint32), high(uint32), high(uint32), uint32(high(int32))])
   NegOne* = Int128(udata: [0xffffffff'u32, 0xffffffff'u32, 0xffffffff'u32, 0xffffffff'u32])
 
 template low*(t: typedesc[Int128]): Int128 = Min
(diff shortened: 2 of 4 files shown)


compiler/int128.nim
```diff
@@ -540,7 +540,14 @@ proc ldexp(x: float64, exp: cint): float64 {.importc: "ldexp", header: "<math.h>
...
 template bitor(a, b, c: Int128): Int128 = bitor(bitor(a, b), c)
 
+proc inInt128Range*(arg: float64): bool =
+  const lowerBound = -pow(2'f64, 127)
+  const upperBound = +pow(2'f64, 127) # 2^127-1 can't be represented with float64 precision
+  lowerBound <= arg and arg < upperBound
+
 proc toInt128*(arg: float64): Int128 =
+  # actually enabling this assert will cause /tests/misc/tconv.nim to fail
+  # assert(inInt128Range(arg), "out of range")
```
alaviss: WDYT about producing a saturating conversion as a fallback? ie. if the value is too large then construct the largest int128, and similarly the smallest int128 for values that are too small. That way it'd make sure that every callsite is protected from the undefined behavior of this conversion.

alaviss (commented):
I recall there's a test for the `int128` module itself, can you add some testcases for the new function there?

And does this PR change VM behavior at all or just fixing literal construction?

saem:
I had a giant comment and I think it totally got lost. :( Oh well, it can wait.

compiler/int128.nim
krux02: that is actually a good idea.

krux02: done

compiler/int128.nim
```diff
@@ -540,7 +540,26 @@ proc ldexp(x: float64, exp: cint): float64 {.importc: "ldexp", header: "<math.h>
...
+const lowerBoundF64 = -pow(2'f64, 127)
+# 2^127-1 can't be represented with float64 precision, therefore upper bound is
+# Exclusive
+const upperBoundF64 = +pow(2'f64, 127)
+
+proc inInt128Range*(arg: float64): bool =
+  ## Simple range check. Of coures NaN is defined not to be in range.
+  lowerBoundF64 <= arg and arg < upperBoundF64
+
 proc toInt128*(arg: float64): Int128 =
```
alaviss: I'd suggest adding some docs to describe this function behavior.

```suggestion
proc toInt128*(arg: float64): Int128 =
  ## Return an `Int128` representing the non-factional part of `arg`.
  ##
  ## Return the maximum or minimum `Int128` when `arg` cannot be represented.
```

compiler/int128.nim
```diff
@@ -540,7 +540,26 @@ proc ldexp(x: float64, exp: cint): float64 {.importc: "ldexp", header: "<math.h>
...
+  # actually enabling this assert will cause /tests/misc/tconv.nim to fail
+  # assert(inInt128Range(arg), "out of range")
+
+  # saturating at `Max` and `Min` values isn't correct, but better than returing
+  # complete garbage.
+  if arg < lowerBoundF64:
+    return Min
+  if upperBoundF64 <= arg:
+    return Max
+
```
alaviss: We should cover NaN, too (will need an import from `math`):

```suggestion
  if isNan(arg):
    return 0

```

The pick of `0` is motivated by [WASM](https://github.com/WebAssembly/spec/blob/main/proposals/nontrapping-float-to-int-conversion/Overview.md#design) and [Rust](https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast).

alaviss (commented):
A couple of tweaks and this should be good to go:

- Commit message:
  - `128` -> `int128`?
  - Can you mention the new saturating conversion behavior in the message body too?

- Please add tests for the new float behaviors in `tests/compiler/tint128.nim`.

compiler/int128.nim
krux02: I would have picked 0xDEADBEEF or something. But 0 is fine for me as well.

krux02:
checks have passed. you can take a look again for the merge.

alaviss (approved):
lgtm

bors r+

PR #161, opened 2022-01-16
remove some when compiles calls from tables/deques

krux02:
using when compiles is considered a bad practice. With this change, nimskull uses less of it.

 *   `checkIfInitialized` tableimpl is moved down to `tables` and `sharedtables`
 * `deques.defaultInitialSize` just exists. No point to check for it.

## Opinion

I think the file `tablesimpl` should be eleminated entirely. It provides little dryness of code, but really hurts code readability. Not in this PR though.


saem:
Could you amend the the commit with the excellent info of the PR body. It explains everything nicely.

Then feel free to bors.

krux02:
@saem done

saem (approved):
Bors r+

PR #166, opened 2022-01-17
std: generic diff and levenshtein algorithms

haxscramper:
- Add implementation of the *generic* myers diff and levenshtein edit
  distance algorihtms to be used in `testament` and `spellSuggest`.

Existing implementation of the diff algorithm is not generic and can be hardly made into one, due to heavily hardcoded types and unclear API, so I added new one - it might not be as optimized, but it would be much easier to reuse for writing custom diff formatters (e.g. colored)


tests/stdlib/tdiff.nim
```diff
@@ -69,5 +104,96 @@ proc main() =
...
+proc main() =
+  stringAndIntDiff()
+  levenshteinDiff()
+
 main()
-static: main()
+# FIXME when compiled to js backend `formatDiffed` fails with 'cannot
+# evaluate at compile-time' error. This is not a regression, mainline works
+# the same way.
+myersDiff()
```
saem, line 196: Is there a simpler test case we can put into the VM tests rather than in here?

haxscramper: I didn't try to minimize it yet

saem: Not the worst, at least you described it in a comment.

saem (approved):
Bors r+

PR #168, opened 2022-01-17
ci: add macOS into the test matrix

alaviss:
This introduces a shared target matrix system that will be used to
further customize each targets for nightlies in the future.

The upload-compiler action is also modified to not use `readarray` as that
feature is not available in the macOS version that is being tested.


saem (approved):
Bors r+

PR #169, opened 2022-01-17
add colored text handling library

haxscramper:
Add colored text handling library for use in message reporting, `treeRepr` pretty printing, `compiler/debugutils.render()`, testament diff, spellsuggest corrections and more. Implementation-wise it is complete (ready for review), but missing some other parts:

- [x] Text formatting tests, at least for basic SGR ANSI escape codes
- [x] Documentation for public procs
- [x] Adequate commit messages
- [x] Examples of usage

Basic design overview: `seq[Rune + Styling]` is a "colored text". `Rune` is used to allow proper formatting of the Unicode text. Using just escape codes with string would've been highly impractical due to the need of filtering them out later with `--colors=off`. Current implementation provides necessary abstraction layer that allows `cli_report` generator to always generate colored reports texts that are simply rendered as non-colored ones.

### DIFF
--- a/lib/experimental/colortext.nim
+++ b/lib/experimental/colortext.nim
@@ -1,0 +1,772 @@
+import std/[terminal, unicode, sequtils, strformat, strutils]
+
+## This module provides implementation of the basic colored text
+## abstractions. Text is represented using sequence of colored unicode
+## runes.
+##
+## The module is designed to be easily reusable for quick formatting - most
+## basic usage is `"text" + fgRed` - merge text with given formatting
+## style.
+##
+## In addition to the string construction basic modification options are
+## added - indent, align left/right/center, split by lines, add
+## indentation. This module is designed to /generate/ formatted text, so no
+## `find/replace` and other search-based procedures are added. It is
+## assumed that all necessary formatting is done beforehand.
+##
+## ..note:: If you implement string formatting functions that return
+##   `ColText` (which is a recommended approach, since regular string
+##   embedded ansi escapes does not compose properly) you need to `export`
+##   stringification operator as well - ``export $``, otherwise
+##   `echo treeRepr()` and similar operations become a little more
+##   annoying to use.
+
+export
+  # Export most commonly used styles so people won't have to import
+  # `std/terminal` all over the place.
+  terminal.ForegroundColor,
+  terminal.BackgroundColor,
+  terminal.Style
+
+type
+  TermColorBg* = distinct uint8
+    ## Terminal background color. 16-colors are represented as the first 16
+    ## values (with 0 being the default), everything higher maps to 256
+    ## colors
+
+  TermColorFg* = distinct uint8
+    ## Terminal foreground color. Mapping is identical to the background
+    ## version
+
+func `$`(fg: TermColorFg): string =
+  ## Convert foreground color to human-readble representation
+  result = $fg.uint8
+  if fg.uint8 < 15:
+    result.add "($1)" % $ForegroundColor(low(ForegroundColor).uint8 + fg.uint8)
+
+func `$`(bg: TermColorBg): string =
+  ## Convert background color to human-readable representation
+  result = $bg.uint8
+  if bg.uint8 < 15:
+    result.add "($1)" % $BackgroundColor(low(BackgroundColor).uint8 + bg.uint8)
+
+func `==`*(a, b: TermColorBg | TermColorFg): bool =
+  ## Color equality comparison
+  a.uint8 == b.uint8
+
+type
+  ColStyle* = object
+    ## Styling options
+    fg*: TermColorFg ## Foreground color
+    bg*: TermColorBg ## Background
+    style*: set[Style] ## Other styling options (italic, underline, dim,
+                       ## bright etc.)
+
+  ColRune* = object
+    ## Single unicode rune with applied coloring
+    rune*: Rune
+    style*: ColStyle
+
+  ColText* = object
+    ## Sequence of colored unicode runes
+    runes*: seq[ColRune]
+
+  ColRuneLine* = seq[ColRune]
+  ColRuneGrid* = seq[ColRuneLine]
+  ColModifier* =
+    ForegroundColor | BackgroundColor | TermColorFg | TermColorBg |
+    set[Style] | Style | ColStyle ## Typeclass, contains all types that can
+                                  ## be used for text styling
+
+
+
+
+func len*(text: ColText): int =
+  ## Number of *runes* in text
+  text.runes.len
+
+func add*(text: var ColText, rune: ColRune | seq[ColRune]) =
+  ## Append one or more colored runes to the text
+  text.runes.add rune
+
+iterator items*(text: ColText): ColRune =
+  ## Iterate over all runes left to right
+  for item in text.runes:
+    yield item
+
+iterator ritems*(text: ColText): ColRune =
+  ## Iterate over all runes right to left
+  var idx = text.runes.high
+  while 0 <= idx:
+    yield text.runes[idx]
+    dec idx
+
+func `==`*(s1, s2: ColStyle): bool =
+  ## Compare styles for equality - foreground, background and style
+  s1.fg == s2.fg and s1.bg == s2.bg and s1.style == s2.style
+
+func contains*(ps: ColStyle, s: Style): bool =
+  ## Check if coloring contains specific style option
+  ps.style.contains(s)
+
+func termColor*(bg: BackgroundColor): TermColorBg =
+  ## Convert `std/terminal.BackgroundColor` to terminal background color
+  if bg != bgDefault:
+    result = TermColorBg(bg.uint8 - low(BackgroundColor).uint8)
+
+func termColor*(bg: ForegroundColor): TermColorFg =
+  ## Convert `std/terminal.ForegroundColor` to terminal foreground color
+  if bg != fgDefault:
+    result = TermColorFg(bg.uint8 - low(ForegroundColor).uint8)
+
+func `==`*(f1: TermColorFg, f2: ForegroundColor): bool =
+  ## Compare for equality with `std/terminal` color
+  termColor(f2) == f1
+
+func `==`*(f1: TermColorBg, f2: BackgroundColor): bool =
+  ## Compare for equality with `std/terminal` color
+  termColor(f2) == f1
+
+
+func initColStyle*(
+    fg: ForegroundColor = fgDefault,
+    bg: BackgroundColor = bgDefault,
+    style: set[Style] = {}
+  ): ColStyle =
+  ## Initialize color style with given options
+  ColStyle(fg: fg.termColor, bg: bg.termColor, style: style)
+
+func colStyle*(fg: ForegroundColor): ColStyle =
+  ## Convert to style with only foreground option set
+  ColStyle(fg: fg.termColor)
+
+func colStyle*(bg: BackgroundColor): ColStyle =
+  ## Convert to style with only background option set
+  ColStyle(bg: bg.termColor)
+
+func colStyle*(fg: TermColorFg): ColStyle =
+  ## Convert to style with only foreground color set
+  ColStyle(fg: fg)
+
+func colStyle*(bg: TermColorBg): ColStyle =
+  ## Convert to style with only background color set
+  ColStyle(bg: bg)
+
+func colStyle*(style: set[Style]): ColStyle =
+  ## Convert to style with given set of styling options
+  ColStyle(style: style)
+
+func colStyle*(style: Style): ColStyle =
+  ## Convert to style with single styling option set
+  ColStyle(style: {style})
+
+func colStyle*(style: ColStyle): ColStyle =
+  ## Passthrough proc, used to simplify mapping of the `ColModifier`
+  ## typeclass to the `ColStyle` object.
+  style
+
+func default*(rune: typedesc[ColRune]): ColRune =
+  ## Default value of the colored rune
+  ColRune(style: initColStyle())
+
+func default*(style: typedesc[ColStyle]): ColStyle =
+  ## Default value of the color style
+  ColStyle(fg: fgDefault.termColor(), bg: bgDefault.termColor())
+
+func uc*(s: static[string]): Rune =
+  ## Create single unicode rune from string literal - `uc"⮰"`
+  runeAt(s, 0)
+
+func isDefault*(col: TermColorFg | TermColorBg): bool =
+  ## Check if foreground or background color have default value
+  col.uint8 == 0
+
+func `+=`*(s1: var ColStyle, s2: ColStyle) =
+  ## Merge two styling optons, overriding target if the source has
+  ## non-default value for background or foreground colors
+  s1.style = s1.style + s2.style
+  s1.fg = if s2.fg.isDefault(): s1.fg else: s2.fg
+  s1.bg = if s2.bg.isDefault(): s1.bg else: s2.bg
+
+func `+`*(s1, s2: ColStyle): ColStyle =
+  ## Merge two styles. Second one overrides colors if they are non-default.
+  result = s1
+  result += s2
+
+func `+`*(a: ColModifier, b: distinct ColModifier): ColStyle =
+  ## Merge two color modifiers and generate color style
+  ## (`fgRed+{styleUnderline}`)
+  colStyle(a) + colStyle(b)
+
+func toColRune*(rune: Rune, style: ColStyle): ColRune =
+  ## Convert colered rune to styled one with `style`
+  ColRune(rune: rune, style: style)
+
+func toColText*(text: string, style: ColStyle = default(ColStyle)): ColText =
+  ## Convert text colored one using given style
+  for rune in runes(text):
+    result.add toColRune(rune, style)
+
+func `+`*(text: string, style: ColModifier): ColText =
+  ## Convert text colored one using given style
+  toColText(text, colStyle(style))
+
+func `+`*(ch: char, style: ColModifier): ColRune =
+  ## Convert character to colored rune with given style
+  toColRune(Rune(ch), colStyle(style))
+
+func `+`*(r: Rune, style: ColModifier): ColRune =
+  ## Convert rune to colored one with given style modifier
+  toColRune(r, colStyle(style))
+
+func setStyle*(
+    text: var ColText,
+    new: ColStyle,
+    override: bool = true,
+  ) =
+  ## Set style for all runes in text, optionally overriding non-defaulted
+  ## values (or resetting them completely)
+
+  for ch in mitems(text.runes):
+    if override:
+      ch.style.fg = new.fg
+      ch.style.bg = new.bg
+      ch.style.style = new.style
+
+    else:
+      if isDefault(ch.style.fg):
+        ch.style.fg = new.fg
+
+      if isDefault(ch.style.bg):
+        ch.style.bg = new.bg
+
+      for s in new.style:
+        ch.style.style.incl s
+
+func `+`*(text: sink ColText, style: ColModifier): ColText =
+  ## Apply styling options to colored text
+  result = text
+  result.setStyle(colStyle(style), false)
+
+func toColText*(text: ColText): ColText =
+  ## Passthrough implementation of converter to colored text
+  text
+
+func toColText*(rune: ColRune): ColText =
+  ## Convert single rune to colored text
+  ColText(runes: @[rune])
+
+
+func clt*(str: string): ColText =
+  ## Shorthand to construct colored text string with default value
+  ## (`clt"default"`)
+  str + default(ColStyle)
+
+func clt*(ch: char): ColText =
+  ## Shorthand to construct colored text with single character
+  toColText(ch + default(ColStyle))
+
+func clr*(ch: char): ColRune =
+  ## Construct colored rune with default styling
+  ch + default(ColStyle)
+
+
+func isNewline*(rune: ColRune): bool =
+  ## Check if rune is a newline rune
+  rune.rune == Rune(int('\n'))
+
+func toLower*(text: sink ColText): ColText =
+  ## Convert colored text to lowercase
+  result = text
+  for rune in mitems(result.runes):
+    rune.rune = toLower(rune.rune)
+
+func toUpper*(text: sink ColText): ColText =
+  ## Convert colored text to uppercase
+  result = text
+  for rune in mitems(result.runes):
+    rune.rune = toUpper(rune.rune)
+
+
+func alignLeft*(
+    text: sink ColText, length: int, padding: ColRune = clr(' ')
+  ): ColText =
+  ## Align colored text left, using padding rune to fill in missing
+  ## elements
+
+  result = text
+  if result.len < length:
+    result.runes.add padding.repeat(length - result.len)
+
+func alignCenter*(
+    text: sink ColText, length: int, padding: ColRune = clr(' ')
+  ): ColText =
+  ## Center colored text, using padding rune to fill in missing elements
+
+  if result.len < length:
+    let
+      diff = length - text.len
+      left = diff div 2
+      right = diff - left
+
+    result.runes.add padding.repeat(left)
+    result.add text.runes
+    result.runes.add padding.repeat(right)
+
+func alignRight*(
+    text: ColText, length: int, padding: ColRune = clr(' ')
+  ): ColText =
+  ## Align colored text right, using padding rune to fill in missing
+  ## elements
+
+  if text.len < length:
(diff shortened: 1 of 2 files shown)


haxscramper:
Using https://github.com/nim-works/nimskull/pull/166 it is possible to implement many helper procedures, like this one, that can be used in lots of different reports (commands mismatches, typos, spelling correction for linter, diff implementations)

```nim
proc stringMismatchMessage*(
    input: string,
    expected: openarray[string],
    colored: bool = true,
    fixSuggestion: bool = true,
    showAll: bool = false,
  ): ColoredText =
  ## - TODO :: Better heuristics for missing/extraneous prefix/suffix

  let expected = deduplicate(expected)

  if expected.len == 0:
    return clt("No matching alternatives")


  var results = stringMismatchCandidates(input, expected).
    sortedByIt(it.distance)

  let best = results[0]

  if best.distance > int(input.len.float * 0.8):
    result = &"No close matches to {toRed(input, colored)}, possible " &
      namedItemListing(
        clt("alternative"),
        results[0 .. min(results.high, 3)].mapIt(it.target.toYellow()),
        clt("or")
      )

  else:
    result = clt(&"Did you mean to use '{toYellow(best.target, colored)}'?")

    if fixSuggestion:
      if best.edits.len < min(3, input.len div 2):
        result &= " (" & getEditVisual(
          toSeq(input),
          toSeq(best.target),
          best.edits,
          dollar[char]
        ) & ")"

      else:
        result &= clt(
          &" ({toRed(input, colored)} -> {toGreen(best.target, colored)})")

    if showAll and expected.len > 1:
      result &= " ("
      for idx, alt in results[1 ..^ 1]:
        if idx > 0:
          result &= " "

        result &= (toItalic(alt.target, colored) & "?") + tcGrey63

      result &= ")"

```

image

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,765 @@
...
+func col8*(color: uint8): TermColor8Bit = TermColor8Bit(color)
+func bg*(color: TermColor8Bit): TermColor8BitBg = TermColor8BitBg(color)
+func fg*(color: TermColor8Bit): TermColor8BitFg = TermColor8BitFg(color)
+
+func `==`*(a, b: TermColor8Bit): bool = a.uint8 == b.uint8
+
+
+type
+  ColStyle* = object
+    case use8Bit*: bool
```
krux02: so you store per rune a boolean that says if it uses 8 bit or not? Seems very inefficient if you ask me.

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,765 @@
...
+      of true:
+        fg8*: TermColor8Bit
+        bg8*: TermColor8Bit
+
+      of false:
+        fg*: ForegroundColor
+        bg*: BackgroundColor
```
krux02: 8 bit colors are usually used to save storage space, but with this condition, no storage space will be saved at all.

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,765 @@
...
+  str + (fg + bgDefault)
+
+func `+`*(rune: Colrune, style: ColStyle): ColRune =
+  result = rune
+  result.style = style
+
+func `+`*(text: ColRuneLine, style: ColStyle): ColRuneLine =
+  for ch in text:
+    result.add ch + style
+
```
krux02, line 213: so many `+` operators. Nim uses `&` for string concatenation. Why differ here?

krux02 (changes requested):
Generally speaking, I am not a fan of this. It adds several data structures and 765 lines of code just to handle colored text in a very inefficient form. Then it seems to oversee many attributes of unicode and how styling text is done in general, that I can't approve of this addition.

I've added colored to Nim as well, in [astalgo](https://github.com/nim-works/nimskull/blob/devel/compiler/astalgo.nim) all it does is it declares some constants in AnsiEscapes and adds them to the strings when needed. No additional types necessary.

From me this is a no to this PR, sorry.

saem:
> Generally speaking, I am not a fan of this. It adds several data structures and 765 lines of code just to handle colored text in a very inefficient form. Then it seems to oversee many attributes of unicode and how styling text is done in general, that I can't approve of this addition.

Design improvements, if we're running into performance issues makes sense. With that said, reading the PR description we've got colouring in many locations, so implementing that in 4 plus places I think some abstraction makes sense.

I'm trying to think about the API and if it precludes optimization. But it seems to create a template and we interpolate into it to get coloured output.

lib/experimental/colortext.nim
haxscramper: It uses 8-bit colors styling, not 8 bits per color

lib/experimental/colortext.nim
haxscramper: Because I'm not aiming to save storage space here, this is not a high-performance string handling library, it is a convenience wrapper that would make it easier to write colored output

lib/experimental/colortext.nim
haxscramper: Because it is not a string concatenation, it is a `text + style` operation. I don't specifically care what operator to use here, bit `&` wasn't used specifically because it is not a concatenation.

haxscramper:
> I've added colored to Nim as well

You mean this one, correct? https://github.com/nim-lang/Nim/commit/20a21aa1848726a60493432b7337ecbfd491f7ac

> Then it seems to oversee many attributes of Unicode and how styling text is done in general,

I don't understand what attributes of Unicode I'm missing, or what I'm missing from text styling. I do support all 16-color and 256-color schemes, as well as all regular text styling attributes.

> I've added colored to Nim as well, in astalgo all it does is it declares some constants in AnsiEscapes and adds them to the strings when needed.

That suffers from exactly the same issue as I've described in the PR title - all code is littered with logical checks, you can't remove colors after they have been added, you can't do anything else on the text after it has been colorized (alignment does not work properly, you can't pad things left/right, get length etc., only append new text)

Again, this is not a high-performance string handling library, it is a create-and-write-out solution. I don't plan to store these strings anywhere, or try to constantly rearrange them beyond what is minimally necessary to produce formatted output. I can try and squeeze a couple of bytes from the color handling, for example by using 64-bit integer and some bit hacking if you want this, but considering it is almost guaranteed that I will call `stdout.write` or `stderr.write` afterwards for each constructed string, I see little to no reason to do these micro-optimizations and the do IO that would dwarf all the runtime.

haxscramper:
@saem 

> But it seems to create a template and we interpolate into it to get coloured output.

```nim
proc treeRepr*(
    pnode: PNode,
    pathIndexed: bool     = false,
    positionIndexed: bool = true,
    maxdepth: int         = 120,
    maxlen: int           = 30,
    lineInfo: bool        = false
  ): ColoredText =

  coloredResult()

  proc aux(n: PNode, level: int, idx: seq[int]) =
    let pref = joinPrefix(level, idx, pathIndexed, positionIndexed)

    add pref
    if isNil(n):
      add "<nil>" + fgRed
      return

    if level > maxdepth:
      add " ..."
      return

    add hshow(n.kind)

    proc addComment(sep: bool = true) =
      if n.comment.len > 0:
        add "\n"
        for idx, line in enumerate(
          split(n.comment.strip(leading = false), '\n')
        ):
          if idx > 0: add "\n"
          add pref & "  # " & line + fgCyan

      elif sep:
        add " "

    proc addFlags() =
      if not isNil(n.typ):
        add " "
        add n.typ.lispRepr(colored)

      if n.flags.len > 0:
        add " nflags:" & to8Bit($n.flags, 2, 0, 3)


    if lineInfo:
      add "@"
      add $n.info.fileIndex.int + fgBlue
      add "/"
      add $n.info.line + fgCyan
      add ":"
      add $n.info.col + fgCyan
      add " "

    case n.kind:
      of nkStrKinds:
        add " "
        add "\"" & n.getStrVal() + fgYellow & "\""
        addFlags()
        addComment()

      of nkIntKinds:
        add " "
        add $n.intVal + fgBlue
        addFlags()
        addComment()

      of nkFloatKinds:
        add " "
        add $n.floatVal + fgMagenta
        addFlags()
        addComment()

      of nkIdent:
        add " "
        add n.getStrVal() + fgGreen
        addFlags()
        addComment()

      of nkSym:
        add " "
        add n.getStrVal() + fgCyan
        add " sk:"
        add ($n.sym.kind)[2 ..^ 1] + fgCyan

        if n.sym.flags.len > 0:
          add " flags:" & to8Bit($n.sym.flags, 2, 0, 3)

        if n.sym.magic != mNone:
          add " magic:" & to8Bit($n.sym.magic, 2, 0, 5)

        addFlags()
        addComment()

      of nkCommentStmt:
        addFlags()
        addComment()

      else:
        discard


    if n.kind notin nkTokenKinds:
      addFlags()
      if n.len > 0:
        add "\n"

      addComment(false)

      for newIdx, subn in n:
        aux(subn, level + 1, idx & newIdx)
        if level + 1 > maxDepth:
          break

        if newIdx > maxLen:
          break

        if newIdx < n.len - 1:
          add "\n"

  aux(pnode, 0, @[])
```

example API usage - I've decided to not make `add` varargs, because template does not work very good with these. Maybe `add` should've been a proc though

lib/experimental/colortext.nim
krux02: > > I've added colored to Nim as well
> 
> You mean this one, correct? [nim-lang/Nim@20a21aa](https://github.com/nim-lang/Nim/commit/20a21aa1848726a60493432b7337ecbfd491f7ac)

Well define what you see as "correct". It is correct in terms of ANSI color escapes, and it is very tiny. But it is certainly not correct in terms of platform agnostic. It just supports one single (standardized) type of text coloring in a minimalistic way.

> > Then it seems to oversee many attributes of Unicode and how styling text is done in general,
> 
> I don't understand what attributes of Unicode I'm missing, or what I'm missing from text styling. I do support all 16-color and 256-color schemes, as well as all regular text styling attributes.

https://www.compart.com/en/unicode/U+200F
grapheme clusters will break vertical alignment as far as I can see


> That suffers from exactly the same issue as I've described in the PR title - all code is littered with logical checks, you can't remove colors after they have been added, you can't do anything else on the text after it has been colorized (alignment does not work properly, you can't pad things left/right, get length etc., only append new text)

Well except from the logical checks, that are cetainly bad/annoying. This is designed to be minimal, yes. Only implement what is actually necessary.

Maybe you should start with a spec/requirements sheet first, so that I can judge according to the spec.
 
> Again, this is not a high-performance string handling library, it is a create-and-write-out solution. I don't plan to store these strings anywhere, or try to constantly rearrange them beyond what is minimally necessary to produce formatted output. I can try and squeeze a couple of bytes from the color handling, for example by using 64-bit integer and some bit hacking if you want this, but considering it is almost guaranteed that I will call `stdout.write` or `stderr.write` afterwards for each constructed string, I see little to no reason to do these micro-optimizations and the do IO that would dwarf all the runtime.

Actually I did measurements in the past, string handling and measurements are a performance bottleneck in the compiler.

lib/experimental/colortext.nim
krux02: Please don't use the mathematical `+` operator for this. Using operators this way, was the primary reason many programming languages abandoned operator overloading entirely.

haxscramper: What operator do you want to use then? Except for `&` which is used to denote concatenation, which is not what is going on here.

lib/experimental/colortext.nim
haxscramper: > string handling and measurements are a performance bottleneck in the compiler.

Which part of the string handling - error reporting or PSym/PIdent/nkStrLit handling? Those are two different use cases, and this library is not going to be used to represent all existing string handling operations from now on, so general measurements hardly apply here. Yes, it is slower than concatenating `\e[31m` from const, but whether this becomes a bottleneck is another question.

haxscramper: >  grapheme clusters will break vertical alignment as far as I can see

Is that common issue that comes up in handing of the compiler messages (which are mostly English text) and some basic Unicode drawing characters?

To be honest, I'm not even sure if I understand the issue at hand, I would need to read Unicode spec/documentation or something of that nature.

lib/experimental/colortext.nim
haxscramper: `+` is already used for sets, and for `style + style` operations it is exactly what happens - I merge to sets of formatting options.

Using `&` for `text + style` will prevent code like `"red" + fgRed & "blue" + fgBlue` and require `("red" & fgRed) & ("blue" & fgBlue)` for no apparent benefit

`%` might be used for `text % style` it has higher precedence, and already used for string interpolation in `std/strutils` and generally is not considered to be commutative (C uses it for mod etc.), so probably is a good solution here

lib/experimental/colortext.nim
haxscramper: > This is designed to be minimal, yes. Only implement what is actually necessary.

Yes, it is minimal and does exactly what was necessary at the time, which is to colorize a couple string elements with hardcoded styles. That's where things usually start, and if there is no demand to change anything, it is generally fine. But in this specific case, I'm trying to add a module that would make it easier to quickly progress and iterate through different error message formats and styles, without getting constantly bogged down in details. 

How about we add red underline for the error message location, like clang/rust/elm etc. do? Or implement inline type diffs like clang does on templates? Maybe some other way of error formatting - without being able to even align text blocks to each other, we won't get anywhere if we try to cling to the "minimal" approach here. 

- https://clang.llvm.org/diagnostics.html
- https://elm-lang.org/news/compiler-errors-for-humans

Try printing out highlighted code on "but expression has type" instead of garbled monochrome mess. Or do something for the abysmal quality of the type mismatch errors - try styling mismatched parts differently, or doing some smarter highlighting *inside* the failed candidate.

If we have to constantly drag around `if conf.useColors(): wantedTypeColor & typeToStr(arg.typ) & ansiCloso else: typeToStr(arg.typ)` or even `conf.maybeWrap(typeToStr(arg.typ), wantedTypeColor)` for the sake of some ephemeral performance gains in an area where easy iterative improvements are much more important, we won't get anywhere.

lib/experimental/colortext.nim
haxscramper: Actually mixing `%` and `+` suffers from the same issues - need to add extraneous `()` all over the place to avoid compilation errors, so it can be any *single* operator that is not `&` and has a higher precedence.

krux02: I am not sure yet. I don't know about any established style operators yet. Generally I would prefer a real verb instead of an operator. From my experience in language design it is as important to read something as it is to write it.

lib/experimental/colortext.nim
krux02: > Is that common issue that comes up in handing of the compiler messages (which are mostly English text) and some basic Unicode drawing characters?
> 
> To be honest, I'm not even sure if I understand the issue at hand, I would need to read Unicode spec/documentation or something of that nature.

The question is, is this for compiler internal use only, or is it supposed to be a string coloring library exposed for the public to be used? If it is compiler internal only, I think you would get almost away with ASCII only, When it is a public API, you have to dig into the Unicode spec/documentation, there is no way to avoid its complexity.

krux02:
Just as an inspiration, nos as, this is how it should be done.

In emacs a styled string is a tuple of a string, and a data structure that contains all styling for that string. So the split is at a higher level, rather than at a lower level. This that also allows an easy style removal while also being effective when there is little style to be applied. The cost isn't spent per character.

But this is also only how emacs does it for string literals, styled text within buffers is probably stored somewhat differently, otherwise insertion would break all these styling properties all the time.

There is certainly something to read about how text styling and alignment can be done.

lib/experimental/colortext.nim
haxscramper: *For now* it is most for compiler use, just like `experimental/diff`, in the future it can be expanded as necessary and moved out of experimental

haxscramper:
Yes, *if* this actually happens to be a bottleneck (which I *really* doubt), we can invest more time optimizing this. Pretty sure there are a lot of other places where we can speed up the compiler, and I'd rather prefer to make accent of making error message better rather than failing faster. I don't especially care how quickly a compiler can tell me my code is wrong, I care how much time do I need to spend fixing it up (and understanding what was wrong) - and by extensions, it should be easy for compiler authors to iterate on the messages.

krux02:
> Yes, if this actually happens to be a bottleneck (which I really doubt), we can invest more time optimizing this. Pretty sure there are a lot of other places where we can speed up the compiler, and I'd rather prefer to make accent of making error message better rather than failing faster. I don't especially care how quickly a compiler can tell me my code is wrong, I care how much time do I need to spend fixing it up (and understanding what was wrong) - and by extensions, it should be easy for compiler authors to iterate on the messages.

I've actually measured the compiler performance in the past and inefficient data structures for text generation (mostly tiny ropes for backend code) is actually affecting the performance measurably. I know this is for error messages and not for code generation, but it just happend to be two days ago that I actually went in the compiler and just print something on all the hidden reports that the compiler generates but doesn't print on screen (for whatever reason). I wanted to find out why tunittests doesn't print a deprecation warning for the usage of `macros.callsite`, which is used in unittests. And damn it was a lot of messages, I did not expect it. I didn't save them though because I focused on that deprecation warning alone, so sorry I can't give you an exact number. I can't do it right now, but maybe later I can come back with an exact number.

Anyway, I think I was annoying enough for now.

haxscramper:
> print something on all the hidden reports that the compiler generates but doesn't print on screen (for whatever reason)

Yes, with the old approach to error generation it would've been a more noticeable performance hit, but with structured reports it becomes completely irrelevant, as we only pay for messages that actually are going to be displayed and nothing else. 

The old system eagerly formatted all the reports and had randomly placed 'optimization' hacks for `compiles()`, `tryConstExpr()`, `presentFailedCandidates()` to partially mitigate this.

haxscramper:
Checklist done, ready to be reviewed and merged.

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+##
+## ..note:: If you implement string formatting functions that return
+##   `ColText` (which is a recommended approach, since regular string
+##   embedded ansi escapes does not compose properly) you need to `export`
+##   stringification operator as well - ``export $``, otherwise
+##   `echo treeRepr()` and similar operations become a little more
+##   annoying to use.
+
+export
+  # Export most commonly using styles so people won't have to import
```
saem:
```suggestion
  # Export most commonly used styles so people won't have to import
```

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+  # `std/terminal` all over the place.
+  terminal.ForegroundColor,
+  terminal.BackgroundColor,
+  terminal.Style
+
+type
+  TermColorBg* = distinct uint8
+    ## Terminal background color. 16-colors are reprsented as a first 16
```
saem:
```suggestion
    ## Terminal background color. 16-colors are represented as the first 16
```

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+
+  TermColorFg* = distinct uint8
+    ## Terminal foreground color. Mapping is identical to the background
+    ## version
+
+func `$`(fg: TermColorFg): string =
+  ## Convert foreground color to human-readble representation
+  result = $fg.uint8
+  if fg.uint8 < 15:
+    result.add "("
```
saem: For string concatenation, when we have string format included please favor interpolation over successive adds.

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+
+  ColText* = object
+    ## Sequence of colored unicode runes
+    runes*: seq[ColRune]
+
+  ColRuneLine* = seq[ColRune]
+  ColRuneGrid* = seq[ColRuneLine]
+  ColModifier* =
+    ForegroundColor | BackgroundColor | TermColorFg | TermColorBg |
+    set[Style] | Style | ColStyle ## Typeclass, contains all types that can
```
saem, line 78: For doc blocks that continue such as this please align the starting `##`

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+
+iterator ritems*(text: ColText): ColRune =
+  ## Iterate over all runes right to left
+  var idx = text.runes.high
+  while 0 <= idx:
+    yield text.runes[idx]
+    dec idx
+
+func `==`*(s1, s2: ColStyle): bool =
+  ## COmpare styles for equality - foreground, background and style
```
saem:
```suggestion
  ## Compare styles for equality - foreground, background and style
```

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+  s1.fg == s2.fg and s1.bg == s2.bg and s1.style == s2.style
+
+func contains*(ps: ColStyle, s: Style): bool =
+  ## Check if coloring contains specific style option
+  ps.style.contains(s)
+
+func term*(bg: BackgroundColor): TermColorBg =
```
saem: Unless we absolutely litter code with this let's do it termColor, at the very least.

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+func colStyle*(fg: ForegroundColor): ColStyle =
+  ## Convert to style with only foreground option set
+  ColStyle(fg: fg.term)
+
+func colStyle*(bg: BackgroundColor): ColStyle =
+  ## Convert to style with only background option set
+  ColStyle(bg: bg.term)
+
+func colStyle*(fg: TermColorFg): ColStyle =
+  ## Convert to stylewith only foreground color set
```
saem:
```suggestion
  ## Convert to style with only foreground color set
```

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+func colStyle*(bg: TermColorBg): ColStyle =
+  ## Convert to style with only background color set
+  ColStyle(bg: bg)
+
+func colStyle*(style: set[Style]): ColStyle =
+  ## Convert to style with given set of styling options
+  ColStyle(style: style)
+
+func colStyle*(style: Style): ColStyle =
+  ## Convert tot style with single styling option set
```
saem:
```suggestion
  ## Convert to style with single styling option set
```

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+
+func default*(style: typedesc[ColStyle]): ColStyle =
+  ## Default value of the color style
+  ColStyle(fg: fgDefault.term(), bg: bgDefault.term())
+
+func uc*(s: static[string]): Rune =
+  ## Create single unicode rune from string literal - `uc"⮰"`
+  runeAt(s, 0)
+
+func isDefault*(col: TermColorFg | TermColorBg): bool =
```
saem, line 180: Right about here is where you confirmed my thinking that we could cut down a lot of repetition by using type classes.

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+  result = s1
+  result += s2
+
+func `+`*(a: ColModifier, b: distinct ColModifier): ColStyle =
+  ## Merge two color modifiers and generate color style
+  ## (`fgRed+{styleUnderline}`)
+  colStyle(a) + colStyle(b)
+
+func toColRune*(rune: Rune, style: ColStyle): ColRune =
+  ## Convert coled rune to styled one with `style`
```
saem: Coles should be coloured?

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+  ## elements
+
+  result = text
+  if result.len < length:
+    result.runes.add padding.repeat(length - result.len)
+
+func alignCenter*(
+    text: sink ColText, length: int, padding: ColRune = clr(' ')
+  ): ColText =
+  ## Center colored text, using padding rune to fill in missing elemments
```
saem:
```suggestion
  ## Center colored text, using padding rune to fill in missing elements
```

lib/experimental/colortext.nim
```diff
@@ -0,0 +1,781 @@
...
+
+func add*(colored: var ColText, other: ColRuneLine) =
+  ## Add colored rune line (sequence of runes) to colored text
+  colored.runes.add other
+
+func add*(colored: var ColText, rune: ColRune) {.inline.} =
+  ## Add colored rune to the text
+  colored.runes.add rune
+
+func add*(colored: var ColText, ch: string) {.inline.} =
```
saem: I would type class this and the next func

saem (changes requested):
I did a quick skim of the docs. Will go through it again tomorrow, as in struggling to hold things in my head right now (tired).

saem (approved):

PR #171, opened 2022-01-18
niminst: add target suffix to binary archive

alaviss:
Since these archives can only be used on their respective platform, add
a suffix of the target to the archive so as to distinguish them from
generic source archives.

Right now we use the host information unconditionally, since
cross-compiling the compiler is not supported yet.


saem (approved):
Bors r+

PR #172, opened 2022-01-18
koch.py: add download bootstrap source command

alaviss:
This is used for reproducibility tests where the time might be
advanced far into the future, making download fails due to TLS
errors.

With this, the hack used in reproducibility CI can be dropped.

And as a QoL change, inform the user if they run `./koch.py` with
`--help` or `-h` before the bootstrap compiler was built as that step
might take a long time and that the user should be patient.


saem (approved):
Bors r+

PR #175, opened 2022-01-19
implement new versioning scheme for the compiler

### LINKED ISSUE #155: Roll nimskull version back to 0.0.0 or 0.1.0
This is to make sure that nightlies doesn't start with 1.6, which is the version inherited from Nim and cause trouble later on.

alaviss:
This commit implements the following scheme:

```
  <major>.<minor>.<patch>[-dev.<distance>][+dirty]
```

Where:

* `<major>.<minor>.<patch>` is the "base" version. This version is the
  compiler version being developed (ie. after 0.1.0, this will be bumped
  to 0.1.1 or 0.2.0).

* `-dev.<distance>` signifies a pre-release of the base version. Any
  version with this tag is considered to be older than the base version,
  and is not subjected to any stability requirements.

  The `<distance>` part specifies the number of commits since the last
  tag or the first commit if there are no tags.

* `+dirty` signifies that the compiler was built on modified source that
  was not committed.

With this commit, the compiler version is now rolled back to 0.1.0. The
advertised `NimVersion` remains at 1.6.0.

Implementation wise, this is done in two parts:

* The compiler maintains the base version and the `Version` structure
  with support for the suffix to be inputted by the build tool.

* Build tooling (koch specifically) maintains the addition of the
  version suffixes and ensure that the base version is correct with
  regards to the repository tags.

Known problems:

* There is no mechanism to obtain the compiler version from the code
  being compiled.

* Need a full clone with tags to compile with correct information.
  CI has been modified accordingly.

* The development status is embedded into the compiler, which makes transparent promotion impossible (ie. promoting `0.1.0-dev.100` -> `0.1.0` requires a rebuild).

Differences to https://github.com/nim-works/nimskull/discussions/150:

* `rc` is dropped.

* `dev` replaces `pre`, uses commit distance instead of commit date for
  easier comparision.

Closes https://github.com/nim-works/nimskull/issues/155

### DIFF
--- a/tools/koch/koch.nim
+++ b/tools/koch/koch.nim
@@ -29,10 +29,10 @@ import std/[json, os, strutils, parseopt, osproc]
   # If this fails with: `Error: cannot open file: std/os`, see
   # https://github.com/nim-lang/Nim/pull/14291 for explanation + how to fix.
 
+import ".."/".."/compiler/nversion
+
 import kochdocs
 import deps
-
-const VersionAsString = system.NimVersion
 
 const
   HelpText = """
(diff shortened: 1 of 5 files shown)


tools/koch/koch.nim
```diff
@@ -152,14 +168,87 @@ proc getSourceMetadata(): tuple[hash, date: string] =
...
+
+          versionSuffix = ""
+
+        # If `-` is in the tag, then the tag is of a previous release.
+        else:
+          let
+            # Git produces the following format: <tag>-<distance>-g<short commit id>
+            #
+            # We are interested in the tag and the commit distance.
+            splittedOutput = nearestReleaseTagCall.output.strip().rsplit('-')
```
saem:
```suggestion
            splitOutput = nearestReleaseTagCall.output.strip().rsplit('-')
```
Splitted isn't a word, but it also is a bit more clear that it's past tense, grammar objections aside. I can go either way.

saem (approved):
Very minor thing that I could go either way on, LGTM.

PR #178, opened 2022-01-19
minor refactor of unittest implementation

krux02:
generally speaking unittesting is very important, but the unittest
modules causes nothing but problems. The problem is, unittest pretty
much uses all problematic metaprogramming APIs at the same time. This
should be changed, step by step.

 * `template print(a,b: untyped)` -> `proc print[T](name: string, value: T)`
 * `newIdentNode(":c" & $counter)` -> `genSym`
 * `template assign` -> `newVarStmt`


saem (approved):
Already much improved. 🥳

Bors r+

PR #179, opened 2022-01-19
`cli_reporter` / `typesrenderer` refactor 

### LINKED ISSUE #167: Refactor: move all string formatting for `P(Node|Sym|Type)` to `cli_reporter`
Search through `sem*` and `types` modules in order to find leftover procs that were used for error message "formatting". If these procs are only used in `cli_reporter`, move them to this module instead. 

Example: `getProcHeader` - defined in `types.nim`, used only in `cli_reporter.nim`

Procs like these are likely to return `string` type, so it can be used as a starting hint.

tandy-1000:
Small refactor moving all leftover string formatting for P(Node|Sym|Type) from types.nim to typesrenderer.nim.
Removed some dead code, cleaned up some imports, reshuffled some consts & procs.
String formatting / rendering procs should not be in types.nim, and this should be fixed by this PR.

fixes #167

### DIFF
--- a/compiler/ast/reports.nim
+++ b/compiler/ast/reports.nim
@@ -1136,7 +1136,6 @@ type
   SemTypeMismatch* = object
     formalTypeKind*: set[TTypeKind]
     actualType*, formalType*: PType
-    descriptionStr*: string
 
   SemCallMismatch* = object
     ## Description of the single candidate mismatch. This type is later
--- a/compiler/ast/types.nim
+++ b/compiler/ast/types.nim
@@ -20,7 +20,6 @@ import
     trees,
     renderer,
     lineinfos,
-    astmsgs,
     errorhandling,
     reports
   ],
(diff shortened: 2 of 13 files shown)


saem:
Keep it up, @tandy-1000!

BTW you can make things even better by adding missing docs for exported procs or if after the move they no longer need to be exported, remove the export flag `*`.

tandy-1000:
rebased on devel.. 
edit: thanks to @haxscramper

saem:
I'm not sure types renderer should be in the AST package, seems misplaced. I'd like a follow-on refactor PR that fixes that, the heavy lifting being thinking about naming and placement.

saem:
Please double check to see if the currently exported procs etc still need exporting.

tandy-1000:
> I'm not sure types renderer should be in the AST package, seems misplaced. I'd like a follow-on refactor PR that fixes that, the heavy lifting being thinking about naming and placement.

Not sure if I'll have time to execute this, but I'll try.

> Please double check to see if the currently exported procs etc still need exporting.

All done :).

saem:
Cool, last remaining bit based on your check list is the docs

haxscramper:
had to make a minor fix for the `cli_reporter` logic, main work has already been approved, so

bors r+

PR #184, opened 2022-01-20
disable MsgOrigin at default verbosity

alaviss:
This is very noisy and doesn't help anyone who is not developing the
compiler.


haxscramper (approved):

haxscramper:
I have no objections, and this should be the only place that needs to be changed to remove `MsgOrigin` from default output, so no extra changes are required here.

saem (approved):
Bors r+

alaviss:
Messed up a setting, let's try this again

bors r+

PR #185, opened 2022-01-20
reproducible: generate HTML output for diffoscope

alaviss:
diffoscope text output is very hard to parse due to its size and lack of
coloring, so emit an HTML documentation instead and upload it.

For an example of what a failing run looks like with this, see: https://github.com/alaviss/nimskull/actions/runs/1724664069


saem (approved):
Bors r+

PR #186, opened 2022-01-20
koch: set SOURCE_DATE_EPOCH for docgen

alaviss:
Default generated docs embed the time at generation into its docs, which
is not reproducible.

Set SOURCE_DATE_EPOCH to the commit date to ensure the reproducibility
of generated docs.

Added docs building to reproducibility test. The generated documentation
will be picked up by the binary archive command.

### DIFF
--- a/tools/koch/koch.nim
+++ b/tools/koch/koch.nim
@@ -24,7 +24,7 @@ when defined(amd64) and defined(windows) and defined(vcc):
 when defined(i386) and defined(windows) and defined(vcc):
   {.link: "icons/koch-i386-windows-vcc.res".}
 
-import std/[json, os, strutils, parseopt, osproc]
+import std/[json, os, strutils, parseopt, osproc, times]
   # Using `std/os` instead of `os` to fail early if config isn't set up properly.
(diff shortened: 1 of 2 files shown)


alaviss:
bors try

alaviss:
bors try

tools/koch/koch.nim
```diff
@@ -652,6 +652,15 @@ when isMainModule:
...
     localDocsOnly = false
     localDocsOut = ""
+
+  # Set SOURCE_DATE_EPOCH to cover other tooling that might make use of the
+  # current time. Currently these tools are known to use the current time:
+  #
+  # - nim doc
+  let epoch = getSourceMetadata().date.parse("yyyy-MM-dd", zone = utc())
+                                      .toTime.toUnix()
+  echo epoch
```
saem: We want to leave the echo in?

saem:
One question, otherwise good to go.

tools/koch/koch.nim
alaviss: Oh lol I forgot about it

Issue #183, opened 2022-01-20
Move predicate logic out of `cli_reporter` report hook


haxscramper:
Move predicate logic out of direct implementation of the `cli_reporter.reportHook`, and instead add `WritabilityClassification` or something similar - `enum CanWrite, Ignore, CanForceWrite` - for regular  allowed/ignored reports, and mandatory force debugging reports that can't be filtered out.

PR #187, opened 2022-01-20
debug: clean up formatting for the vmgen debugging

haxscramper:
- Improve formatting for the vmgen code listing reports
- Split VM debuggint defines in two - `nimVMDebugExecute` to trace actually
  executed vm opcodes, and `nimVMDebugGenerate` to see generated code.
- Allow filtering out generated code by proc name, using
  `--define:expandVmListing=vmTarget`
```nim
type
  Obj = object
    charField: char
    intField: int

proc vmTarget(arg: Obj) =
  echo arg.charField.int + arg.intField

static:
  vmTarget(Obj())

```

```
nim c --filenames:canonical --define:expandVmListing=vmTarget file.nim
```

```
Code listing for vmTarget
  LdConst      r3     $     1279                system.nim(2005, 30)
  LdObj        r6     r1     r0                 file.nim(7, 11)
  NodeToReg    r5     r6     r0                 file.nim(7, 11)
  Conv         r6     r5     int   char         file.nim(7, 21)
  LdObj        r7     r1     r1                 file.nim(7, 31)
  NodeToReg    r5     r7     r0                 file.nim(7, 31)
  AddInt       r4     r6     r5                 file.nim(7, 26)
  IndCallAsgn  r2     r3     #2                 file.nim(7, 26)
  Echo         r2     r1     r0                 file.nim(7, 26)
  Ret          r0     r0     r0                 file.nim(7, 8)
  Eof          r0     r0     r0                 file.nim(7, 8)
```

### DIFF
--- a/compiler/ast/lineinfos.nim
+++ b/compiler/ast/lineinfos.nim
@@ -58,9 +58,13 @@ proc computeNotesVerbosity(): tuple[
     # debug report for transition of the configuration options
     result.base.incl {rdbgOptionsPush, rdbgOptionsPop}
 
-  when defined(nimVMDebug):
+  when defined(nimVMDebugExecute):
     result.base.incl {
-      rdbgVmExecTraceFull, # execution of the generated code listings
+      rdbgVmExecTraceFull # execution of the generated code listings
+    }
+
+  when defined(nimVMDebugGenerate):
+    result.base.incl {
       rdbgVmCodeListing    # immediately generated code listings
     }
 
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -19,7 +19,8 @@ import
     strformat,
     tables,
     intsets,
-    json
+    json,
+    strtabs
   ],
   ast/[
     lineinfos,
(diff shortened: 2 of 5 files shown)


compiler/cli_reporter.nim
```diff
@@ -3475,6 +3506,8 @@ proc reportHook*(conf: ConfigRef, r: Report): TErrorHandling =
     # Return without writing
     return
 
+
+
```
saem: New lines between conditional branches, depending upon context are often breaking my reading flow. I end up thinking the code ended, only to pick it up again.

compiler/cli_reporter.nim
```diff
@@ -3442,7 +3456,24 @@ proc reportHook*(conf: ConfigRef, r: Report): TErrorHandling =
   # comment
   assertKind r
 
-  if (
+  if (r.kind == rdbgVmCodeListing) and (
```
saem: Could we use a series of let assignments to shrink this down?

saem (approved):
Minor things, not blockers.

Nice to have is a bit in the internals doc for how to debug the VM with this using the examples from the PR message body.

compiler/cli_reporter.nim
haxscramper: This is a follow-up Todo https://github.com/nim-works/nimskull/issues/183

saem (approved):
A+, would merge again.

Bors r+

PR #188, opened 2022-01-21
separate documentation publishing from building

### LINKED ISSUE #182: Split documentation publisher out
Currently docs are published as part of the CI run, which means that we can only publish docs when CI runs on `devel`. This requires a re-run of the CI on the same commit, which is rather inefficient.

A revised scheme should be:

- Tests and artifacts are produced during main CI, which is run during push to `staging` by bors.

- Once `devel` is fast-forwarded to `staging`, publish artifacts produced during the run.

This publisher will be extended to handle binary artifacts once we start building them.

alaviss:
Since docs were published as a part of the main CI run, we were forced
to re-run CI on devel even though bors has performed the check and built
the documentation. This wastes resources especially since our test
matrix is expanding.

This commit implemented a separate publisher that publishes artifacts
built during staging (currently docs). The publisher uses a pull
architecture, that is, to query producers and obtain necessary artifacts
from them.

Main CI will now always build and upload generated documentation, which
has a nice bonus of allowing us to review docgen changes more throughly
in PRs.

An another adjustment made is that the "Edit" link will now be
hard-coded to the `devel` branch instead of being inferred from the
branch being built on. This lets us reuse docs built during the time
spent in staging branch.

Closes #182

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,9 @@ on:
   push:
     # Empty configuration means use default (ie. test all branches)
     branches-ignore:
+      # Everything would have passed bors before going into devel
+      - devel
+      # Bors temporary branches
       - staging.tmp
       - trying.tmp
       - staging-squash-merge.tmp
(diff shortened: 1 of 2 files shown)


saem (approved):
Bors r+

PR #189, opened 2022-01-21
doc: reformat comments, add readmes in subdirs

### LINKED ISSUE #180: Convert all non-documentation comments to documentation
Most of the compiler modules use single comment for documentation of the different procedures, fields, toplevel modules and so on. They won't show up in the generated documentation for modules.

```nim
type
  Ctx = object
    g: ModuleGraph
    fn: PSym
    stateVarSym: PSym # :state variable. nil if env already introduced by lambdalifting
    tmpResultSym: PSym # Used when we return, but finally has to interfere
    unrollFinallySym: PSym # Indicates that we're unrolling finally states (either exception happened or premature return)
    curExcSym: PSym # Current exception

    states: seq[PNode] # The resulting states. Every state is an nkState node.
    blockLevel: int # Temp used to transform break and continue stmts
    stateLoopLabel: PSym # Label to break on, when jumping between states.
    exitStateIdx: int # index of the last state
    tempVarId: int # unique name counter
    tempVars: PNode # Temp var decls, nkVarSection
    exceptionTable: seq[int] # For state `i` jump to state `exceptionTable[i]` if exception is raised
    hasExceptions: bool # Does closure have yield in try?
    curExcHandlingState: int # Negative for except, positive for finally
    nearestFinally: int # Index of the nearest finally block. For try/except it
                    # is their finally. For finally it is parent finally. Otherwise -1
    idgen: IdGenerator
```

Convert comments like these into documentation ones where applicable (not every single comment is actually a documentation, sometimes they are TODO etc.)

This also applies to the `discard """` comments.

```nim
discard """

one major problem:
  spawn f(a[i])
  inc i
  spawn f(a[i])
is valid, but
  spawn f(a[i])
  spawn f(a[i])
  inc i
is not! However,
  spawn f(a[i])
  if guard: inc i
  spawn f(a[i])
is not valid either! --> We need a flow dependent analysis here.

However:
  while foo:
    spawn f(a[i])
    inc i
    spawn f(a[i])

Is not valid either! --> We should really restrict 'inc' to loop endings?

The heuristic that we implement here (that has no false positives) is: Usage
of 'i' in a slice *after* we determined the stride is invalid!
"""

```

haxscramper:
Reformat comments to use double hash (to make them visible in
documentation), add readme.rst in compiler subsystem directories.

Mostly closes https://github.com/nim-works/nimskull/issues/180 although
there might be some bits and pieces that can be formatted better.

### DIFF
--- a/compiler/ast/nimlexbase.nim
+++ b/compiler/ast/nimlexbase.nim
@@ -7,10 +7,10 @@
 #    distribution, for details about the copyright.
 #
 
-# Base Object of a lexer with efficient buffer handling. In fact
-# I believe that this is the most efficient method of buffer
-# handling that exists! Only at line endings checks are necessary
-# if the buffer needs refilling.
+## Base Object of a lexer with efficient buffer handling. In fact
+## I believe that this is the most efficient method of buffer
+## handling that exists! Only at line endings checks are necessary
+## if the buffer needs refilling.
 
 import
   llstream, strutils
@@ -28,12 +28,17 @@ const
   VT* = '\x0B'
 
 const
-  EndOfFile* = '\0'           # end of file marker
-                              # A little picture makes everything clear :-)
-                              #  buf:
-                              #  "Example Text\n ha!"   bufLen = 17
-                              #   ^pos = 0     ^ sentinel = 12
-                              #
+  EndOfFile* = '\0' ##[
+end of file marker
+
+A little picture makes everything clear
+
+.. code-block:: nim
+
+  "Example Text\n ha!"   bufLen = 17
+    ^pos = 0     ^ sentinel = 12
+
+  ]##
   NewLines* = {CR, LF}
 
 type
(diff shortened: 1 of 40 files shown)


compiler/ast/nimlexbase.nim
```diff
@@ -42,12 +47,12 @@ type
...
-    stream*: PLLStream        # we read from this stream
-    lineNumber*: int          # the current line number
-                              # private data:
-    sentinel*: int
-    lineStart*: int           # index of last line start in buffer
-    offsetBase*: int          # use ``offsetBase + bufpos`` to get the offset
+    stream*: PLLStream        ## we read from this stream
+    lineNumber*: int          ## the current line number
+                              ## private data:
+    sentinel*: int#
```
ynfle:
```suggestion
    sentinel*: int
```
I assume this was a typo

haxscramper: yes, fixed now

saem (approved):
bors r+

would have preferred not rst, but that can be changed anytime really, they're so tiny. I thinking the content needs a bit of rework, some vague idea of someone walking around the code base, a bit lost, giving them advice as to which modules to look at first and/or rough package architecture -- breakout your ascii art! whatever, lofty ideas of later.

PR #190, opened 2022-01-21
refactor unittest.expect

krux02:
current implementation of expect uses following features:

  * `body` argument after ``varargs[typed]``
  * exceptions for nnkExceptionBranch branch wrapped in nnkBracket

The argument after varargs is underspecified and badly implemented. In
other words it has bugs and it is unclear if they can be fixer or the
feature needs to be deprecated and removed.

Wrapping exceptions in nnkBracket for nnkExceptionBranch is in conflict
with astspec. According to the spec tha exceptions must be added
directly to nnkExceptionBranch. It is questionable how and why it even
compiles.

You could argue that changing the names for `expect` is a breaking change. But I argue that by saying that `expect` was never inteded to be used with named arguments.

### DIFF
--- a/lib/pure/unittest.nim
+++ b/lib/pure/unittest.nim
@@ -730,7 +730,7 @@ template require*(conditions: untyped) =
     check conditions
   abortOnError = savedAbortOnError
 
-macro expect*(exceptions: varargs[typed], body: untyped): untyped =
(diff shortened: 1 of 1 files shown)


saem (approved):
bors r+

PR #191, opened 2022-01-21
debug: implement `treeRepr()` for `PNode`

haxscramper:
debug: implement `treeRepr()` for `PNode`

- ADD implementation of the `treeRepr()` proc for the `PNode`, placed in
  the new `utils/astrepr` file. New file was added to avoid mixing "ast
  algorithms" with "debug printing".
- ADD New overload for `addInDebugUtils` - `PNode -> PSym` trace. This was
  necessary to write an example on fixing a specific bug.
- FIX/ADD return support for AST printing output in the sem trace - it was
  present before structured report refactor but then got removed. Not it is
  back again.
- REMOVE old implementaiton of the `debug` calls - their functionality is
  fully replaced by the new `treeRepr()`
- REMOVE implementation of the `treeToYaml` converter - their functionality
  has been fully replaced with the `treeRepr()` which also has almost
  identical representation. They were not used anywhere in the compiler or
  tests, so it would be easy to assume they were not used for anything
  besides debugging.
- DOC add example of the updated sem tracer. Move the 'compiler debugging'
  documentation to a separate file - it is already pretty long and does not
  really fit into 'compiler architecture' that is `intern.rst`

Remaining todo items:

- [x] Finish documentation comments for the AST repr
- [x] Write adequate commit message
- [x] Add missing printing information
- [x] `treeRepr` for the PSym and PType
- [x] Declaration location for nodes
- [x] Restore missing functionality of the tree printer for the sem tracer reports - treeRepr, astRepr and others can be used there.


doc/intern.rst
```diff
@@ -292,15 +313,40 @@ It has a lot of information, but general parts for each call parts are:
...
+
+
+.. tip::
+
+    Sometimes it is necessary to print processed nodes for a specific part
+    of the compiler procedure, but it is called multiple times on the code
+    that does not show any issues, clobbering the final output and making
+    it harder to figure out what was wrong.
+
+    In that case you can wrap problemantic (input) code with
```
saem:
```suggestion
    In that case you can wrap problematic (input) code with
```

compiler/sem/semexprs.nim
```diff
@@ -3276,4 +3276,10 @@ proc semExpr(c: PContext, n: PNode, flags: TExprFlags = {}): PNode =
   of nkError: discard "ignore errors for now"
   else:
     localReport(c.config, n, reportSem rsemInvalidExpression)
-  if result != nil: incl(result.flags, nfSem)
+
+  if result != nil:
+    if c.config.isCompilerDebug():
```
saem: Personally not a fan of this over simply using and

saem (dismissed):
One typo, looks good to me.

compiler/sem/semexprs.nim
haxscramper: This is a random hack for debugging I added to get some real AST to print

compiler/ast/ast.nim
```diff
@@ -139,6 +139,7 @@ const
 
   nkSymChoices* = {nkClosedSymChoice, nkOpenSymChoice}
   nkStrKinds* = {nkStrLit..nkTripleStrLit}
+  nkIntKinds* = {nkCharLit .. nkUInt64Lit}
```
saem, line 142: Feels like this and the above line should move up to float literal and the literal nodes should be composed of those definitions.

compiler/utils/astrepr.nim
```diff
@@ -0,0 +1,442 @@
...
+    conf: ConfigRef,
+    typ: PType,
+    flags: set[TReprFlag] = defaultTreeReprFlags,
+    indent: int = 0
+  ): ColText =
+
+  coloredResult(1)
+
+  var res = addr result
+  genFields(res[], indent, flags)
```
saem, line 225: Why the ref and then deref?

compiler/utils/astrepr.nim
```diff
@@ -0,0 +1,442 @@
...
+        add n.ident.s + fgGreen
+        addFlags()
+        addComment()
+
+      of nkSym:
+        add " "
+        add n.sym.name.s + fgCyan
+        hfield "sk", substr($n.sym.kind, 2) + fgBlue
+
+        if not n.sym.owner.isNil() and n.sym.owner.kind in {
```
saem: Rather than breakup the set, I would put the second boolean expression on the next line.

compiler/utils/tree_repr_doc.rst
```diff
@@ -0,0 +1,143 @@
...
+
+Procedure arguments:
+
+- `conf`: some configuration options are inferred from it, such as
+  `--filenames:canonical`, but it is mostly used to get full paths from
+  file ids
+- `pnode`: node to be printed - can be `nil`, recursive, `nkError` or any
+  other kind form or shape.
+- `flags`: Set of the formatting options for the procedure. Several
+  built-in constants are defined in this module that migh tbe suitable for
```
saem:
```suggestion
  built-in constants are defined in this module that might be suitable for
```

compiler/front/cli_reporter.nim
```diff
@@ -47,6 +51,32 @@ import
...
+    ## unnecessary switches, or adding more magic `define()` blocks
+    semStack*: bool  ## Show `| context` entries in the call tracer
+
+    reportInTrace*: bool ## Error messages are shown with matching indentation
+    ## if report was triggered during execution of the sem trace
+
+    semTraceData*: bool ## For each sem step show processed data, or only
+    ## procedure calls.
+
+
```
saem: Too much space considering the two things are strongly related and the docs in the type inform the values in the const

haxscramper:
ready to be merged

compiler/front/cli_reporter.nim
```diff
@@ -3092,19 +3126,123 @@ proc reportBody*(conf: ConfigRef, r: DebugReport): string =
...
+        res[].add ":"
+        res[].add value
+
+      let enter = s.direction == semstepEnter
+      if controller.semTraceData and
+         s.kind != stepTrack #[ 'track' has no extra data fields ]#:
+        field("kind", $s.kind)
+        case s.kind:
+          of stepNodeToNode:
+            if enter:
```
saem, line 3184: Not a blocker, but I prefer a direction enum and an exhaustive case, as we may wish to introduce various intermediates states that are neither enter or leave.

compiler/utils/astrepr.nim
```diff
@@ -0,0 +1,442 @@
...
+  if not isNil(typ.sym):
+    result &= " sk:" & ($typ.sym.kind)[2..^1] + fgCyan
+
+  if not isNil(typ.n):
+    let t = $typ.n
+    if '\n' notin t:
+      result &= " " & t + fgRed
+
+
+# proc treeRepr*(
```
saem: Dead code

compiler/utils/tree_repr_doc.rst
```diff
@@ -0,0 +1,143 @@
...
+  different debugging or AST exploration tasks.
+- `maxDepth`: Ignore all nodes that are placed deeper than that. Useful to
+  view only high-level overvew for large nodes, such as full proc bodies
```
saem:
```suggestion
  see a high-level overview for large nodes, such as full proc bodies
```

compiler/utils/tree_repr_doc.rst
```diff
@@ -0,0 +1,143 @@
...
+- `maxLength`: on each level, print only that much subnodes. Useful to cut
```
saem:
```suggestion
- `maxLength`: on each level, print upto that many subnodes. Useful to cut
```

compiler/utils/tree_repr_doc.rst
```diff
@@ -0,0 +1,143 @@
...
+  out parts on large `case` trees, toplevel nodes for modules, statement
+  lists and such.
+
+
+.. note:: Output examples in the documentation might be slightly different
+          from the actual output due to
```
saem: Missing next part of the sentence?

compiler/utils/tree_repr_doc.rst
```diff
@@ -0,0 +1,143 @@
...
+  vmTarget(Obj())
+
+- `.0`, `.1` and `.1.0` show subnode positions of the analyzed input. This
+  is very useful when you need to index into input tree - you can
+  immediately see necessary positions.
+
+  **TODO**: Allow printing out special defines, like `ast.pragmasPos`
+  instead of the hardcoded indices
+
+- After node position somes the node symbol - `Call`, `Sym`, `ObjConstr`
```
saem: Sorry, could figure out what you meant by "somes" to offer a suggestion.

doc/debug.rst
```diff
@@ -0,0 +1,318 @@
...
+
+    {.define(nimCompilerDebug).}
+    vmTarget(Obj())
+    {.undef(nimCompilerDebug).}
+
+You will get all the call entries traced
+
+.. important::
+
+     Both *compiler* and and your file must be compiled with defines - main
```
```suggestion
     Both *compiler* and your file must be compiled with defines:
```

doc/debug.rst
```diff
@@ -0,0 +1,318 @@
...
+     compiler requires `nimDebugUtils` (this guard exists to avoid heavy
```
```suggestion
     - compiler requires `nimDebugUtils` (this guard exists to avoid heavy
```

doc/debug.rst
```diff
@@ -0,0 +1,318 @@
...
+     performance hits), and your test compiler need to be run with
```
```suggestion
     performance hits), and your file needs to be compiled with
```

doc/debug.rst
```diff
@@ -0,0 +1,318 @@
...
+   |      | |                   Location of the `addInNimDebugUtils()` - the proc itsemf
+   |      | Name of the proc
+   |      Whether proc has been entered or exited
+   Depth of the traced call tree
+
+If test compiler is executed with `-d:nimCompilerDebugTraceDir=/some/dir`
+option the reports stored between different sections are written into
+separate files in this directory. This is extremely helpful if you want to
+track bugs where you can clearly find "this works" and "this doesn't work"
+versions.
```
```suggestion
versions. This debugging technique is referred to differential debugging or sometimes differential diagnosis.
```

saem: So we build common terminology in the community.

doc/debug.rst
```diff
@@ -0,0 +1,318 @@
...
+             to sym:
+               Macro
+                 flags: {Used, Global}
+                 offset:2
+                 typ:   Proc (args):
+                 owner:  kind:skModule name:file
+
+
+As you can clearly see now, the return values of this procedure are
+different here. We have successfully localized the location of the bug from
```
```suggestion
different here. We have successfully localized the bug from
```

doc/debug.rst
```diff
@@ -0,0 +1,318 @@
...
+
+          # Call any debugging logic you can think of
+          echo c.config.treeRepr(result, maxPath = 1)
+
+
+.. warning::
+
+     Specific details of processing for modules in presence of multiple
+     compile-time statements (such as `{.define().}`) is yet to be properly
+     specified and testsed, so when used for **cross-module** semantic
```
```suggestion
     specified and tested, so when used for **cross-module** semantic
```

lib/experimental/colortext.nim
```diff
@@ -636,23 +644,26 @@ func `$`*(colr: ColRuneGrid): string =
 
     result.add toString(line)
 
+
+
 func termFg*(r, g, b: range[0 .. 5]): TermColorFg =
   ## Create 256-terminal color with given red, green and blue coloring.
   ## Colors are mapped to 216-element color cube.
-  TermColorFg(16 + b + g * 6 + (6 * 6) * r)
+  TermColorFg(offset256 + b.uint8 + g.uint8 * scaleGreen + scaleRed * r.uint8)
```
saem, line 652: Almost inclined to say converter, but not yet.

saem (changes requested):
@haxscramper did a review, sorry I'm a little off today. But I hope the feedback is useful. Let me know if you have any concerns.

This is incredible by the way, we can teach people now. It's bonkers how fast we got here.

compiler/utils/astrepr.nim
haxscramper: "result cannot be captured" in procs that `genFields()` generate

compiler/front/cli_reporter.nim
haxscramper: Then we would immediately want to format them (otherwise they would be useless), and in that case we will add `case` here.

compiler/utils/astrepr.nim
saem: Crap, know I forgot something, thanks

compiler/utils/tree_repr_doc.rst
haxscramper: `comes`, but it took me another two mintes to figure this out as well

compiler/front/cli_reporter.nim
saem: Yeah, we would switch to case. Again not a blocker just a general stylistic remark, more for in the future.

doc/debug.rst
haxscramper: no, it means that main compiler compiles tests compiler, there is no error. Previous suggestion removing `main` is also incorrect.

haxscramper: I don't know how to word this properly, but considering we are compiling the compiler to compile the file there is bound to be some weird phrasing

compiler/ast/ast.nim
haxscramper: Can be done later, there is a lot of random `const XXX = YYY` defines, I'd prefer to clean them up in one go and document them/use in definitions.

doc/debug.rst
saem: Then I need to follow through the instructions again, because I'm not getting the mechanics.

I would think we'd want to flip switches (defines) that would produce a test compiler that'll output all the things. Then when we go to use that test compiler we provide more switches (defines) that would control the output.

saem: My understanding is:
1. Create a compiler that knows to trace (compile a test compiler with defines)
2. Use the test compiler created in the prior step to compile your program with tracing parameters to control the output

doc/debug.rst
```diff
@@ -0,0 +1,319 @@
...
+
+You will get all the call entries traced
+
+.. important::
+
+     Both *compiler* and your file must be compiled with defines - *main*
+     compiler requires `nimDebugUtils` (this guard exists to avoid heavy
+     performance hits), and your *test* compiler need to be run with
+     `nimCompilerDebugCalltrace` to control what exact parts of the default
+     debugger will be executed.
```
```suggestion
     Both *compiler* and your file must be compiled with defines. The
     compiler must be built with `nimDebugUtils` (this guard avoids heavy
     performance hits). Your file must be compiled with
     `nimCompilerDebugCalltrace` to control which exact parts of the
     default debugger will be executed.
```

saem (approved):
bors r+

🎉

PR #196, opened 2022-01-23
msgs: refactor cli_reporter filtering

### LINKED ISSUE #183: Move predicate logic out of `cli_reporter` report hook

haxscramper:
- Refactor `cli_reporter.reportHook` implementation by moving the logic for
  classifying reports into different writability kinds into `options.nim`.
  This commit closes https://github.com/nim-works/nimskull/issues/183
- Moved hardcoded compile-time boolean switches into `ConfigRef` - in the
  future this type will be expanded into `--debug=<configure debugging>`
  switch and `{.debug().}` pragma. For now everything just stayed as
  hardcoded as previously, just moved type definitions around.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -51,26 +51,6 @@ import
 
 import front/options as compiler_options
 
-type
-  HackController* = object
-    ## additional configuration switches to control the behavior of the
-    ## debug printer. Since most of them are for compiler debugging, you
-    ## will most likely recompile the compiler anyway, so toggling couple
-    ## hardcoded constants here is easier than dragging out completely
-    ## unnecessary switches, or adding more magic `define()` blocks
-    semStack*: bool  ## Show `| context` entries in the call tracer
-
-    reportInTrace*: bool ## Error messages are shown with matching indentation
-    ## if report was triggered during execution of the sem trace
-
-    semTraceData*: bool ## For each sem step show processed data, or only
-    ## procedure calls.
-
-const controller = HackController(
-  semStack: off,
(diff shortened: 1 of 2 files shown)


compiler/front/cli_reporter.nim
```diff
@@ -3597,63 +3574,19 @@ proc reportHook*(conf: ConfigRef, r: Report): TErrorHandling =
...
-  ))):
-    echo conf.reportFull(r)
+  let
+    wkind = conf.writabilityKind(r)
 
-  elif r.kind == rdbgVmCodeListing or (
-    # Optionally Ignore context stacktrace
-    not controller.semStack and r.kind == rdbgTraceLine
-  ):
+  if wkind == writeDisabled:
```
saem, line 3587: Feels like if we got it right, then we should first and foremost case on the wkind, then whatever additional logic.

saem (changes requested):

compiler/front/cli_reporter.nim
haxscramper: debug reports can be both enabled and force enabled, and sem tracer first needs to be checked for the trace group rotation. So adding a case here is not really useful, since report writability kind does not necessarily dictate **how** it is written, just whether it **can/must/cannot** be written

saem (approved):
Bors r+

PR #198, opened 2022-01-23
debug: refactor debugutils tracer templates

### LINKED ISSUE #194: sem trace segfaults when executed without `--stacktrace` enabled (ex: `-d:release`, `-d:danger`).
The current implementation assumes that stacktrace exists, but this is not the case for release/danger builds, and those sometimes need to be debugged as well. This is a follow-up todo for the https://github.com/nim-works/nimskull/pull/191 - it adds another debug utils step kind, the template implementation needs to be refactored into a more manageable structure where check can be added in a single place, right now it is impossible to do cleanly.

haxscramper:
- REFACTOR implementation of the debug tracer templates from `debugutils`.
  Add generalized template that allows modifiying remaining bits that are
  needed to produce full debug step information.
- FIX debug trace segfault/out-of-bounds error for `--stacktrace=off`. This
  commit closes https://github.com/nim-works/nimskull/issues/194

### DIFF
--- a/compiler/utils/debugutils.nim
+++ b/compiler/utils/debugutils.nim
@@ -140,7 +140,89 @@ template addInNimDebugUtilsAux(conf: ConfigRef; prcname: string;
   else:
     discard # noop if undefined
 
-const instDepth = -4
+
+type
+  StepParams* = object
+    ## Parameters necessary to construct new step of the execution tracing.
+    c*: ConfigRef
+    kind*: DebugSemStepKind
+    indentLevel*: int
+    action*: string
+    info*: InstantiationInfo
+
+proc stepParams*(
+    c: ConfigRef,
+    kind: DebugSemStepKind,
+    indentLevel: int,
+    action: string
+  ): StepParams =
+
+  StepParams(
+    c: c,
+    kind: kind,
+    indentLevel: indentLevel,
+    action: action
+  )
+
+const hasStacktrace = compileOption"stacktrace"
+
+template traceStepImpl*(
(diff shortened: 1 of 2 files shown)


compiler/utils/debugutils.nim
```diff
@@ -140,36 +140,108 @@ template addInNimDebugUtilsAux(conf: ConfigRef; prcname: string;
...
+
+const instDepth = -5
+template traceEnterIt*(
+    params: StepParams,
+    body: untyped,
+    templateDepth: int = instDepth
+  ): untyped =
+  ## Convenience wrapper around the `traceStepImpl`. If called from user
+  ## code `templateDepth` parameter must be specified as well - it controls
+  ## depth of the template instantiatoin location that is need to be
```
saem:
```suggestion
  ## depth of the template instantiation location that is need to be
```

compiler/utils/debugutils.nim
```diff
@@ -140,36 +140,108 @@ template addInNimDebugUtilsAux(conf: ConfigRef; prcname: string;
...
+  ## accounted for. With current implementation of the debugutils this
+  ## value is set to `-5` (`instDepth` default), for you code it might be
```
saem:
```suggestion
  ## value is set to `-5` (`instDepth` default), for your code it might be
```

saem (changes requested):
Just a couple typos, it's change and fewer crashes

saem (approved):
LGTM 🥳

Issue #193, opened 2022-01-22
semchecked ast passed down as untyped macro argument


krux02:
### Specification

Whenever symbols and semchecked ast nodes get passed to a macro as "untyped" it is a bug. Typechecked nodes are structured very differently than untyped nodes and therefore will cause unexpected behavior in the macro.

### Example

```nim
import macros

macro check(args: varargs[untyped]): untyped =
  echo args.treeRepr
  
proc foo1() = 
  let check = 123
  var a = 1
  var b = 1
  # for some reason this is fully semchecked
  check(a == b)

# ArgList
#   Infix
#    Sym "=="
#    Sym "a"
#    Sym "b"

proc foo2() =
  let check = 123
  # this "untyped" ast contains Error nodes
  check(a == b)

# ArgList
#   Error
#     Error
#       Infix
#         Ident "=="
#         Error
#           Ident "a"
#           IntLit 365
#           StrLit "lookups.nim"
#         Ident "b"
#       IntLit 202
#       StrLit "semcall.nim"
#     IntLit 372
#     StrLit "semexprs.nim"
#     Infix
#       Ident "=="
#       Error
#         Ident "a"
#         IntLit 365
#         StrLit "lookups.nim"
#       Ident "b"

proc foo3() =
  var a = 1
  var b = 1
  # this is what it should be
  check(a == b)

# ArgList
#   Infix
#     Ident "=="
#     Ident "a"
#     Ident "b"

proc foo4() =
  # this is fine, too
  check(a == b)

# ArgList
#   Infix
#     Ident "=="
#     Ident "a"
#     Ident "b"
```

### Actual Output

```
ArgList
  Infix
    Sym "=="
    Sym "a"
    Sym "b"
ArgList
  Error
    Error
      Infix
        Ident "=="
        Error
          Ident "a"
          IntLit 365
          StrLit "lookups.nim"
        Ident "b"
      IntLit 202
      StrLit "semcall.nim"
    IntLit 372
    StrLit "semexprs.nim"
    Infix
      Ident "=="
      Error
        Ident "a"
        IntLit 365
        StrLit "lookups.nim"
      Ident "b"
ArgList
  Infix
    Ident "=="
    Ident "a"
    Ident "b"
/tmp/scratch.nim(47, 7) Hint: 'a' is declared but not used [XDeclaredButNotUsed]
/tmp/scratch.nim(48, 7) Hint: 'b' is declared but not used [XDeclaredButNotUsed]
ArgList
  Infix
[...]

PR #199, opened 2022-01-23
convert unittest.suite into a macro with structure

krux02:
* remove usage of macros.callsite()
 * add workaround for bug https://github.com/nim-works/nimskull/issues/193
 * add structural checks for unittest.suite

### DIFF
--- a/lib/pure/unittest.nim
+++ b/lib/pure/unittest.nim
@@ -460,11 +460,52 @@ proc suiteEnded() =
   for formatter in formatters:
     formatter.suiteEnded()
 
+proc testStarted(testName: string) =
+  for formatter in formatters:
+    formatter.testStarted(testName)
+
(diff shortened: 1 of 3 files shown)


lib/pure/unittest.nim
```diff
@@ -493,32 +534,59 @@ template suite*(name, body) {.dirty.} =
...
+      elif typ.eqIdent("teardown"):
+        command.expectLen 2
+        if teardown == nil:
+          teardown = command
+        else:
+          error("double definition of teardown", command)
+      else:
+        error("illegal construct for unittest suite", command)
+    else:
+      warning("please only put `test`, `setup` and `teardown` at top level of `suite`", command)
```
saem, line 571: In the future I think supporting more nesting would be great, needs some further adjustments, of course.

saem (approved):
Bors r+

PR #202, opened 2022-01-24
Issues 192

canelhasmateus:
Implemented GC stats
Preface: i've written very little nim, bear with me :)

Lots of questions, honestly

### DIFF
--- a/compiler/ast/reports.nim
+++ b/compiler/ast/reports.nim
@@ -1685,6 +1685,23 @@ type
     `out`*: string
     nimcache*: string
     hints*, warnings*: seq[tuple[name: string, enabled: bool]]
+  
+  InternalGCStats* = object
+    cycleCollections*: int    
+    cycleTableSize*: int      
+    maxCycleTableSize*: int64
+    maxPause*: int64          
+    maxPauseTime*: int64
+    maxStackCells*: int       
+    maxThreshold*: int        
+    stackCells* : int64
+    stackScans*: int          
+    occupiedMem* : int
+    totalMem*: int
+    zctCapacity*: int64
+    maxStackSize*: int        
+    stackBottom* : int64
+
 
   InternalCliData* = object
     ## Information used to construct messages for CLI reports - `--help`,
@@ -1721,7 +1738,8 @@ type
 
       of rintCliKinds:
         cliData*: InternalCliData
-
+      of rintGCStats:
+        gcStats*: InternalGCStats
       else:
         discard
 
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -2841,7 +2841,7 @@ To create a stacktrace, rerun compilation with './koch temp $1 <file>'
       assert false, "is a configuration hint, should not be reported manually"
 
     of rintGCStats:
-      result = r.msg
+      result = r.msg #Should the formatting logic be moved / duplicated here?
 
     of rintQuitCalled:
       result = "quit() called"
--- a/compiler/nim.nim
+++ b/compiler/nim.nim
@@ -85,8 +85,14 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef) =
 
   mainCommand(graph)
   if conf.hasHint(rintGCStats):
+    let gcData = GC_getData()
+    let internalGcData  = InternalGCStats( 
+        )
+
     conf.localReport(InternalReport(
-      kind: rintGCStats, msg: GC_getStatistics()))
+            kind: rintGCStats,
+            msg: GC_getStatistics() ,
+            gcStats: internalGcData) )
 
   if conf.errorCounter != 0: return
   when hasTinyCBackend:
--- a/compiler/nim.nim
+++ b/compiler/nim.nim
@@ -86,7 +86,21 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef) =
   mainCommand(graph)
   if conf.hasHint(rintGCStats):
     let gcData = GC_getData()
-    let internalGcData  = InternalGCStats( 
+    let internalGcData : InternalGCStats  = InternalGCStats(
+         cycleCollections : gcData.cycleCollections,
+         cycleTableSize : gcData.cycleTableSize,
+         maxCycleTableSize : gcData.maxCycleTableSize,
+         maxPause : gcData.maxPause,
+         maxPauseTime : gcData.maxPauseTime,
+         maxStackCells : gcData.maxStackCells,
+         maxThreshold : gcData.maxThreshold,
+         stackCells : gcData.stackCells,
+         stackScans : gcData.stackScans,
+         occupiedMem : gcData.occupiedMem,
+         totalMem : gcData.totalMem,
+         zctCapacity : gcData.zctCapacity,
+         maxStackSize : gcData.maxStackSize,
+         stackBottom : gcData.stackBottom
         )
 
     conf.localReport(InternalReport(
--- a/lib/system/gc.nim
+++ b/lib/system/gc.nim
@@ -61,24 +61,6 @@ comparisons).
 
 {.push profiler:off.}
 
-type
-  GCStatsData = object
-    cycleCollections: int    # number of performed full collections
-    cycleTableSize: int      # max entries in cycle table
-    maxCycleTableSize: int64
-    maxPause: int64          # max me
-    maxPauseTime: int64
-    maxStackCells: int       # max stack cells in ``decStack``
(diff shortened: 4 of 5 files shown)


compiler/ast/reports.nim
```diff
@@ -1721,7 +1738,8 @@ type
 
       of rintCliKinds:
         cliData*: InternalCliData
-
+      of rintGCStats:
+        gcStats*: InternalGCStats
```
canelhasmateus, line 1742: The issue description said to add a new Kind. However, there was an existing one with a fitting description. Am i correct to assume that there is no need to create a new one?

compiler/front/cli_reporter.nim
```diff
@@ -2841,7 +2841,7 @@ To create a stacktrace, rerun compilation with './koch temp $1 <file>'
       assert false, "is a configuration hint, should not be reported manually"
 
     of rintGCStats:
-      result = r.msg
+      result = r.msg #Should the formatting logic be moved / duplicated here?
```
canelhasmateus, line 2844: As described in the issue, i implemented the gc_statsdescription as getData().formatMessage(). 
However, given that message formatting is a domain of cli_reporter, i'll assume that it should be implemented here. 
Is this the case? I left it at the gc because there was a lot of calls to the getDescription.

compiler/nim.nim
```diff
@@ -85,8 +85,28 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef) =
...
+         stackCells : gcData.stackCells,
+         stackScans : gcData.stackScans,
+         occupiedMem : gcData.occupiedMem,
+         totalMem : gcData.totalMem,
+         zctCapacity : gcData.zctCapacity,
+         maxStackSize : gcData.maxStackSize,
+         stackBottom : gcData.stackBottom
+        )
+
     conf.localReport(InternalReport(
```
canelhasmateus, line 106: I had some questions here. 
First, how does GC_getStatistics() come into scope here? 
  I had a hard time figuring out how to get GC_getData() to be in scope.
  I'll assume some kind of compile / global magic, since it happened after adding the method to gc_interface

A second point: How should the data structures be handled here? Is it ok for cli_report to depend on the gc_interface GCStatsData, or is it more appropriate for it to rely only on its own data types? I basically copy a object here.

lib/system/gc.nim
```diff
@@ -900,22 +903,49 @@ when not defined(useNimRtl):
...
+             "[GC] cycle collections: " & $data.cycleCollections & "\n" &
+             "[GC] max threshold: " & $data.maxThreshold & "\n" &
+             "[GC] zct capacity: " & $data.zctCapacity & "\n" &
+             "[GC] max cycle table size: " & $data.maxCycleTableSize & "\n" &
+             "[GC] max pause time [ms]: " & $(data.maxPauseTime div 1_000_000) & "\n"
 
     # this caused memory leaks, see #10488 ; find a way without `repr`
     # maybe using a local copy of strutils.toHex or snprintf
     when defined(logGC):
-      result.add "[GC] stack bottom: " & gch.stack.bottom.repr
```
canelhasmateus, line 918: These comments left me a little bit confused.

lib/system/gc_interface.nim
```diff
@@ -36,8 +53,10 @@ when hasAlloc and not defined(js) and not usesDestructors:
     ## sweep phase may take a long time and is not needed if the application
     ## does not create cycles. Thus the mark and sweep phase can be deactivated
     ## and activated separately from the rest of the GC.
-
+  
+  proc GC_getData*(): GCStatsData {.rtl, benign.}
```
canelhasmateus, line 57: I had to add GC_getData here, to have it available in the scope of nim.nim. 
Surely, adding methods to this interface means i should also add an mock implementation for every other gc?

canelhasmateus:
@alaviss @haxscramper

lib/system/gc_interface.nim
canelhasmateus: Judging by the results of the CI tests and the sea of red "Implementation expected", i'll safely assume that is the case

lib/system/gc_interface.nim
```diff
@@ -13,6 +13,23 @@ when hasAlloc:
...
+type
+  GCStatsData* = object
+    cycleCollections*: int    # number of performed full collections
+    cycleTableSize*: int      # max entries in cycle table
+    maxCycleTableSize*: int64
+    maxPause*: int64          # max me
+    maxPauseTime*: int64
+    maxStackCells*: int       # max stack cells in ``decStack``
+    maxThreshold*: int        # max threshold that has been set
+    stackCells* : int64
```
alaviss, line 25: You should keep the `:` next to the identifier. There are many other places with this style issue in this PR as well.

```suggestion
    stackCells*: int64
```

lib/system/gc_interface.nim
```diff
@@ -13,6 +13,23 @@ when hasAlloc:
       gcOptimizeTime,    ## optimize for speed
       gcOptimizeSpace    ## optimize for memory footprint
 
+type
+  GCStatsData* = object
```
alaviss, line 17: I'm not sure if we should use an object like this. Not all supported GC even have some of the concepts in this object, let alone data for it.

I think what should be done instead is to define a set of data querying operations, like `GC_getMemoryUsage`, `GC_getCycleStats`, etc. and expose only the interfaces supported by the GC in use.

We should then be able to check if a given statistic is available via something like:

```nim
when declared(GC_getCycleStats):
  let cycleStats = GC_getCycleStats()
  <do stuff with this data>
```

An alternative is to use a lot of `when` in the `GCStats` object, for example:

```nim
type GCStats = object
  # Common data goes here
  totalMem*: int ## Total amount of memory allocated from the OS
  occupiedMem*: int ## The amount of memory in use
  when defined(gcRefc) or defined(gcMarkAndSweep) or defined(gcOrc):
    cycleCollections*: int    ## number of performed full collections (orc, refc, markAndSweep only)

  when defined(gcRefc) or defined(gcMarkAndSweep):
    cycleTableSize*: int      ## max entries in cycle table (refc, markAndSweep only)
    maxCycleTableSize*: int64 ## Only available on refc, markAndSweep GCs.
    maxPause*: int64          ## Only available on refc, markAndSweep GCs.
    maxPauseTime*: int64      ## Only available on refc, markAndSweep GCs.
    maxStackCells*: int       ## Only available on refc, markAndSweep GCs.
```

Why `when`s? It's because you cannot change the GC on-the-fly, so it makes little sense to have all the data for cycle collections when you use arc, for example.

compiler/ast/reports.nim
```diff
@@ -1685,6 +1685,23 @@ type
     `out`*: string
     nimcache*: string
     hints*, warnings*: seq[tuple[name: string, enabled: bool]]
+  
+  InternalGCStats* = object
```
alaviss, line 1689: My comment about `GCStatsData` also applies here.

lib/system/gc_interface.nim
haxscramper: You can't choose gc on the fly, but you certainly can try to analyze the data from two program versions built with different compiler options, and for that (if we dumped data in a structured manner) we need all fields to be accessible via variant object that switches on used gc type.

lib/system/gc.nim
saem: Did the referenced issue shed any light at all?

compiler/front/cli_reporter.nim
saem: Yeah, that sounds right.

Question is why is get description get called so much?

compiler/nim.nim
saem: Likely in system some how and comes in behind some conditionals.

First identify the module is in. Then find the import or include of that module, I bet it chases back to system.

compiler/ast/reports.nim
saem: Yeah, probably not.

lib/system/gc_interface.nim
saem: Modules expose an interface (all exported items), so if we're doing some sort of static interface through conditional compilation then yes that interface should be informed implemented.

The feedback on this instance showing up as a sea of red.

saem:
You're figuring it out, keep going, @canelhasmateus

saem:
@canelhasmateus should this go back into drafts while you keep plugging away?

canelhasmateus:
@saem definitely. I was under the impression i already had.

saem:
This is now stake, please reopen when you start working on it again.

PR #206, opened 2022-01-26
Testament diff rework - unstructured, structured

haxscramper:
wip, todo:

- [x] adequate commit message. I could not bear the sight of some testament parts, specifically `addResult`, so I split it up at least a little, and there were a couple of other entries that I had to reformat because I could not meaningfully work with them any other way, because they were **SO** badly implemented.
- [x] squash edits
- [x] `--msgFormat` documentation
- [x] examples of the structured messages for the compiler output, add documentation to `nimc.rst`
- [x] Remove temporary attempts for message formatting - I'm sure that `toHtml` for colored text is not needed
- [x] Add better documentation for `#[tt.` inline annotations testament
- [x] More explanation comments about formatting logic - those are almost always hard to reason about, random string bits placed all over the code don't help to understand anything either
- [x] make unified diff join adjacent edit lines as well, the `-+-+` is less readable compared to `--++`. Make this configurable, sometimes `-+-+` with inline change annotations is more useful than `--++` that was joined.
- [x] more tests for structural comparison - no difference, has difference, more given output with a single expected one, equally weighted options
- [x] remove debug prints from tests, they fail megatest hack
- [x] nimsuggest fails because of missing sexp import - it was moved, this needs to be documented
- [x] add more descriptions for testament spec fields, nobody is going to do this "look in the code" bullshit if there are no comments there even. And the link is constantly broken anyway, and it doesn't even link to the type definition, instead links to the source code of the parse for whatever reason.
- [?] if annotation is inline, actually show the surrounding code and place arrows around maybe, because it is not clear how location was inferred from the position of the ^. Possibly this is an overkill, but again, at least trying this out is necessary.
- [x] Examples of test files that use new structured reporting
- [x] grouped and ungrouped unified diffs

Changes for regular diff:

Convert this:

image

to this

image

and also make testament support structured reports that would allow you to get this

image

Out of this

```nim
discard """
nimoutFormat: sexp
cmd: "nim c --msgFormat=sexp --skipUserCfg --hints=on --hint=all:off --hint=User:on --filenames:canonical $file"
nimout: '''
(User :str "User Hint" :location ("tfile.nim" 8 _))
'''
"""

{.hint: "User hint".}

{.hint: "Another hint".} #[tt.Hint
      ^ (User :str "Another hint") ]#
```

Note how structured diff properly handles both inline and `nimout` error messages whereas with unstructured diff testament can only report half of the problems.

### DIFF
--- a/compiler/front/sexp_reporter.nim
+++ b/compiler/front/sexp_reporter.nim
@@ -1,0 +1,182 @@
+## Implementation of the structured CLI message generator. Using
+## `--msgFormat=sexp` will make compiler switch to the report hook
+## implemented in this module.
+
+import
+  experimental/[
+    sexp,
+    diff,
+    colortext,
+    sexp_diff
+  ],
+  ast/[
+    lineinfos,
+    ast,
+    reports
+  ],
+  front/[
+    options,
+    msgs
+  ],
+  std/[
+    strutils
+  ]
+
+import std/options as std_options
+
+var writeConf: ConfigRef
+
+
+proc addFields[T](s: var SexpNode, r: T, ignore: seq[string] = @[])
+
+
+
+proc sexpItems*[T](s: T): SexpNode =
+  result = newSList()
+  for item in items(s):
+    result.add sexp(item)
+
+proc sexp*[T: object | tuple](obj: T): SexpNode =
+  result = newSList()
+  addFields(result, obj)
+
+proc sexp*[T: object | tuple](obj: ref T): SexpNode = sexp(obj[])
+proc sexp*[E: enum](e: E): SexpNode = newSSymbol($e)
+proc sexp*[T](s: seq[T]): SexpNode = sexpItems(s)
+proc sexp*[R, T](s: array[R, T]): SexpNode = sexpItems(s)
+proc sexp*[I](s: set[I]): SexpNode = sexpItems(s)
+proc sexp*(s: cstring): SexpNode = sexp($s)
+
+proc sexp*(v: SomeInteger): SexpNode = newSInt(BiggestInt(v))
+proc sexp*(id: FileIndex): SexpNode =
+  sexp(writeConf.toMsgFilename(id))
+
+
+iterator sexpFields[T: object | tuple](obj: T, ignore: seq[string] = @[]): SexpNode =
+  for name, value in fieldPairs(obj):
+    var pass = true
+    when value is ref or value is ptr:
+      if isNil(value):
+        pass = false
+
+    when value is seq or value is string:
+      if len(value) == 0:
+        pass = false
+
+    when value is TLineInfo:
+      if pass and value == unknownLineInfo:
+        pass = false
+
+    when value is ReportLineInfo:
+      if pass and not value.isValid():
+        pass = false
+
+    if pass and name in ignore:
+      pass = false
+
+    if pass:
+      yield newSKeyword(name, sexp(value))
+
+
+proc add*(self: var SexpNode, str: string, expr: SexpNode) =
+  self.add newSSymbol(":" & str)
+  self.add expr
+
+proc sexp*[T](o: Option[T]): SexpNode =
+  if o.isNone: newSNil() else: sexp(o.get())
+
+proc addFields[T](s: var SexpNode, r: T, ignore: seq[string] = @[]) =
+  for item in sexpFields(r, ignore):
+    s.add item
+
+proc sexp*(i: ReportLineInfo): SexpNode =
+  convertSexp([
+    writeConf.formatPath(i.file).sexp(),
+    sexp(i.line),
+    sexp(i.col)
+  ])
+
+proc sexp*(i: TLineInfo): SexpNode =
+  convertSexp([sexp(i.fileIndex), sexp(i.line), sexp(i.col)])
+
+proc sexp*(e: StackTraceEntry): SexpNode =
+  result = newSList()
+  result.addFields(e, @["filename"])
+  result.add newSKeyword(
+    "filename", writeConf.formatPath($e.filename).sexp())
+
+
+proc sexp*(typ: PType): SexpNode =
+  if typ.isNil: return newSNil()
+  result = newSList()
+  result.add newSSymbol(($typ.kind)[2 ..^ 1])
+  if typ.sons.len > 0:
+    result.add("sons", sexp(typ.sons))
+
+proc sexp*(node: PNode): SexpNode =
+  if node.isNil: return newSNil()
+
+  result = newSList()
+  result.add newSSymbol(($node.kind)[2 ..^ 1])
+  case node.kind:
+    of nkCharLit..nkUInt64Lit:    result.add sexp(node.intVal)
+    of nkFloatLit..nkFloat128Lit: result.add sexp(node.floatVal)
+    of nkStrLit..nkTripleStrLit:  result.add sexp(node.strVal)
+    of nkSym:                     result.add newSSymbol(node.sym.name.s)
+    of nkIdent:                   result.add newSSymbol(node.ident.s)
+    else:
+      for node in node.sons:
+        result.add sexp(node)
+
+proc sexp*(t: PSym): SexpNode =
+  convertSexp([
+    substr($t.kind, 2).newSSymbol(),
+    name = sexp(t.name.s),
+    info = sexp(t.info)
+  ])
+
+
+proc reportHook*(conf: ConfigRef, r: Report): TErrorHandling =
+  writeConf = conf
+  let wkind = conf.writabilityKind(r)
+
+  if wkind == writeDisabled:
+    return
+
+  else:
+    var s = newSList()
+    s.add newSSymbol(multiReplace($r.kind, {
+      "rsem": "Sem",
+      "rpar": "Par",
+      "rlex": "Lex",
+      "rint": "Int",
+      "rext": "Ext",
+      "rdbg": "Dbg",
+      "rback": "Bck",
+    }))
+    s.add newSSymbol(":severity")
+    s.add sexp(conf.severity(r))
+
+    let f = @["kind"]
+
+    case r.category:
+      of repLexer:    s.addFields(r.lexReport, f)
+      of repParser:   s.addFields(r.parserReport, f)
+      of repCmd:      s.addFields(r.cmdReport, f)
+      of repSem:
+        if r.kind == rsemProcessingStmt:
+          s.addFields(r.semReport, f & "node")
+
+        else:
+          s.addFields(r.semReport, f)
+
+      of repDebug:    s.addFields(r.debugReport, f)
+      of repInternal: s.addFields(r.internalReport, f)
+      of repBackend:  s.addFields(r.backendReport, f)
+      of repExternal: s.addFields(r.externalReport, f)
+
+    if wkind == writeForceEnabled:
+      echo s.toLine().toString(conf.useColor)
+
+    else:
+      conf.writeln(s.toLine().toString(conf.useColor))
--- a/doc/nimc.rst
+++ b/doc/nimc.rst
@@ -141,6 +141,20 @@ Level  Description
        for compiler developers.
 =====  ============================================
 
+Compiler message formats
+------------------------
+
+The compiler can output messages in both unstructured (plaintext) and
+structured (S-expressions) forms. S-expressions were chosen mostly for
+integration with testament, in the future json support will be added as
+well.
+
+You can select message format using `--msgFormat=text|sexp`:option: switch
+in the compiler. Unstructured compiler reports are formatted for higher
+readability and used by default. Structured reports are formatted as
+S-expressions, one per line. Every single compiler report is wrapped in
+structured data, including ``echo`` messages at compile-time.
+
 
 Compile-time symbols
 --------------------
--- a/doc/testament.rst
+++ b/doc/testament.rst
@@ -91,87 +91,46 @@ you have to run at least 1 test *before* generating a report:
 Writing Tests
 =============
 
-Example "template" **to edit** and write a Testament unit:
 
-.. code-block:: nim
+``description`` - textual description of the test. **Highly** recomended to
+add one - in the future testament might use this to provide better teardown
+reports, or notify about ``knownIssue`` state transitions.
 
-  discard """
+Test execution options
+----------------------
 
-    # What actions to expect completion on.
-    # Options:
-    #   "compile": expect successful compilation
-    #   "run": expect successful compilation and execution
-    #   "reject": expect failed compilation. The "reject" action can catch
-    #             {.error.} pragmas but not {.fatal.} pragmas because
-    #             {.fatal.} pragmas guarantee that compilation will be aborted.
-    action: "run"
+- ``action`` - What action(s) to expect completion on.
+  - ``"compile"``: expect successful compilation
+  - ``"run"``: expect successful compilation and execution
+  - ``"reject"``: expect failed compilation. The "reject" action can catch
+    `{.error.}` pragmas but not `{.fatal.}` pragmas because `{.fatal.}` pragmas
+    guarantee that compilation will be aborted.
 
-    # The exit code that the test is expected to return. Typically, the default
-    # value of 0 is fine. Note that if the test will be run by valgrind, then
-    # the test will exit with either a code of 0 on success or 1 on failure.
-    exitcode: 0
+- ``batchable``: Can be run in batch mode, or not.
 
-    # Provide an `output` string to assert that the test prints to standard out
-    # exactly the expected string. Provide an `outputsub` string to assert that
-    # the string given here is a substring of the standard out output of the
-    # test.
-    output: ""
-    outputsub: ""
+- ``joinable``: Can be run Joined with other tests to run all togheter, or
+  not. Defaults to `true`
 
-    # Whether to sort the output lines before comparing them to the desired
-    # output.
-    sortoutput: true
+- ``timeout`` Timeout seconds to run the test. Fractional values are supported.
 
-    # Each line in the string given here appears in the same order in the
-    # compiler output, but there may be more lines that appear before, after, or
-    # in between them.
-    nimout: '''
-  a very long,
-  multi-line
-  string'''
+- ``cmd``: Command used to run the test. If left out or an empty
+  string is provided, the command is taken to be: ``"nim $target --hints:on
+  -d:testing --nimblePath:build/deps/pkgs $options $file"`` You can use the
+  ``$target``, ``$options``, and ``$file`` placeholders in your own
+  command, too.
 
-    # This is the Standard Input the test should take, if any.
-    input: ""
+  example: ``"nim c -r $file"``
 
-    # Error message the test should print, if any.
-    errormsg: ""
+- ``targets`` supported backend compilation targets for test into (c,
+  cpp, objc, js).
 
-    # Can be run in batch mode, or not.
-    batchable: true
+- ``matrix`` flags with which to run the test, delimited by `;`
 
-    # Can be run Joined with other tests to run all togheter, or not.
-    joinable: true
+- ``disabled`` Conditions that will skip this test. Use of multiple
+  "disabled" clauses is permitted.
 
-    # On Linux 64-bit machines, whether to use Valgrind to check for bad memory
-    # accesses or memory leaks. On other architectures, the test will be run
-    # as-is, without Valgrind.
-    # Options:
-    #   true: run the test with Valgrind
-    #   false: run the without Valgrind
-    #   "leaks": run the test with Valgrind, but do not check for memory leaks
-    valgrind: false   # Can use Valgrind to check for memory leaks, or not (Linux 64Bit only).
+  .. code-block:: nim
 
-    # Command the test should use to run. If left out or an empty string is
-    # provided, the command is taken to be:
-    # "nim $target --hints:on -d:testing --nimblePath:build/deps/pkgs $options $file"
-    # You can use the $target, $options, and $file placeholders in your own
-    # command, too.
-    cmd: "nim c -r $file"
-
-    # Maximum generated temporary intermediate code file size for the test.
-    maxcodesize: 666
-
-    # Timeout seconds to run the test. Fractional values are supported.
-    timeout: 1.5
-
-    # Targets to run the test into (c, cpp, objc, js).
-    targets: "c js"
-
-    # flags with which to run the test, delimited by `;`
-    matrix: "; -d:release; -d:caseFoo -d:release"
-
-    # Conditions that will skip this test. Use of multiple "disabled" clauses
-    # is permitted.
     disabled: "bsd"   # Can disable OSes...
     disabled: "win"
     disabled: "32bit" # ...or architectures
@@ -179,10 +138,74 @@ Example "template" **to edit** and write a Testament unit:
     disabled: "azure" # ...or pipeline runners
     disabled: true    # ...or can disable the test entirely
 
-  """
-  assert true
-  assert 42 == 42, "Assert error message"
+- ``knownIssue``
 
+Compiler output assertions
+--------------------------
+
+- ``errormsg``: Error message the test should print, if any.
+
+
+- ``nimout`` Each line in the string given here appears in the same order
+    in the compiler output, but there may be more lines that appear before,
+    after, or in between them. Note that specifying multiline strings for
+    testament spec inside of the `discard """` section requires using
+    triple single quotes `'`
+
+    .. code-block:: nim
+
+        nimout: '''
+        a very long,
+        multi-line
+        string'''
+
+- ``nimoutFull``: true/false, controls whether full compiler output must be
+  asserted, or only presence of error messages
+
+- ``maxcodesize``: Max side of the resulting codegen file for a test
+
+In addition to ``nimout`` message annotations testament also allows to
+supply hints, warnings and error messages directly in the source code using
+specially formatted comments, starting with ``#[tt.``. For example, if you
+want to assert that error message is genrated, you can write a following
+test:
+
+.. code-block::
+
+   {.error: "Error message".} #[tt.Error
+     ^ "Error message"
+   ]#
+
+File, line and column information are automatically inferred from the
+position of the ``^`` marker in the annotation body.
+
+
+Binary output assertions
+------------------------
+
+- ``exitcode``: The exit code that the test is expected to return.
+  Typically, the default value of 0 is fine. Note that if the test will be
+  run by valgrind, then the test will exit with either a code of 0 on
+  success or 1 on failure.
+
+- ``output``, ``outsub``: Provide an `output` string to assert that the
+  test prints to standard out exactly the expected string. Provide an
+  `outputsub` string to assert that the string given here is a substring of
+  the standard out output of the test.
+
+- ``sortoutput`` Whether to sort the output lines before comparing them to
+  the desired output.
+
+- ``input``: this is the Standard Input the test should take, if any.
+
+
+- ``valgrind`` On Linux 64-bit machines, whether to use Valgrind to check
+    for bad memory accesses or memory leaks. On other architectures, the
+    test will be run as-is, without Valgrind.
+
+    - ``true``: run the test with Valgrind
+    - ``false``: run the without Valgrind
+    - ``"leaks"``: run the test with Valgrind, but do not check for memory leaks
 
 * As you can see the "Spec" is just a `discard """ """`.
 * Spec has sane defaults, so you don't need to provide them all, any simple assert will work just fine.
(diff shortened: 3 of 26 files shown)


haxscramper:
bors try

haxscramper:
image

Sweep-grouping inline diffs is certainly necessary

that's what was generated before

image

saem:
I tried to review, little too tired to digest this tonight, will try again tomorrow.

compiler/front/sexp_reporter.nim
```diff
@@ -0,0 +1,185 @@
...
+    msgs
+  ],
+  std/[
+    strutils
+  ]
+
+
+var writeConf: ConfigRef
+
+import std/options as std_options
```
saem: Not a blocker or hard requirement, more a stylistic note: I really prefer imports at the top. Much easier to spot a module with far too many dependencies.

compiler/front/sexp_reporter.nim
```diff
@@ -0,0 +1,185 @@
...
+
+proc addFields[T](s: var SexpNode, r: T, ignore: seq[string] = @[])
+
+proc sexp[T: object | tuple](obj: T): SexpNode =
+  result = newSList()
+  addFields(result, obj)
+
+proc sexp[T: object | tuple](obj: ref T): SexpNode =
+  result = newSList()
+  addFields(result, obj[])
```
saem: After the deref can we not simply call the proc above?

compiler/front/sexp_reporter.nim
```diff
@@ -0,0 +1,185 @@
...
+
+proc sexp*[E: enum](e: E): SexpNode = newSSymbol($e)
```
saem: It's a run of procs like these where I wish we had more of the Haskell style declarations, oh well, inspiration for down the road.

compiler/front/sexp_reporter.nim
```diff
@@ -0,0 +1,185 @@
...
+
+proc sexpItems*[T](s: T): SexpNode =
```
saem, line 34: Can we move this below the other s-exp procs, it flows better and if the T is one of the ones below it'll not be confusing if some declares a new one but this doesn't take it into account up here.

Also, seriously this probably shouldn't compile.

compiler/front/sexp_reporter.nim
```diff
@@ -0,0 +1,185 @@
...
+proc sexp*[R, T](s: array[R, T]): SexpNode = sexpItems(s)
+proc sexp*[I](s: set[I]): SexpNode = sexpItems(s)
+proc sexp*(s: cstring): SexpNode = sexp($s)
+
+proc sexp*(v: SomeInteger): SexpNode = newSInt(BiggestInt(v))
+proc sexp*(id: FileIndex): SexpNode =
+  sexp(writeConf.toMsgFilename(id))
+
+
+iterator sexpFields[T](obj: T, ignore: seq[string] = @[]): SexpNode =
```
saem: So weird that this works without constraints, feels like a bug.

I think T should have an a type constraint of `object | tuple`

compiler/front/sexp_reporter.nim
```diff
@@ -0,0 +1,185 @@
...
+        pass = false
+
+    if pass and name in ignore:
+      pass = false
+
+    if pass:
+      yield newSKeyword(name, sexp(value))
+
+
+proc add*(other: var SexpNode, str: string, expr: SexpNode) =
```
saem:
```suggestion
proc add*(self: var SexpNode, str: string, expr: SexpNode) =
```

compiler/front/sexp_reporter.nim
```diff
@@ -0,0 +1,185 @@
...
+  other.add newSSymbol(":" & str)
+  other.add expr
+
+proc sexp*[T](o: Option[T]): SexpNode =
+  if o.isNone: newSNil() else: sexp(o.get())
```
saem, line 86:
```suggestion
  sexp(o.getOr(nil))
```

Does this work?

saem (changes requested):
Reviewed code, will do docs tomorrow 🤞

compiler/front/sexp_reporter.nim
haxscramper: > Also, seriously this process shouldn't compile.

what 'process' are you talking about and why it shouldn't compile?

compiler/front/sexp_reporter.nim
haxscramper: Why does it feel like a bug? Generic iterator, accepts generic type `T`, works as intended

> I think T should have an a type constraint of object | tuple

Can be clarified later, but the fact it works without it doesn't really make it a bug IMO

haxscramper:
I will update the PR for review comments after you review the docs. Also, please do read the tests as well - I added some explanations about how diff is intended to function, and it's behavior there.

compiler/front/sexp_reporter.nim
saem: Typo, yeah it's strange that it's allowed by the compiler.

compiler/front/sexp_reporter.nim
saem: It absolutely is a bug because there is no guarantee you can iterate over T?

doc/nimc.rst
```diff
@@ -141,6 +141,20 @@ Level  Description
        for compiler developers.
 =====  ============================================
 
+Compiler message formats
+------------------------
+
+Compiler can output messages in both unstructured (plaintext) and
```
saem:
```suggestion
The compiler can output messages in both unstructured (plaintext) and
```

doc/testament.rst
```diff
@@ -91,98 +91,120 @@ you have to run at least 1 test *before* generating a report:
 Writing Tests
 =============
 
-Example "template" **to edit** and write a Testament unit:
 
-.. code-block:: nim
+``description`` - textual description of the test. **Highly** recomended to
```
saem, line 95:
```suggestion
``description`` - textual description of the test. **Highly** recommended -
```

doc/testament.rst
```diff
@@ -91,98 +91,120 @@ you have to run at least 1 test *before* generating a report:
...
+add one - in the future testament might use this to provide better teardown
```
saem, line 96:
```suggestion
in the future testament might use this to provide better teardown
```

doc/testament.rst
```diff
@@ -91,98 +91,120 @@ you have to run at least 1 test *before* generating a report:
...
-    sortoutput: true
-
-    # Each line in the string given here appears in the same order in the
-    # compiler output, but there may be more lines that appear before, after, or
-    # in between them.
-    nimout: '''
-  a very long,
-  multi-line
-  string'''
+- ``action`` - What actions to expect completion on.
```
saem:
```suggestion
- ``action`` - what action(s) to execute and expect.
```

doc/testament.rst
```diff
@@ -91,98 +91,120 @@ you have to run at least 1 test *before* generating a report:
...
+- ``joinable``: Can be run Joined with other tests to run all togheter, or
+  not. Defaults to `true`
 
-    # Can be run in batch mode, or not.
-    batchable: true
+- ``timeout`` Timeout seconds to run the test. Fractional values are supported.
 
-    # Can be run Joined with other tests to run all togheter, or not.
-    joinable: true
+- ``cmd``: Command the test should use to run. If left out or an empty
```
saem:
```suggestion
- ``cmd``: Command used to run the test. If left out or an empty
```

doc/testament.rst
```diff
@@ -91,98 +91,120 @@ you have to run at least 1 test *before* generating a report:
...
-    valgrind: false   # Can use Valgrind to check for memory leaks, or not (Linux 64Bit only).
+  example: ``"nim c -r $file"``
 
-    # Command the test should use to run. If left out or an empty string is
-    # provided, the command is taken to be:
-    # "nim $target --hints:on -d:testing --nimblePath:build/deps/pkgs $options $file"
-    # You can use the $target, $options, and $file placeholders in your own
-    # command, too.
-    cmd: "nim c -r $file"
+- ``targets`` Targets to run the test into (c, cpp, objc, js).
```
saem:
```suggestion
- ``targets`` supported backend compilation targets for the test (c, cpp, objc, js).
```

doc/testament.rst
```diff
@@ -191,10 +213,129 @@ Example "template" **to edit** and write a Testament unit:
 * Has some built-in CI compatibility, like Azure Pipelines, etc.
 * `Testament supports inlined error messages on tests, basically comments with the expected error directly on the code. <https://github.com/nim-works/nimskull/blob/9a110047cbe2826b1d4afe63e3a1f5a08422b73f/tests/effects/teffects1.nim>`_
 
+Reading test outputs
+====================
+
+Testament supports two different modes of interaction with the compiler -
+structured and unstructured. Unstructured interaction mode (currently
+default) allows user to specify exact compiler output that should be
+produced by input file and then compares it based on the ``nimoutFull``
```
saem:
```suggestion
produced by the test and then compares it based on ``nimoutFull``
```

doc/testament.rst
```diff
@@ -191,10 +213,129 @@ Example "template" **to edit** and write a Testament unit:
...
+configuration options.
+
+If tests are mismatched, failure message is generated, showing failed
```
saem:
```suggestion
If there is a mismatch, a failure message is generated, showing the
```

doc/testament.rst
```diff
@@ -191,10 +213,129 @@ Example "template" **to edit** and write a Testament unit:
...
+.. note:: expected (on the left) outout has two lines deleted - trailing
+          ``'''`` in the testament spec is placed on the next line, so it
+          is considered to be a string literal of ``"Expected unstructured
+          compiler output\n"``
+
+
+Structured mismatches
+---------------------
+
+In structured output mode, compiler writes out S-expressions for each
```
saem:
```suggestion
In structured output mode, the compiler writes out S-expressions for each
```

doc/testament.rst
```diff
@@ -191,10 +213,129 @@ Example "template" **to edit** and write a Testament unit:
...
+output diagnostic entry, one per line.
+
+If testament is used in structured mode, all expected compiler reports -
+both inline and written in ``nimout`` are collected in a single list that
+is matched against produced output directly. Failure message shows best
```
saem:
```suggestion
is matched against produced output directly. The failure message shows the best
```

doc/testament.rst
```diff
@@ -191,10 +213,129 @@ Example "template" **to edit** and write a Testament unit:
...
+possible mismatch annotations for the given output. For example, given test
```
saem:
```suggestion
possible mismatch annotations for the given output. For example, given the test
```

doc/testament.rst
```diff
@@ -191,10 +213,129 @@ Example "template" **to edit** and write a Testament unit:
...
+like this, testament output will contain two mismatches for both failures.
```
saem:
```suggestion
below, testament output will contain two mismatches for both failures.
```

doc/testament.rst
```diff
@@ -191,10 +213,129 @@ Example "template" **to edit** and write a Testament unit:
...
+
+Compiler printed reports
+
+.. code-block:: nim
+
+    (User :severity Hint :str "User hint" :location ("tfile.nim" 8 6))
+    (User :severity Hint :str "Another hint" :location ("tfile.nim" 10 6))
+
+And they were matched against full list of expected entries. For the first
+entry there is a mismatch in ``:location[2]``, and for secdon one there is
```
saem:
```suggestion
entry there is a mismatch in ``:location[2]``, and for second one there is
```

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+
+  if prev < curr:
+    result.add str[prev ..< curr]
+
+proc formatDiffed*[T](
+    ops: seq[SeqEdit],
+    oldSeq, newSeq: seq[T],
+    conf: DiffFormatConf
+  ): tuple[oldLine, newLine: ColText] =
+  ## Generate colored formatting fothe levenshtein edit operation using
```
saem:
```suggestion
  ## Generate colored formatting for the levenshtein edit operation using
```

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+        newLine.add conf.chunk(newSeq[op.targetPos], sekInsert)
+        unchanged = 0
+
+      of sekReplace:
+        oldLine.add conf.chunk(oldSeq[op.sourcePos], sekReplace, sekDelete)
+        newLine.add conf.chunk(newSeq[op.targetPos], sekReplace, sekInsert)
+        unchanged = 0
+
+      of sekNone:
+        assert false
```
saem: Please add a string to this assert and use that was a way to describe why we should never got this branch.

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+    else: ($ch, $ch)
+
+proc toVisibleNames(conf: DiffFormatConf, str: string): string =
+  ## Convert all characters in the string into visible ones
+  for ch in str:
+    result.add conf.explainChar(ch)
+
+
+proc toVisibleNames(conf: DiffFormatConf, split: seq[string]): seq[string] =
+  ## Convert all charactersw in all strings into visible ones.
```
saem:
```suggestion
  ## Convert all characters in all strings into visible ones.
```

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+  if 0 < split.len():
+    for part in split:
+      result.add conf.toVisibleNames(part)
+
+const Invis = { '\x00' .. '\x1F', '\x7F' }
+
+func scanInvisible(text: string, invisSet: var set[char]): bool =
+  ## Scan string for invisible characters from right to left, updating
+  ## active invisible set as needed.
+  var chIdx = text.high
```
saem: Can we use a for and a countdown instead?

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+  while 0 <= chIdx:
+    if text[chIdx] in invisSet:
+      return true
+
+    else:
+      invisSet = Invis
```
saem, line 195: I don't get it, we immediately flip it over.

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+    dec chIdx
+
+func hasInvisible*(text: string, startSet: set[char] = Invis + {' '}): bool =
+  ## Does string have significant invisible characters?
+  var invisSet = startSet
+  if scanInvisible(text, invisSet):
+    return true
+
+func hasInvisible*(text: seq[string]): bool =
+  ## Do any of strings in text have signfificant invisible characters.
```
saem:
```suggestion
  ## Do any of strings in text have significant invisible characters.
```

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+  var idx = text.high
+  var invisSet = Invis + {' '}
+  while 0 <= idx:
+    # Iterate over all items from righ to left - until we found first
```
saem:
```suggestion
    # Iterate over all items from right to left - until we find the first
```

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+    # visible character space is also considered significant, but removed
```
saem:
```suggestion
    # visible character, space is also considered significant, but removed
```

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+    else:
+      chunks.add chunk
+
+  let groups =
+    if conf.groupInline:
+      # Group edit operations by chunk - `[ins], [ins], [ins] -> [ins, ins, ins]`
+      #
+      # This is not specifically important for insertions and deletions,
+      # but pretty much mandatory for the 'change' operation, if we don't
+      # want to end up with the `h->He->El->Lo->O` instead of
```
saem:
```suggestion
      # want to end up with `h->He->El->Lo->O` instead of
```

lib/experimental/colordiff.nim
```diff
@@ -0,0 +1,709 @@
...
+  coloredResult()
+  var first = true
+  proc addl() =
+    if not first:
+      add "\n"
+    first = false
+
+  if conf.groupLine:
+    # Grouping line edits is not possible in the side-by-side
+    # representation, so going directly for a
```
saem: Cut off?

lib/experimental/diff.nim
```diff
@@ -530,6 +531,30 @@ proc shiftDiffed*[T](
         result.newShifted.add((sekKeep, line.targetPos))
 
 
+iterator zipToMax*[T](lhs, rhs: seq[T], fill: T = default(T)):
+  tuple[lhs, rhs: T, rhsDefault, lhsDefault: bool, idx: int] =
+  ## Iterate each argumen to the end, filling in missing values with `fill`
```
saem:
```suggestion
  ## Iterate each argument to the end, filling in missing values with `fill`
```

lib/experimental/diff.nim
```diff
@@ -530,6 +531,30 @@ proc shiftDiffed*[T](
...
+  ## argument. This is an opposite of the std built-in `zip` which iterates
+  ## up until `min(lhs.len, rhs.len)`.
+  ##
+  ## This is an internal iteratr, exported to be shared with `colordiff`
```
saem:
```suggestion
  ## This is an internal iterator, exported to be shared with `colordiff`
```

lib/experimental/diff.nim
```diff
@@ -530,6 +531,30 @@ proc shiftDiffed*[T](
...
+  ## implementaion, but if you are implementing your own diff formatter,
+  ## you might consider using it as well.
```
saem: So is it internal or not?

lib/experimental/sexp_diff.nim
```diff
@@ -0,0 +1,406 @@
...
+    algorithm
+  ]
+
+type IdxCostMap* = Table[(int, int), int]
+
+proc randomKey[K, V](table: OrderedTable[K, V]): K =
+  for k, v in pairs(table):
+    return k
+
+proc stablematch*(
```
saem:
```suggestion
proc stableMatch*(
```

lib/experimental/sexp_diff.nim
```diff
@@ -0,0 +1,406 @@
...
+    proc(a, b: ((int, int), int)): int =
+      if a[1] == b[1]:
+        cmp(a[0], b[0])
+      elif order == Descending:
+        cmp(a[1], b[1])
+      else:
+        -cmp(a[1], b[1])
+  )
+
+  var seenLset: IntSet
```
saem:
```suggestion
  var seenLeft: IntSet
```

Seems more clear to me, but maybe I'm missing something?

lib/experimental/sexp_diff.nim
```diff
@@ -0,0 +1,406 @@
...
+
+  for idx in 0 ..< lhsLen:
+    if idx notin seenLset:
+      result.lhsIgnore.add idx
+
+export `$`, toString
+
+type
+  SexpPathPartKind = enum
+    ## Kind of the s-expression mismatch path art
```
saem:
```suggestion
    ## Kind of the s-expression mismatch path part
```

lib/pure/strutils.nim
```diff
@@ -2660,41 +2660,73 @@ func findNormalized(x: string, inArray: openArray[string]): int =
               # security hole...
   return -1
 
-func invalidFormatString() {.noinline.} =
-  raise newException(ValueError, "invalid format string")
+func invalidFormatString(explain: string) {.noinline.} =
+  ## Raise value erorr for invalid interpolation format string
```
saem:
```suggestion
  ## Raise value error for invalid interpolation format string
```

lib/pure/strutils.nim
```diff
@@ -2660,41 +2660,73 @@ func findNormalized(x: string, inArray: openArray[string]): int =
...
+  raise newException(ValueError, "invalid format string - " & explain)
```
saem, line 2665: Not a blocker but I prefer more specific error kinds, make more precise exception handling possible without shenanigans.

lib/pure/strutils.nim
```diff
@@ -2660,41 +2660,73 @@ func findNormalized(x: string, inArray: openArray[string]): int =
...
+    addfText ## Regular text fragment
+    addfPositional ## Positional fragment `$#`
+    addfIndexed ## Indexed fragment `$1`
+    addfDollar ## Dollar literal `$$`
+    addfBackIndexed ## Negative indexed fragment `$-1`
+    addfVar ## Interpolated variable `$name`
+    addfExpr ## Expression in braces `${some expr}`
+
+  AddfFragment* = object
+    ## `addf` format string fragment - can be used to write own text
```
saem:
```suggestion
    ## `addf` format string fragment - can be used to write your own text
```

testament/specs.nim
```diff
@@ -218,13 +226,22 @@ proc extractSpec(filename: string; spec: var TSpec): string =
   var col = 1
   while i < s.len:
     if (i == 0 or s[i-1] != ' ') and s.continuesWith(specStart, i):
-      # `s[i-1] == '\n'` would not work because of `tests/stdlib/tbase64.nim` which contains BOM (https://en.wikipedia.org/wiki/Byte_order_mark)
+      # `s[i-1] == '\n'` would not work because of
+      # `tests/stdlib/tbase64.nim` which contains BOM
+      # (https://en.wikipedia.org/wiki/Byte_order_mark)
```
saem, line 231: Nice!

testament/specs.nim
```diff
@@ -218,13 +226,22 @@ proc extractSpec(filename: string; spec: var TSpec): string =
...
-        raise newException(ValueError, "testament spec violation: duplicate `specStart` found: " & $(filename, a, b, line))
+        raise newException(
+          ValueError,
+          "testament spec violation: duplicate `specStart` found: " &
+            $(filename, a, b, line))
       elif line > lineMax:
-        # not overly restrictive, but prevents mistaking some `specStart` as spec if deeep inside a test file
-        raise newException(ValueError, "testament spec violation: `specStart` should be before line $1, or be indented; info: $2" % [$lineMax, $(filename, a, b, line)])
+        # not overly restrictive, but prevents mistaking some `specStart`
+        # as spec if deeep inside a test file
```
saem:
```suggestion
        # as spec if deep inside a test file
```

testament/testament.nim
```diff
@@ -32,10 +33,110 @@ var optVerbose = false
...
+        "Unexpected given annotation:\n\n? $#\n\n",
+        tcmp.expectedReports[give].node.toLine(sortfield = true)
+      )
+
 ## Blanket method to encaptsulate all echos while testament is detangled.
-## Using this means echo cannot be called with separation of args and must instead
-## pass a single concatenated string so that optional paramters can be
-## included
+## Using this means echo cannot be called with separation of args and must
+## instead pass a single concatenated string so that optional paramters can
```
```suggestion
## instead pass a single concatenated string so that optional parameters can
```

testament/testament.nim
```diff
@@ -398,32 +614,154 @@ proc checkForInlineErrors(r: var TResults, expected, given: TSpec, test: TTest,
...
+      r.expectedReports.add TOutReport(node: parseSexp(line))
+
+  for line in splitLines(given.nimout):
+    if 0 < line.len:
+      r.givenReports.add TOutReport(node: parseSexp(line))
+
+  var map = r.diffMap
+  proc reportCmp(a, b: int): int =
+    # Best place for further optimization and configuration - if more
+    # comparison speed isneeded, try starting with error kind, file, line
```
```suggestion
    # comparison speed is needed, try starting with error kind, file, line
```

testament/testament.nim
```diff
@@ -398,32 +614,154 @@ proc checkForInlineErrors(r: var TResults, expected, given: TSpec, test: TTest,
...
+    # comparison, they doing a regular msg != msg compare and only then
```
```suggestion
    # comparison, then doing a regular msg != msg compare, and only then
```

testament/testament.nim
```diff
@@ -398,32 +614,154 @@ proc checkForInlineErrors(r: var TResults, expected, given: TSpec, test: TTest,
...
+  return r
+
+proc cmpMsgs(
+    r: var TResults, expected, given: TSpec, test: TTest, target: TTarget
+  ) =
+  ## Compare all test output messages. This proc does structured or
+  ## unstructured comparison comparison and immediately reports it's
+  ## results.
+  ##
+  ## It is used to for performting 'reject' action checks - it compares
```
```suggestion
  ## It is used to for performing 'reject' action checks - it compares
```

testament/testament.nim
```diff
@@ -398,32 +614,154 @@ proc checkForInlineErrors(r: var TResults, expected, given: TSpec, test: TTest,
...
+  # Check for produced and given error message locations
+  elif expected.line != given.line and
+       expected.line != 0 or
+       expected.column != given.column and
+       expected.column != 0:
+    # Report error for the location mismatch
     r.addResult(test, target, $expected.line & ':' & $expected.column,
                       $given.line & ':' & $given.column, reLinesDiffer)
+
+  # None of the unstructructured checks found mismatches, reporting thest
```
```suggestion
  # None of the unstructured checks found mismatches, reporting the test
```

testament/testament.nim
```diff
@@ -398,32 +614,154 @@ proc checkForInlineErrors(r: var TResults, expected, given: TSpec, test: TTest,
...
+  # as passed.
   else:
     r.addResult(test, target, expected.msg, given.msg, reSuccess)
     inc(r.passed)
 
 proc generatedFile(test: TTest, target: TTarget): string =
+  ## Get path to the genrated file name from the test.
```
```suggestion
  ## Get path to the generated file name from the test.
```

testament/testament.nim
```diff
@@ -432,10 +770,19 @@ proc generatedFile(test: TTest, target: TTarget): string =
...
-                  given: var TSpec) =
+proc codegenCheck(
+    test: TTest,
+    target: TTarget,
+    spec: TSpec,
+    expectedMsg: var string,
+    given: var TSpec
+  ) =
+  ## Check for any codegen mismatches in file generated from `test` run.
+  ## Only file that was immediately generated is testsed.
```
```suggestion
  ## Only file that was immediately generated is tested.
```

tests/stdlib/tcolordiff.nim
```diff
@@ -0,0 +1,258 @@
+import experimental/[diff, colordiff, colortext]
```
saem, line 1: Not a blocker, but I feel like this should use unit test.

saem (changes requested):
Reviewed. There is plenty about the differencing I don't understand. So that'll be fun to dig into later.

Mostly typos and doc adjustments.

compiler/front/sexp_reporter.nim
haxscramper: > it's strange that it's allowed by the compiler.

again, **why**?

compiler/front/sexp_reporter.nim
haxscramper: What do you mean "can't iterator over T"? The `iterator` is defined for `T` therefore I can iterate over it, I don't see any bugs here

compiler/front/sexp_reporter.nim
haxscramper: no

tests/stdlib/tcolordiff.nim
haxscramper: just like every single test out there, right, when unittest becomes usable we can use it

krux02:
what and why printing category `""`?

haxscramper:
I don't understand the question. The check for `tests/**/t**.nim` file was removed to allow `testament run tlocalfile.nim` - is that what you mean?

krux02:
> I don't understand the question. The check for `tests/**/t**.nim` file was removed to allow `testament run tlocalfile.nim` - is that what you mean?

What I mean is the empty quotes in ``Test "tfile_regular.nim" in category ""``

haxscramper:
Fixed empty quotes, now it prints only test name for empty categories

compiler/front/sexp_reporter.nim
saem: Because a type `T` is not guaranteed to have an `items`, yes?

compiler/front/sexp_reporter.nim
saem: `T` is not guaranteed to have fields for `fieldPairs` to work, it might be out of scope for this PR, but it's silly that between the stdlib and the compiler that this is allowed.

compiler/front/sexp_reporter.nim
haxscramper: It won't compile when I instantiate the proc like every other generic, so I guess this is not allowed? I just don't get why this remark is attached to a definition. sexp two lines above can have $ defined as {.error.} and it would've been an instantiation error as well

compiler/front/sexp_reporter.nim
haxscramper: You are talking about need for 'HasFields' concept or something like that here, otherwise it is a perfectly fine generic that would generate a perfectly fine compilation error on instantiation saying that there are no fields in here. I don't understand where the "bug" part comes from here. Just like the sexpItems above - it is instantiation error, just like every other generic proc. I can put object|tuple there, but that would just make code clearer, it won't make it more correct (as in - reject malformed program that would've compiled previously) in any way. If there is no field pairs the code does not compile, pointing to "instantiation of X...". If there are field pairs the code compiles.

saem (approved):
Bors r+

Merge me up, Scotty!

PR #177, opened 2022-01-19
Sort compiler files into more manageable structure

haxscramper:
Organized compiler source code into smaller sections - if we plan to untangle implementation parts from each other, we can start small, with simple file organization and later continue to remove unnecessary cross-directory imports and so on. 

current moved files:

```
 { => ast}/ast.nim 
 { => ast}/ast_types.nim 
 { => ast}/astalgo.nim 
 { => ast}/astmsgs.nim 
 { => ast}/enumtostr.nim 
 { => ast}/errorhandling.nim 
 { => ast}/errorreporting.nim 
 { => ast}/layouter.nim 
 { => ast}/lexer.nim 
 { => ast}/lineinfos.nim 
 { => ast}/linter.nim 
 { => ast}/llstream.nim 
 { => ast}/ndi.nim 
 { => ast}/nimlexbase.nim 
 { => ast}/nimsets.nim 
 { => ast}/parser.nim 
 { => ast}/renderer.nim 
 { => ast}/reports.nim 
 { => ast}/syntaxes.nim 
 { => ast}/trees.nim 
 { => ast}/treetab.nim 
 { => ast}/types.nim 
 { => ast}/typesrenderer.nim 
 { => ast}/wordrecg.nim 
 { => backend}/ccgcalls.nim 
 { => backend}/ccgexprs.nim 
 { => backend}/ccgliterals.nim 
 { => backend}/ccgmerge_unused.nim 
 { => backend}/ccgreset.nim 
 { => backend}/ccgstmts.nim 
 { => backend}/ccgthreadvars.nim 
 { => backend}/ccgtrav.nim 
 { => backend}/ccgtypes.nim 
 { => backend}/ccgutils.nim 
 { => backend}/cgen.nim 
 { => backend}/cgendata.nim 
 { => backend}/cgmeth.nim 
 { => backend}/extccomp.nim (99%)
 { => backend}/jsgen.nim 
 { => backend}/jstypes.nim 
 { => backend}/tccgen.nim 
 { => font}/cli_reporter.nim 
 { => font}/cmdlinehelper.nim 
 { => font}/commands.nim 
 { => font}/condsyms.nim 
 { => font}/main.nim 
 { => font}/msgs.nim 
 { => font}/nimconf.nim 
 { => font}/options.nim 
 { => font}/scriptconfig.nim 
 { => modules}/depends.nim 
 { => modules}/importer.nim 
 { => modules}/magicsys.nim 
 { => modules}/modulegraphs.nim 
[...]

PR #207, opened 2022-01-27
Unbundle drnim

haxscramper:
- Removes main drnim directory along with all the hacks added into `sem*`
  implementation - special keywords for
  `ensures/requires/assume/invariant`, more logic built into sempass and so
  on. Sem reports that were added in the main compilation.
- External tooling for code analysis should better be implemented via
  completely external modifications, instead of clobbering the main
  compiler codebase with logic for barely functioning tooling. With DOD
  refactor this task becomes much easier, since extra metadata can be
  attached to any symbol.
- We definitely want to move in the direction of automatic code correctness
  and static analysis, but right now this is not a priority, and
  considering dubious implementation of the tool, it would be easier to
  remove it for now and come up with a proper implementation later on, or
  when can turn all hacks back into sem.
- drnim itself Not tested as a part of CI so right now it can't even be built, due to
  file reorganization https://github.com/nim-works/nimskull/pull/177.
- Adds reports for `--staticBounds:on|off` that were mistakenly assumed to
  be drnim-only during earlier refactor
  https://github.com/nim-works/nimskull/pull/94 for structured reports.

Closes https://github.com/nim-works/nimskull/pull/130 as this pull removes
`drnim` as well, and the author of that PR is no longer responsive. Due to
multiple merge conflicts it is highly unlikely the old PR will be revived.


saem (approved):
LGTM

bors r+

PR #209, opened 2022-01-28
Refactor active configuration handing

### LINKED ISSUE #158: Refactor active configuration handling
Refactor active configuration handling in the `options.ConfigRef` - right now every single field is placed on the same level, including both mutable non-configuration values (error counter, list of symbols) and active/passive configurations (list of active notes, local and global options). 

Field that are used to store explicit configuration values (backend/target/options/globalOptions) need to be factored out into a separate type definition.

- This would allow treating compiler (at least on conceptual level) as a function `proc compile(config: ActiveOptions)`. 
- This type can be reused in testament to serve as a configuration cell in test matrix
- Command-line configuration reading can be simplified as well - instead of using a whole `ConfigRef` object and mutating it in-place, it can simply return `ActiveOptions` object

Old code should continue to work using getters and setters. Additional upside - this change would provide a single point of tracing for the active configuration and reduce number of direct access to fields (which cannot be debugged with `echo`, cannot be asserted and so on).

```diff
type
+ ActiveOptions = object
+   options*: TOptions

  ConfigRef* {.acyclic.} = ref object 
-   options*: TOptions
+   active: ActiveOptions

+ func options*(conf: ConfigRef): TOptions = conf.active.options
+ proc `options=`*(conf: ConfigRef, opts: TOptions) = conf.active.options = opts
```

Repeated for all "input" fields: `backend`, `target`, `options`, `filenamesOption`

haxscramper:
- Move active configuration handing into a separate type that should
  contain all the required configuration parameters. Use this type in
  `ConfigRef` module, and use accessor procs for getting and setting values
  of different options, thus removing unguarded access for fields in a
  number of cases.
- Move report kind and category enums into `report_enums` in order to make
  them accessible from the `in_options` - it must be a lightweight
  dependency since it is supposed to be reused by other tools, like
  testament, that have to directly interact with the compiler.

Closes https://github.com/nim-works/nimskull/issues/158


saem:
Did a first quick read, there are a few things I need to look at again in detail, including doc review. I'll read it again tonight/tomorrow, but overall seems reasonable.

haxscramper:
There is a 'config var' system, in addition to the regular active configuration https://github.com/nim-works/nimskull/blob/829ffa1fb6a557bd7e65485383273bafcbde2378/compiler/front/options.nim#L366 field that is accessed via https://github.com/nim-works/nimskull/blob/829ffa1fb6a557bd7e65485383273bafcbde2378/compiler/front/options.nim#L1091-L1098 and contains more random bits of user configurations 

https://github.com/nim-works/nimskull/blob/829ffa1fb6a557bd7e65485383273bafcbde2378/compiler/backend/extccomp.nim#L336

https://github.com/nim-works/nimskull/blob/829ffa1fb6a557bd7e65485383273bafcbde2378/compiler/backend/extccomp.nim#L431

https://github.com/nim-works/nimskull/blob/829ffa1fb6a557bd7e65485383273bafcbde2378/compiler/backend/extccomp.nim#L436

https://github.com/nim-works/nimskull/blob/829ffa1fb6a557bd7e65485383273bafcbde2378/compiler/backend/extccomp.nim#L441

mixed in with docgen

https://github.com/nim-works/nimskull/blob/829ffa1fb6a557bd7e65485383273bafcbde2378/compiler/tools/docgen.nim#L891-L895

https://github.com/nim-works/nimskull/blob/829ffa1fb6a557bd7e65485383273bafcbde2378/compiler/tools/docgen.nim#L1360

of course all of this is barely documented, but it controls important configuration values for the compiler. Maybe having some form of 

```nim
type
  ExtraVarsEnum = enum
    DocSectionTok = "doc.section.toc" ## Documentation for this `--doc.section.tok=` flag
```

is possible, but there are ~11x(4+) options for each compiler and so on, they must be somehow documented as well, but adding a giant copy-pasted enum does not really seem like a good solution. Maybe some form of nested variant for this sort of thing instead?

```nim
type
  ExtraVars = object
    case kind*: ExraVarsSectino
      of ExtraDoc:
         section: tuple[tok: string, ...]
      of ExtraCC:
         array: array[TSystemCC, ExtraCCOption]
```

And so on?

haxscramper:
That part can be done after https://github.com/nim-works/nimskull/issues/213

compiler/ast/reports.nim
```diff
@@ -29,942 +29,13 @@ export
   options.Option,
   int128.toInt128
 
+
+from front/in_options import TOption, TOptions
 type InstantiationInfo* = typeof(instantiationInfo())
 
+import report_enums
+export report_enums
```
saem, line 37: mid-file imports are usually a smell; doubly so with exports, but at least it's showing that there is an issue here, so better than hiding it by shoving them at the top.

compiler/front/in_options.nim
```diff
@@ -0,0 +1,318 @@
...
+    cnMainPackage
+    cnForeign
+    cnWarnAsError
+    cnHintAsError
+    cnCmdline ## notes that have been set/unset from cmdline
+    cnModifiedy ## notes that have been set/unset from either
+                ## cmdline/configs
+
+type
+  ActiveConf* = object
```
saem:
```suggestion
  CurrentConf* = object
```

I suggest renaming this to `CurrentConf` as it's the current understanding of the compiler configuration/input.

As a stretch goal rename ConfigRef to `CompilerStateRef`, because it's a hilarious global -- or we can tackle that separately.

compiler/front/options.nim
```diff
@@ -315,71 +135,78 @@ type
                           ## fields marked with '*' are subject to
                           ## the incremental compilation mechanisms
                           ## (+) means "part of the dependency"
-    backend*: TBackend ## set via `nim x` or `nim --backend:x`
-    target*: Target       # (+)
+
+    # active configuration handling
+    active*: ActiveConf
```
saem:
```suggestion
    # REFACTOR - rename `ConfigRef` to something like `CompilerStateRef` because it's just global
    # data, then `active` can be renamed `conf`.
    active*: CurrentConf
      ## current compiler configuration based on initial inputs
```

We can leave this as `active`, but `current` doesn't sound better either. The problem is that `ConfigRef` isn't right, it's just global compiler state dump... as suggested above if you want to tackle the `ConfigRef` to something like `CompilerStateRef` then that works, in which case this might just be `conf` or something.

Personally, I think that's a big change, I'd rather leave a refactor comment here outlining it, hence my suggestion (line wrap might be off).

compiler/front/options.nim
```diff
@@ -432,13 +254,103 @@ type
       debugUtilsStack*: seq[string] ## which proc name to stop trace output
       ## len is also used for output indent level
 
+template passField(fieldname, fieldtype: untyped): untyped =
+  proc `fieldname`*(conf: ConfigRef): fieldtype =
+    conf.active.fieldname
+
+  proc `fieldname=`*(conf: ConfigRef, val: fieldtype) =
+      conf.active.fieldname = val
```
saem:
```suggestion
    conf.active.fieldname = val
```

compiler/front/options.nim
```diff
@@ -944,66 +845,54 @@ proc initConfigRefCommon(conf: ConfigRef) =
...
+      backend:        backendInvalid,
+      cppDefines:     initHashSet[string](),
+      features:       {},
+      legacyFeatures: {},
+      cCompiler:      ccGcc,
+      macrosToExpand: newStringTable(modeStyleInsensitive),
+      arcToExpand:    newStringTable(modeStyleInsensitive),
+      outFile:        RelativeFile"",
+      outDir:         AbsoluteDir"",
+      dllOverrides: newStringTable(modeCaseInsensitive),
```
saem:
```suggestion
      dllOverrides:   newStringTable(modeCaseInsensitive),
```

compiler/front/options.nim
```diff
@@ -432,13 +254,103 @@ type
...
+  proc `fieldname`*(conf: ConfigRef): fieldtype =
+    conf.active.fieldname
+
+  proc `fieldname=`*(conf: ConfigRef, val: fieldtype) =
+      conf.active.fieldname = val
+
+passField(backend, TBackend)
+passField(target, Target)
+
+template passSetField(fieldname, fieldtype, itemtype: untyped): untyped =
```
saem, line 265: Can we collect all these helper templates together that way you just scroll past them and read the defines all in one continuous chunk rather than being interrupted.

saem (changes requested):
Thanks, this is starting to place the state in better areas, and also for the walkthrough in chat. 👍🏽 

Small name change and refactor comment for those who come back to it, otherwise g2g.

compiler/ast/reports.nim
haxscramper: There is no issue here, enums are just reused for both reports and active configuration, actual reports are reused only inside of the compiler, so it does not make sense to import them in active configuration as well. But report enums are used together with reports, so having to double each of the ~100 imports would've been a lot worse.

Clyybber (approved):

saem (approved):
bors r+

yay! 🎉

PR #212, opened 2022-01-28
tgetprotobyname: remove "ip" test query

krux02:
getProtoByName aparently fails on NetBSD. It also fails on my Arch
machine and my Manjaro machine. Removing this test is probably a good
idea.

### DIFF
--- a/tests/stdlib/tgetprotobyname.nim
+++ b/tests/stdlib/tgetprotobyname.nim
@@ -1,9 +1,5 @@
 import nativesockets
 
-when not defined(netbsd):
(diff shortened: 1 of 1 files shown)


alaviss:
LGTM

Please change the commit title to the same as that of the edited PR title then you can bors it.

I think the new title is more clear on what the commit does without diving too deep.

saem:
Bors r+

This is definitely conflating multiple things in one test, other bits can be figured out later.

krux02:
I applied the commit message change now the full test suite runs again. Would be nice if github would realized that only the commit message changed and therefore it can't affect the result of tests.

alaviss:
bors r+

Don't worry, just bors it immediately, it runs the full test suite there anyway

PR #214, opened 2022-01-28
`genDepend` improvements

haxscramper:
- Track optional includes
- Generate more readable graph
- Generate graph in two stages - collecting data and then writing out it in
  an analyzed form.

before (empty file, show default processed modules)

tfile

after (same empty file, but also shows included files, modules that might've been included, with necessary conditions)

tfile

todo

- [ ] allow tracking module starting from a specific file in the project instead of tracking every possible dependency
- [ ] Filter out most common least common nodes - make it possible to actually see the patterns in the graph, instead of unconditionally producing unmaintainable mess because `ast.nim` is imported literally **everywhere** - 
image does not help anyone to understand anything. Everyone imports 'reports' because everyone writes reports, I already know this, and I don't care about this, I want to get down to the other parts of the structure. Everyone has `utils` modules for common stuff. The compiler should actually consider the most common use cases here.
- [ ] output data in the structured manner - we can't predict every possible use case, nor should `genDepend` turn into an all-purpose, all-powerful code graphing tool. A set of sensible defaults and improved graph formatting + documented json output is a nice compromise. 
- [ ] Group nodes into directories based on the project structure - graphviz supports `subgraph cluster_XXX` - can be used in that case.
- [ ] Allow omitting the interdirectory dependencies. For example, you want to check what is going on inside the sem/ directory, but you don't care about intermodule dependencies outside this part, you just want to know if sem imports tables for example, or something similar. In that case, this 
image is most likely better than this 
image - latter version also would include the world of dependencies outside, but it is hardly necessary, I only care about the mess we have in `sem/*`, it is already bad enough
- [ ] Fix `-o` for `genDepend` - right now it is hardcoded to write to the same file as source project https://github.com/nim-works/nimskull/blob/829ffa1fb6a557bd7e65485383273bafcbde2378/compiler/front/main.nim#L84-L93
- [ ] Commit message
- [ ] Documentation for the use cases, examples, for trivial projects.
- [ ] Use html styling on edge labels
- [ ] Documentation on the implementation
- [ ] Writing dependencies with empty nimcache directory won't create the directory, failing the analysis
- [ ] Generate dependency graph with unique nodes. This would allow to use https://graphviz.org/docs/layouts/twopi/ to analyze the dependency layers. Maybe color-code nodes to make it easier to spot things.

### DIFF
--- a/compiler/front/main.nim
+++ b/compiler/front/main.nim
@@ -75,8 +75,9 @@ proc writeDepsFile(g: ModuleGraph) =
   f.close()
 
 proc commandGenDepend(graph: ModuleGraph) =
+  registerPass(graph, genDependPreSemPass)
   semanticPasses(graph)
-  registerPass(graph, gendependPass)
+  registerPass(graph, genDependPostSemPass)
   compileProject(graph)
   let project = graph.config.projectFull
   writeDepsFile(graph)
--- a/compiler/modules/depends.nim
+++ b/compiler/modules/depends.nim
@@ -11,22 +11,38 @@
 
 import
   ast/[
-    ast
+    ast,
+    lineinfos,
+    renderer
   ],
   modules/[
     modulepaths,
     modulegraphs
   ],
   front/[
-    options
+    options,
+    msgs
   ],
   utils/[
     ropes,
-    pathutils
+    pathutils,
+    astrepr
   ],
   sem/[
     passes
+  ],
+  std/[
+    tables,
+    sets,
+    hashes,
+    sequtils,
+    strutils
   ]
+
+import std/options as std_options
+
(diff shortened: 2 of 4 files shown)


haxscramper:
I think this PR will be eventually superseded by https://github.com/nim-works/nimskull/pull/224, since it tracks the same data, and graphs can be generated from documentable database instead, via basic `SQL` queries.

haxscramper:
Closing because conflicts started to pile up and it is **clearly** something that should be done as a part of https://github.com/nim-works/nimskull/pull/224 or on top of it. I'm still planning to implement all the outlined features (eventually), but this particular implementation is likely a dead end.

PR #215, opened 2022-01-29
automatic support for backwards index

krux02:
All types that have the `[]` operator and the `len` overloaded can now
also be accesed using a backwards index (^1). Old versions of Nim
already supported this, but it got lost over time. So this is actually a
regression fix.

What it takes away though is accessing enum indexed
arrays (``array[MyEnum, T]``) via backwards index. Not sure why that was
ever supported, after all you can't even access them with normal forward
indices, so why should backwards indices be allowed?

This is actually a revival of my old PR here: https://github.com/nim-lang/Nim/pull/7265

### DIFF
--- a/lib/system.nim
+++ b/lib/system.nim
@@ -2542,8 +2542,40 @@ template spliceImpl(s, a, L, b: untyped): untyped =
   # fill the hole:
   for i in 0 ..< b.len: s[a+i] = b[i]
 
-template `^^`(s, i: untyped): untyped =
-  (when i is BackwardsIndex: s.len - int(i) else: int(i))
+template `^^`*(s: typed; i: BackwardsIndex): int =
+  ## General converter template to convert the Backwards index back to
+  ## a normal integer index. If your own type has a special conversion
+  ## from Backwards index to int, you can write your own overload,
+  ## otherwis it will be just default to this: `s.len - int(i)`.
+  ##
+  ## .. code-block:: nim
+  ##
+  ##   type
+  ##     MyType = object
+  ##
+  ##   template `^^`(a: MyType, b: BackwardsIndex): int =
+  ##     ## my conversion of `BackwardsIndex` for `MyType`
+  ##     1000 + int(b)
+  ##
+  ##   proc `[]`(arg: MyType; idx: int): int =
+  ##     idx
+  ##
+  ##   var mt: MyType
+  ##   echo mt[^1] # 1001
+  s.len - int(i)
+
+template `^^`*[Idx, T](a: array[Idx, T], i: BackwardsIndex): int =
+  ## Converter template specialized for the array type.
+  int(a.low) + a.len - int(i)
+
+template `^^`*[Idx, T](a: array[Idx, T], i: Idx): int =
+  ## Converter template specialized for the array type.
+  int(i)
+
+template `^^`*(s: typed; i: int): int =
+  ## No operation in case the index is not a backwards index. For
+  ## generic programming only.
+  i
 
 template `[]`*(s: string; i: int): char = arrGet(s, i)
 template `[]=`*(s: string; i: int; val: char) = arrPut(s, i, val)
@@ -2586,9 +2618,11 @@ proc `[]`*[Idx, T; U, V: Ordinal](a: array[Idx, T], x: HSlice[U, V]): seq[T] =
   ##    var a = [1, 2, 3, 4]
   ##    assert a[0..2] == @[1, 2, 3]
   let xa = a ^^ x.a
-  let L = (a ^^ x.b) - xa + 1
+  let L: int = (a ^^ x.b) - xa + 1
   result = newSeq[T](L)
-  for i in 0..<L: result[i] = a[Idx(i + xa)]
+  for i in 0..<L:
+    let idx: Idx = Idx(i + xa)
+    result[i] = a[idx]
 
 proc `[]=`*[Idx, T; U, V: Ordinal](a: var array[Idx, T], x: HSlice[U, V], b: openArray[T]) =
   ## Slice assignment for arrays.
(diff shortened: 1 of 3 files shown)


krux02:
regarding `tests/concepts/tmapconcept.nim`. it will fail, because concepts.

saem:
I think we need one more piece of information before assuming backwards indexing, not quite sure what it is.

Some explicit indicator for backwards indexing as allowed is a requirement -- well could be indicated subtlety and then done through concepts.

Anyhow, the reason is for example infinite sequences, the len requirement nicely takes care of that.

There are streams or collections that are only deterministically ordered in one direction because of the peculiarities of their traversals. Comparison operator presence in the inner type in a generic variety of such thing would also not be sufficient because the produced items might be well ordered otherwise. Although maybe that's encouraging poor design/under specification.

lib/system.nim
```diff
@@ -2542,8 +2542,40 @@ template spliceImpl(s, a, L, b: untyped): untyped =
   # fill the hole:
   for i in 0 ..< b.len: s[a+i] = b[i]
 
-template `^^`(s, i: untyped): untyped =
-  (when i is BackwardsIndex: s.len - int(i) else: int(i))
+template `^^`*(s: typed; i: BackwardsIndex): int =
```
alaviss, line 2545: I find `^^` to be a weird operator for this purpose (I mean, just look at this: `s ^^ ^1`, it doesn't look like something that make a lot of sense). Since this is not the kind of operator that has to be short (only used by `[^x]` and friends), I think it should have a name instead. I don't have any good name ideas now, though.

lib/system.nim
```diff
@@ -2542,8 +2542,40 @@ template spliceImpl(s, a, L, b: untyped): untyped =
...
+  ##   echo mt[^1] # 1001
+  s.len - int(i)
+
+template `^^`*[Idx, T](a: array[Idx, T], i: BackwardsIndex): int =
+  ## Converter template specialized for the array type.
+  int(a.low) + a.len - int(i)
+
+template `^^`*[Idx, T](a: array[Idx, T], i: Idx): int =
+  ## Converter template specialized for the array type.
+  int(i)
```
alaviss, line 2573: Shouldn't these emit `Idx`? Else it won't work with non-int based array.

lib/system.nim
krux02: any suggestions? I also prefer names.

lib/system.nim
krux02: it needs to be `int`, otherwise you can't subtract it from `len`. And yes backwards indexing doesn't work for `non-int` based array. But why should it, forward indexing `[0]` also doesn't work.

krux02:
> I think we need one more piece of information before assuming backwards indexing, not quite sure what it is.

The backwards index is the thing to index arrays from the back. Instead of `x[x.len - 1 - i]` just `x[^1]`. At least it is exactly that for me. Nothing more and nothing less. Unfortunately Nim has these arrays that don't start at index 0, but at an arbitrary number. I have no idea what the point of that is, but support for it is hacked in. Then Nim also supports enums as array index. Backwards indices are not that much necessary as much as I can tell. Maybe I can reenable them, for now I just reenabled this PR, didn't look to much if I can actually improve it.

> Some explicit indicator for backwards indexing as allowed is a requirement -- well could be indicated subtlety and then done through concepts.

I do not know what you mean here. The entire point of this PR is that you do not need to enable backwards index for your type. It just works out of the box, but you can customize backwards index behavior if you want or need to. The idea here is provide advance access, even though the library only provides a basic interface.
 
> Anyhow, the reason is for example infinite sequences, the len requirement nicely takes care of that.

I don't know about any infinite sequences in the Nim universe. I know about them in haskell but I don't see them to be a thing in the Nim world. Do we plan to support them?

> There are streams or collections that are only deterministically ordered in one direction because of the peculiarities of their traversals. Comparison operator presence in the inner type in a generic variety of such thing would also not be sufficient because the produced items might be well ordered otherwise. Although maybe that's encouraging poor design/under specification.

Pleas rephrase this. I do not know what you want to tell me here. How are streams related to backwards index or even the index operator?

saem:
I covered a bunch of stuff, let me cut it down.

Yes, infinite sequences, sparse collections/streams/whatever are supported and not to be second classed.

This doesn't work for sparse collections, for example a map with int keys qualifies but it doesn't make sense to assume a sort.

lib/system.nim
alaviss: But you can convert it back after you're done subtracting? It's how the original implementation for arrays work, too.

```nim
type
  E = enum
    A
    B
    C
    D

var x: array[E, int]

x[C] = 42
echo x[^2]
```

This compiles right now and won't with your version.

alaviss:
Case in point for opt-in support: `Table[int, T]` which can be sparse and have `len()` defined.

Example:

```nim
import std/tables

var x: Table[int, string]

x[10] = "string"

echo x[^1] # <- This will resolve to `x[0]` due to len(x) == 1
           # Backward indexing doesn't make sense for this type.
```

lib/system.nim
alaviss: How about `toIndex(s, i)`? Though that's probably too generic...

saem: toIndex is fine, the export in system is the issue, only a few library authors would really need this.

lib/system.nim
saem: The index at least would have to be an ordinal, then if it covers to a contiguous integer range we can subtract.

Otherwise, I suppose we could go out on a limb and say ordinal is a stable order for the collection itself, this isn't strictly true and hence my overall reservations. I do like what this is trying to do, but our type system needs to catch-up to the ambitions of the PR.

Clyybber: A type system that would enable this PR without the issues would need a way to annotate collections to mark them as eglible for backwards indexing.
But we already have a way to mark collections as such, simply by implementing \`[]\`(i: BackwardsIndex).
I don't think this PR is a good idea, the concept of backwards indexing is not universial and thus needs to be opt-in for collections and implementing the appropriate overloads is the simplest and most flexible (allows skipping len call in a linked list for example) way to opt-in.

saem:
Given the discussion, not moving ahead with this, hopefully with concepts/interfaces or some such this should be easy to implement in a type safe manner.

PR #216, opened 2022-01-29
debug: detailed information about visited locations

haxscramper:
- Add more detailed information about visited locations in the cyclic
  ASTs - show node numbers, and display `<visisted Kind in ID>` (e.g.
  `<visited Sym in 0>`
- Ignore repeated visitations of the `Empty` node - they are reused, so it
  only adds confusion, since ast looks like it is insanely cyclic.
- Fix duplicated `.comment` annotations from the `CommentStmtNode` -
  `addFlags()` and `addComment()` was called twice.

### DIFF
--- a/compiler/utils/astrepr.nim
+++ b/compiler/utils/astrepr.nim
@@ -14,7 +14,7 @@ import
   ],
   experimental/colortext,
   std/[
-    sets,
+    tables,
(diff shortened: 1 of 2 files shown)


saem (approved):
LGTM! :D

PR #219, opened 2022-01-31
niminst: generate archive manifest after build

alaviss:
This allows tools to figure out what was created without doing guesswork
on file name.

This will be used by the continuous delivery pipeline.

### DIFF
--- a/tools/niminst/niminst.nim
+++ b/tools/niminst/niminst.nim
@@ -22,6 +22,7 @@ const
   deinstallShFile = "deinstall.sh"
   csourcesReleaseFile = "csources-release"
   releaseFile = "release.json"
+  archiveManifestFile = "archive.json"
 
 type
   AppType = enum appConsole, appGUI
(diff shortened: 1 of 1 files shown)


tools/niminst/niminst.nim
```diff
@@ -769,6 +770,51 @@ proc checkedShellExec(cmd: string) =
   if exitCode != 0:
     raise newExternalProgramError(exitCode)
 
+proc computeChecksum(file: string): string =
+  ## Calculate the SHA256 hash for `file`.
+  let p = startProcess(
```
saem, line 775: Hah, I thought we'd have a standard library call for this, oh well.

tools/niminst/niminst.nim
```diff
@@ -769,6 +770,51 @@ proc checkedShellExec(cmd: string) =
...
+
+  defer: close p
+
+  let exitCode = p.waitForExit()
+
+  if exitCode != 0:
+    stderr.write p.outputStream.readAll()
+    raise newExternalProgramError(exitCode)
+  else:
+    # OpenSSL produce an output in this format:
```
saem:
```suggestion
    # OpenSSL produces output in this format:
```

tools/niminst/niminst.nim
```diff
@@ -769,6 +770,51 @@ proc checkedShellExec(cmd: string) =
...
+    version: string ## The project version
+    case kind: ArchiveKind ## The archive type, in the generated JSON this field
+                           ## will be a string.
+    of Binary:
+      os: string ## The binary archive target OS
+      cpu: string ## The binary archive target CPU
+    of Source:
+      discard
+
 proc archiveDist(c: var ConfigData) =
```
saem:
```suggestion
proc createArchiveDist(c: var ConfigData) =
```

If we only day it a few times then it's not worth saving the characters and it completely obviates the comment.

saem (changes requested):
Minor, otherwise g2g

PR #225, opened 2022-02-02
fix hints and warnings

krux02:
no actual code changes. Just dead code that gets deleted.

### DIFF
--- a/lib/pure/strutils.nim
+++ b/lib/pure/strutils.nim
@@ -1493,7 +1493,7 @@ func delete*(s: var string, slice: Slice[int]) =
     a.delete(1..<1) # empty slice
     assert a == "ad"
   when compileOption("boundChecks"):
-    if not (slice.a < s.len and slice.a >= 0 and slice.b < s.len):
+    if (slice.a > s.len) or (slice.a < 0) or (slice.b >= s.len):
       raise newException(IndexDefect, $(slice: slice, len: s.len))
   if slice.b >= slice.a:
     var i = slice.a
@@ -1626,7 +1626,7 @@ func removePrefix*(s: var string, chars: set[char] = Newlines) {.rtl,
 
   var start = 0
   while start < s.len and s[start] in chars: start += 1
-  if start > 0: s.delete(0, start - 1)
+  if start > 0: s.delete(0 ..< start)
 
 func removePrefix*(s: var string, c: char) {.rtl,
     extern: "nsuRemovePrefixChar".} =
@@ -1654,7 +1654,7 @@ func removePrefix*(s: var string, prefix: string) {.rtl,
     answers.removePrefix("yes")
     doAssert answers == "yes"
   if s.startsWith(prefix):
-    s.delete(0, prefix.len - 1)
+    s.delete(0 ..< prefix.len)
 
 func removeSuffix*(s: var string, chars: set[char] = Newlines) {.rtl,
     extern: "nsuRemoveSuffixCharSet".} =
@@ -2449,7 +2449,7 @@ func trimZeros*(x: var string; decimalSep = '.') =
     var pos = last
     while pos >= 0 and x[pos] == '0': dec(pos)
     if pos > sPos: inc(pos)
-    x.delete(pos, last)
+    x.delete(pos .. last)
 
 type
   BinaryPrefixMode* = enum ## The different names for binary prefixes.
--- a/lib/system/assign.nim
+++ b/lib/system/assign.nim
@@ -40,19 +40,20 @@ proc genericAssignAux(dest, src: pointer, n: ptr TNimNode,
   #  echo "ugh memory corruption! ", n.kind
   #  quit 1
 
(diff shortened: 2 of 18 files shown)


lib/system/assign.nim
```diff
@@ -40,19 +40,20 @@ proc genericAssignAux(dest, src: pointer, n: ptr TNimNode,
   #  echo "ugh memory corruption! ", n.kind
   #  quit 1
 
-template deepSeqAssignImpl(operation, additionalArg) {.dirty.} =
-  var d = cast[ptr NimSeqV2Reimpl](dest)
-  var s = cast[ptr NimSeqV2Reimpl](src)
-  d.len = s.len
-  let elem = mt.base
-  d.p = cast[ptr NimSeqPayloadReimpl](newSeqPayload(s.len, elem.size, elem.align))
+when defined(nimSeqsV2):
```
saem, line 43: In a future PR should we simply permanently switch to seqv2 and be done with v1?

lib/system/chcks.nim
```diff
@@ -32,11 +32,11 @@ proc raiseFieldError(f: string) {.compilerproc, noinline.} =
 when defined(gcdestructors):
   proc raiseFieldError2(f: string, discVal: int) {.compilerproc, noinline.} =
     ## raised when field is inaccessible given runtime value of discriminant
-    sysFatal(FieldError, f & $discVal & "'")
+    sysFatal(FieldDefect, f & $discVal & "'")
```
saem: Keep these as errors for now.

saem (changes requested):

lib/system/assign.nim
krux02: I have no opinion on nimSeqsV2, but I do have an opinion on nim-gdb. And that part was written with current seq implementation. I personally think that gdb support is more important than a new seq version. I also wrote a test for it
in `tests/untestable/gdb/gdb_pretty_printer_test_run.sh`. It is in `untestable` because I wasn't able to integrate it into `testament` at that point in time. That should be done though. Currently it is manual testing only.

lib/system/chcks.nim
krux02: Any reasons why?

lib/pure/strutils.nim
```diff
@@ -1493,7 +1493,7 @@ func delete*(s: var string, slice: Slice[int]) =
     a.delete(1..<1) # empty slice
     assert a == "ad"
   when compileOption("boundChecks"):
-    if not (slice.a < s.len and slice.a >= 0 and slice.b < s.len):
+    if (slice.a > s.len) or (slice.a < 0) or (slice.b >= s.len):
```
krux02, line 1496: This one has a slight semantic change. An empty slice (`0 ..< 0` or `0 .. -1` is it is internally) is now valid.

saem: Empty slice makes sense to me, we got a test for this?

lib/system/chcks.nim
saem: Because defects is half-baked and even when it seems irrecoverable, turns out it is recoverable in a number of circumstances. Runtime errors suck, wise yet when we make them a chore to handle.

krux02: Yea, I wasn't really involved in the discussion, I thought it was just a rename. I will revert the change, but the warning is still there to use Defect instead of Error. So I don't know what to do about it.

lib/pure/strutils.nim
krux02: yes I added a test in `tstrutils` for it.

krux02:
It's ready

saem:
Sorry reviewing this while very tired, is the delete change required for fixing the warnings and hints?

At the very least the commit/PR should mention that in the title; or be a separate PR. It would easily be overlooked in putting together a changelog.

krux02:
> Sorry reviewing this while very tired, is the delete change required for fixing the warnings and hints?

Yes it is. I fixed a deprecation warning by replacing `delete` with another overload of `delete` just to find out that they are not equivalent. This PR makes `delete` align with `delete`.

> At the very least the commit/PR should mention that in the title; or be a separate PR. It would easily be overlooked in putting together a changelog.

Yea I can do that. But fixing "delete" was not the intention of this PR. Addressing warnings was the intention.

krux02:
ready to merge. CI reruns the same tests again.

saem (approved):
Bors r+

lib/pure/strutils.nim
Clyybber: The only change in logic here (because the rest is De Morgan's law) is that instead of checking `slice.a >= s.len` (`not slice.a < s.len` in the pre-PR code) now `slice.a > s.len` is checked.
So this PR makes `0..<0` or `0..<-1` valid for a delete operation on an empty string. On non-empty strings they were already valid. As a side-effect this also makes `4..<4` valid on a string with length 4.
`4..<5` will still raise though even though it's an empty slice too (preexisting behaviour tho).
Just thought I'd put this clarification here

PR #226, opened 2022-02-03
test and generate release binaries

alaviss:
## Summary

This change facilitates the generation of source archive as well as
release archives in nimskull's CI. By integrating with main CI, we
test the binaries that will be released directly so they are
guaranteed to work.

## Details

There are several changes packed in this:

- CI now build release compiler and tools, the same kind in a release
  archive. This is to make sure that we test the exact binaries that we
  ship.

  - To help with this, new subcommand `all` is added to `koch` which
    builds binaries the same way as a release build done by
    `unixrelease` would.

- Source archive and release binaries will be packaged and uploaded to
  artifacts. This allow for testing of these if required and will be
  picked up and published via the continous delivery pipeline.

- A new test has been added: Verifying that the same release binaries
  can be built from git and source archive. This test is done via
  diffoscope, so in the event that it doesn't pass, a comprehensive
  report will be uploaded to help trace the problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,7 @@ jobs:
     outputs:
       skip: ${{ steps.skip_result.outputs.result }}
       target_matrix: ${{ steps.matrix.outputs.result }}
+      shared_builder: ${{ steps.matrix.outputs.shared }}
 
     steps:
       - id: run_cond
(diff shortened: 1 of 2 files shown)


saem (approved):
bors r+

saem:
bors r+

(what happened buddy?)

saem:
bors r+

(come on buddy_

PR #227, opened 2022-02-06
debug: improve tree repr prints for `TType`

haxscramper:
- add global configuration object that can be used to provide default
  values for `treeRepr` - now semantic tracer and `debugAst` (used in
  gdb) can be configured as well.
- Move print configuration into `TReprConf` object instead of passing
  four different fields to `treeRepr` for PNode
- Print `n` and `sym` fields for `TType`
- Print `align` field, hide `.size` by default
- Move styling configurations into `const style`, easier to make small
  tweaks in the formatting.
- Add link to `astrepr` in the compiler debugging documentation
- Make `itemId` print more explicit for `PType`
- Show full owner chain instead of the single parent one
- Explicitly print `sons:` field for TType, fix missing newline bug
- Annotate `debug*` procedures with deprecated message

### DIFF
--- a/compiler/utils/astrepr.nim
+++ b/compiler/utils/astrepr.nim
@@ -1,4 +1,23 @@
-## `treeRepr()` implementation for the `PNode` tree structures
+## `treeRepr()` implementation for the `PNode` tree structures. This module
+## provides helper procedures that can be used for debugging purposes -
+## dumping compiler IR in the textual form.
+##
+## For each IR type (``PType``, ``PSym``, ``PNode``) there are several
+## procs defined:
+#
+## - `treeRepr` - main implementation to generate the `ColText` chunk for the
+##   object.
+## - `debug` - convenience ovelroads that print generated text immediately.
+##   There are two version of the `debug` - one that accept `ConfigRef` object,
+##   and one that can work without it. Both call `treeRepr` internally, but
+##   with `ConfigRef` present more information can be printed.
+## - `debugAst`, `debugSym`, `debugType` - procs added for use in `gdb` debugging.
+##   They have `{.exportc.}` annotations, so can be used directly by name there
+##
+## `treeRepr` and all other procedures in the module accept `TReprConf`
+## object (default value is provided via `implicitTReprConf` global, for
+## gdb procs it is used implicitly) that controls which fields specifically
+## will be printed.
 
 import
   ast/[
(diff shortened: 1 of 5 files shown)


compiler/utils/astrepr.nim
```diff
@@ -1,4 +1,23 @@
-## `treeRepr()` implementation for the `PNode` tree structures
+## `treeRepr()` implementation for the `PNode` tree structures. This module
```
saem, line 1: This is awesome.

compiler/utils/astrepr.nim
```diff
@@ -16,63 +35,120 @@ import
...
-  trfShowNodeLineInfo,
-  trfShowSymLineInfo,
-  trfShowSymOptions,
-}
+const treeReprAllFields* = { trfShowSymFlags .. trfShowNodeTypes } ## Set
+                           ## of flags to print all fields in all tree
+                           ## reprs
+
+const style = (
+  kind: fgBlue,
```
saem, line 108: Is there a particular approach or set of principles behind the colours chosen?

compiler/utils/astrepr.nim
```diff
@@ -16,63 +35,120 @@ import
...
+  strLit: fgYellow,
+  floatLit: fgMagenta,
+  dim: termFg(15),
+  errKind: termFg(5, 2, 0),
+  err: fgRed,
+  comment: termFg(4, 2, 1),
+  owner: termFg(1, 3, 0)
+)
+
+const defaultTReprConf* = TReprConf(
```
saem, line 122:
```suggestion
# tip: see `compactReprConf` below for an example of how to derive configs
const defaultTReprConf* = TReprConf(
```

compiler/utils/astrepr.nim
```diff
@@ -16,63 +35,120 @@ import
...
+    trfShowTypeBaseId ## `module` and `item` from `itemId` field
+
+    trfShowNodeLineInfo ## Node location information
+    trfShowNodeFlags ## `.flags`
+    trfShowNodeIds ## `.id` field, available when compiler is built with `-d:nodeIds`
+    trfShowNodeComments ## `.comment` field
+    trfShowNodeErrors ## Embedded `nkError` reports
+    trfShowNodeTypes ## Flat render of the node type
+
+  TReprConf* = object
```
saem, line 85: Directional/long term comment: The 'T' prefix isn't something I want to continue, Hungarian notation is dead, doesn't have to be this PR, but I'd like to work towards phasing it out.

compiler/utils/astrepr.nim
```diff
@@ -16,63 +35,120 @@ import
...
+  block:
+    var base = defaultTReprConf
+    base.flags.incl trfPackedFields
+    base
+  ## Compacted tree repr configuration
+
+var implicitTReprConf*: TReprConf = defaultTReprConf ## global
+  ## configuration object that is implicitly used by `debugAst` and
+  ## `debugType`. Can be used in order to configure behaviour of the
+  ## debugging functions that could later be called from `gdb` environment
```
saem, line 149: This is so on the money, the Python gdb and lldb scripts have to do too much, we really just need to compile in the procs and those scripts should be very simple shims.

compiler/utils/astrepr.nim
```diff
@@ -82,18 +158,20 @@ const IntTypes = {
 
 const CompressedBuiltinTypes = {tyBool, tyChar, tyPointer, tyString} + IntTypes
 
+func contains(rconf: TReprConf, flag: TReprFlag): bool = flag in rconf.flags
```
saem, line 161: 'has' if you want a shorter term.

compiler/utils/astrepr.nim
```diff
@@ -122,124 +200,175 @@ proc format[I](s: set[I], sub: int): string =
 
   result.add "}"
 
-const setStyle = termFg(2, 1, 3)
+func format(i: SomeInteger): ColText = $i + style.number
+
+proc ownerChain(sym: PSym): string =
+  var own: seq[PSym] = @[sym]
```
saem, line 206:
```suggestion
  var chain: seq[PSym] = @[sym]
```

compiler/utils/astrepr.nim
```diff
@@ -122,124 +200,175 @@ proc format[I](s: set[I], sub: int): string =
...
+  var sym = sym
```
saem, line 207: 'curr', for current might be better.

I think there might be an iterator for this already, I think it's hiding in AST or maybe module graph

compiler/utils/astrepr.nim
```diff
@@ -122,124 +200,175 @@ proc format[I](s: set[I], sub: int): string =
...
+  while not sym.owner.isNil():
+    sym = sym.owner
+    own.add sym
 
-func format(i: SomeInteger): ColText = $i + fgCyan
+  for idx in countdown(own.high, 0):
+    if idx < own.high:
+      result.add "."
+
+    result.add own[idx].name.s
```
saem, line 216: Mildly faster and sightly more readable would be a string format, with '%'. I care about readability more, of course.

compiler/utils/astrepr.nim
```diff
@@ -122,124 +200,175 @@ proc format[I](s: set[I], sub: int): string =
...
   ): ColText =
 
   coloredResult(1)
   var res = addr result
-  genFields(res[], indent, flags)
+  genFields(res[], indent, rconf)
 
-  if trfShowSymFlags in flags and sym.flags.len > 0:
-    field("flags", format(sym.flags, 2) + setStyle)
+  if trfShowSymFlags in rconf and
```
saem, line 234: Don't have to do this now, but could leave a small refactor comment for others to follow up on, something like:

1. Create a const with sym related trfReprConf set, if it doesn't exist
2. Change if/Elif to a case statement on the kind
3. Embed the remaining conditionals within each statement branch
4. In the else guard to make sure no kind in the const is missed in prior conditions (other side of exhaustiveness)

compiler/utils/astrepr.nim
```diff
@@ -368,76 +493,77 @@ proc treeRepr*(
...
+          field("err", substr($report.kind, 4) + style.errKind)
+          hfield("errid", $n.reportId.int + style.err)
 
         let (file, line, col) = n.compilerInstInfo()
-        field("info", formatPath(conf, file) + fgBlue)
-        add "(", $line + fgCyan
-        add ".", $col + fgCyan
+        field("info", formatPath(conf, file) + style.ident)
+        add "(", $line + style.number
+        add ".", $col + style.number
```
saem, line 566: Could we use commas, because I think dots are for tree navigation while commas are for text coordinates, also means from an IDE (thinking vscode extension) this is more expected so it's less churn for now?

All ears for counter arguments.

If it's too much for this PR we can at least agree and leave a note for the future.

compiler/utils/astrepr.nim
```diff
@@ -122,124 +200,175 @@ proc format[I](s: set[I], sub: int): string =
...
 
-  if not isNil(conf) and trfShowSymLineInfo in flags and sym.info != unknownLineInfo:
-    field("info", conf.toMsgFilename(sym.info.fileIndex) + fgBlue)
-    add "(", $sym.info.line + fgCyan
-    add ".", $sym.info.col + fgCyan
+  if not isNil(conf) and
+     trfShowSymLineInfo in rconf and
+     (sym.info != unknownLineInfo or rconf.defaulted()):
+    field("info", conf.toMsgFilename(sym.info.fileIndex) + style.ident)
+    add "(", $sym.info.line + style.number
```
saem, line 261: Seeing similar patterns, maybe not enough occurrences, but would it make sense to create a string format template or two to absorb this?

No worries if not.

compiler/utils/astrepr.nim
```diff
@@ -16,63 +35,120 @@ import
...
-} + treeReprAllFields - {
-  trfShowNodeLineInfo,
-  trfShowSymLineInfo,
-  trfShowSymOptions,
-}
+const treeReprAllFields* = { trfShowSymFlags .. trfShowNodeTypes } ## Set
+                           ## of flags to print all fields in all tree
+                           ## reprs
+
+const style = (
```
saem, line 107: Big fan of the named styles approach! 🥳

compiler/utils/astrepr.nim
```diff
@@ -248,115 +377,111 @@ proc treeRepr*(
...
+      for idxIdx, pos in idx:
+        if idx.len - rconf.maxPath <= idxIdx:
+          indent += tern(first, 0, 1)
 
-            case pos:
-              of 0 .. 9: indent += 1
-              of 10 .. 99: indent += 2
-              else: indent += 3
+          case pos:
+            of 0 .. 9: indent += 1
```
saem, line 406: LoL, I tried to think of descriptors for these ranges, before giving up (don't think we need them) here is where I got to:

- we still got room
- little bit of spill over, this is fine
- look, just lock the door and no one will ever use that room again, it doesn't exist

compiler/utils/astrepr.nim
```diff
@@ -449,27 +575,31 @@ proc treeRepr*(
...
+        if newIdx > rconf.maxLen:
           break
 
         if newIdx < n.len - 1:
           add "\n"
 
   aux(pnode, @[])
 
-proc debugAst*(it: PNode) {.exportc.} =
+{.pragma: dbg, deprecated: "Debug proc, should not be used in the final build".}
```
saem, line 600: Thank you!

compiler/utils/tree_repr_doc.rst
```diff
@@ -2,28 +2,21 @@ Generate tree representation from `PNode`.
...
-- `flags`: Set of the formatting options for the procedure. Several
-  built-in constants are defined in this module that might be suitable for
-  different debugging or AST exploration tasks.
-- `maxDepth`: Ignore all nodes that are placed deeper than that. Useful to
-  see a high-level overview for large nodes, such as full proc bodies
-- `maxLength`: on each level, print upto that many subnodes. Useful to cut
-  out parts on large `case` trees, toplevel nodes for modules, statement
-  lists and such.
-- `maxPath`: maximum path length for subnodes
+- ``rconf``: print configuration - all nested calls to ``treeRepr`` also used
```
saem, line 10:
```suggestion
- ``rconf``: print configuration - all nested calls to ``treeRepr`` also use
```

saem (commented):
LoL, has some review comments, they're not super blockers.

compiler/utils/astrepr.nim
haxscramper: I wanted 'in' to work

compiler/utils/astrepr.nim
haxscramper: Close to general syntax highlighting colors in most ide/editor themes

compiler/utils/astrepr.nim
haxscramper: It stands for Tree Repr Config here

compiler/utils/astrepr.nim
haxscramper: I don't understand, what do you want to change here. I have a series of conditionals, each check whether the field is enabled, and if so, they add it. Wrapping this all in case statement will only needlessly increase the nesting.

compiler/utils/astrepr.nim
haxscramper: this was a typo

PR #229, opened 2022-02-09
verify and publish binaries generated by CI

alaviss:
## Summary

This is the final portion of continous delivery pipeline which creates a
release based on artifacts created during a CI run done as part of a
merge via bors.

The release created will contain a JSON manifest and release binaries.
The JSON manifest follows this schema:

```json
{
  "manifestVersion": 0,
  "version": "version identifier (ie. 0.1.0-dev.20000)",
  "source": {
    "name": "source-archive-name",
    "sha256": "checksum"
  },
  "binaries": [
    {
      "target": "the target triplet (eg. x86_64-linux-gnu)",
      "name": "binary-archive-name",
      "sha256": "checksum"
    }
  ]
}
```

This manifest allow tools to easily pick out the artifact that they need
depending on the target operating system, and it also prevent the need
to hard-code a specific file name format in such tools.

## Details

This PR comes with:

- A release manifest manipulation tool: This tool is used to create a
  release manifest from archive manifests generated by niminst. At the
  moment it does not have many features other than the ones required by
  the pipeline. Refer to the tool built-in help for details on how to
  use it.

- An additional test in main CI to verify that the release manifest tool
  can create a release manifest from binaries generated there.

- Extra steps in publisher that download and create a release out of
  artifacts created by main CI.

---

## Notes for Reviewers
- Currently it does not support automated creation of a non-development release. This will come later.

- The release description is very plain at the moment. Suggestions on what to put in them would be appreciated.

- This pipeline will create a release on the main repository, which makes it a bit scary as tags are not removable (they can, but should never be done).

- A test release created by this pipeline can be found here: https://github.com/alaviss/nimskull/releases/tag/0.1.0-dev.20056

### DIFF
--- a/tools/release_manifest.nim
+++ b/tools/release_manifest.nim
@@ -1,0 +1,576 @@
+#
+#     Copyright (c) 2022 Leorize <leorize+oss@disroot.org>
+#
+#        See the file "copying.txt", included in this
+#        distribution, for details about the copyright.
+#
+
+## A tool to build and manage a release manifest for the compiler.
+##
+## Designed for use in CI, thus not having many interactive components.
+
+import std/[json, options, os, parseopt, strutils, tables]
+
+type
+  Database = object
+    ## The artifact database
+    file: seq[string] ## The artifacts, which are unique
+    sha256: seq[string] ## The SHA256 checksum of the artifacts
+    triplet: seq[string] ## The target triplet of the artifacts
+    version: string ## The version of all artifacts
+
+  ArtifactDataV0 = object
+    ## Data about an artifact in the manifest
+    name: string ## The file name of the artifact
+    sha256: string ## The SHA256 checksum of the artifact
+
+  BinaryArtifactDataV0 = object
+    ## Data about a binary artifact in the manifest
+    target: string ## The target of the artifact
+    name: string ## The file name of the artifact
+    sha256: string ## The SHA256 checksum of the artifact
+
+  ReleaseManifestV0 = object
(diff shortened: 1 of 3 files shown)


tools/release_manifest.nim
```diff
@@ -0,0 +1,564 @@
...
+  ## :file:
+  ##   The file to be added
+  ##
+  ## :version:
+  ##   The version of the artifact
+  ##
+  ## :triplet:
+  ##   The triplet it is targeting
+  if file.len == 0:
+    raise newException(ValueError):
```
saem, line 68: Future directional thought: We really should bring result/try into stdlib.

For maximum performance DOD land, the database should probably just have error storage that needs to be resolved and cleared prior to using it further -- not recommend here.

tools/release_manifest.nim
```diff
@@ -0,0 +1,564 @@
...
+    )
+
+  # Add all artifacts
+  for artifact in manifest.binaries.items:
+    result.addArtifact(
+      artifact.name, artifact.sha256, manifest.version, artifact.target
+    )
+
+func toTriplet(os, cpu: string): string =
+  ## Convert Nim's os/cpu pair into a LLVM-style target triplet.
```
saem, line 157: Did you have any go to reference links you used for triplet study when devising this?

If so please link then in comments for those modifying this in the future.

saem (commented):
For the release description I think the only improvement would be a change log, commit message titles perhaps?

saem (approved):
bors r+

saem:
We gunna get nightlies! 🎉

PR #230, opened 2022-02-09
Fix invalid push pragmas creating cyclic AST

Clyybber:
## Summary
The added test would previously produce cyclic AST
which would then trigger asserts. With this fix it
produces the expected error message.

Also cleaned up ast.copyNode to reuse copyNodeImpl.

## Details
It fixes it by checking if the node returned by prepareSinglePragma (which in turn invokes processPush) in the case of an error wraps the whole pragma node, and not just one of it's childs and handles it approriately (not assigning the child to it's parent)/

---

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -806,25 +806,29 @@ proc delSon*(father: PNode, idx: int) =
   for i in idx..<father.len - 1: father[i] = father[i + 1]
   father.sons.setLen(father.len - 1)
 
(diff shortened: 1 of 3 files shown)


saem (approved):
Bors r+

I knew there was a cycle somewhere, just couldn't find it when I last went through the code, glad to have it found and fixed.

PR #232, opened 2022-02-14
astrepr: location filtering, custom data support

haxscramper:
- Move line info formatting logic to a separate helper proc
- Allow symbol/type name/id to be shown
- Add callbacks that can be used to provide additional information
  for nodes/symbols/types printed. With planned move to the
  data-oriented AST it is likely that not *all* data will be
  readily available from the `P(Node|Sym|Type)` itself.
- Minor changes to the default formatting options to reduce the
  verbosity.
- In case of the cyclic AST node numbers are now put on the leftmost
  column instead of being mixed in with the rest of the identifiers.
  This makes it easier to figure out what node "cyclic" annotation
  refers. Also when pasting things to the matrix it is easier to
  explain what you are talking about - (aka "see node number 3").
- Overloads for set operations on the `TReprConf` - include and
  exclude flags just like you do with regular sets.
- Add 'verbose' trepr config - print every possible field
- show 'unknown' location data in the info as well.
- Optionally show symbol AST - was missing from dump
- Check if node is `inFile` or `inLines`
- Support setting implicit config ref as well - otherwise
  `debugSym` in the gdb is unable to print node file paths
  properly.

### DIFF
--- a/compiler/utils/astrepr.nim
+++ b/compiler/utils/astrepr.nim
@@ -1,23 +1,46 @@
-## `treeRepr()` implementation for the `PNode` tree structures. This module
-## provides helper procedures that can be used for debugging purposes -
-## dumping compiler IR in the textual form.
-##
-## For each IR type (``PType``, ``PSym``, ``PNode``) there are several
-## procs defined:
-#
-## - `treeRepr` - main implementation to generate the `ColText` chunk for the
-##   object.
-## - `debug` - convenience ovelroads that print generated text immediately.
-##   There are two version of the `debug` - one that accept `ConfigRef` object,
-##   and one that can work without it. Both call `treeRepr` internally, but
-##   with `ConfigRef` present more information can be printed.
-## - `debugAst`, `debugSym`, `debugType` - procs added for use in `gdb` debugging.
-##   They have `{.exportc.}` annotations, so can be used directly by name there
-##
-## `treeRepr` and all other procedures in the module accept `TReprConf`
-## object (default value is provided via `implicitTReprConf` global, for
-## gdb procs it is used implicitly) that controls which fields specifically
-## will be printed.
+##[
+
+`treeRepr()` implementation for the `PNode` tree structures. This module
+provides helper procedures that can be used for debugging purposes -
+dumping compiler IR in the textual form.
+
+For each IR type (``PType``, ``PSym``, ``PNode``) there are several procs
+defined:
+
+- `treeRepr` - main implementation to generate the `ColText` chunk for the
+  object.
+
+- `debug` - convenience ovelroads that print generated text immediately.
+  There are two version of the `debug` - one that accept `ConfigRef`
+  object, and one that can work without it. Both call `treeRepr`
+  internally, but with `ConfigRef` present more information can be printed.
+
+- `debugAst`, `debugSym`, `debugType` - procs added for use in `gdb`
+  debugging. They have `{.exportc.}` annotations, so can be used directly
+  by name there
+
+`treeRepr` and all other procedures in the module accept `TReprConf` object
+(default value is provided via `implicitTReprConf` global, for gdb procs it
+is used implicitly) that controls which fields specifically will be
+printed.
+
+Configuration object also includes callbacks that would allow you to
+provide additional information for the printed targets. For example, you
+want to perform some analysis, and don't have `ConfigRef` object passed
+around (because your code *only* collects some data from the nodes). But
+then you need to figure out the actual location of where node comes from.
+In that case you simply assign to `implicitTReprConf.extraNodeInfo`, and
+next `debug` you call will provide needed information.
+
+.. code::nim
+
+  implicitTReprConf.extraNodeInfo = proc(node: PNode): ColText =
+    result.add "node location " & (config $ node.info)
+
+
+
+]##
+
 
 import
   ast/[
@@ -35,13 +58,17 @@ import
   std/[
     tables,
     strutils,
-    strformat
+    strformat,
+    math
   ]
 
 import std/options as std_options
 
+export colortext
+
 type
   TReprFlag* = enum
+    ## Configuration options for treeRepr debugging
     trfReportInfo ## Show information about embedded semantic report errors
     trfPackedFields ## Put fields horizontally instead of arranging them
                     ## out veritcally. Should be used only when a small
@@ -56,6 +83,7 @@ type
 
     trfShowSymFlags ## Show symbol `.flags` field
     trfShowSymLineInfo ## Line information
+    trfShowSymName
     trfShowSymTypes ## Flat render of the symbol type
     trfShowSymMagic ## used symbol magic
     trfShowSymKind ## `.kind`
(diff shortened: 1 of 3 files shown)


compiler/utils/astrepr.nim
```diff
@@ -203,34 +265,75 @@ proc format[I](s: set[I], sub: int): string =
...
+    result.add substr($chain[idx].kind, 2)
     result.add ")"
 
 func defaulted(r: TReprConf): bool = trfShowDefaultedFields in r
 
+proc infoField(
+    conf: ConfigRef, info: TLineInfo, rconf: TReprConf,
+    indent: int, name: string): ColText =
+  var res = addr result
+  genFields(res[], indent, rconf)
```
saem, line 290: I might be misunderstanding things, but `genFields` is a template, so the `addr result` shouldn't be necessary correct?

I'm assuming that's for the whole var param/value bit.

compiler/utils/astrepr.nim
```diff
@@ -203,34 +265,75 @@ proc format[I](s: set[I], sub: int): string =
...
   ): ColText =
 
   coloredResult(1)
   var res = addr result
   genFields(res[], indent, rconf)
 
+
+  if not rconf.extraSymInfo.isNil():
+    let text = rconf.extraSymInfo(sym)
+    if 0 < len(text):
```
saem, line 330: Not a blocker for this PR, but I find the inverted checks such as this awkward to read and spend a fair bit of time double checking the logic. Going forward I think we should stick with `if text.len > 0:`.

I thought I'd get use to it, but since rebasing the sexp changes into the testament PR (boulder?) I'm still slowing down a lot. In terms of semantics we're saying "the string text is of length greater than 0", or "the string text is not empty".

In both cases, the subject is `text` and not `0` and it really doesn't flow.

compiler/utils/astrepr.nim
```diff
@@ -374,10 +487,29 @@ proc treeRepr*(
 
 
 
+proc cyclicTreeAux(n: PNode, visited: var seq[PNode], count: var int): bool =
+  if n == nil: return
```
saem: Doesn't have to change this PR -- stylistic.

Earlier there was `if not rconf.extraTypInfo.isNil():`, which would read more linearly (don't have to remember to apply the `not`) if it was written as: `if rconf.extraTypInfo != nil:`.

Here taking what I think you were leveraging above (but the `not` makes it awkward), `if n == nil: return` would be better as `if n.isNil: return`.

compiler/utils/astrepr.nim
```diff
@@ -374,10 +487,29 @@ proc treeRepr*(
...
+  for v in visited:
+    if v == n:
+      return true
+
+  inc count
+  if not (n.kind in {nkEmpty..nkNilLit}):
```
saem:
```suggestion
  if n.kind notIn {nkEmpty..nkNilLit}:
```

similar to before

compiler/utils/astrepr.nim
```diff
@@ -609,41 +765,63 @@ proc debugAst*(it: PNode) {.exportc, dbg.} =
...
+proc inFile*(
+    node: PNode | PSym,
+    file: string,
+    lrange: Slice[int] = low(int) .. high(int)
+  ): bool {.dbg.} =
+  ## Check if file name of the `info` has `filename` in it. For debugging
+  ## purposes as it is pretty slow. Should be used like this:
+  ##
+  ## ..code-block::nim
+  ##   if `??`(conf, node.info, "filename"):
```
saem:
```suggestion
  ##   if inFile(nodeOrSym, "filename"):
```

doc/debug.rst
```diff
@@ -14,9 +15,9 @@ How to debug different subsystems of the compiler using built-in tooling.
 Special defines
 ---------------
 
-Debugging functionality can usually be accessed only when compiler itself
-has been built with special defines - this is made to avoid runtime
-overhead, because some debugging tools are not exactly cheap to run.
+Debugging functionality is available only when compiler itself built with
```
```suggestion
Debugging functionality is available only when compiler itself is built with
```

doc/debug.rst
```diff
@@ -50,12 +51,11 @@ manner. For more documentation see the module itself.
...
-trace how semantic analysis is performed on this block of code. Each
-procedure that is annotated with one of the `addInNimDebugUtils()`
-overloads is going to be traced in the execution tree. For example -
-`sem.semOverloadedCall`, triggered when it is necessary to analyze call to
-overloaded procedure.
+With `nimDebugUtils` you can annotate section of the code to trace how
+semantic analysis works on this block of code. Each procedure that uses
+with one of the `addInNimDebugUtils()` overloads is going appear in the
+execution tree. For example - `sem.semOverloadedCall`, triggered when it is
+necessary to analyze call to overloaded procedure.
```
```suggestion
With `nimDebugUtils` defined, you can then enable tracing for compiler semantic
analysis over sections of code using `nimCompilerDebug` -- code example below.
Not all compiler routines are enabled with first class tracing, if tracing is
missing for a particular compiler procedure then add a call to one of the
`addInNimDebugUtils()` overloads in the compiler procedure definition. An
example of a compiler procedure with tracing enabled is `sem.semOverloadedCall`.
This procedure will appear in any trace where the code being analysed includes a
call to an overloaded procedure.

See below for an example of how to add tracing to a procedure, followed by an
example of how to enable tracing for a section of code being compiled.
```

doc/debug.rst
```diff
@@ -115,19 +114,19 @@ It has a lot of information, but general parts for each call parts are:
...
-separate files in this directory. This is extremely helpful if you want to
-track bugs where you can clearly find "this works" and "this doesn't work"
-versions. This debugging technique is referred to differential debugging or
-sometimes differential diagnosis.
+If test compiler runs with `-d:nimCompilerDebugTraceDir=/some/dir` option
+the reports stored between different sections are written into separate
+files in this directory. This is helpful if you want to track bugs where
+you can get find "this works" and "this doesn't work" versions. This
+debugging technique called differential debugging or sometimes differential
+diagnosis.
```
```suggestion
you have two pieces or versions of code where one works and the other doesn't.
Collect traces for both and compare them to see how the compiler analysis
differs between them and use that to guide your investigation. This technique is
called differential debugging or sometimes differential diagnosis.
```

saem (changes requested):
Some changes for docs; the rest are non-blocking (sytlistic, so we can discuss in chat or something).

The `inFile` is so much clearer. srsly, `??`... wut. The `postLiteral` takes away a good amount of noise, too.

saem:
BTW, feel free to kick-off bors after the docs changes.

It's easier to debug already.

compiler/utils/astrepr.nim
haxscramper: I put all the comparisons in code in one direction - `a < b`, `a < b && b < c` etc.

saem: I know, I've tried to get used to it but that form of consistency doesn't communicate as clearly as I tried to point out.

Not a blocker for this PR. I'm just asking to start adjusting the style going forward in other PRs.

compiler/utils/astrepr.nim
saem: Stylistic, not a blocker

PR #234, opened 2022-02-16
Finalize 'active' configuration subset split

haxscramper:
Due to how this thing is implemented, it is hardly possible to separate the mutation part from everything else. So I ended up just writing more documentation and some tests for these parts.

### DIFF
--- a/tests/compilerunits/confread/tcli_parsing.nim
+++ b/tests/compilerunits/confread/tcli_parsing.nim
@@ -1,0 +1,81 @@
+discard """
+description: ""
+joinable: false
+"""
+
+import
+  ast/[
+    reports
+  ],
+  front/[
+    options,
+    commands
+  ],
+  std/[
+    unittest,
+    strtabs,
+    strutils,
+    os,
+    algorithm,
+    sequtils
+  ]
+
+
+var reported: seq[Report]
+
+proc hook(conf: ConfigRef, report: Report): TErrorHandling =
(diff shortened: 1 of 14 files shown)


tests/compilerunits/confread/tcli_parsing.nim
```diff
@@ -0,0 +1,81 @@
...
+
+  test "Warning configuration":
+    let conf = parse("--warning=all:off --warning=UnusedImport:on").conf
+    check(conf.noteSets[cnCurrent] * repWarningKinds == {rsemUnusedImport})
+
+# import hmisc/other/hpprint
+
+let csd = currentSourcePath().parentDir()
+
+suite "Path options specification":
```
saem, line 68: Anecdote/Tangent:

This sort of testing is an important step forward, pretty early when I started working on `nimsuggest` I was told about `findNimProjectFile` or some proc like that. Anyhow, the intention was detecting a project file so the vscode extension can infer things in a better way than having to specify a config.

Long story short, `findNimProjectFile` is busted and the whole thing is buggy as well. It really needs a spec and this test case shows a pretty good way of how I can now enumerate that.

So at the time I didn't have a great way of testing it and no one was a fan of adding these sort of tests to the compiler so I ended up dropping it. This PR reminded me of that, so now at least there is a way forward. 🎉

saem (approved):
bors r+

PR #237, opened 2022-02-22
Enable debugging unreported errors

Clyybber:
## Summary
Enabled with -d:nimDebugUnreportedErrors
Doesn't keep track of where the nodes were created yet.
Also fix semIf not handling nkError nodes correctly.
Also cleanup buildErrorList/walkErrors roundabout way
of doing post-visit DFS.
Also improve rsemTypeKindMismatch error message.

### DIFF
--- a/compiler/ast/errorhandling.nim
+++ b/compiler/ast/errorhandling.nim
@@ -43,6 +43,9 @@ import
   front/[
     msgs,
     options
+  ],
+  std/[
+    tables
   ]
 
 proc errorSubNode*(n: PNode): PNode =
@@ -101,6 +104,10 @@ proc newError*(
 
   for a in args:
     result.add #[ 3+ ]# a
+
+  when defined(nimDebugUnreportedErrors):
+    if errorKind != rsemWrappedError:
+      conf.unreportedErrors[result.reportId] = result
 
 proc newError*(
     conf: ConfigRef,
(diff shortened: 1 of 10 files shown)


haxscramper:
Please update `debug.rst` documentation accordingly, so people could know compiler reacts to `nimDebugUnreportedErrors` define. Example the "unreported" error that is printed would also be nice (along the lines of other examples I made for sem tracer and debug definitions)

compiler/ast/errorhandling.nim
```diff
@@ -171,10 +175,10 @@ proc buildErrorList(config: ConfigRef, n: PNode, errs: var seq[PNode]) =
   of nkEmpty .. nkNilLit:
     discard
   of nkError:
-    errs.add n
     buildErrorList(config, n[wrongNodePos], errs)
+    errs.add n
   else:
-    for i in countdown(n.len - 1, 0):
+    for i in 0..<n.len:
```
haxscramper, line 185: I personally would put explanation comment about why exactly iteration is done this way and not the other. Of course this is not a blocker, but given this is one of the most important procs that defines the whole logic behind error walker, it would still be nice to see more explanations.

haxscramper:
Aside from required documentation fixes, everything else looks good to me.

Clyybber:
> Example the "unreported" error that is printed would also be nice (along the lines of other examples I made for sem tracer and debug definitions)

The example would have to demonstrate a compiler bug as unreported errors only manifest due to those, and then I'd probably rather want to fix the demonstrated bug :)

PR #252, opened 2022-03-13
Replace `PCtx` parameters with `TCtx` in VM code

zerbina:
* Where `TCtx` is now used, also use `func` instead of `proc`, if applicable

Not all usages of `PCtx` are replaced. `PCtx` is still used in procedures
that form the border between non-VM/VM code. `vmops` also still uses
 `PCtx`, as some ops need to capture a reference to the context

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -72,7 +72,7 @@ const
   errIllegalConvFromXtoY = "illegal conversion from '$1' to '$2'"
 
 proc stackTraceImpl(
-    c:          PCtx,
+    c:          TCtx,
     sframe:     PStackFrame,
     pc:         int,
     lineInfo:   TLineInfo,
@@ -80,7 +80,7 @@ proc stackTraceImpl(
     recursionLimit: int = 100
   ) =
 
-  proc aux(sframe: PStackFrame, pc, depth: int, res: var SemReport) =
+  proc aux(c: TCtx, sframe: PStackFrame, pc, depth: int, res: var SemReport) =
     if sframe != nil:
       if recursionLimit < depth:
         var calls = 0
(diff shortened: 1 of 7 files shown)


compiler/vm/vm.nim
```diff
@@ -540,7 +540,7 @@ proc recSetFlagIsRef(arg: PNode) =
   for i in 0..<arg.safeLen:
     arg[i].recSetFlagIsRef
 
-proc setLenSeq(c: PCtx; node: PNode; newLen: int; info: TLineInfo) =
+func setLenSeq(c: TCtx; node: PNode; newLen: int; info: TLineInfo) =
```
alaviss: The compiler has spoken :P

```suggestion
proc setLenSeq(c: TCtx; node: PNode; newLen: int; info: TLineInfo) =
```

compiler/vm/vmdef.nim
```diff
@@ -189,17 +189,22 @@ proc newCtx*(module: PSym; cache: IdentCache; g: ModuleGraph; idgen: IdGenerator
...
   c.prc = PProc(blocks: @[])
   c.loopIterations = c.config.maxLoopIterationsVM
   c.idgen = idgen
 
-proc registerCallback*(c: PCtx; name: string; callback: VmCallback): int {.discardable.} =
+proc registerCallback*(c: var TCtx; name: string; callback: VmCallback): int {.discardable.} =
   result = c.callbacks.len
   c.callbacks.add((name, callback))
 
+# For a transition period, keep an overload of `registerCallback` that still
```
saem: Maybe mark the template as deprecated with the current comment as the deprecated warning? Either that or make it a doc comment instead of a comment.

saem (changes requested):
Minor item, the other thing I noticed is that some functions have effects, will need to change those to procs or break them up. I'd go the proc route, personally.

zerbina:
Addressed review. I used `proc` again for the functions that fail the strict side-effect test.

I also fixed the failure of the ``tcompilerapi.nim`` test. It failed due to `PCtx` still being used as arguments in ``nimeval.nim``.

Interestingly enough, in the CI, the test failure was reported to be due to the compiler crashing with a nil access. I ran the test on my machine (Windows) and it also failed with a compiler crash. But, instead of an invalid memory access, I got an assertion raised in ``semcall.nim``.

 The compiler crash only happened when importing `nimeval`. Adjusting the `PCtx` usage made the crash go away too.

saem:
I did a quick search in finding usage in semdata and semmagic, but those are at the boundary as mentioned by the PR.

saem (approved):
And so begins the revamp of the VM in devel. 🎉 

Bors r+

alaviss:
Weird, let's try again

bors r+

PR #255, opened 2022-03-18
Rework stack-frame handling in the VM

zerbina:
Instead of having the stack frames as `ref` objects that are chained
together via their `next` field, store them in a `seq` in `TCtx`. This
makes it easier to reason about them, and also connects the frames
to their owning context.

While the frame list should ideally be first-in last-out (stack), this
isn't currently possible due to how exception handling is implemented.

In addition, the stack trace now always includes the entry function
and also the number of skipped frames. A nil access error when
no entry function exists (happens when a statement is evaluated)
is also fixed.

---

**Notes for reviewers**:
* The name "tos" is still used for variables/parameters storing the stack frame index. I was not able to figure out what it's supposed to mean/stand for. If wanted, I can replace it's usages with something more descriptive

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -73,32 +73,47 @@ const
 
 proc stackTraceImpl(
     c:          TCtx,
-    sframe:     PStackFrame,
+    sframe:     StackFrameIndex,
     pc:         int,
     lineInfo:   TLineInfo,
     infoOrigin: InstantiationInfo,
     recursionLimit: int = 100
   ) =
-
-  proc aux(c: TCtx, sframe: PStackFrame, pc, depth: int, res: var SemReport) =
-    if sframe != nil:
-      if recursionLimit < depth:
-        var calls = 0
-        var sframe = sframe
-        while sframe != nil:
-          inc calls
-          sframe = sframe.next
-
-        return
-
-      aux(c, sframe.next, sframe.comesFrom, depth + 1, res)
-      res.stacktrace.add((sym: sframe.prc, location: c.debug[pc]))
+  ## Generate and report the stack trace starting at frame `sframe` (inclusive).
+  ## The generated trace length is capped at `recursionLimit`, but always includes
+  ## the entry function
+  assert c.sframes.len > 0
 
   var res = SemReport(kind: rsemVmStackTrace)
   res.currentExceptionA = c.currentExceptionA
   res.currentExceptionB = c.currentExceptionB
 
-  aux(c, sframe, pc, 0, res)
+  block:
+    var i = sframe
+    var count = 0
+    var pc = pc
+    # Traverse the stack formed via the `next` field
+    while i >= 0:
+      let f = c.sframes[i]
+      i = f.next
+
+      if count < recursionLimit - 1 or i == 0:
+        # The `i == 0` is to make sure that we're always including the
+        # bottom of the stack (the entry function) in the trace
+
+        # Since we're walking the stack from top to bottom,
+        # the elements are added to the trace in reverse order 
+        # (the most recent function is first in the list, not last).
+        # This needs to be accounted for by the actual reporting logic
+        res.stacktrace.add((sym: f.prc, location: c.debug[pc]))
+
+      pc = f.comesFrom
+      inc count
+
+    if count > recursionLimit:
+      res.skipped = count - recursionLimit
+
+  assert res.stacktrace.len() <= recursionLimit # post condition check
 
   let action = if c.mode == emRepl: doRaise else: doNothing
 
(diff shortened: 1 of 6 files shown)


compiler/vm/vm.nim
```diff
@@ -73,26 +73,44 @@ const
...
+  for i in top..bottom:
+    let f = c.sframes[i]
+    let fpc = if i < bottom: c.sframes[i+1].comesFrom else: pc
+    res.stacktrace.add((sym: f.prc, location: c.debug[fpc]))
+  ]#
+
+  proc aux(c: TCtx, sframe: StackFrameIndex, pc, depth: int, res: var SemReport) =
+    if sframe != -1:
       if recursionLimit < depth:
+        # TODO: Skipped number of entries don't get reported? How is/was this intended to be handled?
```
zerbina: As far as I can tell, stack frames with a depth below the recursion limit get skipped here, but aren't getting reported

zerbina:
Hmm, it seems like `shallowCopy` for `seq` doesn't work as I expected it to with ORC

compiler/vm/vm.nim
```diff
@@ -572,21 +590,51 @@ template takeAddress(reg, source) =
...
-  var tos = tos
+  var tos = sframe
   # Used to keep track of where the execution is resumed.
   var savedPC = -1
-  var savedFrame: PStackFrame
+  var savedFrame: StackFrameIndex
   when defined(gcArc) or defined(gcOrc):
-    template updateRegsAlias = discard
-    template regs: untyped = tos.slots
+    # TODO: does shallowCopy for seqs still work under --gc:arc?
```
alaviss: It doesn't, unfortunately. `shallowCopy` is gone for `seq` and `strings` in new runtime. It's semantics was never defined properly, so it wasn't really missed :p

compiler/vm/vm.nim
```diff
@@ -73,26 +73,44 @@ const
...
     infoOrigin: InstantiationInfo,
     recursionLimit: int = 100
   ) =
-
-  proc aux(c: TCtx, sframe: PStackFrame, pc, depth: int, res: var SemReport) =
-    if sframe != nil:
+  ## Generate and report the stack trace starting at frame `sframe` (inclusive)
+  assert c.sframes.len > 0
+
+  # TODO: use the below once exception handling is reworked, and stack frames
```
saem: Random musing: Bonkers that it formed a tree before... it's a stack, folks. Anyhow, even in a bright cps future we'd want still a core pure stack or at least structured cactus stack -- needs more invariants in place.

compiler/vm/vm.nim
```diff
@@ -73,26 +73,44 @@ const
...
+  ## Generate and report the stack trace starting at frame `sframe` (inclusive)
+  assert c.sframes.len > 0
+
+  # TODO: use the below once exception handling is reworked, and stack frames 
+  # don't form a tree any longer
+  #[
+  let bottom = high(c.sframes)
+  let top = max(0, bottom - recursionLimit)
+  
+  # TODO: Skipped number of entries don't get reported? How is/was this intended to be handled?
```
saem: I think the might be a bug, but for a comment on the whole proc it shouldn't be a recursive implementation, it can easily be a loop:

The intention:
- print the most recent frame last
- descend the stack and do so again
- until we run out of stack or reach depth
- if any frames remain, say how many we skipped

The above results in a message such as: "skipped x frames, calls that led up to printing".

compiler/vm/vm.nim
saem: Yes, it's for pruning

saem:
I think tos stands for "top of stack"

saem (changes requested):
This looks good to me.

I do think simplifying the stack frame printing with a trivial loop and a comment saying what it's supposed to do would help a fair bit. See my review comment.

compiler/vm/vmdef.nim
```diff
@@ -152,19 +153,20 @@ type
...
-  PStackFrame* = ref TStackFrame
-  TStackFrame* {.acyclic.} = object
+  StackFrameIndex* = int
+
+  TStackFrame* = object
     prc*: PSym                 # current prc; proc that is evaluated
     slots*: seq[TFullReg]      # parameters passed to the proc + locals;
                               # parameters come first
-    next*: PStackFrame         # for stacking
+    next*: StackFrameIndex         # for stacking
```
zerbina, line 162: `next` could use `Option[StackFrameIndex]` instead of just `StackFrameIndex`. This would have the advantage of the default value meaning "no previous frame" (not the case right now), but would also require more space and make access a bit slower.

compiler/vm/vm.nim
```diff
@@ -2321,10 +2362,20 @@ proc rawExecute(c: var TCtx, start: int, tos: PStackFrame): TFullReg =
...
+  assert c.sframes.len == 0
+  c.sframes.add frame
+  result = rawExecute(c, start, 0).regToNode
+  assert c.sframes.len == 1
+  c.sframes.setLen(0)
+
+
 proc execute(c: var TCtx, start: int): PNode =
-  var tos = PStackFrame(prc: nil, comesFrom: 0, next: nil)
+  # XXX: instead of building the object first and then adding it to the list, it could
```
zerbina, line 2374: What's the preferred style here? Adding to the list first and then modifying the element might be a bit more efficient (due to no copy/move having to take place)

zerbina:
I transformed the recursive call into a loop and also implemented reporting for the number of skipped frames. There was also a possible nil access error that I fixed.

Compilation under ORC should also succeed again.

Since error reporting is the domain of @haxscramper, maybe they can weigh in on if they're okay with the reporting related changes.

haxscramper:
If you consider it necessary, you can add or remove report kinds and change how things are printed. If you think it would be better, then go for it. IIRC error messages from the VM are barely tested (if tested at all), so any improvement would be a win.

compiler/vm/vmdef.nim
saem: A sentinel index would work as well?

compiler/vm/vm.nim
saem: It's not a hot loop in my opinion.

For the thought exercise:
Since it's a seq of capacity x, then it's automatically initialized to the zero value per element for each frame. Can we of we set the len to 1 afterwards, then just set the one field to -1, should do it?

compiler/vm/vmdef.nim
saem: Hmm, the real trick would be extending optional, or at least a new data type, that allows for "in range" sentinel/nil like values.

Can do this later, but it's a pretty common thing in DOD land.

compiler/vm/vm.nim
```diff
@@ -73,32 +73,47 @@ const
...
+  ## the entry function
+  assert c.sframes.len > 0
 
   var res = SemReport(kind: rsemVmStackTrace)
   res.currentExceptionA = c.currentExceptionA
   res.currentExceptionB = c.currentExceptionB
 
-  aux(c, sframe, pc, 0, res)
+  block:
+    var i = sframe
```
saem, line 92: I prefer the var/let block syntax, but don't change it, mostly fyi

compiler/vm/vm.nim
```diff
@@ -73,32 +73,47 @@ const
...
-        var sframe = sframe
-        while sframe != nil:
-          inc calls
-          sframe = sframe.next
-
-        return
-
-      aux(c, sframe.next, sframe.comesFrom, depth + 1, res)
-      res.stacktrace.add((sym: sframe.prc, location: c.debug[pc]))
+  ## Generate and report the stack trace starting at frame `sframe` (inclusive).
```
saem, line 82: Nailed it! 🤩

saem (approved):
Bors r+

PR #256, opened 2022-03-18
Add language specification for `finally` clause

zerbina:
## Summary
This is an attempt at expanding the specification with tests for everything `finally` related. It's not in a mergeable state yet, but I thought that it might be useful already.

---

The additions to `t06_procedure_calls.nim` are more or less structured, while `t07_finally.nim` is currently more of a dumping ground for various test cases than anything else.

**To-do:**
* add more tests for interactions between `finally` and both `{.inline.}` and `{.closure.}` iterators
* add documentation
* re-organize test cases
* file issues for test cases that fail / put them into separate files with "knownIssue" markings
* there are still missing cases
* it might make sense to add some kind of mini-DSL via macros, in order to write the test cases in a way that makes them easier to comprehend and modify. They're mostly just copy-and-pasted right now.
* a macro or script that auto-generates lots of different permutations might also be a good idea

### DIFF
--- a/tests/lang/s01_basics/s08_control_flow/t07_finally.nim
+++ b/tests/lang/s01_basics/s08_control_flow/t07_finally.nim
@@ -1,0 +1,786 @@
+discard """
+  targets: "c cpp"
+"""
+
+## A finally clause is executed once control-flow leaves the `finally`'s corresponding `try` and (if present) `except clause. This 
+## includes normal, interrupted (`break`, `continue`, `yield`, `return`) and exceptional (`raise`) control flow.
+
+block:
+  var values: seq[string]
+
+  proc p() =
+    block a:
+      block b:
+        try:
+          values.add("1")
+          try:
+            values.add("2")
+            return
+          finally:
+            values.add("3")
+            break a
+        finally:
+          values.add("4")
+          break b
+        values.add("5")
+      values.add("6")
+    values.add("7")
+
+  p()
+
+  doAssert values == @[
+    "1",
+    "2",
+    "3",
+    "4",
+    "6",
+    "7",
+  ]
+
+
+block:
+  var values: seq[string]
+  block a:
+    try:
+      try:
+        values.add("1")
+        raise (ref CatchableError)()
+      finally:
+        values.add("2")
+        break a
+    except:
+      values.add("3")
+    finally:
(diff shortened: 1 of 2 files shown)


tests/lang/s01_basics/s08_control_flow/t07_finally.nim
```diff
@@ -0,0 +1,786 @@
...
+
+block:
+  iterator iter(): int {.closure.} =
+    try:
+      raise (ref CatchableError)()
+    finally:
+      yield 1
+
+  var it = iter
+  # TODO: what should happen here?
```
haxscramper, line 128: Per current behavior - `yield` should take priority, like any other 'jump' in finally.
Per recent discussion on how things *should* be implemented later on - `raise` should take priority.

haxscramper (commented):
Per discussion in `#main` - the closest we got to the "everyone sort of agrees it is the way things should be" is (mostly per @alaviss, and I think I can probably agree to it as well).


- If try body exit normally and finally exit for reason R, the block exit for reason R
- If try body exit for reason S and finally exit normally, the block exit for reason S
- If try body exit abruptly for reason S and finally exit abruptly for reason R
  - If S is an unhandled exception and R is not an unhandled exception, the block exit for reason S
  - If S and R are unhandled exceptions, R is chained to S and the block exit for reason R -- R is chained to S (S will be the new toplevel exception in chain)
  - Otherwise, the block exit for reason R and S is discarded

This would allow avoiding dropping exceptions while mostly maintaining the "finally takes priority" behavior that we have now.

**NOTE**: it is just a tl;dr to keep the [context](https://matrix.to/#/!alsWvZtOfWkQlHwcHC:envs.net/$HnYPPddJmRj2R1mEhdMKK6z2LWrW_ER_kHXM575sECM?via=libera.chat&via=matrix.org&via=envs.net) in place - the PR is perfectly fine as it is now, I think.

haxscramper:
I think I can work on this later on, after https://github.com/nim-works/nimskull/pull/476 and https://github.com/nim-works/nimskull/pull/135 are merged and general test cleanup is complete.

zerbina:
Sure, feel free to take this over. I believe the to-do list is still up-to-date.

PR #259, opened 2022-03-26
testament: no longer a general tool

saem:
Testament isn't ready to be a general test runner by any means. This
removes testament from releases and part of tools build. In addition it
drops the pattern option as categories are now a must.

Part of a broader effort to streamline testament.

Changes:
- removes testament from installer and koch tools build
- removes doc references for pattern option
- removes testament references as a tool
- unittest module no longer recommends testament
- remove stale reference to async category in readme.md

### DIFF
--- a/compiler/installer.ini
+++ b/compiler/installer.ini
@@ -66,7 +66,6 @@ Files: "doc/html"
 Files: "tools"
 Files: "tools/nim-gdb.py"
 Files: "nimpretty"
-Files: "testament"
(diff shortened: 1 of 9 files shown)


saem:
I'll rebase this onto `devel` which should reduce it to one commit, until then need to know if I've updated the requisite bits for removing testament as a general tool.

alaviss (approved):
bors r+

PR #260, opened 2022-03-28
Compiler Tracing: fixed compiler execution tracing

saem:
- enabled intermediate stack traces by default
- updated compiler internals doc to make it easier to get started
- added tracing to a few more procs, noteably `semletOrVar` and `pragma`
- removed redundant `HackController` declaration

Miscellaneous Changes:
- lots of little white space removal, code was harder to follow
- removed echos in `airthm/tarithm` test

The Issue:
- defer is broken when inside a pragma block, this happened with both
  `cast` and `line`
- this caused the pop of the `config.debugUtilsStack` to happen for
  the block instead of the routine the template was being called in
---

## Notes for Reviewers
* note sure the exact cause
* leaving this PR here so I can come back to it or someone can help

### DIFF
--- a/compiler/ast/reports.nim
+++ b/compiler/ast/reports.nim
@@ -448,6 +448,7 @@ type
   DebugSemStepKind* = enum
     stepNodeToNode
     stepNodeToSym
+    stepSymNodeToNode
     stepNodeFlagsToNode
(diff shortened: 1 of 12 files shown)


saem:
Bors try

PR #261, opened 2022-03-29
Improve `compiler/utils/bitsets.nim` module

zerbina:
* Add tests for all functions (there were none previously)
* Turn procs into funcs
* All `bitset` functions use set views now. This makes them
  usable in cases where the bit-set is not stored as a `seq`
* Add `bitSetInclRange`, which aims to be more efficient than a
  simple `for x in a..b: s.bitSetIncl(x)`
* Add tests for `bitSetInclRange`
* Adjust functions in `nimsets` to take bit-set views as arguments

The `bitSetContains(x, y)` function previously calculated if "x is a
subset of y" instead of if "y is a subset of x". This is now changed
in order to make `bitSetContains` consistent with `contains`

### DIFF
--- a/compiler/ast/nimsets.nim
+++ b/compiler/ast/nimsets.nim
@@ -64,10 +64,11 @@ proc someInSet*(s: PNode, a, b: PNode): bool =
         return true
   result = false
 
-proc toBitSet*(conf: ConfigRef; s: PNode): TBitSet =
+proc inclTreeSet*(result: var TBitSetView, conf: ConfigRef; s: PNode) =
+  ## Includes all elements from tree-set `s` into `result`
+  assert result.len == int(getSize(conf, s.typ))
   var first, j: Int128
   first = firstOrd(conf, s.typ[0])
-  bitSetInit(result, int(getSize(conf, s.typ)))
   for i in 0..<s.len:
     if s[i].kind == nkRange:
       j = getOrdValue(s[i][0], first)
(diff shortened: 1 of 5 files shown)


compiler/utils/bitsets.nim
```diff
@@ -15,58 +15,108 @@ type
   TBitSet* = seq[ElemType]    # we use byte here to avoid issues with
                               # cross-compiling; uint would be more efficient
                               # however
+  TBitSetView* = openArray[ElemType] ## A view into a bit set. This doesn't
+                                     ## necessarily have to be a `TBitSet`,
+                                     ## but can be any byte sequence storing
+                                     ## representing a bit set
```
alaviss, line 21: If you want to alias `openArray[X]`, you have to write it out explicitly for now as the compiler might interpret this as a concrete view type, which is an unstable feature.

zerbina: Even when only used as a parameter type? That's a bit annoying, but I'll use `openArray[ElemType]` for now then

compiler/ast/nimsets.nim
```diff
@@ -64,10 +64,11 @@ proc someInSet*(s: PNode, a, b: PNode): bool =
         return true
   result = false
 
-proc toBitSet*(conf: ConfigRef; s: PNode): TBitSet =
+proc inclTreeSet*(result: var openArray[ElemType], conf: ConfigRef; s: PNode) =
```
alaviss: You're probably missing an import or so?

alaviss (commented):
LGTM, can merge once the compilation failure is fixed.

compiler/utils/bitsets.nim
```diff
@@ -11,62 +11,116 @@
...
+
+func bitSetInit*(b: var TBitSet, length: int) =
+  ## Creates a new bitset in `b` with a size-in-bytes of `length * ElemSize` .
+  ## The resulting bitset allows for elements in the
+  ## range `0..<(length * ElemSize * 8)`
   newSeq(b, length)
 
-proc bitSetUnion*(x: var TBitSet, y: TBitSet) =
+func bitSetUnion*(x: var openArray[ElemType], y: openArray[ElemType]) =
+  ## Calculate the union between sets `x` and `y
```
alaviss:
```suggestion
  ## Calculate the union between sets `x` and `y`
```

That should fix docgen...

zerbina: Ah, thanks. I wasn't able to figure out why docgen was failing while all other test passed.

Should I revert the `TBitSetView` -> `openArray[ElemType]` change then, as it didn't cause the issue?

alaviss: Feel free to

saem (approved):
Bors r+

Woot, this is gaining steam!

PR #265, opened 2022-04-03
Testament clean up targets

saem:
## Summary
* testament now tests with js and cpp targets more widely
* consequently test target specification needed to be adjusted

## Details
* this touches many test's specification
* it's a long road, so not everything is going to be perfectly categorized

---

## Notes for Reviewers
* some of the later changes for target support were more quick and dirty
  * it's just waaaay to much to slog through in one go

I'm considering introducing the following things:
* `native` target, which uses the same as the compiler
* `!js` and other such target exclusion capabilities


saem:
note to self, rebase so the commit list / diff is a bit cleaner.

saem:
This is effectively done elsewhere and no longer necessary. If someone feels otherwise leave a comment.

PR #268, opened 2022-04-06
Improve internal reporting of errors in `vm`/`vmgen`

### LINKED ISSUE #160: Refactor `stackTrace` in the `vm` implementation
VM `rawExecute` uses an implicitly injected `return` statement that is added by `template stackTrace()`. This makes code much harder to read, since you can't properly see where the loop is terminated. 

```diff
template stackTrace(
    c: PCtx,
    tos: PStackFrame,
    pc: int,
    sem: ReportTypes,
  ) =
  stackTraceImpl(c, tos, pc, c.debug[pc], instLoc())
  localReport(c.config, c.debug[pc], sem)
- return
```

Remove implicitly injected `return` and add it in a way that is visible to the reader.

```diff
    of opcLdArr:
      # a = b[c]
      decodeBC(rkNode)
      if regs[rc].intVal > high(int):
        stackTrace(c, tos, pc, reportVmIdx(regs[rc].intVal, high(int)))
+       return
```

Hint: in order to track all the usages of the `stackTrace` template, you can add `{.hint: "used here".}` or `{.warning: "used here".}`

zerbina:
Reporting of guest code related errors was done in an inconsistent
manner via either `stackTrace` or `globalReport`, where `stackTrace` used
`localReport` and an injected `return` statement to terminate the
execution loop. `stackTrace` was sometimes used in nested functions,
thus not terminating execution at all.

### Changes
Both the execution engine and `vmgen` now use exceptions for error
propagation internally, translating them into a result value at their
respective interface edges. In case of an error, the result value stores the
error report. Otherwise, it stores the actual result.

In addition, some usages of `globalReport` are also replaced with the
 new error propagation mechanism.

Depending on whether or not a VM invocation is expected to return
something, errors reported from VM execution and/or `vmgen` are either
turned into a `nkError` node or handled directly via `handleReport`

Fixes https://github.com/nim-works/nimskull/issues/160

---

This implements an adjusted/simplified version of the proposal from the VM discussion thread. Error reporting should be a bit more consistent and streamlined now, making future changes to it a bit easier. 

## Notes for Reviewers
* I'm not sure if I got the details regarding `newError` usage right
* I was also not sure what the best approach to putting the guest stack-trace into the `nkError` was, so the stack-traces are simply dropped right now

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -449,15 +449,11 @@ proc tryConstExpr(c: PContext, n: PNode): PNode =
   c.config.m.errorOutputs = {}
   c.config.errorMax = high(int) # `setErrorMaxHighMaybe` not appropriate here
 
-  try:
-    result = evalConstExpr(c.module, c.idgen, c.graph, e)
-    if result == nil or result.kind == nkEmpty:
-      result = nil
-    else:
-      result = fixupTypeAfterEval(c, result, e)
-
-  except ERecoverableError:
+  result = evalConstExpr(c.module, c.idgen, c.graph, e)
+  if result == nil or result.kind in {nkEmpty, nkError}:
     result = nil
+  else:
+    result = fixupTypeAfterEval(c, result, e)
 
   c.config.errorCounter = oldErrorCount
   c.config.errorMax = oldErrorMax
@@ -475,13 +471,20 @@ proc semConstExpr(c: PContext, n: PNode): PNode =
   if result == nil:
     #if e.kind == nkEmpty: globalReport(n.info, errConstExprExpected)
     result = evalConstExpr(c.module, c.idgen, c.graph, e)
-    if result == nil or result.kind == nkEmpty:
-      if e.info != n.info:
+    assert result != nil
+    case result.kind
+    of {nkEmpty, nkError}:
+      let withContext = e.info != n.info
+      if withContext:
         pushInfoContext(c.config, n.info)
+
+      if result.kind == nkEmpty:
         localReport(c.config, e.info, SemReport(kind: rsemConstExprExpected))
-        popInfoContext(c.config)
       else:
-        localReport(c.config, e.info, SemReport(kind: rsemConstExprExpected))
+        localReport(c.config, result)
+
+      if withContext:
+        popInfoContext(c.config)
       # error correction:
       result = e
     else:
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -17,6 +17,8 @@ import
     parseutils
   ],
   ast/[
+    errorhandling,
+    errorreporting,
     lineinfos,
     renderer, # toStrLit implementation
     trees,
@@ -54,12 +56,14 @@ import
     vmprofiler,
     gorgeimpl,
     vmdeps,
+    vmerrors,
     vmgen,
     vmdef
   ]
 
 
 import ast/ast except getstr
+import std/options as stdoptions
 
 
 const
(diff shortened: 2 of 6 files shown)


compiler/sem/sem.nim
```diff
@@ -475,13 +471,19 @@ proc semConstExpr(c: PContext, n: PNode): PNode =
   if result == nil:
     #if e.kind == nkEmpty: globalReport(n.info, errConstExprExpected)
     result = evalConstExpr(c.module, c.idgen, c.graph, e)
-    if result == nil or result.kind == nkEmpty:
-      if e.info != n.info:
+    assert result != nil
+    if result.kind in {nkEmpty, nkError}:
```
saem: Stylistic point, please prefer case statements, it nudges the compiler in the direction of simple successive pattern matches.

compiler/vm/vm.nim
```diff
@@ -587,11 +561,18 @@ template takeAddress(reg, source) =
   when defined(gcDestructors):
     GC_ref source
 
-proc rawExecute(c: var TCtx, start: int, sframe: StackFrameIndex): TFullReg =
-  assert sframe == c.sframes.high
+proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): TFullReg =
+  ## Runs the execution loop, starting in frame `tos` at program counter `pc`.
```
saem, line 565: Thank you!

compiler/vm/vmerrors.nim
```diff
@@ -0,0 +1,44 @@
+## This module contains the definitions for error reporting inside
```
saem, line 1: Excellent, thanks

compiler/vm/vmgen.nim
```diff
@@ -1482,18 +1528,11 @@ proc setSlot(c: var TCtx; v: PSym) =
...
-  # MUST raise an exception that is captured by `sem.tryConstExpr` in sem. In
-  # the future this needs to be removed, `checkCanEval` must return a
-  # `true/false` bool.
-  #
-  # For more elaborate explanation of the related code see the comment
-  # https://github.com/nim-works/nimskull/pull/94#issuecomment-1006927599
-  #
-  # This code must not be reached
-  raiseRecoverableError("vmgen.cannotEval failed")
+func cannotEval(c: TCtx; n: PNode) {.noinline, noreturn.} =
```
saem, line 1531: This isn't a hack as in we shouldn't avoid the raise?

compiler/vm/vmops.nim
```diff
@@ -152,25 +152,6 @@ when defined(nimHasInvariant):
     of MultipleValueSetting.cincludes: copySeq(conf.cIncludes)
     of MultipleValueSetting.clibs: copySeq(conf.cLibs)
 
-proc stackTrace2(c: PCtx, report: SemReport, n: PNode) =
```
saem, line 155: Kill it, kill it!  :D

compiler/vm/vm.nim
```diff
@@ -2362,22 +2353,98 @@ proc rawExecute(c: var TCtx, start: int, sframe: StackFrameIndex): TFullReg =
...
+    let errKind = res.report.kind
+
+    let stId = config.addReport(wrap(res.stackTrace))
+    let rId = config.addReport(wrap(res.report))
+
+    if not report:
+      # TODO: put the stack-trace into the error somehow, nothing happens
+      # with it right now. Maybe use report's `context`?
+
+      result = newError(config, node, errKind, rId, instLoc())
```
saem, line 2415: IIRC, error node now has specialized storage so it might be possible to put a report object in there, in which case creates report object to store the trace. Otherwise, serialize to a string, shove it in there like a son.

(on my phone, so hard to look things up)

compiler/vm/vm.nim
```diff
@@ -2362,22 +2353,98 @@ proc rawExecute(c: var TCtx, start: int, sframe: StackFrameIndex): TFullReg =
...
+    result = res.result
+    if node != nil and result.info.line < 0:
+      result.info = node.info
+  else:
+    let errKind = res.report.kind
+
+    let stId = config.addReport(wrap(res.stackTrace))
+    let rId = config.addReport(wrap(res.report))
+
+    if not report:
```
saem: If you have both sides of the if, please prefer avoiding negation and simplify the boolean expression. It's easier to quickly scan/skim.

saem (changes requested):
Code is looking much improved, the more uniform handling is very welcome.

I suggested a possible way to handle putting the exception trace in the nkError node, on the phone so it's a bit light on details. I can dig in more on the weekend (slim chance earlier) if it comes up short.

The minor style remarks aren't do or die, shortly evolving the code style, eventually as it cements it can go into the linter.

compiler/vm/vmgen.nim
zerbina: I don't think so, no. `vmgen` has to report about code that is not evaluate-able by the VM.

For example, `cannotEval` would be invoked for the `x` inside the when's condition expression:
```nim
let x = 0
when x == 0:
  discard
```
Resulting in this error message: `cannot evaluate at compile time: x`

An alternative would be to check for unevaluate-able expressions during sem-check instead of during VM code-gen.

saem: I meant the exception based signaling, but it works for now. Thanks for providing your perspective.

compiler/vm/vm.nim
zerbina: So, should I add the stack-trace's report ID as a `nkIntLit` to the error node then? Serializing it to a string, would require going through `cli_reporter`

saem: Yes, that makes sense. Adding it in as an additional argument is likely the way to go.

zerbina:
Requested style adjustments are applied. In addition, the stack-trace's report id is now added to the error node as a `nkIntLit`.

In the future, VM execution related errors should use their own report type (e.g. `VmReport`)

saem (approved):
Bors r+

PR #272, opened 2022-04-10
Implement variant object branch changing for `gc:refc`

zerbina:
This was previously forbidden and rejected at run-time. In order to get
`refc` closer to `arc`'s behaviour, it is now allowed.

The new logic is implemented in cgen and thus does not apply to the JS
back-end. It works by injecting reset logic for the assigned-to
discriminator's record-case in front of the assignment.

Behaviour is now the same as with `arc`, which means that:
* a branch reset now also happens when changing the discriminator value
  to a new value which refers to the same of-branch as the old one
* only the 'from' (source) branch is reset. This can lead to unexpected
  behaviour or crashes when modifying padding bits/bytes

---

As discussed on Matrix.

With `refc`, changing the discriminator value to something that referred to the same of-branch didn't perform a reset previously, so this is a breaking change. However, it is consistent with `arc` now.

### DIFF
--- a/compiler/backend/ccgstmts.nim
+++ b/compiler/backend/ccgstmts.nim
@@ -9,6 +9,8 @@
 
 # included from cgen.nim
 
+from ast/enumtostr import searchObjCase
+
 const
   RangeExpandLimit = 256      # do not generate ranges
                               # over 'RangeExpandLimit' elements
@@ -1565,17 +1567,61 @@ proc asgnFieldDiscriminant(p: BProc, e: PNode) =
   initLocExpr(p, e[0], a)
   getTemp(p, a.t, tmp)
   expr(p, e[1], tmp)
-  if optTinyRtti notin p.config.globalOptions:
+
+  if p.config.selectedGC == gcRefc:
+    # TODO: the branch reset logic should be generated into a standalone
+    # function to which a call is then emitted here
+
+    # Setup a loc for the object that holds the discriminator (only needed
+    # for the accesor)
+    var le: TLoc
+    initLocExpr(p, dotExpr[0], le)
+    var accessor = rdLoc(le)
+
+    let field = dotExpr[1].sym
+    var ty = dotExpr[0].typ
+    # Find the object type of which the discriminator is part
+    # XXX: duplicates the logic in `ccgexprs.lookupFieldAgain`
+    while ty != nil:
+      ty = ty.skipTypes(skipPtrs)
+      assert(ty.kind in {tyTuple, tyObject})
+      if lookupInRecord(ty.n, field.name) != nil:
+        break
+      if not p.module.compileToCpp:
+        accessor.add(".Sup")
+      ty = ty[0]
+
+    internalAssert(p.config, ty != nil, field.info)
+    assert ty.kind == tyObject
+
+    # Find the discriminator's corresponding nkRecCase node
+    let recCase = searchObjCase(ty, field)
+
+    # only reset when the new discriminator is different
+    linefmt(p, cpsStmts, "if($1!=$2){$n", [rdLoc(a), rdLoc(tmp)])
+    # Emit case reset logic
+    specializeResetN(p, accessor, recCase, ty)
+
+    # We assume that no padding bits were modified, so no need to clear
+    # the destination branch
+    #specializeResetN(p, rdLoc(tmp), recCase, objTyp)
+
+    lineF(p, cpsStmts, "} $n", [])
+
+
+  elif optTinyRtti notin p.config.globalOptions:
     let field = dotExpr[1].sym
     genDiscriminantCheck(p, a, tmp, dotExpr[0].typ, field)
     localReport(p.config, e, reportSem rsemCaseTransition)
+
   genAssignment(p, a, tmp, {})
 
 proc genAsgn(p: BProc, e: PNode, fastAsgn: bool) =
   if e[0].kind == nkSym and sfGoto in e[0].sym.flags:
     genLineDir(p, e)
     genGotoVar(p, e[1])
-  elif optFieldCheck in p.options and isDiscriminantField(e[0]):
+  elif isDiscriminantField(e[0]) and
+    (optFieldCheck in p.options or p.config.selectedGC == gcRefc):
     genLineDir(p, e)
     asgnFieldDiscriminant(p, e)
   else:
--- a/tests/objvariant/tbranch_change.nim
+++ b/tests/objvariant/tbranch_change.nim
@@ -1,0 +1,159 @@
+discard """
+  description: "Tests to make sure branch changing works correctly"
+  targets: "c cpp"
+  matrix: "--gc:refc; --gc:arc; --gc:orc"
+"""
+
+# XXX: remove this together with arc/orc related workarounds once
+# `tbranch_reset_arc_orc_issue.nim` works
+const isArcOrOrc = defined(gcArc) or defined(gcOrc)
+
+block:
+
+  # This is also a test for generic variant objects
+
+  type A[T] = object
+    case kind: bool
+    of false:
+      a: T
+    of true:
+      b: T
+
+
+  # variant with simple types (no destructors exist)
+  when not isArcOrOrc:
+    block:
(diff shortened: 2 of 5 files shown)


zerbina:
This should probably be mentioned/changed in the manual too

saem:
How tricky would it be to make work in jsgen?

compiler/backend/ccgstmts.nim
```diff
@@ -1565,17 +1567,61 @@ proc asgnFieldDiscriminant(p: BProc, e: PNode) =
   initLocExpr(p, e[0], a)
   getTemp(p, a.t, tmp)
   expr(p, e[1], tmp)
-  if optTinyRtti notin p.config.globalOptions:
+
+  if p.config.selectedGC == gcRefc:
+    # TODO: the branch reset logic should be generated into a standalone
```
saem, line 1572: Along with the inline pragma/hint, then we can have llvm/gcc figure out the rest.

compiler/backend/ccgstmts.nim
```diff
@@ -1565,17 +1567,61 @@ proc asgnFieldDiscriminant(p: BProc, e: PNode) =
...
+    # for the accesor)
+    var le: TLoc
+    initLocExpr(p, dotExpr[0], le)
+    var accessor = rdLoc(le)
+
+    let field = dotExpr[1].sym
+    var ty = dotExpr[0].typ
+    # Find the object type of which the discriminator is part
+    # XXX: duplicates the logic in `ccgexprs.lookupFieldAgain`
+    while ty != nil:
```
saem, line 1585: I wonder if this will work for distincts, mind you in sure that's broken elsewhere too if it is here.

(Thinking out loud, not sure how to even think about that yet)

tests/objvariant/tbranch_change.nim
```diff
@@ -0,0 +1,150 @@
+discard """
+  description: "Tests to make sure branch changing works correctly"
+  targets: "c cpp"
+  matrix: "--gc:refc"
```
saem: Might as well cover arc and orc here too, the parts that don't work under them can be put into a separate test with a `knowIssue` in its testament spec. It will presently disable the test, but in the future they'll be run and checked to see if they've been fixed.

saem:
Just had a look through the tests, looks like hot code reload tests are failing. I'm wondering if we're injecting clears when we shouldn't during reload. Either that or maybe the rtti that we're referencing is a mismatched version than what we need for the particular variant?

saem (changes requested):
Minor change for some extra test coverage, but this is pretty cool!

tests/objvariant/tbranch_change.nim
zerbina: That would be more or less all tests then. `arc`/`orc` doesn't emit a branch reset for types that don't have a destructor (fixable, but would be another breaking change).

I could add a dummy `string` field to all variants in order to force `arc`/`orc` to generate the reset logic.

zerbina:
> Just had a look through the tests, looks like hot code reload tests are failing. I'm wondering if we're injecting clears when we shouldn't during reload

I can't reproduce the test failures on my machine unfortunately. For me, they succeed (tried it with `debug`, `release` and `danger` compiler builds)

> Either that or maybe the rtti that we're referencing is a mismatched version than what we need for the particular variant?

The branch reset logic doesn't use RTTI, so that shouldn't lead to any problems

zerbina:
> How tricky would it be to make work in jsgen?

Not much, I'd say. If wanted, I could do it as part of this PR.

A new PR would probably be better though, I think

tests/objvariant/tbranch_change.nim
saem: > That would be more or less all tests then. `arc`/`orc` doesn't emit a branch reset for types that don't have a destructor (fixable, but would be another breaking change).

Maybe I'm being especially naive, but doesn't the lack of reset logic being emitted seem like a bug? As in if we allow reset and it only clears the fields sometimes is inconsistent. I'm guessing it's all because of how things "happen to have been implemented", and inject destructors is blind to this stuff.

> I could add a dummy `string` field to all variants in order to force `arc`/`orc` to generate the reset logic.

I'd prefer we flag the type perhaps and work off that if we can?

saem:
I'll try the tests locally, might need to specify both the c and CPP targets... Or there is something really weird going on.

zerbina:
Okay, I was able to reproduce the issue. It only happens when running the whole `dll` category, not when running the test alone.

The issue is with `newOp` in `server.nim` depending on the previous "no branch reset when same active branch" behaviour.

We can either:
* setup the discriminator in the constructor expression (it seems like the separate discriminator assignment was intentional though)
* wrap the assignment in a `cast(uncheckedAssign)` block
* only do the branch reset when changing the active branch
* drop this PR

tests/objvariant/tbranch_change.nim
zerbina: > Maybe I'm being especially naive, but doesn't the lack of reset logic being emitted seem like a bug?

The intention behind how it's implemented with `arc` seems to be to only prevent memory leaks and crashes due to corruption of objects with destructors. Since variant objects without destructors don't (directly) exhibit the aforementioned problems, branches are simply not reset on change for those.

If branch switching is to be officially allowed then I'd say it is a bug.

> I'd prefer we flag the type perhaps and work off that if we can?

I'm not sure I understand what you mean by flagging the type? Putting the failing test blocks behind a `when not defined(gcArc)`?

saem:
> Okay, I was able to reproduce the issue. It only happens when running the whole `dll` category, not when running the test alone.

Nice on the reproduction, I forgot to mention that testament had this silly special category handling. DLL is one of those special categories, it actually does a whole bunch of test runs and lies in the output about it being one test.

> The issue is with `newOp` in `server.nim` depending on the previous "no branch reset when same active branch" behaviour.

OK, so I see a few things and I blech... here goes:
- `result` auto initializes for a return value
- if the programmer then starts setting values changing the order changes the reset
- the current behavior feels like a hack to sorta kinda work in some cases, but not really
- during hot code reload we need to not inject resets and/or allow disabling it over regions, but we can revisit hcr so that's fine

No matter what this is yet more inconsistencies between refc and arc that subtly breaks code... Joy

> We can either:
> 
> * setup the discriminator in the constructor expression (it seems like the separate discriminator assignment was intentional though)
> * wrap the assignment in a `cast(uncheckedAssign)` block
> * only do the branch reset when changing the active branch
> * drop this PR

My preference, only do the reset if the branch changes or just say this is the behaviour now and make the test match.

I'm trying to suss out what's better, thoughts?

tests/objvariant/tbranch_change.nim
saem: By flagging the type I meant a tyFlag on the type data, but I'm sure that has issues.

zerbina: Getting `injectdestructors` to also emit a branch reset for types without destructors only requires removing a single `if` check

zerbina:
Personally, I'd prefer the "only reset when changing the active branch" behaviour since I find it more intuitive. Implementing this requires more work though, and, depending on how it's done, is also less efficient at run-time.

saem:
> Personally, I'd prefer the "only reset when changing the active branch" behaviour since I find it more intuitive. Implementing this requires more work though, and, depending on how it's done, is also less efficient at run-time.

Would you prefer to wait, maybe put this into draft and then revisit in a month or two to see if we gain further insight, maybe easier to implement, or someone picks it up?

I'm far more buried in test cleanup than I thought I would be, but I can dig in or someone else might be all over it on the revisit.

zerbina:
I'd like to continue working on this, but would temporarily put this PR on hold, at least until the big VM PR is merged.

zerbina:
Implementing the branch-switching logic in the code-generator was a flawed idea from the start. The code-generator should only be responsible for mapping a branch *reset* (i.e. zero'ing the memory the record-case occupies) into the target language -- figuring out where and when to perform such resets should be the responsibility of an earlier stage.

In addition, the planned removal of the legacy GC options (this includes `refc`) renders this PR/feature stale.

PR #274, opened 2022-04-11
testament: internals steps and assertions

saem:
## Summary
* introduce test steps and assertions to testament's internal data model

## Details
* 
---
## Notes for Reviewers
* very much a work in progress


zerbina:
What do you think about also including artifact related assertions? E.g. assertions for binary size, generated code, etc.

saem:
> What do you think about also including artifact related assertions? E.g. assertions for binary size, generated code, etc.

Very much so, they're already supported in the current incarnation and I think they're a must for ensuring certain subtle behaviours about generated source.

PR #242, opened 2022-02-28
vm: store data as raw contiguous bytes

zerbina:
## Summary

This replaces the `PNode`-based data representation in the VM with
storing the data as raw contiguous bytes, similar to how it works on
the c/cpp targets. `VmType` encodes the information necessary to
interpret the raw bytes.

## Details

- since `PType`s store a lot of information irrelevant for the VM, they
 are first translated into a format (`VmType`) that's faster and easier to
 traverse/query. The mapped-to-types are de-duplicated, meaning that
 if two VM types have the same ID (currently the same instance), they're
 also the same type, making type equality checks very fast. This
 respects structural-vs-nominal types, but currently not `distinct` and
 named-vs-unnamed tuples.
- location type information is stored separately from the actual data
- `const` values are still stored using `PNode` representation, which
 requires a translation step whenever data crosses the compiler/VM
 border
- it is enforced that locations are only accessed via handles of
 compatible type, reporting an error on violation
- all memory access is now also ensured to not go outside of
 VM-owned memory, with an error being reported on violation

### Known changes in observable behaviour

- `repr` output in compile-time/NimScript contexts changed a bit
 when compiling with `--mm:refc`. The output is more closer to
 that with `--mm:arc`.
- when passing non-`NimNode` values to non-`static` parameters with
 `genAst`/`getAst`, the invoked macro always receives it as typed AST
 now (even if the parameter is of `untyped` type)

---

The initial goal was for this PR to be _the_ VM refactoring PR, but it shifted to being about unblocking both the DOD AST and a `vmgen` rewrite.

I'm not happy with the overall quality, but I think it has to suffice for the first step. The biggest issues:
* the code in `vm.nim` is too complex
* `vmgen.nim` was in a bad state before, but it's even worse now
* there's lots of redundancy in both the instruction set and the implementation
* I had to introduce lots of temporary hacks/workarounds

Lots of parts/components are in a transitional state and not finished yet. I do have a clear plan on how to further proceed from here. In the bigger VM refactoring picture, this PR marks the first big step, with further steps being:
* use a IR for `vmgen`; rewrite `vmgen` in general
* revise the instruction set; do more work at code-gen time as opposed to during execution
* architectural improvements
* restructure/refactor the execution loop
* bring all unfinished efforts to completion (type layout, type creation, allocator, etc.)
I'm going to move to a more incremental development approach (smaller, more frequent PRs) for the rest of the VM refactoring from this PR on.

[...]

PR #275, opened 2022-04-11
Remove experimental compile-time C FFI

zerbina:
The feature was not tested as part of CI and it's interactions with the
new VM are also not completely clear. If wanted/needed, it could be brought
back in an adjusted form after the VM refactoring is finished.

This is a breaking change.

---

Discussed [here](https://github.com/nim-works/nimskull/pull/242#discussion_r841113759)


saem:
Looks like a usage snuck into a random test (strtabs). 😬

Once that passes I'm all good to merge, it's a lot less code. 🎉🥳

zerbina:
Hmm, `test_nimscript.nims` should have also caught this. Looks like it's not executed.

The issue is with some modules importing `fenv`, which has some importc'ed globals. I guess declaration (but not usage) of `importc` variables has to be allowed for now.

saem:
Yay!

Bors r+

PR #277, opened 2022-04-16
do node consistency

krux02:
do nodes are now consistently lambda expressions no matter if they have arguments or not.
stmtLists are no longer lifted implicitly to lambda expressions

## Summary
 * do nodes without arguments are not implicitly converted to bare `stmtList` anymore.
 * bare `stmtList` is not lifted to a lambda expression anymore.
 * The behavior that got removed was a hack that cause more problems than it solved. 

## Details

It caused problems in my macro library, especially the implicit lifting of `stmtList` to a λ expression is a liability that is really hard to work around. It is not a feature to be proud upon. It is a breaking change, especially cases when `do` blocks need to be treated as bare `stmtList` and not λ expressions.

---

This is a change I've made a long time ago on my own fork of Nim.

### DIFF
--- a/compiler/ast/reports.nim
+++ b/compiler/ast/reports.nim
@@ -14,6 +14,7 @@
 ## reused by external tooling - custom error pretty-printers, test runners
 ## and so on.
 
+import macros
 import std/[options, packedsets]
 
 import
@@ -410,11 +411,18 @@ func reportSym*(
 
   SemReport(kind: kind, ast: ast, str: str, typ: typ, sym: sym)
 
-template withIt*(expr: untyped, body: untyped): untyped =
-  block:
-    var it {.inject.} = expr
-    body
-    it
+# template withIt*(expr: untyped, body: untyped): untyped =
+#   block:
+#     var it {.inject.} = expr
+#     stripDoNode(body)
+#     it
+macro withIt*(expr: untyped, body: untyped): untyped =
+  let body = if body.kind == nnkDo: body[^1] else: body
+  let it = ident"it"
+  quote:
+    var `it` {.inject.} = `expr`
+    `body`
+    `it`
 
 template tern*(predicate: bool, tBranch: untyped, fBranch: untyped): untyped =
   ## Shorthand for inline if/else. Allows use of conditions in strformat,
--- a/compiler/front/in_options.nim
+++ b/compiler/front/in_options.nim
@@ -213,6 +213,10 @@ type
       ## Historically and especially in version 1.0.0 of the language
       ## conversions to unsigned numbers were checked. In 1.0.4 they
       ## are not anymore.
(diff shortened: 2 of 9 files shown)


krux02:
@haxscramper the code I broke is mostly code you wrote, so I would really like to hear your opinion about it.

alaviss:
Can you provide an example for the behavior change?

compiler/ast/reports.nim
```diff
@@ -410,11 +411,18 @@ func reportSym*(
...
-  block:
-    var it {.inject.} = expr
-    body
-    it
+# template withIt*(expr: untyped, body: untyped): untyped =
+#   block:
+#     var it {.inject.} = expr
+#     stripDoNode(body)
+#     it
+macro withIt*(expr: untyped, body: untyped): untyped =
```
haxscramper, line 419: Not sure if changing to `do` section is necessary - it worked perfectly fine without that, at least of the use cases I had. The template is supposed to emulate rust's field update uses. 

If `withIt` stops working due to behavior changes and now has to be written via macro, I'm doubting this *API* change is necessary.

compiler/front/in_options.nim
```diff
@@ -213,6 +213,10 @@ type
       ## Historically and especially in version 1.0.0 of the language
       ## conversions to unsigned numbers were checked. In 1.0.4 they
       ## are not anymore.
+    optOldDoNode
```
haxscramper, line 216: Do we need to provide backward compatibility support right now? As in - no reason to start piling up more dialects and legacy feature support during refactoring.

haxscramper (commented):
Aside from `withIt` template, most of the code was not written by me, so I don't think I can provide any in-depth input on this. So just two minor questions here

compiler/ast/reports.nim
krux02: it was used somewhere with an explicit `do` argument and that do node needs to be taken out again. I wanted to change the usage to not use `do` but it was weird.

compiler/front/in_options.nim
krux02: If you tell me to drop supporting the old behavior, I do it right now. I just did this backwards compatibility, because originally before I got banned, I wanted it to be a PR for Araq-Nim. And that is what I preserved here.

saem:
This video from racket conf offers a better solution for alternatives and handling multiple groups to be associated with a previous line: https://youtu.be/UUWYTW3Xck8

Preview:
```
foo(stuff):
  | One statement
    Still part of one statement
  | Second statement, both one and two are arguments to foo
```

Without something as the above we can't move forward with this do change.

saem:
This has gone stale and the multiple statements via pipes is required first. It can be resurrected if and when needed.

PR #281, opened 2022-04-17
vm: move `ExpandToAst` logic into own instruction

zerbina:
The logic for template expansion as needed by the `mExpandToAst` magic
(used by `quote`) is now implemented as a separate instruction instead
of as part of `IndCallAsgn`. This makes the role of template expansion
less confusing ("why are templates 'called' inside the VM?", "weren't
they expanded already?") and also allows for a different instruction
argument make-up. The latter is required for both a future overhaul
of macro expansion and the refactored VM.

User-facing behaviour stays the same

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -1287,6 +1287,34 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): TFullReg =
               ast[i] = a.sym.ast[i]
             ast[bodyPos] = transformBody(c.graph, c.idgen, a.sym, cache=true)
             ast.copyTree()
+    of opcExpandToAst:
+      decodeBC(rkNode)
+
+      let
+        callExprAst = c.constants[regs[rb].intVal.int]
+        prc =         callExprAst[0].sym
+        prevFrame =   c.sframes[tos].next
+
+      assert callExprAst.kind in nkCallKinds
+      assert prc.kind == skTemplate
+
+      let genSymOwner = if prevFrame > 0 and c.sframes[prevFrame].prc != nil:
+                          c.sframes[prevFrame].prc
+                        else:
+                          c.module
+      var templCall = newNodeI(nkCall, c.debug[pc])
+      templCall.add(newSymNode(prc))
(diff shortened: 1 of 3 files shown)


compiler/vm/vm.nim
```diff
@@ -1287,6 +1287,32 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): TFullReg =
               ast[i] = a.sym.ast[i]
             ast[bodyPos] = transformBody(c.graph, c.idgen, a.sym, cache=true)
             ast.copyTree()
+    of opcExpandToAst:
+      decodeBC(rkNode)
+
+      let callExprAst = c.constants[regs[rb].intVal.int]
```
saem: Please favour let blocks

compiler/vm/vm.nim
```diff
@@ -1287,6 +1287,32 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): TFullReg =
...
+                          c.module
+      var templCall = newNodeI(nkCall, c.debug[pc])
+      templCall.add(newSymNode(prc))
+      for i in 1..rc-1:
+        let node = regs[rb+i].regToNode
+        node.info = c.debug[pc]
+        templCall.add(node)
+
+      var a = evalTemplate(templCall, prc, genSymOwner, c.config, c.cache, c.templInstCounter, c.idgen)
+      if a.kind == nkStmtList and a.len == 1: a = a[0]
```
```suggestion
      if a.kind == nkStmtList and a.len == 1: # flatten if a single statement
        a = a[0]
```

Also, reading this code, what if it's empty? Should this be an nkEmpty then?

saem (changes requested):
Make makes more sense that it works this way.

One question about the statement list handling, style bit if you're in there.

compiler/vm/vm.nim
zerbina: I don't know. In order to preserve the previous behaviour, I left this as is

saem: OK, let's stick with it for now, it does feel like a bug waiting to happen. 😬

saem (approved):
Bors r+

PR #285, opened 2022-04-22
stdlib: remove deprecated `NimIdent` and `NimSym`

zerbina:
They were deprecated since as far back as 0.18.0 and are
not compatible with the new VM in their current form

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -716,8 +716,8 @@ type
     mNccValue, mNccInc, mNcsAdd, mNcsIncl, mNcsLen, mNcsAt,
     mNctPut, mNctLen, mNctGet, mNctHasNext, mNctNext,
 
-    mNIntVal, mNFloatVal, mNSymbol, mNIdent, mNGetType, mNStrVal, mNSetIntVal,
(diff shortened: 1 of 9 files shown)


zerbina:
An alternative to removing them would be to turn them into `distinct NimNode`

haxscramper:
If I'm not mistaken `magic: "NIdent"` can be removed as well, from the implementation - `mNIdent`, `opcNIdent` in the VM, opcodes and TMagic. `mNSetIdent`, `opcNSetIdent` and other magics that were added for the deprecated API. 

Otherwise, the change looks perfectly fine to me - `macros.nim` API certainly needs some cleanup from the ancient code that it accumulated.

zerbina:
> If I'm not mistaken `magic: "NIdent"` can be removed as well, from the implementation - `mNIdent`, `opcNIdent` in the VM, opcodes and TMagic. `mNSetIdent`, `opcNSetIdent` and other magics that were added for the deprecated API.

Good catch, thanks!

zerbina:
The corresponding magics and implementation in `vm`/`vmgen` are now also removed

saem (approved):
Yaaas, burn it!

Bors r+

PR #287, opened 2022-04-23
ci: remove "Nim in Action" related dependencies

zerbina:
The "Nim in Action" test category was removed in commit `59e382f2f`


saem (commented):
Good catch, that'll shave some time I hope.

Bors r+

PR #288, opened 2022-04-23
stdlib: add experimental ``results`` module

zerbina:
This extends the standard library with a `Result` type.

Based on https://github.com/disruptek/badresults with the following
changes:
* removed windows `{.inline.}` work-around
* documentation adjustments
* added `lent` and `sink` annotations
* used `func` instead of `proc` where applicable
* added functional-style procedures (e.g. `map`, `filter`, etc.)
* added some `runnableExamples`
* adjusted the tests to work without ``balls``
* the tests are also run inside the VM

---

### Notes for Reviewers
* the documentation is probably a bit insufficient. Any help in this area is appreciated!
* is it possible to include the original author/contributor information somehow?

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/lib/experimental/results.nim
+++ b/lib/experimental/results.nim
@@ -1,0 +1,283 @@
+##[
+This module implements a result-type which encapsulates both a success and a
+failure in one value. Only one of the two states can be active.
+
+Based on https://github.com/disruptek/badresults, which was itself based on
+https://github.com/arnetheduck/nim-result.
+
+Exception translation
+=====================
+
+Using `get` or `value` raises an exception if the `Result` stores a failure.
+For `Result[T, E]` the exception raised is:
+ * if `E` is a `ref` type that inherits from `Exception`, the `error` value
+ * if a custom overload of `toException <#toException, E>`_ exists, the result
+  of `toException(r)`
+ * otherwise, a new `ResultError[E]` with `error` as the payload
+
+]##
+
+import std/options
+
+type
+  ResultError*[E] = object of ValueError
+    error: E
+
+  ResultErrorRef*[E] = ref ResultError[E]
+
+  Result*[T, E] = object
+    ## A `Result` is typically used to augment a return value of type `T`
+    ## indicating that a computation (proc, etc...) is uncertain and should
+    ## return a value of type `T` or end in an error of type `E`. This is
+    ## useful to avoid encoding entirely expected error cases as values instead
+    ## of resorting to exception handling for program control flow or various
+    ## side-channel data.
+    case o: bool
+    of false:
+      e: E
+    of true:
+      v: T
+
+template reportNoError() =
+  raise ResultErrorRef[void](msg: "Result does not contain an error")
+
+func error*[T, E](self: Result[T, E]): lent E {.inline.} =
+  ## Retrieves the error from a `Result`; raises `ResultError[void]`
+  ## if the `Result` is not an error.
+  if self.isErr:
+    result = self.e
+  else:
+    reportNoError()
+
+func unsafeGet*[T, E](self: var Result[T, E]): var T {.inline.} =
+  ## Fetches the value of a `Result` if set, undefined behavior if unset.
+  ##
+  ## See also:
+  ## * `unsafeGet <options.html#unsafeGet,Option[T]>`_
+  assert not isErr(self)
+  result = self.v
+
+func unsafeGet*[T, E](self: Result[T, E]): lent T {.inline.} =
+  ## Fetches the value of a `Result` if set, undefined behavior if unset.
+  ## See also:
+  ## * `unsafeGet <options.html#unsafeGet,Option[T]>`_
+  assert not isErr(self)
+  result = self.v
+
+func unsafeGet*[E](self: Result[void, E]) {.inline.} =
+  ## Does nothing if `Result` is set, undefined behavior if unset.
+  ## See also:
+  ## * `unsafeGet <options.html#unsafeGet,Option[T]>`_
+  assert not self.isErr
+
+func newResultError[E](e: sink E;
+                       s: sink string): ResultErrorRef[E] {.inline, nimcall.} =
+  ## capturing ResultError...
+  ResultErrorRef[E](error: e, msg: s)
+
+template toException*[E](err: E): ResultErrorRef[E] =
+  ## Refer to `this <#exception-translation>`_ for more information.
+  ##
+  ## The returned exception is required to be a `ref`.
+  mixin `$`
+  when compiles($err):
+    newResultError(err, "Result isErr: " & $err)
+  else:
+    newResultError(err, "Result isErr; no `$` in scope.")
+
+template raiseResultError[T, E](self: Result[T, E]) =
+  mixin toException
+  when E is ref Exception:
+    if self.error.isNil: # for example Result.default()!
+      raise ResultErrorRef[void](msg: "Result isErr; no exception.")
+    else:
+      raise self.error
+  else:
+    type ET = typeof(self.error.toException)
+    when typeof(ET) is ref:
+      raise self.error.toException
+    else:
+      # In theory, the non-ref exception could be turned into a ref
+      # exception here
+      {.error:
+        "The provided `toException` overload doesn't produce a `ref` value".}
+
+func ok*[E](R: typedesc[Result[void, E]]): Result[void, E] =
+  ## Returns a result as success.
+  R(o: true)
+
+func initSuccess*[E](self: var Result[void, E]) =
+  ## Sets a result to success.
+  runnableExamples:
+    var r: Result[void, string]
+    r.initSuccess()
+    doAssert r.isOk
+
+  self = ok[E](typeOf(self))
+
+func ok*[T, E](R: typedesc[Result[T, E]]; v: sink T): Result[T, E] =
+  ## Returns a result with a success and value.
+  runnableExamples:
+    let r = Result[int, string].ok(42)
+    doAssert r.value == 42
+
+  R(o: true, v: v)
+
+proc initSuccess*[T, E](self: var Result[T, E]; v: sink T) =
+  ## Sets the result to success and updates value.
+  self = ok[T, E](typeOf(self), v)
+
+func err*[T, E](R: typedesc[Result[T, E]]; e: sink E): Result[T, E] =
+  ## Returns a result with an error.
+  runnableExamples:
+    let r = Result[int, string].err("uh-oh")
+    doAssert r.error == "uh-oh"
+
+  R(o: false, e: e)
+
+func initFailure*[T, E](self: var Result[T, E]; e: sink E) =
+  ## Sets the result as an error.
+  self = err[T, E](typeOf(self), e)
+
+func isOk*(self: Result): bool {.inline.} = self.o
+func isErr*(self: Result): bool {.inline.} = not self.o
+
+func `==`*(a, b: Result): bool {.inline.} =
+  ## Equal if both `a` and `b` have the same ok/err status, then compares the
+  ## value or error pairs using available `==` for those type of values.
+  if a.isOk == b.isOk:
+    if a.isOk: a.v == b.v
+    else:      a.e == b.e
+  else:
+    false
+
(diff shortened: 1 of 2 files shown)


lib/experimental/results.nim
```diff
@@ -0,0 +1,187 @@
...
+  of `toException(r)`
+ * otherwise, a new `ResultError[E]` with `error` as the payload
+
+]##
+
+type
+  ResultError*[E] = ref object of ValueError
+    error: E
+
+  Result*[T, E] = object
```
saem, line 28:
```suggestion
  Result*[T, E] = object
    ## a `Result` is typically used to augment a return value of type `T` indicating that a computation (proc, etc...)
    ## is uncertain and should return a value of type `T` or end in an error of type `E`. This is useful to avoid encoding
    ## entirely expected error cases as values instead of restoring to exception handling for program control flow or
    ## various side-channel data.
```

lib/experimental/results.nim
```diff
@@ -0,0 +1,187 @@
...
+    case o: bool
+    of false:
+      e: E
+    of true:
+      v: T
+
+proc error*[T, E](self: Result[T, E]): lent E {.inline.} =
+  ## Retrieves the error from a `Result`; raises `ResultError[void]`
+  ## if the `Result` is not in error.
```
saem:
```suggestion
  ## if the `Result` is not an error.
```

lib/experimental/results.nim
saem: I'd favor having two distinct types, `Success` and `Failure`, that are based on `Result`. Then when it's statically known that a `Result` is one or the other a coercion can be performed. Meanwhile the compiler can catch any obvious errors.

lib/experimental/results.nim
```diff
@@ -0,0 +1,187 @@
...
+    let r = Result[int, string].err("uh-oh")
+    doAssert r.error == "uh-oh"
+
+  R(o: false, e: e)
+
+proc err*[T, E](self: var Result[T, E]; e: sink E) =
+  ## Sets the result as an error.
+  self = err[T, E](typeOf(self), e)
+
+proc isOk*(self: Result): bool {.inline.} = self.o
```
saem: These can be funcs?

zerbina:
It seems like the exception's `name` field is always set to `null` for the JS target, leading to a crash when trying to convert it to a Nim `string`.

This needs a separate fix. For now, I'd simply prevent the failing test-case from running when targeting JS

lib/experimental/results.nim
saem: Predicate are fine, but a result type is much nicer when we have map, flatMap, mapErr, etc... and perhaps filter and filterErr which move to Option.

With the distinct types mentioned earlier, recover as a special case flatMapErr which converts to a Success means the compiler can help book keep exhaustiveness.

tests/stdlib/tresults.nim
```diff
@@ -0,0 +1,209 @@
...
+    get(b)
+    check c.error == 5
+    check d.error == 3
+    check f.error is ref IOError
+    check $a == "Ok(5.3)"
+    check $b == "Ok()"
+    check $e == "5.3"
+    when nimvm:
+      # Issue: Exception names are not properly set inside the VM
+      check $f == "BIO:/uh-oh"
```
zerbina: This issue is fixed once the refactored VM is merged

lib/experimental/results.nim
```diff
@@ -0,0 +1,187 @@
...
+proc isErr*(self: Result): bool {.inline.} = not self.o
+
+func `==`*(a, b: Result): bool {.inline.} =
```
saem, line 145:
```suggestion
func `==`*(a, b: Result): bool {.inline.} =
  ## equal if both `a` and `b` have the same ok/err status, then compares the value or error pairs using available
  `==` for those type of values.
```

saem (changes requested):
Some doc suggestions.

The map and distinct type is more for discussion

lib/experimental/results.nim
```diff
@@ -0,0 +1,187 @@
...
+  mixin toException
+  when E is ref Exception:
+    if self.error.isNil: # for example Result.default()!
+      raise ResultError[void](msg: "Result isErr; no exception.")
+    else:
+      raise self.error
+  else:
+    raise self.error.toException
+
+proc ok*[E](R: typedesc[Result[void, E]]): Result[void, E] =
```
saem: These ok procs feel far too benignly named for how easily they can break logic. They're outright nuking a result, for someone reading code using Result they'd rightly be shocked as to the severity of the mutation.

zerbina: How about `initSuccess`/`initFailure`?

saem: Sounds good to me!

tests/stdlib/tresults.nim
```diff
@@ -0,0 +1,209 @@
+discard """
+  description: "Tests for the std/experimental/results module"
+  matrix: "--gc:refc; --gc:arc"
+  targets: "c cpp js"
+"""
+
+# This file was based on the ``test.nim`` file
```
saem, line 10: Would it be asking too much to use `unittest` inside here?

I think a unit testing library is a better direction for stdlib unit testing imo, rather than mostly stylistic structures through blocks and the lik. <-- feel free to disagree

tests/stdlib/tresults.nim
```diff
@@ -0,0 +1,209 @@
...
+
+macro check(msg: string, body: untyped) =
+  result = newStmtList()
+
+  body.expectKind nnkStmtList
+  for n in body:
+    result.add(newCall(bindSym"check", n, msg))
+
+template expect(e: typed, body: untyped) =
+  doAssertRaises e, body
```
saem, line 30: These are pretty much identital to `std/unittest`

tests/stdlib/tresults.nim
zerbina: > Would it be asking too much to use `unittest` inside here?

I don't think so, no. Unfortunately, `std/unittest` doesn't work at compile-time (yet), since it uses non-compile-time globals and methods.

I'm not against using `unittest` here, but it would mean that `Result` can't be tested in the VM for now.

saem: Oh, that's silly. Ok, looks like outside of this unit test needs to be fixed. Nevermind for now.

lib/experimental/results.nim
zerbina: Hmm, wouldn't that require lots of manual conversion? If, for example, `ok` returns a `Success`, one would have to do convert it back to a `Result` prior to assignment. A `converter` could be used of course, but I dislike their implicit nature

lib/experimental/results.nim
```diff
@@ -0,0 +1,187 @@
...
+For `Result[T, E]` the exception raised is:
+ * if `E` is a `ref` type that inherits from `Exception`, the `error` value
+ * if a custom overload of `toException <#toException, E>`_ exists, the result
+  of `toException(r)`
+ * otherwise, a new `ResultError[E]` with `error` as the payload
+
+]##
+
+type
+  ResultError*[E] = ref object of ValueError
```
zerbina: Should `ResultError` be a `object of ValueError` instead? With the current version, wanting to use the underlying `object` type requires some helper logic

alaviss: Yea, it should be converted to `object`.

lib/experimental/results.nim
```diff
@@ -0,0 +1,273 @@
...
+  else:
+    raise ResultError[void](msg: "Result does not contain an error")
+
+func error[T, E](self: var Result[T, E]): var E {.inline.} =
+  ## Retrieves the error from a `Result`; raises `ResultError[void]`
+  ## if the `Result` is not an error. For internal use only
+  if self.isErr:
+    result = self.e
+  else:
+    raise ResultError[void](msg: "Result does not contain an error")
```
alaviss: Is this used for anything at all? I think direct `.e` access would be better for internal code.

lib/experimental/results.nim
```diff
@@ -0,0 +1,273 @@
...
+
+func ok*[T, E](R: typedesc[Result[T, E]]; v: sink T): Result[T, E] =
+  ## Returns a result with a success and value.
+  runnableExamples:
+    let r = Result[int, string].ok(42)
+    doAssert r.value == 42
+
+  R(o: true, v: v)
+
+proc initSuccess*[T, E](self: var Result[T, E]; v: sink T) =
```
alaviss, line 126: While @saem [prefers](https://github.com/nim-works/nimskull/pull/288#pullrequestreview-951120827) this long name, I think this is terrible for usability.

This sort of thing is meant to reduce keystrokes. With the rename it went from:

```nim
proc foo(): Result[int, string] =
  result.ok 10
```

to

```nim
proc foo(): Result[int, string] =
  result.initSuccess(10)
```

Which is almost as long as `Result[int, string]`! I think a shorter name should be used.

Alternatively you can consider a `Success[T]`/`Failure[E]` type pair which wraps a value so that they can be transparently assigned to `Result[T, E]` (via converters, I would assume). It would at least enable something like this:

```nim
proc foo(): Result[int, string] =
  ok(10) # This returns `Success[int]`
```

saem: The short name is a terrible idea, because the design without the distinguished type is off. Why are we suddenly mutating it so casually. No, the API isn't right.

Saving key strokes is meaningless, it's not the big blocker, reading is. The compiler is full of convenient to write code, let's optimize reading.

In the latter half as you point out with distinguished types we tend towards a better API. At that point instead of mutating, `ok` is constructing a result. That's much better. By having the mutation version have an ugly name we make it clear we're doing something suspect.

lib/experimental/results.nim
zerbina: It's only used for `takeErr`. I'll move the `raise` part into a template instead and remove this overload

lib/experimental/results.nim
zerbina: > Alternatively you can consider a `Success[T]`/`Failure[E]` type pair which wraps a value so that they can be transparently assigned to `Result[T, E]` (via converters, I would assume). It would at least enable something like this:

I tried it, but converters (currently) can't be used for these kinds of implicit conversions.
```nim
type A[B, C] = object
  b: B
  c: C

converter conv[B, C](x: B): A[B, C] =
  result.b = x

proc p(): A[int, string] = 1
discard p()
```
The compiler is unable to instantiate `C` in this example.

saem: Not sure why we need converters, this is all phantom type information.

zerbina: How would you implement alaviss example without converters?

saem: Yup you're right, I started to write it and then I realized how it wouldn't work. There is nothing between alias and distinct type without the use of a converter or inheritance, barring some potential hackery with a third generic phantom param and that's pretty cursed.

Sigh, in which case:
- `initSuccess` could be `toOk` or `asOk` (little weird as it's mutating it)
- while `ok` constructs a new result

Aside/Reflection: This particular Result type feels a lot more like an `Either`, which often is implemented as a generalization. Result in that context would be implemented as a subtype mapping to Either as "left" (wrong) vs "right" (right, hah!). As implemented presently `Result` is sightly less general and one step further might be having a `Try`which assumes the left/wrong is an exception (subtype/inheritance).

Enough of that, back to reality:
A single Result type and no subtypes is fine for now, and the ok that creates and something else but a bit shorter to mutate works.

lib/experimental/results.nim
saem: There should be a subtype relationship, but that turned out to be a pain.

saem: Leaving this contestation unresolved so it's easily referenced (retrospectively) that which we did not sort out.

saem (approved):
Bors delegate+

alaviss (approved):
LGTM. As @saem said, the main part is fine as is, we can add the ergonomics after once we figured out how it is used in the compiler.

bors r+

PR #292, opened 2022-04-30
Consistently initialize `Exception.name` across back-ends

zerbina:
Creating an `Exception` without raising it left it's `name` field as nil
on the JS and VM back-ends. For the C targets, the name initialization
was injected during code-gen.

Initializing the `name` field now either happens in `newException` or
when raising the exception and thus works the same across all back-ends.
Creating an exception without `newException` doesn't set the `name`
field on the C back-ends anymore.

In addition, when raising an exception on the JS target, the exception's
name is not overridden if it's non-nil.

This is a breaking change.

---

If the exception's name should also be automatically set when not using `newException`, I'd propose doing this in `transf` instead of during code-gen

### DIFF
--- a/compiler/backend/cgen.nim
+++ b/compiler/backend/cgen.nim
@@ -429,16 +429,6 @@ proc genObjectInit(p: BProc, section: TCProcSection, t: PType, a: var TLoc,
       var r = if mode == constructObj: addrLoc(p.config, a) else: rdLoc(a)
       linefmt(p, section, "#objectInit($1, $2);$n", [r, genTypeInfoV1(p.module, t, a.lode.info)])
 
-  if isException(t):
-    var r = rdLoc(a)
-    if mode == constructRefObj: r = "(*$1)" % [r]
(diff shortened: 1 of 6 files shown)


saem:
This feels more consistent, if some platform compatibility bits come up we can consider those as they arise and revisit.

Bors r+

zerbina:
The failing cpp test is due to the recent `sem` [change](https://github.com/nim-works/nimskull/pull/290)

saem:
> The failing cpp test is due to the recent `sem` [change](https://github.com/nim-works/nimskull/pull/290)

Exclude CPP from the test and leave a comment in the test with a comment (`# knownIssue: ...`).

zerbina:
The failing cpp test should be fixed by [https://github.com/nim-works/nimskull/pull/294](https://github.com/nim-works/nimskull/pull/294). I'll rebase, once it's merged

saem:
Merge in progress. 🎉

PR #293, opened 2022-05-01
Turn some more localReports into newError

Clyybber:
## Summary
* Turn some more localReports into newError
* Remove nkComesFrom


saem:
Boo, test failures, I wanted to merge this.

PR #294, opened 2022-05-01
cpp: fix `var` parameter regression

zerbina:
Fixes at least some of the code-gen issues introduced by
https://github.com/nim-works/nimskull/pull/290. All previously
disabled tests work again.

For more details, refer to ``tests/ccgbugs/tvar_param.nim``.

Since temporaries are now also used in places where they previously
weren't (for the C back-end), this is a breaking change.


saem:
This is much more sensible than before, nicely done!

Bors r+

PR #297, opened 2022-05-03
vm: improve aliasing of `slots` under ORC

zerbina:
Indirectly accessing `slots` via a `TStackFrame` pointer was done to
work around `shallowCopy` not being available for `seq` with ORC.

The shallow copy is now replicated with the `cursor` pragma, saving
one pointer indirection

---

I ran the test suite with an ORC enabled compiler and there were no (VM related) failures. This change allows to get rid of `SlotStorage` in the VM refactoring [PR](https://github.com/nim-works/nimskull/pull/242)


saem:
It's a good sign when we're seeing a few lines for a win.

Bors r+

PR #298, opened 2022-05-03
koch: use epoch 0 if metadata is not available

alaviss:
## Summary
In cases where no metadata is available (ie. a source dump from Github),
use epoch 0 instead of attempting to parse an empty string.
    
This fixes a crash when the Github-generated source tarball is used
(which doesn't have the metadata).


saem:
Bors r+

Nice that's one less "weird error" to debug and bit of trivia to recall.

saem:
Bors r+

Nice that's one less "weird error" to debug and bit of trivia to recall.

PR #301, opened 2022-05-07
stdlib: fix `styleCheck` errors in ``results``

zerbina:
---

It would probably be a good idea to add a test similar to ``tstrict_funcs_imports.nim`` (just with `styleCheck:error` instead), in order to automatically catch these kinds of issues/regressions


saem:
You can add the test of you want.

Bors delegate+

zerbina:
The test is added [here](https://github.com/nim-works/nimskull/pull/305). Once that PR is merged, I'll rebase and bors this one

PR #304, opened 2022-05-08
qualifying compiler imports as such

disruptek:
## Summary
I cannot build nimph with nimskull because I want to simply `--path="$nim"` and then import compiler modules as in `import compiler/utils/pathutils`.  This changes the code so that pattern works.

## Details
Compiler imports are now qualified with a `compiler` prefix. This
restores the ability to use the compiler API, previously resulted in
errors due to failed imports.

Due to the nature of this change many files had to be modified.

This particular fix was motivated in order to have `disruptek/nimph`
working under nimskull. Compiler API consumers can now use the API by
including `--path:"$nim"` in their cfg.

Misc:
- `tests/compilerapi/tcompilerapi` should catch regressions now
- doc work around, disabled example in `compiler/utils/nimdebugutils`,
    required because runnable examples doesn't account for the `path`
----
I don't see why I should need this `nim.cfg` to include `--path="$config"` in the repository root.  This behavior feels intentional, though.  What's the thinking here?

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -10,15 +10,15 @@
 ## abstract syntax tree + symbol table
 
 import
-  ast/[
+  compiler/ast/[
     lineinfos, # Positional information
     idents, # Ast identifiers
     ast_types # Main ast type definitions
   ],
-  front/[
+  compiler/front/[
     options
   ],
-  utils/[
+  compiler/utils/[
     ropes,
     int128 # Values for integer nodes
   ],
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1,7 +1,7 @@
-import utils/ropes
+import compiler/utils/ropes
 import std/[hashes]
 
-from front/in_options import TOption, TOptions # Stored in `PSym`
+from compiler/front/in_options import TOption, TOptions # Stored in `PSym`
 
 
 type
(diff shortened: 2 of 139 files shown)


haxscramper:
@saem I think I asked about `compiler/ast` vs simple `ast/` imports when I was doing code rearrangements in my original PR, and you said https://github.com/nim-works/nimskull/pull/177#pullrequestreview-857898738

> If we have imports like `compiler/foo`then that implies we could just as easily end up with `testament/bar` as an import. At least right now we'd spot the `..` and know it's broken more easily than a myriad of directories one might import.

After using compiler code (in my projects) for some time, I found this to be unnecessarily annoying, breaking uses of the compiler API in several annoying ways - in my own code I import `compiler/ast`, but inside the compiler it is written as `ast/`, so I have to make sure there are two different import paths for my code to work properly each time. 

I imagine @disruptek is making this change for the exact same reason, and I see almost no value in keeping it as `ast/` out of the fear someone might import testament or something else they were not supposed to be doing.

saem:
@haxscramper live and learn, the lesser evil here is `import compiler/ast/ast`.

saem:
@alaviss can you take a peek at what I've done to the `*.cfg`s?

Honestly, couldn't follow the matrix discussion after a while because it was all over the place. 😬

alaviss:
@saem The configs looks good to me

saem:
Hopefully that (PR open/close) will sort the bot.

bors r+

saem:
blech, have to wait until the whole thing is done 😒

saem:
Bors r+

saem:
Bors r+

Come on buddy.

PR #307, opened 2022-05-12
stdlib: move `varargsLen` to `macros.nim`

zerbina:
It could only be implemented in `system` due to a misuse of a
compiler magic, making it dependent on a VM implementation detail.

### Changes
- `varagsLen` is now located in `macros.nim` and doesn't depend on a
 compiler  magic anymore
- the `tvarargslen.nim` test is moved to the `macros` category

---

The macro was originally added to `system` in order to make it available with no extra imports. There also was a discussion in the original [PR](https://github.com/nim-lang/Nim/pull/12907) about whether or not this macro should even be needed in the first place, but the PR was merged without the discussion concluding.


saem:
Bors r+

zerbina:
I missed adding an import as well as a wrong white-space fix, sorry

saem:
Bors r+

PR #308, opened 2022-05-12
internal: don't use VM callback for `addFloatRoundtrip`

zerbina:
The function doesn't need to be overridden with a VM callback
implementation in order to work in the VM. In addition, the usage of
a callback was problematic in this case, as the callback has no way
to know if the input is a `float32` or `float64` and always treated
it as the latter

---

Fixes the `tfloats.nim` test in the data layout [PR](https://github.com/nim-works/nimskull/pull/242)


saem (approved):
Bors r+

PR #309, opened 2022-05-14
add an int id field to TNode

saem:
## Summary
* added an id field to TNode
* migrated comment storage to use the id field instead of address
* small step in adding more data oriented concepts
---


saem:
Bors r+

PR #311, opened 2022-05-14
Fixes #263

### LINKED ISSUE #263: testament in `sexp` mode fails on Windows
Running testament on Windows with `sexp` output formatting fails with `reMsgsDiffer` due to testament treating the backslash characters in the file path as the beginning of a string escape sequence.

### Example

`tests\new\tfile.nim`:
```nim
discard """
  nimoutFormat: sexp
  cmd: "nim c --msgFormat=sexp --filenames:canonical $file"
"""

{.hint: "hint".} #[tt.Hint
     ^ (User :str "hint") ]#
```

Platform: Windows
Commit: 365a680bd8e7bcba0b59af40e43096dcb510d5e0

Steps to reproduce:
* run `koch.py boot -d:release`
* run `koch.py test r tests/new/tfile.nim`

### Actual Output

```
Failure: reMsgsDiffer

Expected inline Hint annotation at tests\new\tfile.nim(6, 6):

- (User :location ("tests\new\tfile.nim" 6 6) :severity Hint :str "hint")

Given:

+ (User :location ("tests
ew      file.nim" 6 6) :reportFrom ("pragmas.nim" 1540 20) :reportInst ("pragmas.nim" 1540 20) :severity Hint :str "hint")


  :location[0] expected "tests\\new\\tfile.nim", but got "tests\u000Aew\u0009file.nim" (["tests\\new\\tfile.nim" -> "tests\u000Aew\u0009file.nim"])
```

### Expected Output

Test success

canelhasmateus:
## Summary
Changes to Sexp parsing. 

We are changing it because it does not  work as expected when the string contains literal \n , \t, \r. this is very common in path  expressions, and are currently parsed as literals. 

I'm not sure about the message, nor the effects.

### DIFF
--- a/lib/experimental/sexp.nim
+++ b/lib/experimental/sexp.nim
@@ -143,15 +143,6 @@ proc parseString(my: var SexpParser): TTokKind =
       of 'f':
         add(my.a, '\f')
         inc(pos, 2)
-      of 'n':
-        add(my.a, '\L')
-        inc(pos, 2)
-      of 'r':
-        add(my.a, '\C')
-        inc(pos, 2)
-      of 't':
-        add(my.a, '\t')
-        inc(pos, 2)
       of 'u':
         inc(pos, 2)
         var r: int


saem:
I don't think the problem is the parsing, more likely the issue is the encoding. The thing generating the sexp should escape the string prior to making a sexp string.

So whenever we go from a string to a sexp string, we should escape it or require that the user escapes it. Regardless, there should be a way to create an escaped or unescaped (if it was done elsewhere) sexp string.

saem (changes requested):
I think the issue is the API and usage.

We need to support string escaping in the API, and that should be used by the structured error generator.

canelhasmateus:
makes sense. I also didn't think this is was a good solution, but I opened the PR to get the ball rolling

PR #312, opened 2022-05-15
Fix crash with "type expected" report

zerbina:
`reportBody` always expected a symbol, but it's not always present (non
nil)


haxscramper:
The structure of the reports is still kind of rough, so in the future we might need to decrease branching in the reporters (because branching *inside* the `of <report kind>` implies there should be more report kinds), but that's a question for another time. 

Otherwise it looks perfectly fine to me, so when tests pass it can be merged

PR #317, opened 2022-05-16
Fix type field related issues with variant objects

zerbina:
With the c/c++ back-end, during object construction, variant objects had
their type fields initialized _before_ any discriminators were assigned
to, meaning that only the default branches had valid type fields.

In the case that a default branch wasn't also the selected branch, the
previously initialized type fields aliased with fields of the selected
branch, breaking the zero-initialization guarantee. Both `refc` and
ARC/ORC were affected.

Object initialization (`ccgexprs.genObjConstr`) now performs type field
initialization _after_ assigning all discriminator values.

---

The `doInitObj` extra parameter is a bit ugly, but it was the simplest solution not requiring larger refactorings that I could come up with.

@haxscramper originally suggested that the test for this fix should go into `ccgbugs` (which makes sense), but after the following discussion regarding test suite organization, I think it makes more sense to put it under `objvariants` while also testing it with all back-ends/GCs

### DIFF
--- a/compiler/ast/types.nim
+++ b/compiler/ast/types.nim
@@ -227,6 +227,10 @@ proc searchTypeForAux(t: PType, predicate: TTypePredicate,
   else:
     discard
 
+proc searchTypeNodeFor*(n: PNode, predicate: TTypePredicate): bool =
+  var marker = initIntSet()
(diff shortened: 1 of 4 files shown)


saem:
Bors r+

PR #318, opened 2022-05-19
vm: add back support for non-ordinal `case` selectors

zerbina:
`case` statements using _slicelists_ in their `of` branches now compile
and run again when the selector is of `float` or `string` type


saem (approved):
Bors r+

PR #319, opened 2022-05-20
vm: don't depend on `PType` for exception handling

zerbina:
- use `VmType`s type relation analysis during exception type matching
- pass the name of the raised exception's static type as a constant.
 This makes exception name setting behaviour the same as on other
 back-ends
- `HeapSlot.nType` is removed since it's no longer needed

In addition, a long-standing bug with re-raising in a function with
non-void return type is also fixed

This is the first step towards decoupling VM execution from compiler
state

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -189,10 +189,25 @@ template toException(x: DerefFailureCode): untyped =
   ## `Result` -> exception translation
   toVmError(x, instLoc())
 
-proc reportException(c: TCtx; sframe: StackFrameIndex, raised: LocHandle, excType: PType) =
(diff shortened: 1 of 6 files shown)


compiler/vm/vm.nim
```diff
@@ -189,10 +189,25 @@ template toException(x: DerefFailureCode): untyped =
   ## `Result` -> exception translation
   toVmError(x, instLoc())
 
-proc reportException(c: TCtx; sframe: StackFrameIndex, raised: LocHandle, excType: PType) =
+proc reportException(c: TCtx; sframe: StackFrameIndex, raised: LocHandle) =
```
saem, line 192: Yay, less parameters!

saem (approved):
Bors r+

PR #320, opened 2022-05-24
vm: make `types` list append-only

zerbina:
## Summary
A separate hash-table overlay is used for looking up structural types, making
the previous list re-orderings used for speeding up the linear search obsolete.

This is an important step towards replacing `PVmType` with integer IDs

### Future work:
- `genTuple` should also return the already computed structural-type hash
- the equality function used by `getOrIncl` should be made override-able

### DIFF
--- a/compiler/vm/vmtypegen.nim
+++ b/compiler/vm/vmtypegen.nim
@@ -38,6 +38,71 @@ template paddedSize(info: VmType): untyped =
   ## Round the size up so that `info.sizeInBytes % (1 shl info.alignment) == 0`
   paddedSize(info.sizeInBytes, 1'u shl info.alignment)
 
+func hash(t: PVmType): Hash {.inline.} =
+  # the ref itself already acts as a unique id
+  hash(cast[int](t))
+
+func hash(t: VmType): Hash =
+  ## Creates a hash over the fields of `t` that are relevant for structural
+  ## types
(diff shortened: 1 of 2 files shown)


compiler/vm/vmtypegen.nim
```diff
@@ -38,6 +38,71 @@ template paddedSize(info: VmType): untyped =
...
+    of akSet:                 hash(t.setLength)
+    of akCallable, akClosure: hash(t.funcTypeId.int)
+    of akDiscriminator:       hash(t.numBits)
+    of akArray:               hash(t.elementCount) !& hash(t.elementType)
+    of akObject:
+      var r = 0
+      for (_, ft) in t.objFields.items:
+        r = r !& hash(ft)
+      r
+    of akInt, akFloat, akString, akPNode:
```
saem, line 62: So we wouldn't match a type who's first element is an int with an int parameter in a function?

saem:
Left a comment if a non issue then we're good to go.

Bors d

saem:
Bors d+

compiler/vm/vmtypegen.nim
zerbina: No, they wouldn't match. Procedural types use a separate process that turns them into unique IDs first (`getOrCreateFuncType`)

PR #321, opened 2022-05-24
vm: improve how types are passed to execution

zerbina:
Run-time-type-information in the form of `PType` is now passed to
execution via a separate list, as most instructions only need the
`PVmType`.

This makes `vmgen` a bit faster, since type generation/lookup doesn't
require a slow linear search with `sameType` anymore.

---

The PR is ready and only marked as a draft in order to not waste CI resources. Each commit only depends on it's previous commits, so if wanted, this PR could be split into a series of PRs

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -496,7 +496,7 @@ proc findExceptionHandler(c: TCtx, f: var TStackFrame, raisedType: PVmType):
       while c.code[pc].opcode == opcExcept:
         let excIndex = c.code[pc].regBx - wordExcess
         let exceptType =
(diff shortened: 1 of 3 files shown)


zerbina:
I forgot to mark the PR as a draft on submit, so the test suite is still being run 🤦

saem:
Separate would be nice, but a squash works.

Bors d+

zerbina:
The first three commits are split off here:
- https://github.com/nim-works/nimskull/pull/322
- https://github.com/nim-works/nimskull/pull/323
- https://github.com/nim-works/nimskull/pull/324

The fourth split off PR:
- https://github.com/nim-works/nimskull/pull/325

PR #322, opened 2022-05-25
vm: store built-in `VmType`s in `types`

zerbina:
This makes all existing types available in the `types` seq, which is a
requirement for the planned `ref -> integer ID` change


saem:
Bors r+

PR #323, opened 2022-05-25
vm: split up `opcSetType`

zerbina:
The instruction was used for both setting the type of an address
register (pointer) and for internally setting the type of a `NimNode`.

Both usages require different kinds of type information (`PType`
vs. `VmType`), so it makes sense to split them up


saem:
Bors r+

Issue #41, opened 2021-11-06
Clean up and reorder tests


haxscramper:
Moved to documentation https://nim-works.github.io/nimskull/contributing.html#writing-or-improving-tests

Original issue description

<details>

Todo list/guide for cleaning up existing tests

- [ ] Check if test name makes sense - `t123123_b.nim` does not make sense, change it to something matching what is being tested
- [ ] Reduce number of `echo`-based error testing. If you see direct `echo` in test consider changing it to the `doAssert` check instead
- [ ] If possible, provide explanation to the test logic
- [ ] Link relevant issues in the test description (`description:` field) or in comments
- [ ] Add test labels [link](https://github.com/haxscramper/nimskull/blob/test-naming-cleanup-1/doc/testament.rst#labels)

--------------

- [ ] Get rid of completely illegible file names like `t10489_a.nim`, `t10489_b.nim` - if you want to help someone to get to a concrete issue easier, you can make a *comment in a file* with a *proper link*. And that' where you can get a proper file name - original issue title.
- [ ] Structure tests around categories, not around some barely navigable mountain of folders that people won't ever be able to figure out which tool/language/stdlib this part belongs to. Top-level tool categories should be
  - [ ] language (https://github.com/nim-works/nimskull/projects/2)
  - [ ] stdlib
  - [ ] compiler - CLI interface, input handling
  - [ ] other tools - nimsuggest, nim doc
- [ ] Provide `description:` field for the test, more documentation comments and general explanation of the features tested.
- [ ] Reduce usage of `echo` in tests - `doAssert`, `==` are more than sufficient in a lot of cases. Not *all* of them, but their usage increase locality of the code - no need to look at this header in the 289-line file when this could've been replaced with `return "some string"` and that is later checked with `doAssert()`. [link](https://github.com/nim-works/nimskull/blob/b0933ba30ebb5760190471202bc3529b60bd8264/tests/casestmt/tcasestmt.nim#L46-L64) of how things should *not* be tested.

  ```nim
  discard """
  output:
  '''
  Not found!
  Found!
  1
  compiles for 1
  i am always two
  default for 3
  set is 4 not 5
  array is 6 not 7
  default for 8
  an identifier
  OK
  OK
  OK
  ayyydd
  '''
  """
  ```

Tests should be easy to understand for a new contributor, because you change something, then the test fails, and you have *no idea* what this thing was even testing - what could be more frustrating than trying to not only *why* it fails but *what* fails in the first place.

[...]

PR #327, opened 2022-05-26
Rename Tests in suite

canelhasmateus:
Just a bit of love for the test suite. 
The idea is to rename some tests, according to #41 . 
Alongside, possibly introduce better organization , via a possible "tag" key in the tests.

Everything is still *very* experimental - many test are a bit hard to rename.


canelhasmateus:
Ok, i kinda finished with the renaming, with the exception of :

* Arc tests with multiple tests inside
* Macro folders
* Error Messages, which should be changed by hax very soon

Of course, its very dirty , so further modifications are required before the merge.

Some points i would like to point out are:

* Aside from the starting `t` , are there any other naming conventions? Here i am using camelCase.

* The `description` field is pretty noisy - i included many titles and sometimes relevant commentaries. I'm inclined to leave it like this in terms of the contents - with some minor formatting - instead of trying to summarize / synthesize better descriptions based on the PR information.
  
In regards to the proposed `tags` field - i did some quick and dirty tagging without much thought behind it - just letting things bubble up and see where we get.

The 'official' ones - if they come to be, of course - could possibly come from this initial effort. After parsing it a little, here are some tags that seem promising / common. 

```
Features:

    async ; atomics ; defer;
    iterators  ; iterables ; closure_iterators
    import
    implicit_return



    type ; subtyping ; inheritance ; override ; field ; nesting ;
    var ; ref ; static ; ptr;
    type-inference ; type-alias ; distinct ; variants ; union ; untyped
    generics ; overload

Runtime

    case ; proc  ; branches; if ; for ;  loop ; variable ; expression ; asssignment ; recursion;  deref
    effect-system ; func ; discard ;
    scopes ; top_level ; global ; {.global.}

    cast; converters; copy ; move ; lent ;
    destructor ;  instantiation ; {.requiresinit.} ; unintialized ; reset
    gc; gc_arc ;  gc_arc ;  gc_orc ;  newruntime

    template ; macro ; pragma ; {.compiletime.} ; compiles ;
    semcheck ; codegen ; name_mangle ;

ARch / Backend

    arch_32-bit ; arch_openbsd
    backend_c ;  backend_cpp ;  backend_js; backend_vm ; js ;
    gcc_optimization; compiler_gcc ; {.inline.}

    {.exportc.} ; dll
    mode_release ; mode_debug


Errors

    test_regression
    error_compilation ; error_compile; error_compile?; error_fail_compilation;  error_fail_compile ; error_fail_runtime; error_runtime
    should_error ; stacktrace ; message ;  exception ; exception_handling


StdTypes ; Stdlib

    nimdoc ; runnableexamples
    stdlib ; sequtils ; typetraits
    array ; char ; datetime ; namedtuple ; literal ; floating_point ; int ; openarray
    string ; tables ; seqs ;  sets ; range
    repr ; ord ; bitwise ;

___

    {.noreturn.}
    definition

```

I'm open to further approaches to come up with these - i thought about scraping other prs / test contents for keywords and what-not , but that seemed a bit over-engineered and low value / effort.

zerbina:
Not a blocker, but I think using the full issue link in the test description might be better (e.g. `https://github.com/nim-lang/nim/issues/16558` instead of `#16558`) since it prevents possible future ambiguities.

saem:
Descriptions, Style, and Naming:
- *thank you*, survey/wide work like this shine a light on the little frictions
  that so often never get dealt with
- *line width to 80 characters* please; allows having side by side files open,
  as cross-referencing is often required while developing
- *issue references as urls* `#123` is ambiguous, please use the full url
- *snake_case_test_names* they're used in `lang` which is the latest, don't
  bother with the `t01_` prefix, it can just the `tmy_test_for_whatever`

Tag Taxonomy:
- *drop redundant information* by example, lets drop `backend_c` given we have
  target, this duplicate sources of truth means more to edit/maintain/etc
- *structured information takes precedence* reiterating the above, but if we
  know backend, feature, testing for failure, whatever via the testament spec or
  category, then don't replicate it by hand in the tag
- *limit names to [a-z_]* so, `{.requiresinit.}` to `pragma_requiresinit`
- *language spec for naming inspiration* see if we have names in `lang` tests if
  so let's try to be consistent with that, with simplified characters for names

> Ok, i kinda finished with the renaming, with the exception of :
> 
> * Arc tests with multiple tests inside
> * Macro folders
> * Error Messages, which should be changed by hax very soon

Multiple tests inside one are ok, as I mentioned there is an overhead to each
test. So combining them is beneficial up to a point, for example most tests
take about `0.25s` on my machine to compile and run, slower ones are multiple
seconds. If it takes longer than a second I think it's too much, so relatively
speaking I would run a category like lang get a rough average and then more than
4x that is probably too slow and I wouldn't combine further.

> Of course, its very dirty , so further modifications are required before the merge.
> 
> Some points i would like to point out are:
> 
> * Aside from the starting `t` , are there any other naming conventions? Here i am using camelCase.

It's a mismash, but I would take some cues from `lang`, it's probably the
cleanest.

> * The `description` field is pretty noisy - i included many titles and sometimes relevant commentaries. I'm inclined to leave it like this in terms of the contents - with some minor formatting - instead of trying to summarize / synthesize better descriptions based on the PR information.

Yup, the description helps, outside of line count I think that's good. If
someone can read the description and know the spirit of the test then they can
interpret, debug, judge, edit/rewrite, whatever. Thanks.

> In regards to the proposed `tags` field - i did some quick and dirty tagging without much thought behind it - just letting things bubble up and see where we get.

It's very instructive, thanks. I'd prefer fewer tags, but I'm good as long as we
can easily change our mind/query them. tuning up testament dry-run to act as tag
querying would be super helpful. As long as the output works with simple text
filter CLI tools, we're golden. Adding an output option of something like line
separated json where one can use `jq` might be all we need. 

> The 'official' ones - if they come to be, of course - could possibly come from this initial effort. After parsing it a little, here are some tags that seem promising / common.

Again, big thanks!

> ```
> Features:
> 
>     async ; atomics ; defer;
>     iterators  ; iterables ; closure_iterators
>     import
>     implicit_return
>  
>     type ; subtyping ; inheritance ; override ; field ; nesting ;
>     var ; ref ; static ; ptr;
>     type-inference ; type-alias ; distinct ; variants ; union ; untyped
>     generics ; overload
>
> Runtime
> 
>     case ; proc  ; branches; if ; for ;  loop ; variable ; expression ; asssignment ; recursion;  deref
>     effect-system ; func ; discard ;
>     scopes ; top_level ; global ; {.global.}
> 
>     cast; converters; copy ; move ; lent ;
>     destructor ;  instantiation ; {.requiresinit.} ; unintialized ; reset
> ```

Let's go with these, I'm hoping over time that their occurrences will fade over
time and we can shrink the list. The language spec should take care of most of
these.

> ```
>     gc; gc_arc ;  gc_arc ;  gc_orc ;  newruntime
> ```

These should be treated separately as structured data in other fields to do with
the test matrix

>     template ; macro ; pragma ; {.compiletime.} ; compiles ;
>     semcheck ; codegen ; name_mangle ;
> 
> ARch / Backend
> 
>     arch_32-bit ; arch_openbsd
>     backend_c ;  backend_cpp ;  backend_js; backend_vm ; js ;
>     gcc_optimization; compiler_gcc ; {.inline.}
> 
>     {.exportc.} ; dll
>     mode_release ; mode_debug

The mode ones, release and debug, are probably good categories to add to a
structured testament spec field that can then be queried.

> Errors
> 
>     test_regression
>     error_compilation ; error_compile; error_compile?; error_fail_compilation;  error_fail_compile ; error_fail_runtime; error_runtime
>     should_error ; stacktrace ; message ;  exception ; exception_handling

A number of these tags I think should be deriveable from the test spec rather
than typed in by a human I feel. We can have one list of tags, maybe separate
into ones humans enter vs ones testament auto-adds, then they can be queried via
a testament dry-run or whatever. Or visually inspect the human ones.

> 
> StdTypes ; Stdlib
> 
>     nimdoc ; runnableexamples
>     stdlib ; sequtils ; typetraits
>     array ; char ; datetime ; namedtuple ; literal ; floating_point ; int ; openarray
>     string ; tables ; seqs ;  sets ; range
>     repr ; ord ; bitwise ;
> 
> ___
> 
>     {.noreturn.}
>     definition
> ```

This category would benefit most from tags, at least the ability to filter on
them.

> I'm open to further approaches to come up with these - i thought about scraping other prs / test contents for keywords and what-not , but that seemed a bit over-engineered and low value / effort.

I'm thinking from a use case perspective, so here are test tags I'm not seeing
but would be immediately useful:

- `slow`: nobody likes slow tests when they want quick answers
- `io.file` & `io.network`: when there is file/network io, which besides being
  slow I vaguely remember a few of these being hdd intensive and people might
  not want to hammer their ssds. Notice how I introduced `.`, so yeah, tags
  _could_ get hierarchy... `io` is a tag, but so is `io.file`, so filtering on
  `io` does `io` as a prefix
- `benchmark`: correlates with `slow` but this is for all the measures

The two big uses cases I believe are:
- run all relevant tests for a change I made, but I need rapid feedback (fast!)
- run all tests that check language feature x, where x is `macros`, `generics`

The former because the test suite is so damn slow and categories don't help,
while the latter avoids running 10 categories or scouring them and lang spec
will only do so much even down the road.

saem:
Big Picture:
------------

So I bounced around reading the code, your comment, adding in my own thoughts.
The big thing about tags is querying/filtering and the ability to manage them,
which is also helped by querying and filtering. Get that right and everything
else becomes easier.

There are two aspects to this specific PR:
1. Descriptions, test names, naming/style, these I think are pretty close
2. tag taxonomy and rollout

The first I think can be put together and good to go, this PR effectively, the
second I would make as a separate PR to implement tagging in testament. To be
clear, the work you did in this PR was _instrumental_ in understanding how this
should work. It's really easy to see now thanks to your efforts, very much
appreciated.

The merge ordering would likely be the tag/filtering PR would land first; unless
you want to drop tags from this one. Maybe there is a better way?

Testament Tag PR:
-----------------

I'd recommend getting the following up and running, this would put you and
others in a very fast and productive loop once implemented:
- add `user` and `derived` tag field on the `TSpec` testament data type
- populate derived tags, not everything, just one or two
- start with a simple tag string validator lower case alphabet, underscore
  separators, and maybe `.` hierarchy, I think it's: `[a-z_]+((\.[a-z_]+)*)`;
  you can drop the `.` thing if it's easier
- make `dry-run` or a new `query` sub-command that can report and filter tags,
  for now it can be the derived ones and maybe `slow` as one tag
- make `all`, `category`, and `pcat` allow tag filtering, something like
  `testSpec.tags.anyIt(it.startsWith("io"))`

I believe that to be small and focused with good feedback/progress loop on its
own. But once merged you, or anyone, can try all sorts of tagging very quickly
derived or user defined and just _try things_.

Because we can query and filter them, if we don't like the tags at a given time,
just query to get a list and a bit of typing to fix it. Example, let's say we
decided to make the `slow` tag, then later on thought nope, all the slow tests
are basically `io`. We query + do some typing and we're done. Then later on
someone feels it should be `io`, `io.fs`, and `io.network`... same thing.

Not only that, but PRs for tag related changes are easy to accept because the
cost of improper categorization/whatever is super super low.

tests/ccgbugs/tDunnoAboutAName.nim
```diff
@@ -1,6 +1,8 @@
 discard """
   output: "[p = nil]"
   targets: "c cpp"
+  description: "C++ Atomics in union: ::<unnamed union>::<unnamed struct>::field with constructor not allowed in anonymous aggregate #13062"
```
saem:
```suggestion
  description: '''C++ Atomics in union:
::<unnamed union>::<unnamed struct>::field with constructor not allowed in
anonymous aggregate https://github.com/nim-lang/nim/issues/13062'''
```

example for how to do multi-line descriptions

tests/ccgbugs/tcodeGenDistinctUInt64On32Bit.nim
```diff
@@ -1,4 +1,7 @@
-
+discard """
+description: ""distinct uint64" type corruption on 32-bit, when using {.borrow.} operators ; The problem only appears when borrowed operators for a "distinct uint64" type force a type inference."
```
saem: extra double quote.

tests/parallel/tbetterSpawnErrorMessage.nim
```diff
@@ -1,6 +1,7 @@
 discard """
   matrix: "--threads:on"
   errormsg: "'spawn' takes a call expression; got: proc (x: uint32) = echo [x]"
+  description: "Poor error message for spawn when a procedure (without calling it) is passed, issues #7535"
```
saem: Good news the parallel category got murdered in `devel`, so you don't have to worry about that one. :D

saem (changes requested):
For descriptions it's definitely just formatting.

For tags if the testament tag support lands separately then if the current ones aren't perfect we can do a quick pass to drop any we don't need and add/change them incremnetally.

Thanks again, this is really good and a big lift!

tests/array/t_array_doesnt_support_typedesc.nim
```diff
@@ -0,0 +1,17 @@
...
+cmd: "nim check $file"
+errormsg: "invalid type: 'typedesc[int]' in this context: 'array[0..0, typedesc[int]]' for var"
+nimout: '''
+tArrayTypeDesc.nim(10, 5) Error: invalid type: 'type' in this context: 'array[0..0, type]' for var
+tArrayTypeDesc.nim(11, 5) Error: invalid type: 'typedesc[int]' in this context: 'array[0..0, typedesc[int]]' for var
+'''
+description: '''
+  . From https://github.com/nim-lang/Nim/pull/13261,
+    https://github.com/nim-lang/Nim/issues/7331,
+    https://github.com/nim-lang/Nim/issues/9932,
```
canelhasmateus: Given the number of issues that reference this behaviour, this should be inside a typedesc spec?

saem:
> This seem like a great test to refactor, using labeled blocks and asserting literal strings instead of using stdout.
> 
> Do you know of an easier way to put these comments in my review, btw?

I think the way to do it is to make an inline comment by clicking on the little plus button that appears when you hover the mouse near the line numbers in a file diff. Then it should correctly associate to the file/line.

saem:
> This seem like a great test to refactor, using labeled blocks and asserting literal strings instead of using stdout.
> 
> Do you know of an easier way to put these comments in my review, btw?

I think the way to do it is to make an inline comment by clicking on the little plus button that appears when you hover the mouse near the line numbers in a file diff. Then it should correctly associate to the file/line.

canelhasmateus:
> > This seem like a great test to refactor, using labeled blocks and asserting literal strings instead of using stdout.
> > Do you know of an easier way to put these comments in my review, btw?
> 
> I think the way to do it is to make an inline comment by clicking on the little plus button that appears when you hover the mouse near the line numbers in a file diff. Then it should correctly associate to the file/line.

What i was refering to is: while developing, there are things that come to mind - "I should comment about this at the pr"
Currently, i need to remember these places and add the comments at the pr. Or make them comments in the code, like i did there. I dislike both options, and was wondering if there is any kind of mechanism to automate this process - Such as a "#todo: ... " that gets flagged and removed.

tests/destructor/tdestructor_manual_invocation.nim
```diff
@@ -0,0 +1,32 @@
+discard """
+  matrix: "--gc:refc; --gc:arc"
+  output: '''
```
canelhasmateus: Didn't manage to remove these echos here. 
Could make =destroy push to a seq , and check its status later ; 
But then, need to make the =destroy get called by the runtime - is this possible by adding scopes?

tests/destructor/tmacro_many_branches.nim
```diff
@@ -0,0 +1,42 @@
...
+macro bigIfElseExpr(): untyped =
+  result = nnkIfExpr.newTree()
+
+  for x in 0 ..< 1000:
+    result.add nnkElifExpr.newTree(newLit(false), newStrLitNode($x))
+
+  result.add nnkElseExpr.newTree(newStrLitNode("other"))
+
+proc test(arg: string): string =
+  doAssert bigIfElseExpr() == "other"
```
canelhasmateus, line 37: this test is odd. Trying to remove the echos, i added that assert.  
It works, but changing the literal string in line 26 to something other than 'other' made it fail.
It seems like the second macro doesn't get tested , either.

tests/destructor/tseq_copy_unpack_empty.nim
```diff
@@ -4,6 +4,16 @@ discard """
...
 '''
+description: '''
+ . From https://github.com/nim-lang/Nim/issues/12037
+   Newruntime: seq copy invalidates original seq
+
+ . From https://github.com/nim-lang/Nim/issues/12820
+   @[] is a problem for --gc:destructors
+
+ . From https://github.com/nim-lang/Nim/issues/12989
+   ARC: Unpacking tuple with seq causes segfault
```
canelhasmateus, line 11: As i understand, half of these tests have nothing to do with destructors

tests/destructor/tdestructor_manual_invocation.nim
zerbina: You could wrap the top-level statements in a function in order to have the destructors called in a deterministic way. That'd be a change in semantics, but it shouldn't interfere with what the test is testing.

> Could make =destroy push to a seq , and check its status later ;

In most cases, that's the way to go 👍

tests/destructor/tmacro_many_branches.nim
zerbina: I tried it locally and changing `"other"` in `bigCaseStmt` worked there.

> It seems like the second macro doesn't get tested , either.

Judging by the issue's description, I believe the test here is more along the lines of "does it compile?" than "does it produce the correct output?".

tests/init/tinit_var_tracking_tuples.nim
```diff
@@ -0,0 +1,28 @@
+discard """
+  nimout: '''
+tInitVarTrackingTuples.nim(14, 7) Hint: BEGIN [User]
```
canelhasmateus: As i understand, this 'nimout' check seems to be useless. Is that right?

tests/exception/texception_message_null_byte.nim
```diff
@@ -1,3 +1,11 @@
+discard """
+description: '''
+  . From https://github.com/nim-lang/Nim/issues/13115
```
canelhasmateus, line 3: this tests seems it uses a define clause to do things. They look like other test files - but i renamed some of them. 
This test still passes - although it is slow - but his makes me a bit worried that something of this nature can happen in other tests. 
w/e, will cross that bridge when it shows up.

tests/init/tinit_var_tracking_tuples.nim
saem: It's making sure there is no intervening/disordered output and it does finish.

There are better ways.

tests/exception/texception_message_null_byte.nim
saem: Just making sure, but the test calls itself with the define in order to execute various combinations. So long as the code is resilient to renamed that's fine. If you haven't already mentally trace the execution to learn more.

In this case it programmatically gets the file name and it's not hardcoded.

tests/destructor/tseq_copy_unpack_empty.nim
saem: Sorta, the tricky thing is that semantics of these operations changed between gc, as the ringtones vary.

For unpacking it's not always destructors but there could allocations/copies in which case destruction and or moves.

tests/array/t_array_doesnt_support_typedesc.nim
saem: Yeah, I think so.

tests/init/tinit_var_tracking_tuples.nim
canelhasmateus: Considering  i removed the stdout checking, is this still relevant?

saem: So given that you took out the stdout checking what are the implications of that, given what I said above?

tests/proc/tproc_untyped_arg.nim
```diff
@@ -6,3 +6,4 @@ discard """
 
 template something(op: proc (v: untyped): void): void =
```
canelhasmateus, line 7: by removing the knownIssues:, i get
```
Expected:
'untyped' is only allowed in templates and macros or magic procs

Gotten:
return type 'untyped' is only valid for macros and templates
```

Is this the message that is supposed to be raised? If so , can we enable again?

saem: No, I don't think that makes sense. Can you check the original bug report?

The expected says the proc param shouldn't be untyped, while the actual is taking about return types.

canelhasmateus: yeah, that is what i gathered from the discussion too.

canelhasmateus:
Ok, basically finished. Now, to bash my head against the ci until it passes.

tests/init/tinit_var_tracking_tuples.nim
```diff
@@ -0,0 +1,28 @@
...
+'''
+description: '''
+  . From https://github.com/nim-lang/Nim/issues/8314
+    Incorrect 'Cannot prove that 'y' is initialized' caused by tuple assignment
+  . If a normal assignment is used instead, the warning disappears.
+  . Fixed by https://github.com/nim-lang/Nim/pull/8321
+'''
+"""
+
+{.hint: "BEGIN".}
```
saem: Not sure why we need these hints?

Likely just need an assert to make sure there is an actual call and test

saem:
Why are there a bunch of lines in the descriptions staying with a '.'?

canelhasmateus:
> Why are there a bunch of lines in the descriptions staying with a '.'?

Just formatting, to indicate the start of a new paragraph.

tests/init/tinit_var_tracking_tuples.nim
canelhasmateus: https://github.com/nim-works/nimskull/pull/327#discussion_r902722393

canelhasmateus:
It seem that the CI reports some errors that i'm unable to reproduce, and from the output of the "print test errors" stage, they seem to be all caused by ''Result: reKnownIssue".
Not sure what is going on - Can anyone shine a light here? @alaviss @saem @zerbina

saem:
> It seem that the CI reports some errors that i'm unable to reproduce, and from the output of the "print test errors" stage, they seem to be all caused by ''Result: reKnownIssue". Not sure what is going on - Can anyone shine a light here? @alaviss @saem @zerbina

First thing I thought of is that you're not rebased on the latest devel. Maybe you found a bug in testament itself.

Can you reproduce it locally by running a single test?

canelhasmateus:
I have rebased on latest devel. 
I can't reproduce it locally : the mentioned failed tests are simply skipped.

I have been "bisecting" the change though, and confirmed that the changes in the "ccggen" category causes CI to fail. 
I'm now trying to minimize this scope, and find out all the 'breaking changes'.
Edit: If you have a better approach to find the roots, feel free to tell me. I profoundly dislike abusing CI time :(

canelhasmateus:
Luckily that was everything. 
Will check later what exactly is going on with the ccggen changes. 
Apart from that, i have largely finished , and don't aim to make further changes in this PR. 

The implementation of tags, alongside the categorization of tests, will be introduced in future (hopefully shorter) PRs - i hate long lived branches.
Are there any un-addressed points you'd like changed?

saem:
> I have rebased on latest devel. I can't reproduce it locally : the mentioned failed tests are simply skipped.
> 
> I have been "bisecting" the change though, and confirmed that the changes in the "ccggen" category causes CI to fail. I'm now trying to minimize this scope, and find out all the 'breaking changes'. Edit: If you have a better approach to find the roots, feel free to tell me. I profoundly dislike abusing CI time :(

Here are some stating the obvious items, just in case:
* when running the test locally, ensure you're specifying both the the C and C++ targets, `./koch.py test --targets:'c cpp' r tests/cat/test.nim`
* also koch.py deals with this, but just in case you still can't reporduce locally, checking to make sure you have a recent build via a boostrap would rule that out

Other bits:
* could you list out the tests in a comment here that are troublemakers so it's less spelunking for folks trying to drive by help
* these are likely due to cpp related breakages because we removed the hack in sem

tests/arc/tarc_import_inline_iterators.nim
```diff
@@ -0,0 +1,12 @@
+discard """
+  cmd: "nim c --gc:arc $file"
```
saem, line 2: Can we convert these to use a single entry `matrix`?

(can be in a follow-up PR)

tests/concepts/t3330.nim
```diff
@@ -1,5 +1,5 @@
 discard """
-description: "explain reporting isn't working during the transition to nkError; revise and ressurrect once fixed"
+description: "From https://github.com/nim-lang/Nim/issues/3330"
```
saem, line 2: I guess that's better than before. 😆

tests/init/tinit_var_tracking_tuples.nim
saem: I still don't get it, but we can figure that out of band. 😆

tests/js/tjs_reset_nested_table.nim
```diff
@@ -0,0 +1,13 @@
+discard """
+description: '''
+  . From https://github.com/nim-lang/Nim/issues/11697
+    JS backend crashes when trying to reset a Table inside a Table
```
saem, line 4: xzibit.gif

tests/parser/tparse_confusing_identation.nim
```diff
@@ -0,0 +1,62 @@
+discard """
+  description: " Confusing 'invalid indentation' errors reported when indentation is not at fault, issues #15667"
```
saem: convert to a URL please.

saem (approved):
I found one missed url, but that's it.

tests/concepts/t3330.nim
canelhasmateus: didn't have the courage to rename due to the implications on changing nimout. This is passing, btw

tests/init/tinit_var_tracking_tuples.nim
canelhasmateus: I removed the hint and added another doAssert based on the output of the function, just to be sure.

tests/arc/tarc_import_inline_iterators.nim
canelhasmateus: Absolutely. 
I have the intention of going over every remaining test, adding proper descriptions, tags and gathering what can / should be moved to the specs. This can be done in the same effort.

How would this specific entry be expressed as a matrix? Does this extend to every other use of "cmd:"?

saem: Yay, awesome!

going from memory, but I believe this is it: `matrix: "--gc:arc"`

Yeah, most `cmd` can be converted to this, but not all.

saem (approved):
bors r+

yay!

this was a big lift, thanks, @canelhasmateus

Issue #328, opened 2022-05-28
Wrong behaviour at runtime with templates and `var`/`lent` tuples


zerbina:
Wrapping an expression of `var` or `lent` tuple type `T` in a template who's return type is `T` directly (without `var` or `lent`), silently leads to wrong runtime behaviour with the C and C++ targets. Both `refc` and `arc`/`orc` are affected

### Example

* commit: https://github.com/nim-works/nimskull/commit/e41a50fbe1ebd0ebc92b2337aed398209ba0daf5
* command line: `nim c -r`

```nim
type Tup = tuple
  a: int

func get(x: Tup): lent Tup = x

template getT(p: Tup): Tup = get(p)

block:
  let t = (a: 1)
  let v = getT(t)
  doAssert v[0] == 1
```

### Actual Output

The assertion fails

### Expected Output

Success

### Additional Information

Printing the AST for `getT(t)`after `transf` yields:
```
HiddenSubConv # type: Tup
  Empty
  Call # type: lent Tup
    Sym "get"
    HiddenDeref
      Sym "x"
```

The `HiddenSubConv` is likely injected here:
https://github.com/nim-works/nimskull/blob/e41a50fbe1ebd0ebc92b2337aed398209ba0daf5/compiler/sem/sigmatch.nim#L2157-L2162

`cgen` then handles the `HiddenSubConv` here:
https://github.com/nim-works/nimskull/blob/e41a50fbe1ebd0ebc92b2337aed398209ba0daf5/compiler/backend/ccgexprs.nim#L2286-L2291

Since `Tup` and `lent Tup` are not the same backend type, `genSomeCast` is used, producing the following C code (shortened for readability):
```c
Tup* T1_;
T1_ = (Tup*)0;
T1_ = get__tissue_5((&t));
v = (*(Tup*) ((&T1_))); // <- issue: Tup** is reinterpreted as Tup*
```

Changing the return type of `getT` to `untyped` makes the example work. No `HiddenSubConv` is injected around the call.

zerbina:
Fixed by #1006, but the issue wasn't auto-closed.

PR #329, opened 2022-05-28
fix: make `Result.value` work with tuple values

zerbina:
Work around a compiler bug by changing the return type of the
`value` template to `untyped`.

The bug in question is: https://github.com/nim-works/nimskull/issues/328

### DIFF
--- a/lib/experimental/results.nim
+++ b/lib/experimental/results.nim
@@ -200,8 +200,11 @@ func `$`*[E](self: Result[void, E]): string =
   if self.isOk: "Ok()"
   else: "Err(" & $self.e & ")"
 
-template value*[T, E](self: Result[T, E]): T = get self
-template value*[T, E](self: var Result[T, E]): T = get self
+# XXX: the `value` templates only use `untyped` in order to work around a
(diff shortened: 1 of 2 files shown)


lib/experimental/results.nim
```diff
@@ -200,8 +200,8 @@ func `$`*[E](self: Result[void, E]): string =
   if self.isOk: "Ok()"
   else: "Err(" & $self.e & ")"
 
-template value*[T, E](self: Result[T, E]): T = get self
-template value*[T, E](self: var Result[T, E]): T = get self
+template value*[T, E](self: Result[T, E]): untyped = get self
```
saem, line 206: can you leave a comment here to indicate that this should become typed once the bug is fixed?

zerbina: With typed, do you mean `lent T` and `var T`? Or just `T` as was previously?

saem: I meant `T`

zerbina:
I added a comment regarding the change in the code.

bors r+

PR #330, opened 2022-05-28
vm: replace custom result-like types with `Result`

zerbina:
`VmGenResult` and `ExecutionResult` are both kept as aliases.

In addition:
- the handling of the result values is simplified/cleaned up a bit
- errors are always wrapped in `nkError` nodes and no longer reported
 during result unpacking

### Future work:
- `evalStaticStmt`  and `setupCompileTimeVar` need to further propagate
 the errors instead of `localReport`-ing them

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -722,12 +722,12 @@ proc opConv(c: var TCtx; dest: var TFullReg, src: TFullReg, dt, st: (PType, PVmT
     else:
       asgnValue(c.memory, dest, src)
 
-proc compile(c: var TCtx, s: PSym): int =
-  let r = vmgen.genProc(c, s)
-  if unlikely(not r.success):
-    raiseVmError(r.report)
+func toException(r: sink SemReport): ref VmError {.noinline.} =
+  new(result)
+  result.report = r
 
-  result = r.start
+proc compile(c: var TCtx, s: PSym): int =
+  result = vmgen.genProc(c, s).value()
 
   when debugEchoCode:
     c.codeListing(s, nil, start = result)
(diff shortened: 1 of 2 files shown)


compiler/vm/vm.nim
```diff
@@ -3319,51 +3304,49 @@ proc execute(c: var TCtx, start: int, frame: sink TStackFrame; cb: proc(c: TCtx,
...
-  ## new `nkError` wrapping `node`. Otherwise, sets `start` to the instruction
-  ## offset given by the result and returns `nil`
-  if r.success:
-    result = nil
-    start = r.start
+template returnOnErr(res: VmGenResult, config: ConfigRef, node: PNode): int =
+  ## Unpacks the vmgen result. If the result represents an error, exits the
+  ## calling function by either returning a new `nkError` wrapping `node`
+  ## or, if the calling function has no return value, via
+  ## `handleReport`+`return`.
```
saem: Can we not use an nkError in both cases except one is an empty?

compiler/vm/vm.nim
```diff
@@ -3319,51 +3304,49 @@ proc execute(c: var TCtx, start: int, frame: sink TStackFrame; cb: proc(c: TCtx,
...
-  ## directly (via `handleReport`)
-  if res.success:
-    result = res.result
+  if res.isOk:
+    result = res.take
     if node != nil and result.info.line < 0:
       result.info = node.info
   else:
-    let errKind = res.report.kind
+    let err = res.takeErr
```
saem: Nonblocking style remark: as you go can you favour let and var blocks?

saem (commented):
This is super exciting!

We're on the cusp of being able to compile and instead of the current sem approach where we get a boolean, we can instead get the AST that worked out is an nkError!

compiler/vm/vm.nim
zerbina: `evalStmt` now also return a `PNode`, making the special `handleReport`+`return` logic here unnecessary.

zerbina:
Fixed the failing tests (the VM stack-trace wasn't printed anymore) and also addressed the style remarks.

*Edit:* @saem If the tests pass (they should) and you're okay with the adjustments, feel free to bors already

saem:
heck yeah, bors away! :D

PR #332, opened 2022-05-29
Initial implementation of the newstyle error reporting

haxscramper:
I managed to get some free time to write a proof-of-concept implementation of the cleaned-up error formatting implementation using colored text messages and reporting. 

For some reason, it took me quite a while to realize it is not mandatory to have a full conversion of the test suite (to structured error messages) before I can start working on the formatted version. At the start, it is quite possible to work on the newstyle errors (`--errorformat=new`) while keeping old implementation in place.

Right now, function call mismatch messages are implemented partially, and I added contextual information for the generic procedure instantiation context messages.

This is a draft PR - I will continue working on it when I have enough time to dive into it, but if anyone is interested in picking it up - you are more than welcome, so don't hesitate to ask for directions or suggestions.

before:

image

after (proof-of-concept, so formatting and representation is guaranteed to change):

image

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -431,6 +431,21 @@ proc getStr*(a: PNode): string =
     #internalError(a.info, "getStr")
     #result = ""
 
+
+func getIdentStr*(s: PSym): string = s.name.s
+func getIdentStr*(i: PIdent): string = i.s
+
+proc getIdentStr*(n: PNode): string =
+  ## Get string from identifier or symbol node. Raise recoverable error for
+  ## all other node kinds
+  case n.kind:
+    of nkIdent: result = n.ident.s
+    of nkSym: result = n.sym.name.s
+    else:
+      raiseRecoverableError(
+        "Cannot extract 'identifier' string from node of kind " & $n.kind)
+
+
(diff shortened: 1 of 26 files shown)


haxscramper:
something something readability

image

image

although with more compact error messages formatting becomes trickier, because you need to think about placement, readability and maybe missing information.

haxscramper:
https://github.com/zesterer/ariadne should be ported for code annotation generation, we have almost all parts of the infrastructure to easily integrate it (colored runes for errors, enum for message generation, code formatting solution with range outputs)

haxscramper:
The idea is relatively sound, but the number of conflicts and the pace of cleanup for the nodes suggest I should close this PR for the time being -- in any scenario it would be easier to start based on the improved code rather than try to rebase this PR. For the ariadne rewrite I mentioned -- I had rewritten this to the C++ (thanks to the ChatGPT assistance, otherwise I'm almost 100% sure it would've been impossible), but there are still a lot of places where original Rust code is hard to understand. Also some pieces of logic need to be rewritten and cleaned up before this can be moved to nimskull.

https://github.com/haxscramper/haxorg/blob/master/src/sem/ErrorWrite.hpp the C++ code is here.

PR #336, opened 2022-06-06
experimental: removed of spawn and parallel

saem:
## Summary
* remove parallel and spawn experiemental feature
* it creates a cyclic dependency between sem and the backend
* removed associated tests that relied on these features

---
## Notes for Reviewers
*  not marked as draft so I can see what happens in CI
* removal of `nkParForStmt` likely requires a csources reset
* other than `nkParForStmt` let me know if I missed something


CyberTailor:
What to use instead?

saem:
@CyberTailor currently there isn't a "blessed" alternative that we've put into place. For further discussion please use the chat.

PR #337, opened 2022-06-06
Remove nkParForStmt node

Clyybber:
## Summary
Removes nkParForStmt node which is unused since https://github.com/nim-works/nimskull/commit/60d96a55953b8d2778e3d461858f515e45369c4a

## Details
Bootstrapping is made to work by changing the ordinal mapping of node kinds in lib/macros depending on wether we compile with a compiler that has or doesn't have nkParForStmt, similar to https://github.com/nim-works/nimskull/commit/63bea9dcda428b47ca0885d660e13c980a9c1f10

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -3080,7 +3080,6 @@ proc expr(p: BProc, n: PNode, d: var TLoc) =
           # by ensuring it's no inner proc (owner is a module).
           # Generate proc even if empty body, bugfix #11651.
           genProc(p.module, prc)
-  of nkParForStmt: genParForStmt(p, n)
   of nkState: genState(p, n)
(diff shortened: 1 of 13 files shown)


compiler/backend/ccgexprs.nim
```diff
@@ -3080,7 +3080,6 @@ proc expr(p: BProc, n: PNode, d: var TLoc) =
           # by ensuring it's no inner proc (owner is a module).
           # Generate proc even if empty body, bugfix #11651.
           genProc(p.module, prc)
-  of nkParForStmt: genParForStmt(p, n)
```
saem, line 3083: Dead code, can remove `genParFor` in a subsequent PR.

saem (approved):
Bors r+

PR #339, opened 2022-06-09
vm: rename `FunctionTypeId` to `RoutineSigId`

zerbina:
The old name was confusing and didn't communicate the purpose/usage
very well. `RoutineSigId` better reflects the type's purpose.

Field and function names at `RoutineSigId` usage sites are also adjusted

### Misc:
- documentation improvements
- a redundant table lookup in `makeSignatureId` is eliminated


saem (approved):
This reads so much better!

Bors r+

PR #341, opened 2022-06-11
vm: improve `unreachable` template

zerbina:
## Summary

- add an optimized overload for enum values (they're often used with
 `unreachable`)
- use the `enum` overload where applicable
- move the `raise` statement and message formatting code into
 a `.noinline` function in order to improve code size

## Details

Since the `raise` is not inside a template anymore, the original line
information is also added to the message in order to still have some
context when stact-traces are disabled

---

The `unreachable` helper template is not strictly related to only the VM and is also useful in general, so maybe it should be moved somewhere else?


saem (approved):
I don't have a suggestion for a better place to put it. I would maybe leave a comment for someone down the road to consider it, but it's good by me.

zerbina:
Left a comment in the code.

bors r+

PR #343, opened 2022-06-12
More context in `undeclared identifier` error messages

zerbina:
## Summary
- add back hint about present recursive imports
- add corresponding error message test
- use `createUndeclaredIdentifierError` in place of manual
 re-implementation

### DIFF
--- a/compiler/ast/reports.nim
+++ b/compiler/ast/reports.nim
@@ -268,11 +268,11 @@ type
       of rsemExprHasNoAddress:
         isUnsafeAddr*: bool
 
-      of rsemUndeclaredIdentifier,
-         rsemCallNotAProcOrField,
-           :
-        potentiallyRecursive*: bool
+      of rsemUndeclaredIdentifier:
+        recursiveDep*: seq[tuple[importer, importee: string]] ## The list of
+        ## present recursive module imports
 
+      of rsemCallNotAProcOrField:
         explicitCall*: bool ## Whether `rsemCallNotAProcOrField` error was
         ## caused by expression with explicit dot call: `obj.cal()`
         unexpectedCandidate*: seq[PSym] ## Symbols that are syntactically
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1973,6 +1973,14 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
 
     of rsemUndeclaredIdentifier:
       result = "undeclared identifier: '" & r.str & "'"
+      if r.recursiveDep.len > 0:
+        result.add "\nThis might be caused by a recursive module dependency:"
+        result.add "\n "
+        for dep in r.recursiveDep.items:
+          result.addf("'$1' -> ", dep.importer)
+
+        result.addf("'$1'", r.recursiveDep[^1].importee)
+
       if r.spellingCandidates.len > 0:
         result.add "\n"
         result.add presentSpellingCandidates(
--- a/compiler/sem/lookups.nim
+++ b/compiler/sem/lookups.nim
@@ -562,7 +562,7 @@ proc errorUndeclaredIdentifier*(
     kind: rsemUndeclaredIdentifier,
     str: name,
     spellingCandidates: candidates,
-    potentiallyRecursive: c.recursiveDep.len > 0
+    recursiveDep: c.recursiveDep
   ))
 
   if c.recursiveDep.len > 0:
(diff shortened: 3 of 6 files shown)


haxscramper:
`recursiveDep` should probably be stored directly in the `SemReport` error and checked for `0 < .len` in the formatting, this would allow `"A -> B -> C -> A"` elaboration in the error message later on. Otherwise, I see no issues with this change.

haxscramper (approved):

saem:
Do we already have a test for that particular error path?

Aside for later: need to think about how to make it easier to see the test vs error coverage.

zerbina:
> Do we already have a test for that particular error path?

Skimming over the tests, it doesn't look like it. I'm going to add one as part of this PR.

I only noticed the missing hint due to an `undeclared identifier` error happening because of the cyclic import dependency between `transf` and `lambdalifting`

haxscramper:
> need to think about how to make it easier to see the test vs error coverage.

All errors go through a single callback, so you can copy the output into a secondary channel and then collect results after the whole tests suite is run. Error branch coverage (for changes like in this PR - where error might be different) we can get information about all error instantiations at compile-time (IIRC it also does go through a single template, or a few of those) and compare it with actually triggered errors (report stores information about its instantiation location).

zerbina:
`recursiveDep` is now stored as part of the report and also printed during reporting. In addition, I added a test for the error message.

The test fails right now, since the hint about the recursive import dependency is missing. From a cursory look into the issue, it seems like a `rsemUndeclaredIdentifier` error is constructed twice, with only the second one being reported. Since each construction of a `rsemUndeclaredIdentifier` currently consumes the active `recursiveDep` of `PContext`, the second error gets an empty `recursiveDep`.

I'm rather busy with all the VM related work and currently don't have the bandwidth to further look into this, so I converted the PR to a draft for now. If someone else wants to take over, they're welcome to do so.

saem:
I'll try and look on this weekend, my dance card is pretty full, too. Unless Hax beats me to it.

saem:
My currently terrible idea as to how to fix this would be to go into `sempass2.track` and for the `nkError` branch in the case statement do the reset of `TContext.recursiveDep` there, instead of in `lookups`.

Here is a snippet lying around in a branch as I'm futzing around with cleaning up `importer` and removing `localReport` in lieu of nkError.

It's ok if no one is enthused by this approach.

```
# around line 1322 in sempass2
  of nkError:
    let report = tracked.config.m.reports.getReport(n.reportId)
    if report.category == repSem and
      report.semReport.kind == rsemUndeclaredIdentifier:
      # xxx: recursiveDep are tracked for better error reporting for undeclared
      #      identifiers, but until we report the error we can't clear this
      #      data. This isn't the ideal place for it, but it's not too far off
      #      wrt to lifetimes. Errors are sometimes more eagerly reported, but
      #      that's more a temporary issue and they should all pass through
      #      here eventually.
      if tracked.c.recursiveDep.len > 0:
        # prevent excessive errors for 'nim check'
        tracked.c.recursiveDep.setLen 0
    
    localReport(tracked.config, n)
```

zerbina:
Hmm, I guess the missing hint in the error message is not too critical for now. Maybe this PR should simply be postponed until further `sem` refactorings happened? If not, I'm also okay with @saem's proposal.

---

I continued to think about this for a bit and here's an idea I came up with:

Introduce a `ModuleReportContext` (placeholder name) that stores the collected recursive dependencies. Each module that's part of the compilation gets it's own instance of this context. Whenever a cyclic import is detected, it's appended to a `recursiveDeps` seq in the `ModuleReportContext` of the module in which the import happened.

If an `undeclared identifier` report is constructed, instead of storing the recursive dependency information in the report directly, the current length of the context's `recursiveDeps` together with the current module's ID is stored in the report.

When processing the report, the `ModuleReportContext` is looked up via the module ID stored in the report and only the recursive dependencies stored in `recursiveDeps[0..<activeLen]` are included in the message.

This removes the need to reset the recursive dependency information after report construction; makes `TContext.recursiveDep` unnecessary; and also provides more context for the message (all active cyclic imports instead of just the latest one)

saem:
We know if there is a recursion problem in `myImportModule` but for reasons that I don't understand we don't act on it... that seems goofy. Am I missing something?

In any case, I think what you're saying is in effect to store each recursive deps independently then we can reference the latest one per error. Like a stack of stacks.

zerbina:
> We know if there is a recursion problem in `myImportModule` but for reasons that I don't understand we don't act on it... that seems goofy. Am I missing something?

Hm, how do we know if there's a recursion problem in `myImportModule`? Cyclic imports are allowed and when importing module `a` in `b` (where `a` also imports `b`), it's not know on import whether `b` uses a symbol from `a` that wasn't sem-checked yet.

> In any case, I think what you're saying is in effect to store each recursive deps independently then we can reference the latest one per error. Like a stack of stacks.

Yes, sounds good.

zerbina:
This PR is superseded by #505.

PR #344, opened 2022-06-13
sem: clean-up TContext docs, readers, and writers

saem:
## Summary
* documented each field for TContext, purpose, writers, and consumers
* docs to understand the compiler internals and what needs fixing
* remove dead code, fix bugs/issues in writers/consumer of the data

## Details
* removed dead fields: evalContext, unknownIdents, symMapping, etc
* documented refactor strategy to allow this to be a longer lived refactor

---

## Notes for Reviewers
* the comment blocks are kinda ugly/chunky right now to first see the mess
* the idea with the above is after seeing the mess actually clean-up starts
* this will likely be a long running refactor
* after each field is documented, will figure out which clean-ups to include

### DIFF
--- a/compiler/sem/semdata.nim
+++ b/compiler/sem/semdata.nim
@@ -91,81 +91,595 @@ type
       exceptSet*: IntSet         # of PIdent.id
 
   PContext* = ref TContext
+
+  # Refactoring of `TContext` in progress:
+  # --------------------------------------
+  #
+  # TContext is used for a lot of stuff that it probably shouldn't be. In terms
+  # of compiler architecture this is post parsing, so we're taking parsed code
+  # and evaluating it. The result of this evaluation should be a value, an
+  # error, or code that can be compiled (think of the exe as a partially
+  # evaluated function, waiting for params to finish and become a value).
+  #
+  # So we're analysing, interpreting, and compiling, etc... that means we'll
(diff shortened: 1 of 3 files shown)


compiler/sem/semdata.nim
```diff
@@ -91,39 +91,339 @@ type
...
+    # end: caches/common defs, likely whole program, per modules sandboxing?
+    # -------------------------------------------------------------------------
+
+
+    # -------------------------------------------------------------------------
+    # start: not entirely clear why, function pionters for certain sem calls?
+    # -------------------------------------------------------------------------
     semConstExpr*: proc (c: PContext, n: PNode): PNode {.nimcall.} # for the pragmas
     semExpr*: proc (c: PContext, n: PNode, flags: TExprFlags = {}): PNode {.nimcall.}
     semTryExpr*: proc (c: PContext, n: PNode, flags: TExprFlags = {}): PNode {.nimcall.}
```
zerbina: I believe this is done as a way to make various modules (e.g. `pragmas`) importable (without import cycles) instead of having to use `include`

saem: Ah thanks, I knew this fact at some point and then forgot while doing this work.

Aside: English hilarity, I first read the "I'm" in "importable" as it is in the word "impossible". Took a second to parse your comment as intended. 😄

PR #346, opened 2022-06-13
testament: enable logging and pass it to runners

alaviss:
## Summary
Re-enable logging by default, as it was disabled (probably a mistake)
by the refactoring.

Also forward the logging switch to runners spawned from testament.

## Details
* This is required by `ci_testresults` to function, and is also required by `--failing` and friends
* The feature is supposed to be enabled by default (according to cmdline docs)


saem:
Whoops and thanks!

Bors r+

saem (approved):

PR #347, opened 2022-06-13
compiler: add a first-class VM back-end

zerbina:
## Summary
- the corresponding compiler commands are `vm` and `compiletovm`
- the backend option is `--backend:vm`
- the produced executable uses the `.nimbc` (Nim Bytecode) file
 extension. They're executed via the standalone `vmrunner` program and
 have no run-time dependencies on the source-code they were built from
- testing for the VM target can be done via `defined(vm)`

All alive code is processed and translated to byte-code during the build
step; the runner doesn't make use of just-in-time code-generation.

Having a dedicated VM back-end makes it easier to grow test coverage
for the VM and it's corresponding code-generator (`vmgen`). With the
`vm` testament target, running tests in the VM that don't depend on
being run at compile-time can now be done without the
`template code = ...; static: code; code` pattern.

## Details
- adjust `vmgen` to support non-compile-time code
- include the `vm` target in the default targets for the 'lang'
 test category
- add a test for module initialization order
- the `vmrunner` program is built by `koch tools` and used by
 `nim r` and testament
- the `passes` interface is not used for code-generation; only for
 collecting the active modules and their top-level statements.
 `vmbackend.generateCode` is responsible for running code-generation
 and writing the output executable file
- when building for the VM target, the `when nimvm` compile-time branch
 is also taken in the executable. This is meant as transition helper,
 but the plan is to make it behave like it does on the other
 targets later on
- dead-code-elimination uses `vmgen`'s existing dependency
 collection facility
- a `.nimbc` file is a `PackedEnv` stored in the existing `rodfiles`
 format
- `PackedEnv` is a packed and easily serialize-able representation of
 all static VM environment state necessary for running the program,
 plus the program's byte-code

---

## Notes for reviewers
- I'm not sure if the doc-comment quality is good enough. They likely need some extra attention

### DIFF
--- a/compiler/front/in_options.nim
+++ b/compiler/front/in_options.nim
@@ -139,6 +139,7 @@ type
     backendCpp = "cpp"
     backendJs = "js"
     backendObjc = "objc"
+    backendNimVm = "vm"
     # backendNimscript = "nimscript" # this could actually work
     # backendLlvm = "llvm" # probably not well supported; was cmdCompileToLLVM
 
@@ -150,6 +151,7 @@ type
     cmdCompileToCpp
     cmdCompileToOC
     cmdCompileToJS
+    cmdCompileToVM
     cmdCrun        ## compile and run in nimache
     cmdTcc         ## run the project via TCC backend
     cmdCheck       ## semantic checking for whole project
--- a/compiler/vm/packed_env.nim
+++ b/compiler/vm/packed_env.nim
@@ -1,0 +1,965 @@
+## This module contains the type definitions for a packed representation of
+## all the state required for loading and running a standalone VM instance.
+##
+## In addition, procedures for storing and loading the original objects to and
+## from the packed environment object are also located here.
+##
+## For reading and writing `PackedEnv` to files, a ``rodfile``-based solution
+## is provided.
+##
+## First collecting/packing into a data structure that is then written to disk
+## in a separate is not as resource efficient (both in terms of memory
+## consumption and time) as doing both steps in a combined manner. The first
+## (and current) approach is less complex however, which is why it's chosen
+## over the latter.
+
+import
+  std/[
+    hashes,
+    tables
+  ],
+  compiler/ast/[
+    ast_query,
+    ast_types
+  ],
+  compiler/ic/[
+    bitabs,
+    rodfiles
+  ],
+  compiler/utils/[
+    pathutils # for `AbsoluteFile`
+  ],
+  compiler/vm/[
+    vmdef
+  ]
+
+from compiler/vm/vmaux import findRecCase, findMatchingBranch
+from compiler/vm/vmobjects import `$`, packDiscr
+from compiler/vm/vmtypes import alignedSize
+
+export RodFileError
+
+type
+  PackedDataKind* = enum
+    pdkInt
+    pdkIntLit ## an embedded literal
+    pdkFloat
+    pdkString
+    pdkPtr # also includes ref
+    pdkObj
+    pdkArray
+    pdkSet
+    pdkField
+
+  # XXX: `PackedDataNode` is memory-usage wise quite inefficient. Not only at
+  #      the layout level (3 out of it's 8 bytes are wasted for padding), but
+  #      also for representing data in general. For example: `array[256, byte]`
+  #      requires 2048 bytes!
+  #      A possible different approach: store the data in it's
+  #      raw-byte-serialized representation, with additional metadata
+  #      describing where strings and seqs (i.e. values with non-automatic
+  #      storage) should be placed. Maybe also only store the non-zero parts
+  PackedDataNode* = object
+    ## A single node in a depth-first linear tree that is meant for storing
+    ## complex data. If a node has children, they follow directly after their
+    ## parent node. `PackedDataNode` can be seen as a specialized
+    ## version of `PackedNodeLite`
+    kind*: PackedDataKind
+    pos*: uint32 ## for {pdkInt, pdkFloat, pdkString}: the respective `LitId`
+                 ## for {pdkObj, pdkArray, pdkSet}: the number of children
+                 ## for pdkPtr: `0` if nil, `1` otherwise
+                 ## for `pdkField`: the field's position
+                 ## for `pdkIntLit`: a direct literal that fits into 4 byte
+
+  # PackedSymLite, PackedNodeLite, etc. are similiar to the types in
+  # ``packed_ast.nim``, with the difference that the types here only store the
+  # data necessary for their use-case (they're used to provide type
+  # information for `repr`) and are thus more compact.
+
+  SymId = distinct int32
+  TypeId = distinct int32
+  NodeId = distinct int32
+
+  PackedSymLite = object
+    kind: TSymKind
+    magic: TMagic
+    name: LitId
+    typ: TypeId
+    ast: NodeId
+    flags: TSymFlags
+    position: int
+    offset: int
+
+  PackedTypeLite = object
+    kind: TTypeKind
+    callConv: TCallingConvention
+    flags: TTypeFlags
+    types: seq[TypeId]
+    n: NodeId
+    sym: SymId
+    #typeInst*: TypeId
+
+  PackedNodeLite = object
+    kind: TNodeKind
+    flags: TNodeFlags
+    operand: int32  # for kind in {nkSym}: SymId
+                    # for kind in {nkStrLit, nkIdent, nkNumberLit}: LitId
+                    # for non-atom kinds: the number of children
+    typeId: TypeId
+
+  SliceListType* = BiggestInt|BiggestFloat|ConstantId
+
+  PackedVmType* = object
+    # XXX: inefficient storage. For statically sized atoms, `size` and `align`
+    #      are redundant, wasting space. `VmType` has the same problem
+    size*: uint32
+    align*: uint8
+    kind*: AtomKind
+    numFields*: uint16
+
+  # XXX: the rodfile interface for `BiTable`s stores both `keys` and `values`,
+  #      even though `keys` can be reconstructed during loading (at least in
+  #      our use-case here). This makes the resulting file larger than
+  #      necessary
+
+  PackedEnv* = object
+    strings*: BiTable[string]
+    numbers*: BiTable[BiggestInt] # also includes floats
+
+    files*: seq[string]
+    infos*: BiTable[TLineInfo]
+    dbgSyms*: seq[tuple[name, info: LitId]] # for debugging only
+
+    nodes*: seq[PackedDataNode] ## complex data. Currently used for storing
+                                ## complex constants
+    consts*: seq[(ConstantKind, uint32)]
+    cconsts*: seq[tuple[typ: VmTypeId, packedId: uint32]] ##
+      ## Packed `TCtx.complexConsts`. The constants type together with an
+      ## index referencing a sub-tree in `nodes`
+
+    tfields*: seq[tuple[offset: uint32, typId: uint32]]
+    tbranches*: seq[BranchListEntry]
+    types*: seq[PackedVmType]
+
+    globals*: seq[VmTypeId] ## All globals. Only their types are stored
+    functions*: seq[tuple[sym: uint32, sig: RoutineSigId, t1, t2: VmTypeId, kind: CallableKind, a, b: uint32]]
+    callbacks*: seq[string]
+
+    code*: seq[TInstr]
+    debug*: seq[uint32] # Packed version of `TCtx.debug`. Indices into `infos`
+
+    # rtti related data:
+    nimNodes: seq[PackedNodeLite]
+    nimSyms: seq[PackedSymLite]
+    nimTypes: seq[PackedTypeLite]
+    typeInfos: seq[tuple[nt: TypeId, t: VmTypeId]] ##
+      ## Packed version of `TCtx.rtti`
+
+    entryPoint*: FunctionIndex
+
+  TypeInfoDecoder = object
+    types: seq[PType]
+    syms: seq[PSym]
+    nodes: seq[PNode]
+
+  TypeInfoEncoder = object
+    types: Table[ItemId, TypeId]
+    syms: Table[ItemId, SymId]
+
+    pendingTypes: seq[(PType, TypeId)]
+    pendingSyms: seq[(PSym, SymId)]
+
+  PackedEncoder* = object
+    typeInfoEnc: TypeInfoEncoder
+
+    # XXX: `typeMap` is only necessary until `PVmType` gets replaced
+    #      by `VmTypeId`
+    typeMap*: Table[PVmType, VmTypeId] ## Maps each type instance to it's ID
+
+  DataEncoder* = object
+    ## Contextual state needed for turning data `PNode`-trees into
+    ## `PackedDataNode` trees and storing them into packed environment
+    routineSymLookup*: #[lent]# Table[int, LinkIndex] ##
+      ## symbol-id -> routine's link-index. Only read, not mutated
+    i: int ## the index in `PackedEnv.nodes` where the next item is to be stored
+
+const
+  EmbeddedUInts = {nkCharLit, nkUInt8Lit..nkUInt32Lit}
+  EmbeddedInts = {nkInt8Lit..nkInt32Lit} # these also fit into a `uint32`
+  ExternalInts = {nkIntLit, nkInt64Lit, nkUIntLit, nkUInt64Lit}
+
+  NilSymId = -1.SymId
+  NilTypeId = -1.TypeId
+
+
+# -------- general utilities ---------------------------------------------
+
+func growBy[T](x: var seq[T], n: Natural) {.inline.} =
+  x.setLen(x.len + n)
+
+iterator lpairs[T](x: openArray[T]): tuple[key: int, val: lent T] =
+  ## Custom `pairs` iterator that uses `lent`. As of this comment, the stdlib
+  ## one doesn't
+  var i = 0
+  let L = x.len
+  while i < L:
+    yield (i, x[i])
+    inc i
+
+template mapList*[D, S](d: seq[D], s: openArray[S], it: untyped, code) =
+  ## `s` and `d` get evaluated multiple times, so beware
+  bind lpairs
+  d.newSeq(s.len)
+  for i, it {.inject.} in lpairs(s):
+    d[i] = code
+
+# -------- accessors and shortcuts ---------------------------------------
+
+func hash(x: PVmType): int {.inline.} =
+  # Don't use `x` directly, as the lower 2-3 bits are always empty (due to
+  # the 4 or 8 byte alignment of `VmType`)
+  hash(cast[int](x))
+
+func hash(x: TypeId): Hash {.borrow.}
+func `==`(a, b: TypeId): bool {.borrow.}
+
+func hash(x: SymId): Hash {.borrow.}
+func `==`(a, b: SymId): bool {.borrow.}
+
+template genAccessors(t: type, f: untyped, idTyp: type) {.dirty.} =
+  template `[]`(self: t, id: idTyp): untyped =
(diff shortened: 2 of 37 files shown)


zerbina:
I rebased the PR onto `devel`, since most of the commits that could be split off are merged there now. This should already make review a bit easier.

saem:
Giving it anotger read.

I'm letting my comments to a minimum of at all possible to not end up blocked.

saem:
> * What is the future of NimScript and `when nimvm`?

How the nimvm analysis is handled is a terrible hack and needs fixing, but doesn't have to be now.

I'd like a NimScript-like thing where we can use the language for scripting, but I'm not married to the implementation and/or current philosophy. Whatever we do eval code plus a sandbox where we can deny effects (io, time, etc) and access is super handy. Doesn't need to be solved now.

> How and what should be their interactions with the VM back-end?

VM should be what gets used by the complier to make nimvm happen and the scripting engine to feed code to run in a sandbox (they'll need to work together for the sandboxing to be practical).

> While `when nimvm` could previously be used to detect if some code is run at compile-time (or, in the case of NimScript, configuration time), this breaks apart for the VM target.

It's not really the VM's problem, a protected read only environment and code region might be all that's needed in order for nimscript or the compiler to say what mode one is in.

> * Should the binaries be redistributable and if yes, to what extent (version compatibility, cross-compilation, etc.)?

Not now, but I'd like to be greedy and want all that on a very long long timeline.

> The runner currently assumes that the binary comes straight from the compiler which was built with the same compiler source as the runner itself.

Good.

compiler/front/in_options.nim
```diff
@@ -139,6 +139,7 @@ type
     backendCpp = "cpp"
     backendJs = "js"
     backendObjc = "objc"
+    backendNimVm = "vm" # XXX: NimskullVm instead?
```
saem: I'm almost entirely settled on `cyo` as the new name, BTW. Calling it nimskull might make it easier to search, though. No change expected on my end.

compiler/vm/packed_env.nim
```diff
@@ -0,0 +1,945 @@
...
+  ## data (except `nodes`, `numbers`, and `strings`) is thrown away. The only
+  ## parts of `dst` not touched by `storeEnv` are: `cconsts`, `globals`,
+  ## `code`, and `entryPoint`. These have to be filled in separately.
+  # store all types:
+  dst.types.newSeq(c.types.len - 1) # -1 since the invalid type is not included
+  for i in 1..<c.types.len:
+    let t = c.types[i]
+    dst.types[i - 1] = storeVmType(enc, dst, t)
+
+  # TODO: maybe make `dst.consts` a seq of indices into `dst.nodes` instead?
```
saem: Maybe, I think the real savings might come in when array data can be more densely stored in nodes and then const can leverage that as well. Otherwise the separation might have some benefits for now.

compiler/vm/packed_env.nim
```diff
@@ -0,0 +1,945 @@
...
+  #       have to perform a manual copy instead of a move here
+  mapList(dst.callbacks, c.callbackKeys, c):
+    c.string
+
+  mapList(dst.files, c.config.m.fileInfos, fi):
+    fi.fullPath.string
+
+  storeTypeInfos(enc, dst, c.rtti)
+
+proc writeToFile*(p: PackedEnv, file: AbsoluteFile): RodFileError =
```
saem, line 877: Tests and/or template might help here to ensure the symmetry.

compiler/vm/vmaux.nim
```diff
@@ -119,8 +118,8 @@ proc initProcEntry*(c: var TCtx, prc: PSym): FuncTableEntry =
 
   # Create the env parameter type (if an env param exists)
   result.envParamType =
-    if (let envP = prc.getEnvParam(); envP != nil):
-      c.getOrCreate(envP.typ)
+    if tfCapturesEnv in prc.typ.flags:
```
saem: Rebase should catch this yes?

saem (commented):
Part way through this pass, breaking for lunch.

compiler/vm/vmbackend.nim
```diff
@@ -0,0 +1,449 @@
...
+    vmtypegen
+  ],
+  experimental/[
+    results
+  ]
+
+import std/options as stdoptions
+
+type
+  CodeFragment = object
```
saem, line 55: Very cool.

Total tangent, but DOD experiement lessons and digging into things like Racket, I think fragments are definitely an unsung hero. Where it's not quite as decomposed as a sea of nodes with little context or so monolithic as whole program singular AST, but logical chunks of AST + context all wrapped into one fragment. Which can be composed or decomposed as needed.

Glad to see we're nudging in that direction here.

saem:
Reading through the code and thinking about backends, targets, pltaform detection etc...

*Timing & Targets*

I think we want an explicit `compileTime` (or something) flag that one uses to detect whether the compiler/interpreter is "pre-executing" the code vs program/script runtime. Which is different from target detection, which can be `vm` (or something), which should answer whether the target backend is the virtual machine or not.

I think these two things should cover the distinction we need to make:
- compile time detection
- virtual machine target detection

With that clean slate thinking written down, now need to think through how does today map to that and/or did something get missed?

*Some Mapping Thoughts*

Some partial thinking about mapping: nimscript could be thought of as a target itself, I think it would run in a slightly more restrictive fashion, whereas the vm target is potentially distinct from that as we don't care about sandboxing nearly to the same extent.

Current Summary:

This might be wrong, but maybe it's easier to discuss?

| Type | Name | Old Name | Description |
|--------|----------|---------------|-----------------|
| target | `vm` | `vm`| backend target like any other, c, js, cpp |
| target | `nimscript` | `nimscript` | a special target, the compiler is used like an interpreter, more sandboxing than the `vm` target and less than macro evaluation |
| timing/stage | `compileTime` | `nimvm` | detects when something is running/timing, but the target is still vm/c/js/cpp/nimscript |

testament/testament.nim
```diff
@@ -957,7 +957,7 @@ proc testSpecHelper(r: var TResults, run: var TestRun) =
         given.err, givenSpec = given.addr)
     else:
       let isJsTarget = run.target == targetJS
-      var exeFile = changeFileExt(run.test.name, if isJsTarget: "js" else: ExeExt)
+      var exeFile = changeFileExt(run.test.name, if isJsTarget: "js" elif run.target == targetVM: "vm" else: ExeExt)
```
saem: BTW, for testatment there is a func `defaultTargets`, which takes a category and returns the default targets for that category. Easy way to bump up coverage category by category, then you only have to change a test here or there for includes/excludes.

saem:
Additional bits from the PR comment body:

> After the runner is built, `nim c --backend:vm -r` can be used to build and run a NimSkull project

Would it be reasonable to ask for `nim vm` (compile) or `nim vm -r` (compile and run)?

> ## Things that don't work yet:
> ... snip ...

Your todos, including the ones in the code, are comprehensive I had little to add that didn't feel like nitpicking.

> * the binaries use the `.vm` extension (open for discussion)

Not a deal breaker, we can change it later, but I was thinking calling it that refers more to bytecode as the output isn't so much a virtual machine as what can be consumed by one. In that vein `.nimbc` or `.nbc`, where bc is for byte code -- I'm very imaginative, thank you. :D

> * the stdlib changes are quickly thrown together and just here for testing

There is plenty to change about it, so as long as tests pass and it's not destructive, which they didn't seem to be.

Outside of this PR, I think the stdlib APIs need some serious rethink, regardless. So if the tests work and people can refactor relatively easily then that's not hampered. The stdlib really should be a truly minimal set that works across all targets, then breaking into native vs js targets, and finally truly target/OS specific.

> * the compiler front-end changes are also rather thrown-together

I didn't see anything that set off alarms

> * the test adjustment have large overlap with the changes [here](https://github.com/nim-works/nimskull/pull/265). It would make sense to finish the mentioned PR first, I believe

That PR might be a lot easier with the existing Testament changes, let me take a peek.

saem (commented):
Whoops, forgot to post these yesterday. 🤦🏽

zerbina:
> _Timing & Targets_
> ... snip ...

I agree. Compile-time detection and VM target detection are currently conflated and need to be separated.

> _Some Mapping Thoughts_
> ... snip ...

`nimscript` being treated as a (special) target makes sense.

> Current Summary:
> This might be wrong, but maybe it's easier to discuss?

The first branch in a `when nimvm` is currently taken whenever the VM is used. That is:
* during compile-time execution (macros, static, etc.)
* when executing NimScript
* when compiling code for the standalone VM target

Replacing `nimvm` with something like `compileTime` and, like you said, only taking the positive branch during compile-time execution is a good way forward, I believe. The rename to `compileTime` can happen separately from changing the meaning of `nimvm`.

Most usages of `nimvm` in the stdlib that I've seen so far query whether the VM is used, so the change would require a fair amount of adjustments. Maybe it makes sense to change the semantics in a separate PR? There would be no easy way to test the change in a scalable manner then however (until the first-class VM back-end is merged). 

> Would it be reasonable to ask for `nim vm` (compile) or `nim vm -r` (compile and run)?

Sure, it's a simple change. 

> Not a deal breaker, we can change it later, but I was thinking calling it that refers more to bytecode as the output isn't so much a virtual machine as what can be consumed by one. In that vein `.nimbc` or `.nbc`, where bc is for byte code -- I'm very imaginative, thank you. :D

I think we should change it now. `.vm` was just a placeholder used to make it easy to ignore the artifacts in ``git``. For what it's worth, I actually really like `.nimbc` :P

> There is plenty to change about it, so as long as tests pass and it's not destructive, which they didn't seem to be.

The adjustments currently boil down to an additional `defined(vm)` in some `when` statements. The changes are not exhaustive and only cover some modules and procedures that are used in the tests I enabled the VM target for. I guess it would work for now.

One thing I noticed is that there's target or OS detection used in places where the wrapped code actually only cares about whether or not a procedure or variable is declared. I changed some `when` statements that I had to  modify anyways to use `declared(x)` instead

> That PR might be a lot easier with the existing Testament changes, let me take a peek.

For the most part, it's about the `stdout.write` -> `echo` (or using assertion) change, as the VM currently doesn't support `stdout` and friends. Then there's also the addition of the `targets` field to the various test specifications

saem:
> > _Timing & Targets_
> > ... snip ...
> 
> I agree. Compile-time detection and VM target detection are currently conflated and need to be separated.
> 
> > _Some Mapping Thoughts_
> > ... snip ...
> 
> `nimscript` being treated as a (special) target makes sense.

Cool, I don't think it makes sense to scope it into here, we'll eventually need to deal with:
- current OS based conditionals stay as is (win, js, nodejs, etc)
- yeah js is special and nodejs windows _is_ different than nodjs linux
- the above OS based conditionals are for target, but the VM and nimscript care about both host and target

Only takeaway for now is leave some affordance in today's  naming convention a it's all target oriented and there will be an othogonal host oriented variety. I think it'll be used less frequently, and it could be as simply as a `host_` prefix.

> > Current Summary:
> > This might be wrong, but maybe it's easier to discuss?
> 
> The first branch in a `when nimvm` is currently taken whenever the VM is used. That is:
> 
> * during compile-time execution (macros, static, etc.)
> * when executing NimScript
> * when compiling code for the standalone VM target
> 
> Replacing `nimvm` with something like `compileTime` and, like you said, only taking the positive branch during compile-time execution is a good way forward, I believe. The rename to `compileTime` can happen separately from changing the meaning of `nimvm`.

Absolutely, it should totally be separate.

> Most usages of `nimvm` in the stdlib that I've seen so far query whether the VM is used, so the change would require a fair amount of adjustments. Maybe it makes sense to change the semantics in a separate PR? There would be no easy way to test the change in a scalable manner then however (until the first-class VM back-end is merged).

Yup, strongly agree that this PR is a prerequisite. My thinking:
1. Get this merged
2. Chase coverage and bugs with the VM
3. Rename to compileTime and do step 2 for it

> > Would it be reasonable to ask for `nim vm` (compile) or `nim vm -r` (compile and run)?
> 
> Sure, it's a simple change.

Yay! 🤩🎉

> > Not a deal breaker, we can change it later, but I was thinking calling it that refers more to bytecode as the output isn't so much a virtual machine as what can be consumed by one. In that vein `.nimbc` or `.nbc`, where bc is for byte code -- I'm very imaginative, thank you. :D
> 
> I think we should change it now. `.vm` was just a placeholder used to make it easy to ignore the artifacts in `git`. For what it's worth, I actually really like `.nimbc` :P

🔥😁, Ok `.nimbc` wins.

> > There is plenty to change about it, so as long as tests pass and it's not destructive, which they didn't seem to be.
> 
> The adjustments currently boil down to an additional `defined(vm)` in some `when` statements. The changes are not exhaustive and only cover some modules and procedures that are used in the tests I enabled the VM target for. I guess it would work for now.

It does, it'll be a good chance to streamline and improve test coverage.

> One thing I noticed is that there's target or OS detection used in places where the wrapped code actually only cares about whether or not a procedure or variable is declared. I changed some `when` statements that I had to modify anyways to use `declared(x)` instead

That seems better, so let's run with it.

> > That PR might be a lot easier with the existing Testament changes, let me take a peek.
> 
> For the most part, it's about the `stdout.write` -> `echo` (or using assertion) change, as the VM currently doesn't support `stdout` and friends. Then there's also the addition of the `targets` field to the various test specifications

Ah, @canelhasmateus' changes get into that too. Maybe some simple category by category changes might be good, I'll try to spend an hour on that today and see how far I get. Then we can use the default category bit to significantly bump coverage for this change.

canelhasmateus:
Minor comment: [Can we add the vm to testament help doc?](https://github.com/nim-works/nimskull/blob/dd631f7b41c4af112bbc2746a92d54fed9782f91/testament/testament.nim#L44) Is objc even a backend still?

zerbina:
> Minor comment: [Can we add the vm to testament help doc?](https://github.com/nim-works/nimskull/blob/dd631f7b41c4af112bbc2746a92d54fed9782f91/testament/testament.nim#L44)

Yes, good catch, thanks! `nimc.rst` also needs to be adjusted.

> Is objc even a backend still?

The compiler still recognizes it as as such and there's dedicated Obj-C handling in the C back-end. From a quick scan, it seem like only a single test makes use of the back-end however.

saem:
> > Is objc even a backend still?
> 
> The compiler still recognizes it as as such and there's dedicated Obj-C handling in the C back-end. From a quick scan, it seem like only a single test makes use of the back-end however.

If the test passes leave it, if it doesn't pass and takes more than 5 minutes to figure out/fix then just delete the whole objc backend. It's firmly in the category of nice to have.

compiler/vm/vmrunner.nim
```diff
@@ -0,0 +1,256 @@
...
+    # value of ``programResult``, which we use as the runner's exit code
+    result = r.unsafeGet.intVal.int
+  else:
+    let err = r.takeErr
+    # an uncaught error occurred
+    c.config.localReport(err.stackTrace)
+    c.config.localReport(err.report)
+    result = 1
+
+programResult = main(commandLineParams())
```
saem:
```suggestion
when defined(isMainModule):
  programResult = main(commandLineParams())
```

I think that should be fine and allow for direct module unit testing / inclusion?

saem:
Did another pass, it's down to the todos this is really coming together!

compiler/vm/vmrunner.nim
zerbina: Done. `main` is now also exported .

zerbina:
The missing parts are now implemented.

Doing symbol alive detection in a separate pre-pass (similar to how `cbackend` does it) instead of during code-generation in `vmgen` would greatly reduce the complexity of the logic in `vmbackend`. The change would have other benefits too (and also some drawbacks), but I believe that it is better done in a separate PR.

Splitting up `TCtx` into smaller parts will help with further reducing the complexity in `vmbackend`.

compiler/vm/packed_env.nim
zerbina: Added a test (`tests/compiler/tvmbackend.nim`).

compiler/vm/vmdef.nim
```diff
@@ -614,7 +629,7 @@ template types*(c: TCtx): untyped =
   ## Transition helper
   c.typeInfoCache.types
 
-proc init(cache: var TypeInfoCache) =
+proc init*(cache: var TypeInfoCache) =
```
saem, line 632: maybe I'm missing something, but shouldn't we call the template instead of simply defining it in the body?

saem: nevermind... github cut off the rest of the definition... sigh

compiler/vm/vmrunner.nim
```diff
@@ -0,0 +1,274 @@
...
+      deref(dst).seqVal.newVmSeq(dst.typ, L, mem)
+
+    var i = 0
+    while i < L:
+      result += loadConst(s, idx+1+result, getItemHandle(dst, i), mem)
+      inc i
+
+  of akString:
+    assert n.kind == pdkString
+    # TODO: `VmString` needs to be able to store string literals...
```
saem: Just confirming are you using `TODO` to indicate handle before the PR is done -- just going by history?

compiler/vm/packed_env.nim
```diff
@@ -0,0 +1,964 @@
...
+  var dec: TypeInfoDecoder
+  dec.types.newSeq(p.nimTypes.len)
+  dec.syms.newSeq(p.nimSyms.len)
+
+  mapList(result, p.typeInfos, x):
+    VmTypeInfo(internal: types[x.t],
+               nimType:  dec.loadType(p, x.nt))
+
+proc loadEnv*(dst: var TCtx, src: PackedEnv) =
+  ## Loads all data from `src` int `dst` for which no further/extra processing
```
saem:
```suggestion
  ## Loads all data from `src` into `dst` for which no further/extra processing
```

saem (approved):
Found a minor typo but I think it's good to go, pulling it out of draft and clearing CI is all that remains.

saem:
Marked it ready to review based on chat, if it's too premature then it can go back to draft, I just wanted to peek at how CI is doing.

compiler/vm/vmrunner.nim
zerbina: I'm mostly using `TODO` to indicate 'handle before PR is done', but sometimes also to document a concrete step that's planned in a refactoring/change split across multiple PRs.

zerbina:
Both the typo and the `tcan_compile_nim` failure are fixed and I also added the `vm` target to the testament command line help output (thanks @canelhasmateus).

Due to the removal of the stdlib related adjustments, the tests in the 'lang' category are failing now, since `system.nim` includes various code and modules that don't work on the `vm` target.

The two possible ways forward that I see are:
- adjust `system` and `io` in a separate PR; merge it; rebase this PR onto `devel` afterwards
- do the adjustments as part of this PR

I'd personally go with the first approach, although it has the downside that the `system` and `io` module changes would have no direct test coverage on introduction (since there'd be no `vm` target yet). What do you think @saem ?

saem:
I prefer the first option, as well.

Part of the solution might be extracting bits out of system into their own module. Forcing some much needed decomposition of system.

tests/lang/s02_core/s01_user_defined_data_types/t03_inheritance.nim
```diff
@@ -52,8 +52,9 @@ block derive_from_ref_root_obj:
   ## To convert base type back to derived, you can use object conversion
   doAssert Derived(base).f2 == 12
 
-  when not defined(js):
+  when not defined(js) and not defined(vm):
    when not defined(js) or defined(tryBrokenSpecification):
```
zerbina: @saem The outer `when` seems redundant?

saem: yeah, that was a hasty add, pretty sure that can be consolidated.

zerbina:
I rebased the PR onto `devel` and adjusted the `isNimVmTarget` definition.

bors r+

PR #349, opened 2022-06-13
vm: remove lazy loading of `const` values

zerbina:
## Summary

Values for `const`s are now loaded into VM memory on their first
_appearance_ instead of on their first _access_. This has no effect
on guest-visible semantics.

The responsibility of loading the constant value is moved outside the
execution engine, simplifying the latter and making the loading logic
easily interchangeable. In addition, it also makes implementation of
the VM back-end simpler.

## Details

- no dedicated layer responsible for updating the execution
 environment with the dependencies exist yet, so the loading of the
 `const` values is temporarily done in `vmgen`

---

The mentioned dedicated layer already exists in the VM back-end PR


compiler/vm/vmgen.nim
```diff
@@ -2556,12 +2552,29 @@ proc gen(c: var TCtx; n: PNode; dest: var TDest; flags: TGenFlags = {}) =
...
-      discard c.getOrCreate(s.typ)
+      if s.ast.kind in nkLiterals:
+        let lit = genLiteral(c, s.ast)
+        c.genLit(n, lit, dest)
+      else:
+        let
+          next = c.complexConsts.len
+          i = c.symToConst.mgetOrPut(s.id, next)
+
+        # XXX: allocating and initializing the constant during code-gen is a
```
saem, line 2563: At first I wondered if it's also an issue because of typed macros, we might have a proc body with constants that are defined and then completely replaced after a further macro does a rewrite and remove them.

But I believe that doesn't apply:
1. We generate code for the macro itself, which is more defined
2. We do not generate code for the subject of the macro that is getting rewritten, that's AST nodes

As in, we really on the language property that code already semantically analysed is effectively static, not self modifying, and said static code is what the VM runs and may run that code over data over code that hasn't yet been fully semantically analysed (subject of a macro, constant expression, etc).

Can you poke at my understanding?

saem:
@zerbina please confirm my understanding, if it's legit then merge away. It's the only potential issue I could think of.

compiler/vm/vmgen.nim
zerbina: Your understanding is correct, during compile-time evaluation, code is only translated to VM byte-code right before it's executed, never earlier.

As you said, the whole logic depends on the fact that a symbol's associated AST (e.g. function body, constant data, etc.) doesn't change after the symbol is fully semantically analysed

saem: Thanks for confirming.

saem:
Bors r+

PR #351, opened 2022-06-14
vm: callback related clean-ups

zerbina:
## Summary

- replace usage of `split` function with iterator
- split up the `callbacks` list because `key` and `value` both have
 different lifetimes as well as readers and writers
- don't cache the callback index in the routine symbol's `position`
 field. This makes callback lookup side-effect free

Part of the ongoing effort to split up `TCtx` into smaller parts.

## Future work

Depending on how the planned bigger callback rework goes, a dedicated
data structure to accelerate the identifier pattern matching might make
sense, as the current `seq`-based implementation is rather inefficient - in
terms of both memory consumption and run time

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -2140,7 +2140,7 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): RegisterInd
         for i in (rb+1)..<(rb+rc):
           checkHandle(regs[i])
 
-        c.callbacks[bb.cbOffset].value(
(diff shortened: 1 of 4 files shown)


saem (approved):
Nice, separating based on lifetime+write/read seems to be a winning strategy in many cases. Modularity being one such thing, I'm aiming to do similar stuff with TContext, so glad to see we're pulling in the same direction.

The non-lent pairs is interesting, unlikely we'd want it to work that way all the time with sequences, but this sort of stdlib of thing will need a rethink soon.

Bors r+

saem:
Bors r+

PR #352, opened 2022-06-14
vm: change how string constants are stored

zerbina:
## Summary

Store string VM constants as `string` instead of `VmString`.

This previous approach required a layering violation and an unwanted
dependency, since the VM's allocator (a run-time resource) needed for
creating `VmString`s was used and modified during code-generation
(i.e. compile-time).

## Details

As a consequence of this change, string literals usage (in both `var`
and `let` contexts) always means an additional allocation + copy for
now. Implementing support for foreign memory in the VM will provide
the framework to fix this.

---

For a perspective, the first steps of foreign memory support as would be required for storing string literals in `VmString` are currently planned after the `vmgen` rewrite/overhaul (which itself is planned after the first-class VM back-end is merged).

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -2217,10 +2217,17 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): RegisterInd
         for s in list.items:
           if v in s: return true
 
+      func cmp(a: string, b: VmString): int =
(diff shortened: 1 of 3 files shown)


zerbina:
The `stdlib/tresult.nim` test failing is due to an issue in `vmgen` that is unrelated to this change

saem (approved):
Bors r+

PR #353, opened 2022-06-16
vm: don't use `PSym.position` for slot indices

zerbina:
## Summary
Use a lookup table that's stored together with the other code-gen state
for mapping a symbol to it's register slot index instead of using the
symbols' `position` field for this.

This is another step towards removing AST modifications from `vmgen`.

### DIFF
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -913,6 +913,21 @@ proc genCall(c: var TCtx; n: PNode; dest: var TDest) =
 template isGlobal(s: PSym): bool = sfGlobal in s.flags and s.kind != skForVar
 proc isGlobal(n: PNode): bool = n.kind == nkSym and isGlobal(n.sym)
 
+func local(prc: PProc, sym: PSym): TDest {.inline.} =
+  ## Returns the register associated with the local variable `sym` (or -1 if
+  ## `sym` is not a local)
+  let
+    s = sym
+    r = prc.locals.getOrDefault(s.id, -1)
+  # Problem: in macro bodies, copies of the parameters' symbols with differnt
+  # IDs are used (see `addParamOrResult`), meaning that these symbols have no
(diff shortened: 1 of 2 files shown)


compiler/vm/vmgen.nim
```diff
@@ -2770,12 +2782,20 @@ proc genExpr*(c: var TCtx; n: PNode, requiresValue = true): VmGenResult =
   #echo renderTree(n)
   #c.echoCode(result)
 
-proc genParams(c: var TCtx; params: PNode) =
-  # res.sym.position is already 0
-  setLen(c.prc.regInfo, max(params.len, 1))
-  c.prc.regInfo[0] = RegInfo(refCount: 1, kind: slotFixedVar)
+proc genParams(prc: PProc; s: PSym) =
```
saem, line 2784: All of a sudden the params make way more sense!

compiler/vm/vmgen.nim
```diff
@@ -913,6 +913,22 @@ proc genCall(c: var TCtx; n: PNode; dest: var TDest) =
 template isGlobal(s: PSym): bool = sfGlobal in s.flags and s.kind != skForVar
 proc isGlobal(n: PNode): bool = n.kind == nkSym and isGlobal(n.sym)
 
+template local(sym: PSym): TDest =
```
saem: I personally much prefer passing in all parameters instead of late binding with mixin. I imagine type might be a thing, so I tend to use a type class.

saem (commented):
Minor style thing, but I'm good to merge.

compiler/vm/vmgen.nim
zerbina: I do too. To be honest, I have no idea why I even used a template here in the first place 😅

zerbina:
`local` is now a `func` (it made no sense for it to be a template)

bors r+

PR #356, opened 2022-06-18
vm: don't depend on `lambdalifting` module

zerbina:
## Summary
`vmcompilerserdes` and `vmaux` both imported `lambdalifting` for
the `getEnvParam` function. Due to the cyclic import dependency
between `transf` and `lambdalifting`, this could cause
`undeclared identifier` errors.

## Details
Usage of `getEnvParam` is replaced with manual lookup (it's only two
lines of simple code and there are only two usage sites).

### DIFF
--- a/compiler/vm/vmaux.nim
+++ b/compiler/vm/vmaux.nim
@@ -21,7 +21,6 @@ import
   ]
 
 from compiler/ast/nimsets import overlap
-from compiler/sem/lambdalifting import getEnvParam
 
 # XXX: this proc was previously located in ``vmgen.nim``
 func matches(s: PSym; x: IdentPattern): bool =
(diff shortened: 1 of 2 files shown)


saem (commented):
I'd recommend putting it in a common dependency and mark it as a poor place for it.

----

I'm thinking lifting etc needs to be broken up as a few modules eventually. The precise structures aren't clear yet as to how. But I digress and the module breakup isn't in scope of this PR.

zerbina:
> I'd recommend putting it in a common dependency and mark it as a poor place for it.

Which version? The manual re-implementation now used across the VM modules or the `getEnvParam` from `lambdalifting`? They're not identical and support different use-cases:
- `lambdalifting.getEnvParam` is meant for retrieving the env param's symbol before the lambda lifting pass is completed, but still works after
- the `tfCapturesEnv` approach only works after lambda lifting for the respective routine is completed, but doesn't depend on implementation details (i.e. the name of the env parameter) of `lambdalifting`

The `tfCapturesEnv` version could be moved to the ``types`` module. Replacing the only external usage of `lambdalifting.getEnvParam` in `itersgen.nim` with the then new proc in ``types`` would allow for it to be not-exported

saem:
I forgot that there were two versions. 😅 

It would be for the `tfCapturesEnv` version, some where in a vm related module might be fine because we can assume that lambdalifting should be done by that point so it's a property of any ast we're encountering?

If that's off, I'm good with a merge too.

saem (approved):
Awesome!

Bors r+

PR #359, opened 2022-06-24
tests: prep lang category for the vm backend

saem:
## Summary
- ready to add the vm backend as lang categories default target
- lang category now runs for all categories

Changes:
- default the lang category to run on all relevant backends
- fixed test target declarations
- disabled some tests for broken backends like cpp

---

### DIFF
--- a/tests/lang/s02_core/s02_procedures/t02_argument_passing.nim
+++ b/tests/lang/s02_core/s02_procedures/t02_argument_passing.nim
@@ -2,9 +2,10 @@ discard """
 description: '''
 Covers the multitude of ways you can pass arguments to procedures.
 '''
+targets: "!cpp"
 """
 
-##
+# knownIssue: broken in cpp
 
 
 block different_number_of_arguments:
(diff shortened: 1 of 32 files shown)


saem:
@zerbina PTAL, this should help prep this category for the vm. I'm updating a few categories at a time, make it a bit easier for @canelhasmateus. :D

zerbina (approved):
Looks good to me, thanks! The ability to just include the VM target for relevant categories in ``defaultTargets`` will make things much easier.

tests/lang/s02_core/s02_procedures/t02_argument_passing.nim
```diff
@@ -2,9 +2,10 @@ discard """
 description: '''
 Covers the multitude of ways you can pass arguments to procedures.
 '''
+targets: "c !cpp js"
```
zerbina: Would `targets: "!cpp"` also work?

saem: Ah, good catch, most likely

saem: fixed.

PR #360, opened 2022-06-26
internals: semDeref rewritten to case/let style

saem:
## Summary
- rewrote semDeref like the recent semVarOrLet refactor

https://github.com/nim-works/nimskull/commit/1d7def279c28067035487530403bc6b452873213

## Details
- more explicit node kind checks
- use let to establish factos
- distinguish sem vs eval parts
- generally more in keeping with structured programming

---

## Notes for Reviewers
* hopefully a bit more self-explanatory

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -1049,7 +1049,7 @@ proc semIndirectOp(c: PContext, n: PNode, flags: TExprFlags): PNode =
   result = nil
   checkMinSonsLen(n, 1, c.config)
   if n.kind == nkError: return n
-  var prc = n[0]
+  let prc = n[0]
   if n[0].kind == nkDotExpr:
     checkSonsLen(n[0], 2, c.config)
     let n0 = semFieldAccess(c, n[0])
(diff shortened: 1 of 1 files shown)


saem:
@haxscramper similar to the other refactors (semVarOrLet, semExportExcept, etc)... is this more understandable?

(if yes, please bors)

compiler/sem/semexprs.nim
```diff
@@ -1605,21 +1605,72 @@ proc buildOverloadedSubscripts(n: PNode, ident: PIdent): PNode =
...
-  if a != nil:
-    if a.kind == nkNilLit:
-      localReport(c.config, n, reportSem rsemDisallowedNilDeref)
-    n[0] = a
-  result = n
-  var t = skipTypes(n[0].typ, {tyGenericInst, tyVar, tyLent, tyAlias, tySink, tyOwned})
-  case t.kind:
-    of tyRef, tyPtr:
-      n.typ = t.lastSon
+  ## analyse deref such as `[]foo`, when given an nkBracketExpr converts it to
```
haxscramper: IIRC deref expressions are written as `foo[]`

compiler/sem/semexprs.nim
```diff
@@ -1605,21 +1605,72 @@ proc buildOverloadedSubscripts(n: PNode, ident: PIdent): PNode =
...
+  ## an nkDerefExpr and evals it, if given an nkDerefExpr evals it, nkError is
+  ## passed through.
+  ##
+  ## it to a  behaves differently depending upon whether
```
haxscramper: Speaking of "simple English" just mentioned in the `#main` - maybe it could be rewritten as "There are two different cases/behaviors: if target is a constant expression (static) and otherwise (dynamic)".

Can leave it as it is now, I just noticed it might be a good example of "let's write in a simplified English".

compiler/sem/semexprs.nim
```diff
@@ -1605,21 +1605,72 @@ proc buildOverloadedSubscripts(n: PNode, ident: PIdent): PNode =
...
+      isTargetConstExpr = targetAsConstExpr != nil
+      isTargetConstNil = isTargetConstExpr and
+                         targetAsConstExpr.kind == nkNilLit
+      semmedTarget =
+        if isTargetConstExpr:
+          targetAsConstExpr
+        else:
+          derefTarget
+      typ = skipTypes(semmedTarget.typ,
+                      {tyGenericInst, tyVar, tyLent, tyAlias, tySink, tyOwned})
```
haxscramper: It might be a bit of a stretch goal for now, but I thought about moving all hardcoded `skipTypes` sets like these in the constants, even if they are used only once in the specific PR. In the future, it might be easier to consolidate them, and right now you, as a PR author can give a meaningful name to it, like `const tyIgnoredInDeref = {tyGenericInst, tyVar, tyLent, tyAlias, tySink, tyOwned}`

haxscramper (approved):
It is more readable (if only because of the comments and better formatting). I added all suggestions I could think of, but those are mostly cosmetic ones.

compiler/sem/semexprs.nim
saem: thanks, fixed.

compiler/sem/semexprs.nim
saem: rewritten.

compiler/sem/semexprs.nim
saem: I added a local const and I noticed a possible issue for later investigation.

PR #224, opened 2022-02-02
WIP draft of the documentation generator rework

haxscramper:
For now, I'm just testing out different implementation ideas, opening PR in case someone is interested to comment or talk about different parts of the implementation. Right now it is *super* experimental.

PR #361, opened 2022-06-26
add initial block formatter implementation

haxscramper:
Added implementation of the text layouter for error message or code formatter. Split from the #224

This PR provides test and implementation of the generic text block layout algorithm, which can be used for error message formatting and general code layout. Split from aforementioned documentation generator PR was done in order to simplify latter one and because formatter implementation is usable on its own (specifically needed for #332).

Reimplements https://github.com/google/rfmt

### DIFF
--- a/lib/experimental/dod_helpers.nim
+++ b/lib/experimental/dod_helpers.nim
@@ -1,0 +1,184 @@
+import std/[macros, strutils, options]
+
+template declareIdType*(
+    Name: untyped,
+    addHash: static[bool] = false,
+    BaseType: typed = uint64
+  ): untyped {.dirty.} =
+
+  type
+    `Name Id`* = distinct BaseType
+
+  const `Empty Name Id`* = `Name Id`(0)
+
+  func `==`*(i1, i2: `Name Id`): bool = i1.int == i2.int
+  func isNil*(i: `Name Id`): bool = i == `Empty Name Id`
+
+  func `$`*(id: `Name Id`): string =
+    if id == `Empty Name Id`:
+      result = "<empty- " & astToStr(Name) & "Id>"
+
+    else:
+      result = "<" & astToStr(Name) & "-" & $id.int & ">"
+
+  func toIndex*(id: `Name Id`): int =
+    assert not isNil(id), $id
+    result = int(BaseType(id) - 1)
+
+  func `to Name Id`*(idx: int): `Name Id` =
+    result = `Name Id`(idx + 1)
+
+  when addHash:
+    func hash(id: `Name Id`): Hash = Hash(id.int)
+
+import std/bitops
+
+proc bits[T](t: T): string =
+  let v = cast[uint64](t)
+  for bit in countdown(high(BitsRange[T]), 0):
+    if testBit(v, bit.int):
+      result.add "1"
+    else:
+      result.add "0"
+
+
+template declareHighMasking*(
+    Name: untyped,
+    highMaskRange: static[range[0..64]] = 0,
+    BaseType: typed = uint64
+  ): untyped =
+    const Max = sizeof(BaseType) * 8
+    const shift = Max - highMaskRange
+    const highMask =
+      block:
+        var mask: BaseType
+        for idx in countdown(Max - 1, Max - highMaskRange):
+          mask.setBit(idx)
+
+        mask
+
+    type
+      `Name Id Mask`* {.inject.} = distinct BaseType
+
+    func `==`*(m1, m2: `Name Id Mask`): bool {.inject.} =
+      BaseType(m1) == BaseType(m2)
+
+    func getMask*(id: `Name Id`): `Name Id Mask` {.inject.} =
+      `Name Id Mask`((BaseType(id) and highMask) shr shift)
+
+    func popMask*(id: `Name Id`): `Name Id` {.inject.} =
+      `Name Id`((BaseType(id) and not highMask))
+
+    func setMask*(id: var `Name Id`, mask: `Name Id Mask`) {.inject.} =
+      id = `Name Id`(BaseType(id) or (mask.BaseType shl shift))
+
+
+template declareStoreType*(Name: untyped): untyped {.dirty.} =
+  type
+    `Name Store`* = object
+      data: seq[Name]
+
+  func len*(store: `Name Store`): int = store.data.len
+
+  func add*(store: var `Name Store`, item: `Name`): `Name Id` =
+    result = `to Name Id`(store.data.len)
+    store.data.add item
+
+  template `[]`*(store: var `Name Store`, index: `Name Id`): Name =
+    store.data[toIndex(index)]
+
+  iterator items*(store: `Name Store`): Name =
+    for item in items(store.data):
+      yield item
+
+  iterator mitems*(store: var `Name Store`): var Name =
+    for item in mitems(store.data):
+      yield item
+
+  iterator pairs*(store: `Name Store`): (`Name Id`, Name) =
+    for idx, item in pairs(store.data):
+      yield (`to Name Id`(idx), item)
+
+  iterator mpairs*(store: var `Name Store`): (`Name Id`, var Name) =
+    for idx, item in mpairs(store.data):
+      yield (`to Name Id`(idx), item)
+
+template declareStoreField*(Type, field, Name: untyped): untyped {.dirty.} =
+  func add*(main: var Type, typ: `Name`): `Name Id` = main.field.add typ
+  func `[]`*(main: Type, id: `Name Id`): Name = main.field[id]
+  func `[]`*(main: var Type, id: `Name Id`): var Name = main.field[id]
+
+
+macro declareStoredTableField*(Db, field, Name, Typ: untyped): untyped =
+  let name = field.strVal().capitalizeAscii()
+  let
+    (hasName, getName, setName) = (
+      ident("has" & name), ident("get" & name), ident("set" & name))
+
+    nameId = ident(Name.strVal() & "Id")
+
+  result = quote do:
+    func `hasName`*(main: `Db`, id: `nameId`): bool = id in main.`field`
+    proc `getName`*(main: `Db`, id: `nameId`): `Typ` = main.`field`[id]
+    func `setName`*(main: var `Db`, id: `nameId`, val: `Typ`) =
+      main.`field`[id] = val
+
+    func `setName`*(main: var `Db`, id: `nameId`, val: Option[`Typ`]) =
+      if isSome(val):
+        main.`field`[id] = get(val)
+
+
+macro declareStoredTableSeqField*(Db, field, Name, Typ: untyped): untyped =
+  let name = field.strVal().capitalizeAscii()
+  let
+    hasName = ident("has" & name)
+    getName = ident("get" & name)
+    setName = ident("set" & name)
+    addName = ident("add" & name)
+    lenName = ident("len" & name)
+    nameId = ident(Name.strVal() & "Id")
+
+  result = quote do:
+    func `hasName`*(main: `Db`, id: `nameId`): bool = id in main.`field`
+    proc `getName`*(main: `Db`, id: `nameId`): seq[`Typ`] = main.`field`[id]
+    func `setName`*(main: var `Db`, id: `nameId`, val: seq[`Typ`]) =
+      main.`field`[id] = val
+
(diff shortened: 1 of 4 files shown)


haxscramper:
The PR implementation is ready for the review, I will add test now.

lib/experimental/dod_helpers.nim
```diff
@@ -0,0 +1,184 @@
+import std/[macros, strutils, options]
```
saem, line 1:
```suggestion
## Contains various types, procs and templates for data oriented design style
## programming/memory layouts. See: https://www.dataorienteddesign.com/dodbook/

import std/[macros, strutils, options]
```

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1884 @@
+## Nim reimplementation of layout algorithm from
+## https://github.com/google/rfmt.
+
+# Thanks to nim being syntactically close to python this is mostly
+# just blatant copy-paste of python code with added type annotations.
```
saem:
```suggestion
## Nimskull reimplementation of layout algorithm from
## https://github.com/google/rfmt.

# Thanks to nimskull being syntactically close to python this is mostly
# just blatant copy-paste of python code with added type annotations.
```

easier when we need to do a big ol' rename.

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1884 @@
...
+  lenientops,
+  options,
+  hashes,
+  math,
+  sugar,
+  intsets
+]
+
+const
+  infty = 1024 * 1024 * 1024 * 1024
```
saem:
```suggestion
  infty = 1024 * 1024 * 1024 * 1024
    ## info type bit mask thingy??
```

not sure what it should say yet, but that was my first guess besides "infinity"

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1884 @@
...
+    ("\n", "\\n"),
+    ("\\", "\\\\")
+  ])
+
+#*************************************************************************#
+#****************************  Format policy  ****************************#
+#*************************************************************************#
+#===========================  Type definition  ===========================#
+
+import ./dod_helpers
```
saem: boo, relative imports!

(I'm guessing it's due to the main module scenario)

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1751 @@
...
+  LayoutElementKind = enum
+    lekString
+    lekNewline
+    lekNewlineSpace
+    lekLayoutPrint
+
+  LayoutElement = ref object
+    ## An element of a layout object - a directive to the console.
+    ##
+    ## This class sports a collection of static methods, each of which
```
saem:
```suggestion
    ## This object sports a collection of static methods, each of which
```

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1751 @@
...
+    ## returns an anonymous function invoking a method of the console
+    ## to which it is applied.
+    ##
+    ## Refer to the corresponding methods of the LytConsole class for
```
saem:
```suggestion
    ## Refer to the corresponding methods of the LytConsole object for
```

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1751 @@
...
+
+    knots: seq[int] ## a list of ints, specifying the margin settings at
+    ## which the layout changes. Note that the first knot is required to be
+    ## 0.
+    spans: seq[int] ## a list of ints, giving for each knot, the width of
+    ## the corresponding layout in characters.
+    intercepts: seq[float] ## constant cost associated with each knot -
+    ## computed cost of the outputting this layout solution at each know in
+    ## the `knots` list
+    gradients: seq[float] ## at each knot, the rate with which the layout
```
saem, line 144: Might be worth making distincts?

tests/stdlib/ttext_layouter.nim
```diff
@@ -0,0 +1,205 @@
...
+    var blocks = mapIt(0 .. i, lTX(&"arg{it}: int{it}"))
+    let bl = lH(lTX("proc ("),
+      lC(join(lH(blocks), lTX(", ")), join(lV(blocks), lTX(","))),
+      lTX(")"))
+
+    echo toString(bl)
+
+let comm = s.str(", ")
+
+echo toString(
```
canelhasmateus: Aside from the ones inside `for`, are these `echo` ammenable to being transformed into `doAssert`?

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1751 @@
...
+    ##
+    ## This object corresponds to a cost function in the main article.
+    ## Layout is constructed from this function.
+
+    id {.requiresInit.}: int
+
+    knots: seq[int] ## a list of ints, specifying the margin settings at
+    ## which the layout changes. Note that the first knot is required to be
+    ## 0.
+    spans: seq[int] ## a list of ints, giving for each knot, the width of
```
saem, line 139: is this meant to be one entry in `knots` will always have an entry in `spans` or is the relationship not strict 1-to-1?

tests/stdlib/ttext_layouter.nim
canelhasmateus: oh, not easily. Just noticed the toString definition above.

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1751 @@
...
+    id*: LytStrId
+    len*: int
+
+  LytStrSpan* = object
+    ## Span of multiple layout strings
+    strs: seq[LytStr]
+    len: int
+
+  LytBlock* = ref object
+    layoutCache: Table[Option[LytSolution], Option[LytSolution]]
```
saem, line 173: wait... even the key has an option?

as in we can have a single nil key with a possibly nil (optional) layout?

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1751 @@
...
+        str*: LytStr
+
+      of layEvSpaces:
+        spaces*: int
+
+      of layEvNewline:
+        discard
+
+
+  Layout* = ref object
```
saem, line 120: Can many `LayoutElements` refer to the same `Layout`?

tests/stdlib/ttext_layouter.nim
haxscramper: Yes, will do now

haxscramper: moved all checks into doAssert

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1754 @@
...
+#===========================  Type definition  ===========================#
+
+import ./dod_helpers
+
+import std/bitops
+declareIdType(LytStr)
+declareHighMasking(LytStr, highMaskRange = 8)
+
+
+type
```
saem, line 60: A preamble here might help, this is me making a guess, but I'm just trying to draw a "map" of
relationship between the types.

```suggestion
type
  ## These are types for formatting text. Text formatting doesn't have a perfect solution that works
  ## in all cases and we have to resort to heuristics to choose the most approriate text layout on a
  ## case by case basis.
  ##
  ## Text is initially described via a DSL as a set of blocks (`LytBlock`). When it's time to output the
  ## text a layout (`LytLayout`) determines the exact strings (`LytStr`) to output. Text layout requires
  ## heuristics to choose between various options we use a solver ('LytSolution') which is used to
  ## determine the precise layout choosen for a given block.
```

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1754 @@
...
+    id*: LytStrId ## Id of the original piece of text
+    len*: int ## It's lengh in units (units are specified - can be ASCII or
+              ## unicode or anything else)
+
+  LytStrSpan* = object
+    ## Span of multiple layout strings
+    strs: seq[LytStr]
+    len: int
+
+  LytBlock* = ref object
```
saem, line 172: do we expect to have have many things refer to the same block?

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1754 @@
...
+    layoutCache: Table[Option[LytSolution], Option[LytSolution]]
```
saem, line 173: If we can pop the `layoutCache` out, by having a block id and a separate data structure to store the cache itself, then we could have the DSL defined in one chunk/module and the layout solving/pipeline entirely separately.

If the break-up works then someone can read the DSL part alone for just consuming/using the API, might be easier to learn it over time.

saem (commented):
remarked on a few data type related matters, need to look at the algorithms.

I think it might be very helpful if we can use the DOD bits to split:
* dsl and types to build blocks
* transformer/solver
* the very simple output bits that can consume from the previous bit

lib/experimental/text_layouter.nim
```diff
@@ -0,0 +1,1762 @@
...
+  self.gradients.add gradient
+  self.layouts.add layout
+
+
+#=====================  LytSolution manipulation logic  =====================#
+
+
+proc minSolution(solutions: seq[LytSolution]): Option[LytSolution] =
+  ## Form the piecewise minimum of a sequence of LytSolutions.
+
```
saem: observation:

I find these visual breaks have me jumping back up to ensure I suddenly didn't lose context -- destroys my skim speed.

It's not a problem when I'm reading carefully, but quickly skimming is really hard because I'm purposefully not reading everything carefully, so I suddenly doubt things and wonder if I missed something and now I've fallen out of skim mode and am back tracking.

lib/experimental/text_layouter.nim
haxscramper: one to one

lib/experimental/text_layouter.nim
haxscramper: I don't understand the question. A layout block is a single block in the text layout. It does not refer to anything nor does anyone refer to it, aside from parent combinator blocks like line/stack/combinator/choice. If you are asking why the type is a `ref` - because it forms a tree that is supposed to be randomly added to or added into another trees, similar to how AST is constructed.

lib/experimental/text_layouter.nim
haxscramper: I don't understand what "visual breaks" do you refer to. Do you mean the `#=====================  LytSolution manipulation logic  =====================#` comment above the line you commented on?

lib/experimental/text_layouter.nim
saem: Follow-up: can one have many things refer to the exact same solution?

It's a ref type which is why I'm asking.

lib/experimental/text_layouter.nim
saem: yup, it was because it's a ref.

lib/experimental/text_layouter.nim
saem: this is because of python porting/blackbox.

lib/experimental/text_layouter.nim
saem: OK, this is a look at separately/later if the value types make sense.

lib/experimental/text_layouter.nim
saem: for later refactoring.

lib/experimental/text_layouter.nim
saem:
```nim
  ## Form the piecewise minimum of a sequence of LytSolutions.
  # (blank line is the visual break)
  ## more docs
```

The blank line in the block doc comment, I left it as an observation.

It's merely a data point so over some period of time (bigger than this PR) we eventually figure out a style guide.

lib/experimental/text_layouter.nim
saem: need to understand the black box better.

lib/experimental/text_layouter.nim
saem: need to understand the transform internals better, again.

deferred.

lib/experimental/text_layouter.nim
haxscramper: fixed to continuous comments now

saem:
Small thing would be add some explanation around motivation to the PR message body, but either way I'm ok to bors.

haxscramper:
updated the description

PR #366, opened 2022-06-28
Added 'Labels' key to testament spec

canelhasmateus:
Love for the test suite

Continuing with the theme of enhancing our test suite,
this commits adds another key to the testament spec.

The `labels` key is a space-separated string of metadata
describing the test, with purpose of being used
during development.

Since a full run of the test suite takes a long time,
it makes sense to run only a subset at a time. The
current `cat` folders does a very poor job , being
coarse and arbitrary. This key aims to lay a
foundation to fill the gaps.

We also continue with the efforts setted by the
previous commits:
. Remove almost all illegible names in files.
. Add a meaningful 'description' for the changed tests.
. Reduce the usage of 'echo' in the tests.

https://github.com/nim-works/nimskull/issues/41.


tests/ccgbugs/tcodegen_anonymous_aggregate.nim
```diff
@@ -0,0 +1,34 @@
+discard """
+targets: "c cpp"
```
saem, line 2: This might be handled by defaultCategory in testament.Nim, have a look and it might default to these targets.

tests/ccgbugs/tcodegen_anonymous_aggregate.nim
```diff
@@ -0,0 +1,34 @@
...
+labels: "backend_cpp , pragma , atomics , union"
```
saem: Backend CPP is deriveable

saem (commented):
Did a quick pass

tests/ccgbugs/tcodegen_anonymous_aggregate.nim
canelhasmateus: The way i set those up is that the labels contain the `origin` of the bug, not the backend in which it runs. Is this a meaningful  or useful differentiation? not sure. will remove for now

tests/ccgbugs/tcodegen_anonymous_aggregate.nim
canelhasmateus: the standardTargets is defined as either c or cpp , but not both. 
The defaultTargets func does not seen to cover this case, unless i'm mistaken.

canelhasmateus:
PTAL

tests/actiontable/tactiontable.nim
```diff
@@ -33,5 +30,5 @@ const
     "C": action3,
     "D": action4}.toTable
 
-actionTable1["C"]("arg")
-actionTable2["C"]("arg")
+doAssert actionTable1["C"]("arg") == "action 3 arg"
```
canelhasmateus: I probably should change these names to actionTableVar and actionTableConst

tests/alias/talias.nim
```diff
@@ -1,4 +1,9 @@
-# Test the alias analysis
+discard """
+labels: "alias"
+description: '''
+  . Test the alias analysis
```
canelhasmateus: Wasn't able to gather much information about this one.

tests/align/talign.nim
```diff
@@ -39,15 +43,13 @@ proc foobar() =
   doAssert (cast[uint](addr(mylocal2)) and 127) == 0
   doAssert (cast[uint](addr(mylocal3)) and 127) == 0
 
-  echo "align ok"
```
canelhasmateus: This echo is completely useless right? it is already preceded by doAsserts, and is not dependent on any computed value.

tests/align/talign.nim
```diff
@@ -39,15 +43,13 @@ proc foobar() =
...
-
 foobar()
 
-# bug #13122
+block issue_13122:
```
canelhasmateus: Put this second test inside a block. 
Sometimes i'm worried this changes the semantics, since top-level seems to be treated differently, but it doesn't look like the case here.

tests/ambsym/mresolution_identifier_1.nim
```diff
@@ -0,0 +1,9 @@
+import mresolution_identifier_2
```
canelhasmateus: The changes to the `ambsym` category are the biggest i've made so far - i'm getting more confident in these changes. 
So please, take a more careful look at these.

tests/ambsym/tresolution_identifier_2_ok.nim
```diff
@@ -0,0 +1,17 @@
+discard """
+labels: "enum identifier proc resolution"
+description: '''
+  . mresolution_identifier1 defines TExport, an enum with x, y, z.
+  . mresolution_identifier2 defines TExport, an enum with a, b, c.
+  . By expliciting a namespace alongside the type name, we are able to
+    resolve the ambiguity.
+  . This goes for both the TExport and the `foo` call. The `foo` call should
+    be resolved to the one defined in identifier_2, *not* in identifier_2_copy.
```
canelhasmateus: adding a `foo` proc to mresolution_identifier_1 seems like a good idea. 
Anyway, this whole category reads like a spec, and is probably made redundant by those. Is this the case, @haxscramper ?

tests/ambsym/tresolution_identifier_local.nim
```diff
@@ -0,0 +1,32 @@
...
+var
+  m: TMyType
+
+const size = 7
+const content = "1234"
+
+m.len = size
+m.data = content
+
+doAssert doLen(m, 5) == size
```
canelhasmateus: I'm not too keen on what was being tested here to begin with. 
Is it the overloading of the `len` stdlib proc to our own type? or is it the disambiguation of the x.len (field) vs x.len (UFCS)?
Why does `doLen` even has a second argument?
Does the {.inline.} pragma change the semantics?

canelhasmateus (commented):
More changes. This is the deepest one i made so far, so please take a more careful look at the changes for the `ambsym` category. 
In my mind, these changes wouldy be better made in a second PR, but i didn't want to open a new one before closing this. 
Not sure if pulverizing these incoming changes over multiple PRs is better / worse for reviewing - please share your opinions.

canelhasmateus:
pinging @zerbina into reviews as requested :P

tests/alias/talias.nim
zerbina: It's an internal test for the alias analysis in `sem/aliases.nim` using the undocumented `IsPartOf` magic.

tests/align/talign.nim
zerbina: Besides it's usage as a progress indicator during debugging, I don't think it serves any purposes, so it's removal is fine.

tests/align/talign.nim
zerbina: A `block:` (`static:` too) only opens a new scope, the content is still treated as top-level.

Top-level variables (globals) have various differences in semantics compared to local variables. For example, they're not captured in closures and are copied into `sink` arguments (instead of moved) by default.

canelhasmateus: TIL

zerbina:
> pinging @zerbina into reviews as requested :P

Review request accepted :D

> In my mind, these changes wouldy be better made in a second PR, but i didn't want to open a new one before closing this.
> Not sure if pulverizing these incoming changes over multiple PRs is better / worse for reviewing - please share your opinions.

I think splitting the `ambsym` category changes into a separate PR would be a good idea too. Smaller and separate PRs often allow for changes to be merged earlier

canelhasmateus:
> I think splitting the `ambsym` category changes into a separate PR would be a good idea too. Smaller and separate PRs often allow for changes to be merged earlier

Done. Removed the second commit from this one, and opened a second PR.
Do you mind reviewing / borsing this one @zerbina ?

tests/ccgbugs/tcodegen_bare_return.nim
```diff
@@ -0,0 +1,30 @@
+discard """
+labels: "codegen error_runtime exception"
+description: '''
+  . From https://github.com/nim-lang/Nim/issues/9286
+    Exception doesn't clear result value
+  . The second time should return none, but isSome(opt) is always equals to true
+  . Wrong heuristic in codegen, A bare return may trigger the insertion
```
zerbina:
```suggestion
  . Wrong heuristic in codegen. A bare return may trigger the insertion
```

tests/ccgbugs/tcodegen_char_cast_overflow.nim
```diff
@@ -0,0 +1,30 @@
...
+
+let c = '\128'
+
+block case1:
+  doAssert data[c.int] == 'y'
+  doAssert seq2[c.int] == 'y'
+
+
+block case2:
+   proc play(x: openArray[char]) =
```
zerbina:
```suggestion
  proc play(x: openArray[char]) =
```

tests/ccgbugs/tcodegen_distinct_uint64_32bit.nim
```diff
@@ -0,0 +1,20 @@
+discard """
+labels: "32_bit distinct var"
+description: '''
+  . From https://github.com/nim-lang/Nim/issues/13902
+    distinct uint64" type corruption on 32-bit, when using {.borrow.} operators
```
zerbina:
```suggestion
    "distinct uint64" type corruption on 32-bit, when using {.borrow.} operators
```

tests/ccgbugs/tcodegen_generic_openarray_import.nim
```diff
@@ -10,6 +10,13 @@ output: '''
 @[(v: -1), (v: 2), (v: 3)]
 @[(v: -1), (v: 2), (v: 3)]
 '''
+labels : "codegen generic import openArray"
+description: '''
+  . From https://github.com/nim-lang/Nim/issues/9578
+    openArray generates incorrect C code with "incomplete type"
+  . OpenArray[T] generates incorrect C code, if the containing module
```
zerbina:
```suggestion
  . openArray[T] generates incorrect C code, if the containing module
```

tests/ccgbugs/tcodegen_release_gcc_volatile.nim
```diff
@@ -0,0 +1,23 @@
+discard """
+matrix: "-d:release"
+labels: "code_gen gcc mode_release"
```
zerbina:
```suggestion
labels: "code_gen mode_release"
```

Not related to `gcc`.

tests/ccgbugs/tcodegen_signed_zero.nim
```diff
@@ -0,0 +1,16 @@
+discard """
+targets: "c js"
+labels: "backend_VM codegen floating_point"
```
zerbina:
```suggestion
targets: "c js"
labels: "codegen floating_point"
```

This issue is not related to the VM.

tests/ccgbugs/tcodegen_template_with_generic.nim
```diff
@@ -0,0 +1,30 @@
+discard """
+labels: "codegen error_compilation generics template"
+description: '''
+  . From https://github.com/nim-lang/Nim/issues/6756
+    Error in template when using the type of the parameter inside it
+  . The problem is in the line var res: A[T]. The assert fails because symbol is
+    A_105025: ["sfUsed"], [], ["issue.nim", 5, 15], skType
```
zerbina:
```suggestion
  . The problem is in the line `var res: A[T]`. The assert fails because symbol is
    `A_105025: ["sfUsed"], [], ["issue.nim", 5, 15], skType`
```

tests/ccgbugs/tgc_bug.nim
```diff
@@ -0,0 +1,16 @@
+discard """
+matrix: "--gc:refc -d:fulldebug -d:smokeCycles"
+output: '''@["a"]'''
+labels: "32_bit_only codegen gc gcc seq"
```
zerbina:
```suggestion
labels: "codegen gc seq"
```

The issue is not related to the C compiler or processor architecture.

tests/ccgbugs/xarray9578.nim
```diff
@@ -1,7 +1,8 @@
-import t9578
+import tcodegen_generic_openarray_import
 
 proc testArray*(x: var array[3,mytype]) =
   f(x[0].addr)
 
 proc testArray2*(x: var ptr array[3,mytype]) =
   f(x[0].addr)
+
```
zerbina, line 8: Some files end with a new-line and some don't. I think this should be done consistently, but I'm not sure which style is preferred.

zerbina (changes requested):
Some labels don't apply, but after that's fixed I'm good to merge.

I also left some suggestions for improved readability, but these are non-blocking.

zerbina:
Also, could you remove the template `Summary`/`Details` part from the PR message body?

tests/ccgbugs/tgc_bug.nim
canelhasmateus: In my [defense](https://github.com/nim-lang/Nim/issues/6279#issuecomment-324914791) :P

canelhasmateus:
There you go

zerbina (approved):
Nice, thank you. Changing the final new-line style (if wanted, that is) can be done in a separate PR.

bors r+

canelhasmateus:
This timeout is probably my fault. By accepting each of your suggestions separately i inadvertently kicked off multiple CI runs

canelhasmateus:
Very nice! Thank you for your attention.

PR #368, opened 2022-06-29
Ambsym

canelhasmateus:
More illustrative and detailed examples for the `ambsym`
category. We also add labels and descriptions for the
`actiontable` , `alias` , and `align` categories.

https://github.com/nim-works/nimskull/issues/41


canelhasmateus:
I think that ambsym is a suboptimal name.
IMHO, `identifier resolution` is a better name.

saem:
> I think that ambsym is a suboptimal name. IMHO, `identifier resolution` is a better name.

The original intention was `ambiguous symbol(s)`, which is perhaps better in that this doesn't look at all aspects of identifier resolution.

canelhasmateus:
how about a nice compromise in `symbol resolution`?

saem:
sure, it doesn't take much to beat `ambsym`

saem:
bors r+

thanks

PR #369, opened 2022-06-30
vm: clean up callback registration (`vmops`)

zerbina:
## Summary
- split up registration into multiple functions
- improve documentation in `vmops`
- adjust meaning of `vmopsDanger` option. All functions that can modify
 the host's environment are now no-ops by default
- `gorge` and `gorgeEx` are still available without
 specifying `experimental:vmopsDanger`.
- the operations registered by default when using the compiler API
 now only include the basic system and math ops

The split up makes callback registration more modular and allows
the newly introduced categories to be registered independently from
each other (or not at all).

This is an intermediate change, aimed at making callback handling
easier for the VM back-end.

## Details
Operations that are now no-ops without `vmopsDanger`:
- `putEnv`, `delEnv`

Operations that were no-ops without `vmopsDanger` but now aren't:
- `getCurrentDir`, `getTime`

### DIFF
--- a/compiler/front/scriptconfig.nim
+++ b/compiler/front/scriptconfig.nim
@@ -63,7 +63,7 @@ proc setupVM*(module: PSym; cache: IdentCache; scriptName: string;
   # For Nimble we need to export 'setupVM'.
   result = newCtx(module, cache, graph, idgen)
   result.mode = emRepl
-  registerAdditionalOps(result)
+  registerBasicOps(result[])
   let conf = graph.config
 
   # captured vars:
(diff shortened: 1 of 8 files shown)


zerbina:
I moved `gorgeEx` from `registerOs2Ops` to `registerCompileTimeOps` and also excluded it from the set of dangerous ops. Having to pass `experimental:vmopsDanger` to the compiler in order to use `gorgeEx` probably breaks quite some code and is also inconsistent with `gorge` (which doesn't require `vmopsDanger`).

Turning `gorgeEx` into a dangerous op is better done in a separate PR.

compiler/front/scriptconfig.nim
```diff
@@ -254,6 +254,20 @@ proc runNimScript*(cache: IdentCache; scriptName: AbsoluteFile;
...
+  # the VM instance used for NimScript execution is also used for the
+  # compile-time evaluation, so we have to register the macro/compile-time
+  # ops on the instance. Combined with how callback registration works,
+  # this has the following consequences:
+  # - CTFE and macro evaluation happening during semantic analysis of the
+  #  NimScript file are run with NimScript privileges, e.g. file-system write
+  #  access
+  # - the ``vmopsDanger`` option has no effect on callbacks registered
+  #  during `setupVM`
+  # - NimScript has access to the macro/compile-time APIs
```
saem, line 269: yup, stuck with this until we clarify compile time vs runtime, vm vs nimscript, host vs target, etc... oh well, that's ok for now.

compiler/front/scriptconfig.nim
```diff
@@ -254,6 +254,20 @@ proc runNimScript*(cache: IdentCache; scriptName: AbsoluteFile;
   var m = graph.makeModule(scriptName)
   incl(m.flags, sfMainModule)
   var vm = setupVM(m, cache, scriptName.string, graph, idgen)
+  let isDisallowDanger =
```
saem:
```suggestion
  let disallowDanger =
```

not strictly necessary, but it was done in `vm.nim` and it reads a tiny bit better.

saem (commented):
This is much nicer. 👍🏽 

I put in one minor suggestion but I'm also entirely OK to merge.

Feeling a bit inpsired so doing to take a look at that define matrix we discussed about targets, compile vs runtime, etc... and see if there is some progress to be made.

compiler/front/scriptconfig.nim
zerbina: Each NimScript execution uses a separate VM instance (since a separate `ModuleGraph` is used) and while that is quite a waste of resources, it at least means that the side-effects listed here don't carry over to normal compilation.

PR #370, opened 2022-07-01
Refactor `array` and `arithm` test categories

canelhasmateus:
Refactor `array` and `arithm` test category.

 
Added labels and descriptions for the `array` and `arithm`, as well as completely removing echo usage from the impacted tests.

https://github.com/nim-works/nimskull/issues/41


canelhasmateus:
@zerbina

tests/arithm/tarithm.nim
```diff
@@ -1,21 +1,30 @@
 discard """
+labels: "arithmetic bitwise conversion int range system"
+description: '''
+  . From https://github.com/nim-lang/Nim/issues/5216
```
zerbina, line 4: Hmm, since this test file contains both regression tests and general tests, only putting the issues' summaries in the description feels off. Maybe it'd be better to use a description like:
```
description: '''
  Tests for arithmetic operations and issues related to them.
  - issue link 1
  - issue link 2
  - etc.
'''
```
and then put the issue summary in a comment next to the test case? I'm not sure.
Alternatively, the regression tests could be moved into a separate `tarithm_issues.nim` file.

I don't consider this a blocker, so it's totally fine if it stays the way it currently is.

tests/array/tarraycons_idx_lit_3.nim
```diff
@@ -0,0 +1,18 @@
...
+  line: 16
+  labels: "array constructor index"
+  description: '''
+    . Its possible to sparsely specify arrays initial values.
+      In these cases, The following indexes are based of these.
+    . Here, we create an array of size 2^32. And set
+      ( 2^32 - 1 )th value as 1,
+      ( 2^32     )th value is inferred as 2,
+      ( 2^32 + 1 )th value is inferred as 3.
+    . We obviously exceed the array size, prompting an error.
```
zerbina:
```suggestion
    . Here, we create an array with 3 elements. The
        1st element is at index `2^31 - 1` (explicit)
        2nd element is at index `2^31 + 0` (inferred)
        3rd element is at index `2^31 + 1` (inferred)
    . the index value of the last element (2^31 + 1) doesn't fit
      into an int32, prompting an error.
```
Slight correction: `high(int32)` is '2^31' and the array has a size of '3'.

tests/array/tarraycons_ptr_generic2.nim
```diff
@@ -1,7 +1,15 @@
...
   file: "tarraycons_ptr_generic2.nim"
-  line: 17
+  line: 25
+  labels: "array generic ptr ref subtyping"
+  description: '''
+    . From https://github.com/nim-lang/Nim/issues/7601
+      array construction failed generic object of ptr relation subtype with
+    . note: ref object and ref generic is ok, only ptr generic failed
+    . this bug was found during https://github.com/nim-lang/Nim/issues/4799
+      investigation
```
zerbina: The issue description doesn't apply for this test, even though the latter was introduced as part of the PR fixing the mentioned issue. I'm not sure what the description should be, as the test is for a compiler implementation detail (it also succeeds without the changes in the corresponding PR).

Maybe it's best to leave it empty for now?

tests/array/tarraycons_idx_lit_2.nim
```diff
@@ -1,6 +1,10 @@
 discard """
   errormsg: "expected ordinal value for array index, got '\"string\"'"
-  line: 6
+  line: 10
   column: 10
+  labels: "array constructor index"
+  description: '''
+    . Arrays indexes should be ordinal values.
```
zerbina:
```suggestion
    . Array indices need to be ordinal values.
```
Small wording adjustment, since array indices being ordinal values is a requirement.

tests/array/tarraycons_enum.nim
```diff
@@ -0,0 +1,32 @@
+discard """
+  errormsg: "invalid order in array constructor"
+  file: "tarraycons_enum.nim"
+  line: 25
+  labels: "enum array constructor"
+  description: '''
+    . When specifying an array, it is necessary to inform its static size.
+    . Since enums are statically sized , it is possible to use them to specify
```
zerbina, line 8:
```suggestion
    . When creating an array, it's size needs to be known at compile-time.
    . Since enums are statically sized, it is possible to use them to specify
```
Makes it a bit easier to understand, I believe.

tests/array/tarraycons_idx_lit_3.nim
```diff
@@ -0,0 +1,18 @@
+discard """
+  errormsg: "size of array exceeds range of index type 'range 2147483646..2147483647(int32)' by 1 elements"
+  line: 16
+  labels: "array constructor index"
+  description: '''
+    . Its possible to sparsely specify arrays initial values.
+      In these cases, The following indexes are based of these.
```
zerbina:
```suggestion
    . It is possible to only specify the index of the first element
      and let the compiler infer the other elements' indices.
```
A bit more precise.

zerbina (commented):
Nice work :+1:. I left some wording suggestions/corrections and a thought regarding the `tarithm.nim` test.

`tarray.nim` is not reviewed yet.

tests/array/tarraycons_ptr_generic2.nim
canelhasmateus: there are plenty of cases like this - some implementation thrown together haphazardly without much in the way of "what" or "why". Pretty hard to add meaningful descriptions to these. 
In these cases, I'm just taking whatever steps i can gather , be from the commit log or any comments.

I agree with the sentiment that this description is sub-par and dis-honest. But the alternative is to give no information. 

I don't mind leaving it empty.

tests/array/tarraycons_idx_lit_3.nim
zerbina: Sorry, my mistake: `high(int32)` is `2^31 - 1`, so the first element is at `2^31 - 2` and not at `2^31 - 1`.

tests/array/tarray.nim
```diff
@@ -1,116 +1,124 @@
...
+     block:
+      var arr: array[-1, int]
+      ))
 
-  x = [4, 5, 6]
-  echo repr(varArr1(x))
 
-  y = x
-  echo repr(mul(x, y))
+    proc mul(a, b: TMyarray): TMyArray =
```
zerbina, line 135: A `block` seems to be missing here before the proc? The tests using `mul`, `arrayGet`, and `arrayId` are for issue https://github.com/nim-lang/Nim/issues/106.

tests/array/tarray.nim
```diff
@@ -1,116 +1,124 @@
...
+    let arr3: array[0, string] = []
+    var arr4: array[0, string] = []
 
-  var
-    x, y: TMyArray
-    o: TObj
+    doAssert(arr1.len == 0)
+    doAssert(arr2.len == 0)
+    doAssert(arr3.len == 0)
+    doAssert(arr4.len == 0)
```
zerbina, line 123: Did you add the `arr4` test here? I can't seem to find it in the previous version. It's no problem if it is, I just want to make sure.

tests/array/tarray.nim
```diff
@@ -1,116 +1,124 @@
...
-  # bug #1669
-  let filesToCreate = ["tempdir/fl1.a", "tempdir/fl2.b",
-              "tempdir/tempdir2/fl3.e", "tempdir/tempdir2/tempdir3/fl4.f"]
+    proc getPos(r: TMyRecord): int =
+      result = r.x + r.y
 
-  var found: array[0..filesToCreate.high, bool]
+    doAssert sum([1, 2, 3, 4]) == 10
+    doAssert sum([]) == 0
+    doAssert getPos( (x: 5, y: 7) ) == 12
```
zerbina: Regarding the "not_sure_what_this_tests": the first two seem to test passing constant `array` values to `openArray` parameters and the other one doesn't seem to be related to arrays at all (rather to tuples). These tests date back to the very first commit, so there doesn't exists more context unfortunately.

tests/array/tarray.nim
```diff
@@ -119,58 +127,70 @@ block tarraycons:
...
 
-  type
-    Vehicle[T] = object of RootObj
-      tire: T
-    Car[T] = object of Vehicle[T]
-    Bike[T] = object of Vehicle[T]
+    var i = 0
+    for current in fruits:
+      doAssert current.name == fruits[i][].name
+      i += 1
```
zerbina:
```suggestion
    var names: seq[string]
    for it in fruits:
      names.add(it.name)
    doAssert names == ["apple", "banana", "Fruit"]
```
The `current.name == fruits[i][].name` assertion doesn't catch if the pointers point to the wrong object somehow.

tests/array/tarray.nim
```diff
@@ -119,58 +127,70 @@ block tarraycons:
...
+        book = Book[string](cover: "none")
+        hardBook = Hard[string](cover: "skin")
+        softBook = Soft[string](cover: "paper")
+    let books = [book, hardBook, softBook]
 
 
+    var i = 0
+    for current in books:
+      doAssert current.cover == books[i].cover
+      i += 1
```
zerbina: Same suggestion as above.

tests/array/tarray.nim
```diff
@@ -190,70 +210,85 @@ block tarraylen:
...
+    proc fromOpenArray(x: openArray[int]) : int =
+      return x[^2]
 
-  proc mutOpenarray(x: var openArray[string]) =
-    x[^2] = "new one"
+    doAssert fromOpenArray([1, 2, 3]) == 2
 
-  useOpenarray([1, 2, 3])
+  block:
+    proc mutOpenarray(x: var openArray[string] , val : string) =
```
zerbina:
```suggestion
    proc mutOpenarray(x: var openArray[string], val: string) =
```
Tiny code-style correction.

tests/array/tarray.nim
```diff
@@ -298,12 +333,12 @@ block troofregression:
     MySeq[T] = ref object
       data: seq[T]
 
-  proc test[T](sx: MySeq[T]) =
+  proc test[T](sx: MySeq[T]) : T =
```
zerbina:
```suggestion
  proc test[T](sx: MySeq[T]): T =
```

tests/array/tarray.nim
```diff
@@ -536,19 +566,20 @@ block t7818:
...
 block trelaxedindextyp:
   # any integral type is allowed as index
-  proc foo(x: ptr UncheckedArray[int]; idx: uint64) = echo x[idx]
-  proc foo(x: seq[int]; idx: uint64) = echo x[idx]
-  proc foo(x: string|cstring; idx: uint64) = echo x[idx]
-  proc foo(x: openArray[int]; idx: uint64) = echo x[idx]
+  proc foo(x: ptr UncheckedArray[int]; idx: uint64) : int = x[idx]
+  proc foo(x: seq[int]; idx: uint64) : int =  x[idx]
+  proc foo(x: string|cstring; idx: uint64) : int = x[idx]
+  proc foo(x: openArray[int]; idx: uint64) : int = x[idx]
```
zerbina:
```suggestion
  proc foo(x: ptr UncheckedArray[int]; idx: uint64): int = x[idx]
  proc foo(x: seq[int]; idx: uint64): int =  x[idx]
  proc foo(x: string|cstring; idx: uint64): int = x[idx]
  proc foo(x: openArray[int]; idx: uint64): int = x[idx]
```

tests/array/tarray.nim
```diff
@@ -562,20 +593,16 @@ block arrayLiterals:
 
 
 block t8316:
-  # https://github.com/nim-lang/Nim/issues/8316
   proc myAppend[T](a:T):string=
-    echo "a:", a
```
zerbina, line 567: The test also needs to make sure that `myAppend` is only called three times as was done previously via the testament output comparsion.

I'd suggest a simple `int` counter that gets incremented in `myAppend` and is asserted to be `3` after the `append2` assert.

tests/array/tarray.nim
```diff
@@ -562,20 +593,16 @@ block arrayLiterals:
 
 
 block t8316:
```
zerbina: For consistency, this should be `block issue_8316:` I believe?

tests/array/tarray.nim
```diff
@@ -562,20 +593,16 @@ block arrayLiterals:
...
     for a in args:
-      echo "ret:", ret
       ret.add(a)
     ret
 
-  let foo = append2("1", "2", "3")
-  echo foo
+  doAssert append2("1", "2", "3") == "123"
 
 block t12466:
```
zerbina: Rename the lables of this and following blocks to `issue_x`?

You probably just didn't get to it yet, but the issue links should likely be moved into the description as is done with the others, right?

tests/array/tarray.nim
```diff
@@ -190,70 +210,85 @@ block tarraylen:
...
-  takeA(addr c[^1][0])
-  c[^1][1] = 5
-  echo c
+    proc deref(x: ptr int) : int = return x[]
+    doAssert deref(addr c[^1][0]) == 3
 
-  proc useOpenarray(x: openArray[int]) =
-    echo x[^2]
+  block:
+    proc fromOpenArray(x: openArray[int]) : int =
```
zerbina:
```suggestion
    proc fromOpenArray(x: openArray[int]): int =
```

tests/array/tarray.nim
```diff
@@ -190,70 +210,85 @@ block tarraylen:
...
 
-  proc takeA(x: ptr int) = echo x[]
+    var c: seq[(int, int)] = @[(1,2), (3,4)]
+    c[^1][1] = 5
+    doAssert c == @[(1, 2), (3, 5)]
 
-  takeA(addr c[^1][0])
-  c[^1][1] = 5
-  echo c
+    proc deref(x: ptr int) : int = return x[]
```
zerbina:
```suggestion
    proc deref(x: ptr int): int = x[]
```
In general, there's no space inserted between the parenthesis and the colon. If a function only consists of a single expression, the `return` is also omitted most cases.

tests/array/tarray.nim
```diff
@@ -119,58 +127,70 @@ block tarraycons:
...
 
-  type
-    Book[T] = ref object of RootObj
-      cover: T
-    Hard[T] = ref object of Book[T]
-    Soft[T] = ref object of Book[T]
+    var i = 0
+    for current in vehicles:
+      doAssert current.tire == vehicles[i][].tire
+      i += 1
```
zerbina: Same suggestion as above, just with a `seq[int]`.

zerbina (changes requested):
Finished the review of `tarray`. Again, sorry for the delay.

There are some tests cases that don't test the same thing as they did before, plus some minor issues that need addressing, but after those are fixed, I'm good to merge.

Also, thank you for giving functions and variables more descriptive names :)

tests/array/tarraycons_ptr_generic2.nim
zerbina: I've thought about it for a bit and I think the following description fits for the test:
```
Defensive test to make sure that the fix for https://github.com/nim-lang/Nim/issues/7601 doesn't break unrelated code.
```

tests/array/tarraycons_idx_lit_3.nim
canelhasmateus: `high(int32)` is not `2^32 - 1`?

tests/array/tarraycons_ptr_generic2.nim
canelhasmateus: I'll be adding this to the description.

tests/array/tarray.nim
canelhasmateus: Yes, i added it. It seemed logical since we were already testing for const and let arrays.

tests/array/tarray.nim
canelhasmateus: I'm moving the openArray to its own block and renaming this block to `leftoverTests`. I'm a fan of letting things that don't really gel together have explicitly bad names, since it eases further discussions about the validity of those. Feel free to suggest another approach.

canelhasmateus:
Added the requested changes. Sorry for the delay, i was travelling.

tests/array/tarraycons_idx_lit_3.nim
zerbina: No. `high(uint32)` is `2^32 - 1`, but `high(int32)` is not.

tests/array/tarraycons_idx_lit_3.nim
```diff
@@ -0,0 +1,19 @@
...
+  line: 17
+  labels: "array constructor index"
+  description: '''
+    . It is possible to only specify the index of the first element
+      and let the compiler infer the other elements' indices.
+    . Here, we create an array with 3 elements. The
+        1st element is at index `2^31 - 1` (explicit)
+        2nd element is at index `2^31 + 0` (inferred)
+        3rd element is at index `2^31 + 1` (inferred)
+    . the index value of the last element (2^31 + 1) doesn't fit
```
zerbina:
```suggestion
        1st element is at index `2^31 - 2` (explicit)
        2nd element is at index `2^31 - 1` (inferred)
        3rd element is at index `2^31 + 0` (inferred)
    . the index value of the last element (2^31 + 0) doesn't fit
```

zerbina (commented):
> Added the requested changes. Sorry for the delay, i was travelling.

No worries :)

After the off-by-one issue with the one issue description is fixed (sorry for my first suggestion being wrong), I'm good to merge.

tests/array/tarraycons_idx_lit_3.nim
zerbina: Hmm, you marked the suggestion as resolved, but it wasn't applied yet. Maybe I'm missing something?

canelhasmateus: damned be ammends and staging areas

zerbina (approved):
bors r+

canelhasmateus:
what has happened here? bors broke?

zerbina:
The `bors` issue seems to be related to PRs that were moved out of the `draft` state. Closing and re-opening the PRs helped in the past.

bors r+

PR #371, opened 2022-07-02
internals: semexprs rewrites to case/let style

saem:
## Summary
- rewrite a few more procs in the case/let style
- many of the procs in semexprs are now converted
- next steps is cleaning up other modules similarly

## Details
- tuple constructor and par handling cleaned up
- eliminated the need for semexprs.checkPar
- dotTransformation is now nkError ready
- semTupleConstr now handles all tuple construction

### Misc Changes
- added sym id to trace output
- added in more tracing into various procs
- removed step kind in tracing, too much spam
- explain output in ill formed error messages

next steps is cleaning up other modules similarly

---

## Notes for Reviewers
* the rest doesn't clean-up all that easily
* this should become easier after other files are cleaned and the APIs can change

### DIFF
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -577,7 +577,7 @@ template semReportIllformedAst*(
   handleReport(
     conf,
     wrap(
-      SemReport(kind: rsemIllformedAst, ast: node ),
+      SemReport(kind: rsemIllformedAst, ast: node, str: explain),
       instLoc(),
       node.info),
     instLoc(),
@@ -590,12 +590,13 @@ proc joinAnyOf*[T](values: seq[T], quote: bool = false): string =
     else:
       s
 
-  if len(values) == 1:
+  case len(values)
+  of 1:
     result.add q($values[0])
-  elif len(values) == 2:
+  of 2:
     result.add q($values[0]) & " or " & q($values[1])
   else:
-    for idx in 0 ..< len(values) - 1:
+    for idx in 0..<len(values) - 1:
       if idx > 0:
         result.add ", "
       result.add q($values[idx])
@@ -603,18 +604,17 @@ proc joinAnyOf*[T](values: seq[T], quote: bool = false): string =
     result.add " or "
     result.add q($values[^1])
 
-template semReportIllformedAst*(
-  conf: ConfigRef, node: PNode, expected: set[TNodeKind]): untyped =
+template createSemIllformedAstMsg*(node: PNode,
+                                   expected: set[TNodeKind]): string =
   var exp: seq[TNodeKind]
   for e in expected:
     exp.add e
 
-  var msg = "Expected "
-  msg.add joinAnyOf(exp)
-  msg.add ", but found "
-  msg.add $node.kind
+  "Expected $1, but found $2" % [joinAnyOf(exp), $node.kind]
 
-  semReportIllformedAst(conf, node, msg)
+template semReportIllformedAst*(
+  conf: ConfigRef, node: PNode, expected: set[TNodeKind]): untyped =
+  semReportIllformedAst(conf, node, createSemIllformedAstMsg(node, expected))
 
 template localReport*(conf: ConfigRef, info: TLineInfo, report: ReportTypes) =
   handleReport(conf, wrap(report, instLoc(), info), instLoc(), doNothing)
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -442,7 +442,8 @@ proc isOpImpl(c: PContext, n: PNode, flags: TExprFlags): PNode =
   ## implements `is`, for `x is Y` where x is an expression and `Y` is a type
   ## or an expression whose type is compared with `x`'s type.
   c.config.internalAssert:
-    n.len == 3 and n[1].typ != nil and n[2].kind in {nkStrLit..nkTripleStrLit, nkType}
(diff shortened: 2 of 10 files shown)


compiler/front/msgs.nim
```diff
@@ -590,31 +590,31 @@ proc joinAnyOf*[T](values: seq[T], quote: bool = false): string =
...
 
-  if len(values) == 1:
+  case len(values)
+  of 1:
     result.add q($values[0])
-  elif len(values) == 2:
+  of 2:
     result.add q($values[0]) & " or " & q($values[1])
   else:
-    for idx in 0 ..< len(values) - 1:
```
Clyybber, line 598: I think `..<` is preferable even with a suffixed `-1` since it's clearer that it's `1` last element we skip.

compiler/sem/semexprs.nim
```diff
@@ -1867,26 +1871,48 @@ proc takeImplicitAddr(c: PContext, n: PNode; isLent: bool): PNode =
...
+        
+        if x.sym.kind == skResult and varLentOrView:
+          n[0] = x # 'result[]' --> 'result'
+          n[1] = takeImplicitAddr(c, ri, x.typ.kind == tyLent)
+          x.typ.flags.incl tfVarIsPtr
+          #echo x.info, " setting it for this type ", typeToString(x.typ), " ", n.info
+        elif sfGlobal in x.sym.flags:
+          x.typ.flags.incl tfVarIsPtr
+      else:
+        discard
```
Clyybber: This is not equivalent to the original logic, if you have `x.sym.kind != skResult and sfGlobal in x.sym.flags` the old logic would still do `x.typ.flags.incl tfVarIsPtr`, the new one would lead to `discard`.
For this proc I think the old version is more readable.

compiler/sem/semexprs.nim
```diff
@@ -2094,32 +2120,53 @@ proc semProcBody(c: PContext, n: PNode): PNode =
...
+          of nkExprColonExpr:
             tupleConstr[i][1] = takeImplicitAddr(c, tupleConstr[i][1], e.kind == tyLent)
           else:
             tupleConstr[i] = takeImplicitAddr(c, tupleConstr[i], e.kind == tyLent)
         else:
-          semReportIllformedAst(c.config, n[0], {nkPar, nkTupleConstr})
+          hasError = true
+          tupleConstr[i] = c.config.newError(result,
+            reportAst(rsemIllformedAst, result,
+                      createSemIllformedAstMsg(n, {nkPar, nkTupleConstr})))
```
Clyybber, line 2161:
```suggestion
                      createSemIllformedAstMsg(tupleConstr, {nkPar, nkTupleConstr})))
```

Clyybber: This might create cyclic ast I think.

compiler/front/msgs.nim
haxscramper: Yes, I also think `..<` is more approachable. Or use `.. high(values) - 1`, because we have a specific "last existing index" function for almost all containers anyway.

compiler/sem/semexprs.nim
```diff
@@ -2687,17 +2744,32 @@ type
...
+    result =
+      case n.kind
+      of nkTupleConstr:
+        case n[0].kind
+        of nkExprColonExpr:
+          paTupleFields
+        else:
+          paTuplePositions
+      else:
+        paSingle         # (expr)
```
Clyybber: Subtle change in logic, but doesn't matter since https://github.com/nim-lang/Nim/pull/13793 .
`paSingle` can be removed for example, since the disambugation can now happen in `semExpr` by checking wether it's `nkPar` or `nkTupleConstr`.
Then `checkPar` would only be needed to be called in the `nkTupleConstr` case, which means we don't need to check for `nkTupleConstr` here anymore, merging some of the logic of the `of 1` with the `else` case.
The whole logic `checkPar` could just be put into `semTupleConstr` which should then call out to `semTupleFieldsConstr` and `semTuplePositionsConstr`. Then we can also return proper nkError nodes in the `paNone` cases.

saem: That's so much better, decide once for the win. I'm hoping cleanup like this makes this stuff more obvious to spot. Adding that to my list of stuff to spruce up today.

compiler/sem/semexprs.nim
saem: Whoops, that is totally wrong, yeah it should be at the same level as the node being assigned. Nice catch

compiler/sem/semexprs.nim
saem: Boo old style, this is tired me making an obvious error, let me see what another go might like like.

Interesting how no tests failed...

compiler/sem/semexprs.nim
saem: made this change.

turns out there are still a number of places where nkPar with mroe than one element are passed in. I purposefully left the code ugly to make it easier to revisit.

compiler/sem/semexprs.nim
saem: fixed

compiler/sem/semexprs.nim
saem: had another go at it, bit cleaner, but I still think this code and its collaborators need a revisit.

compiler/front/msgs.nim
saem: done.

PR #373, opened 2022-07-03
update astrepr and documentation

haxscramper:
- Optionally show kind types in the tree repr representation, to make it
  easier to differentiate between symbol kinds, node kinds and type kinds
- allow showing `<nil>` fields
- Make it possible to hide non-nil type AST
- Make it possible to request a full symbol types
- Optionally print flag names that will control certain field showing
- If verbose symbol printing is not enabled, print the name as a quoted
  string
- Add text layouter docs, missed them in earlier commit

https://github.com/nim-works/nimskull/discussions/372#discussioncomment-3070575


saem (approved):
bors r+

PR #374, opened 2022-07-03
add AST builder procs

haxscramper:
initial collection of the procedure to see whether the idea is useful
enough to extend it further.

per earlier discussion in matrix, @saem 

> yeah, this is closer, I think the API should feel a bit like you're writing jsx in react. You simply write the tree as it should be produced, in a single expression. Breaking it up would mean function call in the expression but otherwise just "draw" it like it should be.

### DIFF
--- a/compiler/ast/ast_build.nim
+++ b/compiler/ast/ast_build.nim
@@ -1,0 +1,80 @@
+## This module provides simplified procedures for the AST construction
+
+import ast, lineinfos
+
+proc newIf*(cond, body: PNode, orElse: PNode = nil): PNode =
+  ## Construct new `if` statement
+  result = newTree(nkIfStmt, newTree(nkElifBranch, cond, body))
+  if not isNil(orElse):
+    result.add newTree(nkElse, orElse)
+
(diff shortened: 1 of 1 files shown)


compiler/ast/ast_build.nim
```diff
@@ -0,0 +1,80 @@
+## This module provides simplified procedures for the AST construction
+
+import ast, lineinfos
+
+proc newIf*(cond, body: PNode, orElse: PNode = nil): PNode =
```
saem, line 5: Hmm, this brings up an interesting question, do we have more expressions or statements, and also what should we favor?

saem (changes requested):
This could work for untyped AST building, the one missing piece is line information handling. Without line info it either won't compose or we'll have tons of issues were it was missed.

For typed we'd need to consider typ and friends, of course. I think we should cover that separately. For now can we prefix this API so we know it's untyped and we leave room for a typed version?

haxscramper:
I temporarily converted this into a draft PR, I will return to it later on when I have more time to (approximately July 22nd or later)

PR #375, opened 2022-07-03
initial implementation of the descriptive `[]`

haxscramper:
Initial implementation to test validity/usability of the more
descriptive accessor procedures: instead of writing `[0..^3]`
`SliceAllIdents` can be used.

Right now rough estimates of the `\[[0-9]+\]` usage total to 5498
instances in the `compiler/` directory, with 394 in VM, 2359 in `sem/`
etc.

This PR provides core part/implementation that should be expanded as it
is introduced in the different parts of the compiler.

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -663,3 +663,73 @@ proc skipAddr*(n: PNode): PNode {.inline.} =
 proc isNewStyleConcept*(n: PNode): bool {.inline.} =
   assert n.kind == nkTypeClassTy
   result = n[0].kind == nkEmpty
+
+
+
+type
+  NodePosName* = enum
+    ## Named node position accessor
+    PosLastIdent ## Last item in the identifier
+    PosType ## Type position for variable declaration or a procedure. For
+            ## procedure returns it's formal parameters (signature)
+    PosInit ## Initialization expression
+    PosProcBody ## Procedure body
+    PosProcReturn ## Return value of the procedure in formal parameters
+    PosProcArgs ## Formal parmeters in the procedure
+    PosTypeBody ## Type definition body
+    PosName ## Procedure name
+    PosBody ## Generic statement body
+    PosPragma ## Pragma position in the node body
+
+  NodeSliceName* = enum
+    ## Named node slice accessor
+    SliceAllIdents ## All identifiers in the ident defs
+    SliceAllArguments  ## All arguments in the formal parammeters
+    SliceAllBranches ## All case statement branches
+    SliceBranchExpressions ## All expressions in the `of` branch
+
+proc `[]`*(node: PNode, pos: NodePosName): PNode =
+  ## Get subnode by named position
+
+  # NOTE: further expansion and clarifications of this node should happen
+  # on as-needed basis. When adding clarifications for accessing different
+  # *node kinds* please do update the documentation on the respective named
+  # position enum values as well.
+  case pos:
+    of PosLastIdent: node[^3]
+    of PosType:
+      case node.kind:
+        of routineDefs:
+          node[3]
+
+        else:
+          node[^2]
+
+    of PosInit: node[^1]
+    of PosProcBody: node[6]
+    of PosTypeBody: node[2]
+    of PosProcArgs: node[3]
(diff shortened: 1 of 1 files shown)


compiler/ast/ast_query.nim
```diff
@@ -662,4 +662,50 @@ proc skipAddr*(n: PNode): PNode {.inline.} =
 
 proc isNewStyleConcept*(n: PNode): bool {.inline.} =
   assert n.kind == nkTypeClassTy
-  result = n[0].kind == nkEmpty
\ No newline at end of file
+  result = n[0].kind == nkEmpty
+
+
+type
+  NamedNodePos* = enum
```
```suggestion
  NamedNodePos* = enum
    ## nodes that have a name like proc definitions, identifier definitions in let sections, etc
```

I think that captures the intention?

compiler/ast/ast_query.nim
```diff
@@ -662,4 +662,50 @@ proc skipAddr*(n: PNode): PNode {.inline.} =
...
+    PosName ## Procedure name
+    PosBody ## Generic statement body
+
+  NamedNodeSlice* = enum
+    SliceAllIdents ## All identifiers in the ident defs
+    SliceAllArguments  ## All arguments in the formal parammeters
+    SliceAllBranches ## All case statement branches
+    SliceBranchExpressions ## All expressions in the `of` branch
+
+func `[]`*(node: PNode, pos: NamedNodePos): PNode =
```
saem: I think this should dispatch on n.kind then process the pos?

So PosType can be the return type for a proc (not a silly default) and for var let const it's the second last.

Also for procs the last one isn't always the body iirc, I think we put things on the end in certain cases getting a position 7 and 8. Can't recall the circumstances

compiler/ast/ast_query.nim
saem: A pragma pos would be nice, the index varies per kind but it'd be super helpful.

saem (commented):
Added some thoughts.

After seeing it, I see the potential and two things seem key:
1. Careful naming so it's very evident what they do, this will be the first thing people read all over sem etc
2. Dispatch on node kind so it's not merely indirection but meaningful traversal abstraction, encodes node structural characteristics (a bit like lenses)

compiler/ast/ast_query.nim
haxscramper: Pragma position for different entry types, like procedure, type declaration etc?

saem: Yup

saem:
Only way I can think of providing deep meaningful feedback is by using it, I can see that working on one of two ways (feel free to suggest others):

1. I pull down the PR and try it in a few scenarios and come back with feedback, likely this weekend
2. We move the code out of a general module like ast_query and put in into a module where we'd use it a bunch (ast_repr?), then extract in a future PR
3. We merge now and give it a shot, but don't have everyone try all at once in the beginning

The point of either is to see any things that might be less apparent than just thinking about it -- what I'm currently doing.

I personally prefer 2, but I'm good either way you want to roll.

haxscramper:
I want to merge now and start using it, and I never had any intentions to "have everyone try all at once in the beginning".

haxscramper:
And I have absolutely no idea why these tests fail

saem:
It's the tcan_compile_nim test, one of the procs can have side-effects so it's failing.

compiler/ast/ast_query.nim
```diff
@@ -662,4 +662,74 @@ proc skipAddr*(n: PNode): PNode {.inline.} =
...
 proc isNewStyleConcept*(n: PNode): bool {.inline.} =
   assert n.kind == nkTypeClassTy
-  result = n[0].kind == nkEmpty
\ No newline at end of file
+  result = n[0].kind == nkEmpty
+
+
+
+type
+  NamedNodePos* = enum
```
saem: In this or a follow-up PR, can we rename the enum from `NamedNodePos` to `NodePosName` and `NamedNodeSlice` to `NodeSliceName`.

saem: If this can be covered in a follow-up PR (as part of actually using it) then let's merge this one, alternatively it can go into this PR.

PR #376, opened 2022-07-05
compiler/front: add depfile option

CyberTailor:
### Abstract

Add a new command-line option (e.g. `--depfile`) for the `nim compile` command. It accepts an argument (output path for a depfile, generated for a given target).

### Motivation

This change would improve integrations with language-independent build
systems. Particularly, get better edit-compile cycle times.

Currently a build system needs to rebuild a binary if _any_ of source files has
been changed, even if they aren't used by the binary. The full list of files
that a given source file depends on can only be discovered by the compiler.

It doesn't need to be a standalone command because if the file has never been
compiled, it must be built anyway, generating the header dependencies as a side
effect.

### Depfile file format

From GCC manual (`-M` option):

> The preprocessor outputs one 'make' rule containing the object file name for
> that source file, a colon, and the names of all the included files, including
> those coming from '-include' or '-imacros' command-line options.

Or, in the BNF notation:  
https://cmake.org/cmake/help/latest/command/add_custom_command.html

It is recognized by all common build systems.

### Before
*(no command to produce a depfile)*

### After
```sh
$ nim c --depfile main.d main
$ head -4 main.d # system imports not shown
main: \
	/home/project/main.nim \
	/home/project/src/file1.nim \
	/home/project/src/file2.nim \
```

#### Use the depfile in Makefile

```Makefile
main: main.nim
	nim c $(NIMFLAGS) --depfile $@.d $<

-include main.d
```

#### Use the depfile in Ninja

```Ninja
rule nimc:
  deps = gcc
  depfile = $out.d
  command = nim c $nimflags --depfile $out.d $in

build main: nimc main.nim
```

#### Use the depfile in CMake

```CMake
...

add_custom_command(
    OUTPUT main
    COMMAND nim c ${NIMFLAGS} --depfile main.d main.nim
    DEPFILE main.d
    VERBATIM
)
```

#### Use the depfile in Meson

```Meson
...

nim = find_program('nim')
custom_target(
  input : 'main.nim',
  output : 'main',
  depfile : '@BASENAME@.d',
  command : [nim, 'c', '--depfile', '@DEPFILE@', '@INPUT@'],
  build_by_default : true
)
```

### Backwards Compatibility

This feature introduces no backward-incompatible changes.

### DIFF
--- a/compiler/front/depfiles.nim
+++ b/compiler/front/depfiles.nim
@@ -1,0 +1,37 @@
+#
+#    See the file "copying.txt", included in this
+#    distribution, for details about the copyright.
+#
+
+## Functions for writing target's dependencies in various formats.
+
+import
+  std/[strutils, tables]
+
+import
+  compiler/utils/[pathutils],
+  compiler/modules/[modulegraphs],
+  compiler/ast/[ast_types],
+  compiler/front/[msgs, options]
+
+proc writeDepsFile*(g: ModuleGraph) =
+  let fname = g.config.nimcacheDir / RelativeFile(g.config.projectName & ".deps")
+  let f = open(fname.string, fmWrite)
+  for m in g.ifaces:
+    if m.module != nil:
+      f.writeLine(toFullPath(g.config, m.module.position.FileIndex))
+  for k in g.inclToMod.keys:
+    if g.getModule(k).isNil:  # don't repeat includes which are also modules
+      f.writeLine(toFullPath(g.config, k))
+  f.close()
+
+proc writeGccDepfile*(depfile: File, target: string, paths: openArray[string]) =
+  ## Outputs one `make` rule containing target's file name, a colon, and the
+  ## names of all specified paths. Spaces and hashes are escaped.
+  let target = target.multiReplace((" ", "\\ "), ("#", "\\#"))
+  depfile.write(target & ": \\" & '\n')
+  for it in paths:
+    let path = it.multiReplace((" ", "\\ "), ("#", "\\#"))
+    if path.len == 0:
+      continue
+    depfile.write('\t' & path & " \\" & '\n')
--- a/changelog.md
+++ b/changelog.md
@@ -84,3 +84,6 @@ becomes an alias for `addr`.
 
 - There is a new switch `--nimMainPrefix:prefix` to influence the `NimMain` that the
   compiler produces. This is particularly useful for generating static libraries.
+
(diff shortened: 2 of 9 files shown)


saem:
@CyberTailor, welcome to the community and thanks for the PR.

I'm ok with the change overall, before merging is like two changes please:

1. tests, at least one
2. Decouple our history from Nim's

---

The latter is one that needs some background so let me expand upon that here. We are due for a rename, this might be sweeping possibility beyond repo name alone.

I need our data be more and more in our repo, issues, discussion etc to facilitate a clean migration, preservation of history, etc. In particular the source of truth for such decisions, especially as captured in our commit history needs to be clean in this regard.

Here are some concrete suggestions as to how to accomplish this:
- use the content of the RFC, with some touch ups, for this PR and commit message
- start a discussion (GitHub discussions) and then summarize from there
- something else that is equivalent, but the litmus test is someone should be able to understand what and why while remaining in this repo 

----

Again, thanks for the PR, much appreciate you taking the time to put it together, let me know if you have any questions.

haxscramper:
Also want to thank for the PR, it would certainly be pretty useful for things like https://github.com/haxscramper/hcparse/issues/20 and general integration of the nim project into C/C++ build system properly, instead of a `glob(nimcache/*.c)` and like.

------------

On the commit/history side mentioned by @saem - clarification from my side: since the commit itself introduces a *finished* feature, you can just re-use the Abstract and Motivation side of your readme. Also, I think it would be fitting to put examples of usage from your RFC (I think they can be used as-is, since they are already pretty good) into permanent documentation, so people can easily refer to them when they need it. You can put it in the compiler user guide (`nimc.rst`)

CyberTailor:
> tests, at least one

My code needs some refactoring before I can enable tests (so it can use relative paths).

> Decouple our history from Nim's

Done.

CyberTailor:
Added a test.

compiler/front/depfiles.nim
```diff
@@ -0,0 +1,41 @@
+#
+#
+#           The Nim Compiler
+#        (c) Copyright 2015 Andreas Rumpf
```
saem: I missed this, sorry. Please remove these two lines.

After that it's good to go.

saem:
Bors r+

Do the dance Bors.

Congrats on the first PR to our little repo, @CyberTailor! 🎉

CyberTailor:
Done.

PR #377, opened 2022-07-06
doc: add missing documentation for `--depfiles`

haxscramper:
Add missing documentation from the body of the 
https://github.com/nim-works/nimskull/pull/376 into compiler documentation guide 
that can be read by the users directly, without digging through the GitHub 
discussions.


haxscramper:
minor documentation update that includes already written & committed examples from the RFC/PR mentioned in the commit message

PR #378, opened 2022-07-10
stdlib: prepare `system` and `io` for the VM target

zerbina:
Replace `defined(nimscript)` in `when` statements with the new
`isNimVmTarget` constant. Only the `isNimVmTarget` definition needs to
be modified as part of the VM target introduction commit then.

### DIFF
--- a/lib/std/private/digitsutils.nim
+++ b/lib/std/private/digitsutils.nim
@@ -46,7 +46,7 @@ func addChars[T](result: var string, x: T, start: int, n: int) {.inline.} =
     for i in 0..<n: result[old + i] = x[start + i]
   when nimvm: impl
   else:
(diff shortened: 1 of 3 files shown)


lib/std/private/digitsutils.nim
```diff
@@ -46,7 +46,7 @@ func addChars[T](result: var string, x: T, start: int, n: int) {.inline.} =
     for i in 0..<n: result[old + i] = x[start + i]
   when nimvm: impl
   else:
-    when defined(js) or defined(nimscript): impl
+    when not declared(copyMem): impl
```
saem, line 49: I personally prefer flipping clauses to remove negations, I understand that this keeps the change smaller.

saem (approved):
Bors r+

Yay!

PR #379, opened 2022-07-11
Internals: simplify let var section analysis

saem:
## Summary

- let and var section semantics clarified:
  - section normalization rules
  - macro pragmas now allow for a complete rewrite
  - nkError generation
  - normalize symbol pragma data storage
- block semantic analysis uses case/let style
- stmt list semantic analysis clarified:
  - collapses nested stmtLists, unless a block arg
- pragma block analysis less likely to break defer
- number of other fixes

## Details

let and var section semantics clarifications:
- section asts are normalized per identdef and var tuple
- macro pragmas are processed on the normalized ast
- macro pragmas receive one normalized section at a time
- then each normalized section is semantically analysed
- identifier related pragmas are processed here
- other pragmas remain and are processed on the symbol during codegen
- identdef symbols always have a pragmaexpr, even if empty

let and var section internal change remarks:
- semVarOrLet renamed to semLetOrVar
- uses nkError exclusively, resulting in some cascade changes
- production nodes (result) are now entirely distinct from input nodes
  this results in more allocations but it'll make DOD easier in the end
- introduced `ast.newTypeError` to create error types with the `n` field
  containing an nkError, establishing a tyError convention
- documented analysis in proc comment

cascade internal changes:
- vmgen and transf handling of let/var sections and normalized data
- `msgs.illformedAstReport` introduced to stop forced `localReport`

cascade changes template evaluation:
- template evaluation watches and reports nkErrors -- not ideal

cascade changes block semantic analysis:
- `semBlock`: now produces nkErrors
- production and input nodes are now separated
- added some docs

cascade changes stmtlist semantic analysis:
- `semStmtList`: now produces nkErrors
- production and input nodes are now separated
- flattens nested stmts, unless a `do:` `nfBlockArg`
- added some docs

cascade changes type sections:
- semTypeNode and semTypeOf/2 are becoming nk/tyError aware

cascade changes semPragmaBlock:
- pragmas blocks are less likely to screw up defer processing
- pragmas block items disappear after processing from the production
- updated `exception/tdefer1` test with a `{.line.}` pragma for this

miscellaneous changes:
- adding tracing to more sem procs (block, stmtlist, when, const expr, ...)
- added many more asserts and checks to clarify input expectations
- tracing output is more compact and also faster
- introduced `ast.newSymNode2` that is skError aware, future refactors
  will likely have this used nearly everywhere in the compiler
- slightly more precise error messages for let and var sections
- fixed bug in astrepr leading to NPE for type printing
- fixed bug so error nodes now generate node ids
- fixed bug in stdlib.macros to handle empty nkPragmaExprs

next steps for later:
- bring semConstSection in line with this section
- `nfBlockArg` is a hack, should be removed for a node
- error msg in `errmsgs/twrongcolon` wasn't good before not great now

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -494,6 +494,8 @@ proc isOpImpl(c: PContext, n: PNode, flags: TExprFlags): PNode =
   result.info = n.info
 
 proc semIs(c: PContext, n: PNode, flags: TExprFlags): PNode =
+  addInNimDebugUtils(c.config, "semIs", n, result, flags)
+
   if n.len != 3:
     result = c.config.newError(n, semReportCountMismatch(
       rsemWrongNumberOfArguments, 1, 2, n))
@@ -505,8 +507,11 @@ proc semIs(c: PContext, n: PNode, flags: TExprFlags): PNode =
 
   n[1] = semExprWithType(c, n[1], flags + {efDetermineType, efWantIterator})
 
-  if n[2].kind in {nkStrLit..nkTripleStrLit}:
+  case n[2].kind
+  of nkStrLit..nkTripleStrLit:
     n[2] = semExpr(c, n[2])
+  of nkError:
+    discard # below we'll wrap the result in an error
   else:
     let t2 = semTypeNode(c, n[2], nil)
     n[2] = newNodeIT(nkType, n[2].info, t2)
@@ -527,7 +532,7 @@ proc semIs(c: PContext, n: PNode, flags: TExprFlags): PNode =
       liftLhs = false
 
   var lhsType = n[1].typ
-  if n[1].isError:
+  if n[1].isError or n[2].isError:
     result = wrapError(c.config, n)
   elif lhsType.kind == tyTypeDesc and (lhsType.base.kind == tyNone or
      (c.inGenericContext > 0 and lhsType.base.containsGenericType)):
@@ -2643,8 +2648,10 @@ proc semMagic(c: PContext, n: PNode, s: PSym, flags: TExprFlags): PNode =
     result = semDirectOp(c, n, flags)
 
 proc semWhen(c: PContext, n: PNode, semCheck = true): PNode =
-  # If semCheck is set to false, ``when`` will return the verbatim AST of
+  # If semCheck is set to false, `when` will return the verbatim AST of
   # the correct branch. Otherwise the AST will be passed through semStmt.
+  addInNimDebugUtils(c.config, "semWhen", n, result)
+
   result = nil
 
   template setResult(e: untyped) =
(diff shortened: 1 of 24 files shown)


compiler/sem/semstmts.nim
```diff
@@ -462,11 +462,508 @@ proc errorSymChoiceUseQualifier(c: PContext; n: PNode) =
...
+
+proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
+  ## semantically analyses let or var sections, analysis follows these steps:
+  ## 0. results are accumulated into a statement list and holds the section
+  ## 1. for each entry in the section to analyse:
+  ##    - if a single ident in the identdef it's analysed
+  ##    - multiple idents are decomposed and analysed one at a time
+  ##    - nkVarTuple:
+  ##      1. rhs assigns to a single intermediary identdef and then analysed
+  ##      2. for each unpack item, one identdef, analysed one at a time
```
saem: I'm going to drop the lowering approach. Pretty sure I figured out a way around it.

saem:
Still ugly, the pragma data on the symbol AST is missing/needs to be fixed.

Then a large amount of cleanup.

compiler/sem/semexprs.nim
```diff
@@ -2932,26 +2940,88 @@ proc semTupleConstr(c: PContext, n: PNode, flags: TExprFlags): PNode =
...
+    if bodyRes.kind != nkError:
+      result.typ = bodyRes.typ
+
+    if isEmptyType(result.typ):
+      result.transitionSonsKind(nkBlockStmt)
+    else:
+      result.transitionSonsKind(nkBlockExpr)
+
+    if lablRes.kind == nkError or bodyRes.kind == nkError:
+      result = c.config.wrapError(result)
```
saem, line 2996: @Clyybber so fun discovery, turns out nkError is a problem when the `efNoSemCheck` flag is passed and causes all sorts of issues with the current compiler design. Maybe I'm wrong about the broad reaching impact, but here goes.

The command to make the current PR fail is `./koch.py temp -d:debug c -r tests/stdlib/tsequtils.nim` It'll spew a giant pile of debug on screen, it's the body of `mapIt` from `sequtils` because it's failing in the test on line 432.

What's happening is the `var it {.inject.} = ...` in `mapIt` is producing an nkError, great but then semBlock isn't doing anything with it. So I changed up `semBlock` to do that, and honestly the code makes way more sense from an analysis and production perspective.

The problem is that it's still failing and that's because `semStmtList` in `semstmts` is failing to to produce an nkError, because it's guarded by a check for `efNoSemCheck` flag. The issue now is that we're producing an AST riddled with nkErrors.

There are a bunch of issues but if I keep propogating nkError we're basically doing no semcheck whatsoever until we get up to the callsite that issued the `efNoSemCheck` in the first place. Then there is the issue we can't separate type vs untyped analysis. Can't recover the error because we're not consistent enough with `wrongNode`, our report kinds aren't sufficiently categorized for this.

Thoughts?

saem: @zerbina in case this piques your interests or you have some ideas.

saem: Hmm, rereading through history and where it's used `efNoSemCheck` might not be as serious as it sounds and it seems to only really matter wrt to governing whether to run after macro call analysis. It might just be as simple as propagating it through? 🤔

(I'm trying this, but the whole thing feels off)

zerbina: I haven't tried reproducing the issue yet, but here are some first thoughts:

If I haven't missed a usage site, `efNoSemCheck` is only passed to `semTemplateExpr` and `semMacroExpr` (as you said). The `efNoSemCheck notin flags` test in `semStmtList` currently always evaluates to true (and `isError` should thus be set to true in case an `nkError` node is encountered), since both `semTemplateExpr` and `semMacroExpr` prevent post-expansion semantic analysis if `efNoSemCheck` is present.

Aside: Maybe `efNoSemCheck` should be removed and both `semTemplateExpr` and `semMacroExpr` be split into two procedures, where one only performs the expansion and the other one performs the expansion plus `semAfterMacro`?

saem: Thanks, for sharing your perspective, it corroborates what I'm experiencing/thinking.

Right now I'm trying to sort out where the nkError get swallowed -- got it progagating it further. The nkError wedge seems to be pushing in the same direction that you're hinting at, which is to split the expansion vs check.

saem: Not pretty, but it seems to work.

saem:
Took this out of draft, I'll do a squash and proper commit message after some review.

saem:
Going to start a rebase and merge this in.

PR #381, opened 2022-07-14
Convert semgnrc to nkError

Clyybber:
Rename wrapErrorInSubTree to just wrapError.
Add a debug checker for empty wrapErrors.
Add a captureError template, only used in semgnrc for now
Minor clean up of addDecl/addDeclAt in lookups.nim

### DIFF
--- a/compiler/sem/semgnrc.nim
+++ b/compiler/sem/semgnrc.nim
@@ -22,6 +22,10 @@ proc getIdentNode(c: PContext; n: PNode): PNode =
   of nkPostfix: result = getIdentNode(c, n[1])
   of nkPragmaExpr: result = getIdentNode(c, n[0])
   of nkIdent, nkAccQuoted, nkSym: result = n
+  of nkError:
+    # Passing it through; considerQuotedIdent in
+    # newSymS will handle and output this error
+    result = n
   else:
     semReportIllformedAst(c.config, n, {
       nkPostfix, nkPragmaExpr, nkIdent, nkAccQuoted, nkSym})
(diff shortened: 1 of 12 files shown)


compiler/sem/semgnrc.nim
```diff
@@ -18,6 +18,7 @@
 ## included from sem.nim
 
 proc getIdentNode(c: PContext; n: PNode): PNode =
+  # XXX nkError? Also should callsites of this proc check?
```
saem: I think here is fine, we can just immediately return n and move on.

compiler/sem/semgnrc.nim
```diff
@@ -324,19 +329,24 @@ proc semGenericStmt(c: PContext, n: PNode,
     # Consider 'when declared(globalsSlot): ThreadVarSetValue(globalsSlot, ...)'
     # in threads.nim: the subtle preprocessing here binds 'globalsSlot' which
     # is not exported and yet the generic 'threadProcWrapper' works correctly.
-    let flags = if mixinContext: flags+{withinMixin} else: flags
-    for i in first..<result.safeLen:
-      result[i] = semGenericStmt(c, result[i], flags, ctx)
+    if not result.isError:
+      captureError c.config, result:
+        let flags = if mixinContext: flags+{withinMixin} else: flags
+        for i in first..<result.safeLen:
```
saem, line 351: Personally, I think the real pattern might be transform children (or a range), with some map proc/block, and on error wrap the whole thing.

It's a combination of flatmap and fold applying a semantic analysis function. I can't even think of a name.

result = foo(result, range, proc)

Could make the first param a var, but that made me feel like a monster.

saem:
I think the only thing I'd change is the nkError handling in the comment I identified as it's swallowing the error otherwise.

Otherwise the code is better than before and good to merge.

compiler/sem/semgnrc.nim
Clyybber: Yeah, just often it's not cleanly over a range, but some custom iteration, hence I chose the template approach.

PR #383, opened 2022-07-20
use atomic reference counting when threading

disruptek:
This is a reimplementation of the atomic reference counting patch.
When using threads, reference counting uses atomic operations.
This change also removes the `discardable` pragma from `atomicInc` and `atomicDec`, as well as adding an optional `AtomMemModel` parameter to these functions.

This code is tested extensively via https://github.com/sesco-llc/gread and a more complete version, which includes identical patching against `refs_v2` in mainline Nim, is available in https://github.com/nim-works/gitnim/tree/1.6.7.

### DIFF
--- a/lib/system/atomics.nim
+++ b/lib/system/atomics.nim
@@ -211,13 +211,24 @@ elif someVcc and hasThreadSupport:
         importc: "_InterlockedExchangeAdd", header: "<intrin.h>".}
 
 else:
+  type AtomMemModel* = distinct cint
+
+  const
+    ATOMIC_RELAXED = 0.AtomMemModel
+    ATOMIC_CONSUME = 1.AtomMemModel
+    ATOMIC_ACQUIRE = 2.AtomMemModel
+    ATOMIC_RELEASE = 3.AtomMemModel
+    ATOMIC_ACQ_REL = 4.AtomMemModel
+    ATOMIC_SEQ_CST = 5.AtomMemModel
+
   proc addAndFetch*(p: ptr int, val: int): int {.inline.} =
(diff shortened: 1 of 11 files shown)


lib/system/atomics.nim
```diff
@@ -211,26 +211,38 @@ elif someVcc and hasThreadSupport:
...
 else:
+  type AtomMemModel* = distinct cint
+
+  const
+    ATOMIC_RELAXED = 0.AtomMemModel
+    ATOMIC_CONSUME = 1.AtomMemModel
+    ATOMIC_ACQUIRE = 2.AtomMemModel
+    ATOMIC_RELEASE = 3.AtomMemModel
+    ATOMIC_ACQ_REL = 4.AtomMemModel
+    ATOMIC_SEQ_CST = 5.AtomMemModel
```
zerbina, line 222: These should be exported, I think? The `ATOMIC_` constants in the `vcc` branch should probably be too.

disruptek: Yeah.  I asked in IRC but I figured that this would be the simplest and least impactful change.  But I cannot think of a reason not to expose the type fully; especially since we do allow invocation of the inc/dec without regard to threading...

disruptek: Yeah.  I asked in IRC but I figured that this would be the simplest and least impactful change.  But I cannot think of a reason not to expose the type fully; especially since we do allow invocation of the inc/dec without regard to threading...

saem: this can be covered later, I say merge and revisit if/when needed.

PR #388, opened 2022-07-30
macro pragmas: support template and const defs

saem:
## Summary:
* constdef annotation handling same as identdefs in let or var
* annotation processing on template definitions like other routine kinds

## Details:
* reenabled `tests/pragmas/tpragmas_misc`
* updated manual for macros pragmas
* `semstmts.semConst` reworked to be like `semstmts.semLetOrVar`
* constdef annotation handling same as identdefs in let or var
* `semLetOrVarAnnotation` works on const defs, and renamed
* added tracing to `semTemplateDef`
* added tests for let/var section processing
    
Mostly a port of https://github.com/nim-lang/Nim/pull/19406

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -488,24 +488,23 @@ proc copyExcept(n: PNode, i: int): PNode {.inline.} =
   for j in 0..<n.len:
     if j != i: result.add(n[j])
 
-proc semLetOrVarAnnotation(c: PContext, n: PNode;
-                           validPragmas: TSpecialWords): PNode =
-  ## analyses normalized let or var sections for pragma annotations and applies
-  ## the first macro pragma producing the transform wrapped in an nkStmtList,
-  ## if none are applied `n`, otherwise an error. If an nkStmtList is returned
-  ## the call must unwrap it execute a semExpr prior to inserting it into the
-  ## AST.
+proc semConstLetOrVarAnnotation(c: PContext, n: PNode): PNode =
+  ## analyses normalized const, let, or var section for pragma annotations and
+  ## applies the first macro pragma producing the transform wrapped in an
+  ## nkStmtList, if none are applied `n`, otherwise an error. If an nkStmtList
+  ## is returned the caller must unwrap it and execute a semExpr prior to
+  ## inserting it into the AST.
 
-  addInNimDebugUtils(c.config, "semLetOrVarAnnotation", n, result)
+  addInNimDebugUtils(c.config, "semConstLetOrVarAnnotation", n, result)
 
   # we don't preemptively set `result` to `n` so we can check to see if we did
   # produce something new or not via a nil check on `result`.
 
-  internalAssert(c.config, n.kind in nkVariableSections,
-    "only let or var sections allowed, got: " & $n.kind)
+  internalAssert(c.config, n.kind in nkVariableSections + {nkConstSection},
+    "only const, let, or var sections allowed, got: " & $n.kind)
   checkSonsLen(n, 1, c.config)
 
-  assert n[0].kind == nkIdentDefs
+  assert n[0].kind in {nkConstDef, nkIdentDefs}
 
   let pragExpr = n[0][0]
   if pragExpr.kind != nkPragmaExpr:
@@ -521,9 +520,9 @@ proc semLetOrVarAnnotation(c: PContext, n: PNode;
                 prag
 
     # we only want to process macro pragmas (ast transforms)
-    # xxx: this dance to filter through them is another sign that pragmas are
+    # xxx: this dance to filter through them is another sign that pragmas
     #      shouldn't be associated to symbols as the syntax hints
-    if whichPragma(prag) == wInvalid:
+    if whichPragma(prag) == wInvalid and key.kind in nkIdentKinds:
       # a custom pragma as opposed to a built-in
       let (ident, err) = considerQuotedIdent(c, key)
 
@@ -668,6 +667,7 @@ proc semNormalizedLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
         # xxx: replace with an emptyType?
         nil
       of nkError:
+        # hasError is set further on
         defTypePart.typ
       else:
         semTypeNode(c, defTypePart, nil)
@@ -739,17 +739,7 @@ proc semNormalizedLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
   if c.matchedConcept != nil:
     typFlags.incl taConcept
 
-  let t = typeAllowed(typ, symkind, c, typFlags)
-  if t != nil:
-    typ = newTypeError(typ, nextTypeId(c.idgen)):
-            c.config.newError(def, 
-                              SemReport(
-                                kind: rsemTypeNotAllowed,
-                                allowedType: (
-                                  allowed: t,
-                                  actual: typ,
-                                  kind: symkind,
-                                  allowedFlags: typFlags)))
+  typ = typeAllowedOrError(typ, symkind, c, def, typFlags)
 
   # always construct a `producedDecl`, even on error, as we still need to
   # include all the preceding child nodes from `defPart`, without potentially
@@ -954,33 +944,8 @@ proc semNormalizedLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
     discard
 
   if hasError:
-    # xxx: hasError is here until we've converted the whole proc to use nkError
-    #      until then we check at the end to ensure we capture any nkErrors
-    #      embedded within the section. Even after full conversion more work
-    #      will be required to eliminate it.
+    # wrap the result if there is an embedded error
     result = c.config.wrapError(result)
-
-    # if c.config.inFile(n, "tsequtils.nim", 1006..1069):
-    #   var errors: seq[PNode] = @[]
-    #   for error in c.config.anyErrorsWalk(result):
-    #     debug error
-  else:
-    discard
-    # if c.config.inFile(n, "sequtils.nim", 1006..1069):
-    #   var errors: seq[PNode] = @[]
-    #   for error in c.config.anyErrorsWalk(result):
-    #     debug error
-
-      # debug n
-      # writeStackTrace()
-      # debug result
-
-      # if defPart.isErrorLike or
-      #   defInitPart.isErrorLike or initExpr.isErrorLike or initType.isErrorLike or
-      #   defTypePart.isErrorLike or defTypePart.isErrorLike or givenTyp.isErrorLike:
-      #   debug n
-      #   debug initType
-      #   debug givenTyp
 
 
 proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
@@ -988,12 +953,12 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
   ##
   ## 0. results are accumulated into a statement list and holds the section
   ## 1. for each entry in the section to analyse:
-  ##    - if a single ident in the identdef or vartuple it's analysed
+  ##    - if it's an identdefs with a single ident or a vartuple it's analysed
   ##    - multiple idents are decomposed and analysed one at a time
-  ## 2. analysis is first on the untyped version (macro pragma)
+  ## 2. analysis is first done on the untyped version (macro pragma)
   ## 3. typed analysis (interleaved with typed macro pragmas):
   ##    1. expansion: rhs and lhs are analysed
-  ##    2. reduction: combined into an identdef, then final analysis
+  ##    2. reduction: combined into an identdefs, then final analysis
   ##
   ## If successful a singleton let/var section with a singleton identdefs or
   ## varTuple is produced, otherwise nkError. IdentDefs's name part is
@@ -1051,7 +1016,9 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
         singletonNode.add singletonDef
 
         # process pragmas and based on that direct further analysis
-        let pragmad = semLetOrVarAnnotation(c, singletonNode, validPragmas)
+        let pragmad = semConstLetOrVarAnnotation(c, singletonNode)
+          ## a single def after pragma macros (annotations) are applied, which
+          ## can be anything after transformation
         case pragmad.kind
         of nkEmpty:
           discard # skip adding it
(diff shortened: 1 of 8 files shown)


compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+  if haveInit and initType.kind == tyTypeDesc and
+      c.p.owner.kind != skMacro:
+    typFlags.incl taProcContextIsNotMacro
 
-    # don't evaluate here since the type compatibility check below may add a converter
-    var def = semExprWithType(c, a[^1])
+  def = semConstExpr(c, def)
+  if def == nil:
+    def = c.config.newError(defInitPart, reportSem rsemConstExpressionExpected)
+    hasError = true
```
zerbina: The node returned by `semConstExpr` is never `nil`, so this check can be omitted.

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
 
-    if def.kind == nkSym and def.sym.kind in {skTemplate, skMacro}:
-      typFlags.incl taIsTemplateOrMacro
-    elif def.typ.kind == tyTypeDesc and c.p.owner.kind != skMacro:
-      typFlags.incl taProcContextIsNotMacro
+  if def.kind != nkNilLit:
+    # xxx: this seems wrong, instead of guarding on nil, maybe it should be
+    #      inferred from `typ`? then we still do the rest of the checks?
```
zerbina, line 1180: The `nkNilLit` check here is done in order to support nil `ref`, `ptr` and `pointer` values for `const`s (they would be rejected by `typeAllowed` otherwise).

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+    if t != nil:
+      typ = newTypeError(typ, nextTypeId(c.idgen)):
+              c.config.newError(def,
+                                SemReport(
+                                  kind: rsemTypeNotAllowed,
+                                  allowedType: (
+                                    allowed: t,
+                                    actual: typ,
+                                    kind: skConst,
+                                    allowedFlags: typFlags)))
```
zerbina: Do you think it'd make sense to add a `checkTypeAllowedOrError` that returns an error type in the "not allowed"-case and the input type otherwise and use that here instead? So basically:

```nim
typ = typeAllowedCheck(c, typ, skConst, typFlags)
```

It would reduce code duplication a bit once the other usage of `typeCheckAllowed` start to move away from `localReport`.

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+      hasError = true
+
+    if not hasError and sfCompileTime in v.flags:
+      var x = newNodeI(result.kind, v.info)
+      x.add producedDecl
+      vm.setupCompileTimeVar(c.module, c.idgen, c.graph, x)
+
+    if v.flags * {sfGlobal, sfThread} == {sfGlobal}:
+      # this is just logging, doesn't need to be converted to an error
+      localReport(c.config, v.info, reportSym(rsemGlobalVar, v))
```
zerbina: The same for `sfGlobal`.

Edit: this comments was intended to follow after the ones concerning the `.compileTime`, `.threadvar` pragmas.

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+        producedDecl[^1] = c.config.newError(n[^1],
+                                             reportSem rsemThreadvarCannotInit)
+    
+    if producedDecl[i].isError:
+      hasError = true
+
+    if not hasError and sfCompileTime in v.flags:
+      var x = newNodeI(result.kind, v.info)
+      x.add producedDecl
+      vm.setupCompileTimeVar(c.module, c.idgen, c.graph, x)
```
zerbina: I think this is not correct.

```nim
const a {.compileTime.} = x
```

doesn't make sense (and should be disallowed).

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+        if actualType.kind in {tyObject, tyDistinct} and
+            actualType.requiresInit:
+          defaultConstructionError2(c, v.typ, r)
+        else:
+          checkNilableOrError(c, producedDecl[i])
+    else:
+      if sfThread in v.flags:
+        hasError = true
+        producedDecl[^1] = c.config.newError(n[^1],
+                                             reportSem rsemThreadvarCannotInit)
```
zerbina: I don't think `sfThread` on a `const` should be allowed and instead be an error.

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+    discard
 
   if hasError:
-    # xxx: hasError is here until we've converted the whole proc to use nkError
-    #      until then we check at the end to ensure we capture any nkErrors
-    #      embedded within the section
+    # wrap the result if there is an embedded error
     result = c.config.wrapError(result)
-
+
```
```suggestion

```

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+    
   dec c.inStaticContext
 
 
+proc semConst(c: PContext, n: PNode): PNode =
+  ## semantically analyses const sections, analysis follows these steps:
+  ##
+  ## 0. results are accumulated into a statement list and holds the section
+  ## 1. for each entry in the section to analyse:
+  ##    - if a single ident in the identdef or vartuple it's analysed
```
```suggestion
  ##    - if there's only a single ident in the identdef or vartuple, it's analysed
```

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+  ##    - multiple idents are decomposed and analysed one at a time
+  ## 2. analysis is first on the untyped version (macro pragma)
```
```suggestion
  ## 2. analysis is first done on the untyped version (macro pragma)
```

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+  ## 3. typed analysis (interleaved with typed macro pragmas):
+  ##    1. expansion: rhs and lhs are analysed
+  ##    2. reduction: combined into an identdef, then final analysis
+  ##
+  ## If successful a singleton const section with a singleton constDef or
+  ## varTuple is produced, otherwise nkError. ConstDef's name part is
+  ## normalized to always contain have a pragmaExprs, even with an empty
```
```suggestion
  ## normalized to always contain a pragmaExprs, even with an empty
```

compiler/sem/semstmts.nim
```diff
@@ -1099,155 +1074,425 @@ proc semLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
...
+          singletonDef.comment = a.comment
+
+        singletonDef[0] = a[j]
+        singletonDef[1] = copyTree(a[^2])
+        singletonDef[2] = copyTree(a[^1])
+
+        singletonNode.add singletonDef
+
+        # process pragmas and based on that direct further analysis
+        let pragmad = semConstLetOrVarAnnotation(c, singletonNode)
```
zerbina, line 1416: The name `pragmad` is a bit confusing. I had to read it multiple times in order to realize that it's likely supposed to mean "def with pragmas applied".

My comment isn't meant as a request for it to change - just as feedback.

doc/manual.rst
```diff
@@ -7811,11 +7812,31 @@ This is translated to:
...
-which will replace the original row in the type section.
+which will replace the original row in the type section. Transformation for
+type sections are ill-defined and error prone.
 
-When multiple macro pragmas are applied to the same definition, the
-compiler will apply them consequently from left to right. Each macro
-will receive as input the output of the previous one.
+For variables and constants, it is largely the same, except a unary node with
+the same kind as the section containing a single definition is passed to
+macros, and macros can return any expression.
```
zerbina:
```suggestion
For variables and constants, it is largely the same, except that an unary
node with the same kind as the section containing a single definition is
passed to macros, and macros can return any statement.
```

doc/manual.rst
```diff
@@ -7811,11 +7812,31 @@ This is translated to:
...
+.. code-block:: nim
+  var a = ...
+  foo:
+    var b {.importc, nodecl.} = ...
+  var c = ...
+
+When multiple macro pragmas are applied to the same definition, the first one
+from left to right will be evaluated. This macro can then choose to keep the
+remaining macro pragmas in its output, and those will be evaluated in the same
+way.
```
zerbina:
```suggestion
When multiple macro pragmas are applied to the same definition, the compiler will
evaluate them from left to right. Each macro can choose to remove the remaining
macro pragmas - they won't be evaluated then.
```
It's subjective, but I think this is easier to parse.

compiler/sem/semstmts.nim
```diff
@@ -488,24 +488,23 @@ proc copyExcept(n: PNode, i: int): PNode {.inline.} =
...
-  ## the first macro pragma producing the transform wrapped in an nkStmtList,
-  ## if none are applied `n`, otherwise an error. If an nkStmtList is returned
-  ## the call must unwrap it execute a semExpr prior to inserting it into the
-  ## AST.
+proc semConstLetOrVarAnnotation(c: PContext, n: PNode): PNode =
+  ## analyses normalized const, let, or var section for pragma annotations and
+  ## applies the first macro pragma producing the transform wrapped in an
+  ## nkStmtList, if none are applied `n`, otherwise an error. If an nkStmtList
+  ## is returned the call must unwrap it execute a semExpr prior to inserting
+  ## it into the AST.
```
```suggestion
  ## is returned the call must unwrap it and execute a semExpr prior to
  ## inserting it into the AST.
```

zerbina (commented):
I left a few comments regarding allowed pragmas on `consts` and also a few suggestions for documentation improvements, but overall, it looks really nice.

saem:
bors cancel

compiler/sem/semstmts.nim
saem: Yeah, but that sounds like a bug with `typeAllowed` that we're working around here?

compiler/sem/semstmts.nim
saem: smort!

Yeah, that's better than what's happening now, gimmie a bit.

compiler/sem/semstmts.nim
saem: fixed locally

compiler/sem/semstmts.nim
saem: Thanks, I added a comment to help, maybe a better name will become evident soon.

compiler/sem/semstmts.nim
saem: Updated locally to:

```
##    - if it's a constdef with a single ident or a vartuple it's analysed
```

compiler/sem/semstmts.nim
saem: Yup, that was me on autopilot, it would already error out in pragmas.

compiler/sem/semstmts.nim
saem: Yup, autopilot, would error out in pragmas

compiler/sem/semstmts.nim
saem: fixed locally, moved the `hasError = true` futher up.

compiler/sem/semstmts.nim
saem: Removed the global var logging, doesn't make sense unlike let and var.

compiler/sem/semstmts.nim
saem: I left an extra space between procs to more clearly demarcate the transition when quickly skimming, not married to it but curious what you think?

compiler/sem/semstmts.nim
saem: fixed.

compiler/sem/semstmts.nim
saem: fixed.

doc/manual.rst
saem: applied locally

doc/manual.rst
saem: I like it, applied locally.

compiler/sem/semstmts.nim
zerbina: Well, `typeAllowed` only looks at the type, so I wouldn't consider it a bug.

Thinking about this a bit more, we don't want to disallow `ref` and `ptr` in general, but only non-`nil` values of these types. With the current implementation,
```nim
type A = object
  x: ref int

const c = A()
```
is rejected, even though it'd work fine.

Rejecting non-`nil` values for `ptr`, `ref`, and closure environments is actually already implemented (the `rsemVmUnsupportedNonNil` report; see `deserializeRef` and `deserialize` in `vmcompilerserdes.nim`), so the `nkNilLit` guard here could be removed (after adjusting `typeAllowed`).

`typeAllowed` is still needed here, in order to reject objects using inheritance (the back-ends don't support those yet if I remember correctly).

I do think it's better to do this in a separate PR however.

compiler/sem/semstmts.nim
saem: Done.

compiler/sem/semstmts.nim
zerbina: I meant the trailing whitespace (I should have mentioned that with the suggestion), the extra line is fine and I agree that it helps with readability.

compiler/sem/semstmts.nim
saem: Totally agree on the separate PR.

As for whether it's a bug or not. `typeAllowed` does know it's a const/let/var. It's also only receiving `typ`, and not `def`. So the only pathway for `def` to influence `typeAllowed` is the earlier derivation of `typ` from `def`. In which case, shouldn't `typ` be correctly constructed such that `typeAllowed` doesn't reject it?

Correctly constructed meaning, given:
```
const foo: SomeRefType = nil
# or
const foo: SomeRefType = evalsToNil()
```
even if `def` is a nil it should make no difference as `typ` should be correctly derived. Unless, `typ` is not being derived correctly from `def` and/or we're missing a user specified type expession and we have a pure nil value (kinda like a void pointer).

Maybe I'm totally misunderstanding it?

saem:
Squashed with all the latest.

doc/manual.rst
metagn: Not important but `an unary` is wrong

saem:
Not sure what exactly broke reproducible builds, have to figure out how to figure that out. 😓  But in the meantime rebased on the latest devel as it contains some workflow updates and perhaps that's might address this. 🤞🏽

PR #391, opened 2022-07-31
ci: typo key for fail on unmatched files

saem:
## Summary
* fixed a typo in ci publishing, should now fail if a glob matches nothing

---


alaviss (approved):
Thanks!

bors r+

PR #392, opened 2022-07-31
internal: remove sons from leaf nodes empty/none

saem:
## Summary
- pulls nkEmpty and nkNone into a branch without sons
- ensures more correctness by design

## Details
- various small fixes to get bootstrap working
- no tests failed, so that's a good{?} sign

---

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -430,8 +430,10 @@ proc addSonNilAllowed*(father, son: PNode) =
   father.sons.add(son)
 
 proc delSon*(father: PNode, idx: int) =
-  if father.len == 0: return
-  for i in idx..<father.len - 1: father[i] = father[i + 1]
+  if father.len == 0:
+    return
+  for i in idx..<father.len - 1:
+    father[i] = father[i + 1]
   father.sons.setLen(father.len - 1)
 
 template copyNodeImpl(dst, src, processSonsStmt) =
@@ -451,6 +453,7 @@ template copyNodeImpl(dst, src, processSonsStmt) =
   of nkSym: dst.sym = src.sym
   of nkIdent: dst.ident = src.ident
   of nkStrLit..nkTripleStrLit: dst.strVal = src.strVal
+  of nkEmpty, nkNone: discard # no children, nothing to do
   else: processSonsStmt
 
 proc copyNode*(src: PNode): PNode =
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -809,6 +809,8 @@ type
       sym*: PSym
     of nkIdent:
       ident*: PIdent
+    of nkEmpty, nkNone:
+      discard
     else:
       sons*: TNodeSeq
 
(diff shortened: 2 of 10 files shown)


compiler/ast/ast.nim
```diff
@@ -506,23 +508,35 @@ proc shallowCopy*(src: PNode): PNode =
 proc copyTree*(src: PNode): PNode =
   # copy a whole syntax tree; performs deep copying
   copyNodeImpl(result, src):
-    newSeq(result.sons, src.len)
-    for i in 0..<src.len:
-      result[i] = copyTree(src[i])
+    case src.kind
+    of nkWithoutSons:
+      discard
+    else:
```
zerbina: Instead of having this logic here, I think it'd be better to add a `of nkNone, nkEmpty: discard` branch to the `case` in `copyNodeImpl`. This would reduce code duplication a bit.

compiler/ast/ast.nim
```diff
@@ -506,23 +508,35 @@ proc shallowCopy*(src: PNode): PNode =
...
 proc copyTreeWithoutNode*(src, skippedNode: PNode): PNode =
   copyNodeImpl(result, src):
-    result.sons = newSeqOfCap[PNode](src.len)
-    for n in src.sons:
-      if n != skippedNode:
-        result.sons.add copyTreeWithoutNode(n, skippedNode)
+    case src.kind
+    of nkWithoutSons:
+      discard
+    else:
```
zerbina: The same here.

zerbina (commented):
Left a small suggestion. 

As an aside, I think it'd make future tracking down of changes (e.g. via `git blame`) easier if the refactorings unrelated to the main topic of the PR (single-line statements to multi-line, `for it in n` instead of `for i in 0..<n.len`, etc) are done in a separate PR. I don't consider this to be a blocker however.

compiler/ast/ast.nim
saem: Wow, talk about tunnel vision, I totally understood what was happening there. Yup, that's much much cleaner and what it should have been all along.

compiler/ast/ast.nim
saem: Should be fixed.

compiler/ast/ast.nim
saem: Done.

saem:
Bots r+

compiler/ast/ast.nim
```diff
@@ -430,8 +430,10 @@ proc addSonNilAllowed*(father, son: PNode) =
   father.sons.add(son)
 
 proc delSon*(father: PNode, idx: int) =
-  if father.len == 0: return
-  for i in idx..<father.len - 1: father[i] = father[i + 1]
+  if father.len == 0:
+    return
+  for i in idx..<father.len - 1:
+    father[i] = father[i + 1]
```
Clyybber, line 436: Preexisting, but that if check is unneccessary, as the loop body will not execute if `father.len` is zero.

saem:
Bots r+

PR #393, opened 2022-08-02
lang/stdlib: remove quick-win deprecated items

saem:
## Summary
* removes many deprecated parts of the language/standard library
* should result in fewer dialects and faster CI to say the least

## Details
* to be filled in

---

## Notes for Reviewers

still need to think of a merging strategy, inclined to cherry pick and merge one at a time
the above is slow from a CI/manual effort perspective, it's cleaner, however. suggestions?
(especially @alaviss)

Need to wrap this up, I left off at the 1.6 release notes.


saem:
@Clyybber got any specific input on these two in particular?

- [pragmas: type pragmas must follow generic params](https://github.com/nim-works/nimskull/pull/393/commits/701c7ef8618a45ef9874b36e219e5e2fc8335d9b)
- [lang: remove deprecated c/C octal prefix](https://github.com/nim-works/nimskull/pull/393/commits/c83b8a1e41a0dda84a6c6f3c1a06185890f77df8)

compiler/ast/parser.nim
```diff
@@ -2063,38 +2063,41 @@ proc parseTypeClass(p: var Parser): PNode =
...
+        
+        if p.validInd and p.tok.tokType == tkBracketLe:
+          p.localError ParserReport(kind: rparPragmaBeforeGenericParameters)
+        
+        res
+      of tkEquals:
+        (p.emptyNode, p.emptyNode)
+        # parsing for everything after the equals is shared, we do that below
+      else:
+        (p.emptyNode, p.emptyNode) # xxx: error?
```
Clyybber, line 2091: The `else` case is handled by sem, `type abc` on it's own is valid in the parser, so I'd merge the `tkEquals` and the `else` branch. If we don't want/need a specialized error message for "pragma before generic param" then we can also remove the tkCurlyDotLe branch and have it handled with the generic "unexpected X" automatically I think.

Clyybber (approved):

compiler/ast/parser.nim
saem: I prefer to start with a focused grammar, then loosen restrictions over time (reasoning for all this is in the 'out loud thinking' below). So having the grammar restricted in the parser seems reasonable and I'd like to introduce more of that.

I likely don't understand the art of grammar construction well enough, so I'm wondering if it's reasonable to say that `type abc` shouldn't be valid here hence the `else` being separate?

As for the special case error message for pragma and position ordering, that's friendlier to humans.

----
(Out loud thinking)

I'm still getting a feel for what should be where, in particular a set of principles in place.

If we know there is an irregularity in the grammar we should reject it early, IMO. We get far more helpful error messages that way in that we've got better context.

There are a bunch of competing ideas in my head, but they boil down to two main "sides":

1. the focused grammar of Nimskull the language itself and building to that, with crisp and clear error messages, quick detection of issues and syntax limiting is more useful, etc
2. a superset grammar, that can represent all legal terms the former does _and_, is more forgiving for the purposes of building DSLs and other things macros should let you do. All within reason, not full reader macro control like lisp/racket.

The focused route (1) has lots of upsides, not the least of which is simplicity in how direct it is -- consider how much people struggled with `|`, which is alternation in regular grammars.

1 and 2 are somewhat mutually exclusive, we can start with one and move to 2, but the other way is much harder. We can also have 1 be the norm and in special sections drop to 2. In the end we need to pick a horse.

I started to sketch out grammar structures for proc and type definitions in my notepad to see what the symmetries are, mostly ignoring current syntax. I'm starting to get some insights, but it's very nascent and even if it was ready as an idea, it's an entirely different matter implementing it.

alaviss:
> still need to think of a merging strategy, inclined to cherry pick and merge one at a time
> the above is slow from a CI/manual effort perspective, it's cleaner, however. suggestions?
> (especially @alaviss)

That's certainly the best when it comes to: "Every commit in devel is green".

If it's too much hassle a squash is certainly possible, though this change is a bit big for a squash, so you'd have to squash by categories.

saem:
All changes merged in, closing PR.

PR #395, opened 2022-08-06
stdlib: remove deprecated modules

saem:
deprecated stdlib module removal:
- removed events, LockFreeHash, parseopt2, and securehash
- associated clean-up in testament
- lib/deprecate/pure now only contains ospaths, an include for os
- remove events related tests


saem:
Bots r+

PR #396, opened 2022-08-06
lang: remove deprecated c/C octal prefix

saem:
## Summary
- can no longer use 0c1
- this has long since been deprecated and not referenced in the manual


saem:
Bots r+

PR #397, opened 2022-08-06
macros: remove deprecated TNimSym/TypeKinds

saem:
## Summary
- Removed macros.TNimSym and macros.TypeKinds

## Details
They are not used anywhere in the code base


saem:
Bots r+

PR #398, opened 2022-08-06
cpp: remove this pragma, breaks codegen

saem:
Removed the this pragma, but codegen is now broken:
- codegen no longer renames parameters that happen to be called `this`
- can't find a reasonable causal chain that explains why
- disabled tests that fail on the cpp backend in the meantime

Added benefit:
- removed the `nfExprCall` flag
- remove the `sfIsSelf` const/flag
- CI should run faster and less cruft in dispatch

It also became clear that `semIndirectOp` and call patterns around it
need a complete rewrite. The results should be simplified dispatch.


saem:
Bots r+

saem:
Bots r+

saem:
Bots r+

PR #399, opened 2022-08-06
dialect: remove old case object define

saem:
## Summary
- removed old case object define

NB reset magic is still present


saem:
Bots r+

PR #400, opened 2022-08-06
pragmas: type pragmas must follow generic params

saem:
## Summary

removed the previous deprecation where pragmas could preceed the generic
parameter list in a type definition, eg:

```
type Foo{.somePragma}[T] = ... # now an error
type Foo[T] {.somePragma.} = ... # acceptable
```

## Details

Due to lots of issues, mostly stylistic, I've pragmas on the RHS of a
type definition as legal. Largely because after changing it I realized a
more significant rethink of type declaration syntax should cover this in
the future.

Fixed a number of test failures along the way due to this change.

### DIFF
--- a/compiler/ast/parser.nim
+++ b/compiler/ast/parser.nim
@@ -1987,7 +1987,7 @@ proc parseObject(p: var Parser): PNode =
   result = newNodeP(nkObjectTy, p)
   p.getTok
   result.add if p.tok.tokType == tkCurlyDotLe and p.validInd:
-               # Deprecated since v0.20.0
+               # Deprecated: type declaration syntax needs overhaul to fix this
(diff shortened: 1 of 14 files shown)


saem:
@alaviss do you know what's up with this, all tests/checks seem good and bors continually is gets stuck. Reopening the PR, syncing, etc, aren't helping. :(

saem:
Sure, now it works.

alaviss:
> @alaviss do you know what's up with this, all tests/checks seem good and bors continually is gets stuck. Reopening the PR, syncing, etc, aren't helping. :(

It could be that the bors server is overloaded. That does happen from time to time.

Once Github Merge Queue become available I will look into it as a bors substitute.

PR #402, opened 2022-08-06
stdlib: removed deprecated module oswalkdir

saem:
functionality has long since been moved to `std/os`


saem:
Bots r+

saem:
I think I punched bors in the face and it doesn't know what to do.

How's it going bors? You ok they're buddy?

PR #404, opened 2022-08-06
stdlib: remove deprecated tables.add and sharedtables

saem:
`add` was deprecated because it allowed mapping multiple values to a key
This was understandably very confusing, now add is removed.

additionally, sharedtables and sharedlists were deprecated, only
sharedtables was removed as sharedlists is still used when threads are
enabled.

Fixed broken tests.


saem:
Bots r+

disruptek:
Ugh, I was in the camp that people should RTFM and we shouldn't cripple a data structure which provides a cost-free feature simply because after 11 years @dom96 still doesn't know about it.

However, to be fair, I mostly use @c-blake's tables now, since they are superior.

Still, I think this is a poor line of reasoning for removing a feature which we may want to exploit in the compiler.

saem:
I see that as a different data structure. I'm not against either, they serve different purposes, but the one that comes up most often I find is one key one value.

I have personally encountered the need for the other structure such as when parsing HTTP and email headers where the key space doesn't associate to singular values.

Having two module, one for each purpose is a better bet than conflating the two interfaces. The general terms I'm used to from prior experience:

- map: a key space member associates with at most one value
- multimap: a key space member can be associated with multiple values

More "formally", one can think of them as such:
- map: `K -> Option[V]`
- multimap: `K -> List[V]`

If we can come up with some good names to differentiate them that'd be ideal. The most obvious one is `multitable` given `table` as it is today. I'm open to better names.

-----------

Relatedly, structures that might help with naming discussion:

- set/hashset: stores unique elements and membership is binary, either is present or not regardless of number of insertions or deletions 
- bag/multiset: stores unique elements and membership is count based on insertions and deletions

canelhasmateus:
In my code bases, i'm quite partial to using the names:

* `map`: K -> Option[V]
* `partition:`  K -> List[V]

saem:
Bors r+

We can do the rework afterwards.

disruptek:
Sure, and I was in favor of a simple rename in mainline.
Also, I think bitabs should be in stdlib.

PR #406, opened 2022-08-06
pragma: remove the deprecated unroll pragma

saem:
supposedly it was ignored by the compiler, the fewer pragmas the better.


saem:
Bots r+

PR #407, opened 2022-08-06
strutils: remove deprecated delete proc

saem:
## Summary

- remove deprecated `delete(var string, int, int)`
- removed associated tests
- updated usage of removed proc in rstgen


saem:
Bots r+

PR #408, opened 2022-08-06
httpclient: request only accepts HttpMethod

saem:
## Summary

`request` proc's `httpMethod` param no longer accepts strings


saem:
Bots r+

saem:
Bors r+

PR #410, opened 2022-08-06
algorithm: remove deprecated reversed proc

saem:
## Summary

- remove deprecated `reversed[T](var openArray[T], int, int): seq[T]`
- removed associated tests


saem:
Bots r+

PR #412, opened 2022-08-06
options: remove deprecated UnpackError

saem:
## Summary

Removed UnpackError, was replaced with UnpackDefect a while back

## Details

Defects in general are a bad idea, but less code is good too


saem:
Bots r+

PR #413, opened 2022-08-06
lang: remove for loop and case statement macros

saem:
There are better ways to do these sorts of things.


saem:
Bots r+

SolitudeSF:
> There are better ways to do these sorts of things.

like what? is there a relevant discussion/motivation i can read?

saem:
> > There are better ways to do these sorts of things.
> 
> like what? is there a relevant discussion/motivation i can read?

No, because there isn't a one to one replacement being considered. Considering the lack of any sort of _working_ interfaces starting with something like enumerable or iterable would be big leap forward. As for case, start making them less limited from the start. Performance issues get mitigated in many situations here and elsewhere with some partial evaluation.

SolitudeSF:
i barely understand what are you trying to say here. im gonna have to ask more directly: why are for loop macros removed?

saem:
> i barely understand what are you trying to say here. im gonna have to ask more directly: why are for loop macros removed?

Sorry, fat fingered the edit button at first.

I don't want this or many other features in the language. The approach is:

1. More obvious things should be done
2. See if those are insufficient
3. Discuss if this or some other approach is required

SolitudeSF:
okay, now i get it.
just unfortunate to see the feature i like go.

saem:
> okay, now i get it. just unfortunate to see the feature i like go.

It's useful to you, so it'd be helpful to capture some of the reasoning why here if you're up for sharing, thanks. https://github.com/nim-works/nimskull/discussions/289#discussioncomment-3344279

PR #416, opened 2022-08-07
cleanup: remove warnings throughout the compiler

saem:
## Summary

- mostly cleaning up unused variables and procs
- consolidated max macro pragma and generic instantiation const

### DIFF
--- a/compiler/ast/lexer.nim
+++ b/compiler/ast/lexer.nim
@@ -96,11 +96,13 @@ type
 
   TokTypes* = set[TokType]
 
+when defined(nimsuggest):
+  const weakTokens = {tkComma, tkSemiColon, tkColon,
+                      tkParRi, tkParDotRi, tkBracketRi, tkBracketDotRi,
+                      tkCurlyRi}
+    ## tokens that should not be considered for previousToken
+
 const
-  weakTokens = {tkComma, tkSemiColon, tkColon,
-                tkParRi, tkParDotRi, tkBracketRi, tkBracketDotRi,
-                tkCurlyRi} # \
-    # tokens that should not be considered for previousToken
   tokKeywordLow* = succ(tkSymbol)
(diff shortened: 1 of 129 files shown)


saem:
~Not worth it.~

saem:
welp.. this is a big change.

Don't need a detailed review but I blitzed through imports and took out a bunch of warnings.

@zerbina or @alaviss?

lib/nimrtl.nim
```diff
@@ -12,8 +12,7 @@ batchable: false
 #
 
 ## Main file to generate a DLL from the standard library.
-## The default Nimrtl does not only contain the `system` module, but these
-## too:
+## The default Nimrtl does contains the `system` module as well as:
```
zerbina:
```suggestion
## The default Nimrtl does contain the `system` module as well as:
```

compiler/ast/lexer.nim
```diff
@@ -748,12 +750,12 @@ proc getEscapedChar(L: var Lexer, tok: var Token) =
...
   template registerLine =
-    let col = L.getColNumber(pos)
-
     when not defined(nimpretty):
-      if col > MaxLineLength:
+      if L.getColNumber(pos) > MaxLineLength:
         L.localReportPos(
           LexerReport(kind: rlexLineTooLong), pos)
+    else:
+      discard
```
zerbina: I think omitting the empty else clause in this case improves readability. Just a suggestion - no change expected from my side.

zerbina (commented):
Found a small typo, but otherwise, looks good!

lib/nimrtl.nim
saem: Thanks.

compiler/ast/lexer.nim
saem: Fair enough, removed.

saem:
Bors r+

PR #421, opened 2022-08-20
internal: add missing reports and clean-up

saem:
## Summary

added missing reports for the parser:
- template missing end close
- invalid expression
- invalid filter

Misc:
- created an enum for compiler verbosity
- removed `rparName`, doesn't seem used at all
- formatting fixes

### DIFF
--- a/compiler/ast/lineinfos.nim
+++ b/compiler/ast/lineinfos.nim
@@ -24,6 +24,15 @@ export FileIndex, TLineInfo
 
 import reports
 
+type
+  CompilerVerbosity* = enum
+    ## verbosity of the compiler, number is used as an array index and the
+    ## string matches what's passed on the CLI.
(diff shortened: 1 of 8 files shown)


saem:
@haxscramper can you take a quick look at the `rparName` removal?

I don't think there is an issue, but just in case I didn't consider something.

haxscramper:
No, no problems with `rparName` removal - IIRC it was created from parser "linter" diagnostics, but the whole thing wasn't used in the first place, apparently.

saem:
hah, left some asserts in there.

saem:
> No, no problems with `rparName` removal - IIRC it was created from parser "linter" diagnostics, but the whole thing wasn't used in the first place, apparently.

Thanks for looking it over -- yeah that was my conclusion too.

PR #424, opened 2022-08-31
WIP: a re-implementation of the compiler backend

zerbina:
## Summary

This PR implements:
* back-end oriented intermediate representations for procedure bodies, types, symbols, etc. - `vmir`, `irtypes`, `irliterals`
* logic for translating post-`transf` AST to the IRs - `irgen`, `irtypes`, `cbackend2`
* transformation and processing logic for the various IRs - `irpasses`, `cpasses`, `typeinfogen`, `markergen`, `typeprocessing`
* a new C back-end based on the components listed above - `cbackend2`, `cgen2`

The overall goal is to design and implement a simple, modular, and data-oriented framework for developing the various back-ends/target with.

---

The `README.md` that exists is outdated - a lot of things changed since it was written. It still acts as a good introduction to the code however.

## Notes for Reviewers

The PR is quite large. I'd suggest starting with looking at `cbackend2.generateCode`, and after getting familiar with how it works on a high-level, continue with the components/sub-systems it uses.

There are a lot of maximum line-length violations (especially in code using the `IrCursor` API). For the most part, these are in parts of the code where I'm not yet happy with the general architecture, and that are thus likely to be rewritten anyway.

I also left a lot of annotations regarding ideas, to-dos, questions, and general problems in the code - feel free to comment on them.

The documentation very likely needs some extra attention. Not all routines are documented yet, and for those that are, I believe that the documentation is sometimes either glossing over too many details or focusing on the wrong things.


zerbina:
The prototype is finished - the back-end is able to build a compiler that is able to build a working compiler. There are still some limitations however:
- it only works on Windows - on Linux, the compiler built by the back-end immediately quits
- the generated C code depends on undefined behaviour and crashes when enabling C compiler optimizations (so `-d:release` or `--opt:speed` don't work)
- the back-end is not able to build itself yet
- only `--gc:refc` is supported by the back-end so far 
- the compiler built by the back-end takes around 40 to 50 minutes to build the compiler itself

As for why the generated code is so inefficient, I believe this is because of the various unfinished parts. The issues that I think contribute the most:
1. if `move` is not explicitly used, every assignment (including `x: 0` in constructor expressions) uses a full copy, no matter what is on the right side. E.g. `let x = f()` will do a full copy of the result of `f()`. 
2. all value-type arguments are unconditionally passed by value (even those that are borrowed from, resulting in the undefined behaviour mentioned earlier)
3. return-value-optimization (RVO) is not employed
4. `copyStringRC1`, `newSeqRC1`, and `newObjRC1` are not used (they were at one point), putting unnecessary pressure on the ZCT
5. a lot of the optimizations done by `cgen` are still missing

## Stats

`tempnim` build with: `--gc:refc -d:danger --exceptions:goto`

```
./bin/tempnim.exe c --nimcache:./nimcache/temp --checks:off --stacktrace:off -d:cbackend2 --compileOnly ../nimskull-devel/compiler/nim.nim
Hint: gc: refc; opt: none; 0 lines; 12.230s; 681.344MiB peakmem
Real peakmem: ~706 MB

./bin/tempnim.exe c --nimcache:./nimcache/temp --checks:off --stacktrace:off --compileOnly ../nimskull-devel/compiler/nim.nim
Hint: gc: refc; opt: none; 0 lines; 14.322s; 834.969MiB peakmem
Real peakmem: ~833 MB
```

`tempnim` build with: `--gc:orc -d:danger`

```
./bin/tempnim.exe c --nimcache:./nimcache/temp --checks:off --stacktrace:off -d:cbackend2 --compileOnly ../nimskull-devel/compiler/nim.nim
Hint: gc: refc; opt: none; 0 lines; 9.730s; 829.543MiB peakmem
Real peakmem: ~717 MB

./bin/tempnim.exe c --nimcache:./nimcache/temp --checks:off --stacktrace:off --compileOnly ../nimskull-devel/compiler/nim.nim
Hint: gc: refc; opt: none; 0 lines; 9.398s; 1016.242MiB peakmem
Real peakmem: ~860 MB
```

The peakmem shown by compiler is the peak amount of memory requested from the OS and doesn't represent the "real" peak amount of memory used by allocations.

For a fair comparison, checks and stack-traces are disabled because as of now, the back-end only supports checks partially and stack-traces not at all.

## Module overview (as of 29.09.2022)

### `cbackend2`

This is the entry-point of the back-end. Processing happens roughly as follows:
1. collect the top-level statements for all modules via the `passes` interface
2. transform all alive (and potentially alive) procedures and translate them to the IR used throughout the back-end via `irgen`
3. translate all environmental state of the previously translated code (symbols, types, literal data, etc.) to the representation used by the back-end
4. run multiple transformation and lowering passes (`irpasses`) against the code IR. Types are also lowered
5. cull procedures that turned out to not actually be part of the alive graph (they aren't really "culled" - just not queued for C code-generation)
6. setup the global context for the code-generator (`cgen2`)
7. perform C code-generation for all modules that have alive content

### `vmir`

The type definitions used by the code IR are located here, together with routines for querying and modifying the code. The code IR is not modified in-place, but instead, changes are collected in a separate *changeset* which has to be applied first. `IrCursor` and `Changes` are the APIs used for that.

The deferred approach allows for running multiple passes concurrently and then applying their merged together changesets at once. Mutations to the code IR are also easier to track/manage.

### `irgen`

Implements the translation of semantically-analysed and transformed NimSkull AST to the IR used by the back-end.

### `irpasses` / `cpasses`

These modules implement the code, type, and literal-data transformation/lowering passes. `cpasses` was originally meant for passes only relevant for the C-family of targets, but where a pass is located is mostly arbitrary now.

There's a lot of boilerplate code in there.

### `irtypes`

Contains the type definitions for types, symbols, and procedure-headers together with routines for translating, querying, and modifying them.

### `irliterals`

Implements the IR used for literal-data. Complex data (objects, tuples, arrays, slice-lists) is stored in a flat packed-tree structure. Literal-data is *immutable* - once added to the environment, it can't be modified again (at least via the provided API).

### `typeinfogen` / `markergen`

Implements the compiler-to-run-time interface for types (RTTI).

### `cgen2`

The C code-generator. It operates on modules and performs work as follows:
1. translate the body of every procedure attached to the input module into another IR (`CAst`)
2. figure out dependencies (referenced types, globals, constants, and procedure)
3. emit includes, prototypes for functions, type declarations/definitions, used constants, etc.
4. emit the implementation for every function attached to the module

More execution time is spent in `cgen2.emitModuleToFile` than in the back-end's other parts *combined*. Making the code-generator more efficient is simple - I just didn't get to working on it yet.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+
+proc initCompilerProcs(p: PassEnv, g: ModuleGraph, tgen: var DeferredTypeGen,
+                       procs: var ProcedureEnv, syms: var DeferredSymbols,
+                       data: var LiteralData) =
+  ## Registers all ``.compilerproc`` symbols with `p`
+  # TODO: rename
+  for sym in g.compilerprocs.items:
+    case sym.kind
+    of routineKinds:
+      p.compilerprocs[sym.name.s] = procs.requestProc(sym)
```
haxscramper, line 277: Maybe we should introduce the `func symStr(s: PSym): string = s.name.s` and use it, instead of continuing to write hundreds of naked field accesses in the code.

haxscramper (approved):
Aside from a minor stylistic comment I made, everything else seems to be in order. Sadly, I can hardly promise to provide an in-depth review ATM (IRL stuff), but going from the provided readme I don't think I would have any high-level comments anyway.

compiler/front/main.nim
```diff
@@ -146,22 +148,27 @@ proc commandCompileToC(graph: ModuleGraph) =
...
+  if conf.symbolFiles == disabledSf and not isDefined(graph.config, "cbackend2"):
     registerPass(graph, cgenPass)
 
     if {optRun, optForceFullMake} * conf.globalOptions == {optRun} or isDefined(conf, "nimBetterRun"):
       if not changeDetectedViaJsonBuildInstructions(conf, conf.jsonBuildInstructionsFile):
         # nothing changed
         graph.config.notes = graph.config.mainPackageNotes
         return
+  else:
+    graph.config.exc = excGoto # only goto exceptions are support for the new cbackend
```
saem, line 160: Thinking out loud, a smaller PR we could pull out of this one is removing the other exception types.

compiler/vm/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+Other parts:
+* `genCodeV3Exec`: a very early prototype of a code-generator for the VM, taking the IR as input and producing VM bytecode
+* `IrStore`, `IrNode2`: old attempts at the IR. Still kept around since there are some parts that might be reused
+* `computeInlining`: an early, non-working prototype of computing the most memory efficient IR generation order in the context of procedure inlining
+
+#### IR overview:
+* the IR is a linear node-based representation
+* nodes reference each other via indices. A node can only reference nodes coming before it; reference cycles are forbidden
+* it's still undecided if a node may be referenced multiple times
```
saem, line 40: Within the AST proper, no.

saem: I do see this as useful for something like the empty node, but I'm not convinced if this is the right thing.

compiler/ast/nimsets.nim
```diff
@@ -66,7 +66,9 @@ proc someInSet*(s: PNode, a, b: PNode): bool =
 
 proc inclTreeSet*(result: var TBitSetView, conf: ConfigRef; s: PNode) =
   ## Includes all elements from tree-set `s` into `result`
-  assert result.len == int(getSize(conf, s.typ))
+  # XXX: requiring the length to fit might help in catching some issues, but
+  #      it's too restrictive
+  assert result.len >= int(getSize(conf, s.typ))
```
saem, line 71: Should it instead introduce a new proc that is relaxed while keeping it strict elsewhere?

compiler/sem/sighashes.nim
```diff
@@ -51,9 +51,9 @@ type
     CoDistinct
     CoHashTypeInsideNode
 
-proc hashType(c: var MD5Context, t: PType; flags: set[ConsiderFlag])
+func hashType(c: var MD5Context, t: PType; flags: set[ConsiderFlag])
```
saem, line 54: Read this file, then thought:

🎶 won't you take me to... Func-y town🎶 😁

compiler/vm/README.md
```diff
@@ -0,0 +1,170 @@
...
+* control-flow is represented via gotos and joins. Instead of storing the index of the corresponding `join` target, a `goto` stores an index (`JoinPoint`) into a list storing the actual IR indices. This is aimed at making IR modification simpler, by removing the need to patch goto targets in the IR directly.
+* there exist a few special experimental gotos (goto-link-back, goto-with-continuation, goto-active-continuation) meant for more efficient `finally` handling
```
saem, line 42: Philosophical remark:

This feels like a theme in my experience.

It starts with a very small set of core primitives that are beautiful and all, great for learning and conveying key concepts. Then for making them actually useful a large set of variants are then introduced to encode constraints information.

compiler/vm/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### `hookPass`
+
+Meant to replace assignments and `mDestroy` calls with calls to the `=copy`, `=sink` and `=destroy` hooks (if they are present). This is a general pass that applies to all targets and garbage collectors.
+
+Introduced before `LinearPass2` existed and is thus missing the `mDestroy` patching.
+
+Currently also ignores whether or not a hook is trivial and thus replaces the assignment for types that don't actually need/use a `=copy` hook.
+
+#### `refcPass`
```
saem, line 78: Possibly another PR, we drop refc.

compiler/vm/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+All references to `seq` types need to be rewritten to use the `PGenericSeq` based type. Since seqs are generic, the new types need to be created as part of the pass (implemented).
+
+The types of locals, globals an parametes are currently adjusted (in a rather in-elegant manner), but field types also need to adjusted! This is not easily doable with the current surrounding architecture and a rewrite of how the types are adjusted is planned.
+
+A better approach for the type rewriting part would be to introduce a new kind of pass that operates only on types. `cbackend2` is then responsible for collecting all used types (this makes sense in general). The types could then also be run through a unification step, since most of the time, multiple `PType` instances exist for the exact same type (the VM also does this unification).
+
+As a further improvement, the IR should use it's own representation of types and symbols (a very early prototype exists in `irtypes.nim`).
+
+#### `seqsV2Pass`
```
saem, line 98: Another PR possibility, we drop one off these and only support one. It doesn't have to be 2, we can just have one that's most convenient and then upgrade that to one we want.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+    irpasses,
+    irdbg,
+    typeinfogen,
+    typeprocessing
+  ],
+  experimental/[
+    results
+  ]
+
+from compiler/vm/vmdef import unreachable
```
saem, line 48: Preferred: move `unreachable` to common code.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+    fileIdx: FileIndex
+    flags: TSymFlags
+
+    initProc: ProcId ## the 'init' procedure for the module. May be unset
+    dataInitProc: ProcId ##
+      ## the procedure responsible for:
+      ## * intializing RTTI related to the module
+      ## * running dynamic-library loading logic
+
+    # XXX: ``dataInitProc`` is not relevant to the VM target. Maybe the field
```
saem, line 79: Question: is this simply at present or should we not do so/have that in the future from a design perspective?

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+
+  # TODO: turn into a ``distinct uint32``
+  ModuleId = int ## The ID of a module in the back-end
+
+  ModuleListRef = ref ModuleList
+  ModuleList = object of RootObj
+    modules: seq[Module] ## ``ModuleId`` -> data collected via the ``passes``
+                         ## interface
+
+    # XXX: ``modulesClosed`` and ``moduleMap`` should be split off into a
```
saem, line 90: Thank you for identifying this lifetime separation, they quickly become subtle traps.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+  ModuleRef = ref object of TPassContext
+    ## The pass context for the VM backend. Represents a reference to a
+    ## module in the module list
+    list: ModuleListRef
+    index: int
+
+func growBy[T](x: var seq[T], n: Natural) {.inline.} =
+  x.setLen(x.len + n)
+
+iterator cpairs[T](s: seq[T]): (int, lent T) =
```
saem, line 110: Possible alternative name `pairsGrowable`

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+# TODO: rename to ``collectProcedureSyms``
+func collectRoutineSyms(ast: PNode, syms: var seq[PSym], marker: var IntSet) =
+  ## Traverses the `ast` and collects all referenced symbols of routine kind
+  ## to `syms` and `marker`
+  if ast.kind == nkSym:
+    let s = ast.sym
+    if s.kind in routineKinds:
+      collect(syms, s, marker)
+
+    return
```
saem, line 132: I think using a case here would be better as it then simply falls through sure to structure.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+
+  let tn = transformStmt(c.graph, c.idgen, c.module, ast)
+  let r = c.genStmt(tn)
+
+  if unlikely(r.isErr):
+    config.localReport(r.takeErr)
+
+  c.endProc()
+
+  if c.irs.len > 2:
```
saem, line 160: Recommend: make this a predicate func, as it's used twice.

This might contribute to a longer term encapsulation of irs behind context relevant operations.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+  ## procedure is empty, a 'none' is returned.
+  c.prc = PProc(sym: module.sym)
+  c.irs.reset()
+
+  c.startProc()
+
+  let ast =
+    if module.stmts.len > 1: newTree(nkStmtList, module.stmts)
+    elif module.stmts.len == 1: module.stmts[0]
+    else: newNode(nkEmpty)
```
saem, line 150:
```suggestion
    case module.stmts.len
    of 0: newNode(nkEmpty)
    of 1: module.stmts[0]
    else: newTree(nkStmtList, module.stmts)
```

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+  for id in mlist.modulesClosed.items:
+    let m {.cursor.} = modules[id]
+    if m.dataInitProc != NoneProc:
+      discard result.irCall(result.irProc(m.dataInitProc))
+
+    if sfSystemModule in m.flags:
+      # the 'init' procedure for the ``system`` module must be called right
+      # after the data init procedure
+      # XXX: that's what ``cgen`` does, but what's the reason behind this
+      #      requirement?
```
saem, line 230: I believe that's to ensure that `system` is fully initialized prior to other modules doing their thing. Because `system` is implicitly imported as the first thing this is an easy way to force correct ordering.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+
+# TODO: needs a different name:
+proc resolveTypeBoundOps(p: PassEnv, g: ModuleGraph, tgen: DeferredTypeGen, procs: var ProcedureEnv) =
+  for op, tbl in p.attachedOps.mpairs:
+    for k, v in g.attachedOps[op].pairs:
+      let t = tgen.lookupType(k)
+      if t != NoneType:
+        tbl[t] = procs.requestProc(v)
+      else:
+        discard#echo "missing type for type-bound operation"
```
saem, line 295: Internal error perhaps?

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+
+  if pos[0]:
+    echo "Node position: ", pos[1]
+
+  echo "IR:"
+  printIr(ir, env, calcStmt(ir))
+
+  printTypes(ir, env)
+
+template logError(ir: IrStore3, env: IrEnv, prc: ProcId, code: untyped) =
```
saem, line 328:
```suggestion
template withErrorLogging(ir: IrStore3, env: IrEnv, prc: ProcId, code: untyped) =
```

It's longer, but it might read better?

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+
+  for it in g.compilertypes.mvalues:
+    it = types.map(it)
+
+# TODO: needs a better name
+func processObjects(g: PassEnv, ic: var IdentCache, types: var TypeEnv, syms: var SymbolEnv,
+                    objects: Table[TypeId, PType]) =
+  ## Adds the ``m_type`` field to ``RootObj`` and rebases all non-final object
+  ## types that don't have a base type onto ``RootObj``. `objects` holds the
+  ## mappings from each object type to it's respective source type (i.e. the
```
saem, line 358:
```suggestion
  ## mappings from each object type to its respective source type (i.e. the
```

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+  # from ``RootObj`` plus all ``ntkPathObj`` nodes), the type is changed to
+  # have a negative field offset
+  types.setFieldOffset(rootType, -1)
+  types.replaceRecord(rootType):
+    [(syms.addDecl(ic.getIdent("m_type")), fieldType)]
+
+  for id, t in objects.pairs:
+    if id != rootType and types.base(id) == NoneType and
+       tfFinal notin t.flags and sfPure notin t.sym.flags:
+      # if an object has no ancestor, can be inherited from, and is not marked
```
saem, line 376: Do we have to do this late in the game because of the whole partial definition shenanigans?

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+      # ``RootObj`` (which provides the type field)
+      # XXX: the way this is implemented here differs from how it works in the
+      #      old back-end. There, a type field (a header) was directly
+      #      embedded into the object
+      types.setBase(id, rootType)
+
+func computeTypeFieldStatus(pe: PassEnv, types: TypeEnv,
+                            objects: Table[TypeId, PType]): TypeFieldInfo =
+  ## Computes and returns the type-field status for each type in `types`. This
+  ## procedure must be called after ``processObjects`` as it depends on the
```
saem, line 389: Just a thought: Can we enforce this ordering through lock levels? 🤔

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+    # owning module
+
+    for id in env.syms.items:
+      let s = env.syms[id]
+      case s.kind
+      of skVar, skLet, skForVar:
+        let mIdx = mlist[].lookupModule(env.syms.orig[id])
+        modules[mIdx].syms.add(id)
+
+      else:
```
saem, line 699: It doesn't need to be changed everywhere, but it's really rare that spacing out the case or if branches is more readable. A new line is a logical end, indentation change is another, two new lines is a very strong visual separator.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+        discard
+
+  # HACK: in the context of RTTI, closure types are represented via a "fake"
+  #       tuple type (one not matching the type they're lowered to).
+  # the later RTTI setup logic generation requires the types it processes to
+  # have corresponding entries in the `tfInfo` and `gcInfo` lookup tables,
+  # so the type has to be added to the environment before computing the
+  # tables
+  let fakeClosure = genFakeClosureType(env.types, passEnv)
```
saem, line 708: Not a blocker, just musing about the future: I think that semantic lowering like this would be better served by flagging analysis/lowering on or off in sem.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+
+  # XXX: mutable because they need to be swapped in and out of the ``RefcPassCtx``.
```
saem, line 710: More push for the no more refc case.

compiler/vm/cbackend2.nim
```diff
@@ -0,0 +1,1097 @@
...
+  #      used instead
+  var
+    tfInfo = computeTypeFieldStatus(passEnv, env.types, objects)
+    gcInfo = computeGcLookup(env.types)
+
+  var lpCtx = LiftPassCtx(graph: passEnv, cache: g.cache, env: addr env)
+  var ttc = TypeTransformCtx(graph: passEnv, ic: g.cache)
+  var upc = initUntypedCtx(passEnv, addr env) # XXX: not mutated - should be ``let``
+
+  # XXX: instead of manually figuring out out passes are to be batched
```
saem, line 722:
```suggestion
  # XXX: instead of manually figuring out how passes are to be batched
```

compiler/vm/cgen2.nim
```diff
@@ -0,0 +1,2450 @@
+## `vmir`-based C code-generator. Separated into two phases:
```
saem, line 1: I really like how conceptually simple this is. : 🎉

compiler/vm/irgen.nim
```diff
@@ -0,0 +1,2256 @@
...
+func pushScope(p: var PProc; finalizer, handler = JoinPoint(-1)) =
+  p.scopes.add Scope(firstLocal: p.activeLocals.len,
+                     finalizer: finalizer, handler: handler)
+
+func popScope(p: var PProc) =
+  # TODO: take ``PProc`` instead of ``TCtx``
+  let scope = p.scopes.pop()
+  p.activeLocals.setLen(scope.firstLocal)
+
+# XXX: the emission of ``ntkLocEnd`` instructions is disabled for now. When
```
saem, line 221: I'm still reading/understanding how this should work, but from a quick glance, the following occurred to me. Since we know there will be a start and stop for each could we:

1. preallocate two sequences of length equal to the number of locals (could also be a tuple)
2. location order and sequence order must match
3. each start/end has a positional offset, treat the offset as the instruction being appended with a ntkLocStart/ntkLocStop

saem: Tangentially, it occurs to me that perhaps a bunch of this dramatically simplifies thanks to CPS + structuring, as local lifetime cannot exceed the lifetime of the continuation, unless passed on (move/copy). The CPS transform should be during the semantic analysis phase as it'll change meaning for a number of things. But that should ease this nonetheless as the shape of "procs" you receive will be very small and regular.

Oh well, that's all one fine day right now. 😅

compiler/vm/irpasses.nim
```diff
@@ -0,0 +1,2634 @@
...
+  TypeContext* = object
+    ## Stores the type information for a procedure's body
+
+    # TODO: introduce the ``IdMap[SomeId, T]`` type that's already mentioned
+    #       elsewhere and use it here
+    orig: seq[TypeId] ## ``IRIndex`` -> unmapped type of the expression
+    mapped: seq[TypeId] ## ``IRIndex`` -> mapped type of the expression. A
+      ## mapped type is the type after lowering/transformation.
+
+  TypedPass*[T] = object
```
saem, line 122: `TypePassVistor`

saem (commented):
I've done a first pass. I can only absorb so much in one go and around `irpasses` I was really fading.

----

In terms of how to proceed as a strategy:

1. we need to shrink dialect problems in devel
2. unit tests directly against cbackend2 `generateCode`

My reasoning for the first point:
- even with the code working for say `refc` it's that much more to review and test
- CI will be faster, I suspect test cycles are going to be a limiting factor here
- it'll simplify sem's output, which will clarify/simplify things further here, and likely cycle again reinforcing each other
- we need to do it anyways and it'll bring the goal of getting things merged a lot closer

The reasoning for the second point:
- Cyo will be a new language at this point (grammar and semantics), its frontend will be separate and I'd like to reuse the backend and VM
- tests will enforce and maintain strong separation
- they'll allow for refactoring without having to go through a frontend

----

Practically, I think it should work by us looking at and discussing CI failures and then:

1. if you could makes fixes for the things we think should continue to work
2. I can drop the dialects/whatever... for the things we're going to remove

Simultaneously, for issues you're encountering during implementing fixes:
- if you can add some direct unit tests to hold necessary properties in place
- I can try to make sem less dumb to make your life easier 🤞🏽 

As those lines converge, we end up with all tests passing and an easy merge. Thoughts?

compiler/vm/cgen2.nim
```diff
@@ -0,0 +1,2450 @@
...
+    result.args[i] = (mapTypeV3(p.typ),
+                      ident)
+    inc i
+
+
+template start(): CAstBuilder =
+  var b: CAstBuilder
+  b
+
+template buildAst(code: untyped): CAst =
```
Clyybber, line 622: I think
`start().....fin()` or `fin: start()....` is better than `buildAst: builder....` as it doesn't require knowledge of another template and its `builder` variable.

compiler/front/main.nim
```diff
@@ -146,22 +148,27 @@ proc commandCompileToC(graph: ModuleGraph) =
   let conf = graph.config
   extccomp.initVars(conf)
   semanticPasses(graph)
-  if conf.symbolFiles == disabledSf:
+  if conf.symbolFiles == disabledSf and not isDefined(graph.config, "cbackend2"):
```
Clyybber, line 151: Can be flipped to 
```nim
if isDefined(graph.config, "cbackend2"): ...
elif conf.symbolFiles == disabledSf: ...
```
as you did in line 169.

compiler/vm/README.md
zerbina: As `refc` support is fully implemented already, I'd be a bit reluctant to remove it either before or as part of this PR and instead would rather remove it after. The implementation is then preserved in the git history at least.

One thing to consider is that the `refc`, `markAndSweep`, `boehm`, and `go` GC support share almost all of their implementation in the compiler, so only removing `refc` would not reduce (by a significant amount) the required code/complexity.

compiler/vm/cbackend2.nim
zerbina: I agree, but I haven't yet figured out where to. Ideally, the template should be located in a separate module in `compiler/utils` (or maybe even in the standard library), but I don't know how that module should be named.

compiler/vm/cbackend2.nim
zerbina: I'm not sure I understand the question. The kind of setup/initialization the `DatInit` procedures perform for the C target is performed by the VM on the host side.

compiler/vm/cbackend2.nim
zerbina: Yep, that's better.

compiler/vm/README.md
saem: Overall, ok we don't have to do it just yet.

As for the other GCs I would drop all of those as well.

compiler/vm/cbackend2.nim
saem: `compiler/utils/idioms` might be better than another utils or a 'commons'.

compiler/vm/cbackend2.nim
saem: I was unsuccessfully asking, do we foresee this changing for the VM target as the use cases expand?

It seems unlikely.

compiler/vm/cbackend2.nim
zerbina: The question was in the direction of why the `init` procedure for `system` is called before the `DatInit` procedure of other modules.

Your assessment is correct, it's necessary because the `DatInit` procedures are also responsible for loading used dynamic libraries and for setting up the procedures/globals depending on them, which is something that requires various `system` parts to be initialized (`cstderr`, the GC/allocator, etc.).

compiler/vm/cbackend2.nim
zerbina: The `echo` is a debugging remnant that I forgot to remove.

Not all existing types are part of the alive graph, and only those that are are translated. For types not part of the alive graph, `lookupType` returns `NoneType`, so the `else` case is expected behaviour and not an error.

compiler/vm/cbackend2.nim
zerbina: It reads better, and since the template is also only used for debugging, I don't think the longer name is an issue.

compiler/vm/cbackend2.nim
zerbina: I'm not sure what you mean by "partial definition" in this case?

How information about a location's dynamic type is stored is target-dependent (C/C++ use an extra `m_type` field part of the object, the VM keeps track of them separately, etc.), so I think that it makes sense to have this logic be part of the back-end.

The logic here is meant for the C and C++ targets and is responsible for making sure that inheritance roots that are non-pure have a `m_type` field by modifying them to have `RootObj` as their base type.

compiler/vm/cbackend2.nim
zerbina: Could you elaborate on that? I'm not sure I understand.

I am/was under the impression that sem should be unaware of a target's implementation details. The `fakeClosure` type is a slight hack that is only needed for the `TNimType`-based RTTI, which is only used for the C and C++ targets if `optTinyRtti` is not enabled.

compiler/front/main.nim
zerbina: Yep, that's cleaner, thanks.

compiler/vm/cbackend2.nim
saem: I meant, I thought we could know this information sooner and what's restricting that is the `partial` (hope I got the name right) pragma. Which defines a type over multiple modules.

compiler/vm/cbackend2.nim
saem: Yup, I think sem should never ever ask which backend and then make a decision, so your absolutely correct about that.

What I meant was when the compiler is choosing a backend and then sets up the passes, it will configure sem to apply some optional transforms. These optional transforms are uniform in implementation but aren't universally enabled for all backends (at time of pass setup).

compiler/vm/cbackend2.nim
zerbina: I also don't see this changing. If the VM is going to support the `.dynlib`-related features, the support for them is very likely going to implemented on the host side of things.

compiler/vm/cbackend2.nim
zerbina: Ah, I believe you mean package-level objects (i.e. `.package`)?

I thought about whether the injection of the `m_type` field is something that can be done during the semantic-analysis phase, and in theory, that's possible. The only problem I currently see is that that non-pure objects used in the context of compile-time execution would also have the `m_type` field then, which is unnecessary, as the VM doesn't need it.

compiler/vm/cbackend2.nim
zerbina: To make sure I understand correctly: if a `case|if` branch doesn't end in a nested statement/expression, it's preferred to not use a new line?

compiler/vm/cbackend2.nim
zerbina: The issue is not directly with `refc`, but instead with first-class view-types being unfinished.

Both sequences need to be available to various passes (those using `RefcPassCtx`), but passing them as parameters is not possible due to the `TypedPass` interface, and storing them as part of `RefcPassCtx` directly would be wrong, as they are not something that directly belongs to the context object.

The context object should ideally borrow from the sequences instead (via `lent seq` in this case), but since that's currently not possible, I've used the swap-in-swap-out idiom as a way to mimic borrowing. Using `shallowCopy`/`.cursor` or a pointer don't require the source sequences to be mutable, but they have other downsides.

compiler/vm/irgen.nim
zerbina: Since scopes (in the context of lifetimes) aren't directly encoded in the IR, the original idea was to use `ntkLocEnd` as a way to signal the end-of-life of a location on a control-flow path. This information was meant to be used by the move-analyser and related data-flow analysis, as well as the code-generator for the VM (to simplify register allocation).

Because each local can have more than one associated `ntkLocEnd` instruction, the sequence idea wouldn't work. Attaching extra out-of-band information to instructions is also a bit problematic right now, as the attachment position have to be adjusted separately on each code modification.

Aside: Instructions might need a stable ID for other things, but because of the additional memory they'd take up, I'm still trying to get around having to introduce them.

With the change of plan (i.e. the `injectdestructors` rewrite that I'll be working on first), most of the ideas around `ntkLocEnd` have become obsolete/stale.

compiler/vm/cbackend2.nim
saem: Yes, that's what I meant. Hmmm, I guess nothing to change for now.

compiler/vm/cbackend2.nim
saem: I'm still working through what's tasteful, but that's not a bad way to look at it.

The only time I find separating it out helps is when there are many branches in a single `case` and they don't fit the screen vertical space. An example of this would be the report body printing in cli_reporter.

Sorry this is still a bit fuzzy, I'm refining it as fast as I can. My mental model is, the eye is rapidly skimming down the lines, a blank line is a big interrupt to that, so the conceptual separation between the contiguous lines before and the lines thereafter needs to be very meaningful. Often times with conditional branches they're not really conceptually separate. They are in that report body scenario, because you can mentally elide every other branch and the proc body reads like a specialized print for that one case.

compiler/vm/irgen.nim
saem: Thank you for through explanation.

After coming back to this after some time, I think I get it more than I did before. `ntkLocEnd` is effectively acting as a "consume" instruction being automatically inserted by the compiler. This needs to be branch aware at present... but also leads me back to thinking it simplifies under CPS. 😅 

But don't worry about all that jazz, it's just me thinking out loud.

zerbina:
Update: I'm currently in the process of splitting off (iterated upon versions of) the various bits developed here into separate PRs, as detailed by the plan described [here](https://github.com/nim-works/nimskull/discussions/467).

It's likely that this PR itself is never going to be merged, but I'm leaving it open for now, since some of the discussions here are still relevant.

zerbina:
This PR is outdated and has been superseded by #1333.

Issue #423, opened 2022-08-31
Decouple parser from PNode with DOD approach


saem:
Upgraded parser (and possibly lexer/tokenizer) that conforms to a more data oriented design approach as taken in the VM and new Backend (in progress). The semantic analysis layer won't be able to take advantage of it yet, but other tools like filters and pretty might.

## The approach at the out set:
- decouple the parser from `PNode`
- improve parser data types to have a more data oriented approach and get it working
- look at the DOD approach and see what rough edges/misfeatures need fixing, replacement, or removal; do some of those

NB: like any plan revise and improve the approach as we learn more. At the end we should have a faster parser which is nicer to memory, with more tests/validation against the grammar, fewer oddities in the grammar, clearer node kinds, decoupling from semantic analysis, and a path for tools like source filters, formatters, etc, easily making effective use of the new parser.

### Decouple the parser from `PNode`

The parser is coupled to `PNode` right now which takes a sea of nodes on the heap approach. First step make a `ParsedNode` (or whatever) that's basically a very slim `PNode`, add a proc to convert to `PNode`, and get everything working as before.

### Improve parser data types

Start revising the parser produced data types with a more DOD style. Start with a `ParsedFragment` that is a bit more than a sequence of parsed nodes (minimal object) and some metadata. Internally it should be a DOD layout and be produced by `parseAll`, `parseTopLevelStmt`, etc. A fragment could be some arbitrary string of AST or a while file/module, statement, etc. The idea is this is the parser's production for a single parse action. If a whole file vs fragment distinction proves to be useful then do that.

First cut data type will be easy and then getting the parser code to use it will be where it gets interesting focus on getting it working and keeping the data model as clean as possible. Mark all the weirdness so we know what to come back to next.

### Smooth out rough edges

The above will get it working, now we clean things up so we're learning from the design smells showing up in the data model.

### Final remarks on approach

This doesn't have to be a big bang. Quick wins regardless of this work should be PR'd separately ahead of time. If the first part of the approach doesn't hurt performance too much then that can come in on its own. If the second is clean enough then ditto. Other intermediate milestones might be discovered and get pulled in early.

----------

todo steps added by @haxscramper, collected from the discussion in https://github.com/nim-works/nimskull/pull/425

- [ ] Clean up the parser implementation - this PR. It introduces another slowdown of the parsing stage, but not too drastic, and immediately enables decoupling. 
[...]

PR #425, opened 2022-09-01
decouple parser from PNode

haxscramper:
wip implementation of the https://github.com/nim-works/nimskull/issues/423


haxscramper:
@saem this is a first pass on the implementation, which uncovered some implementation questions that I wrote down in changes - look them over and tell me about your thoughts on direction/solution. I'm considering whether I should also include the change to the DOD lexer in this PR or split it into a separate subsequent one.

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,104 @@
...
+                     # the token, which in turn will store information
+                     # about global positioning (tuple made up of a token
+                     # id and a file ID)
+                     #
+                     # NOTE technically this is not really necessary even
+                     # with the current implementation, but the parser
+                     # consistently copies this information around anyway,
+                     # so I will let it stay this way for now.
+    token*: Token # TODO Replace full token value with an index information
+    kind*: TNodeKind # NOTE/QUESTION - for now the same kind of nodes is
```
saem, line 33: Yes, or at least one will be a super set of the other. Ultimately separate enums are likely best.

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,104 @@
...
+
+    # TODO replace `ref` object tree with begin/end ranges for the nested
+    # trees in the linearized structure.
+    subnodes*: seq[ParsedNode]
+    comment*: string # TODO this should either be a token or a sequence of
+                     # tokens.
+
+    # HACK explicit flags in order to track down all 'extra' information
+    # that is collected during parsing.
+    isBlockArg*: bool # QUESTION add 'nkStmtListBlockArg' or similar node
```
saem, line 46: You found it too!

OK, so this dumb thing is used to tell the semantic analysis layer about a statement list (possibly also block) that are being passed to a macro, template, whatever. During semantic analysis we collapse/flatten nested statement lists into one level, but we also do statement list unwrapping, so if there is only one node in a statement list we unwrap it except if there is a defer (I think this might be a bug/design flaw).

That first flattening is guarded by this dumb flag. I personally think we should always flatten, no matter what, but never unwrap in semStmtList (it's the last/second last proc in semstmts, I rewrote it recently so it should be readable). Without that guard a macro can receive arbitrary chunks of AST not always wrapped in a statement list and it makes the really annoying/painful to write.

I do believe between fixing the unpacking logic and the AST the parser generates this shouldn't be a problem. My guess at the most correct fix is sematic analysis layer should:
- unwrap as determined by the receiving node, so be done on add/index assign based on receiving node kind
- flattening should always take place in semantic analysis of a statement list because that's just always right

Then this dumb guard should not be necessary. Separating the AST used by sem, parsing, and backend means we can change add and index assign behavior and it only impacts sem so we can do destination based unwrapping rather than have it live incorrectly in semStmtList. 🎉

Sorry that was a rather lengthy explanation.

compiler/ast/ast_query.nim
```diff
@@ -97,6 +97,8 @@ const
   callableDefs* = nkLambdaKinds + routineDefs
 
   nkSymChoices* = {nkClosedSymChoice, nkOpenSymChoice}
+  nkFloatKinds* = nkFloatLiterals # QUESTION remove float literals
```
saem, line 100: I don't think so, but maybe I'm missing some key point? Whatchya thinking?

compiler/ast/parser.nim
```diff
@@ -1854,8 +1852,8 @@ proc parseEnum(p: var Parser): PNode =
     var symPragma = a
     if (p.tok.indent < 0 or p.tok.indent >= p.currInd) and p.tok.tokType == tkCurlyDotLe:
       let pragma = optPragmas(p)
-      symPragma = newTreeI(nkPragmaExpr, p.lineInfo, [a, pragma])
-    # nimpretty support here
+      symPragma = p.newNode(nkPragmaExpr, @[a, pragma])
+    # nimpretty support here LOL what is this even
```
saem, line 1844: Yeah, it's likely some dumb leak of Implementation because it was "easier" to do this silly thing. 😄

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,104 @@
...
+                      # and convert it to the `nkStmtList` + `nfBlocArg`
+                      # flags later on? Why do we need the `nfBlockArg`
+                      # flag in the first place?
```
zerbina, line 49: It's purpose is to prevent the flattening of `nkStmtList`s with only one item. Saem started a discussion related to it [here](https://github.com/nim-works/nimskull/discussions/385#discussioncomment-3217640) .

**Edit**: Sorry for the redundant answer. When I started the review, saem's review wasn't published yet :D

saem (commented):
I think I got them all, the PR is already really starting to shape up.

compiler/ast/parser.nim
```diff
@@ -1309,32 +1298,32 @@ proc primary(p: var Parser, mode: PrimaryMode): PNode =
...
   if p.tok.tokType == tkNot:
-    let notOpr = newIdentNodeP(p.tok.ident, p)
+    let notOpr = p.newNode(nkIdent, p.tok)
     p.getTok
     p.optInd(notOpr)
-    newTreeI(nkInfix, p.lineInfo, [notOpr, a, parseExpr(p)])
+    return p.newNode(nkInfix, @[notOpr, a, parseExpr(p)])
   else:
-    a
+    return a
```
zerbina: Hm, what is the reason you changed to the use of `return` here?

In the case an if-expression is not wanted here, I think a `result = ...` would fit better.

compiler/sem/passes.nim
```diff
@@ -182,15 +182,15 @@ proc processModule*(graph: ModuleGraph; module: PSym; idgen: IdGenerator;
 
     while true:
       if graph.stopCompile(): break
-      var n = parseTopLevelStmt(p)
+      var n: PNode = parseTopLevelStmt(p).toPNode()
```
zerbina:
```suggestion
      var n = parseTopLevelStmt(p).toPNode()
```

zerbina (commented):
I answered a question and left some small suggestions regarding code style.

compiler/ast/ast_query.nim
haxscramper: We have node *kinds* for everything and only float *literals*. It is a naming question, and it is a set of *node kinds*, not a set of *literals*. Also `nkLiterals` is inconsistent with Atomic kinds in the `macros.nim`, but that's another question. I think `nkLiteralKinds`, `nkFloatKinds` or even `nkFloatLiteralKinds` make more sense (the last one is the most sensible IMO, but a bit longer)

TBH I always found this enum set definitions a bit weirdly structured and lacking in some places, but it is hard to see the underlying structure immediately from the code, it must be an incremental improvement

haxscramper: https://github.com/nim-works/nimskull/blob/837238f3cbebb6d52be6c328624f6f5598b44788/compiler/ast/ast_query.nim#L687 I think those two must be intrinsically related to each other (the collection of the enum kind sets and the `[]` named indexing operator), but for now that's just a new idea I got while elaborating on the "literals/kinds" distinction above

compiler/ast/parser.nim
haxscramper: I think it would be possible o keep nimpretty in the parser for the time being, but as soon as I'm done with DOD part of the lexer it will have to go away and finally tackle the "teach a system without eyes how to make code visually appealing, how hard can it be? :P" problem.

compiler/ast/parser.nim
haxscramper: ok, I will change it to the `result=`, I don't have any specific preference for the `return` over `result=`

haxscramper:
> I think I got them all, the PR is already really starting to shape up.

@saem I think the optimal plan would be to split in following steps (PRs)

1. Clean up the parser implementation - this PR. It introduces another slowdown of the parsing stage, but not too drastic, and immediately enables decoupling. 
2. Throw away nimpretty hacks - https://github.com/nim-works/nimskull/discussions/113
3. Transition lexer to use DOD approach - I'm thinking of a global `FileIndex -> seq[Token]` mapping that is generated by the lexer. Each token includes a line, a column and a tag information. And an extent range. 
  - `ParsedNode` includes `(FileIndex, TokenIndex)` pair that is used to get the token back from the main storage
  - No tokens are discarded from the tokenizer - Non-documentation comments, obviously included. I'm not *exactly* sure about the INDENT/DEDENT/SAME-INDENT token set, but I think we can also retain them. 
4. Change the `ParsedNode` to use a DOD structure as well

compiler/ast/ast_query.nim
saem: For `nkFloatLiterals` vs `nkFloatKinds`, the former is literals only and the latter might include more, but I can't think of a single thing not covered by the former. The is a difference in meaning but I'm really hard pressed to think of a case where they would not be equivalent. I would check the sites and consolidate personally.

----

> https://github.com/nim-works/nimskull/blob/837238f3cbebb6d52be6c328624f6f5598b44788/compiler/ast/ast_query.nim#L687
> 
> I think those two must be intrinsically related to each other (the collection of the enum kind sets and the `[]` named indexing operator), but for now that's just a new idea I got while elaborating on the "literals/kinds" distinction above

That's exactly the types of insights that form by looking at this stuff.

I agree with the sentiment, the naming and categorization needs to make sense based on purpose and not solely because a new programmer's first instincts are to lump them together.

So to recap:
- naming needs to be more consistent
- naming needs to better convey intention
- sets need to serve a purpose(s), eg: many sites in the compiler need to handle all literals one way (`nkLiterals`) or lots of literal handling is weird about floats vs other numbers (`nkFloatLiterals`)

Doesn't have to be those precise names but I think we're on the same page. For actually changing sets carefully looking through their usage is required because there are unfortunate subtleties.

compiler/ast/parser.nim
saem: After some of these things design issues/boundary violations are resolved, we'll no longer be hobbled by bugs and limitations. As in no more silly compromises. That alone will do wonders for simplifying the formatting problem.

compiler/ast/parser.nim
saem: I like to assign to result once, or at least as few times as possible, and try to avoid return.

Effectively the code should be really linear and a simple fall through. Then the reader can jump to the end and that's also the logical end. With mid-proc returns the logical end and written end don't match. All in an effort to let people (including future me) quickly skim code, get the gist of things, and keep flow.

saem:
> > I think I got them all, the PR is already really starting to shape up.
> 
> @saem I think the optimal plan would be to split in following steps (PRs)

This does look better and I'm sure it'll be revised more as we continue, let's do it. Additional benefit, smaller chunks, easier review, testing, etc.

> 1. Clean up the parser implementation - this PR. It introduces another slowdown of the parsing stage, but not too drastic, and immediately enables decoupling.

Cool, that's what I was thinking things are shaping up around.

> 2. Throw away nimpretty hacks - [`nimpretty` #113](https://github.com/nim-works/nimskull/discussions/113)

Yes, less hacks and more Hax. :D

> 3. Transition lexer to use DOD approach - I'm thinking of a global `FileIndex -> seq[Token]` mapping that is generated by the lexer. Each token includes a line, a column and a tag information. And an extent range.

Yup, will need to support:
- lex me a string (for reasons we have a fragment of code)
- lex me a string which is a fragment of a file

But I'm assuming that's implied. The former is because we might be asked to handle various code fragments like repls and cases where we don't have a file perse. The latter is where we are formatting a subsection of some file. The amount of effort and support needed here might be near zero. Just highlighting two interesting cases besides the majority case of program/file(s) compilation.

> * `ParsedNode` includes `(FileIndex, TokenIndex)` pair that is used to get the token back from the main storage

Yup, then it would only need the left and right index and probably be done.

> * No tokens are discarded from the tokenizer - Non-documentation comments, obviously included. I'm not _exactly_ sure about the INDENT/DEDENT/SAME-INDENT token set, but I think we can also retain them.

Based on what I could glean from the paper and @alaviss 's remarks I think knowing this is important for these reasons:
- need indent level (logical) for preserving meaning
- need actual spaces (physical) for spans, spaces for alignment, run length encoding/compressing the stream

> 4. Change the `ParsedNode` to use a DOD structure as well

Yup, I think it'll be easier if the lexer has had a once over already.

haxscramper:
Step \#5 is to add reduced parsed node set - based on the `TNodeKind`, but without sem-related data. I initially forgot about that part - while `ParsedNode` and `PNode` are structurally similar, the latter one has a lot *more* states. 

`nkError` probably should stay if we want to get parser error correction in later.

compiler/ast/ast.nim
```diff
@@ -604,3 +606,39 @@ proc toHumanStr*(kind: TSymKind): string =
 proc toHumanStr*(kind: TTypeKind): string =
   ## strips leading `tk`
   result = toHumanStrImpl(kind, 2)
+
+
+proc setBaseFlags(n: PNode, base: NumericalBase) =
```
saem, line 612: All your base are belong to us.

haxscramper: :bulb:   There should be a separate node kind for different literals

compiler/sem/passes.nim
```diff
@@ -84,13 +85,26 @@ proc closePasses(graph: ModuleGraph; a: var TPassContextArray) =
       m = graph.passes[i].close(graph, a[i], m)
     a[i] = nil                # free the memory here
 
-proc processTopLevelStmt(graph: ModuleGraph, n: PNode, a: var TPassContextArray): bool =
-  # this implements the code transformation pipeline
-  var m = n
-  for i in 0..<graph.passes.len:
+proc processTopLevelStmt(
```
saem, line 88: Nice, this no longer hurts me when I read it. Thank you for putting an end to the violence.

compiler/ast/parser.nim
```diff
@@ -1358,22 +1324,30 @@ proc postExprBlocks(p: var Parser, x: PNode): PNode =
...
-                   newProcNode(nkDo, stmtList.info, body = stmtList,
+                   newProcNode(
+                     nkDo,
+                     stmtList.info,
+                     body = stmtList,
                      params = if openingParams.kind != nkEmpty: openingParams
-                              else: newTreeI(nkFormalParams, p.lineInfo, p.emptyNode),
-                     name = p.emptyNode, pattern = p.emptyNode, genericParams = p.emptyNode,
-                     pragmas = openingPragmas, exceptions = p.emptyNode)
+                              else: p.newNode(nkFormalParams, @[p.emptyNode]),
```
saem, line 1351: I think this is missing line info

haxscramper: line information is added automatically from the parser in

```nim
proc newNode(
    p: Parser,
    kind: TNodeKind,
    subnodes: sink seq[ParsedNode] = @[],
    info: TLineInfo = p.lineInfo
  ): ParsedNode =
  ## Create a new non-leaf parsed node using line information from the
  ## parser p
  return newParsedNode(kind, info, subnodes)
```

haxscramper: `else: newTreeI(nkFormalParams, p.lineInfo, p.emptyNode),` -> `.newNode(nkFormalParams, @[p.emptyNode]),`

compiler/ast/parser.nim
```diff
@@ -247,30 +247,38 @@ proc indAndComment(p: var Parser, n: PNode, maybeMissEquals = false) =
...
+  ## Create a new ident leaf parsed node using line information from the
+  ## parser and provided identifier as a token.
+  # HACK this is a temporary transition function until ParsedNode can fully
+  # switch over to using token indices
+  #
+  # NOTE this handling loses information about original token indentation
+  # and positioning
+  return newParsedNode(
+    kind = nkIdent,
+    info = p.lineInfo,
```
saem:
```suggestion
    info = info,
```

compiler/ast/parser.nim
```diff
@@ -666,101 +670,57 @@ proc identOrLiteral(p: var Parser, mode: PrimaryMode): PNode =
...
   of tkCustomLit:
     let splitPos = p.tok.iNumber.int
-    let str = newStrNodeP(nkRStrLit, p.tok.literal.substr(0, splitPos-1), p)
-    let callee = newIdentNodeP(getIdent(p.lex.cache, p.tok.literal.substr(splitPos)), p)
-    result = newTreeI(nkDotExpr, p.lineInfo, [str, callee])
+    var tok = p.tok
+    tok.literal = p.tok.literal.substr(0, splitPos - 1)
+    let str = p.newNode(nkRStrLit, tok)
+    let callee = p.newNode(
+      # TODO(lexer) implement custom literals using two tokens - the
```
saem, line 703: Strongly agree

compiler/ast/parser.nim
```diff
@@ -666,101 +670,57 @@ proc identOrLiteral(p: var Parser, mode: PrimaryMode): PNode =
...
+  of tkTripleStrLit: result = p.newNode(nkTripleStrLit, p.tok); p.getTok
+  of tkCharLit:      result = p.newNode(nkCharLit,      p.tok); p.getTok
   of tkCustomLit:
     let splitPos = p.tok.iNumber.int
-    let str = newStrNodeP(nkRStrLit, p.tok.literal.substr(0, splitPos-1), p)
-    let callee = newIdentNodeP(getIdent(p.lex.cache, p.tok.literal.substr(splitPos)), p)
-    result = newTreeI(nkDotExpr, p.lineInfo, [str, callee])
+    var tok = p.tok
+    tok.literal = p.tok.literal.substr(0, splitPos - 1)
+    let str = p.newNode(nkRStrLit, tok)
```
saem, line 701: Maybe I'm totally off base here, but I don't think the before and after are equivalent, in the former code we're substring-ing on the same literal. While in the latter we're doing a sub string on the result of the first to get the second?

saem: Oh maybe it's a value type ... checking

saem: Yup, value type.

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,111 @@
...
+
+iterator pairs*(node: ParsedNode): (int, ParsedNode) =
+  for idx, item in pairs(node.sons):
+    yield (idx, item)
+
+proc add*(node: ParsedNode, other: ParsedNode) =
+  ## Add new element to the sons
+  node.sons.add(other)
+
+proc transitionSonsKind*(n: ParsedNode, kind: TNodeKind) =
```
saem, line 76: I'm wondering if there is unguarded sons.len access in the parser that doesn't look at kind and we're not enjoying sons on certain kinds?

Might be worth adding a check for transition to leaf kind and clear sons.

compiler/sem/passes.nim
```diff
@@ -84,13 +85,26 @@ proc closePasses(graph: ModuleGraph; a: var TPassContextArray) =
...
+  var processingResult = toProcess
+  for i in 0 ..< graph.passes.len:
     if not isNil(graph.passes[i].process):
-      m = graph.passes[i].process(a[i], m)
-      if isNil(m): return false
+      # Iterate over all processing passes, re-assigning the evaluation
+      # results each time.
+      processingResult = graph.passes[i].process(a[i], processingResult)
+      if isNil(processingResult):
+        # QUESTION this is an error/fatal diagnostic or some parts of the
```
saem, line 103: I think sem and backend can both end up doing this especially through legacy code paths.

compiler/sem/passes.nim
```diff
@@ -176,39 +208,99 @@ proc processModule*(graph: ModuleGraph; module: PSym; idgen: IdGenerator;
...
-          sl.add n
-        #echo "-----\n", sl
-        if not processTopLevelStmt(graph, sl, a): break
-        if rest != nil:
-          #echo "-----\n", rest
-          if not processTopLevelStmt(graph, rest, a): break
+
+        # Iterate until non-imperative statement is found (or end is
+        # reached, in which case parser retuns an empty node)
+        while rest.isNil():
```
saem, line 270: OK, I don't think this is equivalent to what was there before. I need to reread it a few times, but we collect chunks until a proc def/type section/etc. It should simply connect all that stuff as one statement list and pass that on, but it looks like we might be breaking it up further and that's causing issues.

compiler/ast/parser.nim
haxscramper:
```
str = thing[0, splitPos - 1]
callee = thing[splitPos]
result = dotExpr[str, callee]
```

In both cases

compiler/ast/ast_parsed_types.nim
haxscramper: I will add it in later PRs when DOD will be involved, maybe even try to get rid of this transition logic in parser

compiler/ast/parser.nim
```diff
@@ -134,7 +135,8 @@ proc openParser*(p: var Parser, fileIdx: FileIndex, inputStream: PLLStream,
     openEmitter(p.em, cache, config, fileIdx)
   p.getTok                   # read the first token
   p.firstTok = true
-  p.emptyNode = newNode(nkEmpty)
+  p.emptyNode = newParsedNode(nkEmpty)
+  p.emptyNode.info = unknownLineInfo
```
saem: You don't need this, it's already set in the constructor.

compiler/ast/ast.nim
```diff
@@ -604,3 +607,42 @@ proc toHumanStr*(kind: TSymKind): string =
...
+      result.strVal = parsed.token.literal
+
+    of nkIdent:
+      result.ident = parsed.token.ident
+
+    else:
+      if parsed.isBlockArg:
+        result.flags.incl nfBlockArg
+
+
```
saem, line 645:
```suggestion

```

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,111 @@
+## Data structure for the parser results
+
+import
+  compiler/ast/[
+    ast_types, # For the node kinds
+    lexer # For the token type definition
+  ]
+
+# NOTE further refactoring considerations for the parser
```
saem, line 9:
```suggestion
# NOTE further refactoring considerations for the parser:
```

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,111 @@
...
+                     # the token, which in turn will store information
+                     # about global positioning (tuple made up of a token
+                     # id and a file ID)
+                     #
+                     # NOTE technically this is not really necessary even
+                     # with the current implementation, but the parser
+                     # consistently copies this information around anyway,
+                     # so I will let it stay this way for now.
+    token*: Token # TODO Replace full token value with an index information
+    kind*: TNodeKind # NOTE/QUESTION - for now the same kind of nodes is
```
saem, line 33: Yes, separate kinds.

See `ast_query.nodeKindsProducedByParse` for a shortened list: https://github.com/nim-works/nimskull/blob/devel/compiler/ast/ast_query.nim#L112

compiler/ast/ast_parsed_types.nim
saem: We could also add a double nkStmtList, then remove unwrapping of stmtLists from semStmtList and have `ast.add(n, child: PNode)` take `n.kind` into account and do unwrapping as required.

This is a more holistic fix and doesn't require a new node -- should be a reasonable standalone PR.

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,111 @@
...
+
+    # HACK explicit flags in order to track down all 'extra' information
+    # that is collected during parsing.
+    isBlockArg*: bool # QUESTION add 'nkStmtListBlockArg' or similar node
+                      # and convert it to the `nkStmtList` + `nfBlocArg`
+                      # flags later on? Why do we need the `nfBlockArg`
+                      # flag in the first place?
+
+func len*(node: ParsedNode): int =
+  ## Number of the sons of a parsed node
```
saem, line 52:
```suggestion
  ## Number of sons of the parsed node
```

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,111 @@
...
+  return node.sons.len()
+
+# NOTE added for the sake of API similarity between PNode
+proc safeLen*(node: ParsedNode): int = node.len()
```
saem, line 56:
```suggestion
func safeLen*(node: ParsedNode): int {.inline.} = node.len()
```

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,111 @@
...
+iterator items*(node: ParsedNode): ParsedNode =
+  for item in node.sons.items():
+    yield item
+
+iterator pairs*(node: ParsedNode): (int, ParsedNode) =
+  for idx, item in pairs(node.sons):
+    yield (idx, item)
+
+proc add*(node: ParsedNode, other: ParsedNode) =
+  ## Add new element to the sons
```
saem, line 73:
```suggestion
  ## append `other` to `node`'s `sons`
```

compiler/ast/ast_parsed_types.nim
```diff
@@ -0,0 +1,111 @@
...
+  return ParsedNode(kind: kind, info: unknownLineInfo)
+
+func newParsedNode*(
+  kind: TNodeKind, info: TLineInfo, sons: seq[ParsedNode] = @[]): ParsedNode =
+  ## Create a new non-leaf parsed node with a specified location
+  ## information and sons.
+  return ParsedNode(kind: kind, info: info, sons: sons)
+
+func newParsedNode*(kind: TNodeKind, info: TLineInfo, token: Token): ParsedNode =
+  ## Create a new leaf parsed node with the specified location information
```
saem, line 96: technically we should probably assert on this.

compiler/ast/lexer.nim
```diff
@@ -517,6 +517,7 @@ proc getNumber(L: var Lexer, result: var Token) =
           # XXX: Test this on big endian machine!
         of tkFloat64Lit, tkFloatLit:
           setNumber result.fNumber, (cast[PFloat64](addr(xi)))[]
+
```
saem, line 520:
```suggestion
```

one less file in the diff.

compiler/ast/parser.nim
```diff
@@ -75,31 +76,31 @@ type
...
-proc parseStmtPragma(p: var Parser): PNode
-proc parsePragma(p: var Parser): PNode
-proc postExprBlocks(p: var Parser, x: PNode): PNode
-proc parseExprStmt(p: var Parser): PNode
-proc parseBlock(p: var Parser): PNode
-proc primary(p: var Parser, mode: PrimaryMode): PNode
-proc simpleExprAux(p: var Parser, limit: int, mode: PrimaryMode): PNode
+proc optInd*(p: var Parser, n: ParsedNode)
+proc indAndComment*(p: var Parser, n: ParsedNode, maybeMissEquals = false)
+# proc setBaseFlags*(n: ParsedNode, base: NumericalBase)
```
saem, line 93:
```suggestion
```

should be good to remove now that this is finalizing.

compiler/ast/reports.nim
```diff
@@ -86,10 +86,12 @@ type
 
     reportInst*: ReportLineInfo ## Information about instantiation location
     ## of the reports - present for all reports in order to track their
-    ## origins.
+    ## origins throught the compiler.
```
saem, line 89:
```suggestion
    ## origin within the compiler.
```

compiler/ast/reports.nim
```diff
@@ -86,10 +86,12 @@ type
...
 
     reportFrom*: ReportLineInfo ## Information about submit location of the
-    ## report
+    ## report. Contains information about the place where report was
+    ## /submitted/ to the system - sometimes report might be created,
+    ## modified to add new information and only them put into the pipeline.
```
saem, line 94:
```suggestion
    ## /submitted/ to the system - sometimes a report is created, modified
    ## to add new information, and only then put into the pipeline.
```

compiler/sem/lookups.nim
```diff
@@ -739,6 +742,7 @@ proc qualifiedLookUp*(c: PContext, n: PNode, flags: set[TLookupFlag]): PSym =
       result = errorExpectedIdentifier(c, ident, wrongNode, errExprCtx)
     elif checkModule in flags:
       result = searchInScopes(c, ident, amb).skipAlias(n, c.config)
+      # debug(result, verboseTReprConf)
```
saem, line 744:
```suggestion
```

compiler/sem/lookups.nim
```diff
@@ -29,7 +29,8 @@ import
     modulegraphs
   ],
   compiler/utils/[
-    debugutils
+    debugutils,
+    astrepr
```
saem, line 33:
```suggestion
    debugutils
```

compiler/sem/passes.nim
```diff
@@ -27,6 +27,7 @@ import
   ],
   compiler/utils/[
     pathutils,
+    astrepr
```
saem, line 30:
```suggestion
    pathutils
```

compiler/sem/passes.nim
```diff
@@ -84,13 +85,26 @@ proc closePasses(graph: ModuleGraph; a: var TPassContextArray) =
...
-proc processTopLevelStmt(graph: ModuleGraph, n: PNode, a: var TPassContextArray): bool =
-  # this implements the code transformation pipeline
-  var m = n
-  for i in 0..<graph.passes.len:
+proc processTopLevelStmt(
+    graph: ModuleGraph,
+    toProcess: PNode,
+    a: var TPassContextArray): bool =
+  ## Main processing pipeline entry point - accepts a toplevel node,
+  ## collection of pass contexts and main module graph which defines passes
```
saem, line 93:
```suggestion
  ## collection of pass contexts, and a main module graph which defines passes
```

compiler/sem/passes.nim
```diff
@@ -144,30 +164,42 @@ proc partOfStdlib(x: PSym): bool =
...
   prepareConfigNotes(graph, module)
-  openPasses(graph, a, module, idgen)
-  if stream == nil:
+  openPasses(graph, passesArray, module, idgen)
+  if defaultStream.isNil():
     let filename = toFullPathConsiderDirty(graph.config, fileIdx)
-    s = llStreamOpen(filename, fmRead)
-    if s == nil:
+    stream = llStreamOpen(filename, fmRead)
+    if stream == nil:
```
saem, line 188:
```suggestion
    if stream.isNil:
```

compiler/sem/passes.nim
```diff
@@ -176,39 +208,99 @@ proc processModule*(graph: ModuleGraph; module: PSym; idgen: IdGenerator;
...
         # read everything until the next proc declaration etc.
-        var sl = newNodeI(nkStmtList, n.info)
-        sl.add n
+        var toplevelStatements = newNodeI(nkStmtList, firstStatement.info)
+        toplevelStatements.add firstStatement
+
+        # Tail data that is would have already been parsed by the time we
+        # collect all the necessary statements - it should be processed as
+        # a separate trailing action.
         var rest: PNode = nil
```
saem, line 266:
```suggestion
        var rest: PNode = nil
          ## we don't know if this collection is finished unless we parse
          ## the next node, but if that node is the end of collection we need
          ## to keep it here so we can add it to the next round of parsing.
```

compiler/sem/passes.nim
```diff
@@ -176,39 +208,99 @@ proc processModule*(graph: ModuleGraph; module: PSym; idgen: IdGenerator;
...
+      if firstStatement.kind in imperativeCode:
         # read everything until the next proc declaration etc.
-        var sl = newNodeI(nkStmtList, n.info)
-        sl.add n
+        var toplevelStatements = newNodeI(nkStmtList, firstStatement.info)
+        toplevelStatements.add firstStatement
+
+        # Tail data that is would have already been parsed by the time we
+        # collect all the necessary statements - it should be processed as
+        # a separate trailing action.
```
saem, line 265:
```suggestion
```

I don't think this is accurate, `rest` stores everything that's not part of this collection but was parsed because we need to parse and then decide whether to continue.

I added a suggestion which documents the rest var itself, see what you think below.

compiler/sem/pragmas.nim
```diff
@@ -38,6 +38,7 @@ import
     ropes,
     pathutils,
     debugUtils,
+    astrepr
```
saem, line 41:
```suggestion
    debugUtils
```

saem (commented):
Did a pass on all the comments.

saem:
Merging now, @haxscramper have a look at the unresolved comment related suggestions.

It might be easiest to simply use the same branch, apply the suggestions, and then do an rebase onto devel.

compiler/sem/passes.nim
haxscramper: original code was


```nim
        var rest: PNode = nil
        while true:
          var n = parseTopLevelStmt(p)
          if n.kind == nkEmpty or n.kind notin imperativeCode:
            rest = n
            break
          sl.add n
        #echo "-----\n", sl
        if not processTopLevelStmt(graph, sl, a): break
        if rest != nil:
          #echo "-----\n", rest
          if not processTopLevelStmt(graph, rest, a): break
```

We don't add it to the next round of parsing in any way, and instead add another process call at the end, "separate trailing action."

PR #426, opened 2022-09-03
cpp: fix parameters named `this` causing compilation failure

zerbina:
## Summary
The `wThis` enum item is restored and the tests that were previously
disabled because of it's removal are enabled again.

## Details
Identifiers that are keywords in the target language need to be escaped
in the generated code. For the C-like targets, the `ccgtypes.isKeyword`
procedure is responsible for testing if an identifier is a keyword  - it
compares the `PIdent.id` against a slice of IDs specified in `wordrecg`
(in this case, `ccgKeywordsLow..ccgKeywordsHigh`).

With the removal of the `wThis` enum item, which - due to it's enum
position - was part of the aforementioned slice and supplied the ID and
keyword name, `this` is no longer recognized as a keyword and is thus
not escaped in the generated C/C++ code anymore.


saem (approved):
Bors r+

Thank you.

PR #427, opened 2022-09-04
internals: semexprs.semConv now emits nkError

saem:
## Summary

Refactored `semConv` to return nkError in all error cases. A further
refactoring to reowrk into a case style is still required.


saem:
Bors r+

PR #431, opened 2022-09-04
improve naming of `sigmatch.paramTypesMatch` vars

haxscramper:
No functional changes - improve naming of the `paramTypesMatch`
procedure and better document its implementation. Replace multiple `==
nil` usages with `.isNil()` checks.

### DIFF
--- a/compiler/sem/sigmatch.nim
+++ b/compiler/sem/sigmatch.nim
@@ -321,7 +321,7 @@ proc concreteType(c: TCandidate, t: PType; f: PType = nil): PType =
     result = t
     while true:
       result = PType(idTableGet(c.bindings, t))
-      if result == nil:
+      if result.isNil():
         break # it's ok, no match
         # example code that triggers it:
         # proc sort[T](cmp: proc(a, b: T): int = cmp)
@@ -411,7 +411,7 @@ proc genericParamPut(c: var TCandidate; last, fGenericOrigin: PType) =
      last.len-1 == fGenericOrigin.len:
     for i in 1..<fGenericOrigin.len:
       let x = PType(idTableGet(c.bindings, fGenericOrigin[i]))
-      if x == nil:
+      if x.isNil():
         put(c, fGenericOrigin[i], last[i])
 
 proc isObjectSubtype(c: var TCandidate; a, f, fGenericOrigin: PType): int =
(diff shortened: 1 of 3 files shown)


compiler/sem/sigmatch.nim
```diff
@@ -2364,68 +2364,116 @@ proc paramTypesMatchAux(m: var TCandidate, f, a: PType,
...
       # ambiguous: more than one symbol fits!
       # See tsymchoice_for_expr as an example. 'f.kind == tyUntyped' should match
       # anyway:
-      if f.kind in {tyUntyped, tyTyped}: result = arg
-      else: result = nil
+      if formal.kind in {tyUntyped, tyTyped}:
+        result = arg
+
+      else:
+        result = nil
```
```suggestion
      result =
        if formal.kind in {tyUntyped, tyTyped}:
          arg
        else:
           nil
```

Spacing these small bits out mechanical hurts readability. Broad spaces intimate a significant logical break, which this is not, we're doing one logical thing and that's setting `result`.

compiler/sem/sigmatch.nim
```diff
@@ -2440,7 +2488,7 @@ proc setSon(father: PNode, at: int, son: PNode) =
 proc prepareOperand(c: PContext; formal: PType; a: PNode): PNode =
   if formal.kind == tyUntyped and formal.len != 1:
     # {tyTypeDesc, tyUntyped, tyTyped, tyProxy}:
-    # a.typ == nil is valid
+    # a.typ.isNil() is valid
```
```suggestion
```

if we're going to change it simply remove it instead.

saem (changes requested):
Couple minor changes, I'll add terms later.

compiler/sem/sigmatch.nim
haxscramper: What do you think about more active usage of the `tern` template for one-line conditionals like this?

`result = tern(formal.kind in {tyUntyped, tyTyped}, arg, nil)`

saem: I'd prefer it if it was `ternary` in full or just looked more like the ternary operator that's broadly available in other languages. The latter is unlikely.

With compact code space sensitivity really sucks because space is really inefficient (by the volume of screen real estate required) at expressing logical subdivisions.

This problem could be somewhat alleviated when the parser becomes grammar driven and we can generate code from the grammar including syntax highlighting. Then perhaps we can use that to accentuate the separation.

PR #432, opened 2022-09-05
internal: case/nkError style for template sem

saem:
## Summary
* rework templates to take nkError into account

## Details

Now using `nkError` in `semtempl` module almost exclusively. `evaltempl`
still has a number of `localReport` rather than error propagation.

This made a fair bit of code more linear, with fall through logic for
cases where `nkError` is encountered. Semantic analysis should be more
complete in error cases or partially working code such as during
suggest.

Added a small test for term rewriting macros, a fuller test suite is
still required for covering the feature as specified. This will also
help find any design flaws.

Some minor changes were made to `hlo` (high level optimizations) to
allow `nkError` nodes being returned.

---

## Notes for Reviewers
* code is gross, suggestions welcome
* likely plenty can be removed using nkError more effectively

### DIFF
--- a/compiler/ast/trees.nim
+++ b/compiler/ast/trees.nim
@@ -160,10 +160,16 @@ proc effectSpec*(n: PNode, effectType: TSpecialWord): PNode =
       return
 
 proc unnestStmts(n, result: PNode) =
-  if n.kind == nkStmtList:
-    for x in items(n): unnestStmts(x, result)
-  elif n.kind notin {nkCommentStmt, nkNilLit}:
-    result.add(n)
+  case n.kind
+  of nkStmtList:
+    # xxx: we don't handle nkStmtListExpr, hmm
+    for x in items(n):
+      unnestStmts(x, result)
+  of nkCommentStmt, nkNilLit:
+    # QUESTION: why don't we drop empties? what about defer?
+    discard
+  else:
+    result.add n
 
 proc flattenStmts*(n: PNode): PNode =
   result = newNodeI(nkStmtList, n.info)
--- a/compiler/sem/evaltempl.nim
+++ b/compiler/sem/evaltempl.nim
@@ -45,9 +45,10 @@ proc copyNode(ctx: TemplCtx, a, b: PNode): PNode =
   if ctx.instLines: result.info = b.info
 
 proc evalTemplateAux(templ, actual: PNode, c: var TemplCtx, result: PNode) =
-  template handleParam(param) =
+  template handleParam(param: PNode) =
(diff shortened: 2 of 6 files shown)


compiler/ast/trees.nim
```diff
@@ -160,10 +160,16 @@ proc effectSpec*(n: PNode, effectType: TSpecialWord): PNode =
...
-    for x in items(n): unnestStmts(x, result)
-  elif n.kind notin {nkCommentStmt, nkNilLit}:
-    result.add(n)
+  case n.kind
+  of nkStmtList:
+    # xxx: we don't handle nkStmtListExpr, hmm
+    for x in items(n):
+      unnestStmts(x, result)
+  of nkCommentStmt, nkNilLit:
+    # xxx: why don't we drop empties? what about defer?
```
haxscramper: Can we start marking questions with the `QUESTION` tag instead of `xxx:`?

compiler/sem/evaltempl.nim
```diff
@@ -47,7 +47,8 @@ proc copyNode(ctx: TemplCtx, a, b: PNode): PNode =
 proc evalTemplateAux(templ, actual: PNode, c: var TemplCtx, result: PNode) =
   template handleParam(param) =
```
haxscramper: `param` should be a `param: PNode`, right?

compiler/sem/evaltempl.nim
```diff
@@ -148,7 +154,13 @@ proc evalTemplateArgs(n: PNode, s: PSym; conf: ConfigRef; fromHlo: bool): PNode
 
   # add any generic parameters
   for i in 1..genericParams:
-    result.add n[givenRegularParams + i]
+    let it = n[givenRegularParams + i]
+    
+    # xxx: propagate nkError
```
haxscramper, line 159: `xxx:` here means `TODO`, correct? Same with `QUESTION` tag - it is not really obvious what the author tried to convey.

haxscramper (changes requested):
High-level comment: for preexisting procedures, would it be possible to add documentation comments on their purpose - including non-exported ones.

compiler/sem/evaltempl.nim
saem: TODO I tend to use for things to do before the PR is done.

saem: I tend to use `TODO` for things in the PR and `xxx` for stuff outside the PR.

saem:
Added doc comments, going to wait for CI to clear.

PR #435, opened 2022-09-06
perf: make reporting faster

saem:
## Summary
* reports now clear up memory-ish

## Details
* uhhh, check the code for now

---

## Notes for Reviewers
* this is probably a silly idea
* if it works it's not silly though
* this is mostly a sketch, need to clean-up with feedback

### DIFF
--- a/compiler/ast/reports.nim
+++ b/compiler/ast/reports.nim
@@ -712,6 +712,7 @@ type
 
   Report* = object
     ## Toplevel wrapper type for the compiler report
+    reportId: ReportId
     case category*: ReportCategory
       of repLexer:
         lexReport*: LexerReport
@@ -759,9 +760,16 @@ static:
     echo "parser reports   = ", len(repParserKinds)
     echo "internal reports = ", len(repInternalKinds)
 
-let reportEmpty* = Report(
+const
+  emptyReportId* = ReportId(0)
+    ## report id for the `emptyReport`
+
+let emptyReport* = Report(
+  reportId: emptyReportId,
   category: repInternal,
   internalReport: InternalReport(kind: repNone))
+  ## used either as a sentinel report value it's usually an error if attempting
+  ## to output an emptyReport.
 
 template eachCategory*(report: Report, field: untyped): untyped =
   case report.category:
@@ -931,17 +939,33 @@ type
 func incl*(s: var ReportSet, id: ReportId) = s.ids.incl uint32(id)
 func contains*(s: var ReportSet, id: ReportId): bool = s.ids.contains uint32(id)
 
-func addReport*(list: var ReportList, report: sink Report): ReportId =
+func isEmptyReport*(report: Report): bool {.inline.} =
+  report.category == repInternal and report.internalReport.kind == repNone
+
+proc addReport*(list: var ReportList, report: sink Report): ReportId =
   ## Add report to the report list
   list.list.add report
   result = ReportId(uint32(list.list.high) + 1)
+
+  # xxx: perhaps make an empty category instead?
+  if not report.isEmptyReport:
+    report.reportId = result
 
 func addReport*[R: ReportTypes](list: var ReportList, report: R): ReportId =
   addReport(list, wrap(report))
 
 func getReport*(list: ReportList, id: ReportId): Report =
-  ## Get report from the report list using it's id
+  ## Get report from the report list using its id
   list.list[int(uint32(id)) - 1]
+
+proc freeReport*(list: var ReportList, report: Report) =
+  ## In non-debug builds replaces the report in the report list with an
+  ## `emptyReport`, should improve performance
+  when defined(debug):
+    discard
+  else:
+    if report.reportId.uint32 > 0:
+      list.list[int(uint32(report.reportId)) - 1] = emptyReport
 
 
 func actualType*(r: SemReport): PType = r.typeMismatch[0].actualType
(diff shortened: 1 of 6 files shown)


compiler/ast/reports.nim
```diff
@@ -712,6 +712,7 @@ type
 
   Report* = object
     ## Toplevel wrapper type for the compiler report
+    reportId: ReportId
```
haxscramper, line 715: Report should not store its ID directly, instead the ID to free should be passed to the cleanup logic.

compiler/ast/reports.nim
```diff
@@ -759,9 +760,16 @@ static:
     echo "parser reports   = ", len(repParserKinds)
     echo "internal reports = ", len(repInternalKinds)
 
-let reportEmpty* = Report(
+const
+  emptyReportId* = ReportId(0)
+    ## report id for the `emptyReport`
+
+let emptyReport* = Report(
```
haxscramper, line 767: If report ID is used directly for the free operation, explicit empty report won't be necessary

compiler/front/cli_reporter.nim
```diff
@@ -3661,9 +3661,11 @@ proc reportHook*(conf: ConfigRef, r: Report): TErrorHandling =
 
   elif r.kind in rdbgTracerKinds and conf.isDefined(traceDir):
     rotatedTrace(conf, r)
+    conf.m.reports.freeReport(r)
```
haxscramper, line 3664: Report freeing should never happen in the report writing, it should happen in the `options.nim` *after* the report has been considered for writing. A good place to consider putting report freeing is 

https://github.com/nim-works/nimskull/blob/309a57c88b332307b5b40b85194dec2feb6a3d13/compiler/front/options.nim#L516

At this location we still have ID information, therefore all changes with `emptyReport` and `reportId` built into the `Report` object will no longer be necessary.

compiler/front/options.nim
```diff
@@ -508,12 +508,12 @@ proc report*(conf: ConfigRef, id: ReportId): TErrorHandling =
...
   ##           IDs, because that's the only way report can (supposedly)
   ##           enter the error message system twice.
-  return conf.report(conf.m.reports.getReport(id))
+  result = conf.report(conf.m.reports.getReport(id))
 
 proc report*(conf: ConfigRef, node: PNode): TErrorHandling =
   ## Write out report from the nkError node
   assert node.kind == nkError
-  return conf.report(node.reportId)
+  result = conf.report(node.reportId)
```
haxscramper, line 516: Instead, report freeing should be placed here, as we have the ID information and can access the report store directly. Requiring the *formatting* function to *cleanup* the report is not ideal.

haxscramper (changes requested):
This can be implemented without storing *id of the report inside the report*, otherwise the approach seems sensible to me as we don't need to keep the data around, and it is clearly providing an unnecessary load on the GC.

haxscramper:
For performance measurements, we can adopt @ringabout's solution https://github.com/nim-lang/Nim/commit/cf1c14936670ef0c175149f6fba4dffc1cf4ba43 which generates PR messages like these https://github.com/nim-lang/Nim/pull/20170#issuecomment-1207436763 Alternatively (and it might even be a much better solution), we can track time for each `koch` step - boot, tests, each tool and so on. This will provide a lot more data points that can be aggregated to track the trends. 

Maybe a bit more structured solution - parsing GitHub comments is very easy, so I can build a historical tracking stats as well.

haxscramper:
https://fast.vlang.io/

https://github.com/vlang/v/blob/master/cmd/tools/fast/fast.v

haxscramper:
Strictly speaking, there is no requirement to store report IDs in a continuously growing sequence and later replace written ones with empty data. We can use a monotonously increasing ID generator and `OrderedTable` from which we delete reports as we write them. Reports are never really needed after they are written, so this is largely an unnecessary memory 

We hardly ever traverse the report sequence directly and if we need to keep all reports in one place for some later statistics we can achieve this by writing a hook that collects them.

compiler/front/cli_reporter.nim
saem: I think the issue is that `handleReport` simply takes the PNode, gets the reportId, looks up the report and passes it on -- we never call that proc. Changing this seemed to be rather tricky.

https://github.com/nim-works/nimskull/blob/747e060e5f6dc16bf3ba9ac5608341e989952dbe/compiler/front/msgs.nim?rgh-link-date=2022-09-06T16%3A21%3A21Z#L504-L512

haxscramper: Reworking it might be less trivial due to some unnecessary interactions, but I would prefer to streamline the reporting pipeline to ensure there is a single point where `id -> report` conversion happens rather than to put cleanup operations on the wrong place, handling earlier design errors that I made. 

Although the code you provided can also contain a cleanup action, replacing `conf.m.writtenSemReports` with a more general `OrderedTable`-based solution that would remove all reports that had already been written, ensuring both memory reclamation and uniqueness of the write action for each report.

saem:
This is far more complicated than it seems, we effecitively cannot tell when to free these reports. We need to know if we've logically passed them by or not and that's only possible if we've in fact output them, while not conflating that with say a nested compliation.

saem:
The error storage needs to be tied to the lifetimes of the AST to which they relate -- it gets complicated if we emit an `nkError` that should then 'poison' additional downstream operations. Perhaps we need to add an error pass which looks at all reported errors, checks for nesting and other bits, and then cleans up as necessary.

Perhaps this clean-up can happen near `sempass2.track`? 🤔 Most likely this should happen in `trackProc` and `trackStmt` which are top level calls.

haxscramper:
> Perhaps this clean-up can happen near `sempass2.track`

Something like `informAboutCertainlyUnnecessaryReports(conf: ConfigRef, reports: seq[ReportId])`

we can still free the reports that we know for sure we wrote out, which happens to be a large number - everything that is explicitly goes through `ID -> Report -> structuredErrorHook`

saem:
This has been superceded by adding in AST diagnostics.

PR #436, opened 2022-09-08
preemptively remove nim"pretty" before refactor

haxscramper:
Remove  nim"pretty" hacks  from the  lexer  and parser  in advance,  before further  lexer
refactoring  takes   place  per  @zerbina's  suggestion   (comment)  in  the  WIP   PR  at
https://github.com/nim-works/nimskull/pull/433.   nim"pretty"   removal  has   also   been
discussed  in https://github.com/nim-works/nimskull/discussions/113.  This tool  should be
re-implemented using standalone  binary that reads in the concrete  syntax tree that *does
not loose positioning  information* instead of the current implementation  which if rather
useless - does not guarantee uniform layout, barely configurable.

This PR:

- Removes nim"pretty" binary and  associated tests - they would have  to be rewritten from scratch anyway, in order to be able to guarantee a consistent formatting.
- Removes all `when  defined(nimpretty)` hacks from the compiler source  - `lexer.nim` and `parser.nim` were  infested the  most, but there  were a couple  of stains  in different files as well.


compiler/ast/parser.nim
```diff
@@ -104,9 +99,7 @@ proc simpleExprAux(p: var Parser, limit: int, mode: PrimaryMode): ParsedNode
 # implementation
 
 template prettySection(body) =
```
zerbina: Since the template is now redundant, could you remove it and it's usages? Feel free to disregard this comment if you're planning to re-use `prettySection` later on.

zerbina (commented):
Very nice - thank you for splitting the change off.

If the suggestion doesn't apply, feel free to merge the PR already.

compiler/ast/parser.nim
haxscramper: Fixed, thanks for pointing this out. Now merging

PR #437, opened 2022-09-13
vm: make the `nimVMDebugGenerate` define work again

zerbina:
## Summary
Fix the compilation error in `vmjit` that happened when
`debugEchoCode` was enabled.

In addition, refactor, clean-up, and document the `codeListing`
procedure and move it to the new module `vmutils`. `codeListing` also
doesn't directly report the generated listing anymore - that's the
responsibility of it's caller now.

## Details
* the `debugEchoCode` constant is removed and replaced with
 direct usage of `defined(nimVMDebugGenerate)`

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -71,6 +71,9 @@ import
     results
   ]
 
+when defined(nimVMDebugGenerate):
+  import compiler/vm/vmutils
+
 # XXX: instead of `assert` a special `vmAssert` could be used for internal
 #      checks. This would allow for also reporting things like the current
 #      executed instruction
@@ -86,8 +89,6 @@ const
 
 const
   errIllegalConvFromXtoY = "illegal conversion from '$1' to '$2'"
-
-
 
 proc createStackTrace(
     c:          TCtx,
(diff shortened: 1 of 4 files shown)


compiler/vm/vm.nim
```diff
@@ -87,7 +87,8 @@ const
 const
   errIllegalConvFromXtoY = "illegal conversion from '$1' to '$2'"
 
-
+when debugEchoCode:
+  import compiler/vm/vmutils
```
haxscramper: Ideally, conditional imports should be placed consistently with the rest of imports - that is, at the start of the file, not before the first usage.

compiler/vm/vmjit.nim
```diff
@@ -38,6 +38,10 @@ export VmGenResult
 const
   debugEchoCode* = defined(nimVMDebugGenerate)
 
+when debugEchoCode:
```
haxscramper: Preferably should be moved to the top of the file as well.

haxscramper (approved):
`debugEchoCode` define is currently missing from the documentation, so if you can, please do add it (at least in the "Debugging Defines - Quick Reference" section) in `debug.rst`. Aside from minor comments on the import placement I don't have anything to say, and since this PR does not seem to change the behavior in any way I think it can be merged.

compiler/vm/vmjit.nim
zerbina: That's only possible if the usage of `debugEchoCode` here is replaced with `defined(nimVMDebugGenerate)`, which would mean that the imports are guarded by a different expression than the dependents of the import, rendering `debugEchoCode` somewhat useless.

I can move the `debugEchoCode` constant to `vmdef` instead.

zerbina:
> `debugEchoCode` define is currently missing from the documentation, so if you can, please do add it (at least in the "Debugging Defines - Quick Reference" section) in `debug.rst`.

Hm, but what for? It's not something that can be separately toggled, as it's just an alias for `defined(nimVMDebugGenerate)`.

compiler/vm/vmjit.nim
haxscramper: Just remove `debugEchoCode` altogether and use defines as the rest of debugging utils does

zerbina:
As requested, I removed `debugEchoCode` and replaced it's usage with `defined(nimVMDebugGenerate)`.

bors r+

PR #438, opened 2022-09-13
Split rsemVm into a separate report object type

haxscramper:
Long overdue refactoring, no intentional functionality changes, only cleanup for the reporting system implementation. Add the `VMReport` report object, new "VM" category and split mixed semantic analysis report kinds into more manageable chunks.

### DIFF
--- a/compiler/ast/report_enums.nim
+++ b/compiler/ast/report_enums.nim
@@ -17,6 +17,8 @@ type
     repCmd = "Cmd" ## Report related to execution of the external command -
     ## start of the command, execution failure, succes and so on.
 
+    repVM = "VM" ## Report related to embedded virtual machine
+
     repDebug = "Debug" ## Side channel for the compiler debug report. Sem
     ## expansion traces and other helper messages designed specifically to
     ## aid development of the compiler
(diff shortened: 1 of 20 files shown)


compiler/ast/report_enums.nim
```diff
@@ -1035,13 +1056,52 @@ const
...
+      repTraceKinds +
+      repErrorKinds +
+      repDataPassKinds +
+      repFatalKinds +
+      { repNone }
+    )
+
+    assert(
+      len(diff) == 0,
+      "Severity grouping is missins some elements: " & $diff
```
saem:
```suggestion
      "Severity grouping is missing some elements: " & $diff
```

saem (approved):
one typo, otherwise it's good to go.

compiler/front/options.nim
```diff
@@ -665,7 +665,10 @@ func isEnabled*(conf: ConfigRef, report: ReportKind): bool =
           result = true
 
         of repTraceKinds:
-          result = true
+          # Semantic trace kinds are enabled by default and probably should
+          # not be changed - but these reports might (in theory) be
+          # modified at runtime.
+          result = report in conf.notes
```
zerbina, line 671: Could you list this (small functional) change in the commit message?

zerbina (commented):
Looks good to me. Thanks!

---

There are some pure style changes in `cli_reporter` that are unrelated to this PR (indentation of `case` statements, line breaks for strings, etc.). Would you be okay with moving those to a separate PR? It'd make traversing the history a bit easier.

I don't consider this to be a blocker however - feel free to disregard my request.

compiler/front/options.nim
haxscramper: Sorry, I just noticed your review as well (together with a notification about PR merge), but this is not a functional change, just part of the reporting enum sets was tangled together in a weird manner and there were two bugs cancelling each other out ((1)some reports were never added to the allowed `notes` and (2) for trace reports writing was always allowed).

haxscramper: Added comment just documents the logic behind the check should have been added there several months ago

zerbina: Thank you for clarifying.

PR #440, opened 2022-09-19
Internals: nkError branch in TNode 

saem:
## Summary
`nkError` has a dedicate branch in `TNode`

## Details
- `reportId` is now in the branch
- nkError has a `kids` node sequences instead of `sons`

The separation of the sequence requires some direct access to `kids`.
This surfaced compiler bugs where traversals went into error nodes.

Next we can replace `reportId` the report and remove gc/memory pressure.

---

## Notes for Reviewers
* getting a weird VM Runner error, not sure how it's happening, tips?
* will rebase after CI is passing
* will handle `reportId` change separately

PR #441, opened 2022-09-19
vm: properly handle `quit` in const/standalone mode

zerbina:
## Summary
The previous behaviour of `opcQuit` was to abort the execution loop
without any clean-up (and without reporting the user-provided
exit-code). Since this looked like a normal exit to the caller of
`rawExecute`, the VM was left in an invalid state, causing an internal
compiler error when `quit` was called in the context of a `const`
evaluation.

When being run in `emConst` or `emStandalone` mode, `opcQuit` now
performs an abnormal exit by raising an error that also stores the
exit-code.

---

@saem reported the issue [here](https://github.com/nim-works/nimskull/pull/440#issuecomment-1250479840) .

### DIFF
--- a/compiler/ast/report_enums.nim
+++ b/compiler/ast/report_enums.nim
@@ -289,6 +289,7 @@ type
     rvmNoType
     rvmNotAField
     rvmUnsupportedNonNil
+    rvmQuit
 
     rvmTooManyIterations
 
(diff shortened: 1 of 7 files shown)


saem:
bors r+

yay! 🎉 

thank you

PR #442, opened 2022-09-24
arc: fix code-gen error when copying pure sub-objects

zerbina:
## Summary
To decide whether a type-field copy is needed, `liftdestructors` used `lackMTypeField` which doesn't consider base types, thus erroneously generating a type-field copy for sub-object types not annotated as `.pure` where the root is `.pure`. This then caused an error at the C compilation stage.

`liftdestructors` now uses the same procedure (`isObjLackingTypeField`) as is used by `cgen` to detect the presence of a type header.

## Details
* move `isObjLackingTypeField` from `ccgtypes` to `types` and document it


saem (approved):
Bors r+

Issue #443, opened 2022-09-25
refactor: compiler output/reports modules design and faster compile time


saem:
## Context

Structured error reporting has helped identify and resolve a number of issues in the compiler, in part by bringing together all the data types required for various report/output generated by the compiler. In doing so a significant issue was revealed where the AST does not own its error data impacting many phases of compilation primarily semantic analysis (sem).

### Current Challenges

The current design/implementation results in:
* the compiler holding a lot of memory during it's execution (imagine suggest!)
* from a design perspective doesn't compose well precluding features like:
  * additional diagnostics and hints based on error inspections
  * error recovery within the compiler
  * compile time error recovery/dispatch `case compile(someExprOrStmt).resultKind ... # pretend the api is nicer`
* semantic analysis can also run in a loop (suggest) and broken untouched code should keep its errors and be repeatedly provided during diagnostics

### Relevant modules/links for context:
* data types for reports are presently here: `compiler/ast/reports` and `compiler/ast/report_enums`
* ast data types are, including `TNode` with the `nkError` branch is in `compiler/ast/ast_types`
* initial PR to bring in structured reports https://github.com/nim-works/nimskull/pull/94

## Problem

We're already on our way to have all errors generated by semantic analysis to be held within `nkError` nodes. These nodes dictate the lifetime of all the error data associated with them. A node can be generated and used and reused in a variety of contexts, one of which is error reporting and whether we report it or not doesn't indicate whether we can reclaim the memory.

```nim
# compiler/ast/ast_types.nim -- abridged
type
  TNode* = object
  case kind*: TNodeKind
  # ...
  of nkError:
    reportId*: ReportId
    kids*: seq[PNode]   # wrapped ast + extra info, see `compiler/ast/errorhandling`
  # ...

# compiler/ast/reports.nim -- abridged
type
  ReportBase* = object of RootObj
    context*: seq[ReportContext] # sem specific contextual info
    location*: Option[TLineInfo]    # location info
    reportInst*: ReportLineInfo     # where in the compiler source it was created
    reportFrom*: ReportLineInfo   # where it was submitted for reporting

  SemReport* = object of ReportBase
    ast*: PNode
    typ*: PType
    sym*: PSym
    str*: string
    spellingCandidates*: seq[SemSpellCandidate]
    case kind*: ReportKind
    # ... branches for various kinds: errors, hints, & warnings

# compiler/ast/lineinfos.nim -- abridged
type
[...]

PR #446, opened 2022-10-09
Internals: Lexer and Nimconf Output and Reporting

saem:
## Summary

Lexing and Nim Config File Handling no longer depend upon the legacy
report infrastructure. This is part of a large effort to fix compiler
diagnostics and output, see this issue for more information:
https://github.com/nim-works/nimskull/issues/443

## Details

Before `lexer` and `nimconf` modules' "reports" were comingled as lexer
reports in the `report_enums` modules under the `rlex` prefix.
Additionally, the `lexer` and `nimconf` module had odd cylic
dependencies with `reports` and `msgs` modules. Lastly, the `reports`
module owned the error/event/output data types when those should be
under `lexer` and `nimconf`.

Now, `lexer` and `nimconf` define "diagnostic" data types that the
caller deals with including integrating output handling. Now they no
longer depend upon the `reports` module.

Changes to Lexer:
- lexer only throws on fatal/internal error
- all other cases (errors, warnings, hints) are diagnostics (LexerDiags)
- the lexer stores all diags in a field for its lifetime
- diags are traversed via an interator `errorsHintsAndWarnings`
- `diagOffset` is used to demarcate "when" the last diags came from

Changes to Nim Config handler:
- similar to Lexer above, wrt diagnostics
- legacy reporting bridge for `nimconf` is in `cmdlinehelper`

Simplifications, Code Clean-up, and Responsibility Changes:
- `options` module:
  - moved `MsgConfig` here was misplaced in `lineinfos`
  - moved helper procs for file path and other handling bits specific to
    options and used by other modules from `msgs`
- `msgs` hanldes `lexer` diagnostic to report mapping as it is the
  compiler messages framework code
- `lineinfos` now contains the `TLineInfo` type as it should, was not
  possible before because `MsgConfig` was misplaced
- introduced `numericbase` to break a cycle between `ast_types` and
  `lexer` modules
- `reports` module cleanup:
  - `ReportBase` no longer contains `ReportContext`, which is a sem/vm
    only issue; longer term all this data should never be here in the
    first place
  - minor: removed some unnecessary new lines and cleaned up formatting
- `cmdlinehelper` handles nimconf diagnostic to report mapping, then
  callers like `main` module can use that

---

## Notes for Reviewers

* this is going to be a big change, that ship has sailed
* the code already feels more modular

### Strategy
* I'm going to create a number of clean-ish commits in this branch
* they might go in one at a time, or get squashed
* rough order of operations:
  * extract all lexer related diagnostics out
  * make the above work:
    * including sorting declarations into better modules
    * break apart cyclic dependencies
    * don't introduce too many bridges
  * repeat for all other report types
* if the above yields clean chunks then they'll merge in one at a time, otherwise they'll get squashed

### Questions

Cycle breaking without introducing function pointers is a useful trick. The problem is
reporting/error handling are tied tightly and it's hard to see how to tease them apart.
Thoughts? Good place to see the issue is the `handleDiag` stuff in `lexer`. They can't
simply fire and forget because the errorhandling strategy as configured injects itself
into the lexer's execution in cases where we abort. So how do we break this while still
using static dispatch (code and no function pointers)?

### DIFF
--- a/compiler/ast/lexer.nim
+++ b/compiler/ast/lexer.nim
@@ -22,11 +22,11 @@ import
     pathutils,
   ],
   compiler/ast/[
+    numericbase,
     wordrecg,
     nimlexbase,
     llstream,
     lineinfos,
-    reports,
     idents
   ],
   std/[
@@ -35,8 +35,7 @@ import
     strutils
   ],
   compiler/front/[
-    options,
-    msgs
+    options
   ]
 
 const
@@ -52,7 +51,7 @@ const
 
 type
   TokType* = enum
-    tkInvalid = "tkInvalid", tkEof = "[EOF]", # order is important here!
+    tkInvalid = "tkInvalid", tkError = "tkError", tkEof = "[EOF]", # order is important here!
     tkSymbol = "tkSymbol", # keywords:
     tkAddr = "addr", tkAnd = "and", tkAs = "as", tkAsm = "asm",
     tkBind = "bind", tkBlock = "block", tkBreak = "break", tkCase = "case", tkCast = "cast",
@@ -96,6 +95,69 @@ type
 
   TokTypes* = set[TokType]
 
+  LexerDiagKind* = enum
+    # internal errors begin
+    lexDiagInternalError ## lexer programming error
+    # internal errors end
+    
+    # users errors begin
+
+    # spacing
+    lexDiagMalformedUnderscores
+    lexDiagMalformedTrailingUnderscre
+    lexDiagInvalidToken
+    lexDiagNoTabs
+
+    # numbers
+    lexDiagInvalidIntegerPrefix
+    lexDiagInvalidIntegerSuffix
+    lexDiagNumberNotInRange
+    lexDiagExpectedHex
+    lexDiagInvalidIntegerLiteral
+
+    # char
+    lexDiagInvalidCharLiteral
+    lexDiagMissingClosingApostrophe
+    lexDiagInvalidUnicodeCodepoint
+    # xxx: differentiate between invalid codepoint empty vs out of range, but
+    #      still be treated within the invalidUnicodeCodepoint "family"
+
+    # string
+    lexDiagUnclosedTripleString
+    lexDiagUnclosedSingleString
+
+    # comments
+    lexDiagUnclosedComment
+
+    # user errors end
+
+    # warnings begin
+    lexDiagDeprecatedOctalPrefix = "OctalEscape"
+    # warnings end
+
+    # linting hints begin
+    lexDiagLineTooLong = "LineTooLong"
+    lexDiagNameXShouldBeY = "Name"
+    # linting hints end
+
+  LexerDiag* = object
+    ## `Diag`nostic data from the Lexer, mostly errors
+    msg*: string
+    location*: TLineInfo        ## diagnostic location
+    instLoc*: InstantiationInfo ## instantiation in lexer's source
+    case kind*: LexerDiagKind:
+    of lexDiagNameXShouldBeY:
+      wanted*: string
+      got*: string
+    else:
+      discard
+
+const
+  LexDiagsFatal*   = {lexDiagInternalError}
+  LexDiagsError*   = {lexDiagMalformedUnderscores..lexDiagUnclosedComment}
+  LexDiagsWarning* = {lexDiagDeprecatedOctalPrefix}
+  LexDiagsHint*    = {lexDiagLineTooLong..lexDiagNameXShouldBeY}
+
 when defined(nimsuggest):
   const weakTokens = {tkComma, tkSemiColon, tkColon,
                       tkParRi, tkParDotRi, tkBracketRi, tkBracketDotRi,
@@ -106,12 +168,8 @@ const
   tokKeywordLow* = succ(tkSymbol)
   tokKeywordHigh* = pred(tkIntLit)
 
-type
-  NumericalBase* = enum
-    base10,                   ## base10 is listed as the first element,
-                              ## so that it is the correct default value
-    base2, base8, base16
 
+type
   Token* = object             ## a Nim token
     tokType*: TokType         ## the type of the token
     indent*: int              ## the indentation; != -1 if the token has been
@@ -126,6 +184,13 @@ type
     literal*: string          ## the parsed (string) literal; and
                               ## documentation comments are here too
     line*, col*: int
+    error*: LexerDiag         ## error diagnostic if `tokType` is `tkError`
+
+  LexerError = object of ValueError
+    ## internal error used to signal the lexer can no longer proceed given the
+    ## input and has aborted
+    diagId: int      ## index into `Lexer.diags`, used to fetch the diag and
+                     ## insert into the error token
 
   Lexer* = object of TBaseLexer
     fileIdx*: FileIndex
@@ -138,6 +203,22 @@ type
     when defined(nimsuggest):
       previousToken: TLineInfo
     config*: ConfigRef
+    diags: seq[LexerDiag]
+
+func diagOffset*(L: Lexer): int {.inline.} =
+  ## return value represents a point in time where all existing diagnostics are
+  ## considered in the past, used in conjunction with `errorsHintsAndWarnings`
+  L.diags.len
+
+iterator errorsHintsAndWarnings*(L: Lexer, diagOffset = 0): LexerDiag =
+  ## iterate over all diagnostics (excluding fatal) from the beginning, or from
+  ## the point in time specificed via `diagOffset`
+  for d in L.diags.toOpenArray(diagOffset, L.diags.high):
+    case d.kind
+    of LexDiagsFatal:
+      continue         # fatals are reported via tkError
+    else:
+      yield d
 
 proc getLineInfo*(L: Lexer, tok: Token): TLineInfo {.inline.} =
   result = newLineInfo(L.fileIdx, tok.line, tok.col)
@@ -170,6 +251,7 @@ proc `$`*(tok: Token): string =
   of tkFloatLit..tkFloat64Lit: $tok.fNumber
   of tkInvalid, tkStrLit..tkCharLit, tkComment: tok.literal
   of tkParLe..tkColon, tkEof, tkAccent: $tok.tokType
+  of tkError: tok.error.msg
   else:
     if tok.ident != nil:
       tok.ident.s
@@ -223,9 +305,10 @@ proc openLexer*(lex: var Lexer, filename: AbsoluteFile, inputstream: PLLStream;
 proc closeLexer*(lex: var Lexer) =
   if lex.config != nil:
     inc(lex.config.linesCompiled, lex.lineNumber)
+  lex.diags.setLen(0)
   closeBaseLexer(lex)
 
-proc getLineInfo(L: Lexer): TLineInfo =
+proc getLineInfo*(L: Lexer): TLineInfo =
   result = newLineInfo(L.fileIdx, L.lineNumber, getColNumber(L, L.bufpos))
 
 proc matchTwoChars(L: Lexer, first: char, second: set[char]): bool =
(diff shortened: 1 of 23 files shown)


zerbina:
The following idea is tailored towards the lexer, but it might be possible to generalize it.

---

For the lexer, there are two kinds of reports:
* fatal error: the input is malformed and the lexer is thus unable to continue it's processing
* non-fatal: hints, warnings, etc. The lexer is able to continue it's processing

I'd propose the following:

If the lexer encounters a fatal error it aborts, non-fatal reports are accumulated. For aborting I see three approaches: exceptions, a dedicated `tkError` with manual unwinding, or a hybrid of both. Personally, I'd go with the hybrid approach. In code:

```nim

type
  Lexer = object of TLexerBase
    ...
    reports: seq[LexerReport] ## accumulates reports. The list is never cleared
                              ## or read from by lexer procedures - it is only
                              ## appended to.

  LexerError = object of ValueError
    report: int ## an index into ``Lexer.reports``

...

func fatal(L: var Lexer, report: sink LexerReport) {.noreturn, noinline.} =
  L.reports.add(report)
  raise (ref LexerError)(reportId: L.reports.high)

func reportNonFatal(L: var Lexer, report: sink LexerReport) =
  ## For hints, warnings, etc.
  L.reports.add(report)

proc matchUnderscoreChars(L: var Lexer, tok: var Token, chars: set[char]): Natural =
  ...
      if L.buf[pos+1] notin chars:
        fatal(L, LexerReport(kind: rlexMalformedUnderscores))


proc rawGetTok*(L: var Lexer, tok: var Token) =
  try:
    rawGetTokInner(L, tok) # the current ``rawGetTok`` procedure is renamed to
                           # ``rawGetTokInner``
  except LexerError as e:
    # XXX: the lexer could be reset to the state it had at the beginning of
    #      ``rawGetTok`` here
    tok = Token(kind: tkError, report: e.report)

```

The lexer is then completely agnostic to what should happen with the report - that is the responsibility of the callsite (in this case, the parser) now.

It'd be cleaner for `reports` to be a mutable parameter to all lexer related procedures that need to report something (either fatal or non-fatal), because:
* whether a procedure can report something can then be deduced from just it's signature
* the `reports` list is only used as an output - it shouldn't be part of the lexer's state

However, for ergonomic reasons (manually passing around the list can become a bit cumbersome), I think that it's okay if the list is stored as part of `Lexer`.

Handling of the reports in `parser.nim` would then look something like this:

```nim

proc handleReports(c: ConfigRef, reports: var seq[LexerReports]) =
  for r in reports.mitems:
    # `mitems` so that `r` can be moved
    # XXX: ``localReport`` is just used as an example
    localReport(c, move r)

  # the reports were all moved elsewhere, their previous storage doesn't need
  # to be kept around
  reports.setLen(0)

proc getTok(p: var Parser) =
  ...
  p.lex.rawGetTok(p.tok)
  handleReports(p.lex.config, p.lex.reports)
  # XXX: ``localReport`` already aborted (either via a `raise` or via `quit`) if an
  #      error report was encountered, so we don't need to explicitly `tkError`
  #      tokens here

```

saem:
Thank you very much, @zerbina. That was really helpful. 🎉 

I think your analysis is correct, from what I can tell only the `rintIce`/`lexDiagInternalError` are "fatal", all other errors progress can technically be made so we only need to emit a `tkError`.

Maybe I should have the Lexer take a handler map at init (with a default), which determines whats a `tkError` and what's side-band telemetry?:

```nim
type
  LexerDiagAction = enum
    ldaError, ldaAsDiag, ldaIgnore

  LexerDiagKind = enum
   lexDiagInternalError # always fatal
    lexDiagMalformedUnderscores
    ...

  LexerDiagConfig = array[LexerDiagKind, LexerDiagAction]

  LexerDiag = object
    case kind: LexerDiagKind
    ... # whatever data

  Lexer = object of TBaseLexer
    ...
    diags: seq[LexerDiag]
    diagHandlingConfig: LexerDiagConfig
```

All diags end up in `diags`, things that are supposed to be errors end up in `tkError` just like you suggested with the index from `tkError` into `diags`. Handle becomes:

```nim
func fatal(L: var Lexer, msg: string) {.noreturn, noinline.} =
  L.diags.add # create internal error, we only do "getNumber", IIRC
  raise (ref LexerError)(errId: L.diags.high)

func diag(L: var Lexer, diag: sink LexerDiag) =
  ## For non-fatal errors, hints, warnings, etc.
  L.diags.add(diag)
  case L.diagHandlingConfig[diag.kind]
  of ldaError: ...
  ...
```

zerbina:
> I think your analysis is correct, from what I can tell only the `rintIce`/`lexDiagInternalError` are "fatal", all other errors progress can technically be made so we only need to emit a `tkError`.
> 
> Maybe I should have the Lexer take a handler map at init (with a default), which determines whats a `tkError` and what's side-band telemetry?:

Since the lexer only produces one token at a time (via `rawGetTok`), I'd consider all diagnostics that are in the `rlexErrorKinds` set as fatal, as code that `localReport`s those expects (lexer) execution to abort (either via a raised exception or via `quit`).

To rephrase: in the context of the lexer, I'd not distinguish between fatal and non-fatal errors, but just abort `rawGetTok` and produce a `tkError` whenever an error (that is, the expectations the lexer has of its input are violated) is encountered. Anything that is not an error is side-band telemetry.

saem:
Sounds good thanks. I'll try that approach first.

saem:
@zerbina I pushed the latest, if you're curious.

It turns out that only truly fatal stuff (`internalError`) should raise, because the parser/lexer rely on not raising due to `check`. I got it working well enough, at least the tests seem to all pass. I'd have to do some serious surgery to the lexer/parser module to make it work with a proper `tkError` unfortunately. Might make more sense when the whole thing gets DODed.

saem:
_for the curious: my thoughts as I'm working along_

### This PR and the Near Term

Right now I'm putting all the bridging code for module specific diagnostics to `Report` in `msgs`. It can get reworked and simplfied once it's all collected. Likely over the course of a few PRs.

So far the code/structure seems a lot better, as the modules are more self-contained and the overall dependencies are decreasing -- always a good sign.

### Long Term Direction

#### handleReport and friends

There are still very significant issues because it works like the old school `localError` and friends conflating output and error handling. The latter can change due to `check` and that just make code much harder to convert. There is going to be an awkward phase still as things get refactored to not rely on the control flow effects that `handleReport` often introduces and 
procs / callers rely upon.

#### `Reports` and friends

Overall it's much easier to refactor this stuff and the real issue is `handleReport` control flow effects.

`Reports` structuring shouldn't be much more than some very generic compiler metadata and that can be figured out outside of specific modules. The rest really should lean on module data type or we convert to more generic structures based on the intention of the output.

The intended output intention should be a point of delineation, something like:

- compiler progress, diagnostics, tracing, telemetry/instrumentation are more metrics and logging
- compile time program execution output
- command output like help messages

This isn't well separated, which makes this especially problematic when it comes to flagging/configuring whether we should get output from hitns/warnings/etc... and the whole `Notes` in `ConfigRef` etc blows up massively.

compiler/front/nimconf.nim
```diff
@@ -17,25 +17,193 @@ import
...
+#      emits diagnostics and is configured as to how it should handle them as
+#      they arise wrt to aborting, etc
+from compiler/ast/report_enums import ReportKind, ReportCategory
+from compiler/ast/reports import Report, DebugReport, ExternalReport,
+                                 InternalReport, LexerReport, ParserReport,
+                                 toReportLineInfo
+from compiler/front/msgs import handleReport
+import std/options as std_options
+
+proc handleConfigEvent(
```
saem: The nimconf bit is pretty lightweight as it is today, but also not necessarily the most testable. Anyways, in order to break it up from reporting I need a way for it to signal the output events.

I'm thinking about doing something like this:

```
# same for loadConfigFile etc

proc readConfigFile(conf: ConfigRef, ...): bool =
  var
    L: Lexer
    tok: Token
    stream: PLLStream

  stream = llStreamOpen(filename, fmRead)
  if stream != nil:
   # move handleRead implementation into the options module (as a callback) then let it do the reporting config.handleRead(cekDebugReadStart, filename.string)

  # code mostly as today
```

Basically all this reporting stuff moves into options, then I chase down any dependency cycles and break it up. Thoughts?

compiler/ast/lexer.nim
```diff
@@ -189,6 +200,23 @@ type
...
+func diagOffset*(L: Lexer): int {.inline.} =
+  ## return value represents a point in time where all existing diagnostics are
+  ## considered in the past, used inconjunction with `errorsHintsAndWarnings`
+  L.diags.len
+
+iterator errorsHintsAndWarnings*(L: Lexer, diagOffset = 0): LexerDiag =
+  ## iterate over all diagnostics (excluding fatal) from the beginning, or from
+  ## the point in time specificed via `diagOffset`
+  if diagOffset < L.diags.len:
+    for d in L.diags[diagOffset..<L.diags.len]:
```
zerbina: Because this allocates a new sequence, I think it'd be better to instead use one of the following approaches:

1.
```nim
for i in diagOffset..<L.diags.len:
   let d = L.diags[i]
```
2.
```nim
for d in L.diags.toOpenArray(diagOffset, L.diags.high):
```
3. introduce and use a dedicated iterator. I did the same thing here: https://github.com/nim-works/nimskull/blob/3e4f3f29d7d2164b8f8ee3d47c3fc8ce2c41d395/compiler/sem/analyser.nim#L61-L64 Maybe the iterator could be moved into `utils/idioms`?

Using any of the above approaches would also make `diagOffset < L.diags.len` guard obsolete.

compiler/ast/lexer.nim
```diff
@@ -323,133 +353,66 @@ template eatChar(L: var Lexer, t: var Token) =
...
+template handleDiag(L: Lexer, diag: LexerDiagKind): untyped =
   doAssert diag notin {lexDiagNameXShouldBeY}
   {.line.}:
-    let d = LexerDiag(location: L.getLineInfo, instLoc: instLoc(), kind: diag)
-    L.config.handleDiagReport(d, instLoc())
+    L.diags.add LexerDiag(
+                  location: L.getLineInfo, 
+                  instLoc: instLoc(),
+                  kind: diag
+                )
```
zerbina: Instead of using a template (both here and for the other `handleDiagX` routines), you could use the following:
```nim
func handleDiag(L: var Lexer, diag: LexerDiagKind; instLoc = instLoc(-1)) =
  doAssert diag notin {lexDiagNameXShouldBeY}
  L.diags.add LexerDiag(
                location: L.getLineInfo, 
                instLoc: instLoc,
                kind: diag
              )
```

This would reduce the binary's size a bit and one no longer has to watch out for double (or multi) evaluation.

compiler/ast/parser.nim
```diff
@@ -104,7 +105,15 @@ proc getTok(p: var Parser) =
   p.lineNumberPrevious = p.lex.lineNumber
   p.lineStartPrevious = p.lex.lineStart
   p.bufposPrevious = p.lex.bufpos
+  p.lexDiagOffset = p.lex.diagOffset
```
zerbina: This seems to be the only usage of `Parser.lexDiagOffset` - could it be turned into a local instead?

compiler/front/nimconf.nim
```diff
@@ -17,25 +17,193 @@ import
...
+    of rextConf:
+      Report(
+        category: repExternal,
+        externalReport: ExternalReport(
+          reportInst: evt.instLoc.toReportLineInfo,
+          kind: kind,
+          msg: evt.msg))
+    else:
+      doAssert false, "totally borked"
+      Report() # assert ensure we don't get here
```
zerbina:
```suggestion
      unreachable("totally borked")
```

zerbina (commented):
The changes to reporting in `lexer` look good to me. Also, you're right. I forgot about `nim check` in my earlier analysis.

I still need to think a bit about the `nimconf` part.

compiler/ast/lexer.nim
saem: Thanks, will do.

I was a bit on autopilot wrt templates as the previous routines did that. Yay, reviews!

compiler/ast/parser.nim
saem: I think I'm being silly and that should be used/updated as the parser progresses the lexer across invocations.

I'll double check and that could well be a bug on my part. Good catch.

saem:
For `nimconf`

- I created a type for a callback in `nimconf`
- I'll have the two exported procs in `nimconf` that need reporting accept callback as an argument
- those procs will set a module private global var to the callback
- all module procedures etc that need handling will use the module's internal var as a target

The callback implementation is with the cli helpers/nimprog module (on the phone can't recall the name). But `main`, `nimsuggest`, etc all use the helper as well as `nimconf` IIRC. So worst case I can export and pass along the same callback implementation for reporting. Better case the API usage can be simplified and routed entirely through the CLI helper and it's cleaner.

saem:
Keeping this focused on the Lexer and NimConf, will change others in follow-up PRs.

compiler/front/options.nim
```diff
@@ -20,6 +20,10 @@ from terminal import isatty
 from times import utc, fromUnix, local, getTime, format, DateTime
 from std/private/globs import nativeToUnixPath
 
+from compiler/ast/ast_types import
+  PSym  # Contextual details of the instantnation stack optionally refer to
```
```suggestion
  PSym  # Contextual details of the instantiation stack optionally refer to
```

compiler/ast/lexer.nim
```diff
@@ -138,6 +203,22 @@ type
     when defined(nimsuggest):
       previousToken: TLineInfo
     config*: ConfigRef
+    diags: seq[LexerDiag]
+
+func diagOffset*(L: Lexer): int {.inline.} =
+  ## return value represents a point in time where all existing diagnostics are
+  ## considered in the past, used inconjunction with `errorsHintsAndWarnings`
```
zerbina:
```suggestion
  ## considered in the past, used in conjunction with `errorsHintsAndWarnings`
```

compiler/ast/reports.nim
```diff
@@ -807,19 +812,19 @@ func reportFrom*(report: Report): ReportLineInfo =
...
 
 func context*(report: Report): seq[ReportContext] =
-  eachCategory(report, context)
+  case report.category
+  of repSem:
+    result = report.semReport.context
+  of repVM:
+    result = report.vmReport.context
+  else:
+    doAssert false, "context get on category: " & $report.category
```
zerbina: Maybe use `unreachable` here?

compiler/ast/reports.nim
```diff
@@ -807,19 +812,19 @@ func reportFrom*(report: Report): ReportLineInfo =
...
-    of repSem:      report.semReport.context = context
-    of repVM:       report.vmReport.context = context
-    of repDebug:    report.debugReport.context = context
-    of repInternal: report.internalReport.context = context
-    of repBackend:  report.backendReport.context = context
-    of repExternal: report.externalReport.context = context
+  case report.category
+  of repSem: report.semReport.context = context
+  of repVM:  report.vmReport.context = context
+  else: doAssert false, "context get on category: " & $report.category
```
zerbina: Same here.

compiler/front/nimconf.nim
```diff
@@ -17,29 +17,161 @@ import
...
+                            reportFrom: InstantiationInfo,
+                            eh: TErrorHandling = doNothing): void
+
+var
+  cfgEvtHandler: NimConfEvtHandler
+    ## event handler ensure this is set when a public proc that can call the
+    ## `handle*` templates is invoked (take it in as a param). Currently, this
+    ## is only set by `loadConfig`. This allows the module to not depend upon
+    ## `reports` infrastructure which is being refactored/simplified
+  lexDiagOffset = 0
```
zerbina: As an alternative to the use of globals here, you could introduce a separate `NimConfParser` (or similar) that would look like this:
```nim
type NimConfParser = object
  lexer: Lexer
  condStack: seq[bool]
  config: ConfigRef

  cfgEvtHandler: NimConfEvtHandler
  lexDiagOffset: int
```

and then pass an instance of this object to each `parseX` procedure.

Also, it seems like `lexDiagOffset` is only ever set to `0`?

zerbina (commented):
In general, this looks good to me - I'm a bit unsure about the use of globals in `nimconf.nim`, however. I left a suggestion for how I think it could be done differently, but it's not a blocker, and can also be adjusted later.

compiler/ast/reports.nim
saem: Good call, will do.

compiler/front/nimconf.nim
saem: Yeah, I looked at this briefly, it does get a bit weird with the "output" `ConfigRef` vs the one on the lexer, IIRC. I'll give it another go.

compiler/ast/reports.nim
saem: Will do.

saem:
Thanks for the review, @zerbina.

PR #449, opened 2022-10-16
sem: overload resolution ast corruption fixes

saem:
## Summary
- signature matching no longer corrupts symbol table

## Details

core changes:
- matchesAux and friends in sigmatch no longer merge broken symbols
  - this means defensive ast copies during overload resolution
  - procs are also more nkError aware and will short circuit
- lookups no longer puts error symbols in the top level
  - this was a hack to stop "cascading errors"
  - mildly better error message for ident lookups resolving to errors
- required fixes in semLetOrVar and semConst:
  - const/let/var symbol definition tracks errors better
  - consolidated top level semLetOrVar and semConst

miscellaneous:
- due to nkError branch changes (`kids`), fixes were required in astrepr
- renamed old error proc `defaultConstructionError2`, dropped the `2`
- `semParamList` now returns a `PType` instead of unnecessarily mutating


saem:
Bors r+

PR #450, opened 2022-10-16
compiler: make `injectdestructors` a MIR pass

zerbina:
First, a short summary of how `injectdestructors` used to operate:
1. an instruction-based control-flow-graph that also contains instruction related to data-flow analysis (i.e. `use` and `def`) is generated for the code fragment (in this case, either the body of a procedure or a fragment of a module's top-level code) the transformations are to be applied to
2. the control- / data-flow-graph is traversed (executed) in order to figure out which uses of locations are "last reads" and which writes to them are "first writes". The `PNode`s corresponding to the `use` and `def` instructions for which a "last read" or "first write" is detected are marked with the respective node flags `nfLastRead` and `nfFirstWrite`.
3. the AST is processed -- this is where the core logic happens. `=destroy` calls and temporaries are injected; assignments are rewritten into either `=copy` calls, `=sink` calls, or shallow copies (i.e. `nkFastAsgn`); copies are introduced for arguments to `sink` parameters where necessary; `wasMoved` calls are inserted when an lvalue is consumed; scopes are wrapped in `try`/`finally` statements/expression if necessary and the `nkIdentDefs` of relevant locals are moved to the start of their scope.
4. the optimizer (`optimizer.nim`) that eliminates `wasMoved(x); =destroy(x)` pairs that cancel each other out is run

With the new implementation, the transformations are split into multiple passes. The control-flow graph still uses an instruction-based representation, but no longer contains the data-flow-related instructions. There are two new instructions, `join` and `loop`, that together enable loop bodies (independent of their nesting) to be evaluated for their effects only once, and also for the graph to be efficiently traversed backwards. The `mirexec` module implements both the CFG construction and the traversal routines.

Instead of analyzing all uses of all locations for "last reads" in one single CFG traversal, each usage in a consume context (e.g. something passed to a ``sink`` parameter) is now analyzed separately. This greatly simplifies the analysis and reduces the amount of state that needs to be tracked, resulting in a large speedup.

The aforementioned analysis is modeled as "solving" ownership. Each value that results from a computation gets an ownership status assigned -- a value is considered to be "owned" if there are no operations coming *after* the operation it resulted from that observe its presence in the location that stores it. Same as with the previous implementation, location access through aliases (i.e. a pointer) is not considered here. That is, accessing a location through a pointer is not considered to be a read or write of the underlying location. If a value is an r-value (e.g. it's the result of a procedure call and is not a view) it is always considered to be owned.

For destructor injection, instead of tentatively injecting a destructor for each location that has a destructor and letting the `optimizer` elide it later, a destructor is now only inserted for locations for which it is not certain whether they store (and own) a value once control-flow exits their scope. This analysis also takes implicit moves into account.

Previously, a scope was wrapped in a `try` and the destructors called inside the corresponding `finally` clause if one of the following was present inside the scope: a call that raises or an explicit raise (`Defect`s were ignored, independent of `--panics`), a `break`/`continue` statement, a `return` statement, another scope that contains one of the aforementioned. A `finally` clause is now only required if at least one location requiring destruction still stores a (owned) value when its scope is exited via *unstructured* control-flow. For example:
```nim
block:
  var s = @[1, 2]
  if cond:
    f_sink(s) # `s` is consumed
    raise ValueError.newException("")
  else:
    discard

```
`s` requires a destructor call at the end of the block, but a `finally` is not needed, since `s` no longer contains a value once control-flow reaches the `raise`. To be able to detect consumes, the alive analysis requires value ownership to be figured out already.

The actual assignment rewriting and destructor injection are performed as two separate but concurrently applied MIR passes. While it's possible to perform the copy injection for arguments that are not owned and are used in a consume context as another separate pass, it is applied as part of the assignment rewriting pass, in order to reduce the number of `MirTree` traversals.

When an assignment doesn't own the source operand and the value represents a resource (i.e. something that has a `=destroy` hook), the assignment is rewritten into a call to the type-bound `=copy` hook. If the source value is owned, depending on whether or not the destination location is empty, the assignment is rewritten into either a bit-wise copy (`nkFastAsgn`) or a call to the type-bound `=sink` hook. If the source is an lvalue (i.e. there exists a name for the containing location), a destructive move has to be performed (via a call to `wasMoved`), but only if it can't be proven that no connected write observes the now no longer owned value being present in the source location. The analysis (see `needsReset` and `isLastWrite`) takes into account if a destructor is present for the source location.

Since destructors are only injected for alive locations and `wasMoved` calls are only generated where necessary, the separate `optimizer` pass is no longer needed.

## Fixes

Assignments to location with destructors where the right-hand side is a complex expression now respect the strict left-to-right evaluation order:
```nim
var
  i = 0
  s = "str"

# works now:
(i = 1; s) =
  if i == 1: "str2"
  else: doAssert false; ""

# also works now:
proc p(s: var string): var string =
  i = 2
  s

p(s) = block:
  doAssert i == 2
  "str2"

```

A mutation of the argument is now considered to happen when control-flow reaches the call, not when the argument is passed to a `var` parameter. The difference is subtle but important:
```nim
proc f_sink(x: var seq[int], y: sink seq[int]) =
  y[0] = 2
  doAssert x == [1, 2] # passes

proc main() =
  var s = @[1, 2]
  f_sink(s, s) # a full copy is now introduced for the second argument

main()
``` 

Globals are now only moved if it can proven that doing so does not affect program semantics. Previously, assigning a location derived from a global would perform a destructive move in certain circumstances, such as:

```nim
# at module scope
var a = "str"
var b = a # moved

proc p() =
  discard

echo a # echoes an empty string
```

Due to the way the new implementation works, it is now possible for locations derived from globals to be consumed (if safe) when passing them to `sink` parameters.

As a side-effect of the MIR's canonicalization, the alias analysis now handles tuple access correctly:
```nim
proc prc(x: var string) =
  var tup = (x: 0, y: "str")
  # `tup[1]` and `tup.y` were treated as two disjoint locations, causing the assignment
  # below to be erroneously optimized into a move
  x = tup[1]
  doAssert tup.y == "str"

var o: string
prc(o)
``` 

### Timings
Timings and peakmem are taken from compiling the compiler itself with the options `-d:release --compileOnly` and no cached compilation artifacts. The original compiler is build with `-d:release --exceptions:goto`.

|compiler from|bootstrapped with|using|time taken|peakmem|
|:------------|:----------------|:----|:---------|:------|
|#d7a70238d0|`--gc:refc`|`--gc:refc`|24.426s|1015.844MiB|
|this commit|`--gc:refc`|`--gc:refc`|24.687s|1014.254MiB|
|#d7a70238d0|`--gc:refc`|`--gc:orc`|45.206s|1.99GiB|
|this commit|`--gc:refc`|`--gc:orc`|29.531s|1.222GiB|
|#d7a70238d0|`--gc:orc`|`--gc:refc`|15.492s|978.938MiB|
|this commit|`--gc:orc`|`--gc:refc`|15.714s|979.023MiB|
|#d7a70238d0|`--gc:orc`|`--gc:orc`|22.218s|1.933GiB|
|this commit|`--gc:orc`|`--gc:orc`|18.266s|1.188GiB|

---

### To-Do
- [x] adjust the test spec for the tests using `--expandArc`
- [x] split up `tests/lang_objects/destructor/ttests.nim`
- [x] either fix/cleanup or remove the MIR unit tests
- [x] move the MIR introduction into a separate PR
- [x] rebase this PR onto `devel`
- [x] remove the `compiler/sem/optimizer.nim` module -- it's unused now
- [x] fix the issue with procedure-level globals
- [x] address the remaining review comments
- [x] improve the documentation of `mirexec.nim` (partially done)
- [x] compile a list of pre-existing issues and newly introduced regressions
- [x] squash the commits and write a PR/commit message

## Notes for reviewers
- I collected a list of issues/notes in the form of review comments


saem:
I think I understand it, going to write that out for confirmation:

injectdestructors:
1. has mirgen generates an initial `MirTree` (`genProcBody`)
2. then comes up with a list of changes it's going to make (`analyse`)
3. applies them to the `MirTree`
4. converts the updated `MirTree` to `PNode`

----

Prior to reading it, what I was looking for:
- that it's basically all lisp-y on the inside, the AST is a program (we briefly discussed it before in chat and it turns out we're very much on the same page)
- translation to any intermediate should "remember" where it came from
- the preivous point is most true wrt to changes made for precise error/warnings (need lineage)

It seems you've already accounted for the majority of it -- if not all and it's just my understanding falling short. I think it's mostly covered with the `ChangeSet` + `Action` which remembers the `PNode` it's acting upon. So errors should be able to trace their lineage back within MIR. The second bullet point I'm a little less sure about because I can't see the how we could correlate original `PNode` to initial `MirNode` -- the final hop.

So I'm not sure how we flow the info back up to Sem n order for us to ensure we can flow constraints (mostly types info) down from Sem and warnings/errors telemetry backup. Apologies if it's just my lack of understanding the code.

zerbina:
> I think I understand it, going to write that out for confirmation:

Your understanding is correct. I apologize for not having provided a high-level overview.

---

I think I insufficiently described my motivation for this PR, and since that information can help in understanding some decisions behind the current implementation, here goes a second attempt:

The main contributors to my decision of working on this PR first are as following:
* the pre-PR `injectdestructors` produces AST that violates various (reasonable) expectations of `irgen`. For example, sometimes a produced `nkStmtListExpr` has no type (i.e. it isn't actually an expression) or it has a type but doesn't end in an expression. While possible to work around these things, additional logic would be required and `irgen` would also have to become less strict about it's input.
* the idea of implementing the move-analyser at the back-end IR level has been around since the new back-end's inception, but I no longer think that it's a good idea. Over time, the additional complexity of having the back-end IR accommodate for the move-analyser started to outweigh the benefits -- the ideas I collected on how the move-analyser could be implemented work better (if at all) with a slightly higher-level IR.
* `irgen` is quite complex. It does a lot of different things, like performing general transformations and lowering some magics. Then there's also the deferred type translation plus the subsequent fix-up of all type IDs required by it.

So to rephrase, the goal here is to introduce an IR that:
* is meant to support `injectdestructors`, `lambdalifting`, and, later on, maybe also other parts of `transf` (e.g. `closureiterators`, although this one will probably become unnecessary with the introduction of CPS transforms) efficiently operating on it
* works as the input to `irgen`

My current plan for the how the flow should look like in the near-term:
```
input ->
  transform() ->               # `transf`
    toMir() ->                 # currently `genProcBody`
      lambdalifting() ->
        injectdestructors() ->
          backend()            # `irgen`
``` 

Using `MirNode` in it's current form as the input to new back-end (or more specifically, `irgen`) would have the following benefits already:
* multiple transformations previously performed during `irgen` (listed in the PR's description) are now performed in `mirgen`. This would simplify `irgen` a lot.
* the elision of nodes irrelevant to the back-end (e.g. `nkBindStmt`, `nkTypeDef`, etc) together with the linear tree structure make scanning of dependencies (types, procedures, constants, literals, etc.) trivial, rendering, among other things, deferred type translation for `irgen` obsolete.

The main reason behind the `MirNode` -> `PNode` translation is to keep the current code-generators working without having to keep two implementations of `injectdestructors` around.

---

> The second bullet point I'm a little less sure about because I can't see the how we could correlate original `PNode` to initial `MirNode` -- the final hop.

A `MirNode`'s origin is currently not kept track of (maybe to a source location, but not to a `PNode`) - this is something that's still missing.

> So I'm not sure how we flow the info back up to Sem n order for us to ensure we can flow constraints (mostly types info) down from Sem and warnings/errors telemetry backup.

I'm not sure I understand. In which case is it required for information to flow back from the MIR to Sem? Sorry if it's obvious.

saem:
Thank you for the detailed explanation. The motivations make a lot of sense and the overall flow is sensible. I didn't need it, but consider me extra convinced. 😁

-----

> > So I'm not sure how we flow the info back up to Sem n order for us to ensure we can flow constraints (mostly types info) down from Sem and warnings/errors telemetry backup.

> I'm not sure I understand. In which case is it required for information to flow back from the MIR to Sem? Sorry if it's obvious.

Somewhat of a guess on my part, as we're doing lifetime analysis here, then any invariant enforcement would also need to be here?

Possibilities aside, at the time of my prior comment I had done a very quick search in the current inject destructors to look for `localReport` and saw a few hits raising sem errors. I'm presuming we'll have to do the same-ish. Not a very thorough investigation admittedly.

Reading between the lines we'll likely end up with origin being tracked and nothing about the existing approach seems to make that an impossibility, I think you nailed it. 🎉

compiler/sem/mirtraverse.nim
```diff
@@ -0,0 +1,525 @@
...
+  ## `state` is the execution state of the active thread.
+  ##
+  ## `T` needs to have the following interface:
+  ## * ``merge(x: var T, y, T)``: join the state of two threads. The first
+  ##   parameter is always the state of the thread that is joined into
+  ## * ``isFixpoint(x: T)``: tests if the given state is such as that it can't
+  ##   change further. The information is used to exit loops early.
+  # TODO: make sure that the terminology used here is correct and adjust it if
+  #       not
+  # XXX: concepts would be nice to use here, but I believe they shouldn't be
```
saem: yeah, they're just waaaaay too buggy to use.

zerbina:
The most recent commit contains a partial redesign of the MIR.

The previous iteration was a simple, linear-tree-based IR, which was structurally similar to `PNode` (albeit with a lot less node kinds). Except for things like symbols, literals, and other atoms, expressions and statements all used forward trees. That is, first comes the head node (`mnkCall` for example), then all child-nodes (which can be other sub-tree), and then the `mnkEnd` node.

One of the main differences when compared to `PNode` AST is the omission of complex expressions (`if`, `case`, `try`, `block`, and statement list expressions) in MIR -- those are transformed into assignments to temporaries. This was done to simplify both data-flow analysis and further code modifications. For example:

```nim
let x =
  if a: true
  else: false
```

get translated to the MIR equivalent of:

```nim
var tmp: bool
if a:
  tmp = true
else:
  tmp = false
let x = tmp
```

A significant issue with this is the loss of information about the control-flow dependencies between separate expressions. Consider:

```nim
proc fsink(a: sink string, b: string) = ...

fsink(notLastRead x, (if cond: y else: z))
```

which got transformed to the MIR equivalent of:

```nim
var tmp: string
if cond:
  tmp = y
else:
  tmp = z

fsink(notLastRead x, tmp)
```

`x` needs to copied, but it's not clear at which point, as the argument expressions `x` and `tmp` are entirely unrelated in the IR. While the observable effects would be minor in this specific case (string copies happening in a different order), injecting a copy operation directly before the call is wrong, as the copy is required to happen as part of evaluating the respective argument expression.

One solution to this problem would be to assign every argument expression to a temporary first, use the temporaries as the actual arguments, and then inject the copy (or any other operations performed as part of `injectdestructors`, e.g. sinking into an argument) at the position where the respective temporary is defined. This has the significant downside that these extra constraints and behaviour are not expressed explicitly in the IR.

There also were other issues, so I opted for a partial redesign instead.

---

The core of the redesign is a shift to a representation where the focus is more on the flow of values. Expressions are now not tree-based anymore but are instead a sequence of applying operations to lvalues and values. Except for `if` and `case`, statements and their representation were left untouched -- they still use sub-trees. 

Each sequence starts with an input (e.g. a literal or the name of some entity), is followed by zero or more in-out operations, and ends in a sink. A sink is a kind of operation that doesn't produce a value. For example, both `if` and `raise` statements are sinks. A procedure call is treated as always returning a value and has to be explicitly followed by a dedicated `void` sink. Multiple arguments to an operation are supported via argument blocks and the `arg` sink.

As a simple demonstration:

```nim
doSomething(x, "a", (echo "a"; y.b.c.int))
```

would be represented in the MIR as:

```
argBlock:
  proc doSomething -> arg
  local x -> arg
  lit "a" -> arg
  argBlock:
    proc echo -> arg
    lit "a" -> arg
  end argBlock ->
  call -> void
  local y -> path -> stdConv -> arg
end argBlock ->
call -> void
```
Note that the callee procedure is now also only an argument to the `call` operation.

Reading the above top-to-bottom left-to-right yields the nodes in the order they're placed in memory. The arrows are only meant for visualizing the flow of values. For documentation purposes, I wrote down the new grammar as a doc comment in ``compiler/sem/mirtrees.nim``.

I originally wanted for path expressions to use their own IR, mainly because this would allow for de-duplicating them and assigning each one a unique ID, making comparing two path expression for equality very efficient.

The separate IR is also how they're currently implemented, but none of the expected benefits have materialized so far (comparing two paths for equality hasn't come up yet). There are many not-yet-solved issues, and de-duplicating them is complicated by the fact that conversions can be part of paths, so `sameType` is required as part of the comparison (or it would be at least -- de-duplication is not yet implemented).

In general, there are still a lot of open questions, and some constructs, such as user-ops, are under- or unspecified. Nonetheless, I believe that in it's current state, the refactored MIR is already better than the previous iteration.

Related to the MIR changes, there's an experiment with a small template-based DSL for making MIR generation easier -- see `mirconstr` for the DSL and `mirgen` for how it's currently applied.

---

I also made general progress in regards to the actual `injectdestructors` rewrite, and did some down-scoping by removing the unfinished attempt of a re-implementation of cursor inference. Half of the tests in the `destructors` and `arc` categories succeed now.

There are some pre-existing issue with some of the rewrite-rules, mainly in how they interact (or not) with unstructured control-flow. I'll create tests for them, but am probably not going to fix them as part of this PR.

compiler/sem/mirtrees.nim
```diff
@@ -0,0 +1,360 @@
...
+              ## to discard its input) or an input yielding the 'void' value
+              ## (used when wanting to provide nothing as an argument to an
+              ## operation)
+    mnkName
+
+    mnkModify ## modify(x)
+    mnkConsume## consume(x);
+              ## - this is the last use of `x`, consumes `x` (what consume
+              ##   means depends on the context) it
+              ## - this is not the last use of `x`: creates a full copy of `x`
```
saem: is the doc comment saying that consume can either mean:
- it's the last use
- it's the last use of the current `x` and any furthe uses will be based on full copies?

compiler/sem/mirtrees.nim
```diff
@@ -0,0 +1,360 @@
+## This module contais the type defitions for the mid-end IR (MIR) together
+## with routines for querying them.
+##
+## Routines for traversing the used flat-tree-structure are also located here.
```
saem: Bringing over the high level description of the MIR approach from your recent PR comment would be a nice to have.

I'm popping back and forth because my crappy working memory has to continually remind itself of what's up. 😅

compiler/sem/mirtrees.nim
```diff
@@ -0,0 +1,360 @@
...
+
+import
+  compiler/ast/[
+    ast_types
+  ]
```
saem: Casual remark:

I had a fun little realization. I know this module is the entry point, in terms of learning, for the MIR "package" because it's got basically no dependencies. Paired with some instructive comment that's a pretty nice cue in terms of navigating this bit of the compiler.

compiler/sem/mirmodify.nim
```diff
@@ -0,0 +1,238 @@
+## This module implements the ``Changeset`` API, which is used for applying
+## multiple modifications to a ``MirTree`` at the same time.
+
+# TODO: the module likely needs a better name
```
saem: `mirchangesets.nim`

compiler/sem/mirtraverse.nim
```diff
@@ -0,0 +1,539 @@
+## This module implements abstract execution of the mid-end IR. The abstract
+## execution is used to implement the various data-flow analysis.
+
+# TODO: rename the module to something more descriptive
```
saem: `mireval.nim` or `mirexec.nim`

compiler/sem/mirtraverse.nim
```diff
@@ -0,0 +1,539 @@
...
+              states[id] = active
+            else:
+              when defined(debug):
+                active.merge(states[id])
+                # the assertion failing can be caused by an incorrect ``merge``
+                # implementation
+                assert states[id] == active
+
+          # XXX: is it possible for an item coming later in `reevaluate` to
+          #      already be in a "certain" state while an earlier one is not?
```
saem: Only if there exist no dependencies between the two parts of the tree, right? If that's is not the case then we cannot guarantee convergence of the fixed point approach. Likely need some sort of assert but I'm not exactly sure how that might work.

I think a scenario where this might occur is in a loop, where the body behaviour is certain but the loop condition is not? I'm likely grossly misunderstanding the MIR translation and processing, though.

compiler/sem/mirconstr.nim
```diff
@@ -0,0 +1,72 @@
+## This module contains constructor procedures for the different kinds of
+## ``MirNode``s plus convenience procedures for generating expressions.
+
+import
+  compiler/ast/ast_types,
+  compiler/sem/mirtrees
```
saem: any particular reason to separate these out from `mirtrees`?

saem:
I'm a little (very) sleep deprived so my review wasn't not very insightful I gather -- feel free to ignore the review comments if they're just unhelpful. I've read the code as best as I can and revisited your comment, I think the new approach is better (operatinos on l-values).

I'll take another peek tomorrow -- hopefully I'll be able to understand things better then. I wouldn't consider that a blocker.

----

Aside: minor thing is that perhaps there might be some vlaue in differentiating between `callReturnValue` and `callUnit` or something like that in order to drop the `void` requirement. But I'm pretty motivated to destroy conditionals/nested code.

compiler/sem/mirtrees.nim
zerbina: Oh, that doc comment is barely readable - sorry. The `consume` operation takes an unnamed value or l-value as input and produces a value. More specifically:
- if the input is an unnamed value (e.g. the result of a procedure call not returning a view), the output value is the *same* (as in identity) as the input value
- if the input is an l-value and it can be proven that it's not used beyond the point of the `consume` operation, an unnamed value *equal* (i.e. same in-memory representation) of the input l-value is produced and the source is reset (but not destroyed)
- if the input is an l-value that is used beyond the `consume` operation, a full copy of the input l-value is produced

In other words, the value produced by the `consume` operation is always guaranteed to be unique, that is, there doesn't exist another value with the same identity. The `consume` operation was originally named `own` (which I now believe fits better), but for reasons I don't remember, I renamed it.

compiler/sem/mirconstr.nim
zerbina: The construction procedures and utilities are only needed when generating new MIR code. For just comprehending and processing MIR (e.g. `mirtraverse.nim`, `astgen.nim`, etc.), they're not needed, so I thought it'd make sense to put them into a separate module.

compiler/sem/mirtraverse.nim
zerbina: Okay, so what the `traverseReverse` iterator does is to "execute" a span of the control-flow-graph in reverse, propagating a property to all join points.

The iterator itself only takes care of propagating the property and figuring out its state at the entry (or in this case exit) of a basic block -- evaluating said basic block is the responsibility of the iterator's user.

When starting execution, the state of each join point is set to a user-provided default state. For example, in the case of ownership analysis, this would be `yes`. The CFG instructions (which are "executed" before yielding a basic block) are evaluated as following:
- `goto(x) = state(x)`
- `fork(x, y) = union(state(x), state(y))`
- `loop(x) = uncertain`

The pseudo-function `state` either evalutes to the state of the given join-point, or, if the join-point lies outside the executed span, the user-provided default state.

If during the evaluation of the basic-block, the current state of the property is sampled and happens to be `uncertain`, the basic-block and the MIR node at which the sampling occurred are remembered.

The `processRetries` iterator is then responsible for propagating the state between join-points until the would-be-input to all basic-blocks that were remembered earlier is not `uncertain` anymore, yielding the now not-uncertain state together with the also remembered node position.

---

For the comment in question, I believe the answer is: "yes, it can happen". Consider this CFG:
```
0: join
fork 1 # <- `while` condition
fork 5
1: join

... # sample position B
2: join
fork 3 # <- `while` condition
fork 4
3: join
... # the state is set to "yes" here
loop 2
4: join # <- inner `while` end

... # sample position A
loop 0
5: join # <- outer `while` end
```
which would be generated for:
```nim
while cond_a:
  discard "sample position B"
  while cond_b:
    ... # the state is set to "yes" here

  discard "sample position A"
```

In this case, when propagation first reaches `B`, the state would be certain, while for `A` it would still be `uncertain`. This only affects the order in which the retry items are yielded however, which in all current use-cases doesn't matter.

compiler/sem/mirconstr.nim
saem: Fair enough, I can go either way on this sort of thing. Maaaaybe consider renaming it `mirtrees_constr.nim` so it feels a bit subordinate to the other module. Might hint at "don't read me first".

compiler/mir/analysis.nim
```diff
@@ -0,0 +1,616 @@
...
+## They're based on the ``mirexec`` traversal algorithms and require a
+## ``Values`` dictionary and a ``ControlFlowGraph`` object, both
+## corresponding to the code fragment (i.e. ``MirTree``) that is analysed.
+##
+## A ``Values`` dictionary stores information about the result of operations,
+## namely, whether the value is owned and, for lvalues, the root. It also
+## stores the lvalue effects of operations. An instance of the dictionary is
+## created and initialized via the ``computeValuesAndEffects`` procedure.
+##
+## XXX: the following section doesn't fit here. Maybe move it to somewhere else
```
saem:
```suggestion
## Implementation Notes:
## ----------------------------------
```

Maybe a heading is sufficient?

compiler/mir/analysis.nim
```diff
@@ -0,0 +1,616 @@
...
+
+import std/packedsets
+
+type
+  # XXX: this type is useful in general (but probably needs a better name)
+  #      and should be moved somewhere else
+  IndexedSeq[I: Ordinal, T] = distinct seq[T]
+    ## Similar to a ``seq``, but only can only be accessed by values of the
+    ## specified type. This is useful in situations where the type of the
+    ## index value is a ``distinct`` integer-like type
```
saem:
```suggestion
  OrdinalSeq[I: Ordinal, T] = distinct seq[T]
    ## Similar to a ``seq``, but can only be accessed by values of the
    ## specified type `I`. This is useful in situations where the type of the
    ## index value is a ``distinct`` integer-like type
```

It's fine to move this into the standard library experimental like `results` and `dod_helpers`.

compiler/mir/analysis.nim
```diff
@@ -0,0 +1,616 @@
...
+
+  Values* = object
+    ## Stores information about the produced values plus the lvalue effects of
+    ## operations
+    values: IndexedSeq[OpValue, ValueInfo]
+    # XXX: `values` currently stores an entry for each *node*. Not every node
+    #      represents an operation and we're also not interested in the value
+    #      of every operation, only of those that appear in specific contexts.
+    #      A ``Table`` could be used, but that would make lookup less
+    #      efficient (altough less used memory could also mean better memory
```
saem:
```suggestion
    #      efficient (although less used memory could also mean better memory
```

compiler/mir/analysis.nim
```diff
@@ -0,0 +1,616 @@
...
+  EffectMap = object
+    ## Accelerator structure that maps each operation to its lvalue effects.
+    list: seq[Effect]
+    map: Table[Operation, Slice[uint32]]
+
+  Values* = object
+    ## Stores information about the produced values plus the lvalue effects of
+    ## operations
+    values: IndexedSeq[OpValue, ValueInfo]
+    # XXX: `values` currently stores an entry for each *node*. Not every node
```
saem, line 98: Tables are really really really slow. :(

They absolutely tanked performance when I was first prototyping a DOD AST with them.

compiler/ast/ast_query.nim
```diff
@@ -746,3 +746,14 @@ proc `[]`*(node: PNode, slice: NodeSliceName): seq[PNode] =
     of SliceAllIdents: node.sons[0..^3]
     of SliceAllArguments, SliceAllBranches: node.sons[1..^1]
     of SliceBranchExpressions: node.sons[0 .. ^2]
+
+iterator branches*(node: PNode): tuple[position: int, n: PNode] =
```
haxscramper: > \#\# Traverse banches of the `case` and `if` nodes, returning *branch position* in the main list of branches

Add documentation that explains that `position` != `subnode index`

compiler/ast/ast_query.nim
```diff
@@ -746,3 +746,14 @@ proc `[]`*(node: PNode, slice: NodeSliceName): seq[PNode] =
...
+  assert node.kind in {nkRecCase, nkCaseStmt, nkIfExpr, nkIfStmt}
+  for i in 1..<node.len:
+    yield (i-1, node[i])
+
+iterator branchLabels*(node: PNode): (int, PNode) =
```
haxscramper: > \#\# Return all labels/expressions used in the  `of`, `elif`, `else`, `except` branches.

haxscramper (approved):

saem (approved):
I'm suggesting small bits like doc comments right now as I'm refreshing my memory, thankfully the complexity is largely in the subject matter and not the implementaiton. 😄 

Observations thus far:
- the code is clean and I feel like if there were issues I could tests and fix them
- similar to above, I can see myself or others extending it
- this obviously took a lot of effort and it shows, thank you thank you!
- my confidence in the change is mostly around CI at this point

🎉

----

I'm marking it approved. I'll keep reading and if I find areas of improvement that meaningfully simplify things or can easily be extracted, I'll note them in the PR.

compiler/sem/mirexec.nim
```diff
@@ -0,0 +1,618 @@
...
+
+template upperBound(c: ControlFlowGraph, node: NodePosition): InstrPos =
+  upperBound(c.instructions, node, compare).InstrPos
+
+func `[]`(c: ControlFlowGraph, pc: SomeInteger): lent Instr {.inline.} =
+  c.instructions[pc]
+
+func computeCfg*(tree: MirTree): ControlFlowGraph =
+  ## Computes the control-flow graph for the given `tree`. This is a very
+  ## expensive operation!
```
saem:
```suggestion
  ## Computes the control-flow graph for the given `tree`. This is a very
  ## expensive operation! The high cost is due to two essential reasons:
  ##
  ## 1. a control flow graph needs to materialize all paths for a given `tree`
  ## 2. the amount of allocations/bookkeeping necessary to do so
```

This is partly me just confirming my understanding, the first point I'm fairly confident about and the second one is more me guessing by looking at the push/pop of sequences and the `Table` in `ClosureEnv`.

Aside, I spent a few minutes trying to reason about replacing that table and the best I could figure was maybe the following:

```
type
  ClosureEnv = object
    # ... existing bits except joins
    joinedNodes: seq[NodePosition]
    nodeIsJoined: PackedSet[NodePosition]`

proc addJoin(env: var ClosureEnv, dest: NodePosition): JoinPoint =
  let (id, isNew) =
    if env.nodeIsJoined.containsOrIncl(dest):
      env.nodeIsJoined.add dest
      (env.joinedNodes.len - 1, true)
    else:
      (env.joinedNodes.find dest, false)  # likely more efficient with a reverse find

  JoinPoint(id: id, node: dest, isNew: isNew)
```

This is mostly a test of my understanding and a bit of fun, I doubt it's an improvement.

compiler/mir/analysis.nim
zerbina: Ah, sorry. The text that this comment was referring to got moved somewhere else already.

compiler/mir/analysis.nim
zerbina: Hm, I'm not sure about the name `OrdinalSeq`. Going by just the name, I'd assume that it's a sequence that stores ordinals.

compiler/sem/mirexec.nim
zerbina: I wouldn't say that the control-flow graph needs to materialize all paths for `tree`, rather only all edges (in the form of instructions). However, I might be misunderstanding what you mean.

Regarding the reason for why `computeCfg` is expensive, the last time I profiled it, the by far largest amount of time was spent inside the tree traversal routines. For a general perspective, using a `-d:release` compiler built with `--gc:orc` to compile itself with `--gc:orc` as the benchmark, more time (~15%) is spent in `computeCfg` than in `mirgen` (i.e. `genProcBody`).

I did another profiling session using the newest commit -- the six most time consuming calls inside `computeCfg` are:
| Procedure | Time spent |
| --------- | ---------- |
| `mirtrees.findEnd` | 0.401 % |
| `mirtrees.parent` | 0.224 % |
| `algorithm.sort` | 0.0476 % |
| `commit` | 0.0174 % |
| `mirtrees.findParent` | 0.0174 % |
| `addEdge` | 0.011 % |

The "Time spent" is a percentage of the total compilation time. Using either recursion or a skip-list (or both) could reduce the amount of time spent on tree traversal.

---

The snippet you posted is, apart from two problems that I believe are unintended, correct. I was curious and ran a comparison between your code and the current implementation. Instrumenting a release compiler that is booted with `--gc:orc` and using it to compile the `devel` compiler with `--gc:orc` yields:

|   | Time taken (in milliseconds) | Allocations | Deallocations |
| - | ---------------------------- | ----------- | ------------- |
| `Table` | 151 | 35219 | 17779 |
| `PackedSet` + `seq` + `find` | 152 - 153 | 36865 | 19425 |
| `PackedSet` + `seq` + `rfind` | 154 - 155 | 36865 | 19425 |

zerbina:
Thank you for the review @haxscramper and @saem!

The `for`-loop PR and should fix the issue that currently causes a C compiler error for `nimgrep`.

zerbina:
Merging the MIR is progressing rather slowly, as there are a lot of separate fixes required first. The most recent ones being fixes to the JS code-generator that are blocking the merging of #498, which is the PR with the fixes for the issue that is making bootstrapping fail in the Linux CI for this PR (the bug causing the `nimgrep` to not compile is already fixed in the MIR PR).

compiler/sem/mirexec.nim
saem: The allocations really dictate performance, kinda sad that `PackedSet`s weren't able to get ahead. I'll have to look closer especially once the AST starts to get the DOD treatment as I'd been hoping they'd do a lot of heavy lifting in a few areas.

compiler/mir/analysis.nim
saem: If I think about `SeqThing` vs `ThingSeq`, I read those two as `sequence of thing inside` and `thing sequence`, respectively. The former is clear and the latter is a bit ambiguous, unless put side by side as above. It also doesn't help that we have things like `TNodeSeq`, which on its own makes sense.

Given all that here is the list of names I can think of in no particular order:

- `IndexedSeq`
- `OrdinalSeq`
- `OrdSeq`
- `SeqByOrd`
- `SeqByOrdinal`
- `ByOrdSeq`
- `ByOrdinalSeq`
- `OrdIndexedSeq`
- `OrdinalIndexedSeq`

zerbina:
The PR is now rebased onto `devel` (a slightly outdated version of it), meaning that the things already split out and merged separately are no longer part of the changeset.

As for the test failures, the relevant ones are caused by the temporary injection pass breaking the assumptions of `mirgen`.

`mirgen` translates the definitions (i.e. `nkIdentDefs`) of procedure-level globals in a way so that the resulting MIR code is turned back into an `nkVarSection` + `nkIdentDefs` by `astgen`, but with the injection of temporaries now happening, the assumption made by `mirgen` that the code it generates reaches `astgen` without modifications is no longer true.

Since the fix does not depend on the this PR and requires changes to `transf` and the code-generators, I'm going implement it in a separate PR.

compiler/mir/analysis.nim
zerbina: I moved it to `utils/containers.nim` and renamed it to `OrdinalSeq`.

compiler/mir/analysis.nim
```diff
@@ -0,0 +1,618 @@
...
+    ## transfered to the operation)
+  UseContext* = {mnkArg, mnkDeref, mnkDerefView, mnkCast, mnkVoid, mnkIf,
+                 mnkCase} + ConsumeCtx
+    ## using an lvalue as the operand to one of these operators means that
+    ## the content of the location is observed (when control-flow reaches the
+    ## operator). In other words, applying the operator result in a read
+    # FIXME: none-lvalue-conversions also count as reads, but because no
+    #        distinction is being made between lvalue- and value-conversions
+    #        at the MIR level, they're currently not considered. This is an
+    #        issue and it needs to be fixed
```
zerbina, line 128: This is a small pre-existing issue that `dfa.nim` also has. In the current implementation it's only a cosmetic issue, but it should eventually be fixed nonetheless. I post-poned it, since the fix requires adjustments of the MIR.

Example:
```nim
var x = float(0.0)
discard int(x)
# `x` is treated as being used when control-flow reaches the `discard`
# statement, not when control-flow reaches the conversion  
```

compiler/sem/injectdestructors.nim
```diff
@@ -7,1094 +7,1390 @@
...
+  ## If the ``=copy`` hook is requested but not available because it's
+  ## disabled, a diagnostic is added to `msgs`.
+  # XXX: the procedure does too much and is thus too complex. Splitting the
+  #      'consume' handling into a separate procedure would makes sense, but
+  #      would likely also be less efficient due to the required extra
+  #      (linear) tree traversal.
+  #      Another possible improvement is moving the actual hook injection
+  #      to a follow-up pass. The pass here would only inject ``mCopyAsgn`` and
+  #      ``mSinkAsgn`` magics, which the aforementioned follow-up pass then
+  #      expands into the hook calls. This would simplify the logic here a bit
```
zerbina, line 971: It is almost certain that this is the approach (without the magics) that I'll go with when rewriting/refactoring the pass. In my opinion, the benefits (composes better, much simpler logic, easier to understand) far outweigh the downsides (one more pass -> likely slower).

compiler/sem/injectdestructors.nim
```diff
@@ -1108,42 +1404,96 @@ proc deferGlobalDestructor*(g: ModuleGraph, idgen: IdGenerator, owner: PSym,
...
+  block:
+    var changes = initChangeset(tree)
+
+    injectTemporaries(tree, changes)
+
+    # the VM implements branch switching itself - performing the lowering for
+    # code meant to run in it would be harmful
+    # FIXME: discriminant assignment lowering also needs to be disabled for
+    #        when generating code running at compile-time (e.g. inside a
+    #        macro)
```
zerbina, line 1444: This is going to be a problem once the `injectdestructors` pass is also used on code meant for running in the VM, but since that's not the case yet, no action is required for now.

compiler/sem/mirexec.nim
```diff
@@ -0,0 +1,744 @@
...
+        # only create an edge for the exit branches that are not the last one
+        if tree[i+1].kind != mnkEnd:
+          goto i, parentEnd(tree, i)
+      of mnkExcept:
+        # after a structured exit of the handler, control-flow continues after
+        # the code section it is attached to. Node-wise, this is the ``end``
+        # node terminating the ``try`` the handler is part of
+        # TODO: with this approach, either the last branch needs to fork
+        #       control-flow to the next exception handler or ``mirgen`` has to
+        #       introduce a catch-all handler
```
zerbina, line 346: This is the second most severe issue (that is somewhat pre-exisiting). In the following example:
```nim
try:
  try:
    discard # 1
    raise SomeError.newException("")
  except OtherError:
    discard
except SomeError:
  discard # 2
```
there will be no path between 1 and 2 in the control-flow graph. The current plan for fixing this is to, as part of `transf`, add what amounts to `except: raise` to an `nkTryStmt` that has at least one exception handler but not a "catch-all" one. Not only would that fix the issue in question here, there's also logic in the C and JS code-generators that can be removed then.

compiler/sem/mirexec.nim
```diff
@@ -0,0 +1,744 @@
...
+      # jump to the end of the loop:
+      s.pc = p
+      s.i = c[p].node
+      # remember the current `top` and prevent blocks already visited during
+      # the first pass from being visited again by setting `top` to the current
+      # time:
+      s.visited[id] = s.top
+      s.top = s.time
+      # XXX: it's not formally proven that not setting `s.time` to `s.bottom`
+      #      before entering the loop is correct
```
zerbina: I'm not sure if it's a good idea to leave this in, even if only temporary. If the optimization is incorrect, the traversal would happen in the wrong order in the case of complex nested loops.

tests/arc/topt_no_cursor.nim
```diff
@@ -11,98 +11,98 @@ doing shady stuff...
...
-`=copy`(saved, sibling.right)
-`=copy`(sibling.right, saved.left)
-`=sink`(sibling.parent, saved)
+var sibling
+var :tmp = target[].parent[].left
+`=copy`(sibling, :tmp)
+var saved
+var :tmp_1 = sibling[].right
+`=copy`(saved, :tmp_1)
+var :tmp_2 = sibling[].right
```
zerbina, line 36: Although it looks like it, no copy is created here. The `var :tmp2 = ...` is actually a `var :tmp2: var T`, but there's an `Empty` node in the `IdentDefs` type slot, so nothing is rendered.

tests/arc/topt_cursor.nim
```diff
@@ -3,33 +3,36 @@ discard """
...
   var f = open("debug.txt", fmRead, 8000)
   try:
     var res
     try:
       res = newStringOfCap(80)
-      block :tmp_1:
-        while readLine(f, res):
+      block label_1:
+        while true:
+          if op(readLine(f, res)):
```
zerbina, line 32: Since `astgen` has no way of knowing the name of the original symbol (apart from maybe inspecting the source node), it turns `mnkMagic` MIR nodes into symbols with the name `op`.

tests/lang_objects/destructor/tobjfield_analysis.nim
```diff
@@ -49,3 +49,23 @@ proc main =
...
+    b: Value
+
+  proc prc() =
+    # a literal constructor expression (which ``initValue`` also expands to)
+    # not used *directly* in a consume context only produces a non-owning
+    # container
+    # TODO: this doesn't match the behaviour the specification describes (i.e.
+    #       constructor expression having the same semantics as procedure calls
+    #       in the context of lifetime hook injection). The behaviour here
+    #       should either be explicitly added to the spec or changed
```
zerbina, line 66: This is pre-existing behaviour that's not described by the specification.

tests/arc/tstrformat.nim
```diff
@@ -2,9 +2,20 @@ discard """
...
 '''
-  cmd: "nim c --gc:arc $file"
+  matrix: "--gc:arc"
+  targets: c
+  knownIssue: '''
+    Semantic analysis duplicates the analysed argument into each usage
+    inside the template. Since it contains definitions of local variables,
+    multiple definitions of the same entity exist in the expanded code, which
+    is illegal at the MIR level, thus making the `injectdestructors` pass fail
+    '''
```
zerbina, line 12: I decided to disable this test for now. It's the most severe issue, but there are so many hacks in `mirgen` already that it felt wrong to introduce yet another one. For now, an entity being defined multiple times will raise an internal compiler error when the code reaches the `injectdestructors` pass.

zerbina (commented):
After fixing the remaining bugs (the backward traversal algorithm didn't work for nested loops, which I only noticed due to a single test failing with Valgrind reporting a leak) all tests pass, and I consider this PR to be done (apart from squashing and writing a commit message).

While fixing many bugs the old implementation had, a few new ones were introduced. I noted them, together with their impact, down via review comments, but in general think that none of them are severe enough to block this PR. The most critical one can be hacked around with a medium amount of effort, though I'd prefer to not do so as part of this PR.

compiler/mir/analysis.nim
saem: So this paired with very aggressive/eager free placement we could introduce a bug where we free before the last use?

I agree that this is minor, IIUC we're not that aggressive currently, nor would we want to be in the future.

compiler/sem/injectdestructors.nim
saem: Agreed, I suspect looking at pass logic coalescing which fuses a bunch of these operations will be easier if each pass's logic is very simple/straightforwad. Not to mention, it's likely easier to get some quick wins with a little bit of pass composition based optimizations, than maintain a bunch of very carefully hand tuned implementations.

compiler/sem/injectdestructors.nim
saem: Agree that it can be deferred. Moreover, that it's likely a good thing, I have a funny feeling about handling the same thing at different levels.

compiler/sem/mirexec.nim
saem: Handling it in a common place makes sense.

Although, not sure I could quite put together the proposed solution, would it effectively be this?
```nim
try:
  try:
    try:
      discard # 1
      raise SomeError.newException("")
    except OtherError:
      discard
  except SomeError:
    discard # 2
except:
  raise
```

compiler/sem/mirexec.nim
saem: Yes, let's back it out.

tests/arc/topt_no_cursor.nim
saem: Hmm, wish there was a way to comment it without mucking up the input. 🤔

tests/arc/tstrformat.nim
saem: This sounds like it should be fixed in semantic analysis, rather than hack up `injectdestructors`.

My next two days are really busy, but I could peek at sem and eval template code on the weekend, not sure what the fix will entail, though

tests/arc/topt_cursor.nim
saem: This is going to require far more fundamental fixes, outside the scope of this PR, of course.

tests/lang_objects/destructor/tobjfield_analysis.nim
saem: It's useful to know if it changed. Noting that this behaviour is not per the spec and one or the other should change is the right thing to do. I want to have more of a think about what it means, then we can discuss what makes sense.

Documenting it and leaving it as is good!

compiler/sem/mirexec.nim
```diff
@@ -0,0 +1,744 @@
...
+    # the join point is the start of a loop
+    let (_, p, isRetry) = s.loops[^1]
+    # note that at this point, `s.active` can only be true if this is the first
+    # we're reaching the loop start while being active. For each following
+    # visit, active will always be 'false'
+    if s.visited[id] == 0 and s.active:
+      # this is the first time we're reaching this loop start -- jump back
+      # to the end of the loop and set `isRetry` for this loop-start to
+      # true so that we know whether it'a a loop exit the next time we reach it
+      s.loops[^1][2] = true
```
Clyybber:
```suggestion
      s.loops[^1].isRun = true
```

zerbina: Thanks, changed it.

compiler/mir/analysis.nim
zerbina: It's very likely that there's never going to be an issue, even if we're freeing very aggressively. The value resulting from a non-lvalue conversion is always used directly after, and there's nothing in between where a destruction could be placed.

compiler/sem/mirexec.nim
zerbina: Almost. The proposed solution would like this:
```nim
try:
  try:
    discard # 1
    raise SomeError.newException("")
  except OtherError:
    discard
  except:
    raise
except SomeError:
  discard # 2
except:
  raise
```
This is roughly what the C and JS code-generators already do, and the control-flow graph generation logic would generate the correct edges without requiring any changes to it.

tests/arc/tstrformat.nim
zerbina: Yep, semantic analysis is the correct place to fix this.

The issue here was what originally motivated me to look into the macro output sanitizer. Turning the typed AST passed to a `typed` template parameter into untyped AST before inserting it into the body could fix the problem, but work on the sanitizer is currently stuck due to issues with gensym'ed symbols.

PR #454, opened 2022-10-25
improve S-expression parser robustness

haxscramper:
- Handle  newline  characters  in  the  input  expression  properly,  using `handleRefillChar`. This allows parsing  more crude auto-generated output produced  by the  `pretty`  option (previously  bidirectional serde  test failed).
- Make spacing between list elements  optional -- factor out space skipping into a separate function.
- Add tests  to ensure  buffer is  refilled properly  for larger  inputs -- previously 8192 characters or more ran  into the lexer buffer limit which caused parser to fail as well.

### DIFF
--- a/lib/experimental/sexp.nim
+++ b/lib/experimental/sexp.nim
@@ -51,8 +51,8 @@ type
     errEofExpected,        ## EOF expected
 
   SexpParser* = object of BaseLexer ## the parser object.
-    a: string
-    tok: TTokKind
+    str: string
+    tok: TTokKind ## Current token that lexer has processed
     kind: SexpEventKind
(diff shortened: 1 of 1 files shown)


saem:
Quick thing I noticed, the commit and PR body use `~`, which Renee as strike through, I don't think that's what you meant having taken a quick scan of the code.

haxscramper:
> Quick thing I noticed, the commit and PR body use `~`, which Renee as strike through, I don't think that's what you meant having taken a quick scan of the code.

A habit from https://orgmode.org/manual/Emphasis-and-Monospace.html I will fix it and merge of there are no other objections.

PR #457, opened 2022-11-03
Skip "failing known issue" in CI post-failing

haxscramper:
- Clean  up implementation  of the  backend reporting:  pass around  proper report object that we already have instead of taking it apart into half a dozen string  arguments that need  to be pieced  back into a  JSON object anyway.
- Add information about known issues to the testament JSON cache so they can be used in later PR for tracking known issues that no longer fail
- Skip failing issues  in the CI post-processing in order  to reduce visual clutter.

### DIFF
--- a/testament/backend.nim
+++ b/testament/backend.nim
@@ -6,7 +6,55 @@
 #    Look at license.txt for more info.
 #    All rights reserved.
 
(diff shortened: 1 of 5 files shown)


alaviss (approved):
lgtm

PR #463, opened 2022-11-12
remove explict mention of WASM runtime

haxscramper:
Rationale: while it is technically/factually correct we don't have *any* tests to back  up this claim. Not all GC runtimes work reliably with wasm (`refc` tends to break  completely on startup), it is not out-of-the-box experience in any way (it might be in the future, but right now it is not) and it is not documented anywhere.

Can be switched back on when we actually have this implemented properly


haxscramper:
Super trivial thing, but I think I should get yes/no confirmation, as it is a good example/data point when it comes to feature advertising in general (how decisions about readmes are made etc).

saem (approved):
We can always add it back later, just with tests and all that good stuff.

(less is more)

PR #464, opened 2022-11-13
compiler: remove support for hot-code-reloading

zerbina:
## Summary
HCR was only tested by a single test and made the code-generator logic more complex. The feature is removed in order to reduce the amount of code in `cgen` and to make future changes in the code-generator easier.

The plan is to re-implement support for hot-code-reloading at a point in the future.

## Details
- remove the HCR compiler support
- remove the HCR runtime
- remove the enabled test. The disabled `nimhcr_integration.nim` test is kept
- remove the `nonReloadable` and `executeOnReload` pragmas

---

If wanted, I can also remove the `nimchr_integration.nim` test. The idea behind keeping it was to already have some existing tests when re-implementing HCR support.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


lib/system.nim
```diff
@@ -529,10 +529,7 @@ proc `..`*[T](b: sink T): HSlice[int, T]
   ##   echo a[.. 2] # @[10, 20, 30]
   result = HSlice[int, T](a: 0, b: b)
 
-when defined(hotCodeReloading):
-  {.pragma: hcrInline, inline.}
-else:
-  {.pragma: hcrInline.}
+{.pragma: hcrInline.}
```
saem: Could we get rid of this too?

saem:
Nice!

Outside of CI failure I think it's good to go.

lib/system.nim
zerbina: Sure. I only kept it in order to reduce the size of the change-set.

saem: Makes sense, but more feels is more points. 😁🥳

lib/system/cgprocs.nim
```diff
@@ -13,8 +13,8 @@ type
   LibHandle = pointer       # private type
   ProcAddr = pointer        # library loading and loading of procs:
 
-proc nimLoadLibrary(path: string): LibHandle {.compilerproc, hcrInline, nonReloadable.}
-proc nimUnloadLibrary(lib: LibHandle) {.compilerproc, hcrInline, nonReloadable.}
-proc nimGetProcAddr(lib: LibHandle, name: cstring): ProcAddr {.compilerproc, hcrInline, nonReloadable.}
+proc nimLoadLibrary(path: string): LibHandle {.compilerproc, nonReloadable.}
```
saem: Personally, I would remove `nonreloadable`, as well.

I know it increases the size, but it more importantly highlights how invasive this whole feature was in the first place.

saem: this is non-blocking, I'm good to merge it.

zerbina: It can go too - the required changes are of an okay amount.

saem: Sweet!

zerbina:
The `.nonReloadable` pragma is now also removed. I missed removing the `executeOnReload` pragma earlier, but that's gone now too.

bors r+

saem:
Yaaaaaasssssssssss!

PR #461, opened 2022-11-08
Clean up `stdlib` test category

haxscramper:
- Create  additional  subdirectories  for   the  standard  library  testing
  category. Instead of a  giant mess with over a hundred  files in the same
  directory it  is now split into  several smaller pieces. Other  test that
  /should/ be in the stdlib category  have not been moved, TODO for another
  PR.
- Add  `"cache"`   command  that  was   missing  from  the   testament:  it
  was  *impossible* to  run  a single  category and  then  retry it  again,
  because `cacheResults()` was only called at the end of `all` (which takes
  a lot more time to run).
- Newly  introduced directories  in  the stdlib  category have  `readme.md`
  attached to them.

Future directions:

- there  are  several completely  useless  categories  along the  lines  of `tactiontable.nim` (I can't even fathom how someone thought it might be a good idea to place this into a separate category) and others that need to be moved to the `stdlib`.
- Provide documentation  on the  end-to-end process  of placing  tests into appropriate directories.
- Start   rearranging   internal   stdlib   tests   `tstdlib_various`   and
  `tstdlib_issues` are absolutely revulsing and should be organized.

Smaller version of https://github.com/nim-works/nimskull/pull/455

PR #465, opened 2022-11-13
test organization

haxscramper:
Follow-up for https://github.com/nim-works/nimskull/pull/461, mostly for imports and some docs

- Replace `import tables`  and `import sets` in tests  (replacement is done using `fd | sd`, so it is not exhaustive yet).
- Handle `testament  cat stdlib/os` properly --  there is no reason  to ban this as  (surprise!) test  directory might  have more  than one  level. I guess level of  orderliness of the core maintainers shows  in that pretty well.  After fifteen  years nobody  had ever  thought that  tests can  be placed into something other than 133+ entries on the same level.
- Add  small entry  for  using  `koch.py` for  test  running  tests in  the documentation.
- Add missing testament link from the documentation
- Comment out `nimpretty` link as it was removed earlier.

### DIFF
--- a/.github/contributing.rst
+++ b/.github/contributing.rst
@@ -84,6 +84,27 @@ forward:
 .. note:: Progress of the test suite improvements is tracked in the github
           `project <https://github.com/nim-works/nimskull/projects/2>`_.
 
+Running test suite locally
+--------------------------
+
+You can use `./koch.py` tool to execute tests:
+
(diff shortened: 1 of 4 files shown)


saem:
I think there is some special handling within `categories` inside `processCategory` proc at the very least, that and others need to become aware of categories now having subdirs. Otherwise it's on the right track.

PR #473, opened 2022-11-20
More test category cleanup

haxscramper:
- Move constraints "category"  to tests/generics/tconstraints.nim -- single file is  not a "category",  you can just  run testament in  a single-file mode, it is not that hard.
- Enums to tests/lang_defs/enum/
- Concepts to tests/lang_experimental/concepts/  -- they *are* experimental right now anyway
- tests/lang_exprs/compiles/ -- compiles is an expression built-in magic?
- tests/lang_stmts/controlflow/   --   was    testing   block   and   other
  statement-related things
- tests/lang_types/ -- ranges and types,  most likely will have some string
  and int-related magic
- tests/stdlib/strings/tcnstseq.nim -- was a single-file category
- tests/stdlib/strings/tconcat.nim   --   another  single-file   "category"
  removed.


haxscramper:
Three single-file categories, FFS who *managed* this test suite before? Oh, right, it was thrown in here **six years ago** and only briefly touched since then https://github.com/nim-lang/Nim/commit/20b5f31c03fb556ec0aa2428a40adbac004d8987

PR #475, opened 2022-11-20
compiler: remove `--newruntime`

zerbina:
## Summary
Remove compiler support for `owned ref|proc` together with the
`--newruntime` switch.

`owned` reference-like types have no specification (aside from the RFC
that proposed them), introduce another dialect, are deprecated already,
and also complicate the rewrite of the `injectdestructors` pass.

Removing them simplifies the compiler a bit, making further refactorings
easier.

## Details
- the type-operator `owned` is kept, but now always treated as a no-op
- remove `tfHasOwned`
- remove `tyOwned`. `TNimTypeKind` is adjusted accordingly
- remove the `mUnown` magic
- remove the related reports
- remove the `--refchecks` switch
- remove the associated tests
- the `tnewruntime_misc.nim` test is adjusted to work with `--gc:arc`


tests/destructor/tnewruntime_misc.nim
```diff
@@ -1,13 +1,13 @@
 discard """
-  cmd: '''nim cpp -d:nimAllocStats --newruntime --threads:on $file'''
-  output: '''(field: "value")
-Indeed
+  matrix: "-d:nimAllocStats --gc:arc --threads:on"
```
saem, line 2: Maybe rename the test to `tarc_alloc_dealloc_bugs.nim`?

saem (commented):
Would it be a ton of work to add in a deprecation warning for `owned`? I think we should remove it entirely. It's likely entirely off from what we want.

A better future approach is that all memory and resource allocations are scoped (structured). This isn't entirely the same thing as a stack, those directives to be sorted as we deeply explore the design.

What this means is that if we allocate memory:
- its lifetime will be tied to region's life
- same for any references and pointers, their life cannot exceed it

What this means for resources (threads, continuation cactus subtree, db connection, etc):
- just like memory will never exceed those lifetimes
- in addition to dealloc, this means cancellation support (parent can kill our child can abort)

```nim
type
  MyData = object
    # lots of state with refs

proc bar(d: MyData) =
  # complicated stuff, allocate

proc foo() =
  var
    data: MyData
    t = runInBackground(bar(data))
```

Without getting into the language features we need and want to add (eg: how to specify `data` and `t` to go into a longer region/lifetime) we can already have some very nice guarantees:

- no matter how complicated data gets it'll die with foo
- no matter what the thread tries to do it cannot do it for longer than foo and to anything outside the region

This isn't the whole picture, of course. But I think fine grained reasoning of ownership isn't the path forward. Structured lifetime provides a step to first batch and constrain this stuff and then we can see about any small bits we need in order to coordinate a few potential things:
- how bar can leave things for foo
- how foo and bar can leave data for each other as they operate (pass without copying, avoiding locks, etc)

I strongly suspect what we'll end up with is sub-region markings to allow stuff like:
- what one thing can write and others can read
- setup region(s) for the parent by the child, where when the child dies those regions go to the parent instead of reclamation

Anyhow, I just don't see owned surviving that, but maybe you disagree.

saem:
Should have clarified, the deprecation would allow us to track down and remove owned over time rather than in this PR.

saem:
If adding the deprecation is a bridge too far, then not a blocker, one of us can follow-up with that later.

🤞🏽 I finish the cpp change, then I hopefully have a solution for that symbol redefinition issue.

tests/destructor/tnewruntime_misc.nim
zerbina: From what I can tell by looking into the fixes, none of them are related to ARC. I renamed the file to the generic `tarc_issues.nim` and left a to-do comment in order to remind me of splitting it up later.

zerbina:
> Would it be a ton of work to add in a deprecation warning for `owned`?

Nope, not at all. There's now a dedicated deprecation warning when using `owned`.

Thank you for the detailed write-up on how you envision memory/resource management to work. I think I understood it.

`owned` was a step in this direction I believe, as it bound the lifetime of the heap-allocated location to the lifetime of the location holding the `owned` value. If the `owned ref|proc` went out of scope, the heap-allocated location was destroyed and deallocated.

saem:
> > Would it be a ton of work to add in a deprecation warning for `owned`?
> 
> Nope, not at all. There's now a dedicated deprecation warning when using `owned`.

Thanks, I'm good to merge, will bors after.

> Thank you for the detailed write-up on how you envision memory/resource management to work. I think I understood it.

Thanks for reading it.

> `owned` was a step in this direction I believe, as it bound the lifetime of the heap-allocated location to the lifetime of the location holding the `owned` value. If the `owned ref|proc` went out of scope, the heap-allocated location was destroyed and deallocated.

Yup, the issue with `owned` is that the underlying implementation and primitives are all off. It's working entirely on locations rather than location and region. Until regions exist owned is just not able to signal the right things.

saem:
Bors r+

PR #476, opened 2022-11-20
Remove C++ and ObjectiveC backends

saem:
## Summary:
* removes the C++ and ObjectiveC backends
* removes the genode os target
* this is the first major step towards greatly simplifying codegen

### Interface/Language Changes:
- removed cpp and objective c related commands from the CLI
- removed `importcpp` pragmas, use `importjs` where necessary for JS
- removed `constructor` pragma
- removed mixed C/C++ mode compilation
- removed `cppCompileToNamespace` CLI
- removed cpp exceptions, replaced with `native`, covers JS
- removed thiscall calling convention
- `importjs` is equivalent to `importcpp` no longer requiring a pattern

### Internals:
- removed type flag `tfVarIsPtr` as this was for C++ only
- removed C++ related error reports
- removed named parameter mapping for ObjectiveC
- removed `ccThisCall` calling convention

### Docs/Misc:
- removed C++ specific tests or generalized where possible
- removed cpp and objc from testament backends
- removed nearly all C++, Objective C, and Gendoe mentions from docs
- removed dead code in vmgen, `sizeOfLikeMsg`

---


compiler/ast/ast_query.nim
```diff
@@ -586,12 +586,12 @@ proc skipStmtList*(n: PNode): PNode =
...
 proc isImportedException*(t: PType; conf: ConfigRef): bool =
   assert t != nil
-  if conf.exc != excCpp:
+  if conf.exc != excNative:
     return false
 
   let base = t.skipTypes({tyAlias, tyPtr, tyDistinct, tyGenericInst})
 
-  if base.sym != nil and {sfCompileToCpp, sfImportc} * base.sym.flags != {}:
+  if base.sym != nil and {sfImportc} * base.sym.flags != {}:
```
```suggestion
  if base.sym != nil and sfImportc in base.sym.flags:
```

compiler/backend/ccgtypes.nim
```diff
@@ -674,8 +616,7 @@ proc getTypeDescAux(m: BModule, origTyp: PType, check: var IntSet; kind: TSymKin
     return
   case t.kind
   of tyRef, tyPtr, tyVar, tyLent:
-    var star = if t.kind in {tyVar} and tfVarIsPtr notin origTyp.flags and
-                    compileToCpp(m): "&" else: "*"
+    var star = "*"
```
zerbina:
```suggestion
    let star = "*"
```
The `star[0] = '*'` in line 628 can also be removed.

compiler/sem/dfa.nim
```diff
@@ -709,6 +709,7 @@ proc isAnalysableFieldAccess*(orig: PNode; owner: PSym): bool =
   # wrong! Then the write barrier for the ``self.root`` assignment would
   # free the old data and all is lost! Lesson: Don't be too smart, trust the
   # lower level C++ optimizer to specialize this code.
+  # TODO: revisit if still relevant and how
```
zerbina: The documented problem is not related to C++. Also, for what it's worth, the new move analyser does exactly what is described to be a problem here and there are no issues so far.

tests/ccgbugs/tvar_param.nim
```diff
@@ -1,12 +1,16 @@
...
 """
 
 # These are tests for the fix for the regressions introduced by PR
 # https://github.com/nim-works/nimskull/pull/290.
 # Since temporaries are only generated with arc/orc (see
 # `ccgcalls.withTmpIfNeeded`) we explicitly build with `gc:arc` here
+#
+# xxx: These were introduced for the C++ backend originally and tested against
+#      the C backend for good measure. Keeping these around as they do a good
+#      job of explaining things. Perhaps they're redundant with everything else
```
zerbina: The test is redundant now.

zerbina:
Looks good to me. I left some minor suggestion that can also be applied separately.

I believe that the `ccThisCall` calling convention together with the associated bits can also be removed now.

saem:
Whoops, missed the comments, let me take a quick look at those.

tests/ccgbugs/tvar_param.nim
saem: Removing

compiler/backend/ccgtypes.nim
saem: updated

compiler/ast/ast_query.nim
saem: updated

PR #479, opened 2022-11-22
repo: use standard license template

haxscramper:


saem (changes requested):
This won't do, I'll look at it this weekend

PR #481, opened 2022-11-22
move discard, let, rationals to stmts/stdlib categories

haxscramper:


saem (changes requested):
Wait, why is there a binary added?

haxscramper:
ah right I guess that's why it fails, I somehow missed it

saem:
Can you check to see if binaries got added in other recent changes to tests you've made?

haxscramper:
there are no binaries in previous PRs

saem (approved):

haxscramper:
bors  r+

PR #484, opened 2022-11-28
Internal remove VM legacy reports

saem:
## Summary

This is the first part of removing legacy reports from the VM code base.
It does so by introducing VM and VM Code Generation Events, that are
used in lieu of 'reports'. This is just the start and much more work is
required prior to completion.

## Details

- Added `VmEvent` and `VmGenDiag` types to capture data, these are
  similar to but `VMReports` cruft, but are owned and defined by the VM
  in `vmdef`
- Source code location tracking (`TLineInfo`) is now minimized and much
  of this has been moved to the caller (`compilerbridge`, `vmrunner`,
  etc)
- Introduced `VmExecTrace` to handle code listing output, the type is
  at the very last more lightweight
- `TCtx` has a `vmExecTracer` callback to handle exec trace output
- use tracer callback in `compilerbridge`, `nimeval`, & `scriptconfig`
- Dropped `VMReport` and `DebugReport` dependencies from `vm`

### Compiler/Runner Bridging

- Introduced a somewhat simpler `VmStackTrace` and migrated
  `ExecErrorReport` to use that
- removed `opcNError`, `opcNWarning`, and `opcNHint`, these are now
  callbacks

### Miscellaneous

- Removed a number of unused import warnings
- Now fewer dependencies on `TCtx.debug`, hopefully easier to remove
- Fixed lots of code formatting issues

### Future Work

- Obviously get rid of the remaining reports cruft
- Sem/Ast reports clean-up, one reason why this clean-up can't done in
  one shot
- Events ease dropping the `TCtx.debug` dependency, finish this
- Slim down `VmGenDiag` data
- `YieldReason`/syscall interface to break remaining reports
  dependencies after Sem/Ast

---

## Notes for Reviewers
* Main goal is good enough to merge
* Legacy reporting is viral so need to stem the spread


saem:
Ugh, the code is so gross. I probably have to break things up a bit more. It's also not going to get much better with all the `SemReport` and other pollution.

I'll see if I can get the `vm` module clear of all the reports cruft, at least. Maybe that'll show a way out.

compiler/vm/vm.nim
```diff
@@ -921,20 +931,21 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): YieldReason
...
     updateRegsAlias
 
-  template guestValidate(cond: bool, msg: string) =
+  template guestValidate(cond: bool, pc: PrgCtr, strMsg: string) =
     ## Ensures that a guest-input related condition holds true and raises
     ## a `VmError` if it doesn't
     if unlikely(not cond): # Treat input violations as unlikely
       # TODO: don't use a string message here; use proper reports
-      raiseVmError(VMReport(kind: rvmErrInternal, str: msg))
+      raiseVmError(VmEvent(kind: vmEvtErrInternal, location: some pc, msg: strMsg))
```
saem: All this location setting should be wrapped up via a constructor.

compiler/vm/vm.nim
```diff
@@ -62,14 +62,14 @@ import
...
-    vmtypes
+    vmtypes,
+    vmlegacy
   ],
   experimental/[
     results
   ]
 
 # xxx: reports are a code smell meaning data types are misplaced
-from compiler/ast/reports_vm import VMReport
```
saem, line 72: One down three more to go.

compiler/vm/vmdef.nim
```diff
@@ -519,6 +580,143 @@ type
...
+        ast*: PNode
+      else:
+        discard
+    ra*: int
+    rb*: int
+    rc*: int
+
+  VmEvent* = object
+    ## Event data from a VM instance, mostly errors
+    ast*: PNode
```
saem: Should be able to move these fields into the appropriate branches now that we don't have a super object.

compiler/vm/vmgen.nim
```diff
@@ -2540,7 +2563,8 @@ proc gen(c: var TCtx; n: PNode; dest: var TDest; flags: TGenFlags = {}) =
       if s.magic != mNone:
         genMagic(c, n, dest, s.magic)
       elif s.kind == skMethod:
-        localReport(c.config, n.info, VMReport(kind: rvmCannotCallMethod, sym: s))
+        c.handleDiag(vmGenDiagCannotCallMethod, s, n.info)
+        # localReport(c.config, n.info, VMReport(kind: vmGenDiagCannotCallMethod, sym: s))
```
zerbina: You can turn this into a `fail`. `vmgen` doesn't and also shouldn't need to continue if an error occurs.

As far as I can tell, doing so would allow removing the `handleDiag` and `handleReport` procedures, plus the `vmlegacy` import.

compiler/vm/vmdef.nim
```diff
@@ -519,6 +580,143 @@ type
...
+    vmEvtTooManyIterations
+    vmEvtQuit
+    # debug / trace
+    vmEvtDebugCodeListing
+    vmEvtDebugExecTraceMinimal
+    vmEvtDebugExecTraceFull
+
+  VmEventKindAccessError* = range[vmEvtAccessOutOfBounds .. vmEvtAccessNoLocation]
+
+  VmDebugCodeEntry* = object
```
zerbina: You can move this type to `vm/vmutils.nim` and remove the `vmEvtDebugCodeListing` enum. A debug code listing is never reported as a `VmEvent`.

compiler/vm/vmdef.nim
```diff
@@ -519,6 +580,143 @@ type
...
+          pc: int,
+          opc: TOpcode,
+          info: TLineInfo,
+          ra, rb, rc: TRegisterKind
+        ]
+      of vmEvtDebugExecTraceMinimal:
+        vmgenExecMinimal*: tuple[
+          info: TLineInfo,
+          opc: TOpcode
+        ]
```
zerbina: I think it'd make sense to already split out the trace-related parts into a `VmTrace` (or similar), as they're not really events and also not reported from the VM as such.

compiler/vm/vm.nim
```diff
@@ -2624,7 +2645,8 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): YieldReason
       if a.kind == nkSym:
         regs[ra].intVal = ord(a.sym.kind)
       else:
-        raiseVmError(VMReport(kind: rvmNodeNotASymbol))
+        raiseVmError(VmEvent(kind: vmEvtNodeNotASymbol, location: some pc,
+                             srcLoc: source(c, pc)))
```
zerbina: Thanks for introducing and making use of `srcLoc`.

compiler/vm/vmdef.nim
```diff
@@ -519,6 +580,143 @@ type
...
+      else:
+        discard
+
+  VmStackTrace* = object
+    currentExceptionA*, currentExceptionB*: PNode
+    traceReason*: VmEventKind
```
zerbina, line 649: I don't think the stack-trace should store the trace reason, as a VM event is not always the reason for the creation of a stack-trace.

Ideally, it also shouldn't store both `currentExceptionA` and `currentExceptionB` (if I remember correctly, those are also never queried from the trace), but that is something that can be done as a later PR.

compiler/vm/compilerbridge.nim
```diff
@@ -152,19 +154,43 @@ proc unpackResult(res: sink ExecutionResult; config: ConfigRef, node: PNode): PN
...
+  ## Creates an `ExecErrorReport` with the `event` and a stack-trace for
+  ## `thread`
+  result.stackTrace = createLegacyStackTrace(c, thread, instLoc(-1))
+  let location =
+    case event.kind
+    of vmEvtUserError:
+      event.srcLoc
+    else:
+      event.location.map(proc (l: PrgCtr): TLineInfo = c.debug[l])
+                    .get(source(c, thread))
```
zerbina: The `TCtx` is captured via a full copy here (which is very expensive, as `TCtx` can potentially own a very large amount of data).

Until the compiler is able to capture by-view and put the environment on the stack in cases like the one here, I think it's better to only use `map`, `flatMap`, etc. when the output only depends on the parameter.

compiler/vm/compilerbridge.nim
```diff
@@ -152,19 +154,43 @@ proc unpackResult(res: sink ExecutionResult; config: ConfigRef, node: PNode): PN
...
+                    stacktrace: st.stacktrace,
+                    skipped: st.skipped,
+                    location: some source(c, thread),
+                    reportInst: toReportLineInfo(instLoc))
+
+proc buildError(c: TCtx, thread: VmThread, event: sink VMReport): ExecErrorReport  =
   ## Creates an `ExecErrorReport` with the `event` and a stack-trace for
   ## `thread`
-  result.stackTrace = createStackTrace(c, thread)
+  result.stackTrace = createLegacyStackTrace(c, thread, instLoc(-1))
```
zerbina: You can change the type of `ExecErrorReport.stackTrace` to `VmStackTrace` instead and then perform the translation in `unpackResult`. That moves the transition point further upwards.

compiler/vm/vm.nim
```diff
@@ -134,18 +134,22 @@ const
 const
   errIllegalConvFromXtoY = "illegal conversion from '$1' to '$2'"
 
+
+proc handleError*(c: TCtx, evt: VmEvent, instLoc = instLoc(-1)) {.inline.} =
+  c.config.handleReport(evt, instLoc)
+
```
zerbina: This seems to be the only reason for the `vmlegacy` import, but it doesn't look like it's used here. Maybe (if not use elsewhere) remove it or move it to `vmlegacy`?

compiler/vm/vm.nim
zerbina: You don't need to. Events only leave the VM through `YieldReason`, which is constructed in `execute`. Setting the location can be done in a single place there.

As discussed on Matrix, I'd prefer not storing `location` as part of `VmEvent`. The call-site of `execute` can query the `TLineInfo` or program counter of the instruction the event happened as part of via either `source(c, thread)` or `thread.currentPc`, and then associate this data with the event however it wants.

In the case of VM execution inside the compiler (i.e. `compilerbridge`), the location can be queried in the `yrkError` branch in `compilerbridge.execute` and then passed to `buildError` where it is either stored directly inside the `ExecErrorReport` or used for the `VMReport`.

compiler/vm/vmrunner.nim
```diff
@@ -275,16 +297,23 @@ proc main*(args: seq[string]): int =
...
-      c.config.localReport(createStackTrace(c, thread))
-      c.config.localReport(r.error)
+      c.config.localReport(createLegacyStackTrace(c, thread))
+      let location = 
+        case r.error.kind
+        of vmEvtUserError:
+          r.error.srcLoc
+        else:
+          r.error.location.map(proc (l: PrgCtr): TLineInfo = c.debug[l])
+                        .get(source(c, thread))
```
zerbina: Using `c` inside the anonymous procedure makes it part of the closure's environment (that is, the `TCtx` instance is now located on the heap instead of the stack). 

`c.debug[l]` always yields the same value as `source(c, thread)`, so you can simply use the latter directly.

compiler/vm/vmrunner.nim
```diff
@@ -225,6 +230,23 @@ proc registerCallbacks(c: var TCtx): bool =
       echo "expected '$#' callback but got '$#'" % [p.string, other[i].string]
       result = false
 
+proc createLegacyStackTrace(
+    c: TCtx, 
+    thread: VmThread, 
+    instLoc: InstantiationInfo = instLoc(-1)
+  ): VMReport =
+  # TODO: duplicated from `compilerbridge`, consolidate after all the reports
+  #       bits are untangled.
```
zerbina, line 239: The stack-trace is directly reported in the case of `vmrunner` -- maybe it makes sense to instead duplicate the formatting logic to here and not use `localReport`? This is only a suggestion, it's fine if it stays like it is for now.

compiler/vm/vmdef.nim
```diff
@@ -519,6 +580,143 @@ type
...
+    sym*: PSym
+    location*: TLineInfo        ## diagnostic location
+    instLoc*: InstantiationInfo ## instantiation in VM Gen's source
+    case kind*: VmGenDiagKind
+      of vmGenDiagCannotCast:
+        typeMismatch*: VmTypeMismatch
+      else:
+        discard
+
+  GenDiagHandler* = proc(c: TCtx, diag: VmGenDiag): void
```
zerbina: Is this type still needed?

compiler/vm/vmdef.nim
```diff
@@ -61,7 +66,63 @@ const
...
+
+      vmDiagTooManyIterations
+
+      # trace
+      vmDiagStackTrace
+
+
 type
+  PrgCtr* = int  ##gram Counter, aliased as it's self-documenting and support
+             ## changing the type in the future (distinct/width)
```
zerbina:
```suggestion
  PrgCtr* = int  ## Program Counter, aliased as it's self-documenting and supports
                 ## changing the type in the future (distinct/width)
```

zerbina (commented):
Thank you for working on this! The data for the VM-related reports not being located in a VM-related module felt off to me in the past, and I like that the data types are now closer to the logic producing instances of them.

I left some suggestions regarding parts that I think could and should be improved before merging.

compiler/vm/vmdef.nim
saem: Yeah, we're not maintaining them as a stream as in the lexer, so that makes sense.

compiler/vm/vmdef.nim
saem: I'll check for `traceReason` usage, I suspect it's redundant data.

compiler/vm/compilerbridge.nim
saem: D'oh, I forgot about that, thanks. I'll stick with if/else.

compiler/vm/compilerbridge.nim
saem: I did that earlier but went back given the dependency on thread for the program counter, etc. I'll give it another go soon because other things are cleaner already.

compiler/vm/vmdef.nim
zerbina: We eventually should maintain the traces as a stream, but that's something for a separate PR.

compiler/vm/vmrunner.nim
saem: Sigh, that's what I get for programming under the weather. Pretty much all the VM events will be raised and so their location and what's in the thread will be the same. Ooph

compiler/vm/vmdef.nim
saem: Unlikely, can remove shortly.

zerbina:
This is really starting to shape up, I like it a lot. Thank you!

compiler/vm/vmdef.nim
saem: Done, also noticed a bunch of simplification wrt how much data they need to carry around -- dropped more `TCtx.debug` access from the `vm` module. 🎉

compiler/vm/compilerbridge.nim
saem: done

compiler/vm/vm.nim
saem: removed.

compiler/vm/vmgen.nim
saem: Done

saem:
bors try

saem:
bors try

saem:
Bors r+

PR #485, opened 2022-11-28
vm: remove `errorFlag`

zerbina:
## Summary
The macro API procedures `parseExpr` and `parseStmt` can both fail, in 
which case they raise an exception. Only `opcRaise` can currently start
stack unwinding, so if an error occurred during parsing, the instructions
implementing the two procedures stored the error report in `errorFlag`.

The report is then turned into a string via the `internalErrorFlag`
(`opcQueryErrorFlag`) magic call and raised as an exception in user
code.

This complicates report-related refactoring, as the `errorFlag` field
introduces a dependency between the `vmdef` and `reports` module.

Instead, the `parseExpr` and `parseStmt` procedures now directly
return the error message (if one exists) via a `var string` parameter,
removing the need for both `opcQueryErrorFlag` and `errorFlag`.

## Details
- add an `err: var string` parameter to both `parseExpr` and `parseStmt`
- remove the `opcQueryErrorFlag` instruction
- remove the `internalErrorFlag` procedure
- move the logic for invoking the parser and handling parser errors into a
  separate procedure in `vmdeps`
- clean up the parser-error handling a bit


saem:
Cool, after this is merged the report can get reworked into diagnostics. Eventually, the macro API can just return nkError and that'll drop the exception from there as well.

This is much better, thanks.

haxscramper (approved):

PR #486, opened 2022-11-30
Testament reimplementation and test directory cleanup

haxscramper:
not fully ready for review but can be checked for port correctness from https://github.com/nim-works/nimskull/pull/135


testament/categories.nim
```diff
@@ -40,21 +40,22 @@ proc isTestFile*(file: string): bool =
 
 # --------------------- DLL generation tests ----------------------------------
 
-proc runBasicDLLTest(c, r: var TResults, cat: Category, options: string) =
+proc runBasicDLLTest(
```
saem, line 79: You can drop all the hot code reload testing.

testament/testament.nim
```diff
@@ -49,16 +66,31 @@ Options:
   --backendLogging:on|off      Disable or enable backend logging. By default turned on.
   --megatest:on|off            Enable or disable megatest. Default is on.
   --skipFrom:file              Read tests to skip from `file` - one test per line, # comments ignored
+  --includeKnownIssues         runs tests that are marked as known issues
+
+On Azure Pipelines, testament will also publish test results via Azure Pipelines' Test Management API
```
saem: Not sure if this still holds

testament/testament.nim
```diff
@@ -49,16 +66,31 @@ Options:
...
+provided that System.AccessToken is made available via the environment variable SYSTEM_ACCESSTOKEN.
 
 Experimental: using environment variable `NIM_TESTAMENT_REMOTE_NETWORKING=1` enables
 tests with remote networking (as in CI).
 """ % resultsFile
 
+proc isNimRepoTests(): bool =
```
saem: Testament is compiler only so I ripped this out, IIRC.

testament/testament.nim
```diff
@@ -105,16 +282,21 @@ type
...
   TestFilter = object
     case kind: TestFilterKind
     of tfkAll, tfkHtml, tfkCache:
       discard
 
     of tfkCategories, tfkPCats:
-      # xxx: currently multiple categories are unsupported
+      # FIXME: currently multiple categories are unsupported
       cats: Categories
+    of tfkGlob:
```
saem, line 296: Glob is a terrible buggy mess

testament/testament.nim
```diff
@@ -897,28 +1093,65 @@ func extraOptions(run: TestRun): string =
...
+    exeExt =
+      if isJsTarget:
+        "js"
+      else:
+        ExeExt
+
+    fileDir = if dirPart.isAbsolute():
+                dirPart
+
+              else:
```
saem: No spacing/breaking apart branches like this the if and the else go together forming one expression for the assignment, there must not be any visual separation as it's not communicating the correct thing.

testament/testament.nim
```diff
@@ -897,28 +1093,65 @@ func extraOptions(run: TestRun): string =
...
+      (if matrixEntry == noMatrixEntry:
+        @[specName, target.cmd]
+      else:
+        @[specName, $matrixEntry, target.cmd]).join("_")
+    exeExt =
+      if isJsTarget:
+        "js"
+      else:
+        ExeExt
+
```
saem, line 546: This is one or block, extra spaces without rhyme or reason just make the code harder to read as it breaks flow, it's one continuous series of let definitions.

testament/testament.nim
```diff
@@ -1196,7 +1433,7 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
     execState.filter = TestFilter(kind: tfkAll)
   of "c", "cat", "category":
     # only specified category
-    # xxx: consider removing pcat concept or make parallel the default
+    # HACK consider removing pcat concept or make parallel the default
```
saem: It's a consideration or something to think about not a hack, the semantics shifted slightly with megatest gone.

testament/testament.nim
```diff
@@ -1220,155 +1457,861 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
 
   execState.userTestOptions = p.cmdLineRest
 
-proc main() =
-  ## Define CLI Options/Args Parsing loops
-  var
-    execState = Execution(flags: {outputColour, logBackend})
-    p = initOptParser()
+if paramCount() == 0:
```
saem: Imperative code arbitrarily placed in the middle of a series of definitions is harder to follow, leave the main in our retractor further so it's all collected once and at the end. The former is more preferable.

testament/testament.nim
```diff
@@ -1220,155 +1457,861 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
...
+  defaultCatId: CategoryId = 0
+
+func requestedTargets(execState: Execution): set[TTarget] =
+  ## get the requested targets by the user or the defaults
+  if execState.targets == noTargetsSpecified:
+    {targetC, targetJS}
+  else:
+    execState.targets
+
+func `<`(a, b: TestFile): bool {.inline.} =
```
saem: It's weird with these not having consistent spacing with its neighbours. Rather than fix the spacing the real takeaway is to move it closer to the type definition.

testament/testament.nim
```diff
@@ -1220,155 +1457,861 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
...
+  a.file < b.file
+func cmp(a, b: TestFile): int {.inline.} =
+  cmp(a.file, b.file)
+
+proc prepareTestFilesAndSpecs(execState: var Execution) =
+  ## for the filters specified load all the specs
+  # IMPLEMENT create specific type to avoid accidental mutation
+
+  # NOTE read-only state in let to avoid mutation, put into types
```
saem, line 607: This is more of a todo

testament/testament.nim
```diff
@@ -1220,155 +1457,861 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
...
+      nimcache = nimcacheDir(testFile.file, matrixOptions & testArgs, target)
+      compileCmd = prepareTestCompileCmd(
+        cmdTemplate = spec.getCmd,
+        filename = testFile.file,
+        options = testArgs,
+        nimcache = nimcache,
+        target = target,
+        extraOptions = matrixOptions,
+        outfile = testRun.exeFile(testFile.file, execState.rootDir))
+
```
saem: Continuous block or lets, the i visual break nterruptions had me checking to see if it's local assignment.

testament/testament.nim
```diff
@@ -1029,7 +1262,7 @@ func defaultTargets(category: Category): set[TTarget] =
   else:
     standardTargets
 
-proc testSpec(r: var TResults, test: TTest) =
+proc testSpec(r: var TResults, test: TTest, execution: Execution) =
```
saem: Not sure what I was thinking weaving through another parameter ('execution') rather than combining it with existing ones. Sigh... sorry.

saem (commented):
Did a quick scan left some comments, it's already coming along really nicely.

testament/testament.nim
haxscramper: So I can also drop it?

saem: Yes, please!

Burn that stuff to the ground.

testament/testament.nim
haxscramper: Ok, I just fixed the woefully undescriptive `xxx:` tags and just misinterpreted this one a bit.

testament/testament.nim
haxscramper: I think this one came from the rebase, don't remember writing it explicitly. I will move it out later then

testament/testament.nim
saem: Sometimes the vagueness of xxx is an asset, often a curse.

testament/testament.nim
saem: I hope I didn't do that or I need to have a conversation with myself.

testament/testament.nim
haxscramper: I will break it into smaller `let` section

testament/testament.nim
haxscramper: should I merge these two? I had a feeling they are effectively the same (state object) but wasn't exactly sure. 

There are a few moments still where I need to bring in `TResults` in as well

testament/testament.nim
haxscramper: Well, not *directly*, more like a leftover https://github.com/nim-works/nimskull/pull/135/files#diff-a161d4667e86146f2f8003f08f765b8d9580ae92ec5fb6679c80c07a5310a551L1248 from the earlier times

testament/testament.nim
saem: I could be really tired and that's yet another issue...

My sleepy brain is thinking (or not) that Execution makes sense as a first parameter. Then a second one to identify the test makes sense and that should be immutable data. Then finally result which can be separate to read only and writeable data. Issue with dod is when you just get this super blob that any proc can do anything to at any time.

testament/testament.nim
saem: Thanks, I feel less bad. 😁

testament/testament.nim
saem: ### My Adivce

If you see a pattern that simplifies across the whole test loop then go for it, otherwise I'd recommend preserving the style:

1. establish facts/read queries
2. if necessary setup local mutables
3. do whatever logic and write out any updates
    - aim to mutate a 'row' of state only in one place (single writer)

### Philosphicaly

I recommend keeping it together, I know it's a smelly, but it's good to keep smelly code smelly. It's counter-intuitive, but the way we simplify the code is by thinking through these lets, what they're trying to do, if there is simplification or not. By keeping it smelly we keep the code honest, don't let it fake a clean look.

### My Intention at the Time

They're all acting as sort of facts/read queries of the state. I was aiming to figure out some Command Query Responsibility Segregation type pattern. Where reads (these queries) and the mutable bits (commands later one) are segregated.

With enough of these let blocks identified across various parts of the code, I was thinking it'd be possible to look across them for common factors. Then see if they're worth pulling together into some other abstraction, my hunch is was it might end up like this:

- have the exec state object similar to today
- have some query procs for common queries
- mutation for any given piece of state should only happen once per overall loop in one spot

But the precise decomposition I was waiting to see more before running ahead with it.

testament/specs.nim
```diff
@@ -106,7 +106,9 @@ type
     ccodeCheck*: seq[string] ## List of peg patterns that need to be
     ## searched for in the generated code. Used for backend code testing.
     maxCodeSize*: int ## Maximum allowed code size (in bytes) for the test.
-    err*: TResultEnum
+    err*: TResultEnum # REFACTOR Separate into 'expected output' (they can
```
haxscramper: test review on the comment

haxscramper:
image

Top-level logic for now if anyone is interested

testament/testament.nim
```diff
@@ -2005,6 +2063,11 @@ proc runTests(execState: var Execution) =
     nextCmdIdToActId = newSeqOfCap[int](batchSize)
     nextCmdIdToActKind = newSeqOfCap[TTestAction](batchSize)
     nextCmdIdToInput = newSeqOfCap[Option[string]](batchSize)
+    # QUESTION ASSUME I assume thread war is added here because of the
```
saem: IIRC dropping `threadvars` made the compiler grumpy.

testament/testament.nim
```diff
@@ -2017,6 +2080,17 @@ proc runTests(execState: var Execution) =
   startTimes = newSeq[float](batchSize)
   endTimes = newSeq[float](batchSize)
 
+  # REFACTOR: this function requires additional context to be available,
```
saem: making the callbacks top level module procs also doesn't help as it's not really part of the public surface area and only ever used here.

testament/testament.nim
```diff
@@ -2017,6 +2080,17 @@ proc runTests(execState: var Execution) =
...
+  # making it harder to move out of body. It mutates some of the state as
+  # well, so one possible solution would be to make the surrounding state
+  # of the `runTest` function *explicit* -- create a private
+  # `RunTestContext` ref object and then create a `initOnTestRunStartCb`
+  # which takes a context and returns a closure.
+  #
+  # NOTE: This approach might also address the [[mutation]] question above,
```
saem: Hmm, I think capturing the test result data (things that the callbacks mutate) in a temporary spot, like a `TestBatch` or whatever, and then integrating it back into the main execution state as a separate step might be ok. Though, I worry it might simply be more code where things can go wrong and not really buying us anything.

testament/testament.nim
haxscramper: I'm thinking about using

```nim
proc createBatchExecutionTriggers(batchStore: BatchStore, batchId: BatchId): tuple[onStart: proc(id: int), onProcess, onComplete: proc(id: int, p: Process)] = 
  result.onProcess = proc(id: int) = 
    batchStore[batchId].startTimes[id] = epochTime()
    # ... etc
```

testament/testament.nim
haxscramper: I don't think there is an explanation in terms of some higher-level abstractions, but when I see the 15-argument with half arguments mutable it feels like a code smell

testament/specs.nim
saem: This should only be one field on the `TSpec`, which is the expected result. The actual result shouldn't be on the spec at all, that's part of the abuse.

haxscramper: Things like `proc codegenCheck(` which accept `given: var TSpec` and then mutate fields will be replaced with report constructors [`88ec8a8` (#486)](https://github.com/nim-works/nimskull/pull/486/commits/88ec8a8a4140abc87b52d6e700504c21c08126f7#diff-a161d4667e86146f2f8003f08f765b8d9580ae92ec5fb6679c80c07a5310a551L1023)

testament/testament.nim
```diff
@@ -1001,7 +1026,13 @@ proc codegenCheck(
   ) =
   ## Check for any codegen mismatches in file generated from `test` run.
   ## Only file that was immediately generated is tested.
+  # REFACTOR convert into procedure that creates `CodegenCheckResult`
```
haxscramper: `CodegenCheckResult` or some other kind of report object -- anything aside from direct mutation of the specification

testament/testament.nim
```diff
@@ -506,7 +699,13 @@ proc logToBackend(param: ReportParams) =
 
 proc addResult(r: var TResults, param: ReportParams, givenSpec: ptr TSpec) =
   ## Report results to backend, end user (write to command-line) and so on.
-  
+
+  # REFACTOR: first of all, name: it does not /add/ it fucking *writes*
+  # output. "write" and "add" are two different things. Second: determine
+  # why `givenSpec` is an optional in the first place -- aren't we printing
+  # out results of the test? And each test has a spec attached to it.
```
saem: I remember this now... the issue is that `addResult` is more like add whatever stuff happened, and occassionally, that doesn't involve a test/spec... I can't quite find the call site(s) at the moment.

testament/testament.nim
```diff
@@ -515,7 +714,8 @@ proc addResult(r: var TResults, param: ReportParams, givenSpec: ptr TSpec) =
 
   logToBackend(param)
 
-  # TODO DOC what is this
+  # REFACTOR what is this, no string interpolation until the data is
```
saem: test results aren't the only thing to note/log, so it's not that straightfoward.

testament/testament.nim
```diff
@@ -821,14 +1048,21 @@ proc codegenCheck(
...
   except IOError:
     given.err = reCodeNotFound
     msg Undefined: getCurrentExceptionMsg()
 
-proc compilerOutputTests(run: var TestRun, given: var TSpec; r: var TResults) =
-  ## Test output of the compiler for correctness
+proc compilerOutputTests(run: TestRun, given: var TSpec; r: var TResults) =
+  ## Test output of the compiler for correctness and add comparison results
+  ## to the `TResults` output.
+  # REFACTOR into a functio that returns result of the output test
```
saem: This could go either way, because some tests do a compile and run vs just a compile. Accumulating results while also supporting the latter in the same structure vs separate structures and stitching them back together.

testament/testament.nim
```diff
@@ -1220,155 +1497,908 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
...
+      let optMatrix = execState.testOpts[testId].optMatrix
+      execState.testSpecs[testId].matrix =
+        case execState.testSpecs[testId].matrix.len
+        of 0:
+          optMatrix
+        else:
+          # (saem)REFACTOR - this is a hack, we shouldn't modify the
+          # spec.matrix
+
+          # (haxscramper)QUESTION Doesn't "specify additional test matrix
```
saem, line 713: The assignment  to `execState.testSpecs[testId].matrix` is a bad idea, we should keep the spec intact.

Ideally, a spec is just what was declared in the file, actually should be called `FileSpec`, `DeclaredSpec`, or something. Then we create actual specs that are truly atomic specs, so a declared spec should generate 1 or more actual specs given matrix options, target, etc declared data.

testament/testament.nim
haxscramper: image

I want to do batches/logging part like this

testament/testament.nim
saem: this is an acceptable compromise for a good while, but it'll eventually run into limits when we want to support multiple specs per file, more complex matricies, and or more complex test steps/orchestration.

testament/testament.nim
haxscramper: and to be honest I couldn't really understand what `.data` does, I will look into it more, maybe this will throw a wrench in this data flow model

testament/testament.nim
```diff
@@ -1220,155 +1497,908 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
...
+  let
+    target = testRun.target
+    matrixEntry = testRun.matrixEntry
+  result = test.file.changeFileExt("").replace(DirSep, '/')
+  result.add '_' & $target
+  if matrixEntry != noMatrixEntry:
+    result.add "[$1]" % $matrixEntry
+  if allowFailure:
+    result.add " (allowed to fail) "
+  # if test.options.len > 0:
```
saem: this maybe an issue wrt to test name uniqueness (parallelism).

testament/testament.nim
haxscramper: Ok, so this is mostly an issue of "MatrixAsSpecfied" vs "MatrixAsActuallyUsed"; got it. `DeclaredSpec` seems like a good name for it

testament/testament.nim
haxscramper: Weird, don't remember removing it; options certainly need to be added as a part of name

testament/testament.nim
```diff
@@ -1220,155 +1497,908 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
...
-    isMainProcess   = true
-    skipFrom        = ""
-  
-  ## First parse options
-  case parseOpts(execState, p)
-  of parseQuitWithUsage: quit Usage
-  of parseSuccess: discard
+    # REFACTOR test run should contianer ID of the specification it was
+    # used for as well as information about it's matrix parameters. The ID
+    # should be used instead of this copy.
```
```suggestion
    # REFACTOR test run should contian the ID of the specification it's
    # associated with as well as information about its matrix parameters.
    # The ID should be used instead of this copy.
```

saem: Although, that might not be enough and it might have to be at the action level.

testament/testament.nim
```diff
@@ -1220,155 +1497,908 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
...
+      # REFACTOR this is an absolute abomination of an implementation, it
+      # should be refactored into something that uses 3-4 arguments, not
+      # fucking 15.
+      #
+      # IDEA a possible solution would be to introduce a "TestBatch" object
+      # which is populated with the required procedures. In this case part
+      # of the function above could also me moved elsewhere and turned into
+      # constructor-like procedure.
+      #
+      # NOTE each batch contains a fair bit of mutable state that gets
```
saem: Pre-batching isn't so easy as you have to shuffle things between batches all the time as you're executing, there is a reason why it works more like a sliding window right now.

The code creates the runnable batch and assumed next batch. Then after seeing the result of the current bacth it updates the assumed next batch, plus any related reporting, to do things like drop dependent actions that are no longer valid.

Batch might not be the right word to use universally and so it's giving the wrong impression, it's more like window of test runs and a window of current actions batch and potential next actions batch.

testament/testament.nim
```diff
@@ -1220,155 +1497,908 @@ proc parseArgs(execState: var Execution, p: var OptParser): ParseCliResult =
...
+    compEffortTotal = compEffortTotal + compEffort
+    effort = effort + compEffort + rt.runEnd - rt.runStart
+    compDurLow = min(compDurLow, compEffort)
+    compDurHigh = max(compDurHigh, compEffort)
+
+  let
+    elapsed = latest - earliest
+    avgCompTime = compEffortTotal / float execState.runTimes.len
+
+  # REFACTOR structured output etc; format using standalone procs --
```
saem: this is mostly debug code, but also generally useful.

haxscramper:
Current state of the implementation: Main data processing pipeline is working as outlined in the big flowchart several posts up. Read files, parse specs, create actions. There are several minor implementation bugs that fail various specs. 

Special categories caused some issues due to their very out-of-band nature, but with the current architecture run/action producer does not have to be very tightly integrated in the pipeline and instead can feed the data in smaller chunks that are processed immediately. Create a small "batch" consisting of a single command, process it, parse results. This way hardcoded special categories are just a series of batches that are passed to the program in strict sequential order.

haxscramper:
Conflicts started to pile up and at the moment I'm not sure I can allocate enough time to properly continue this PR so I'm closing it for the time being in order to avoid confusing about what is being worked on.

PR #487, opened 2022-12-05
semexprs: improve `semYieldVarResult`

zerbina:
## Summary
- replace `localReport` with `nkError` in `takeImplicitAddr`
- make `semYieldVarResult` `nkError` aware
- fix the `nkHiddenAddr` produced by `takeImplicitAddr` for tuple
  elements having the wrong type (i.e. `tyPtr`)
- use a dedicated error report for when a literal tuple constructor is
  expected, instead of the rather meaningless `illformed AST` error

## Details
- add a test for the introduced report
- improve documentation

In addition, a `nkHiddenAddr(nkDerefExpr(x))` expression is no longer
collapsed to just `x`. Doing so is unsound and produces expressions of
the wrong type, which causes problems for the MIR. In general, such
transformations/optimizations are better done in the backend.

The aforementioned change is also visible to macros operating on typed
AST:
```nim
proc p(x: ptr int): var int =
  result = x[]
  # the AST of the right-hand side is now:
  # ``nnkHiddenAddr(nnkDerefExpr(nnkSym))``
  # instead of:
  # ``nnkSym``

iterator iter(x: ptr int): var int =
  yield x[]
  # the AST differs in the same way as for the `result` assignment.

iterator iter2(x: ptr int): (int, var int) =
  yield (0, x[]) # the AST for `x[]` is also affected in the same way
```

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1469,6 +1469,13 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemUnexpectedYield:
       result = "'yield' only allowed in an iterator"
 
+    of rsemYieldExpectedTupleConstr:
+      # TODO: make the error message more helpful by also showing which
+      #       elements are of view type
+      result = "literal tuple constructor expected for 'yield', as the " &
+               "iterator returns a tuple directly containing a view. Got: " &
+               r.ast.render
+
     of rsemCannotReturnTypeless:
       result = "current routine cannot return an expression"
 
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -1892,7 +1892,7 @@ proc propertyWriteAccess(c: PContext, n, a: PNode): PNode =
     #fixAbstractType(c, result)
     #analyseIfAddressTakenInCall(c, result)
 
-proc takeImplicitAddr(c: PContext, n: PNode; isLent: bool): PNode =
+proc takeImplicitAddr(c: PContext, formal: PType, n: PNode): PNode =
   ## See RFC #7373, calls returning 'var T' are assumed to
   ## return a view into the first argument (if there is one):
   addInNimDebugUtils(c.config, "takeImplicitAddr", n, result)
@@ -1900,40 +1900,37 @@ proc takeImplicitAddr(c: PContext, n: PNode; isLent: bool): PNode =
   let root = exprRoot(n)
   if root != nil and root.owner == c.p.owner:
     if root.kind in {skLet, skVar, skTemp} and sfGlobal notin root.flags:
-      localReport(c.config, n.info, reportSym(
+      return newError(c.config, n, reportSym(
         rsemLocalEscapesStackFrame, root, ast = n))
 
     elif root.kind == skParam and root.position != 0:
-      localReport(c.config, n.info, reportSym(
+      return newError(c.config, n, reportSym(
         rsemImplicitAddrIsNotFirstParam, root, ast = n))
 
-  case n.kind
-  of nkHiddenAddr, nkAddr: return n
-  of nkDerefExpr: return n[0]
-  of nkBracketExpr:
-    if n.len == 1: return n[0]
-  of nkHiddenDeref:
-    # issue #13848
-    # `proc fun(a: var int): var int = a`
-    discard
-  else: discard
-  
-  case isAssignable(c, n, isLent)
+  # TODO: what about expressions where ``root == nil`` (e.g. the root is a
+  #       pointer dereference)?
+
+  case isAssignable(c, n, formal.kind == tyLent)
   of arLValue:
     discard
   of arLocalLValue:
-    localReport(c.config, n, reportSem rsemLocalEscapesStackFrame)
+    # FIXME: this is not entirely correct. If inside an iterator (both inline
+    #        and closure) and yielding a local as a ``var|lent``, it doesn't
+    #        escape its stack frame. Yielding a local as a view is not an
+    #        error
+    return newError(c.config, n, reportSem rsemLocalEscapesStackFrame)
   else:
-    localReport(c.config, n, reportSem rsemExprHasNoAddress)
+    return newError(c.config, n, reportSem rsemExprHasNoAddress)
 
-  let typ =
-    case n.typ.kind
-    of {tyVar, tyLent}:
-      n.typ
+  if n.kind == nkHiddenAddr:
+    if sameType(formal, n.typ):
+      result = n
     else:
-      makePtrType(c, n.typ)
-  result = newNodeIT(nkHiddenAddr, n.info, typ)
-  result.add(n)
+      # the input is already a view, but it is of different type. While
+      # correctable, this shouldn't happen in the first place
+      result = newError(c.config, n, reportSem rsemExprHasNoAddress)
+  else:
+    result = newTreeIT(nkHiddenAddr, n.info, formal): n
 
 proc asgnToResultVar(c: PContext, n, le, ri: PNode) {.inline.} =
   # refactor: to return a PNode instead and handle nkError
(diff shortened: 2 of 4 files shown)


compiler/sem/semexprs.nim
```diff
@@ -2141,55 +2137,78 @@ proc semProcBody(c: PContext, n: PNode): PNode =
...
+    result.add takeImplicitAddr(c, t, unwrappedValue)
+    hasError = result[0].isError
   of tyTuple:
-    var illformedAst = false
-      ## in some cases syntax issues that we can diagnose precisely enough
-      ## occur, this happens often with tuples (parentheses) so we use this to
-      ## flag if we end up in such a scenario
+    var
+      notTupleConstr = false
+        # if the tuple contains a view (i.e. ``var|lent``), the yield
```
haxscramper: This comment could be replicated to the documentation of the error kind enum added earlier

compiler/front/cli_reporter.nim
```diff
@@ -1469,6 +1469,10 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemUnexpectedYield:
       result = "'yield' only allowed in an iterator"
 
+    of rsemYieldExpectedTupleConstr:
+      result = "tuple constructor expected for iterator returning a tuple" &
```
haxscramper: "Literal tuple constructor (x, y, z, ...) expected for 'yield' as iterator returns a tuple type which contains a view"

Additional information might include
- annotation showing which part of the tuple is a view type
- what was found instead of a tuple ("but found procedure with a tuple return value")

Rephrasing is optional, the main point is to (1) provide at least some simplistic hint to fix, (2) show the language construct (yield) which causes the local issue, (3) more precise description of what is needed (**literal** tuple if I understand the error message correctly)

Bullet points are just TODO for later, feel free to just copy-paste as literal TODO in the code, don't bother with this for now.

compiler/sem/semexprs.nim
```diff
@@ -1892,48 +1892,44 @@ proc propertyWriteAccess(c: PContext, n, a: PNode): PNode =
...
-  of nkBracketExpr:
-    if n.len == 1: return n[0]
-  of nkHiddenDeref:
-    # issue #13848
-    # `proc fun(a: var int): var int = a`
-    discard
-  else: discard
-  
-  case isAssignable(c, n, isLent)
+  # XXX: what about expressions where ``root == nil`` (e.g. the root is a
```
haxscramper: Its my personal preference but I like to use words such as QUESTION, TODO, NOTE, BUG, HACK and similar in the code as (1) a lot of tools can differentiate between these types, (2) it is easier to search and (3) you can see whether it is a QUESTION or a TODO without drilling down into the text, pulling out extra code context and so on.

tests/errmsgs/tyield_view_tuple.nim
```diff
@@ -0,0 +1,22 @@
+discard """
+  nimoutFormat: sexp
+  cmd: "nim check --msgFormat=sexp --filenames=canonical --hints:off $options $file"
+  action: reject
+"""
+
+proc make(x: var int): (int, var int) =
+  (0, x)
```
haxscramper: "this code is ok because it returns a tuple constructed from the literal"

tests/errmsgs/tyield_view_tuple.nim
```diff
@@ -0,0 +1,22 @@
...
+
+iterator iter(): (int, var int) =
+  var x = 0
+  yield make(x) #[tt.Error
```
haxscramper: \# this code fails to compile because the expression is not a tuple constructor literal

compiler/sem/semexprs.nim
```diff
@@ -1892,48 +1892,44 @@ proc propertyWriteAccess(c: PContext, n, a: PNode): PNode =
...
+  #      pointer dereference)? Those should also be rejected (because they're
+  #      not a view into the first argument)
+
+  case isAssignable(c, n, formal.kind == tyLent)
   of arLValue:
     discard
   of arLocalLValue:
-    localReport(c.config, n, reportSem rsemLocalEscapesStackFrame)
+    # XXX: this is not entirely correct. If inside an iterator and yielding
```
haxscramper: something that implies FIXME or this is a NOTE/HACK for further code readers?

compiler/sem/semexprs.nim
```diff
@@ -2141,55 +2137,78 @@ proc semProcBody(c: PContext, n: PNode): PNode =
...
   of tyTuple:
-    var illformedAst = false
-      ## in some cases syntax issues that we can diagnose precisely enough
-      ## occur, this happens often with tuples (parentheses) so we use this to
-      ## flag if we end up in such a scenario
+    var
+      notTupleConstr = false
+        # if the tuple contains a view (i.e. ``var|lent``), the yield
+        # expression *must* be a tuple constructor
+        # XXX: this doesn't seem right in the presence of first-class view
```
haxscramper: Same as the previous: something that needs to be fixed or just a general observation of the weirdness of this code? Or maybe it is a QUESTION about possible suggestions for refactoring (again, note for latter readers)

haxscramper (approved):
You can ignore the comments on more precise admonitions if you want, the error message should preferably at least mention that *literal* constructor is preferred or maybe show (x,y,z,...) somewhere - a bit more hints for the user would be nice. 

Those are general recommendations for rephrasing, implementation-wise I have no objections/suggestions, looks good to me, so approving the merge, you can pick any pieces of reviews you deem appropriate

compiler/sem/semexprs.nim
saem: The issue with said words is it's also really painful to search/find them across the codebase. Additionally, they can be a combination of said things.

compiler/sem/semexprs.nim
```diff
@@ -1892,48 +1892,44 @@ proc propertyWriteAccess(c: PContext, n, a: PNode): PNode =
...
+proc takeImplicitAddr(c: PContext, formal: PType, n: PNode): PNode =
   ## See RFC #7373, calls returning 'var T' are assumed to
   ## return a view into the first argument (if there is one):
   addInNimDebugUtils(c.config, "takeImplicitAddr", n, result)
 
   let root = exprRoot(n)
   if root != nil and root.owner == c.p.owner:
     if root.kind in {skLet, skVar, skTemp} and sfGlobal notin root.flags:
-      localReport(c.config, n.info, reportSym(
+      return newError(c.config, n, reportSym(
```
saem, line 1903: Yay, more nkError, thank you!

saem (approved):
This was an excellent find, change looks good.

saem:
I'm inclined to bors this. The only reason I'm not is in case you wanted to tweak comments, @zerbina. Otherwise this is awesome and good to go.

zerbina:
> I'm inclined to bors this. The only reason I'm not is in case you wanted to tweak comments, @zerbina. Otherwise this is awesome and good to go.

Yep, I still want to tweak the comments -- only haven't done so because I wanted to first publish the PRs I had already started, so that I can focus on one thing at a time.

compiler/sem/semexprs.nim
haxscramper: and the issue with `xxx` is it's also really painful to search/find a specific ones out of them in the codebase because they are all uniform. 

I haven't thought about that before but `XXX NOTE` is also a thing and considering the comment's body is likely to dwarf any admonition in size it is not such a big hindrance

disruptek:
Can you add a note here with some hints as to the typed macro ast changes?

If not, no big deal, we should be able to figure it out quickly enough.

tests/errmsgs/tyield_view_tuple.nim
zerbina: The `make` procedure is only a helper and not related to the test. I added a doc comment for clarification.

zerbina:
I restructured the `semYieldVarResult` tuple branch in order to make it a bit easier to follow, changed the `XXX` annotations into `TODO`s or `FIXME`s, and also improved the error report's message according to @haxscramper 's suggestions.

> Can you add a note here with some hints as to the typed macro ast changes?

Yep, done. I added a listing with the changes to the PR message.

bors r+

PR #490, opened 2022-12-06
compiler: introduce a mid-end IR (MIR)

zerbina:
Summary
=======

This PR introduces and integrates a **m**id-end **i**ntermediate
**r**epresentation (MIR) into the compiler. It is an IR for code and
meant to host target-independent transformations, analysis, and
lowerings, and is the first step towards implementing the plan detailed
in https://github.com/nim-works/nimskull/discussions/467.

Overview
--------

Before being processed by the code-generators, the `PNode` AST is
first passed to `mirgen`, which translates the AST into MIR code. In
the future, the MIR passes will take place after this step, but there
currently aren't any. After all eventual transformations are applied,
the MIR code is translated back into `PNode` AST. Rewriting the
code-generators to operate on the MIR directly is not feasible (at
least not as part of this commit), so the aforementioned translation is
required.

Until a proper pass management is implemented, `mirbridge` combines the
`mirgen`/`astgen` step into a single procedure, so that the logic can
be reused across all code-generators.

Passing the result of the MIR code -> `PNode`-AST translation to the
`injectdestructors` pass makes the latter produce code that fails the
current tests, so as a temporary workaround, the MIR processing is not
run at all for code for which the `injectdestructors` pass is enabled.

Lessons learned / Takeaways
===========================

`mirgen` has a similar structure to the other code-generators, in that
it also uses `genX` procedures that recursively call each other. It is
much more strict however, i.e. instead of using the top-level
dispatcher procedure everywhere, knowledge about the context is used in
order to call the correct procedure directly. For example, when
processing call arguments, the `gen` procedure for *expressions* is
called directly, as arguments are always (except for an explicitly
handled special-case) expressions. The consequences:
- more decisions happening at *compile-time* instead of at *run-time*.
  Decision making happens earlier -- it is moved outwards
- `mirgen` acts as a small validation layer for the input AST.
  Ill-formed or semantically invalid AST is (in some capacity) detected
  as a side-effect of how `mirgen` operates
- it is easier to reason about what AST forms are valid by only looking
  at the code (i.e. `mirgen` acts as a good overview of what is valid
  and what is not)

This helped uncovering a good amount of issues:
- the unsound collapsing of `HiddenAddr (HiddenDeref x)` node sequences
- `HiddenAddr` nodes having the wrong type in some cases
- when evaluating whether a `concept` applies and the last expression
  in the concept's body is the result of a template expansion, a
  `nkStmtListExpr` with expressions of which the result is not
  discarded is passed to `semConstExpr` (`vmgen` ignored this, but
  `mirgen` doesn't)
- when inserting boolean literals, `closureiters` used the symbol of
  the `bool` enum fields instead of integer literals. Past
  early-`transf` (where the `closureiters` transformation takes place)
  this is illegal, and only didn't fail code-generation because all
  code-generators still handled `skEnumField` symbols for robustness

Other findings:
- semantic analysis uses the VM to evaluate type expression in a lot of
  cases. For example, the expressions `static int` or `T not nil` are
  either partially or fully evaluated with the VM. The main work is
  done by `vmgen` here: an expression that has a `typeDesc` type is
  turned into a `nkType` node constant -- the VM only loads this
  constant into the result register and immediately exits again
- `vmgen` is being used by sem to detect and reject expressions
  containing unresolved generic parameters
- `closureiters` implements the same complex-expression-to-statement
  transformation as `mirgen`. Moving the `closureiters` transformation
  to operate on the MIR will likely simplify the logic there by a lot
- there are almost not tests for lifetime tracking hooks with the JS
  target
- cursor inference or the `injectdestructors` pass don't work with the
  output of `astgen`

MIR design
----------

What turned out to be a very good decision, was to separate the scoping
behaviour (in terms of both lifetime and visibility) from control-flow
constructs. That is, to make constructs like `if` (`mnkIf`) or `case`
(`mnkCase`) not affect the lifetime or visibility of definitions. This
simplified a lot of logic, made things clearer, and made important MIR
code possible that could not be represented before. The takeaway: if
there's no concrete reason to do so, don't combine/conflate unrelated
behaviour into a single construct.

Details
=======

This commit is for the most part self-contained, but multiple bugs and
issues in unrelated modules had to be fixed in order for `mirgen` to
work:
- `closureiters` now correctly emits `nkIntLit` nodes instead of
  `skEnumField` symbols for boolean literals
- `closureiters` now sets the type of a `case` expression to `nil`
  when transforming it into a statement
- cursor inference is separated from the `injectdestructors` pass.
  This is necessary in order for the MIR integration to happen
  in-between both
- prevent semantically invalid `nkStmtListExpr` AST from being
  evaluated with the VM during `concept` application

**General fixes:**
Fix `vmgen` raising an assertion for `x.len` where `x` was not directly
a string (e.g. a `sink string` or alias to `string`).

**Changes to `vmgen`:**
- since `mirgen` now handles type expressions, `vmgen` no longer has
  to. Only `nkType` (type literals) are code-gen'ed, everything else
  type related is rejected (an error is raised)
- the special handling for compile-time-only call arguments is
  removed - `mirgen` handles this now
- magics for which the instruction is selected based on the symbol
  name use exhaustive case statements now. This prevents magic symbol
  having an unexpected name from going unnoticed

**Other additions:**
- add a facility (`OptIndex`) for the ad-hoc creation of nilable
  versions of index-like types to the `dod_helpers` module
- add a simple half-open slice implementation to the `idioms` module
- add branch-related iterators to `ast_query`

**Added modules:**
- `mirtrees` contains the type definitions that make up the MIR, along
  with basic tree traversal routines
- `mirchangesets` implements the `Changeset` API, which is the primary
  way of modifying MIR code
- `mirconstr` contains some routines useful when generating MIR
  constructs. It's in an early stage of development, and the goal is to
  eventually generalize the mini DSL used by `mirgen` and move it here
- `sourcemaps` contains the `SourceMap` data type together with the
  outines for querying it. A `SourceMap` associates each `MirNode` with
  the `PNode` it originated from
- `mirgen` implements the `PNode` AST to MIR code translation
- `astgen` implements the MIR code to `PNode` AST translation
- `mir/utils` contains non-essential routines that are mainly meant to
  be used for inspection during debugging
- `utils/containers` implements various `seq`-based containers that are
  generally useful in DOD contexts

Misc
----

- implement printing the symbol kind in `astrepr`

Future directions
=================

The MIR processing is integrated into the C and JS backend the same way
as `transf` is: the code-generators have to explicitly invoke
both. When the body of a procedure or a top-level statement is
encountered:
1. `transf` is executed for the code
2. the transformed code is passed to the MIR processing step (if used)
3. the AST resulting from the previous step is passed on to the actual
   code-generation logic

This approach has the following problems:
- the code-generators need to know about the existence of both `transf`
  and the MIR processing step
- the logic for invoking both steps is duplicated into each backend
- should the `lambdalifting` pass become a MIR pass, the pass is likely
  going to require access to the MIR code of multiple procedures at
  once. With the current processing pipeline, the required state and
  logic for managing would, to some degree, also need to be duplicated
  into each backend/code-generator

Instead, using an approach similar to the one used by the VM backend
(`vmbackend`) for both the C and JS backend would make sense.
`vmbackend` is responsible for: figuring out which procedure to run
code-generation for, transforming the code (`transf`), applying the
MIR processing, and then passing the resulting AST to the
code-generator (`vmgen`). This could be generalized into a reusable
abstraction that is then used across all backends, removing the
duplication of logic and making backend processing more unified.

---

### To-dos
- [x] integrate the to-and-from translation step into:
  - [x] `cgen`
  - [x] `vmgen`
  - [x] `jsgen`
- [x] implement the things still missing from `mirgen` 
- [x] refactor, cleanup, and document the added modules
- [x] write a proper commit/PR message

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -768,3 +768,21 @@ proc `[]`*(node: PNode, slice: NodeSliceName): seq[PNode] =
     of SliceAllIdents: node.sons[0..^3]
     of SliceAllArguments, SliceAllBranches: node.sons[1..^1]
     of SliceBranchExpressions: node.sons[0 .. ^2]
+
+iterator branches*(node: PNode): tuple[position: int, n: PNode] =
+  ## Returns all branches of the ``case`` statement or expression `node` in
+  ## order of occurrence. `position` is the 0-based position of the branch,
+  ## not the index of the sub-node
+  assert node.kind in {nkRecCase, nkCaseStmt}
+  for i in 1..<node.len:
+    yield (i-1, node[i])
+
+iterator branchLabels*(node: PNode): (int, PNode) =
+  ## Returns all labels of the branch-like constructs (i.e. ``of``, ``if``,
+  ## ``elif``) that `node` represents, together with their position. For
+  ## convenience, ``else`` branches are also allowed: they're treated as having
+  ## no labels
+  assert node.kind in {nkOfBranch, nkElifBranch, nkElse, nkElifExpr,
+                       nkElseExpr, nkExceptBranch}
+  for i in 0..<node.len-1:
+    yield (i, node[i])
--- a/compiler/mir/mirchangesets.nim
+++ b/compiler/mir/mirchangesets.nim
@@ -1,0 +1,321 @@
+## This module implements the ``Changeset`` API, which is the main way of
+## applying changes to a ``MirTree``.
+##
+## Instead of modifying a ``MirTree`` directly, the changes (which can be
+## insertions, replacements, or removals) are first recorded into a
+## ``Changeset``. This allows for recording changes independent of each other
+## concurrently and later apply the changes all at once.
+##
+## A cursor-like API is used. One has to first move the ``Changeset`` to the
+## position that the to-be-recorded change applies to via ``seek``.
+##
+## Before applying a ``Changeset`` to a ``MirTree``, it has to be prepared via
+## a call to ``prepare`` first, after which the ``Changeset`` is sealed an no
+## further changes can be recorded. ``prepare`` is responsible from normalizing
+## the internal representation of the ``Changeset`` and is required for the
+## later application to work.
+##
+## Applying the ``PreparedChangeset`` is done via ``apply`` -- updating the
+## ``MirTree``'s corresponding ``SourceMap`` via ``updateSourceMap``. For
+## efficiency, ``apply`` consumes the changeset, so it is advised to call
+## ``updateSourceMap`` first.
+##
+## Order of application
+## --------------------
+##
+## Nodes are inserted *before* the node at the insertion position, meaning that
+## it is allowed for an insertion to overlap with a removal/replacement at the
+## *start* position. All other forms of overlapping changes are disallowed.
+## If there exist two or more insertion at the same position, they are applied
+## in the order the changes were recorded. That is, the nodes from the insertion
+## recorded first are inserted first, then that of the second recorded, then
+## the third one's, etc.
+
+
+import
+  std/[
+    algorithm
+  ],
+  compiler/mir/[
+    mirtrees,
+    sourcemaps
+  ],
+  compiler/utils/[
+    idioms
+  ]
+
+type
+  Row = object
+    ## A changeset row. This is the building block of a changeset. Each row
+    ## represents a modification, of which there are three kinds:
+    ## - removal: `orig` has a length > 0 and `src` is empty
+    ## - insertion: `orig` has a length of 0 and `src` hasn't
+    ## - replacement: both `orig` and `src` have a length > 0
+    orig: HOslice[NodeIndex] ## the slice of nodes this change affects
+    src:  HOslice[NodeIndex] ## a slice in the buffer of staged nodes
+
+    source: uint32 ## the meaning depends on whether or not this row is part
+                   ## of a ``PrepareChangeset``. If it is, `source` is a
+                   ## ``SourceId``, otherwise it's a ``NodeInstance``.
+                   ## The source information is only relevant for insertions
+
+  Changeset* = object
+    # TODO: split ``Changeset`` into two parts:
+    #       1. the core ``Changeset`` - stores the rows and the node buffer
+    #       2. an object storing the data needed to build the changeset (e.g.
+    #          ``Cursor``, ``ChangeBuilder``, etc.)
+    tree: MirTree
+      ## stored in the changeset so that it's not required to pass the tree to
+      ## each ``Changeset`` routine
+    # TODO: `tree` needs to either become a view as soon as possible, or the
+    #       API of ``Changeset`` adjusted in a way that it's no longer required
+    #       for the tree to be stored as part of ``Changeset``. The full copy
+    #       is not acceptable
+
+    nodes: seq[MirNode]
+    rows: seq[Row]
+
+    numTemps: uint32 ## the number of existing temporaries
+
+    pos: NodePosition ## the cursor position
+
+  PreparedChangeset* = object
+    nodes: seq[MirNode]
+    rows: seq[Row]
+
+    diff: int        ## the number of additions/removals
+    stagingSize: int ## the minimum amount of nodes the working area must be
+                     ## able to hold
+
+func single[T](x: T): HOslice[T] {.inline.} =
+  ## Utility for creating a slice with a single item
+  HOslice[T](a: x, b: succ(x))
+
+func span[T](a, b: T): HOslice[T] {.inline.} =
+  HOslice[T](a: a, b: b)
+
+func empty[T](x: typedesc[HOslice[T]]): HOslice[T] {.inline.} =
+  HOslice[T](a: default(T), b: default(T))
+
+func row(start, fin: NodePosition, src: HOSlice[NodeIndex];
+         source = NodePosition(0)): Row {.inline.} =
+  ## Convenience constructor for ``Row``
+  Row(orig: span(NodeIndex(start), NodeIndex(fin)),
+      src: src,
+      source: source.uint32)
+
+func addSingle(s: var MirNodeSeq, n: sink MirNode): HOSlice[NodeIndex] =
+  s.add n
+  result = single(s.high.NodeIndex)
+
+func initChangeset*(tree: MirTree): Changeset =
+  ## Initializes a new ``Changeset`` instance. Until the resulting
+  ## ``Changeset`` is applied, the associated tree must not be modified
+  result.tree = tree # warning: this creates a full copy
+
+  # count the number of existing temporaries:
+  for i, n in tree.pairs:
+    if n.kind in DefNodes and
+       (let ent = child(tree, i, 0); ent.kind == mnkTemp):
+      result.numTemps = max(ent.temp.uint32 + 1, result.numTemps)
+
+func getTemp*(c: var Changeset): TempId =
+  ## Allocates a slot for new temporary and returns its ID
+  result = TempId(c.numTemps)
+  inc c.numTemps
+
+template position*(cs: Changeset): NodePosition =
+  ## The current position of the cursor
+  cs.pos
+
+func seek*(c: var Changeset, dest: NodePosition) {.inline.} =
+  ## Moves the internal cursor position to `dest`
+  assert c.pos in c.tree
+  c.pos = dest
+
+func replace*(c: var Changeset, n: sink MirNode) =
+  ## Records a change that replaces the node at the current cursor position
+  ## with `n`, inheriting it origin information. If the cursor points to a
+  ## sub-tree, the whole sub-tree is replaced
+  let next = sibling(c.tree, c.pos)
+  c.rows.add row(c.pos, next, c.nodes.addSingle(n), c.pos)
+  c.pos = next
+
+func insert*(c: var Changeset, n: sink MirNode, source: NodeInstance) =
+  ## Inserts `n` at the current cursor position, using `source` as the
+  ## inserted node's origin
+  c.rows.add row(c.pos, c.pos, c.nodes.addSingle(n), source.NodePosition)
+
+template insert*(c: var Changeset, source: NodeInstance, name: untyped,
+                 body: untyped) =
+  ## Records an insertion at the current cursor position, providing direct
+  ## access to the internal node buffer inside `body` via an injected variable
+  ## of the name `name`. `source` is the node to inherit the source/origin
+  ## information from
+  block:
+    let
+      start = c.nodes.len.NodeIndex
+      # evaluate `source` before `body`, as the latter might change what
+      # `source` evaluates to:
+      i = NodePosition source
+
+    var name: MirNodeSeq
+    swap(c.nodes, name)
+    body
+    swap(c.nodes, name)
+
+    c.rows.add row(c.pos, c.pos, span(start, c.nodes.len.NodeIndex), i)
+
+template replaceMulti*(c: var Changeset, name, body: untyped) =
+  ## Records a repacement of the node or sub-tree at the current cursor
+  ## position, providing direct access to the internal node buffer
+  ## inside `body` via an injected variable of the name `name`
+  # TODO: rename to ``replace`` once the sem bug that currently prevents this
+  #       is fixed
+  block:
+    let start = c.nodes.len.NodeIndex
+    var name: MirTree
+    swap(c.nodes, name)
+    body
+    swap(c.nodes, name)
+
+    let next = sibling(tree(c), c.pos)
+    c.rows.add row(c.pos, next, span(start, c.nodes.len.NodeIndex), c.pos)
+    c.pos = next
+
+func remove*(c: var Changeset) =
+  ## Records the removal of the currently pointed to sub-tree
+  let next = sibling(c.tree, c.pos)
+  # use an empty source slice
+  c.rows.add row(c.pos, next, empty(HOSlice[NodeIndex]))
+  c.pos = next
+
+func skip*(c: var Changeset, num: Natural) =
+  # Skips `num` sub-trees
+  for _ in 0..<num:
+    c.pos = sibling(c.tree, c.pos)
+
+func prepare*(c: sink Changeset, sourceMap: SourceMap): PreparedChangeset =
+  ## Prepares `c` for being applied
+
+  # applying the changes can be done much more efficiently if they are ordered
+  # by ascending modification position
+  c.rows.sort proc(a, b: auto): int =
+    result = a.orig.a.int - b.orig.a.int
+    if result == 0:
+      # use the end position as the second-order, so that insertions sharing
+      # their start position with replacements/removals are applied first
+      result = a.orig.b.int - b.orig.b.int
+
+  # calcuate the difference in the amount of nodes plus the required size for
+  # the working/staging area:
+  for row in c.rows.items:
+    result.diff -= row.orig.len
+    result.diff += row.src.len
+    result.stagingSize = max(result.stagingSize, result.diff)
+
+  # the `source` column of each row currently refers to the node to inherit
+  # the source information from (but only if the row represents an insertion
+  # or removal). Since the source-information attachments will become stale
+  # once the source mappings are modified, `source` is translated to the
+  # ``SourceId`` here already
+  for r in c.rows.mitems:
+    r.source = sourceMap[r.source.NodeInstance].uint32
+
+  result.nodes = move c.nodes
+  result.rows = move c.rows
+
+func moveRight[T](x: var openArray[T], src, dest, len: Natural) =
+  assert src < dest
+  for i in countdown(len-1, 0):
+    x[dest + i] = move x[src + i]
+
+func moveLeft[T](x: var openArray[T], src, dest, len: Natural) =
+  assert dest <= src
+  for i in 0..<len:
+    x[dest + i] = move x[src + i]
+
+iterator apply[T](s: var seq[T], diff, stagingSize: int, rows: seq[Row]): tuple[row: lent Row, writePos: NodeIndex] =
+  ## Performs the surrounding actions for integrating changes into `s`
+  ## .. warning:: Exiting the iterator via ``return`` or ``break`` will leave
+  ##              `s` in an inconsistent state
+  let oldLen = s.len
+  var
+    writePos = rows[0].orig.a.int
+      ## the start position of the "staging" area
+    readPos = writePos
+      ## the start position of the "original" area
+    origPos = writePos
+      ## keeps track of the logical position in the original sequeunce where
+      ## we're copying from
+
+  if stagingSize > 0:
+    # make space for the temporary/staging partition
+    let start = rows[0].orig.a.int
+    s.setLen(oldLen + stagingSize)
+
+    readPos = start + stagingSize
+    moveRight(s, start, readPos, oldLen - start)
+
+  assert writePos <= readPos
+
+  # iterate over all changes and apply them. A change either adds, removes, or
+  # replaces nodes.
+  for i, row in rows.lpairs:
+    let start = row.orig.a.int
+
+    assert writePos <= readPos
+    assert start >= origPos, "overlapping changes"
+
+    # first move over the original nodes (if any):
+    if start > origPos:
+      # move the nodes from the "original" area to the "staging" area:
+      let L = start - origPos
+      if readPos != writePos:
+        moveLeft(s, readPos, writePos, L)
+      # commit the nodes to "done" area:
+      readPos += L
+      writePos += L
+      origPos += L
+
+    yield (row, writePos.NodeIndex)
+
+    # skip the nodes we're replacing or removing:
+    readPos += row.orig.len
+    origPos += row.orig.len
+
+    # commit:
+    writePos += row.src.len
+
+  assert origPos <= oldLen
+
+  # move the remaining nodes:
+  if oldLen > origPos and readPos != writePos:
+    moveLeft(s, readPos, writePos, oldLen - origPos)
+
+  if diff != stagingSize:
+    # resize to the correct number of items
+    s.setLen(oldLen + diff)
+
+func updateSourceMap*(m: var SourceMap, c: PreparedChangeset) =
+  ## Updates the source mappings stored by `m` according to the changes
+  ## recorded in `c`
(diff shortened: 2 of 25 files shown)


zerbina:
The translation step is now fully integrated into the C and JS code-generators. For `vmgen`, it is partially integrated through `vmjit` (meaning that the VM back-end doesn't make use of it yet).

As a debugging aid and to help one visualize how the MIR code looks and how the translation affects the AST reaching the code-generator, I added three defines that work similar to `--expandArc`:
- `--define:nimShowMirInput=x`: echoes the `PNode` AST for the procedure/module with name `x` that is translated into MIR code. This is the AST the code-generator would see with the MIR translation step disabled.
- `--define:nimShowMir=x`: echoes the generated MIR code for the procedure/module with the name `x`
- `--define:nimShowMirOutput=x`: echoes the `PNode` AST that is generated from the MIR code for the procedure/module with the name `x`. This is the AST the code-generator operates on

The traces don't use proper diagnostic types yet, but are instead directly written out via `msgs.writeln`. In addition, the names of the defines are not final (feel free to suggest better ones) and they're also not documented in `debug.rst` yet.

Note that due to how the code-generators are implemented, nothing will be echoed for procedures that are not part of the alive code.

---

The largest part of the test suite for the C target works, with most of the remaining failing test being related to `injectdestructors`, possibly because the AST now violates some assumption made by the pass (but there might also be other bugs left still).

If the `injectdestructors` related failures are really only because of AST differences, then I think it'd make sense to disable the MIR translation step for procedures and modules marked with `sfInjectDestructors` (i.e. those that require destructor injection) until #450 is merged.

I'm currently focusing on making the remaining test work, and, after that, will work on refactoring and documenting the code.

zerbina:
The functional part is done and the tests pass. All that remains is cleaning up the code and adding documentation, which, due to the amount of code, is likely going to take a bit of time.

So far, the impact on compile time for me is at an additional 1.5-2 seconds for a single bootstrapping iteration (when using `-d:release --gc:refc`). Peak memory consumption for a single iteration increased by ~90 MB. The by far largest amount of time is spent inside `astgen` (i.e. the MIR to `PNode` translation), which makes sense, as it performs a significant amount of allocations.

zerbina:
Even though I'm still in the process of cleaning up and documenting the code, the PR can already be considered ready for review.

In addition, I've pushed the draft of a high-level MIR overview. My goal here is to get it into a state where it provides a solid foundation for incrementally improving it as part of future PRs. There are multiple ways to help in this area, namely by:
1. proof reading the document
2. asking "why", "what", "how" questions based on the text in the document
3. looking at the output of `-d:nimShowMir` and asking "why", "what", "how" questions

The questions resulting from both 2. and 3. would greatly help me in figuring out what parts still need to be expanded upon.

compiler/ast/ast_query.nim
```diff
@@ -765,3 +765,21 @@ proc `[]`*(node: PNode, slice: NodeSliceName): seq[PNode] =
     of SliceAllIdents: node.sons[0..^3]
     of SliceAllArguments, SliceAllBranches: node.sons[1..^1]
     of SliceBranchExpressions: node.sons[0 .. ^2]
+
+iterator branches*(node: PNode): tuple[position: int, n: PNode] =
```
alaviss, line 772: One of the few times I'd love to have `enumerate` :p

doc/mir.rst
```diff
@@ -0,0 +1,161 @@
...
+Even though the control-flow primitives are simpler and more regular than
+the ones used by the higher-level AST, they're still fairly high-level.
+For example, constructs like ``try`` and ``case`` statements still exist.
+The reason for this is twofold:
+- due to the code-generators still operating on ``PNode`` AST, the MIR code
+  has to be translated back to it. Translating low-level control-flow
+  primitives such as 'goto's back into 'if', 'block', and 'break' is fairly
+  involved and would also introduce additional overhead to the
+  back-translation
+- some transformation passes currently using the MIR are simplre to implement
```
alaviss:
```suggestion
- some transformation passes currently using the MIR are simple to implement
```

doc/debug.rst
```diff
@@ -420,6 +420,29 @@ procedure.
...
      this) it might inhibit unexpected behavior.
 
+MIR Input and Output
+====================
+
+For debugging issues related to the MIR but also code-generator issues in
+general, one can print the input and output to the MIR canonicalization step
+plus the corresponding `PNode`-AST output.
+
+To print the `PNode`-AST that reaches `mirgen`, `--define:nimShowMirInput=name`
```
alaviss, line 430: Nothing to change here, but at one point we should create a kind of annotation you can tag on procs you want to debug (ie. a pragma) and maybe move away from using defines.

compiler/utils/cursors.nim
```diff
@@ -0,0 +1,48 @@
...
+## - works for object fields of all types, not only ``ref``s and closures
+##
+## .. warning:: For ``seq|string|ref`` types that don't use destructors,
+##              copies turned into blit copies will result in full copies due
+##              to a C code-generator issue. In other words, using ``Cursor``
+##              for ``seq`` and ``ref`` types when both are not using a
+##              destructor-based management scheme has no benefit in many cases
+
+type
+  Cursor*[T] = distinct T
```
alaviss: Maybe this should be called `Shallow`. `Cursor` makes me think of pointers, which is not what this module is.

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+  EValue = object
+    ## Encapsulates information about the abstract value that results from
+    ## an operation sequence. It's used as a way to transport said information
+    ## between the generator procedures for operators
+    # XXX: maybe rename to ``Computation``?
+    typ {.cursor.}: PType
+
+  HOslice[T] = object
+    ## Half-open slice
+    a, b: T
```
alaviss: Should be moved out elsewhere. This is duplicated on `mirchangesets` as well.

Also, what utility does this provide compared to `Slice` (ie. why couldn't `Slice` be used instead)?

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+## Operation wise, the input AST is traversed via recursing into statements and
+## expressions, ignoring declarative constructs that are not relevant to nor
+## representable with the MIR.
+##
+## None of the MIR operations that imply structured control-flow produce a
+## value, so the input expression that do (``if``, ``case``, ``block``, and
+## ``try``) are translated into statements where the last expression in each
+## clause is turned into an assignment to, depending on the context where
+## they're used, either a temporary or existing lvalue expression. The latter
+## are forwarded to the generation procedures via ``Destination``.
```
alaviss, line 13: Could be helpful to provide samples of "regular" AST and its corresponding MIR.

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+      for i in 0..<n.len-1:
+        if n[i].kind != nkEmpty:
+          # there are statements in the list -> the list needs to be kept
+          return n
+
+      result = n.lastSon
+    else:
+      result = n
+
+  var n {.cursor.} = n
```
alaviss, line 246: I don't think you should use `{.cursor.}` here. The pragma's semantics is a mess and I'd prefer that we don't touch it if it's not explicitly required.

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+func canonicalExpr(n: PNode): PNode =
+  ## Returns the canonical expression, i.e. the expression without leading
+  ## pragma blocks or empty statement lists
+  func skipped(n: PNode): PNode =
+    case n.kind
+    of nkPragmaBlock:
+      result = n.lastSon
+    of nkStmtListExpr:
+      # see if all nodes except the last one are empty. If they are, the
+      # ``nkStmtListExpr`` is skipped
```
alaviss: Is this a `transf` thing? I'm not a fan of this kind of processing. Macros and `sem` can and do produce nested empty `StmtList`, so this is not the best way to filter through them.

You should document what kind of AST are you expecting so that we know when can this logic be deprecated.

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+      for i in 0..<n.len-1:
+        if n[i].kind != nkEmpty:
```
alaviss: `sequtils.anyIt` is probably a good fit here

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+    n = it
+
+  result = n
+
+func selectWhenBranch(n: PNode, isNimvm: bool): PNode =
+  assert n.kind == nkWhen
+  if isNimvm: n[0][1]
+  else:       n[1][0]
+
+func selectDefKind(s: PSym): range[mnkDef..mnkDefCursor] =
```
alaviss, line 257: You probably don't want a `range` return type.

This means whatever variable you created from calling this function is tagged with that `range`. If you reuse that variable for other purposes, it might trigger a range error.

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+  ## Returns the kind of definition to use for the given entity
+  if sfCursor in s.flags: mnkDefCursor
+  else:                   mnkDef
+
+func isSome(x: Destination): bool {.inline.} =
+  x.kind != dkNone
+
+func initDestination(n: PNode, isFirst: bool): Destination =
+  result = Destination(kind: dkGen, node: n, isEmpty: isFirst, own: not isCursor(n))
```
alaviss:
```suggestion
  Destination(kind: dkGen, node: n, isEmpty: isFirst, own: not isCursor(n))
```

When it's an expression you don't have to use `result`

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+      discard
+    of dkFrag:
+      nodes: NodeSpan ## references a slice of nodes in the staging buffer
+    of dkGen:
+      node {.cursor.}: PNode
+
+    isEmpty: bool ## whether the destination has no value. Used to decide if
+                  ## an ``mnkInit`` or ``mnkAsgn`` is to be used
+    own: bool     ## whether the destination wants ownership over the assigned
+                  ## value. Used to select between a fast and normal assignment
```
alaviss: Consider using flags (enum + set) instead. They're easier to query and have a smaller memory footprint.

Bonus points for easy extension in the future.

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+
+  Block = object
+    ## Information about a ``block``
+    label: PSym ## the symbol of the block's label. 'nil' if the block has no
+                ## label
+    id: LabelId ## the block's internal label ID
+
+  CodeFragment = object
+    nodes: MirNodeSeq
+    source: seq[SourceIndex]
```
alaviss: This can use some docs

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+func halfOpen[T](x: Slice[T]): HOslice[T] {.inline.} =
+  ## Constructs a half-open slice from a normal slice. `x` is expected to
+  ## already represent a half-open slice
+  HOslice[T](a: x.a, b: x.b)
+
+template toOpenArray[T, I](x: seq[T], slice: Slice[I]): openArray[T] =
+  toOpenArray(x, slice.a.int, slice.b.int)
+
+template toOpenArray[T, I](x: seq[T], slice: HOslice[I]): openArray[T] =
+  toOpenArray(x, slice.a.int, slice.b.int - 1)
```
alaviss: I highly recommend caching `slice` in a temporary to avoid multiple evaluations.

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+  f.nodes.add n
+
+template subTree(f: var CodeFragment, n: MirNode, body: untyped) =
+  f.nodes.subTree n:
+    body
+
+func apply(frag: var CodeFragment, id: SourceIndex) =
+  ## Associates all nodes added since the previous call to ``apply`` with the
+  ## origin information identified by `id`
+  assert frag.nodes.len > frag.source.len
```
alaviss, line 308: You should drop this assert. For something like this I consider doing nothing a perfectly valid scenario.

compiler/mir/mirgen.nim
```diff
@@ -0,0 +1,2105 @@
...
+func tempNode(typ: PType, id: TempId): MirNode {.inline.} =
+  MirNode(kind: mnkTemp, typ: typ, temp: id)
+
+
+func nextTempId(c: var TCtx): TempId =
+  result = TempId(c.numTemps)
+  inc c.numTemps
+
+func nextLabel(c: var TCtx): LabelId =
+  result = LabelId(c.numLabels + 1)
```
alaviss, line 293: I'd recommend `high(LabelId)` as the invalid id if you need those instead of shifting.

alaviss (commented):
Initial look. I'll review the rest tomorrow.

compiler/utils/cursors.nim
zerbina: I called it `Cursor` because of the functional similarity to the `.cursor` pragma, and because the type is used in places where the pragma should've been used instead (but couldn't, due to the mentioned limitations).

If wanted, I can rename the type to `Shallow`, but I don't know what name to use for the module in that case.

compiler/mir/mirgen.nim
zerbina: Yep, I know and agree, but I don't know where to. A dedicated module feels off, same as with as a general utility one. The `idioms` module might be a good place to move it to.

As for its utility: its easier to use with unsigned integers (e.g. no special handling is required for empty slices) compared to `Slice`.

compiler/mir/mirgen.nim
zerbina: Cursor inference fails here, so the explicit annotation makes sense. Ignoring the semantics when used for object fields, what about the semantics do you consider to be a mess?

compiler/mir/mirgen.nim
zerbina: I'm not sure I understand what you mean. The idea here is to skip/ignore redundant `StmtListExpr` nodes such as `StmtListExpr (Empty), someExpr` and get the underlying `someExpr`.

compiler/mir/mirgen.nim
zerbina: I do want a `range` return type here, both for documentation purposes and so that one doesn't have to use an `else: unreachable()` branch when using the result with a `case`.

compiler/mir/mirgen.nim
zerbina: I know, but I'm still undecided whether I prefer using an explicit `result` assignment in such cases or not.

Does there exist a general preference?

compiler/mir/mirgen.nim
zerbina: But then the zero-representation is not the 'none'/invalid value, which is a property that I dislike and that can cause issues easily.

compiler/mir/mirgen.nim
zerbina: Calling `apply` when there's nothing to update hints at a logic or programming error somewhere, hence the assert. I would prefer to keep it (but can update the documentation).

compiler/mir/mirgen.nim
alaviss: `utils`, `idioms`, it can be anywhere really. I'm more concerned with not duplicating it.

And this could potentially find a place in `system`, though we might want a different name then :P

compiler/mir/mirgen.nim
alaviss: What I wanted to know is /how/ these nodes come to life. An `StmtListExpr` that have a bunch of `Empty` is pretty weird, no?

And what if that `StmtListExpr` is like this: `StmtListExpr(StmtListExpr() StmtListExpr() Call(...)`? Would `transf` or whatever that comes before collapse the inner empty exprs?

compiler/mir/mirgen.nim
alaviss: `{.cursor.}` documentation is very vague on how it works. It is said to "prevents object construction/destruction pairs" and the only behavior demonstrated is eliding all ref count ops.

It's an optimization that's potentially unsafe (ie. the object being viewed is destroyed), so I'd want to leave it out if it is not necessary, but then we have atomic RC in nimskull rn, so avoiding them is a pretty big thing...

compiler/mir/mirgen.nim
alaviss: I feel like that's a long shot, given how `applySource` just manually do the check instead.

But I don't have strong opinions on this, so feel free to do whatever.

compiler/mir/sourcemaps.nim
```diff
@@ -0,0 +1,69 @@
...
+    containers
+  ]
+
+type
+  SourceIndex* = distinct range[0'u32 .. high(uint32)-1]
+    # TODO: maybe rename to ``SourceInstance``? Whether it's an index is an
+    #       implementation detail
+  SourceId* = distinct uint32
+    ## A ``SourceIndex`` with an explicit "nil" state. This is essentially an
+    ## optimized ``Option[SourceIndex]``
```
alaviss: These two needs distinct names. I kept conflating `Id` and `Index` reading the code.

alaviss: I feel like you might want to just commit to 1-based indices instead of having two types for translation :p

compiler/utils/cursors.nim
alaviss: I think it's fine to keep the name.

But I can see that it's too easy to edit an object shallow copied in `Cursor`, which can have devastating effects, for example: de-synchronized length between a seq and its cursor.

Given how unsafe the API is, I would want a kind of safety net (ie. preventing mutating ops) or something that screams unsafe.

Personally I would like to /not/ have this around. I would take a `ref seq` any day of the week just for the ownership semantics, and it's just a lot more foolproof.

compiler/mir/mirchangesets.nim
```diff
@@ -0,0 +1,332 @@
+## This module implements the ``Changeset`` API, which is the main way of
```
alaviss, line 1: I'm not a fan of this.

This system basically means that our trees are too expensive to modify, and hint at a symptom that has to be handled.

On top of that, the change sets are potentially dangerous, as changes must be batched carefully and you cannot check the state of the tree before the changes are applied. This means MIR processing code have to be extremely careful about backtracking and batching, making the pipeline harder to modify.

compiler/utils/containers.nim
```diff
@@ -0,0 +1,50 @@
...
+template `[]=`*[I; T](x: var Store[I, T], i: I, it: T): untyped =
+  ## Overwrites the item corresponding to `i` with `it`
+  # TODO: convert to ``distinctBase`` instead
+  x.data[int(i)] = it
+
+func add*[I; T](x: var Store[I, T], it: sink T): I {.inline.} =
+  ## Appends a new item to the Store and returns the ID assigned to
+  ## it
+  x.data.add it
+  result = I(x.data.high)
```
alaviss, line 51: There's a potential problem where `I` is smaller than `int`: `I` could overflow and there's no facility to verify that's not the case.

alaviss (commented):
I'll need to comb through this a couple more passes (I haven't fully evaluated `astgen`, `mirgen`, `mirtrees`), but here are my current remarks.

compiler/utils/cursors.nim
zerbina: The `Cursor` type together with the module and all usages is now removed. For `astgen`, I resorted to consuming the tree and source-map in order to get around the copies, but `Changeset` has to perform a full copy for now.

compiler/mir/mirgen.nim
zerbina: The documentation is indeed lacking, but how the `.cursor` pragma works is straightforward: local variables annotated with it are non-owning locations for which all lifetime hooks are disabled. Assigning to cursor locations always uses a bit-wise copy.

For object fields, the pragma only works reliably for `ref` types and closures, but I consider this to be a design issue that can be easily fixed.

compiler/mir/mirgen.nim
zerbina: `sem` transforms `when` statements that don't apply into `Empty` nodes -- same with both `runnableExamples` calls and, during template expansion, `CommentStmt` nodes.

Neither empty `StmtListExpr` nor `StmtList` should get through semantic analysis (they currently do, but that's a separate issue).

Anyhow, in order to not duplicate the check and also handle `CommentStmt`s, I replaced the `anyIt` usage with `ast_query.stupidListExpr`.

compiler/utils/containers.nim
zerbina: I added a `rangeCheck` that makes sure that the resulting index is representable by `I`.

compiler/mir/sourcemaps.nim
zerbina: The old `SourceId` is now gone and replaced with usage of the new `OptIndex` container type.

compiler/mir/mirgen.nim
zerbina: I moved the `HOslice` type plus the associated routines that are generally useful to the `utils/idioms` module for now.

zerbina:
@alaviss I've addressed the blockers you mentioned on Matrix (`Cursor` usage, missing range checks, `Id`/`Index` confusion), but am leaving the PR open so that you can take another look -- if you think it's good to go, feel free to merge it.

zerbina:
I've noticed and fixed multiple problems that weren't caught by the test suite:
- `jsgen` was not invoking `canonicalize` for all input AST
- because of the separation of cursor inference from the `injectdestructors` pass and me forgetting to adjust `jsgen`, cursor inference wasn't performed anymore when using the JS target 
- the `injectdestructors` pass doesn't work with the output of `astgen`

The last one is likely a problem with the implementation of the `injectdestructors` pass, but because of a bug, the MIR processing step was effectively skipped for code being flagged for destructor injection, thus hiding the issue. The to-and-from MIR translation still took place -- the resulting AST was just not used.

As a quick fix, I officially disabled the MIR step for code that is flagged for destructor injection for now. This prevents turning further transformations into MIR passes (since not all code reaching the code-generator is passed to MIR processing first), but fortunately, the PR that turns the `injectdestructors` pass into a MIR pass (#450) is almost ready to be merged.

PR #491, opened 2022-12-10
vm: decouple the execution loop from `vmjit`

zerbina:
## Summary
When using the VM to execute code at compile-time, the bytecode generation for procedures happens on demand. `vmgen` collects all procedures that the AST it generates bytecode for makes calls to, and `vmjit` then registers a stub in the VM's function table for each.

If the VM encounters a stub, it makes a call to `vmjit`, which then generates the bytecode for the not-yet-code-gen'ed procedure, also updating the corresponding function table entry with the information required to execute the procedure.

The problem with this approach is that the VM needs to know about the existance of on-demand code-generation. Implementation wise, this has the following consequences:
1. the `vm` module needs to import `vmjit` which then imports `vmgen`. In other words, the VM has an import dependency on the code-generator. This could be solved via a callback, but using dynamic dispatch here has its own problems.
2. `TCtx` needs to store the data relevant for `vmjit`, which in turn means that the type definitions that could otherwise be located in the `vmjit` module need to now be present in `vmdef`
3. code-generation can fail, in which the case the error needs to be propagated to the callsite of `rawExecute`

Instead, when the VM now encounters a call to a procedure with a stub entry, it yields control back to the caller of `rawExecute` while reporting a dedicated return code plus the entry that is a stub.

This makes it possible for the caller to respond to this situation however it sees fit. Right now, this means invoking `vmjit` and, if there were no errors, resuming execution.

While not done as part of this PR, the change unlocks a large number of improvements, such as, for example, removing the import dependency on  `vmgen` from the `vm` module.

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -82,6 +82,28 @@ from std/math import round
 # these were includes previously, so they're re-exported for compatibility
 export vmhooks, vmops
 
+type
+  ExecResultKind = enum
+    erkDone
+      ## execution is done. There is no more code to execute
+    #ekError # TODO: also communicate whether execution was aborted due to an
+             #       error via ``ExecResult``
+    erkMissingProcedure
+      ## a procedure stub was called. The stub has to be resolved before
+      ## continuing execution
+
+  ExecResult = object
+    ## The internal result of a single execution step (i.e. a call to
+    ## ``rawExecute``)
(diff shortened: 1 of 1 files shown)


saem (approved):
This really highlights what to evolve and how wrt splitting out the vmgen dependency. 🎉

Thank you!

Bors r+

PR #492, opened 2022-12-11
compiler: add the `nkNimNodeLit` node kind

zerbina:
## Summary
AST that represents the content of a `NimNode` tree was not detectable
as such without inspecting the surrounding AST. This complicated logic
inspecting the bodies of macros or `.compileTime` procedures, as it had
to specifically ignore the symbol/symbol-choice nodes part of the AST
for `mNBindSym` magic calls.

The existence of `NimNode` literals inside normal AST is now made
explicit and also detectable without further context via a dedicated
node kind, also preventing it from going unnoticed if erroneously
present in places where it shouldn't be.

## Details
- use `nkNimNodeLit` nodes when deserializing `NimNode`s in
  `vmcompilerserdes`
- static `bindSym` calls are now expanded into a `nkNimNodeLit` during
  semantic analysis
- don't use an extra node for providing source information to
  `opBindSym` -- the source information is already provided by the VM
- the arguments to dynamic `bindSym` calls are now considered during
  constant folding
- remove the `opcNBindSym` VM instruction
- remove the now obsolete `rvmInvalidBindSym` report
- propagate error nodes in `semBindSym`
- include `semBindSym` in compiler tracing

## Changes to typed AST
```nim
...
# inside a compile-time evaluation context (macro, `.compileTime` procedure, or `static` expression/statement)

bindSym("s")
# is now represented in the AST as:
# `NimNodeLit(Sym "s")`
# instead of:
# `Call(Sym "bindSym", Sym "s")`

let ident = ...
bindSym(ident)
# is now represented in the AST as:
# `Call(Sym "bindSym", Sym "ident", Sym "brClosed", IntLit ...)
# instead of:
# `Call(Sym "bindSym", Sym "ident", Sym "brClosed", ..., IntLit ...)
```


compiler/vm/vm.nim
```diff
@@ -2584,15 +2584,23 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): RegisterInd
...
+      #       capturing a ``PContext``, keeping it alive until the ``TCtx``
+      #       instance is destroyed), we could yield back control to ``sem``
+      #       with the information that a symbol lookup should be performed.
+      #       The lookup result is then stored into a register also
+      #       communicated earlier and VM execution is resumed.
+      #       In some way (but not exactly), this is similar to returning a
+      #       single ``nnkIdent`` node from a macro, and then passing the
+      #       resulting semantically analysed AST back into another macro as
+      #       a ``typed`` argument.
+      #       But there is also the question if ``semDynamicBindSym`` is
```
saem, line 2600: We simply should disallow breaking scoping/lookup rules, the more of that we can get rid of the better. Separate PR for that, I'd imagine..

compiler/vm/vm.nim
```diff
@@ -2584,15 +2584,23 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): RegisterInd
...
     of opcNDynBindSym:
       # experimental bindSym
+      # IDEA: instead of calling back into ``sem`` via a VM callback registered
+      #       inside ``semDynamicBindSym`` (which has the severe problem of
+      #       capturing a ``PContext``, keeping it alive until the ``TCtx``
+      #       instance is destroyed), we could yield back control to ``sem``
+      #       with the information that a symbol lookup should be performed.
+      #       The lookup result is then stored into a register also
+      #       communicated earlier and VM execution is resumed.
+      #       In some way (but not exactly), this is similar to returning a
```
saem, line 2596: I really like the symmetry, but prefer removing dynamic bind sym

compiler/vm/vmgen.nim
```diff
@@ -551,15 +551,20 @@ template cmpFloatRep(a, b: BiggestFloat): bool =
...
-    of nkEmpty: result = true # TODO: is this case ever needed?
-    of nkType: result = a.typ == b.typ
-    of nkStrLit..nkTripleStrLit: result = a.strVal == b.strVal
+    of nkSym:           result = a.sym == b.sym
+    of nkIdent:         result = a.ident.id == b.ident.id
+    of nkEmpty:         result = true
+    of nkType:          result = a.typ == b.typ
+    of nkStrKinds:      result = a.strVal == b.strVal
+    of nkIntKinds:      result = a.intVal == b.intVal
+    of nkFloatLiterals: result = cmpFloatRep(a.floatVal, b.floatVal)
```
saem, line 567: So clean!

compiler/vm/vmgen.nim
```diff
@@ -2610,6 +2604,19 @@ proc gen(c: var TCtx; n: PNode; dest: var TDest; flags: TGenFlags = {}) =
...
+    # is correct). ``NimNode``s not marked with `nfSem` can be freely modified
+    # inside macros, so in order to prevent mutations of the AST part of the
+    # constant, we perform a defensive tree copy before assigning the literal
+    # to the destination
+    if dest.isUnset:
+      dest = c.getTemp(n.typ)
+
+    var tmp: TDest = c.getTemp(n.typ)
+    c.genLit(n, c.toNodeCnst(n[0]), tmp)
+    c.gABC(n, opcNCopyNimTree, dest, tmp)
```
saem, line 2618: Good

saem (approved):
I like it, a lot. It's definitely a type of cake that exists at comptime and makes sense to have explicitly.

PR #494, opened 2022-12-12
macros: remove the experimental dynamic `bindSym`

zerbina:
## Why
It allowed for violating the scoping and lookup rules and the implementation is also rather complex.

Implementation wise, the feature worked as follows:
1. when semantically analysing a `bindSym` call, the current `PContxt` and  `PScope` are captured in a closure that performs the same operation as a normal `bindSym` would, just with the `ident` and `rule` arguments supplied by the VM.
2. the closure is directly registered as a VM callback and index of the callback's stored as part of the call's AST as an integer literal.
3. a dedicated instruction (`opcNDynBindSym`) uses said index at run-time to dispatch to the callback.

There are multiple issues:
- changes to the `PScope` that happen after the `mNBindSym` magic is semantically analysed (and the callback registered) are visible and participate in the lookup performed by `bindSym`. This makes the result of a dynamic `bindSym` call dependent on where the surrounding macro/compile-time procedure is invoked
- registered callbacks are kept until the VM environment is destroyed (for the main VM instance, this is currently never), and since `PContext` and `PScope` are `ref`s, all instance of these types (and everything owned by them) captured for a dynamic `bindSym` calls are kept alive until the compiler exits
- `sem` needs to know about implementation details of the VM

## Details
- remove the `dynamicBindSym` feature enum
- remove the `opcNDynBindSym` instruction plus the code-generation logic for it
- remove mentions of the feature from the compiler and manual

As a consequence, the `mNBindSym` magic is now no longer present in the AST past semantic analysis.

---

Nim-CPS makes use of this feature, so I'm also requesting a review from @alaviss, in order for them to weigh in on the impact. If the removal is too disruptive without something to replace it, the PR can also be put on hold.


saem (approved):
I looked at the usage in CPS, it's fine to remove it.

Bors r+

disruptek:
CPS doesn't work in nimskull so there's no blocker here or anywhere going forward.

PR #495, opened 2022-12-12
vm: move the compiler interface into a separate module

zerbina:
## Summary
Move the parts that make up the direct interface between the compiler and the VM into a separate module, in order to reduce the amount of dependencies the `vm` module has on the rest of the compiler.

This is also a step towards completely separating the VM from `vmgen`.

---

@saem If this PR interferes with #484 too much, I can also turn it into a draft PR and wait with merging until #484 is finished.

## Notes for Reviewers
* I'm not entirely sure of the name `compilerbridge`

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -915,7 +915,7 @@ proc semNormalizedLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
     if not hasError and sfCompileTime in v.flags:
       var x = newNodeI(result.kind, v.info)
       x.add producedDecl
-      vm.setupCompileTimeVar(c.module, c.idgen, c.graph, x)
+      setupCompileTimeVar(c.module, c.idgen, c.graph, x)
 
     if v.flags * {sfGlobal, sfThread} == {sfGlobal}:
       # this is just logging, doesn't need to be converted to an error
--- a/compiler/vm/compilerbridge.nim
+++ b/compiler/vm/compilerbridge.nim
@@ -1,0 +1,465 @@
+## This module implements the interface between the VM and the rest of the
+## compiler. The VM is only interacted with through this interface. Do note
+## that right now, the compiler still indirectly interacts with the VM through
+## the ``vmdef.PCtx`` object.
+##
+## The interface includes:
+## - the procedure that sets up a VM instance for use during compilation
+##   (``setupGlobalCtx``)
+## - the routines for executing expressions, statements, and macros with the VM
+## - an implementation of the ``passes`` interface that executes processed
+##   code with the VM (``evalPass``)
+## - the VM-related compilerapi
+
+import
+  std/[
+    tables
+  ],
+  compiler/ast/[
+    ast_types,
+    ast,
+    errorhandling,
+    errorreporting,
+    lineinfos,
+    trees
+  ],
+  compiler/front/[
+    msgs,
+    options
+  ],
+  compiler/modules/[
+    modulegraphs
+  ],
+  compiler/sem/[
+    passes,
+    transf
+  ],
+  compiler/utils/[
+    debugutils
+  ],
+  compiler/vm/[
+    vmcompilerserdes,
+    vmdef,
+    vmjit,
+    vmops,
+    vmtypegen,
+    vm
+  ],
+  experimental/[
+    results
+  ]
+
+from compiler/ast/reports import wrap
+from compiler/ast/reports_vm import VMReport
+from compiler/ast/reports_sem import SemReport
+from compiler/ast/report_enums import ReportKind
+
+# to prevent endless recursion in macro instantiation
+const evalMacroLimit = 1000
+
+proc putIntoReg(dest: var TFullReg; c: var TCtx, n: PNode, formal: PType) =
+  ## Put the value that is represented by `n` (but not the node itself) into
+  ## `dest`. Implicit conversion is also performed, if necessary.
+  let t = formal.skipTypes(abstractInst+{tyStatic}-{tyTypeDesc})
+
+  # XXX: instead of performing conversion here manually, sem could generate a
+  #      small thunk for macro invocations that sets up static arguments and
+  #      then invokes the macro. The thunk would be executed in the VM, making
+  #      the code here obsolete while also eliminating unnecessary
+  #      deserialize/serialize round-trips
+
+  case t.kind
+  of tyBool, tyChar, tyEnum, tyInt..tyInt64, tyUInt..tyUInt64:
+    assert n.kind in nkCharLit..nkUInt64Lit
+    dest.ensureKind(rkInt, c.memory)
+    dest.intVal = n.intVal
+  of tyFloat..tyFloat128:
+    assert n.kind in nkFloatLit..nkFloat128Lit
+    dest.ensureKind(rkFloat, c.memory)
+    dest.floatVal = n.floatVal
+  of tyNil, tyPtr, tyPointer:
+    dest.ensureKind(rkAddress, c.memory)
+    # XXX: it's currently forbidden to pass non-nil pointer to static
+    #      parameters. `deserialize` already reports an error, so an
+    #      assert is used here to make sure that it really got reported
+    #      earlier
+    assert n.kind == nkNilLit
+  of tyOpenArray:
+    # Handle `openArray` parameters the same way they're handled elsewhere
+    # in the VM: simply pass the argument without a conversion
+    let typ = c.getOrCreate(n.typ)
+    dest.initLocReg(typ, c.memory)
+    c.serialize(n, dest.handle)
+  of tyProc:
+    # XXX: a hack required to uphold some expectations. For example,
+    #      `genEnumCaseStmt` would fail without this. Procedural types as
+    #      static macro arguments are underspecified
+    let pt =
+      if t.callConv == ccClosure and n.kind == nkSym:
+        # Force the location to be of non-closure type. This breaks other
+        # assumptions!
+        n.sym.typ
+      else:
+        t
+
+    let typ = c.getOrCreate(pt)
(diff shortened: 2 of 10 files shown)


saem:
Thanks, @zerbina. #484 can wait, it'll have to get updated a bunch because of the report splits anyhow.

compiler/sem/semstmts.nim
```diff
@@ -915,7 +915,7 @@ proc semNormalizedLetOrVar(c: PContext, n: PNode, symkind: TSymKind): PNode =
     if not hasError and sfCompileTime in v.flags:
       var x = newNodeI(result.kind, v.info)
       x.add producedDecl
-      vm.setupCompileTimeVar(c.module, c.idgen, c.graph, x)
+      setupCompileTimeVar(c.module, c.idgen, c.graph, x)
```
saem, line 918: It's so obvious when I read it now. The base VM shouldn't have been doing this all along and the code was telling us.

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+  std/[
+    tables
+  ],
+  compiler/ast/[
+    ast_types,
+    ast,
+    errorhandling,
+    errorreporting,
+    lineinfos,
+    reports,
```
saem: Could you narrow this reports import with a `from` as well?

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+
+# to prevent endless recursion in macro instantiation
+const evalMacroLimit = 1000
+
+proc putIntoReg(dest: var TFullReg; c: var TCtx, n: PNode, formal: PType) =
+  ## Put the value that is represented by `n` (but not the node itself) into
+  ## `dest`. Implicit conversion is also performed, if necessary.
+  let t = formal.skipTypes(abstractInst+{tyStatic}-{tyTypeDesc})
+
+  # XXX: instead of performing conversion here manually, sem could generate a
```
saem, line 65: This is a good little project for someone to dive into and learn.

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+  of tyNil, tyPtr, tyPointer:
+    dest.ensureKind(rkAddress, c.memory)
+    # XXX: it's currently forbidden to pass non-nil pointer to static
+    #      parameters. `deserialize` already reports an error, so an
+    #      assert is used here to make sure that it really got reported
+    #      earlier
+    assert n.kind == nkNilLit
+  of tyOpenArray:
+    # Handle `openArray` parameters the same way they're handled elsewhere
+    # in the VM: simply pass the argument without an conversion
```
```suggestion
    # in the VM: simply pass the argument without a conversion
```

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+
+  # execute new instructions; this redundant opcEof check saves us lots
+  # of allocations in 'execute':
+  if c.code[info.start].opcode != opcEof:
+    result = execute(c, info).unpackResult(c.config, n)
+  else:
+    result = c.graph.emptyNode
+
+proc evalExpr*(c: var TCtx, n: PNode): PNode =
+  # deadcode
```
saem: Mark with a deadcode pragma like debug/deprecated?

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+  #for i in 0..<regCount: tos.slots[i] = newNode(nkEmpty)
+  let cb =
+    if requiresValue:
+      mkCallback(c, r): c.regToNode(r, n.typ, n.info)
+    else:
+      mkCallback(c, r): newNodeI(nkEmpty, n.info)
+
+  result = execute(c[], start, tos, cb).unpackResult(c.config, n)
+
+  c.mode = oldMode
```
saem: Do we trust defer?

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+
+proc evalConstExpr*(module: PSym; idgen: IdGenerator; g: ModuleGraph; e: PNode): PNode =
```
saem: Inline?

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+  of tyStatic:
+    putIntoReg(reg, c, x, typ)
+  else:
+    var n = x
+    if n.kind in {nkHiddenSubConv, nkHiddenStdConv}: n = n[1]
+    # TODO: is anyone on the callsite dependent on this modifiction of `x`?
+    n.typ = x.typ
+    reg = TFullReg(kind: rkNimNode, nimNode: n)
+
+# XXX: why is this routine located here? It not related to the VM at all
```
saem: Could put in AST query?

Aside: got this far in my code review

saem (approved):
Minor changes, didn't get to review it all, but I'm already a big fan.

compiler/vm/compilerbridge.nim
zerbina: Yep, will do. That `report` import is a leftover from before rebasing.

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+    let genericParam = gp[i].sym
+    let posInCall = macroSym.typ.len + i
+    if posInCall < call.len:
+      yield (genericParam, call[posInCall])
+
+proc evalMacroCall*(module: PSym; idgen: IdGenerator; g: ModuleGraph; templInstCounter: ref int;
+                    n: PNode, sym: PSym): PNode =
+  #if g.config.errorCounter > 0: return errorNode(idgen, module, n)
+
+  # XXX globalReport() is ugly here, but I don't know a better solution for now
```
saem, line 273: The global errors could be pushed out as nkError, then have the can raise the global there, keep applying this process inductively (? we're going up so not recursion, right? 😄) and we end up handling/recovering at the right spot.

Also applies to the next global error me thinks.

To be super clear, actual work of making that change doesn't need to be in this PR, just capturing the idea assuming it makes sense.

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+  # put macro generic parameters into registers
+  let gp = sym.ast[genericParamsPos]
+  for i in 0..<gp.safeLen:
+    let idx = sym.typ.len + i
+    if idx < n.len:
+      setupMacroParam(tos.slots[idx], c[], n[idx], gp[i].sym.typ)
+    else:
+      # TODO: the decrement here is wrong, but the else branch is likely
+      #       currently not reached anyway
+      dec(g.config.evalMacroCounter)
```
saem, line 333: If we trust `defer` then we're all good and can drop this line and the one at the end?

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,476 @@
...
+
+  # return value:
+  tos.slots[0] = TFullReg(kind: rkNimNode, nimNode: newNodeI(nkEmpty, n.info))
+
+  # put macro call arguments into registers
+  for i in 1..<sym.typ.len:
+    setupMacroParam(tos.slots[i], c[], n[i], sym.typ[i])
+
+  # put macro generic parameters into registers
+  let gp = sym.ast[genericParamsPos]
```
saem, line 325: Throwaway remark: I no longer recall the mechanics but it's weird that we put the generic params after the regular.

compiler/vm/compilerbridge.nim
zerbina: Done. The comment in question was new, but the rest of the code is for the largest part just copied over from `vm`.

compiler/vm/compilerbridge.nim
zerbina: I removed it instead. It hasn't been in use for some time now, and I don't think this will change in the near future.

compiler/vm/compilerbridge.nim
zerbina: In the sense that it should be used here? I personally wouldn't do so, but am also not strongly against it.

compiler/vm/compilerbridge.nim
```diff
@@ -0,0 +1,458 @@
...
+      let r = execute(c, start, tos, cb)
+      result = r.unpackResult(c.config, c.graph.emptyNode)
+      reportIfError(c.config, result)
+      if result.isError:
+        result = nil
+  else:
+    localReport(c.config, sym.info):
+      VMReport(kind: rvmCallingNonRoutine, sym: sym)
+
+# XXX: the compilerapi regarding globals (getGlobalValue/setGlobalValue)
```
saem, line 414: Good

saem (approved):
Last pass done.

compiler/vm/compilerbridge.nim
zerbina: That's mainly a `vmgen` limitation. Normal parameters are always expected in register `param.position + 1`, so for macros, the generic parameters have to be placed after the normal ones.

compiler/vm/compilerbridge.nim
zerbina: Oh, I now see what you mean. The procedure can return early, in which case the previous `mode` wouldn't be restored. Yep, `defer` definitely makes sense then.

compiler/vm/compilerbridge.nim
zerbina: Actually, the `globalReport` for nested expansion is never reached -- `evalMacroCall` is not recursive. The only way for macros to expand other macros is via `getAst`, but that uses implemented via normal in-VM procedure calls, which already use their own mechanism for preventing recursion (albeit with a much higher limit).

I'm going to make a follow up PR that removes `evalMacroCounter` and the associated report.

PR #496, opened 2022-12-14
vm: clean up the external interface

zerbina:
## Summary

Invoking the VM was previously done through an `execute` procedure that was mainly meant for use by the compiler, as it had on-demand procedure code-gen (via `vmjit`) and VM-data deserialization built in. The final goal is to completely decouple the VM from the rest of compiler, and breaking this point of coupling is another step towards this goal.

To encapsulate the state of a single thread of execution, the `VmThread` object is introduced. `initVmThread` creates a new instance; `execute` runs a thread until it yields; and `dispose` cleans up the resources used by the thread and destroys it.

`ExecResult` is renamed to `YieldReason` and VM errors are now also reported through the `YieldReason` object. The previous `vmjit` integration as well the VM-data -> ``PNode`` translation are moved into a procedure in the `compilerbridge` module that is implemented on top of the `VmThread` primitive.

In addition, the `opcQuit` instruction no longer implements the quit logic itself, but yields execution with a dedicated reason, leaving how a `quit` call is to be handled up the VM's callsite. This is done as an experiment to see if a similar, but more generalized mechanism could in the future be used to replace the current VM callbacks.

As a direct consequence, the `vm` no longer imports `vmjit` (which imported `vmgen`).

## Details
- adjust the `vmrunner` to work with the VM API
- export `createStackTrace` and use a `VmThread` for providing the PC and stack frame position
- document issues with `opcGorge`, `opcEcho`, and the way ``quit`` works at compile-time

### DIFF
--- a/compiler/vm/compilerbridge.nim
+++ b/compiler/vm/compilerbridge.nim
@@ -35,7 +35,8 @@ import
     transf
   ],
   compiler/utils/[
-    debugutils
+    debugutils,
+    idioms
   ],
   compiler/vm/[
     vmcompilerserdes,
@@ -49,13 +50,26 @@ import
     results
   ]
 
-from compiler/ast/reports import wrap
+import std/options as std_options
+
+from compiler/ast/reports import wrap, toReportLineInfo
 from compiler/ast/reports_vm import VMReport
 from compiler/ast/reports_sem import SemReport
+from compiler/ast/reports_internal import InternalReport
 from compiler/ast/report_enums import ReportKind
+
+type
+  ExecErrorReport* = object
+    stackTrace*: VMReport ## The report storing the stack-trace
+    report*: VMReport     ## The report detailing the error
+
+  ExecutionResult* = Result[PNode, ExecErrorReport]
 
 # to prevent endless recursion in macro instantiation
 const evalMacroLimit = 1000
+
+# prevent a default `$` implementation from being generated
+func `$`(e: ExecErrorReport): string {.error.}
 
 proc putIntoReg(dest: var TFullReg; c: var TCtx, n: PNode, formal: PType) =
   ## Put the value that is represented by `n` (but not the node itself) into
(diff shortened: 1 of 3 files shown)


compiler/vm/compilerbridge.nim
```diff
@@ -49,14 +50,27 @@ import
...
+  ExecErrorReport* = object
+    stackTrace*: VMReport ## The report storing the stack-trace
+    report*: VMReport     ## The report detailing the error
+
+  ExecutionResult* = Result[PNode, ExecErrorReport]
+
 # to prevent endless recursion in macro instantiation
 const evalMacroLimit = 1000
 
+# prevent a default `$` implmenentation from being generated
```
```suggestion
# prevent a default `$` implementation from being generated
```

compiler/vm/compilerbridge.nim
```diff
@@ -138,6 +152,73 @@ proc unpackResult(res: sink ExecutionResult; config: ConfigRef, node: PNode): PN
...
+      # execution is finished
+      result.initSuccess cb(c, c.sframes[0].slots[r.reg.get])
+      break
+    of yrkError:
+      result.initFailure buildError(c, thread, r.error)
+      break
+    of yrkQuit:
+      case c.mode
+      of emRepl, emStaticExpr, emStaticStmt:
+        # XXX: should code run at compile time really be able to force-quit
```
saem, line 193: That seems... really really weird. I'm trying to think of a use case that's not awful and I can't. All I can think of is how it doesn't help composition.

compiler/vm/vm.nim
```diff
@@ -9,6 +9,13 @@
 
 ## This file implements the new evaluation engine for Nim code.
 ## An instruction is 1-3 int32s in memory, it is a register based VM.
+## # TODO: update the above -- it doesn' reflect reality anymore
```
```suggestion
## # TODO: update the above -- it doesn't reflect reality anymore
```

compiler/vm/vm.nim
```diff
@@ -9,6 +9,13 @@
...
+##
+## To use the VM once has to first set up an execution environment, which is
+## represented by ``TCtx``. In order to execute something, a ``VmThread``
```
saem, line 15: I kinda want to call it a 'process' instead of 'thread' as we might want some sort of green thread/continuation thing.

saem (approved):
Minor suggestions mostly.

I'm not entirely convinced about 'process' but I'm a big fan of how things are decoupling either way. This weekend is going to be really exciting to come back to the code. 🎉

compiler/vm/vm.nim
zerbina: I originally wanted to name it `VmInstance` (and before that, `ExecState`), but ultimately decided on `VmThread`.

In my opinion, 'process' sounds a bit heavy-weight for what the object actually represents (the execution state for a single thread), but I'm not strongly against renaming it to `VmProcess`.

saem: Let's leave it as is. We can always rename it later and have the compiler help us do it.

saem:
Bors r+

PR #488, opened 2022-12-05
compiler: improve `for`-loop tuple unpacking

zerbina:
## Summary
- fix an issue with `semForVar` where types were modified, leading to unrelated valid code being rejected by semantic analysis (see `tforloop_unpacking_bug.nim`)
- always use a full copy instead of a shallow copy for `for` variables
- refactor `semForVar`
- refactor `transformYield`
- make `semForVar` `nkError` aware

## Details
- fix the `nkHiddenAddr` nodes produced by `transformYield` having the wrong type (the type of the accessed tuple was used)
- split up `semForVar` into multiple procedures in order to reduce code duplication
- add debug-tracing calls to the `for`-loop related procedures
- improve the documentation of the refactored code
- add tests for `for`-loop tuple unpacking

Unconditionally using `nkFastAsgn` (i.e. a shallow assignment) for local variables is unsound without proving that doing so is safe (which did not happen). Figuring out where to use shallow copies should be left to the cursor inference step.

The `nkFastAsgn` change negatively impacts performance when of the compiled code when using `--gc:refc` (or other legacy GCs), because a full copy is materialized where it previously wasn't. This is not the case for `--gc:arc|orc`, as `nkFastAsgn` was/is treated the same as `nkAsgn` (i.e. as a full copy) for locations with destructors.

---

This PR depends on #487 to be merged first.

PR #497, opened 2022-12-15
tests: disable the `realtimeGC` test

zerbina:
## Summary

The test depends on the GC not running at the wrong time, which can happen if allocation behaviour changes (as is the case with https://github.com/nim-works/nimskull/pull/488)

This is caused by the GC instance used for the shared library being initialized with an incorrect stack-bottom value, leading to valid stack cells located below the wrong bottom not being considered during stack scanning. If a heap location is only referenced from said stack cells, it is garbage collected, making further access to it invalid and potentially causing crashes due to heap corruption.

For programs using both shared libraries and `refc`, the `nimrtl` (Nim run-time library) has to be used in order for there to only be a single GC instance per thread, but the test explicitly states that it tests the realtime GC *without* using the `nimrtl`. This can't work, and only did so in the past by accident.

The test is only disabled instead of removed, so that it can be, if desired, corrected in the future without losing the associated history.

### DIFF
--- a/tests/realtimeGC/tmain.nim
+++ b/tests/realtimeGC/tmain.nim
@@ -1,6 +1,7 @@
 discard """
   matrix: "--threads:on -d:release -d:useRealtimeGC"
   joinable:false
+  knownIssue: "the test dependens on the GC not running at the wrong time"
 """
 
 #[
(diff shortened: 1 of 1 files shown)


saem:
Bors r+

PR #498, opened 2022-12-17
sem/transf: don't collapse implict addr/deref nodes

zerbina:
## Summary
The collapsing produces unsound AST in the case that the operand of the
`nkHiddenDeref` doesn't have the same type as the `nkHiddenAddr` node.
For example, if the `nkHiddenDeref` dereferences a `ptr` or `ref` value,
the collapsing would result in a `ptr` or `ref` value to be directly
passed to `var` parameter (because `nkHiddenAddr` nodes are generated
when passing something to a `var` parameter), which is incorrect.

The reason that this didn't cause many problems so far is that `var`,
`ref`, and `ptr` are all pointers underneath for the C target, and that
`cgen` doesn't inspect the type of nodes in many cases.

The collapsing done directly in `semexpr` is removed and the one done in
`transf` is adjusted:
- only `DerefExpr( Addr(x) )` and `Addr( DerefExpr(x) )`  (both are
  expressions explicitly provided by the user) are collapsed -- the
  latter one only when `x` is of `ptr` type.
- since collapsing `HiddenDeref( HiddenAddr(x) )` is always known to be
  safe, the behaviour is kept. Because of a `cgen` issue, it's currently
  also required.

Unconditional collapsing of all address-of + deref expression is now
performed inside the C code-generator, because it is known that the
operands to all dereference operations map to C pointers at that point.

## Details
The compiler targets C89, where dereferencing a pointer requires a
pointer to complete type, even when the result is immediately passed to
an address-of operation (i.e. `&(*x)`). `cgen` now requests the complete
type whenever emitting a deref.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -823,13 +823,9 @@ proc hasUnresolvedArgs(c: PContext, n: PNode): bool =
     return false
 
 proc newHiddenAddrTaken(c: PContext, n: PNode): PNode =
-  case n.kind
-  of nkHiddenDeref:
-    checkSonsLen(n, 1, c.config)
-    result = n[0]
-  else:
-    result = newNodeIT(nkHiddenAddr, n.info, makeVarType(c, n.typ))
-    result.add n
+  ## Wraps the expression `n` in an ``nkHiddenAddr`` and assigns a ``var`` type
+  ## derived from the source type to the resulting expression
+  result = newTreeIT(nkHiddenAddr, n.info, makeVarType(c, n.typ)): n
 
 func isVarParam(t: PType): bool =
   # watch out: ``typeDesc[var int]`` is **not** a 'var' parameter
--- a/compiler/sem/transf.nim
+++ b/compiler/sem/transf.nim
@@ -461,36 +461,15 @@ proc transformYield(c: PTransf, n: PNode): PNode =
     for i, child in changeNode:
       child.info = changeNode.info
 
-proc transformAddrDeref(c: PTransf, n: PNode, a, b: TNodeKind): PNode =
+proc transformAddr(c: PTransf, n: PNode): PNode =
   result = transformSons(c, n)
-  var n = result
-  case n[0].kind
-  of nkObjUpConv, nkObjDownConv, nkChckRange, nkChckRangeF, nkChckRange64:
-    var m = n[0][0]
-    if m.kind == a or m.kind == b:
-      # addr ( nkConv ( deref ( x ) ) ) --> nkConv(x)
-      n[0][0] = m[0]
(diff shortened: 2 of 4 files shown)


compiler/sem/semexprs.nim
```diff
@@ -823,19 +823,15 @@ proc hasUnresolvedArgs(c: PContext, n: PNode): bool =
...
-    result = newNodeIT(nkHiddenAddr, n.info, makeVarType(c, n.typ))
-    result.add n
-    let aa = isAssignable(c, n)
-    if aa notin {arLValue, arLocalLValue}:
-      if aa == arDiscriminant and c.inUncheckedAssignSection > 0:
-        discard "allow access within a cast(unsafeAssign) section"
-      else:
-        result = c.config.newError(n, reportAst(rsemVarForOutParamNeeded, n))
+  ## Wraps the expression `n` in an ``nkHiddenAddr`` or, if the expression is
+  ## not a mutable lvalue, an ``nkError``
```
saem: Thanks for documenting it, as well.

compiler/sem/transf.nim
```diff
@@ -461,36 +461,15 @@ proc transformYield(c: PTransf, n: PNode): PNode =
...
-  else:
-    if n[0].kind == a or n[0].kind == b:
-      # addr ( deref ( x )) --> x
-      result = n[0][0]
-      if n.typ.skipTypes(abstractVar).kind != tyOpenArray:
-        result.typ = n.typ
+  let n = result
+  # collapse ``nkAddr( nkDerefExpr(x) )`` to ``x``, but only if ``x`` is of
+  # pointer type. Performing the collapsing when ``x`` is a ``ref`` would be
+  # incorrect, as there'd be a formal vs. actual type mismatch then
```
saem, line 469: Yup

saem (approved):
Looks good to me. Plenty of comments to understand and improve things over time.

saem:
Once tests pass let's merge.

haxscramper (approved):

zerbina:
Another detour is needed, it seems. The PR that fixed the JS code-generator issue that was the previous blocker also added a test for arguments to `var` parameters, which now fails with the VM back-end. This uncovered a large amount of issues with `vmgen`, so I'm going to fix them first.

Clyybber (approved):

zerbina:
I'm in the process of fixing the mentioned `vmgen` issues, but the fixes there depend on the `nkHiddenAddr`/`nkHiddenDeref` nodes to not be collapsed (which this PR implements).

In order to break the cyclic dependency, I disabled the `tvar_arguments` test here and am going to merge the `vmgen` fixes as a follow up.

bors r+

PR #499, opened 2022-12-18
update the project roadmap diagram

haxscramper:
Replace the older project fixup diagram with a new, revised version that more accurately depicts current state of the project planning.


saem:
- CPS comes much earlier, after untangling the C backend, fundamental runtime
- the JS backend needs a similar rewrite as the C backend, forced clarification native vs foreign semantics
- the above need to happen prior to C++ integration

saem:
There is a dialect removal, which is a stream. Don't have to include it this time around. It's more about ripping out implementation. This would either be permanently removed, replaced with an equivalent on by default, or have indirect/multiple-things that are seen as a "replacement".

It's ok if we leave that stuff as version next.

haxscramper:
> There is a dialect removal, which is a stream. Don't have to include it this time around.

There is a _"Clean up and improve implementation of the compiler, doc, stdlib. Remove dialects. Improve orthogonality."_ on the left side now.

PR #500, opened 2022-12-19
remove experimental implicit deref

haxscramper:
Remove implementation and documentation for the experimental implicit deref. The feature is removed largely because it is not used anywhere, not tested (the only two existing tests have been disabled for the last two years in `when false`) block and most importantly have subpar implementation.


haxscramper:
500th PR 🚀🎉

tests/implicit/timplicit.nim
```diff
@@ -29,21 +27,3 @@ var sp: PPValue = addr x
 sp.a = 2
 if sp.a == 2: echo 2  # with sp[].a the error is gone
```
saem: Is this test really testing anything anymore? Maybe we can kill the whole file.

saem (commented):
Noticed one thing, not a blocker. Thanks.

PR #501, opened 2022-12-20
vmbackend: fix crash related to `nkEmpty` nodes

zerbina:
## Summary
During both serialization and deserialization of `PNode`s used for the
VM's RTTI, `nkEmpty` nodes were not explicitly handled, leading to them
being treated as nodes with children. This caused a `FieldDefect` when
trying to access their (inaccessible) `sons` field.

The problem rendered a large amount of programs unusable with the VM
backend, as either the compiler or the runner would crash when RTTI that
contained empty nodes somewhere was used.

## Details
- add `nkError` to the `nkWithoutSons` set
- add the `nkWithSons` set
- make `case` statements in `storeNode` and `loadNode` exhaustive.
  `nkNone` and `nkError` are now also detected (and rejected)
- add a test to make sure `nkEmpty` nodes are handled correctly

---

As mentioned in the commit message, the issue blocked a lot of tests in the test suite from working with the VM target.

## Notes for Reviewers
* the `nkWithoutSons` set wasn't used until now, so extending it with `nkError` can't break any existing code

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1009,7 +1009,10 @@ const
     {nkStrLit..nkTripleStrLit} +
     {nkSym} +
     {nkIdent} +
+    {nkError} +
     {nkEmpty, nkNone}
+
+  nkWithSons* = {low(TNodeKind) .. high(TNodeKind)} - nkWithoutSons
 
 type
   EffectsCompat* = enum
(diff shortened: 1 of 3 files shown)


saem (approved):
Bors r+

Nice find!

PR #502, opened 2022-12-21
semexprs: refactor 'var' parameter argument fixup

zerbina:
## Summary
This is mostly a refactoring and clean up, but it also adjusts the "is
address taken" analysis to no longer ignore l-value conversions and
statement list expressions. The operand to an `addr` operation now also
uses the same (more correct) analysis as the arguments to `var`
parameters.

`jsgen` depends on the `sfAddrTaken` flag in order to know whether a
location needs to support being modified via a "base + index" (the JS
back-end's way of implementing pointers), so this change fixes a
JS code-gen issue. Only in theory however, as the JS code-generator
implements the relevant bits incorrectly -- although in slightly
different way, wrong code is still generated.

## Details
- rename `analyseIfAddressTakenInCall` to `fixVarArgumentsAndAnalyse`,
  refactor and document it, and make it `nkError` aware
- remove the mutability check from `newHiddenAddrTaken` -- this is now
  the responsibility of the callsite
- move the "is address taken" analysis to a separate procedure
- move the `FakeVarParams` constant to the `ast_query` module
- make `semAddrArg` `nkError` aware, fixing `addr x` reporting an
  `expression has no address` error if expression `x` had an error
- fix the "has no address" error reported for  `addr` always suggesting
  the usage of `unsafeAddr`, even if it was already used
- remove leftover and unused `unsafeAddr`-related logic

---

As mentioned in the code comments, I don't think `jsgen` depending on `sfAddrTaken` is a good idea. The analysis itself should also not happen as part of the after-call actions, but rather during `sempass2`. Both are things that are better implemented in separate PRs, however.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -207,6 +207,17 @@ const
     nkUsingStmt,
   }
 
+  FakeVarParams* = {mNew, mNewFinalize, mInc, mDec, mIncl, mExcl,
+    mSetLengthStr, mSetLengthSeq, mAppendStrCh, mAppendStrStr, mSwap,
+    mAppendSeqElem, mNewSeq, mReset, mShallowCopy, mDeepCopy, mMove,
+    mWasMoved}
+    ## An arguments to these magics never uses ``nkHiddenAddr``, even if the
+    ## corresponding parameter is a 'var' parameter. The reason for this is
+    ## that these magics are lowered into code that, because it gets inlined
+    ## directly, doesn't mutate the arguments through indirection. This also
+    ## implies that the "is address taken" analysis (see ``sfAddrTaken``) must
+    ## not be performed for arguments to these magics.
+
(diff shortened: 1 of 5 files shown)


saem:
Did a quick skim, I like it. The js gen dependency on `sfAddressTaken` is definitely very suss.

saem:
Bors r+

Thank you!

Inspired by this PR, we should discuss what the semantic meaning of various types, handles, locations all should mean. Maybe a good way to do that would be to attempt to make a table to describe it. Something like:

|scope|modifier|type|meaning|
|-------|------|-------|----|
|parameter|var|*|var applies to the location, arg must be a handle to a variable location|

The dimensions of the table and constiuents would be a good way to get things nailed down, surface unnamed or specified concepts, etc.

saem:
Bors r+

PR #503, opened 2022-12-21
jsgen: fix multiple `var` argument and `addr` issues

zerbina:
## Summary
- fix `addr x` and `p_var(x)` causing an internal error when `x` is a
  for-var or an lvalue-conversion expression (e.g. `int(val)`)
- fix incorrect JS code being generated for `addr (stmt; x)` and
  `p_var( (stmt; x) )`
- fix incorrect JS code being generated for `x.y` where `x` is an immutable
  parameter that is used somwhere as an argument to a `var` parameter
  or as the operand of an `addr`

## Details
- refactor, simplify, and document `genAddr`
- add an extensive test for `var` arguments
- extend the `taddr.nim` test

---

Depends on PR #502 and blocks #498 .

### DIFF
--- a/compiler/backend/jsgen.nim
+++ b/compiler/backend/jsgen.nim
@@ -1229,7 +1229,7 @@ proc genFieldAccess(p: PProc, n: PNode, r: var TCompRes) =
 
 proc genAddr(p: PProc, n: PNode, r: var TCompRes)
 
-proc genCheckedFieldOp(p: PProc, n: PNode, addrTyp: PType, r: var TCompRes) =
+proc genCheckedFieldOp(p: PProc, n: PNode, takeAddr: bool, r: var TCompRes) =
   internalAssert p.config, n.kind == nkCheckedFieldExpr
   # nkDotExpr to access the requested field
   let accessExpr = n[0]
@@ -1267,7 +1267,7 @@ proc genCheckedFieldOp(p: PProc, n: PNode, addrTyp: PType, r: var TCompRes) =
     setx.res, tmp, disc.loc.r, if negCheck: ~"!==" else: ~"===",
     makeJSString(msg), genTypeInfo(p, disc.typ))
 
-  if addrTyp != nil and mapType(p, addrTyp) == etyBaseIndex:
+  if takeAddr:
     r.typ = etyBaseIndex
     r.res = makeJSString($field.loc.r)
     r.address = tmp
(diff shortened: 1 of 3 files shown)


compiler/backend/jsgen.nim
```diff
@@ -1377,53 +1370,47 @@ proc genSymAddr(p: PProc, n: PNode, typ: PType, r: var TCompRes) =
...
+  ## Dispatches to the appropriate procedure for generating the address-of
+  ## operation based on the kind of node `n`
+  case n.kind
+  of nkSym:
+    genSymAddr(p, n, r)
+  of nkCheckedFieldExpr:
+    genCheckedFieldOp(p, n, takeAddr=true, r)
+  of nkDotExpr:
+    genFieldAddr(p, n, r)
+  of nkBracketExpr:
```
saem, line 1382: Thinking or loud: This bit really highlights how the AST is just off, it really should be lowered/specialized further. The type based dispatch within the branch seems off.

compiler/backend/jsgen.nim
```diff
@@ -2620,18 +2607,23 @@ proc gen(p: PProc, n: PNode, r: var TCompRes) =
   of nkObjConstr: genObjConstr(p, n, r)
   of nkHiddenStdConv, nkHiddenSubConv, nkConv: genConv(p, n, r)
   of nkAddr, nkHiddenAddr:
-    if n.typ.kind in {tyLent}:
+    if mapType(n.typ) != etyBaseIndex:
+      # the operation doesn't produce an addres-like value (e.g. because the
```
```suggestion
      # the operation doesn't produce an address-like value (e.g. because the
```

tests/lang_types/var/tvar_arguments.nim
```diff
@@ -0,0 +1,200 @@
+discard """
+  target: "c js vm"
```
saem, line 2: IIRC the lang category runs on all targets by default.

tests/lang_types/var/tvar_arguments.nim
```diff
@@ -0,0 +1,200 @@
...
+  labels: "var_arg"
+  description: '''
+    Tests to make sure arguments are properly passed to var parameter and that
```
saem, line 5: These tests are freaking awesome.

saem (commented):
A+, would try merging.

tests/lang_types/var/tvar_arguments.nim
zerbina: Yep, the `lang` category does, but the `lang_xxx` ones do not.

PR #504, opened 2022-12-25
generic graph data structure implementation

haxscramper:
Added implementation of generic graph data structure. The implementation is targeting DOD-oriented code: values are stored outside of the graph IDs are used instead  of vertices. Edge  information is stored in  sparse adjacency matrix.

Implementation also  provides basic graph data  structure algorithms: cycle detection, iterative BFS and DFS,  topological sorting. More algorithms can be added later.

This is an initial version, up for discussion/review -- currently only basic algorithmic toolbox is provided.

### DIFF
--- a/lib/experimental/graph.nim
+++ b/lib/experimental/graph.nim
@@ -1,0 +1,397 @@
+## Generic implementation of graph data structure
+
+import std/[
+  tables,
+  strformat,
+  deques,
+  hashes,
+  options,
+  math,
+  lenientops,
+  strutils,
+  sequtils,
+  algorithm,
+  intsets,
+  sets
+]
+
+type
+  GraphEdge*[Id] = tuple[src, dst: Id]
+
+  Path*[Id] = seq[Id]
+  Graph*[Id] = object
+    isDirected*: bool
+    edges: OrderedTable[Id, seq[Id]]
+
+type
+  GraphError* = ref object of CatchableError
+  GraphCyclesError* = ref object of GraphError
+
+func newGraph*[Id](isDirected: bool = true): Graph[Id] =
+  Graph[Id](isDirected: isDirected)
+
+func newGraph*[Id](
+    values: openarray[(Id, Id)], isDirected: bool = true): Graph[Id] =
+  result = newGraph[Id](isDirected)
+  for (src, dst) in values:
+    result.edges.mgetOrPut(src, @[]).add dst
+
+func default*[Id](graph: typedesc[Graph[Id]]): Graph[Id] =
+  newGraph[Id]()
+
+func contains*[Id](graph: Graph[Id], node: Id): bool =
+  node in graph.edges
+
+
+func contains*[Id](graph: Graph[Id], edge: GraphEdge[Id]): bool =
+  result = (
+    ( edge.src in graph.edges and
+      edge.dst in graph.edges[edge.src] ) or
+    ( not graph.isDirected and
+      edge.dst in graph.edges and
+      edge.src in graph.edges[edge.dst] )
+  )
+
+func getEdge*[Id](graph: Graph[Id], source, target: Id): GraphEdge[Id] =
+  if source.id in graph.structure.outgoingIndex:
+    for outEdge in graph.structure.outgoingIndex[source.id]:
+      if graph.structure.edgeMap[outEdge.id][1] == target:
+        return outEdge
+
+  raise newException(ValueError, "No edge for source/target pair")
+
+func edgeCount*[Id](graph: Graph[Id]): int =
+  for node, targets in pairs(graph.edges):
+    result += targets.len()
+
+func nodeCount*[Id](graph: Graph[Id]): int =
+  graph.edges.len()
+
+iterator nodes*[Id](graph: Graph[Id]): Id =
+  for i, _ in pairs(graph.edges):
+    yield i
+
+proc allNodes*[Id](graph: Graph[Id]): seq[Id] =
+  for node in nodes(graph):
+    result.add node
+
+iterator edgePairs*[Id](graph: Graph[Id]): GraphEdge[Id] =
+  for node, targets in pairs(graph.edges):
+    for target in targets:
+      yield (node, target)
+
+func computeInDegrees*[Id](graph: Graph[Id]): CountTable[Id] =
+  for (src, dst) in edgePairs(graph):
+    result.inc dst
+
+func inDeg*[Id](graph: Graph[Id], node: Id): int {.inline.} =
+  ## Return in degree for @arg{node}
+  result = 0
+  for (_, dst) in edgePairs(graph):
+    if dst == node:
+      inc result
+
+func outDeg*[Id](graph: Graph[Id], node: Id): int =
+  if node in graph.edges:
+    return graph.edges[node].len()
+
+func newId[Id](graph: var Graph[Id]): int {.inline.} =
+  result = graph.structure.maxId
+  inc graph.structure.maxId
+
+proc removeNode*[Id](graph: var Graph[Id], node: Id) =
+  ## Remove `node` from graph.
+  ##
+  ## Note that all edges that `node` is part of are removed too.
+  graph.edges.del node
+  for targets in mitems(graph.edges[node]):
+    targets.excl node
+
+proc removeEdge*[Id](graph: var Graph[Id], edge: GraphEdge[Id]) =
+  ## Remove `edge` from graph. Note that starting/ending nodes will not be
+  ## removed.
+  if edge.src in graph.edges:
+    graph.edges[edge.src].del graph.edges[edge.src].find(edge.dst)
+  # QUESTION error on missing edge?
+
+iterator outEdges*[Id](graph: Graph[Id], source: Id): GraphEdge[Id] =
+  ## Iterate over outgoing edges for `source`
+  if source in graph.edges:
+    for it in graph.edges[source]:
+      yield (source, it)
+
+iterator inEdges*[Id](graph: Graph[Id], target: Id): GraphEdge[Id] =
+  ## Iterate over ingoing edges for `target`
+  for (src, dst) in graph.edgePairs():
+    if dst == target:
+      yield (src, src)
+
+proc `$`*[Id](graph: Graph[Id]): string =
+  var count: int = 0
+  var map: seq[(string, string, string)]
+  for edge in edgePairs(graph):
+    map.add(($graph.sourceVal(edge), $graph[edge], $graph.targetVal(edge)))
+
+  map = map.sortedByIt(it[0])
+  var sourcew = 0
+  for it in map: sourcew = max(sourcew, it[0].len)
+
+  var edgew = 0
+  for it in map: edgew = max(edgew, it[1].len)
+
+  for idx, line in map:
+    if 0 < idx:
+      result.add "\n"
+
+    result.add alignLeft(line[0], sourcew)
+    result.add " -["
+    result.add alignLeft(line[1], edgeW)
+    result.add "]-> "
+    result.add line[2]
+
+
+
+iterator depthFirst*[Id](
+    graph: Graph[Id], root: Id): tuple[id: Id, depth: int] =
+  ## Perform depth-first iteration of **mutable** nodes in graph, starting
+  ## from `node`
+  var
+    visited: HashSet[Id]
+    stack: seq[seq[Id]]
+    yielded: HashSet[Id]
+
+  stack.add @[root]
+
+  while 0 < len(stack) and 0 < len(stack[^1]):
+    let top = stack[^1].pop
+
+    if top notin yielded:
+      yield (top, stack.len())
+      yielded.incl top
+
+    var buf: seq[Id]
+    for (src, node) in graph.outEdges(top):
+      if node notin visited:
+        visited.incl node
+        buf.add node
+
+    if stack[^1].len() == 0:
+      discard stack.pop()
+
+    if 0 < len(buf):
+      stack.add buf
+
+iterator breadthFirst*[Id](graph: Graph[Id], root: Id): Id =
+
+  ## Perform breadth-first iteration of parent graph, starting from `node`
+  var que = initDeque[Id]()
+  que.addLast(root)
+  var visited: HashSet[Id]
+  while 0 < len(que):
+    let top = que.popFirst()
+    yield top
+    for (src, target) in graph.outEdges(top):
+      if target notin visited:
+        visited.incl target
+        que.addLast(target)
+
+
+proc getNoIncoming*[Id](graph: Graph[Id]): seq[Id] =
+  ## Get all nodes with zero in degree of zero
+  let degrees = computeInDegrees(graph)
+  for node in graph.nodes:
+    if degrees[node] == 0:
+      result.add node
+
+proc topologicalOrdering*[Id](
+    graph: Graph[Id], start: openarray[Id] = []): seq[Id] =
+  ## Return graph nodes in topological ordering if possible. Otherwise (if
+  ## graph contains cycles raise `GraphCyclesError`)
+  var active = toHashSet(start)
+  if active.len() == 0:
+    active = toHashSet(getNoIncoming(graph))
+
+  var graph = graph
+
+  while 0 < len(active):
+    let node = active.pop
+    result.add node
+    for (src, target) in graph.outEdges(node):
+      graph.removeEdge((src, target))
+      if graph.inDeg(target) == 0:
+        active.incl target
+
+  if 0 < graph.edgeCount:
+    raise GraphCyclesError(
+      msg: "Topological sorting is impossible, graph contains cycles")
+
+
+template findMaxIt*(s: untyped, op: untyped): untyped =
+  var res: int = 0
+  var idx = 0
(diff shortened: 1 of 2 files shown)


tests/stdlib/algorithm/tgraph.nim
```diff
@@ -0,0 +1,101 @@
+import std/unittest
+import experimental/graph
+import std/osproc
+import std/sets
+import std/algorithm
+import std/sequtils
+
+type Id = int
```
saem:
```suggestion
## tests for a generic graph data structure

type Id = int  ## generic graph will use this type for testing
```

tests/stdlib/algorithm/tgraph.nim
```diff
@@ -0,0 +1,101 @@
...
+
+proc arrangeRepr(graph: HGraph[Id]) =
```
saem: What's this for?

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+
+  proc printCycle(vertex: Id) =
+    var stack2: seq[Id]
+    stack2.add(stack.pop())
+
+    while stack2[^1] != vertex:
+      stack2.add(stack.pop())
+
+    var path: seq[Id]
+    while 0 < len(stack2):
```
saem, line 285: I get the idea behind putting the constant item first when it's an assignment to avoid errors in other languages. I find I have to continually flip it around in my head and then see if it makes sense. It's just not how most folks are conditioned to think.

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+
+#[
+checklist
+
+- [ ] All implementations work both at compile and run-time
+- [ ] Works on JS target
+- [ ] All iterator algorithms have both mutable and immutable implementations
+- [ ] As little side effects as possible
+- [ ] Make node and edge ID's distinct integers to avoid messing them up
+- [ ] Iterateo over node value pairs (source, target: N, edge: E)
```
saem:
```suggestion
- [ ] Iterator over node value pairs (source, target: N, edge: E)
```

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+  lenientops,
+  strutils,
+  sequtils,
+  algorithm,
+  intsets,
+  sets
+]
+
+#[
+checklist
```
saem: Great checklist

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+  cycles (`0 -> 1 -> 0` and `1 -> 2 -> 1`)
+
+]#
+
+type
+  HGraphProperty* = enum
+    gpDirected
+    gpUndirected
+    gpAllowSelfLoops
+    gpIgnoreSelfLoops
```
saem:
```suggestion
    gpSelfLoopsAllowed
    gpSelfLoopsIgnored
```

Not strongly suggesting this, but it doesn't keep the tense parallel with the other options.

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+- IDEA :: Add `->` and `<-` as operators that would return `TempEdge`
+  object that can be added via `graph.add source -> target`.
+
+- TODO :: Test if cycle detection works for graphs that contain overlapping
+  cycles (`0 -> 1 -> 0` and `1 -> 2 -> 1`)
+
+]#
+
+type
+  HGraphProperty* = enum
```
saem: These seem like there are two mutually exlusive choices, should these be split into two enums?

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+  HGraph*[Id] = object
+    flags: set[HGraphProperty]
+    edges: OrderedTable[Id, seq[Id]]
+
+type
+  HGraphError* = ref object of CatchableError
+  HGraphCyclesError* = ref object of HGraphError
+
+func newHGraph*[Id](
+    properties: set[HGraphProperty] = {gpDirected, gpAllowSelfLoops}
```
saem:
```suggestion

const defaultProperties = {gpDirected, gpAllowSelfLoops}

func newHGraph*[Id](
    properties: set[HGraphProperty] = defaultProperties
```

That might be nicer

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+
+func newHGraph*[Id](
+    values: openarray[(Id, Id)],
+    properties: set[HGraphProperty] = {gpDirected, gpAllowSelfLoops}
+  ): HGraph[Id] =
+  result = newHGraph[Id](properties)
+  for (src, dst) in values:
+    result.edges.mgetOrPut(src, @[]).add dst
+
+func default*[Id](graph: typedesc[HGraph[Id]]): HGraph[Id] =
```
saem: maybe `empty` instead of `default`?

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+  newHGraph[Id]()
+
+func contains*[Id](graph: HGraph[Id], node: Id): bool =
+  node in graph.edges
+
+
```
saem, line 45:
```suggestion
```

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+    result += targets.len()
+
+func nodeCount*[Id](graph: HGraph[Id]): int =
+  graph.edges.len()
+
+iterator nodes*[Id](graph: HGraph[Id]): Id =
+  for i, _ in pairs(graph.edges):
+    yield i
+
+proc allNodes*[Id](graph: HGraph[Id]): seq[Id] =
```
saem: Generally recommendation keep the lightweight accessors and traversals (iterators) up top. Then the "heavier" procs like `allNodes` following those. Finally the mutators and "scary" stuff following that.

Typically, when learning/first using a data structure module folks are:
1. How do I make one, easy mode? (unless they already have one then start at the next step)
2. I got one, now what can I do with it? (reads, accessors, traversals)
3. How do I put this inside?
4. OK now more advanced bits

Of course, dependencies between routines might not allow following this 'perfectly'

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+  var map: seq[(string, string, string)]
+  for edge in edgePairs(graph):
+    map.add(($graph.sourceVal(edge), $graph[edge], $graph.targetVal(edge)))
+
+  map = map.sortedByIt(it[0])
+  var sourcew = 0
+  for it in map: sourcew = max(sourcew, it[0].len)
+
+  var edgew = 0
+  for it in map: edgew = max(edgew, it[1].len)
```
saem, line 140: QQ: would `foldl` work or does it whine about types?

`let edgew = map.foldl(max(a, b[1].len), 0)

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+
+      elif vertex in stack:
+        printCycle(vertex)
+
+      elif vertex notin visited:
+        stack.add vertex
+        processDFS()
+
+
+    for (_,vertex) in outEdges(graph, top):
```
saem:
```suggestion
    let top = stack[^1]

    template body(vertex: untyped): untyped {.dirty.} =
      if vertex == top and ignoreSelf:
        discard
      elif vertex in stack:
        printCycle(vertex)
      elif vertex notin visited:
        stack.add vertex
        processDFS()

    for (_,vertex) in outEdges(graph, top):
```

The idea with the above is the space between the let and template declaration interrupt reading flow to emphasize we're changing gears somewhat and defining a helper. Keeping the if/elif branches tightly packed indicates this is one logical unit of templated code. When a break (blank line) is see it's clear we've returned back to our regularly scheduled program of proc declaration. Two new lines as in other parts of this file typically hint at breaking up whole routines. We don't have braces so all this subtlety is in the spacing, fun!

tests/stdlib/algorithm/tgraph.nim
```diff
@@ -0,0 +1,101 @@
...
+
+  return result.sorted()
+
+suite "Graph API":
+  test "Iterators":
+    check dfs({0: 1}) == @[0, 1]
+    check dfs({0: 1, 1: 0}, 0) == @[0, 1]
+    # Using ordered tables and `seq[]` so graph does not loose information
+    # about node insertion order. This helps stable testing and makes graph
+    # structure easyer to understand.
```
saem:
```suggestion
    # Using ordered tables and `seq[]` so graph does not lose information
    # about node insertion order. This helps stable testing and makes the
    # graph structure easier to understand.
```

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+
+type
+  HGraphProperty* = enum
+    gpDirected
+    gpUndirected
+    gpAllowSelfLoops
+    gpIgnoreSelfLoops
+
+  HEdge*[Id] = tuple[src, dst: Id]
+  HGraph*[Id] = object
```
saem: What does the `H` in `HGraph` mean?

saem (commented):
Did a quick first past.

tests/stdlib/algorithm/tgraph.nim
haxscramper: For debugging, I will move it to tests later on, `graph-easy` allows to layout the graph in terminal from the textual description. Graphviz repr could be added as well, debugging graph data structures and algorithms usually involves some kind of visualization, very often, so I decided to include this as well.

lib/experimental/graph.nim
haxscramper: I actually thought about making several boolean fields in graph instead, so I guess it  really is unclear what enums are doing here

lib/experimental/graph.nim
haxscramper: `default` also exists for other types

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+iterator nodes*[Id](graph: HGraph[Id]): Id =
+  for i, _ in pairs(graph.edges):
+    yield i
+
+proc allNodes*[Id](graph: HGraph[Id]): seq[Id] =
+  for node in nodes(graph):
+    result.add node
+
+iterator edgePairs*[Id](graph: HGraph[Id]): HEdge[Id] =
+  for node, targets in pairs(graph.edges):
```
zerbina, line 79: There's nothing that can be done about it from here, but just as a general warning: the `pairs` iterator for all table types performs full copies for each element's key and pair.

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+    for target in targets:
+      yield (node, target)
+
+func computeInDegrees*[Id](graph: HGraph[Id]): CountTable[Id] =
+  for (src, dst) in edgePairs(graph):
+    result.inc dst
+
+func inDeg*[Id](graph: HGraph[Id], node: Id): int {.inline.} =
+  ## Return in degree for @arg{node}
+  return computeInDegrees(graph)[node]
```
zerbina:
```suggestion
  result = 0
  for (_, dst) in edgePairs(graph):
    if dst == node:
      inc result
```
This way, no temporary `CountTable` is allocated.

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+func isDirected*[Id](graph: HGraph[Id]): bool {.inline.} =
+  gpDirected in graph.structure.properties
+
+
+proc removeNode*[Id](graph: var HGraph[Id], node: Id) =
+  ## Remove `node` from graph.
+  ##
+  ## Note that related edges will be removed too, if they completely loose
+  ## all related nodes. So calling `removeId("a -> b", "b")` will not only
+  ## remove node `"b"`, but also an edge.
```
zerbina:
```suggestion
  ## Note that all edges that `node` is part of are removed too.
```
This describes the behaviour better, I think.

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+  for targets in mitems(graph.edges[node.id]):
+    targets.excl node.id
+
+  graph.edges.del node.id
```
zerbina:
```suggestion
  graph.edges.del node.id

  for targets in mitems(graph.edges[node.id]):
    targets.excl node.id
```
One less seq to scan.

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+
+proc getNoIncoming*[Id](graph: HGraph[Id]): seq[Id] =
+  ## Get all nodes with zero in degree of zero
+  let degrees = computeInDegrees(graph)
+  for node in graph.nodes:
+    if degrees[node] == 0:
+      result.add node
+
+proc topologicalOrdering*[Id](
+    graph: HGraph[Id], start: seq[Id] = @[]): seq[Id] =
```
zerbina:
```suggestion
    graph: HGraph[Id], start: openArray[Id] = []): seq[Id] =
```
This allows for providing the start nodes as an array without requiring an allocation.

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+        visited.incl target
+        que.addLast(target)
+
+
+proc getNoIncoming*[Id](graph: HGraph[Id]): seq[Id] =
+  ## Get all nodes with zero in degree of zero
+  let degrees = computeInDegrees(graph)
+  for node in graph.nodes:
+    if degrees[node] == 0:
+      result.add node
```
zerbina:
```suggestion
proc getNoIncoming*[Id, T](graph: HGraph[Id], result: var T) =
  ## Appends all nodes with a zero of degree to `result`
  let degrees = computeInDegrees(graph)
  for node in graph.nodes:
    if degrees[node] == 0:
      mixin add
      result.add node
```
This allows one to reuse containers, and also removes the need for extra conversion, such as the one in `topologicalOrdering`.

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+      "->",
+      name(dst)
+    ]
+
+proc findCycles*[Id](
+    graph: HGraph[Id],
+    ignoreSelf: bool = false,
+    overrideDirected: bool = false,
+    yieldMultiedge: bool = false
+  ): seq[seq[Id]] =
```
zerbina: Maybe add a `Path[Id] = seq[Id]` type and then return `seq[Path[Id]]` here for documentation purposes?

lib/experimental/graph.nim
```diff
@@ -0,0 +1,428 @@
...
+
+  var
+    visited: HashSet[Id]
+    stack: seq[Id]
+    resCycles: seq[seq[Id]]
+
+  proc printCycle(vertex: Id) =
+    var stack2: seq[Id]
+    stack2.add(stack.pop())
```
zerbina, line 279: Using `stack` and `resCycles` in `printCycle` means that both locals are allocated on the heap, and that both `printCycle` and `processDfs` are real closures. I think explicitly passing both locals as `var` arguments would be better, but maybe `findCycles` is not used often enough for this kind of optimization to be relevant.

zerbina (commented):
I also did a quick review and left some suggestions for improvements. None of them are blockers, so feel free to apply them how you see fit.

---

The `topologicalOrdering` procedure looks like it will run into issues because of the `removeEdge` inside the `outEdges` loop (i.e. edges are removed while iterating over them). I tried implementing a version that doesn't have that problem, and here's what I came up with:

```nim
proc topologicalOrdering*[Id](
    graph: HGraph[Id], start: seq[Id] = @[]): seq[Id] =
  var active = start
  if active.len == 0:
    active = getNoIncoming(graph)

  var
    i = 0
    numEdges = 0 # the number of visited edges
    seen: CountTable[Id]
  while i < active.len:
    let node = active[i]
    var reused = false
    result.add node

    for (src, target) in graph.outEdges(node):
      let remaining = seen.getOrDefault(target, -1)
      if remaining == -1:
        seen[target] = graph.inDeg(target) - 1
      else:
        seen.inc(target, -1)

      if remaining == 1:
        # this is the last remaining edge -> add the node to the queue
        if active[i] == node:
          # reuse the current slot
          active[i] = target
          reused = true
        else:
          active.add target

    # only increment `i` if the current slot was not reused:
    i += ord(not reused)

  if numEdges != edgeCount(graph):
    raise HGraphCyclesError(
      msg: "Topological sorting is impossible: graph contains sub-graphs unreachable from `start`")
  else result.len != graph.nodeCount:
    let connected = 0 # the number of nodes with edges
    for targets in 0..<graph.edges.items:
      if targets.len > 0:
        inc connected

    # perform the test again, but ignore freestanding nodes this time
    if connected != result.len:
      raise HGraphCyclesError(
        msg: "Topological sorting is impossible: graph contains cycles")
```
Disclaimer: I haven't tested it yet.

No copy of the graph or mutations thereof are required and `inDeg` is called only once for each node.

lib/experimental/graph.nim
haxscramper: Graph only stores lightweight IDs so this won't be too much of a problem

lib/experimental/graph.nim
haxscramper: sets use `.incl`, not `.add`

PR #509, opened 2023-01-02
vmgen: move `mIs` folding into `semfold`

zerbina:
## Summary
The logic is copied verbatim, and while it is still incorrect, it is now
applied earlier and no longer reaches into the code-generator. In other
words, the issue (type expression that is erroneously not materialized)
is worked around at a point that is closer (in the time sense) to the
origin of the problem than it was previously.

---

While it would be better to fix the underlying issue, I currently don't have the bandwidth to do so. Moving the folding to `semfold` means that `mirgen` doesn't have to employ special handling for the `mIs` magic.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


saem (approved):
Yup, this is definitely closer. There are a lot of big changes about to land so the fundamental fix for this can wait its turn. 😁

PR #510, opened 2023-01-02
sem: refactor `semConstExpr`

zerbina:
## Summary
- split the evaluation part into a separate procedure (`evalConstExpr`)
- use `nkError` nodes for reporting errors from `evalConstExpr`
- for a `const` definition, the expression is now not evaluated if
  its type is illegal in a `const` context

## Details
- `nkError` nodes are now prevented from reaching `vmgen`
- the `typeAllowed` check doesn't reject `ref` types for `const`s
  anymore
- disallowed non-nil values (such as for `ref`s) are now rejected via
  `ensureNilOnly` (which also looks at the *values*, not only at the
   *types*)
---

## Notes for Reviewers
* the main motivation here is to prevent `nkError` nodes from reaching `vmgen`. If they do, they'd have to be passed through `mirgen`, which is not intended to be possible.

<!-- Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -458,10 +458,14 @@ proc fixupTypeAfterEval(c: PContext, evaluated, eOrig: PNode): PNode =
 
 proc tryConstExpr(c: PContext, n: PNode): PNode =
   addInNimDebugUtils(c.config, "tryConstExpr", n, result)
-  var e = semExprWithType(c, n)
-  if e == nil:
+  let e = semExprWithType(c, n)
+  if e.isError:
     return
 
+  # XXX: ``getConstExpr`` can and does report errors via ``localReport``. In
+  #      case it happens, ``tryConstExpr`` doesn't return nil but compilation
+  #      fails. This behaviour does make sense, but non-vmgen error should
+  #      also not be ignored here then
   result = getConstExpr(c.module, e, c.idgen, c.graph)
   if result != nil: return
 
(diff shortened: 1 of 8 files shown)


compiler/sem/sem.nim
```diff
@@ -474,52 +478,67 @@ proc tryConstExpr(c: PContext, n: PNode): PNode =
...
+proc evalConstExpr(c: PContext, n: PNode): PNode =
+  ## Tries to turn the expression `n` into AST that represents a concrete
+  ## value. If this fails, an `nkError` node is returned
+  addInNimDebugUtils(c.config, "evalConstExpr", n, result)
+  assert not n.isError
+
+  # this happens when the overloadableEnums is enabled. We short-circuit
+  # evaluation in this case, as neither ``vmgen`` nor ``semfold`` know what to
+  # do with the sym-choice
+  # XXX: this doesn't seem like a good idea. Maybe report an error instead?
```
saem: This silly bit is in const, let, var, initialization expression analysis, as well. I can't help but feel that should have been handled earlier. I just can't picture the solution quite yet. Something like symbolic type based evaluation should have been done which either resolves the choice or all this is in a higher level context (eg: generic), where ambiguity isn't an issue.

The levels thing is taking a bunch of my brain space and my mind is only trying to see a solution through that lens at the moment. 😬

compiler/sem/semstmts.nim
```diff
@@ -1080,14 +1080,26 @@ proc semNormalizedConst(c: PContext, n: PNode): PNode =
...
-    typ = typeAllowedOrError(typ, skConst, c, def, typFlags)
+  if c.matchedConcept != nil:
+    typFlags.incl taConcept
+
+  # semantically analyse the expression first, so that it can be rejected
+  # early based on the type before trying to evaluate it
+  def = semExprWithType(c, def)
+  typ = typeAllowedOrError(typ, skConst, c, def, typFlags)
+
+  # XXX: ``fitNode`` is never used here. That doesn't seem correct, as the
```
saem: I agree that looks to be an oversight, I think in many cases it'd to unnoticed, but it does seem like a bug.

compiler/sem/semstmts.nim
```diff
@@ -1080,14 +1080,26 @@ proc semNormalizedConst(c: PContext, n: PNode): PNode =
...
+  #      type of the expression is used, no matter what
+  if not typ.isError:
+    let e = evalConstExpr(c, def)
+    if e.isError:
+      # XXX: an error node can't be used as the def for a ``const``, as error
+      #      nodes are not considered when inlining constants. If there occurrs
+      #      an error during analysis of the ``const``, the resulting symbol
+      #      should likely be an ``skError`` symbol
+      localReport(c.config, e)
```
saem: If we keep the error and not report it eagerly then we can propagate that into the productions below and register an skError.

compiler/vm/vmcompilerserdes.nim
```diff
@@ -40,7 +40,10 @@ import
 # xxx: reports are a code smell meaning data types are misplaced
 from compiler/ast/reports_sem import SemReport,
   reportAst
-from compiler/ast/report_enums import ReportKind
+from compiler/ast/report_enums import ReportKind, repVM
```
saem: Damn, so close to removing it. 😅

I'll likely catch it in the rebase.

compiler/vm/vmcompilerserdes.nim
```diff
@@ -66,30 +69,36 @@ proc deserialize(c: TCtx, m: VmMemoryRegion, vt: PVmType, formal, t: PType, info
 
 proc deserializeObject(c: TCtx, m: VmMemoryRegion, vt: PVmType, f, con: PType, info: TLineInfo): PNode
 
-proc deserializeRef*(c: TCtx, slot: HeapSlotHandle, vt: PVmType; f, con: PType, info: TLineInfo): PNode =
+proc deserializeRef(c: TCtx, slot: HeapSlotHandle, vt: PVmType; f, con: PType, info: TLineInfo): PNode =
```
saem: Yay, smaller surface

compiler/vm/vmgen.nim
```diff
@@ -2463,10 +2454,6 @@ proc gen(c: var TCtx; n: PNode; dest: var TDest; flags: TGenFlags = {}) =
   when defined(nimCompilerStacktraceHints):
     setFrameMsg c.config$n.info & " " & $n.kind & " " & $flags
   case n.kind
-  of nkError:
-    # XXX: do a better job with error generation
-    fail(n.info, vmGenDiagCannotGenerateCode, n)
```
saem, line 2485: Cool, these moved out.

saem (approved):
This is better, I think for the normalized const semantic analysis it should be possible to continue without reporting and produce an nk/SK error instead.

compiler/sem/semstmts.nim
zerbina: This comment was an oversight by me. When splitting the evaluation part from `semConstExpr` and replacing its usage in `semNormalizedConst`, I also added the call to `semExprWithType` in order to mirror the behavior of `semConstExpr` also does.

Since a call to `fitNode` has to happen *after* a `semExprX` and there was none following the call to `semConstExpr`, I made the (incorrect) assumption that `fitNode` wasn't used. A few lines above however, the rhs is both analysed *and* fitted. In other words, there was no issue (except maybe that the already semantically analysed expression was analysed again).

compiler/sem/semstmts.nim
zerbina: Yep, but no `skError` symbol is produced right now. Since the logic in `semNormalizedConst` is fairly complex and this PR already does a good amount of other things besides refactoring `semConstExpr`, my thinking was that it'd be better to eagerly report the error for now.

zerbina:
I reverted disallowing all non-nil `ref` values crossing the VM/compiler border again. Too many tests no longer pass then, and this PR is mainly meant as a refactoring. Instead, making sure that there are no non-nil `ref` values (non-nil closure environments are still allowed for compatibility) part of a `const` now happens in `semNormalizedConst`.

@saem If you're okay with the new changes, feel free to merge.

saem:
bors r+

This should be relatively easy to rebase... but I am putting a moritorum on big changes that bring in a lot of absolute hot garbage that is "Reports", until there is a strategy to get us out of this quagmire. I'll post more in the relevant PR and throw up a message in chat.

zerbina:
My intention here was not to make things harder than they already are, so I've converted this PR to a draft for now. The part that is a blocker for the MIR PR (`nkError` nodes reaching into `vmgen`) can be be worked around via an early-out in `evalConstExprAux`.

This way, moving to using an `skError` can also be implemented before later merging this PR, making the workaround that is currently employed obsolete.

zerbina:
I rebased the PR onto `devel` and adjusted it to the reporting-related changes plus to the `skError` usage in `semNormalizedConst`. The `localReport` that was previously used is now no longer necessary and thus removed.

bors r+

PR #511, opened 2023-01-07
add a simple tracing framework

zerbina:
## Summary
The goal is to provide the base for an event-based tracing framework
that can be incrementally improved/expanded upon. In its current form,
it's mainly meant for execution-time tracing.

While generally tailored towards being used as part of the compiler,
it's meant to be general enough to be used for both one-off temporary
instrumentation for use during debugging, and for building a persistent,
maybe even user-facing, compiler tracing facility. For example, it could
be expanded to also support the tracing done by `debugutils`.

This commit only adds the framework and a JSON serializer - both are not
used anywhere in the compiler yet.

## Details
The `tracer` module implements the data types and routines that make up
the framework.

The `trace_dump` module implements a serializer to the "Chrome Trace
Event" JSON Object format, which was chosen due to it being
simple-yet-powerful, with multiple visualizers already existing for it.

### DIFF
--- a/compiler/utils/trace_dump.nim
+++ b/compiler/utils/trace_dump.nim
@@ -1,0 +1,134 @@
+## This module implements a serializer for the trace data recorded with the
+## ``Tracer`` facility into the "Chrome Trace Event" JSON format, which is a
+## simple JSON-based format used for representing trace data. The format is
+## understood by Google Chrome's built-in "Trace Viewer" as well as other
+## trace/flamegraph visualizers, such as "Speedscope".
+##
+## For a specification, see `https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU`_
+##
+## "Complete" events (`X`) were used originally, as it meant less output to
+## write, but because of the issues it caused with visualizers in some cases,
+## "Begin" (`B`) and "End" (`E`) events are now used instead.
+##
+## While not strictly necessary, a manual JSON serializer is used instead of
+## ``std/json`` for efficiency.
+
+import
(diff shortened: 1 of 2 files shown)


compiler/utils/trace_dump.nim
```diff
@@ -0,0 +1,134 @@
...
+    let ts = inNanoseconds(e.timeStamp - start).int / 1000
+
+    stream.write "{"
+    intField(pid, 0); next()   # process ID
+    intField(tid, 0); next()   # thread ID
+    rawField(ph, ph); next()   # the event phase
+    floatField(ts, ts); next() # the timestamp
+
+    # while not it's not required by the specification for "End" events to store
+    # the "name" field, some visualizers won't work if the they don't
```
saem:
```suggestion
    # while it's not required by the specification for "End" events to store
    # the "name" field, some visualizers won't work without 
```

compiler/utils/trace_dump.nim
```diff
@@ -0,0 +1,134 @@
...
+    if not first:
+      stream.write ","
+    else:
+      first = false
+
+    let ph =
+      if e.begin: BeginPhase
+      else:       EndPhase
+
+    # make the timestamp a relative one and change the unit its measured in
```
saem:
```suggestion
    # make the timestamp a relative one and change the unit it's measured in
```

saem (approved):
Two doc items they're not blockers.

This is awesome, thank you for doing it so quickly. 🥰

zerbina:
I applied the suggestions (thanks @saem) and also improved the documentation of the `tracer` module a bit.

bors r+

PR #512, opened 2023-01-07
tables: fix multiple `OrderedTable` bugs

zerbina:
## Summary
The fixed issues:
- for `OrderedTable`s with a single item, the iterators would get stuck in an infinite loop in some cases
- items would sometimes not be yielded when using any of the iterators
- comparing two `OrderedTable` would sometimes ignore items

These only happened for `OrdererdTable` not explicitly initialized via `initOrderedTable` - explicitly initialized ones were not affected.

## The problem
The underlying issue was the same in each case. An `OrderedTable` uses an intrusive linked list internally. Each slot stores a `next` pointer and the table itself the head (`first`) and tail (`last`) pointer.

Since the pointers are indices into the internal data `seq` and not memory addresses, the default value (i.e. 0) does not represent "none" (because the 0-th `seq` item is a valid item), meaning that the pointers have to be explicitly initialized to the "none" state (i.e. `-1`). This happens when using `initOrderedTable`, and at one point did also happen when inserting into a default-initialized table, but the latter behaviour got removed as part of
#161 (likely by accident).

When inserting an item into a default initialized `OrderedTable` and the selected slot is slot `0`, the slot's `next` pointer was set to point to the slot itself, due to the `last` pointer incorrectly pointing to slot `0` at that point. If any of the iterators were used directly after this insertion, they would loop forever (because of the self-cycle).

When inserting an item into a default initialized `OrderedTable` and the selected slot is not `0`, the `next` pointer of the unfilled slot `0` (which is treated as the head of the linked list) is set to the index of the newly filled slot. As most of the `OrderedTable` implementation checks for and skips unfilled slots, this doesn't cause any immediate problems, but once the slot at position `0` becomes used (before the table is first enlarged), all previously inserted items are effectively cut off. They're still part of the table, but are no longer returned by any of the iterators nor considered during comparisons or when using `sort`.

## Details
- add an overload of the `initImpl` template that correctly sets the `first` and `last` pointer to `-1` as part when initializing the `OrderedTable` (either explicitly or lazily)
- remove the `isFilled` guards from logic that traverses the linked list - they're unnecessary
- remove the `rawGetKnownHC`, `rawGetDeep`, and `rawGet` overloads for `OrderedTable`. They were either unused or the same as the more general implementation

---

The issue caused Nim-CPS to sometimes get stuck in an infinite loop at compile-time.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/lib/pure/collections/tables.nim
+++ b/lib/pure/collections/tables.nim
@@ -1206,18 +1206,17 @@ type
 
 # ------------------------------ helpers ---------------------------------
 
-proc rawGetKnownHC[A, B](t: OrderedTable[A, B], key: A, hc: Hash): int =
-  rawGetKnownHCImpl()
-
-proc rawGetDeep[A, B](t: OrderedTable[A, B], key: A, hc: var Hash): int {.inline.} =
-  rawGetDeepImpl()
-
-proc rawGet[A, B](t: OrderedTable[A, B], key: A, hc: var Hash): int =
-  rawGetImpl()
+template initImpl(t: OrderedTable, size: int) =
+  ## Initializes `t` to a valid empty state. Called internally by
+  ## ``checkIfInitialized`` to lazily set up the table before it is used
+  newSeq(t.data, slotsNeeded(size))
+  t.counter = 0
+  t.first = -1
+  t.last = -1
 
 proc rawInsert[A, B](t: var OrderedTable[A, B],
                      data: var OrderedKeyValuePairSeq[A, B],
-                     key: A, val: sink B, hc: Hash, h: Hash) =
+                     key: A, val: sink B, hc: Hash, h: int) =
   rawInsertImpl()
   data[h].next = -1
   if t.first < 0: t.first = h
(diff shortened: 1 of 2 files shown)


lib/pure/collections/tables.nim
```diff
@@ -1206,18 +1206,17 @@ type
...
-  rawGetKnownHCImpl()
-
-proc rawGetDeep[A, B](t: OrderedTable[A, B], key: A, hc: var Hash): int {.inline.} =
-  rawGetDeepImpl()
-
-proc rawGet[A, B](t: OrderedTable[A, B], key: A, hc: var Hash): int =
-  rawGetImpl()
+template initImpl(t: OrderedTable, size: int) =
+  ## Initializes `t` to a valid empty state. Called internally by the
+  ## ``checkIfInitialized`` to set lazily set up the table before it is used
```
saem:
```suggestion
  ## Initializes `t` to a valid empty state. Called internally by
  ## ``checkIfInitialized`` to lazily set up the table before it is used
```

lib/pure/collections/tables.nim
```diff
@@ -1232,22 +1231,23 @@ proc enlarge[A, B](t: var OrderedTable[A, B]) =
...
-      var j: Hash = eh and maxHash(t)
-      while isFilled(t.data[j].hcode):
-        j = nextTry(j, maxHash(t))
-      rawInsert(t, t.data, move n[h].key, move n[h].val, n[h].hcode, j)
+    # every item part of the linked list is filled, so we don't need an
+    # ``isFilled`` guard here
+    var j: Hash = eh and maxHash(t)
+    # search for a free slot:
+    while isFilled(t.data[j].hcode):
+      j = nextTry(j, maxHash(t))
```
saem: dumb question, but it doesn't look like we're mutating `t` in the loop, so do we need to continually recalculate `maxHash(t)`?

(I'm super unfamiliar with this code, again, sorry)

tests/stdlib/types/collections/torderedtables_regression.nim
```diff
@@ -0,0 +1,57 @@
+discard """
+  description: "A regression test for lazily initialized `OrderedTable`s"
+  target: native
```
saem:
```suggestion
  target: native
  timeout: "10"
```

We're testing an infinite loop, I figure an explicit timout make sense? 10 might be too high.

saem (approved):
A few small things, nothing is a major blocker.

lib/pure/collections/tables.nim
zerbina: Thanks for reporting these. I tend to adjust documentation comments multiple times, and sometimes the remnants of previous iterations get past my proof-reading.

lib/pure/collections/tables.nim
zerbina: Your reading is correct, `maxHash` can be precomputed - the efficiency improvement is *likely* going to be minimal however, as `maxHash` is a template that expands to `t.data.high`. It is an improvement nonetheless, so I changed it.

tests/stdlib/types/collections/torderedtables_regression.nim
zerbina: Only a single iteration is expected, so an assertion is raised if a second one happens. In other words, the infinite loop is prevented and no intervention from testament's side is required.

saem: Got it, thanks.

saem:
Bors r+

I didn't take a close look, but just eliminating random failure.

PR #516, opened 2023-01-10
introduce vm syscall - echo

saem:
## Summary

Introduced a 'syscall' concept to the VM, starting with the "echo".

## Details

This is the first 'syscall' for the VM. Meaning the VM will yield control to the caller with the reason `yrkEcho`. The caller, the "system" in this case, is responsible for fullfilling the expectations of the echo, whatever that mean in the particular context, then resuming execution, if it chooses to. Baby CPS (Continuation Passing Style) for the win!

It's pretty simple right now, but it already drops more report related dependencies from the VM. A bit more of this and we'll have excised much of it in no time. :D

---

## Notes for Reviewers
* not looking to rework all the things, rather get a good start on them
* I'm not sure the precise shape of the "syscall" interface, so I didn't go nuts with documenting the concept
* if any quick wins stand out let me know, otherwise, I'll keep rolling with these and see what emerges.

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -25,7 +25,6 @@ import
   ],
   compiler/ast/[
     ast,
-    errorreporting,
     lineinfos,
     renderer, # toStrLit implementation
     trees,
@@ -110,6 +109,13 @@ type
     yrkMissingProcedure
       ## a procedure stub was called. The stub has to be resolved before
       ## continuing execution
+    yrkEcho
+      ## "syscall" of echo, the VM expects the code executing it to handle the
+      ## echo and then resume execution
+      # xxx: once there are more syscalls, consider having a few general
+      #      syscall variants and the data in embedded sub-objects. While on
+      #      the subject it's possible that quit could be considered a syscall,
+      #      one that doesn't expect resuming.
 
   YieldReason* = object
     ## The result of a single execution step (i.e. a call to ``execute``)
@@ -123,6 +129,8 @@ type
       exitCode*: int
     of yrkMissingProcedure:
       entry*: FunctionIndex   ## the entry of the procedure that is a stub
+    of yrkEcho:
+      strs*: seq[string]      ## strings to be echo'd, at least one item
 
 const
   traceCode = defined(nimVMDebugExecute)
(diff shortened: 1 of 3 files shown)


saem:
@zerbina I didn't get too ambitious, I needed a break from `sigmatch` woes... I'm building myself up to refactor `semProcAux` and friends, it's going to be a very large scope refactor, so I figure I should do this stuff for a while, get some momentum going, maybe it'll nibble away at some edges (🤞🏽).

The next thing I'm thinking of tackling is making the parser more independent, like `lexer` and `vm` modules. This would let me free the `vm` module of `SemReport` for `opcParseExprToAst` and `opcParseStmtToAst` opcode handling, as `parseCode` would no longer require `try`/`expect` shenanigans. Seems like a win-win-win.

Then keep proceeding until it's all cleared out, hopefully by that time I'll have gotten some more ideas about how to tackle `semProcAux`. 😅

compiler/vm/vm.nim
```diff
@@ -110,6 +109,13 @@ type
     yrkMissingProcedure
       ## a procedure stub was called. The stub has to be resolved before
       ## continuing execution
+    yrkEcho
+      ## "syscall" of echo, the VM expects the code executing it to handle the
+      ## echo and then resume execution
+      # xxx: once there are more syscalls, consider having a few general
+      #      syscall variants and the data in embedded sub-objects. While on
+      #      the subject it's possible that quit could be considered a syscall,
+      #      one that doesn't expect resuming.
```
zerbina, line 118: I think it'd be better to either remove the "the VM expects" part or to rephrase it in a way where it is clear that the expectations are not coming from the VM. For the VM itself, it doesn't matter whether or not the the syscall is handled nor if execution resumes thereafter.

compiler/vm/vm.nim
```diff
@@ -1941,29 +1949,26 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): YieldReason
...
+        raiseVmError(VmEvent(kind: vmEvtErrInternal,
+                             msg: "echo with $1 args" % $rb))
       else:
-        var outp = ""
-        for i in ra..ra+rb-1:
+        for i in ra..(ra+rb-1):
           checkHandle(regs[i])
-          #if regs[i].kind != rkNode: debug regs[i]
-          outp.add(regs[i].strVal)
+          result.strs.add $regs[i].strVal
```
zerbina, line 1967: A suggestion (can't use GitHub suggestions because the selection is not a single addition):
```nim
assert rb > 0

result = YieldReason(kind: yrkEcho)
result.strs.newSeq(rb)
for i in ra..<(ra+rb):
  checkHandle(regs[i])
  result.strs[i - ra] = $regs[i].strVal
```

compiler/vm/vmrunner.nim
```diff
@@ -379,16 +380,18 @@ proc main*(args: seq[string]): int =
 
   # the execution part. Set up a thread and run it until it either exits
   # normally or abnormally
-  var thread = initVmThread(c, entryPoint.start, frame)
-  block:
+  var
+    thread = initVmThread(c, entryPoint.start, frame)
+    continueExecution = true ## whether we continue to execute after yield
+  while continueExecution:
+    continueExecution = false # default to stop execution on any yield
```
zerbina, line 387:
```suggestion
    exit = false ## whether to exit after yield
  while not exit:
```
I personally think that this is a bit easier to understand and that it better signals continuing as the "default" thing to do.

As an alternative, a `while true` loop could be used and explicit `break`s used when exiting. That'd make the construct easier to understand for control-flow analysis.

I don't consider this to be a blocker - no change is required.

zerbina:
> Baby CPS (Continuation Passing Style) for the win!

Yep, very much! Turning `rawExecute` into what could be viewed as a `.closure` iterator was the goal when introducing `YieldReason` and its associated bits.

For a generalized syscall interface (not something that this PR has to implement, I'm just noting it down), my idea was to use the following:
```nim
...
of yrkSyscall:
  id: SyscallId             ## identifies the syscall (e.g. "echo")
  result: Option[TRegister] ## the register that is expected to store the value 
                            ## resulting from the call (if it returns something)
  args: Slice[TRegister]    ## the arguments 
```
Almost all argument/result value processing is moved outside of the VM this way, and the system/host has a large amount of freedom and flexibility in how it implements them.

Depending on whether or not which procedure requires a "syscall" should be a code-generation-time or (VM) run-time decision (i.e. redirecting procedure calls to be "syscalls" without requiring on-the-fly bytecode modifications), a new instruction (e.g. `opcSyscall`) could be introduced, a mechanism similar to the one used for callbacks (i.e. hooks) be used, or both.

zerbina (approved):
Looks good to me - the improvement suggestions and thoughts I left are  not blocking.

saem:
Thank you kindly for the review, @zerbina. I'll try to incorporate these hopefully before the weekend.🤞

compiler/vm/vmrunner.nim
saem: I reworked it where it by default `break`s at the end of an iteration, then use a continue to override it within a branch. Seems a little bit better. All this to avoid infinite loop errors, I might be being silly.

compiler/vm/vm.nim
saem: I'm not sure about the assert, I felt like the event made sense to signal invalid code issue and then have the caller figure that out. I'm not married to this approach. The biggest thing is that I reordered the branches so it's a little more straightfoward to read.

compiler/vm/vm.nim
saem: Good call, done.

saem:
ugh, forgot to push my changes, oh well, they'll go in the next PR.

PR #518, opened 2023-01-17
semstmts: produce an `skError` for erroneous constants

zerbina:
## Summary
If the type or initializer expression has an error, the produced symbol is
now an `skError`. This prevents the `nkError` node that represents the
content of the constant from being inlined, which would cause the compiler
to crash, as `nkError` nodes are not expected when inlining constants.

## Details
- fix a compiler crash when using tuple unpacking for a `const` and either
  the type or the init expression is erroneous
- remove the check for `nkEmpty` init expressions. They're already handled
  earlier when evaluating the expression

## Misc
- fix a `FieldDefect` crash when converting
  `adSemOnlyDeclaredIdentifierFoundIsError` to a legacy report

### DIFF
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -806,7 +806,7 @@ func astDiagToLegacyReport*(diag: PAstDiag): Report {.inline.} =
         location: std_options.some diag.location,
         reportInst: diag.instLoc.toReportLineInfo,
         kind: rsemOnlyDeclaredIdentifierFoundIsError,
-        str: diag.name,
+        str: diag.identName,
         ast: diag.wrongNode,
         wrongNode: diag.err)
   of adSemCannotImportItself:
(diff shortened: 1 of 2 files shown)


compiler/front/msgs.nim
```diff
@@ -806,7 +806,7 @@ func astDiagToLegacyReport*(diag: PAstDiag): Report {.inline.} =
         location: std_options.some diag.location,
         reportInst: diag.instLoc.toReportLineInfo,
         kind: rsemOnlyDeclaredIdentifierFoundIsError,
-        str: diag.name,
+        str: diag.identName,
```
saem, line 809: That's better, thanks.

compiler/sem/semstmts.nim
```diff
@@ -1177,18 +1177,27 @@ proc semNormalizedConst(c: PContext, n: PNode): PNode =
...
+          of nkError:       def
+          else:             unreachable()
       of nkConstDef:
         # xxx: rework skConst to store the whole def like let/var
         v.ast = producedDecl[^1]
       else:
         internalError(c.config, "should never happen")
 
+    if v.ast.isError:
+      # XXX: although this mirrors the behaviour of ``semNormalizedLetOrVar``,
```
saem, line 1196: Admittedly, the handling of type and symbols is really sloppy in the compiler in general and I sadly repeated this pattern.

saem (approved):
Bors r+

saem:
Bors r+

Do the merge dance, please.

PR #520, opened 2023-01-21
testament: megatest works for all targets

saem:
## Summary
Megatest now benefits all targets, this should reduce the chance of CI timeout

## Details

Megatest used to only work for the c target. Now it works across all of them. Each per target megatest will execute one after the other, but will be in parallel with the other tests.

In the future per target megatest execution can be parallelized, but testament requires more clean-up prior to that.

---

## Notes for Reviewers
* key point is that the tests execute faster
* the code isn't great, but recusing timmy code is always like that
* if you spot a new logic error lmk


saem:
Bors r+

PR #522, opened 2023-02-02
ci: don't use deprecated `set-output` command

zerbina:
## Summary
The workflow command `set-output` was deprecated by GitHub Actions some time ago and is planned to be disabled in the future. The now recommend way is to write the workflow output as key-value pairs to the output environment file (`GITHUB_OUTPUT`).


alaviss (approved):
LGTM

bors r+

saem:
Bors r+

The test failure is ridiculous, that test needs to be reworked, we wouldn't be off pinging Google and hoping for the best.

alaviss:
Third time is the charm

bors r+

PR #523, opened 2023-02-03
perform `.closure` inference in `sempass2`

zerbina:
## Summary
While the `lambdalifting` transformation is responsible for figuring
out what inner procedures capture the content of which memory
locations, semantic analysis already needs to know whether an inner
procedure accesses its environment, as whether it does affects its
calling convention (which is part of a procedure's type).

`semCaptureSym` was responsible for detecting if usage of a symbol
required a capture and for changing all relevant enclosing procedures
between the to use the `.closure` calling convention, but the
implementation had multiple issues.

Instead, whether a procedure captures something is now detected during
`sempass2`. Setting the calling convention is performed there too now.
This fixes:
- inner procedures using the `.closure` calling convention even if
  explicitly specified otherwise
- tentative semantic analysis (such as `compiles` or analysis of
  argument expression during overload resolution) having side-effects
  because of `semCaptureSym`
- symbols resulting from typed template/macro parameter expansion
  weren't considered by `semCaptureSym`, which in some cases caused
  procedures to not be updated to use the `.closure` calling convention,
  which then caused errors in the code-generator or backend compiler

Detecting illegal captures now also happens in `sempass2` (instead of
in `lambdalifting`), which means that they're now *always* reported and
not only for alive code.

A hint is now reported if a procedure uses the `.closure` calling
convention but doesn't capture anything.

## Details

- move illegal capture detection from `lambdalifting` into `sempass2`
- detect whether a procedure captures something during `sempass2`
  instead of in `lambdalifting`
- remove `semCaptureSym` and its usages
- remove the `tfCapturesEnv` type flag. It's use case was detecting
  top-level closure procedures (can only be achieved via anonymous
  procedures), but a hidden environment parameter is added for them
  now, instead
- report a dedicated error when attempting to capture a run-time value
  in a compile-time context

Changes to `lambdalifting`:
- remove the unused "in container" analysis
- remove all logic for permanently changing a procedure's calling
  convention
- run the pass on all non-`.closure` procedures (including top-level
  closure iterators) instead of only on top-level procedures. While
  this doesn't affect the result (all inner procedures that capture
  something are still lifted), it means less recursion depth and a
  smaller working set for each run, as well as less calls to
  `transformBody` (which translates to less work).

### Misc

- change `rsemIllegalMemoryCapture` to use the `sym` field instead of
  `symbols`
- move `isOwnedBy` from `vmgen.nim` to `ast_query.nim`

---

## Todo
- [x] add a test for the `rsemClosureNoEnv` warning
- [x] add tests for the various cases that now work correctly

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


saem:
I think I'm pro-converting it to an error, which can be in a separate PR. It might be a bit fun if someone is doing something terrible with "generics", etc... where in some pathways its a closure and in others is not. Honestly I want to make that type of thing a pain, anyhow. It's just silly how far genetics are from universal types.

compiler/ast/ast_types.nim
```diff
@@ -530,7 +530,6 @@ type
                       ## concrete type (lastSon becomes the concrete type)
     tfRetType,        ## marks return types in proc (used to detect type classes
                       ## used as return types for return type inference)
-    tfCapturesEnv,    ## whether proc really captures some environment
```
saem, line 533: Yasss! Less of these silly flags.

saem:
Took a quick pass, looks good but I really should be asleep and I don't trust my code review this late at night. I'll look at it tomorrow, but it's already so much simpler.

(I also fast fingered the close button it seems, sorry about that)

compiler/ast/ast_query.nim
```diff
@@ -548,6 +548,13 @@ proc skipGenericOwner*(s: PSym): PSym =
            else:
              s.owner
 
+func isOwnedBy*(a, b: PSym): bool =
+  ## Tests if `b` is the transitive owner of `a`
```
```suggestion
  ## Tests if `b` is the transitive owner of `a`, returns true if `a` got owned! :)
```

compiler/sem/sempass2.nim
```diff
@@ -407,6 +421,18 @@ proc useVar(a: PEffects, n: PNode) =
...
       a.init.add s.id
+
+  if s.kind in skLocalVars and
+     sfGlobal notin s.flags and       # must not be a global
+     a.owner.kind in routineKinds and # only perform the check inside procedures
+     s.owner != a.owner and
+     illegalCapture(s):
+    # XXX: it'd be better if this error would be detected during the first
+    #      semantic pass already, but detecting an access to outer entity is
+    #      not easily doable because of typed template/macro parameters
```
saem: Thought for the future:

have explicit block and block argument 'passing', this would mean we can also generalize hoisting, lifting, and capturing. That should set us up to much more easily handle this.

doc/manual.rst
```diff
@@ -2108,6 +2108,11 @@ proc inside of a proc). For an inner proc an analysis is performed whether it
 accesses its environment. If it does so, it has the calling convention
 `closure`, otherwise it has the calling convention `nimcall`.
 
+In the case that a non-`iterator` procedure is explicitly specified to use the
+`closure` calling convention, but the aformentioned analysis shows that no
+environment is accessed, the calling convention is changed to `nimcall`. The
+compiler is allowed to either report a warning or error in this case.
```
saem: Due to generics and conditional compilation I'm not sure how great this interaction might be... I think we should start with this, but something to ponder occasionally and perhaps inspiration will strike. 😆 

I say leave this in for now, we can easily change our mind.

saem (approved):
Much simpler! 🥂 

Less state and fewer scattered checks.

zerbina:
I added some tests for the fixes, cleaned up related imports, applied the doc comment suggestion, and also fixed a small issue with `.compileTime` procedure detection.

bors r+

zerbina:
While discussing the language change with me, @Clyybber brought forward the following problem with the compiler changing `closure` to `nimcall`:
```nim
proc outer() =
  proc inner() {.closure.} =
    static: doAssert inner is "closure" # <- works

  static: doAssert inner is "closure" # <- fails
```

Because there's also uncertainty about how well the `.closure` -> `.nimcall` overriding interacts with generics, it was decided to not introduce the language change for now.

While I originally thought that the reintroduction of the `tfCapturesEnv` would be necessary in order for the changes to `lambdalifting` to be kept, this turned out to not be the case --  no changes were required (I did reword some comments, however).

To counter the slightly increased amount of work that `sempass2.detectCapture` now has to do in the presence of non-capturing `closure` procedures, procedures that were analysed already are remembered and not analysed again, which should make the logic faster in general.

Nonetheless, there are some regressions compared to the original change:
```nim
proc outer() =
  proc inner() {.closure.} =
    discard

  inner() # unnecessarily heap-allocates an empty environment again

  proc inner2() = # inferred again to use require the closure calling convention
    inner()

  inner2() # also requires a heap allocation
```
But I think these are acceptable for now. Before looking into fixing these issues again, the lambda-lifting should be refactored first.

zerbina:
That should fix the test failures. Top-level anonymous procedures using the `closure` calling convention now also get a hidden environment parameter added during `transf`.

Once the CI passes, this PR can be merged.

PR #524, opened 2023-02-05
stdlib: split sexp module into sexp and sexp_parse

saem:
## Summary

- `sexp` contains the high level interface with `SexpNode`
- `sexp_parse` provides the low level token straem interface

## Details

The low level event/stream based interface wasn't really useable by other modules. This separation allows for that. This should also ensure that we have small narrow public interfaces for modules.

A more accessible token/event stream based interface should allow for building fast sexp based IR serdes for simple test description and print debugging.

---

## Notes for Reviewers
* suggestions that help narrow down the public API surface

### DIFF
--- a/lib/experimental/sexp.nim
+++ b/lib/experimental/sexp.nim
@@ -7,7 +7,24 @@
 #    distribution, for details about the copyright.
 #
 
-## **Note:** Import ``nimsuggest/sexp`` to use this module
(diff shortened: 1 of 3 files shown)


saem:
Ugh, some of these CI failures really don't provide a lot of good signal.

saem:
Bors r+

saem:
OK, I'll look at optimizing some bit of CI tonight, this is ridiculous.

PR #526, opened 2023-02-06
transf: move the `ImplicitObjConv` hint to `sempass2`

zerbina:
## Summary
The hint is about program semantics and should thus be reported during semantic analysis. This is a step towards removing all program-semantics-related diagnostics reporting from both the transformation and back-end phase.

As a side-effect of this change, the hint is now produced for all code part of the compilation, instead of only for alive code.

### DIFF
--- a/compiler/sem/transf.nim
+++ b/compiler/sem/transf.nim
@@ -48,11 +48,6 @@ import
     idioms
   ]
 
-# xxx: reports are a code smell meaning data types are misplaced
-from compiler/ast/reports_sem import SemReport,
-  reportAst
-from compiler/ast/report_enums import ReportKind
-
 proc transformBody*(g: ModuleGraph; idgen: IdGenerator, prc: PSym, cache: bool): PNode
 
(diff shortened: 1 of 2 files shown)


compiler/sem/transf.nim
```diff
@@ -48,11 +48,6 @@ import
     idioms
   ]
 
-# xxx: reports are a code smell meaning data types are misplaced
```
saem, line 51: Every time we have less of this the modules become more amenable to change. ♥️

compiler/sem/transf.nim
```diff
@@ -549,20 +544,11 @@ proc transformConv(c: PTransf, n: PNode): PNode =
     else:
       result = transformSons(c, n)
   of tyObject:
-    template convHint() =
```
saem, line 552: Thanks for dropping the needles abstraction.

saem (approved):
Bors r+

Thanks, zerbina. Now, let's just hope that bors doesn't let us down.

PR #527, opened 2023-02-07
disable flakey http test

saem:
It keeps breaking CI as it's attempting to use services like google.
This is of course a bad idea in the first place, marked it as a known
issue very firmly. Just so people are aware how ridiculous the whole
thing is.

---

## Notes for Reviewers
* do not be afraid to remove junk like this


saem:
Bors r+

saem:
Bors r+

PR #529, opened 2023-02-08
compiler/installer.ini: reduce csources list to common platforms

alaviss:
## Summary
Currently the large list take a huge toll on CI time. Trim this down to common platforms to speed CI up.

For platforms removed from this list, bootstrapping from a source archive would require either cross-compilation or via an older archive with the platform included to generate the bootstrap compiler.


saem:
bors r+

Thanks, @alaviss!

saem (approved):

PR #530, opened 2023-02-08
publisher: update output format

alaviss:
## Summary
With the switch to `GITHUB_OUTPUT`, it seems that they no longer uses the old escaping style but rather a new bash-like method documented here:

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

However, we were outputting the old escaped format, making the output unusable by later jobs.

This commit fixes the output format to restore publishing functionality to publisher.

Verified working: https://github.com/alaviss/nimskull/actions/runs/4121576271/jobs/7117396986

### DIFF
--- a/.github/workflows/publisher.yml
+++ b/.github/workflows/publisher.yml
@@ -75,8 +75,13 @@ jobs:
           # Create a new release manifest
           ./release_manifest add *.json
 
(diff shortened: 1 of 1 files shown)


alaviss (commented):
bors r+

saem (approved):
Bors r+

Go Go Gadget Releases!

PR #532, opened 2023-02-10
 Saem exp sigmatch ast corruption

saem:
## Summary

Previously `nkError` nodes were allowed to pass through to parameters, these now result in a match failure.

## Details

Besides the change stated above, `semProcBody` now is `nkError` aware. Removed creating callsite node copies during sigmatch. Finally, error message for a space between a routine call name and parameters now makes sense and fixed the associated test.

Although slow, CPS should now work.
---

## Notes for Reviewers
* feedback I'm mos interested in: glaring issues and quick wins
* if I'm not around please merge

### DIFF
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -2504,6 +2504,7 @@ proc genClosureConstr(c: var TCtx, n: PNode, dest: var TDest) =
 proc gen(c: var TCtx; n: PNode; dest: var TDest; flags: TGenFlags = {}) =
   when defined(nimCompilerStacktraceHints):
     setFrameMsg c.config$n.info & " " & $n.kind & " " & $flags
+
   case n.kind
   of nkSym:
     let s = n.sym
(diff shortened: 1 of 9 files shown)


compiler/vm/vmgen.nim
```diff
@@ -2683,6 +2683,8 @@ proc gen(c: var TCtx; n: PNode; dest: var TDest; flags: TGenFlags = {}) =
       genCastIntFloat(c, n, dest)
   of nkType:
     genTypeLit(c, n.typ, dest)
+  of nkError:
+    assert false
```
zerbina:
```suggestion
```
An `nkError` should not be able to reach here. If one does, then there's a bug with `mirgen`.

tests/errmsgs/tmake_tuple_visible.nim
```diff
@@ -1,6 +1,8 @@
 discard """
-  errormsg: '''Mixing types and values in tuples is not allowed.'''
-  line: 19
+  errormsg: '''type mismatch: got <(typedesc[NimEdAppWindow], int)>'''
+  line: 20
```
zerbina:
```suggestion
  line: 21
```

zerbina (approved):
Looks good to me. I left a small suggestion for `vmgen` and a fix for the test failure.

compiler/vm/vmgen.nim
saem: I should make that an internal error really.

In the future i should have a const that's, `nodeKindsAllowedForVMGen` or something, and then guard it all and all case statements are exhaustive.

tests/errmsgs/tmake_tuple_visible.nim
saem: Thanks

saem:
Bors r+

PR #533, opened 2023-02-11
Parser: remove report dependency

saem:
Summary
=======

This removes direct dependency on `reports`, but an indirect one still
exists via `msgs`. It's pretty trivial indirection at the moment, but
after dropping direct reports dependencies the API can be changed more
drastically.

A number of changes were required in order to make this possible, here
is an overview:

- smaller parser public API & simpler implementation
- added `parse` compiler command, for devs
- parser error message improvements
- fixes to `astrepr` logic and output
- lots of style clean-ups

Details
=======

Slimmer `parser` Public API
---------------------------

Previously the parser had many public procedures (eg: `isOperator`,
`getTok`, `skipComment`, etc) that would allow fine grained control for
other modules.

There are many issues with this:
- there are no consumers of this API
- lots of public API surface to test
- the API itself was bad, it conflated lexing and parsing

The public API surface for the parser has been reduced significantly,
now consisting of:
- `openParser`
- `parseTopLevelStmt`
- `parseAll`
- `closeParser`
- `parseString`

That's it, which frankly reads far more sensibly.

Simplified `parser` Implementation
----------------------------------

- removed `InternalReport`, `reports_parser`, and `reports_enum` imports
- introduced diagnostics for the parser, akin to the lexer, `ParseDiag`
- `ParseDiag` favours data over strings
- `ParsedNode` now has its own kind enum, mostly a subset of
   `TNodeKind`, but entirely compatible

Consolidated a pattern within the parser, where a node was created with
the current token's information, and then the token was immediately
consumed via `getTok` to advance the `lexer`. This is captured in the
newly introduced `newNodeConsumingTok`.

Long-term, itemizing these traversal/consumption patterns will make the
parser logic not only more regular, but also highlight oddities in the
grammar as the implementation will be convoluted.

Parsing/Diagnostics Performance
-------------------------------

`ParsedNode` uses a lightweight `ParsedToken`

Introduce `ParsedToken`, a smaller data type, storing the least amount
of data required from `lexer.Token` for `ParsedNode`. This not only
saves memory, but the runtime performance impact on my machine is
roughly 33% faster full compiler testament run for all targets

- before change: 3+ minutes
- after change:  2+ minutes

Added specialized diagnostic/report kinds for:
- empty accent quote when ident expected
- msg for asm statements without a string literal
These reduces the amount of string data carried around in the compiler.

Improved Custom Numeric Literal Handling
----------------------------------------

- the `lexer` still does silly things for lexing these
- it just does less work and produces better data
- fewer string operations and hacks are required by the `parser`

Parser Diagnostic/Reporting - Invalid Indentation
-------------------------------------------------

- now has correct source line information
- tracks instantiation and submission location
- has the appropriate severity
- improved phrasing for indent error from possible missed `=` character
- adjusted tests for the above

Parser Diagnostic/Reporting - Malformed Call Syntax
---------------------------------------------------

- `parser` detects malformed calls and sets better line info
- net-net the user will have a better chance to find the issues

Parser Diagnostic/Reporting - Misc
----------------------------------

- token rendering call out keywords via prefix, eg: `keyword template`
- inconsistent spacing style check shows the problematic source

Removed unused report kinds:
- `rparIdentOrKwdExpected`
- `rparRotineExpected`
- `rparPragmaAlreadyPresent`

Parse Compiler Command
----------------------

`parse` command:
- added `parse` command, which outputs the parsed ast for a file
- usage: `nim parse foo.nim`
- super useful for diffing parser output changes
- heavily leverages `astrepr`

`astrepr` module:
- `astrepr.treeRepr` now works for `ParsedNode`, was previously broken
- AST trasversal is now exhaustive and breakages less likely to pass CI

`astrepr` output improvements, mainly for `ParsedNode`:
- `astrepr` now shortens ParsedNodeKind enum
- output now includes line and column information
- comments no longer result in excessive new line output
- fixed many formatting issues for `ParsedNode` output
- improved `astrepr`'s output for custom numeric literals

Canonical Filenames Performance Issue
-------------------------------------

Also discovered a performance issue with canonical filenames option and
the `nimdebugstacktrace` option. Removed some of the pain, but canonical
file paths result in significant performance issues due to filesystem
IO. I've fixed part of it and filed an issue:
https://github.com/nim-works/nimskull/issues/546

Other Improvements
------------------

- introduced `debugutils.setFrame` template for frame msg hints
- above `setFrame` avoids the canonical path performance hit
- removed circular dependency between `ast` and `options` module
- document unused parser reports and other outliers
- move `isImportedException` to `ast/types`, whice drops `front/options`
  cyclic depencdency from `ast/ast_query`
- fixed docs in nimlexbase, also easier to understand
- `ast.toPNode` now handles `nil` input
- `syntaxes.parseFile` returns `ParsedNode`, allows avoiding unnecessary
  conversions in future use cases where only `ParsedNode` is required

Special Mentions
----------------

Thanks, clyybber and zerbina for the reviews!

Misc
----
- remove blank space characters from otherwise empty lines
- remove awful code style of `0 < foo.len`
- fixed a number of typos in comments
- adjusted a few tests to ensure they pass

---

## Notes for Reviewers
* done and good to go IMO
* Should be complementary to lexer DOD work

## TODO (non-exhausitive)
- [x] remove direct dependency on `reports` bits
- [x] `parser` sort out data internals approach

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -20,11 +20,11 @@ import
     numericbase       # NumericalBase
   ],
   compiler/front/[
-    options
+    in_options
   ],
   compiler/utils/[
     ropes,
-    astrepr,
+    idioms,
     int128 # Values for integer nodes
   ],
   std/[
@@ -32,7 +32,7 @@ import
     tables # For symbol table mapping
   ]
 
-export ast_types, ast_idgen, ast_query, int128, ast_parsed_types
+export ast_types, ast_idgen, ast_query, int128
 
 var ggDebug* {.deprecated.}: bool ## convenience switch for trying out things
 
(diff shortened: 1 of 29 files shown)


compiler/ast/parser.nim
```diff
@@ -821,8 +838,10 @@ proc parseOperators(p: var Parser, headNode: ParsedNode,
   while opPrec >= limit and p.tok.indent < 0 and not isUnary(p.tok):
     checkBinary(p)
     let leftAssoc = ord(not isRightAssociative(p.tok))
-    var a = p.newNode(nkInfix)
-    var opNode = p.newNode(nkIdent, p.tok) # skip operator:
+    var
+      a = p.newNode(pnkInfix, p.tok) # TODO: confirm before merge, is this the
+                                     #       right token to set?
```
Clyybber: yep

compiler/ast/parser.nim
```diff
@@ -1158,8 +1181,8 @@ proc primary(p: var Parser, mode: PrimaryMode): ParsedNode =
     # Note 'sigil like' operators are currently not reflected in the grammar
     # and should be removed for Nim 2.0, I don't think anybody uses them.
     let isSigil = isSigilLike(p.tok)
-    result = p.newNode(nkPrefix)
-    var a = p.newNode(nkIdent, p.tok)
+    result = p.newNode(pnkPrefix, p.tok) # TODO: before merge confirm `p.tok` is correct
```
Clyybber: yep

compiler/ast/parser.nim
saem: Thanks!

compiler/ast/parser.nim
saem: thanks, noted it locally, will push changes shortly.

compiler/ast/parser.nim
```diff
@@ -1264,22 +1287,27 @@ proc postExprBlocks(p: var Parser, x: ParsedNode): ParsedNode =
...
+                     stmtList.token,
                      body = stmtList,
-                     params = if openingParams.kind != nkEmpty: openingParams
-                              else: p.newNode(nkFormalParams, @[p.emptyNode]),
+                     params =
+                      if openingParams.kind != pnkEmpty:
+                        openingParams
+                      else:
+                        # TODO: before merge confirm p.tok is correct
+                        p.newNode(pnkFormalParams, p.tok, @[p.emptyNode]),
```
Clyybber: Since there are no params in this case there isn't much of a right answer here I think. For the `pnkFormalParams` that `parseParamList` discards, it would be `p.tok` before the call to `parseParamList`.

compiler/ast/parser.nim
```diff
@@ -2004,14 +2036,14 @@ proc parseTypeDef(p: var Parser): ParsedNode =
...
+    if pragmas.kind == pnkEmpty:
       identifier
     else:
-      p.newNode(nkPragmaExpr, @[identifier, pragmas])
+      p.newNode(pnkPragmaExpr, p.tok, @[identifier, pragmas])
   result.add genericParams
   result.add case p.tok.tokType
              of tkEquals:
-               result.info = p.lineInfo
+               result.token = p.tok # TODO: before merge confirm this makes sense
```
Clyybber: It does, in the sense that the `=` of the type declaration will be considered the root of it, similar to a binary operator.

compiler/ast/parser.nim
saem: Yeah, I think there is a convention in place and it needs to become a formalization, presently things are a bit fuzzy when it comes to info/token to fixate on:

- given `foo = bar`, the `pnkAsgn` is at the equals sign, by convention.
- given `foo = 10`, in an enum type definition context, the `pnkEnumFieldDef` is at `foo` or should it be at the assignment, or the tuple?
- given `foo bar`, a `pnkCommand`, in the old parser it's after the `o` in `foo`
- similar to the above, `foo(bar)` a `pnkCall`, uses the `(` position, as that's the "invocation"

Not sure which token/location to fixate on here for the parsed node production, it might be more than one rule but we need something.

compiler/ast/parser.nim
saem: Fair enough, I noted some of the conventions in the reply to the prior comment, I think the gap is a sensible specification. 😆

compiler/ast/parser.nim
Clyybber: > given foo = 10, in an enum type definition context, the pnkEnumFieldDef is at foo or should it be at the assignment, or the tuple?

At the assignment would be consistent with how we do it elsewhere and also not redundant with the RHS or LHS location info (or token in this case).

Clyybber: In this specific case where there are no formal params but we still generate a formal params node, we can use the token "closest" to where the formal params would be if there were any, so I think `p.tok` before the call to `parseParamList` is the way to go.

compiler/ast/parser.nim
```diff
@@ -1264,22 +1286,27 @@ proc postExprBlocks(p: var Parser, x: ParsedNode): ParsedNode =
     p.getTok
     p.skipComment(result)
     if p.tok.tokType notin {tkOf, tkElif, tkElse, tkExcept}:
```
Clyybber: Pre-existing, but I think `tkFinally` might be missing here?

saem: pretty sure it is, I'll add it in later or separately.

compiler/ast/parser.nim
saem: `pnkDo` is now sorted, at least.

I'll take a quick peek at the enum field definition.

compiler/ast/parser.nim
saem: pretty sure it is, I'll add it in later or separately.

compiler/ast/parser.nim
saem: enum is done, too.

Now to figure out this command funny business.

compiler/ast/parser.nim
saem: Removed the comment as confirmation.

compiler/ast/parser.nim
saem: OK, line info for `pnkCommand`s are now fixed.

Aside: it's pre-existing, but right now all `pnkCommand` nodes indicated their info based on the position of the first param, where as the call is between the end of call sym and start of parens. Probably need to think about what makes the most sense.

You can see this happening in `tvarargslen.nim`, all the line infos have a column of `9`, except the `pnkCall` which is `8`.

compiler/ast/parser.nim
saem: The `finally` is a must, a test will confirm it, but without that present it'll never do that same indent loop for finally at the bottom.

----

The "real" question is do we enforce only a single `finally` and it being last? I'm leaning strongly towards yes. Even though it's not for exception handling, we should ensure there is a certain amount of regularity to the code or we'll end up with people not trusting the syntax they're reading, also a matter of identity (that looks like nimskull code). Scala 2 had a lot of issues with this.

Don't have to finalize it now, but something to mull over. Conceptually, I consider this similar to operator precedence. Lastly, it's easier to alleviate a restriction later than add one.

compiler/modules/modulegraphs.nim
```diff
@@ -617,3 +617,17 @@ proc onProcessing*(graph: ModuleGraph, fileIdx: FileIndex, moduleStatus: string,
         importStackLen: graph.importStack.len,
         moduleStatus: moduleStatus,
         fileIdx: fileIdx))
+
+proc isImportedException*(t: PType; conf: ConfigRef): bool =
+  # TODO: find a better home for this, it used to be `ast_query` but that
+  #       module must not depend upon `options` (due to `ConfigRef`). The
+  #       consumers of this proc are cgen, jsgen, semstmts, and tranf at the
+  #       time or writing.
```
zerbina: I think `compiler/ast/types.nim` would be a good fit.

compiler/ast/parser.nim
```diff
@@ -220,46 +236,74 @@ proc lineInfo(p: Parser): TLineInfo =
...
-  return newParsedNode(kind, info, subnodes)
+  newParsedNode(kind, p.lex.fileIdx, token, subnodes)
+
+proc currentTokToNode(
+    p: var Parser,
+    kind: ParsedKindWithSons,
+    subnodes: sink seq[ParsedNode] = @[]): ParsedNode =
+  ## Create a new non-leaf parsed node from the current token and advance the
+  ## parser
+  # TODO: come up with a better name before merging
```
zerbina: Maybe `consumeTok`?

saem:
## Preamble

I'm pretty sure I need to come up with a better data format for the parser. This branch shoves at least one token in nearly all `ParsedNode` and I don't believe that's a good idea to merge. So here is a significant rework of the `parser` module to address that.

So first off, a review of how it's/can be used today.

## Core Parser API

There are three ways to use the parser:

1. Parse a string
2. Parse the next top level statement in a file
3. Parse all or rest of the file

When the parser encounters an error currently it passes control to the reports stupidity, which not only logs the message, but also decides whether to continue, raise, or abort.

## Key Parser Use Case

Of course it's used to turn text to some parsed structured, so that's easy. The last part in the previous section about passing control is most relevant. A number of places in the compiler (ie. VM and suggest) use these and a `structuredReportHook` to pass back `doNothing` to avoid raising/aborting. This has the parser keep making progress regardless of the issues it encounters. This control can also be used to force early termination and avoid unnecessary work -- the biggest impact here is if we're compiling a bunch like in testing where speed, file IO, and memory consumption matter.

Recap:
1. parsing and feeding semantic analysis or source filtering
2. parsing and not having errors abort the whole process
3. parsing while doing little/to no extra work (first for speed, and second for resource usage)

## Architecture Thinking

Given the above interface and use case(s), I think giving up the ability to  making a parser that's very fast and light will mean we can probably easy up on the fine grained control desired by 3. At which point 1 and 2 are easily satisfied.

The current parser implementaiton, which does the usual `reports` silliness in various leaf procs, just isn't written in a way where returning an error node will work. Not to mention, we'll still be stuck with a sea of slow refs. Also this branch remembers all the tokens currently, which is also heavy/slow.

So whatever the approach, it needs to be:
1. light on the memory
2. errors need to be easily encoded but allow progress
3. the parser has the errors readily available for downstream

## Design Thinking

The first point means moving away from a sea of refs and likely some sort of linear structure. So the return type of the various parse procs should be `ParseResult` or something instead of `ParsedNode`. `ParseResult` can be a chunky value as long as it's lighter in aggregate than the `ParsedNode` and saves on pointer chasing.

The second point effectively rules out exceptions and anything else that stops progress. So whatever it is, we need to be able to keep adding to the `ParseResult` and then let downstream sort things out. For sanity's sake that should be relatively easy, perhaps `ParsedResult.areYouOkay` or the like.

Third, an iterator or something that allows for diagnostics (errors, warnings, or hints) will make it easy for downstream consumers to log, abort, whatever if there are issues. The previous approach of passing control back and forth meant that correlation of error and current state of the world was easy to do, so a bit of extra data might be required to support the necessary queries (don't have an exhausitive answer for the nature of these queries).

## Design

This is a quick sketch of the data/layout,hopefully it's sufficiently illustrative.

```nim
type
  TokenData = object
    tokens: ParseTokens # light weight tokens, rather than change the lexer wholesale
    strLiterals: seq[string]
    intLiterals: seq[tuple[num: BiggestInt, base: NumericalBase]]
    fltLiterals: seq[tuple[num: BiggestFloat, base: NumericalBase]]
    spacing: seq[tuple[leading: int8, trailing: int8]] # might not need this
    indent: seq[int16] # might not need this

  ParseResult* = object
    fileIndex: FileIndex
    tokens: TokenData
    parseTree: seq[TreeNode]
    extraData: seq[int32]
    diags: ParseDiags        # lexer diags will need mapping into `ParseDiag`

  TreeNode = object
    tokenIdx: int32 # index into `ParseResult.tokens`
    case kind: ParseNodeDatumKind: # appropriate enum, elided
      else:
        discard # left/right int32 indexing tricks a la Zig

  ParseTokens = seq[ParseToken]

  ParseToken = object
    line: uint16
    col: int16
    kind: TokType # hopefully this is reuseable

  ParseDiags = seq[ParseDiag]

  ParsedNode* = object # value object for iteration/traversal
    id*: int32
    kind*: ParsedNodeKind
    parent: int32 # -1 for no parent
```

At this point the key API `parseString`, `parseAll`, `parseTopLevelStmt` will all return a `ParseResult`, although in the case of `parseTopLevelStmt` it should return a `ref` object instead of the big chunk of data each time. To use a `ParseResult`, the API could be:

```nim
func hasErrors(p: ParseResult): bool
iterator diags(p: ParseResult): ParseDiag
iterator items(p: ParseResult): ParsedNode

# to make `ParsedNode` useable:
template makeParsedNodeUseable(r: ParseResult, n: ParsedNode) =
  template lineInfo(n: ParsedNode): TLineInfo = lineInfo(r, n)
  # remaining templates ...
  # basically, bind by name all the accessors, which require a `ParseResult` for data fetch
  # call this in the iterator body and it safely scopes to the lifetime of the result
```

## Desired Outcome

With the above design have a lightweight parser that can then do extra work without us feeling the performance cost. At the same time not require too much change in various parts of the compiler using the `parser` module, some bits at the call sites are one thing, but impacting their overall internals is another.

## Feedback

I'm looking for feedback, for understanding you might want to focus on the Design section as that's more concrete, bottom to top. However you approach it, when providing feedback please go top to bottom, improving earlier sections will have a cascade effect and ensure everything builds up logically.

compiler/modules/modulegraphs.nim
saem: thanks, going to give that a shot now.

compiler/ast/parser.nim
saem: Thanks, I like `consume`, going to go with a verbose name for now based on that..

saem:
I'm going to go with a more conservative change for the `ParsedNode` data layout internals.  I think the bigger win is converting the `lexer` to a run length encoding based approach first. Then rework the `parser` and propagate up through the compiler.

compiler/ast/ast.nim
```diff
@@ -637,38 +636,206 @@ proc toHumanStr*(kind: TTypeKind): string =
...
   case base
   of base10: discard
   of base2: incl(n.flags, nfBase2)
   of base8: incl(n.flags, nfBase8)
   of base16: incl(n.flags, nfBase16)
 
+func toTNodeKind*(kind: ParsedNodeKind): TNodeKind {.inline.} =
+  case kind
+  of pnkError: nkError
+  of pnkEmpty: nkEmpty
```
zerbina, line 650: Nothing for now, but it might make sense to use constant array like `[pnkError: nkError, ...]` here instead.

compiler/ast/ast.nim
```diff
@@ -637,38 +636,206 @@ proc toHumanStr*(kind: TTypeKind): string =
...
+          "tkIn, tkOut, etc need to be translated, probably in the parser for now"
+    result.ident = parsed.startToken.ident
+
+  of nkAccQuoted:
+    for (ident, line, col) in parsed.idents.items:
+      let info = TLineInfo(fileIndex: parsed.fileIndex, line: line, col: col)
+      result.add newIdentNode(ident, info)
+
+  of nkError:
+    assert false, "IMPLEMENT ME"
```
zerbina:
```suggestion
    unreachable("IMPLEMENT ME")
```

compiler/ast/ast_parsed_types.nim
```diff
@@ -17,44 +22,260 @@ import
...
+  pnkCallKinds* = {pnkCall, pnkInfix, pnkPrefix, pnkPostfix,
+                  pnkCommand, pnkCallStrLit}
 
 func len*(node: ParsedNode): int =
   ## Number of sons of the parsed node
   return node.sons.len()
 
 # NOTE added for the sake of API similarity between PNode
-proc safeLen*(node: ParsedNode): int = node.len()
+proc safeLen*(node: ParsedNode): int =
```
zerbina:
```suggestion
func safeLen*(node: ParsedNode): int =
```

compiler/ast/parser.nim
```diff
@@ -111,28 +127,32 @@ proc getTok(p: var Parser) =
...
                  cache: IdentCache; config: ConfigRef) =
   ## Open a parser, using the given arguments to set up its internal state.
-  ##
   initToken(p.tok)
   openLexer(p.lex, fileIdx, inputStream, cache, config)
   p.getTok                   # read the first token
   p.firstTok = true
-  p.emptyNode = newParsedNode(nkEmpty)
+  p.emptyNode = newEmptyParsedNode(p.lex.fileIdx)
+  p.diags = @[] # clear the diags in case we're reopening an existing parser
```
zerbina:
```suggestion
  p.diags.setLen(0) # clear the diags in case we're reopening an existing
                    # parser
```
In order to reuse the previous memory.

compiler/ast/parser.nim
```diff
@@ -1077,28 +1166,28 @@ proc parseSymbolList(p: var Parser, result: ParsedNode) =
...
     # XXX document this feature!
     let nodeKind =
-      if p.tok.ident.s == "with": nkWith
-      elif p.tok.ident.s == "without": nkWithout
-      else: return
+      case p.tok.ident.s
+      of "with":    pnkWith
+      of "without": pnkWithout
+      else:         return
+    let list = p.newNode(nodeKind, p.tok)
```
zerbina: `newNodeConsumingTok` can be used here.

compiler/utils/astrepr.nim
```diff
@@ -1006,6 +1027,7 @@ proc treeRepr*(
         add ">"
       else:
         add "<visited>"
+        echo cast[int](n)
```
zerbina: It this a leftover from debugging?

saem:
bors cancel

zerbina (approved):
Looks good to me. I'm not super familiar with the parser and surroundings, so I only left a some comments regarding general things.

Nonetheless, I think the decision of doing the data type rework separately was a good call.

compiler/utils/astrepr.nim
saem: whoops, yes it is.

compiler/ast/ast.nim
saem: The logic might not remain entirely straightforward, but I'll see after a few more changes.

Not sure the trade-offs are, though

PR #536, opened 2023-02-12
cgen: make `.global.`s work more according to spec

zerbina:
## Summary
For the C target, globals defined inside procedures that are resource- like (i.e. have a user-defined or lifted destructor) are now initialized in a module's pre-init procedure, instead of each time control-flow reaches the definition. This mirrors what already happened for non-resource-like globals.

In addition, globals defined in `inline` procedures no longer cause linking errors because of duplicated definitions in the generated C code.

## Details
The globals are extracted before translating the AST to MIR code, making the workaround in `mirgen` (that didn't work anyway) obsolete. In order to work towards unifying the architecture of the back-ends, `vmbackend` now also makes use of the pre-extraction, no longer requiring extra logic in the code-generator.

### Known Issues
- the destructor-injection pass is not run for the initializer expression of globals defined inside procedures
- `.global.`s on the JS target still don't work according to specification

---

## Notes for Reviewers
The PR is meant to unblock #450 with the least amount of work and changes to semantics. I think that the introduced regression (no destructor injection for initializers of procedure-level globals) is acceptable until the implementation gets an overhaul.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/injectdestructors.nim
+++ b/compiler/sem/injectdestructors.nim
@@ -1103,6 +1103,14 @@ func shouldInjectDestructorCalls*(owner: PSym): bool =
      {sfInjectDestructors, sfGeneratedOp} * owner.flags == {sfInjectDestructors} and
      (owner.kind != skIterator or not isInlineIterator(owner.typ))
 
+proc deferGlobalDestructor*(g: ModuleGraph, idgen: IdGenerator, owner: PSym,
+                            global: PNode) =
+  ## If the global has a destructor, emits a call to it at the end of the
+  ## section of global destructors.
+  if sfThread notin global.sym.flags and hasDestructor(global.typ):
+    var c = Con(owner: owner, graph: g, idgen: idgen)
+    g.globalDestructors.add genDestroy(c, global)
+
 proc injectDestructorCalls*(g: ModuleGraph; idgen: IdGenerator; owner: PSym; n: PNode): PNode =
   when toDebug.len > 0:
     shouldDebug = toDebug == owner.name.s or toDebug == "always"
--- a/compiler/sem/transf.nim
+++ b/compiler/sem/transf.nim
@@ -1154,3 +1154,60 @@ proc transformExpr*(g: ModuleGraph; idgen: IdGenerator; module: PSym, n: PNode):
     # expressions are not to be injected with destructor calls as that
     # the list of top level statements needs to be collected before.
     incl(result.flags, nfTransf)
+
+proc extractGlobals*(body: PNode, output: var seq[PNode], isNimVm: bool) =
+  ## Searches for all ``nkIdentDefs`` defining a global, appends them to
+  ## `output` in the order they appear in the input AST, and removes the nodes
+  ## from `body`. `isNimVm` signals which branch to select for ``when nimvm``
+  ## statements/expressions.
+  ##
+  ## XXX: this can't happen as part of ``transformBody``, as ``transformBody``
+  ##      is reentrant because of ``lambdalifting`` and it's thus not easily
+  ##      possible to collect something from the body of a single procedure
+  ##      only. There's also the problem that extracting the globals is not
+  ##      wanted when transformation happens for a procedure that's invoked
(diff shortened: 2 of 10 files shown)


compiler/sem/injectdestructors.nim
```diff
@@ -1103,6 +1103,14 @@ func shouldInjectDestructorCalls*(owner: PSym): bool =
      {sfInjectDestructors, sfGeneratedOp} * owner.flags == {sfInjectDestructors} and
      (owner.kind != skIterator or not isInlineIterator(owner.typ))
 
+proc deferGlobalDestructor*(g: ModuleGraph, idgen: IdGenerator, owner: PSym,
+                            global: PNode) =
+  ## If the global has a destructor, emits a call to it to the end of the
```
saem:
```suggestion
  ## If the global has a destructor, emits a call to it at the end of the
```

compiler/sem/transf.nim
```diff
@@ -1154,3 +1154,60 @@ proc transformExpr*(g: ModuleGraph; idgen: IdGenerator; module: PSym, n: PNode):
...
+    discard "ignore declarative contexts"
+  of nkWithoutSons:
+    discard "not relevant"
+  of nkConv, nkHiddenStdConv, nkHiddenSubConv:
+    # only analyse the imperative part:
+    extractGlobals(body[1], output, isNimVm)
+  of nkWhen:
+    # a ``when nimvm`` construct
+    # XXX: this logic duplciates what ``mirgen`` already does. Maybe
+    #      collecting should happen there? Or procedure-level globals should
```
saem: Likely makes sense in sem as a normalization step.

compiler/sem/transf.nim
```diff
@@ -1154,3 +1154,60 @@ proc transformExpr*(g: ModuleGraph; idgen: IdGenerator; module: PSym, n: PNode):
...
+     nkNimNodeLit:
+    discard "ignore declarative contexts"
+  of nkWithoutSons:
+    discard "not relevant"
+  of nkConv, nkHiddenStdConv, nkHiddenSubConv:
+    # only analyse the imperative part:
+    extractGlobals(body[1], output, isNimVm)
+  of nkWhen:
+    # a ``when nimvm`` construct
+    # XXX: this logic duplciates what ``mirgen`` already does. Maybe
```
saem:
```suggestion
    # XXX: this logic duplicates what ``mirgen`` already does. Maybe
```

compiler/sem/transf.nim
```diff
@@ -1154,3 +1154,60 @@ proc transformExpr*(g: ModuleGraph; idgen: IdGenerator; module: PSym, n: PNode):
...
+    #      collecting should happen there? Or procedure-level globals should
+    #      be lifted from procedures during semantic analysis already?
+    let branch =
+      if isNimVm: body[0][1]
+      else:       body[1][0]
+
+    extractGlobals(branch, output, isNimVm)
+  of nkVarSection, nkLetSection:
+    # iterated over all children and extract identdefs of globals:
```
saem:
```suggestion
    # iterates over all children and extract identdefs of globals:
```

compiler/vm/vmbackend.nim
```diff
@@ -177,8 +177,13 @@ proc generateTopLevelStmts*(module: var Module, c: var TCtx,
 
   module.initProc = (start: start, regCount: c.prc.regInfo.len)
 
-proc generateCodeForProc(c: var TCtx, s: PSym): VmGenResult =
+proc generateCodeForProc(c: var TCtx, s: PSym,
+                         globals: var seq[PNode]): VmGenResult =
+  ## Generates and emits the bytecode for the procedure `s`. The globals
+  ## defined in it it are extracted from the body and their identdefs appended
```
saem:
```suggestion
  ## defined in it are extracted from the body and their identdefs appended
```

saem (approved):
Bors r+

Thanks for the detailed explanation, it was really easy to follow.

PR #538, opened 2023-02-13
vmgen: don't materialize unnecessary copies

zerbina:
## Summary
A full copy was materialized when assigning to locals in a few specific cases. This was necessary in order to prevent false alising between locations, back when the `PNode`-based memory representation was still used -- it's no longer needed now.

Removing the logic and usage simplifies `vmgen` a bit and also means that no additional copy is introduced for assignments such as `local.a = complex` or `local[i] = complex`, resulting in a small time efficiency improvement for code using these.

---

## Notes for Reviewers
* this is another preparation for the address / `var` parameter support in the VM

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


saem:
Bors r+

zoom zoom

PR #541, opened 2023-02-14
tests: introduce better category `lang_syntax

saem:
## Summary
Renamed `compiler_integration` to `lang_syntax` as that's what we're really testsing.

The fundamental thing is to ensure we specify the lexical, grammar, syntax, etc. Unlike many other languages it's an API/data interchange protocol.

---

## Notes for Reviewers
* review is probably useful in terms of gleaning direction wrt
  * the language
  * test categorization


saem:
Bors r+

PR #545, opened 2023-02-15
report illegal branch switch during semantic analysis

zerbina:
## Summary
Assigning to a discriminant of a variant object (i.e. switching the active branch) that has a user-provided destructor is a semantic error, but it was checked for and reported during the `injectdestructors` pass, which, as of now, is part of the mid-/back-end phase.

It is now reported in `sempass2`, which also fixes the issue of no error being reported when the offending assignment is not part of the alive code.

---


saem (approved):
Bors r+

PR #548, opened 2023-02-20
vm: improve support for `var`, `lent`, and `addr`

zerbina:
## Summary

- support implicit and explicit to-`pointer` conversions
- the `ptr` value produced by taking the address of a primitive local (e.g. int, float, etc.) can now be assigned to a field or array-slot. An internal error was previously raised by the VM at run-time
- fix many issues in edge-cases related to `var`|`lent` parameters, return values, and views
- fix pointers to primitive locals not being checked for memory access violations (e.g. using a pointer to access a local that went out of scope wasn't detected as a violation)
- fix an evaluation-order issue with `parseBiggestFloat` in the VM
- for-vars (e.g. the `i` in `for i in ...`) are now properly treated as globals
- significantly improve efficiency of accessing `var T` parameters where `T` is a compound type or string
- improve the efficiency of assigning to global variables

Note that all of the above fixes and improvements only apply to code running in the VM -- the C and JS targets are not affected.

### Motivation

The motivation behind the rework is to get rid of the very costly handle <-> address overhead, which was previously incurred on every `var T` parameter access where `T` is a non-primitive type.

Another goal is the removal of the register-addresses (`rkRegAddr`), which, aside from complicating the VM's design, have multiple issues:
- they're not subjected to the standard access checks
- they require registers to be stored at fixed host memory locations, which prevents various optimizations to the VM
- they make pointers in the VM dynamically typed: a pointer can be either a "real" address or a register-address

### `vmgen` changes

- remove the remaining `gfNode` flag. It is replaced with either a `load` parameter or with using separate procedures
- don't unconditionally treat `var` and `lent` types as pointers
  - `var T` parameters and return variables use handles
  - fields and globals that are direct views are stored as pointers
- code generation for view assignments and `nkHiddenDeref` takes the above difference into account
- all locals that need to be addressable via a handle or pointer (they have their address taken or are passed to a ``var`` parameter) are prevented from using register storage. This only applies to locals of primitive type (ints, floats, `NimNode`, and pointers), as the others never use register storage
- since its now obsolete, the logic for "patching assignments" is removed
- magics that mutate their operands but don't use `nkHiddenAddr` require special "write back" logic in order to support their operands using stack-cell storage. This is implemented via `Loc` and `genLoc`

The approach to supporting `addr` and views for locals of primitive type is similar to the one employed by `jsgen`, but instead of relying on the `sfAddrTaken` flag, an AST-based analysis for gathering all locals that need to be addressable via a handle or pointer is run prior to code generation.

Result variable and immutable parameters of primitive type that need to be addressable via handles or pointers require dedicated handling:
- the result variable uses stack-cell storage and is transitioned to register storage prior to non-exceptional procedure exits
- immutable parameter registers are transitioned to stack-cell storage in the procedure preamble

### VM changes

- replace `opcAddrReg` and `opcAddrNode` with the new, simpler `Addr` operation
- remove the `rkRegAddr` register kind and all work-arounds related to it
- remove the `LdGlobalAddr` operation -- it is replaced with usage of `LdGlobal` + `Addr`

### Views vs. `var` parameters

The way both are now implemented by `vmgen` indicates that the notion started by the MIR of them being two different things is the right direction.

The difference is not fully enforced through the AST reaching the code generator yet, which leads to much of the handling being implicit, making the implementation harder to follow.

This will eventually be improved upon by either basing the code generator on the MIR or (more likely) a dedicated IR.


tests/lang_defs/distinct/tdistinct.nim
```diff
@@ -1,5 +1,5 @@
 discard """
-  targets: "c js"
+  targets: "c js vm"
```
saem: Note to self: have a look at the `lang_defs` category and see if it makes sense to just default the whole thing to all targets.

zerbina:
I'm going to try and split off some of the changes here. Not all of them are directly related  to making the address handling work, and the current impact is a bit large.

zerbina:
That should wrap this PR up.

Due to rather large architectural changes to `vmgen`, I think that further changes to both `vmgen` and the VM should wait until this PR is merged.

compiler/vm/vmgen.nim
```diff
@@ -79,6 +88,17 @@ type
...
     diag: VmGenDiag
 
+  Loc = object
+    ## An encapsulation that associates the register storing the value with
+    ## the register storing the handle of the location it was loaded from.
+    ##
+    ## `Loc` is used for convenient write-back handling. A write-back is
+    ## necessary where an instruction operates on and modifies a register that
+    ## stores a simple value (e.g. int, float), and the modification needs to
+    ## be ## reflected on the memory location the value originated from
```
```suggestion
    ## be reflected at the memory location the value originated from
```

compiler/vm/vmgen.nim
```diff
@@ -2506,29 +2754,103 @@ proc genAddr(c: var TCtx, src, n: PNode, dest: var TDest) =
...
+    assert isVarLent(n[0].typ)
+    if isPtrView(n[0]):
+      # we want a handle (``rkHandle``), but the input view uses a pointer
+      # (``rkAddress``) internally. Turn it into a handle by dereferencing it
+      prepare(c, dest, n[0].typ)
+      let tmp = c.genx(n[0])
+      c.gABC(n, opcLdDeref, dest, tmp)
+      c.freeTemp(tmp)
+    else:
+      # the operand is handle already; treat the deref as a no-op
```
```suggestion
      # the operand is a handle already; treat the deref as a no-op
```

compiler/vm/vmgen.nim
```diff
@@ -2142,13 +2284,118 @@ proc genFieldAsgn(c: var TCtx, obj: TRegister; le, ri: PNode) =
...
+
 proc genAsgn(c: var TCtx; le, ri: PNode; requiresCopy: bool) =
   case le.kind
   of nkBracketExpr:
     let typ = le[0].typ.skipTypes(abstractVarRange-{tyTypeDesc}).kind
-    let dest = c.genx(le[0], {gfNode})
+    let dest = c.genx(le[0])
     let idx = if typ != tyTuple: c.genIndex(le[1], le[0].typ) else: le[1].intVal
-    let tmp = c.genx(ri)
+    let tmp = c.genAsgnSource(ri, true)
```
```suggestion
    let tmp = c.genAsgnSource(ri, wantsPtr = true)
```

compiler/vm/vmgen.nim
```diff
@@ -2142,13 +2284,118 @@ proc genFieldAsgn(c: var TCtx, obj: TRegister; le, ri: PNode) =
...
+  else:
+    if usesRegister(c.prc, s):
+      # if the location is backed by a register (i.e., is not in stored
+      # in a memory cell), we don't use a temporary register + assignment
+      # but directly write to the destination register
+      gen(c, ri, local(c.prc, s))
+    else:
+      # an assignment is required to the local is required. Views are always
+      # stored as handles in this case, so a register move is used for
+      # assigning them
```
```suggestion
      # an assignment is required to the local. Views are always stored as
      # as handles in this case, so a register move is used for assigning them
```

saem (commented):
Some minor suggestions, but good to go.

compiler/vm/vmgen.nim
zerbina: Done. I also updated all other occurrences.

zerbina:
The review comments are addressed and the PR is rebased and squashed.

bors r+

PR #549, opened 2023-02-20
compiler: remove support for the legacy GCs

zerbina:
## Summary

There are multiple issues with the legacy memory management strategies
(the ones that are not destructor-based):

* each one introduces another language dialect
* they necessitate the old runtime (more code to maintain)
* they're specific to the C backend, making them not fit in well with
  the direction of having each backend support the full language
* the way they're currently integrated with the compiler introduces a
  large amount of complexity and conditional logic to the C code-
  generator

For the reasons listed above, compiler support for both the legacy GCs
and large parts of the old runtime is removed. The only remaining
selectable GC modes are `orc` and `arc`.

Except for the v1 run-time-type-information, which is still required by
the `deepcopy` implementation and `marshal` module, the old runtime is
now obsolete and can be removed.

## Details

**CLI:**

* adjust the allowed `gc` switch arguments to only include
  `destructors`, `arc`, and `orc`
* remove the obsolete `--seqsv2` option (the version 2 implementation is
  always used now, at least for the C backend)

**C code-generator:**

* remove all usages of both `optSeqDestructors` and `optTinyRtti` --
  they're now unconditionally enabled when using the C backend
* remove `optTinyRtti`, as it has no remaining usage-sites
* remove all code related to the old runtime's `string`s and `seq`s
* remove all support for the garbage collectors using the old runtime
* remove code-generation logic for the unused magics `mIntToStr`,
  `mInt64ToStr`, and `mFloatToStr`
* simplify the code generated for the entry point

**Tests:**

* remove tests specific to the removed GCs
* disable the tests for `--gc:none`
* don't specify `--gc:arc` for some tests where not required, allowing
  the tests to be joined

**Documentation:**

* clean up the memory management documentation (`gc.rst`)
* remove some mentions of the removed GCs and the old runtime

**CI:**

* remove `boehm`-related package dependencies

**Misc:**

* remove `when` statements related to the removed GCs from compiler code
* support `koch` building the first iteration with `--gc:orc`, which is
  a preparation for updating the csources compiler to one that doesn't
  support `--gc:refc`

For the purpose of backwards compatibility, the now removed GC modes can
still be tested for with `compileOption`.

### Removal of `--gc:none`

It is currently not clear how `--gc:none` should work under the new
runtime, as no traditional GC that can be disabled by simply not running
it exists. Sequences and strings are also no longer ref-based, making it
unclear how these should work.

For the time being, the option is removed, but with the intention to
bring it back once the semantics are decided.


compiler/backend/ccgliterals.nim
```diff
@@ -93,9 +60,4 @@ proc genNilStringLiteral(m: BModule; info: TLineInfo): Rope =
   result = ropecg(m, "((#NimStringDesc*) NIM_NIL)", [])
 
 proc genStringLiteral(m: BModule; n: PNode): Rope =
-  case detectStrVersion(m)
-  of 0, 1: result = genStringLiteralV1(m, n)
-  of 2: result = genStringLiteralV2(m, n, isConst = true)
-  else:
-    internalError(
-      m.config, n.info, "cannot determine how to produce code for string literal")
+  result = genStringLiteralV2(m, n, isConst = true)
```
saem, line 63: Removing all the `V2` suffix stuff is part of the runtime removal?

compiler/backend/ccgstmts.nim
```diff
@@ -15,21 +15,6 @@ const
   stringCaseThreshold = 8
     # above X strings a hash-switch for strings is generated
 
-proc getTraverseProc(p: BProc, v: PSym): Rope =
```
saem, line 18: It's fantastic how much code we're dropping here, further drives the idea that our backends should greatly simplify and we move more of the brains earlier in compilation.

compiler/front/commands.nim
```diff
@@ -61,6 +46,14 @@ from compiler/ast/report_enums import ReportKind,
   repWarningKinds,
   repWarningGroups
 
+template bootSwitch(name, expr, userString) =
+  # Helper to build boot constants, for debugging you can 'echo' the else part.
+  const name = if expr: " " & userString else: ""
+
+bootSwitch(usedRelease, defined(release), "-d:release")
```
saem: Being very pedantic the imports can't see these, so timing wise it's happening a bit late. But I don't think there are any real dependencies on them right now. Might be a latent bug.

tests/ccgbugs/tgc_bug.nim
```diff
@@ -1,5 +1,4 @@
 discard """
-matrix: "--gc:refc -d:fulldebug -d:smokeCycles"
```
saem, line 2: Why drop the defines while keeping them in `twrongrefcounting.nim`?

tests/lang_callable/iter/titer2.nim
```diff
@@ -8,9 +8,13 @@ discard """
 2
 3
 4'''
-  cmd: "nim $target --gc:none --hints:on --warnings:off $options $file"
+  #cmd: "nim $target --gc:none --hints:on --warnings:off $options $file"
```
saem:
```suggestion
  #matrix: "--gc:none --hints:on --warnings:off"
```

Just to phasing out more of these `cmd`s; doesn't have to be done figured while the line is being edited regardlesss.

saem (approved):
This is great!

A really nice benefit of this is that we can start standardizing the gdb debug script all that much more easily and include it in our CI testing. That should give us decent debugging and making exploring/learning so much easier.

alaviss:
So with this I persume that we are gonna build the compiler with ORC by default?

If so then we should remove the `orc` job from CI configuration.

compiler/backend/ccgliterals.nim
zerbina: Yep, my thinking was that it's cleaner to remove all the suffixes (both in the compiler and runtime) at once.

compiler/front/commands.nim
zerbina: I'm not sure if that's what you mean, but the imports also weren't able to see these previously, as the constants are not exported.

tests/ccgbugs/tgc_bug.nim
zerbina: I'm not entirely certain, but believe that it's an oversight -- I'm going to restore the defines here.

zerbina:
> So with this I persume that we are gonna build the compiler with ORC by default?

How do you mean? The compiler is already built with `--gc:orc`, see #547. The reason I didn't remove the CI job was to keep ensuring that all three bootstrap iterations work with ORC, as the current csources compiler is not able to build the compiler with `--gc:orc` and thus still uses `--gc:refc`.

> If so then we should remove the `orc` job from CI configuration.

Yep, will do.

compiler/backend/ccgliterals.nim
saem: That does sound cleaner.

compiler/front/commands.nim
saem: Sorry, I got my wires crossed, I barely slept yesterday. Anyhow, what I meant was that the other modules could play around with the defines before we get a chance to capture them here.

saem:
Did we want to cut a sources before or after this change?

compiler/front/commands.nim
zerbina: Hm, that's indeed possible, though I think such behaviour (i.e., defines inside a module persisting) should be disallowed.

I'll move them back up and leave a note.

zerbina: Hm, that's indeed possible, though I think such behaviour (i.e., defines inside a module persisting) should be disallowed.

I'll move them back up and leave a note.

zerbina:
> Did we want to cut a sources before or after this change?

If the aim is full coverage at all times, then it's probably better to update the csources *before* removing the ORC bootstrap job. I'm indifferent, and okay with restoring the `orc` job  again.

saem:
IIRC, the csources are generated every time we merge into `devel` as one of the artifacts. So if the job removal is not part of this PR, then we'll maintain coverage, can use this PR's very own release for csources. Meaning a follow-up change can drop the job and update the csources as an atomic operation.

saem:
I'm pretty sure this is good to go, feel free to merge if you're all set, @zerbina.

zerbina:
/merge

All set. It's time to merge.

PR #551, opened 2023-02-20
compiler: introduce an IR for the code generators

zerbina:
## Summary

Add the `CgNode` intermediate representation, change `astgen` (which is
renamed to `cgirgen`) to output it instead of `PNode`, and update all
code generators to use `CgNode`. In order to keep the changes required
for the transition low, the differences of the new IR compared to
`PNode` are kept minimal.

The intent is to have an IR that can be evolved independently from sem
and the macro API. Only `PNode` is replaced so far, but both `PType`
and `PSym` are planned to also get a dedicated version for use by the
code generators. In addition, `PNode` can now be evolved without the
code generators having to be considered.

## Details

The core of the changes is the introduction of `CgNode`. It, for now,
also uses a `ref`-tree-based data-representation and is very similar,
in both structure and naming, to the `PNode`-subset previously used by
the code generators, but with some small simplifications/renamings
already applied where it makes sense.

**Naming differences:**
- `nkNimNodeLit` -> `cnkAstLit`
- `nkCurly` -> `cnkSetConstr`
- `nkBracket` -> `cnkArrayConstr`
- `nkClosure` -> `cnkClosureConstr`
- `nkHiddenDeref` -> `cnkDerefView`
- `nkHiddentStdConv` -> `cnkHiddenConv`
- `nkDiscardStmt` -> `cnkVoidStmt`
- `nkExprColonExpr` -> `cnkBinding`

**Structural differences:**
- the `while` statement is replaced with the `repeat` statement, since
  conditional loops (i.e. `while` statements) don't exist after the MIR
  phase. A `cnkRepeatStmt` node has a single sub-node, which is the
  loop's body
- there are no multi-branch `if` statements, since those don't exist
  during and after the MIR phase. A `cnkIfStmt` has the same structure
  as an `nkElifBranch`
- a `cnkReturnStmt` node has no sub nodes
- instead of `emit` being represented via a pragma statement, a
  dedicated statement (`cnkEmitStmt`) is used
- `cnkPragmaStmt`s don't have sub-nodes, but instead directly store the
  pragma's name
- `cnkObjConstr` doesn't have an extra type slot in the first position 
- there is only a single literal-node kind for the `int`, `uint`,
  `float`, and `string` types
- there are no dedicated nodes for `of` and `else` in `case` statements,
  both use `cnkBranch`
- there are no `nkVarSection|nkLetSection` + `nkIdentDefs` counterparts.
  Instead, definitions use the `cnkDef` node

Many more simplifications are possible, but they are left for future
PRs.

### Producing the IR

`astgen` is changed to produce `CgNode` instead of `PNode` IR. The
`canUseView` and `flattenExpr` procedures for `PNode` are copied and
adjusted for `CgNode`.

For literals, which are transported through the MIR phase as `PNode`s,
translation-to`CgNode` logic is added (`translateLit`). For integer-like
literals, whether the output node is an `cnkIntLit` or `cnkUIntLit` node
is decided based on the input node's *type*, not on its node *kind* --
this helps catching issues with incorrectly typed integer literals. For
`float32` literals, `translateLit` already narrows the value.

Since the module's name now doesn't apply anymore, it is changed to
`cgirnode`. In addition, the module is moved to the `backend` directory
and `generateAST` is renamed to `generateIR` -- the module's
documentation is also updated.

### Code generation

All three code generators and their surroundings are adjusted to operate
on `CgNode` instead of `PNode`, which for the most part means replacing
occurrences of `PNode` with `CgNode` and adjusting for the structural
difference listed above. Unrelated refactorings and clean-ups are kept
to a minimum.

There are two sources of `PNode`s still reaching into the code
generators: constants and type AST. When generating the code for
constants, their AST is first translated to `CgNode` IR before being
passed on to normal code generation, while for the type AST case,
special-purpose routines that still use `PNode` are added.

Routines still used by the code generators that are only available for
`PNode` are copied into the new `compat` module and adjusted to use
`CgNode`. This is meant to be an interim solution, and they are planned
to be phased out and removed again in the future.

Both the C and JavaScript code generation orchestrators passed the
`PNode` body of the main procedure directly to the code generator. This
doesn't work anymore, and it is replaced with canonicalizing (which
produces a `CgNode` tree) the body first.


zerbina:
Attempting to derive the first revision from the MIR was a mistake, as too much changes to the code generators are required for that to work. I'm going to revert all changes and start over, but now with a focus on keeping both the overall changeset and development of the new IR to a minimum.

However, there is still some decoupling left do to before the work here can resume. This includes things like making dynlib handling part of the unified backend processing pipeline, or replacing/removing dependencies on routines that are not part of the code generators and use `PNode` (e.g., `dfa.aliases`, `TNodeTable`, etc.).

saem:
i really like all the node simplification and rework, one question about it, did you already consider naming `cnkVoidStmt` as `cnkVoidExpr`? Asking because I read the node as `void`ing an expression.

zerbina:
> i really like all the node simplification and rework, one question about it, did you already consider naming `cnkVoidStmt` as `cnkVoidExpr`? Asking because I read the node as `void`ing an expression.

Hm, no, I didn't consider it. My thinking was that while it voids (discards) a value, it itself acts as a statement (returns no value), so, for consistency with the other statements, used the `Stmt` suffix.

compiler/backend/cgir.nim
```diff
@@ -0,0 +1,203 @@
...
+    cnkFinally
+
+    cnkBranch     ## the branch of a ``case`` statement
+
+    cnkDef        ## starts the lifetime of a local and optionally assigns an
+                  ## initial value
+
+    cnkAsgn       ## a = b
+    cnkFastAsgn   ## fast assign b to a
+    # future direction: have ``cnkAsgn`` mean "assign without implying any
```
saem, line 128: I wonder if a better distinction is assign vs initialize?

For context I'm simply broadly remarking upon nomenclature.

compiler/backend/cgir.nim
```diff
@@ -0,0 +1,203 @@
...
+    of cnkIntLit, cnkUIntLit:
+      # future direction: use a ``BiggestUint`` for uint values
+      intVal*: BiggestInt
+    of cnkFloatLit:   floatVal*: BiggestFloat
+    of cnkStrLit:     strVal*: string
+    of cnkAstLit:     astLit*: PNode
+    of cnkSym:        sym*: PSym
+    of cnkPragmaStmt: pragma*: TSpecialWord
+    of cnkWithItems:
+      childs*: seq[CgNode]
```
saem:
```suggestion
      kids*: seq[CgNode]
```

Shorter, grammatically "correcter" (😁), and also not needlessly gendered ("sons").

compiler/backend/cgir.nim
```diff
@@ -0,0 +1,203 @@
...
+
+  # future direction: move to a single-sequence-based, data-oriented design
```
saem, line 158: ♥️

compiler/backend/jsbackend.nim
```diff
@@ -19,6 +19,9 @@ import
   compiler/front/[
     options
   ],
+  compiler/mir/[
+    mirbridge, # for `canonicalize`
```
saem: Is this preferable as compared to a selective import?

compiler/backend/jsgen.nim
```diff
@@ -704,12 +711,12 @@ proc genWhileStmt(p: PProc, n: PNode) =
   p.blocks[^1].isLoop = true
   let labl = p.unique.rope
   lineF(p, "Label$1: while (true) {$n", [labl])
-  p.nested: genStmt(p, n[1])
+  p.nested: genStmt(p, n[0])
```
saem, line 714: The index change is because of the new structure of repeat vs while?

saem (approved):
this really is an incredible milestone in decoupling sem and the backends, along with backend unification! 🎉

compiler/backend/cgir.nim
zerbina: That's a good question. Everything related to assignment is very fuzzy at the moment, mainly because of "assignment" having different meaning depending on the used code generator (they don't operate on the same language level with regards to assignments).

Broadly speaking, `cnkAsgn` combines both "assign" (copy to non-empty destination) *and* "initialize" (copy to empty destination), while `cnkFastAsgn` means "create a shallow copy". Whether combining "assign" and "initialize" is a good idea, I'm not sure (it probably isn't), but `PNode` did it, so for initial compatibility, I carried it over.

saem: Sounds like we're on the same page, needs some more thinking.

compiler/backend/cgir.nim
zerbina: I'll change it.

compiler/backend/jsbackend.nim
zerbina: No, a selective import is, in my opinion, preferable here. I'm unsure why I originally opted to not use one.

compiler/backend/jsgen.nim
zerbina: Yep, the loops body was previously in the second slot, but now it's in the first (and only) one.

zerbina:

PR #552, opened 2023-02-22
compiler: remove some unused routines and variables

zerbina:
## Summary
Remove unused routines and variables from modules related to the VM and code generation. They were either never used or became unused as a side-effect of other removals / cleanup.

### DIFF
--- a/compiler/backend/ccgliterals.nim
+++ b/compiler/backend/ccgliterals.nim
@@ -27,9 +27,6 @@ template detectVersion(field, corename) =
 proc detectStrVersion(m: BModule): int =
   detectVersion(strVersion, "nimStrVersion")
 
-proc detectSeqVersion(m: BModule): int =
-  detectVersion(seqVersion, "nimSeqVersion")
-
 # ----- Version 1: GC'ed strings and seqs --------------------------------
 
 proc genStringLiteralDataOnlyV1(m: BModule, s: string): Rope =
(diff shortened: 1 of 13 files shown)


compiler/backend/ccgliterals.nim
```diff
@@ -27,9 +27,6 @@ template detectVersion(field, corename) =
 proc detectStrVersion(m: BModule): int =
   detectVersion(strVersion, "nimStrVersion")
 
-proc detectSeqVersion(m: BModule): int =
-  detectVersion(seqVersion, "nimSeqVersion")
-
 # ----- Version 1: GC'ed strings and seqs --------------------------------
 
 proc genStringLiteralDataOnlyV1(m: BModule, s: string): Rope =
```
saem, line 32: On the phone and doing sketchy GitHub searches but this proc might be unused now.

compiler/backend/ccgliterals.nim
```diff
@@ -92,17 +89,6 @@ proc genStringLiteralV2Const(m: BModule; n: PNode; isConst: bool): Rope =
 
 # ------ Version selector ---------------------------------------------------
 
-proc genStringLiteralDataOnly(m: BModule; s: string; info: TLineInfo;
-                              isConst: bool): Rope =
-  case detectStrVersion(m)
-  of 0, 1: result = genStringLiteralDataOnlyV1(m, s)
-  of 2:
-    result = getTempName(m)
-    genStringLiteralDataOnlyV2(m, s, result, isConst)
```
saem, line 101: Maybe thing one is unused, too

saem (approved):
There is a slim chance I might have one or two more dead procs, hopefully they're not false positives.

Thanks, the less code the better.

compiler/backend/ccgliterals.nim
zerbina: Not yet, `genStringLiteralV1` still uses it. Removing `refc` and friend will indeed make it obsolete, however.

compiler/backend/ccgliterals.nim
zerbina: It's used by `genStringLiteralV2`.

zerbina:
Thanks for taking a look, @saem.

bors r+

PR #553, opened 2023-02-23
vm: support the return-value optimization

zerbina:
Summary
-------

Allow (non-magic) procedures to modify the destination location directly if the result of the call is immediately assigned to something. While the optimization is never actually made use of right now (a temporary is always used), this change makes it possible to do so in the future.

Details
-------

Instead of the VM allocating the location as part of executing `IndCallAsgn`, `vmgen` now emits bytecode for doing so at the callsite (if the destination is not a valid location).

For values that fit into registers, the result register is setup at the start of a procedure via `LdNullReg` (nothing is emitted at the callsite in this case).

While supporting the return-value optimization is the main goal, the change is also a step towards:
1. allocating locations in a single place (i.e. via `LdNull`), which is a preparation for removing the `rkLocation` register kind
2. not having to store the return type in the function-table entry

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -989,13 +989,13 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): YieldReason
 
         assert c.code[pc].opcode in {opcIndCall, opcIndCallAsgn}
         if c.code[pc].opcode == opcIndCallAsgn:
-          # Move the return value to the destination register on the
+          # move the return register's content (that stores either the simple
+          # value or the destination handle) to the destination register on the
           # caller's frame
-          let p = c.sframes[tos].next
-          let i = c.code[pc].regA
-          let slot = addr c.sframes[p].slots[i]
-          slot[].cleanUpReg(c.memory)
-          slot[] = move regs[0]
+          let
+            p = c.sframes[tos].next
+            i = c.code[pc].regA
+          c.sframes[p].slots[i] = move regs[0]
 
         popFrame()
       else:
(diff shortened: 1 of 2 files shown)


compiler/vm/vm.nim
```diff
@@ -2107,13 +2107,17 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): YieldReason
       of ckCallback:
         # it's a callback:
         if instr.opcode == opcIndCallAsgn:
-          # XXX: instead of doing this manually here, vmgen will emit
-          #      `opcLdNull(Reg)` in the future instead
+          # XXX: the callback should be responsible for initializing the
+          #      result register (we wouldn't need the setup logic here then),
+          #      but in order to be able to do that, owning handles need (i.e.
+          #      ``rkLocation``) need to be removed first
```
```suggestion
          #      ``rkLocation``) to be removed first
```

compiler/vm/vmgen.nim
```diff
@@ -413,6 +430,18 @@ func prepare(c: var TCtx, dest: var TDest, typ: PType) =
...
 
+proc prepare(c: var TCtx, dest: var TDest, n: PNode, typ: PType) =
+  ## If `dest` is not already set or refers to an argument slot, allocates a
+  ## location of `typ` and assigns the register holding the resulting handle or
+  ## value to it. `n` only provides the line information to use for emitted
+  ## instructions
+  if dest.isUnset:
+    dest = c.getFullTemp(n, typ)
+  elif c.prc.regInfo[dest].kind == slotTempUnknown and not fitsRegister(typ):
+    # the destination is an register slot used for argument passing, and the
```
```suggestion
    # the destination is a register slot used for argument passing, and the
```

compiler/vm/vmgen.nim
```diff
@@ -1333,7 +1362,7 @@ proc genSetElem(c: var TCtx, n: PNode, typ: PType): TRegister {.inline.} =
   let first = toInt(c.config.firstOrd(t))
   genSetElem(c, n, first)
 
-proc fitsRegister*(t: PType): bool =
+proc fitsRegister(t: PType): bool =
```
saem, line 1365: Yay, less public surface area!

saem (approved):
Awesome.

I spotted some minor non-blocking typos.

PR #554, opened 2023-02-26
types: fix all non-primitve types being treated as cyclic

zerbina:
## Summary

Due to a logic bug with `canFormAcycle`, *all* non-primitive types (such
as tuples, objects, `array`s, `seq`s, etc.) not explicitly marked with
`.acyclic` were detected as being cyclic.

Since `canFormAcycle` is used to decide whether a type is relevant to
the cycle collector when using `--gc:orc`, all heap locations of
non-primitive type were tracked and scanned by the cycle collector, and
unnecessary `=trace` hooks lifted for them.

As a consequence of the above, moving `ref` values between threads or
holding references to a (non-cyclic) heap location from multiple threads
was effectively impossible (without crashing) when using `--gc:orc`.

## Details

The logic for detecting types that can potentially act as the root in a
reference cycle is rewritten from scratch, now properly handling
structural types and generic instantiations. The `.cursor` annotation
is also taken into account.

Since `canFormAcycle` is still used by logic related to the legacy GCs it
is kept for now, in order to reduce the impact of the fix.

For testing the whether is potentially cyclic, the `isCyclical`
type-trait query procedure is added to `std/typetraits`.

### Misc

- remove leftover debug code from `orc.nim`

---

## Notes for Reviewers
* this is a hard blocker for #547

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/ast/types.nim
+++ b/compiler/ast/types.nim
@@ -28,6 +28,7 @@ import
   ],
   compiler/utils/[
     platform,
+    idioms,
     int128,
   ],
   compiler/modules/[
@@ -351,6 +352,8 @@ proc canFormAcycleAux(marker: var IntSet, typ: PType, startId: int): bool =
   case t.kind
   of tyTuple, tyObject, tyRef, tySequence, tyArray, tyOpenArray, tyVarargs:
     if t.id == startId:
+      # XXX: this check leads to all types not being explicitly marked as
+      #      acyclic to be treated as cyclic!
       result = true
     elif not containsOrIncl(marker, t.id):
       for i in 0..<t.len:
@@ -929,6 +932,14 @@ proc sameBackendType*(x, y: PType): bool =
   c.cmp = dcEqIgnoreDistinct
   result = sameTypeAux(x, y, c)
 
+proc sameLocationType*(x, y: PType): bool =
+  ## Tests and returns whether `x` and `y` are the same when used as the type
+  ## for locations
+  var c = initSameTypeClosure()
+  c.flags.incl IgnoreTupleFields
+  c.cmp = dcEqIgnoreDistinct
+  result = sameTypeAux(x, y, c)
+
 proc compareTypes*(x, y: PType,
                    cmp: TDistinctCompare = dcEq,
                    flags: TTypeCmpFlags = {}): bool =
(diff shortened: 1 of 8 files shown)


compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
...
+  ## Analyses and returns whether `typ` can possibly be the root of a reference
+  ## cycle, or in other words, whether a heap cell of type `typ` can keep
+  ## itself alive.
+  ##
+  ## The analysis is conservative: if a location of `typ` can have shared
+  ## ownership (``ref`` or closure) of a heap location with a dynamic type that
+  ## might not match the static one (i.e. a polymorphic ref or closure, the
+  ## latter uses type erasure), `typ` is treated as possibly cyclic.
+  let t = typ.skipTypes(abstractInst - {tyTypeDesc})
+  if not isFinal(t):
```
saem, line 1552: Won't `isFinal` skip the `tyTypeDesc` internally?

compiler/sem/semexprs.nim
```diff
@@ -474,6 +474,9 @@ proc isOpImpl(c: PContext, n: PNode, flags: TExprFlags): PNode =
       res = t.kind == tyProc and
             t.callConv == ccClosure and
             tfIterator in t.flags
+    of "cyclic":
```
saem: "cyclic" isn't really a type, but a fact about the type. Intuitionally `is` feels natural, but I'm thinking an alternative approach should surface in the future.

I'm ok with this for now but let's not document it in the manual (avoiding "official-ness") and mark it as "up for change" or something.

tests/typerel/tcyclic.nim
```diff
@@ -0,0 +1,100 @@
...
+  x: seq[NonCyclicObj]
+
+check NonCyclicObj isnot "cyclic" # no reference cycle is possible
+check (ref object) isnot "cyclic"
+
+type Cyclic = ref object
+  x: Cyclic
+
+check Cyclic           is    "cyclic"
+# the tuple is not part of the reference cycle, its element is:
```
saem:
```suggestion
# the tuple/seq/array are not part of the reference cycle, their element is:
```

compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
...
+
+proc trav(marker: var IntSet, g: ModuleGraph, t: PType, search: PType,
+          isInd: bool): bool =
+  let t = t.skipTypes(abstractInst-{tyTypeDesc})
+
+  case t.kind
+  of tyTuple, tySequence, tyArray:
+    # value types that can keep something alive
+    for i in 0..<t.len:
+      result = check(marker, g, t[i], search, isInd)
```
saem, line 1502: In case of a tuple we only care about the last field? Should this be successively `or`ed?

tests/typerel/tcyclic.nim
```diff
@@ -0,0 +1,100 @@
...
+check (Cyclic,)        isnot "cyclic"
+check seq[Cyclic]      isnot "cyclic"
+check array[1, Cyclic] isnot "cyclic"
+
+type Cyclic2 = object
+  x: ref (Cyclic2,)
+
+# it's contrived, but the single element tuple can be part of a reference cycle
```
saem: This is because `Cyclic2.x` can refer to a `(Cyclic2,)`?

It would also apply to seq and array assuming `Cyclic2`, or equivalent, had a field that could refer to either of those?

tests/typerel/tcyclic.nim
```diff
@@ -0,0 +1,100 @@
...
+check (Cyclic2,)       is "cyclic"
+check (ref (Cyclic2,)) is "cyclic"
+
```
saem, line 54:
```suggestion

type
  AliasedInt = int
  AliasedCyclic = Cyclic

# type aliases use the aliased type
check AliasedInt isnot "cyclic"
check AliasedCyclic is "cyclic"

```

for completeness?

compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
...
+    return false
+
+  # if the type is marked as acyclic, it is asserted to never being part of a
+  # reference cycle at run-time, meaning that we don't need to follow it further
+  let t = t.skipTypes(abstractInst - {tyTypeDesc})
+  if tfAcyclic in t.flags:
+    return false
+  # we're not interested in the difference between named and unnamed tuples nor
+  # distinct types, so use ``sameBackendType`` instead of ``sameType``
+  if isInd and sameBackendType(t, search):
```
saem: I think the result is currently equivalent, but the intention is different. Perhaps having a proc "sameLocationType" (can alias `sameBackendType`) to separate intention makes sense?

compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
...
+    # cycle to the type we're searching for is possible (or not possible) -- we
+    # have to assume that it's possible
+    result = t.callConv == ccClosure
+  else:
+    result = trav(marker, g, t, search, isInd)
+
+proc travNode(marker: var IntSet, g: ModuleGraph, n: PNode, search: PType,
+              isInd: bool): bool =
+  ## Traverses the fields of the given record `n` and checks if the type
+  ## identified by `startId` is reachable through one of them. Returns the
```
saem: Here and in `check`, `startId` doesn't exist, am I missing something? I see that it's referenced in `canFormAcycle` but that seems to be entirely separate.

compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
...
+  of tyOpenArray, tyVarargs:
+    # these are views; they don't own their items and thus can't keep them
+    # alive
+    result = false
+  of tyRef:
+    assert false
+  else:
+    discard "not relevant"
+
+proc isCyclePossible*(typ: PType, g: ModuleGraph): bool =
```
saem, line 1542: (minor, I tend to favour context params first)

I'd suggest making the `g`raph parameter first, so we get `g.isCyclePossible(t)` as opposed `t.isCyclePossible(g)`, they read differently:
- former: "hey graph, is a cycle possible for this type?"
- latter: "type is there a cycle possible, also here is a graph".

compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
...
+    result = travNode(marker, g, t.n, search, isInd)
+  of tyProc:
+    assert t.callConv != ccClosure
+    result = false
+  of tyOpenArray, tyVarargs:
+    # these are views; they don't own their items and thus can't keep them
+    # alive
+    result = false
+  of tyRef:
+    assert false
```
```suggestion
    unreachable()
```

compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
   ## is not marked as ``.pure`` (the ``sfPure`` flags is not present on it)
   result = (typ.kind == tyObject) and ((tfFinal in typ.flags) and
       (typ[0] == nil) or isPureObject(typ))
+
+proc trav(marker: var IntSet, g: ModuleGraph, t: PType, search: PType,
+          isInd: bool): bool
+
+proc check(marker: var IntSet, g: ModuleGraph, t: PType, search: PType,
```
saem, line 1427: Is `reachable` a better name?

Assuming that sounds good then perhaps these make sense:
- `trav` -> `productReachable`
- `travNode` -> `nodeReachable`

Maybe there is a flavour of naming with the shorter `reach`?

---

Tangentially, I searched for `type reachability` to see if it already meant something and google brought up this video: https://www.youtube.com/watch?v=T_mDt-qnmMk

compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
...
+  case t.kind
+  of tyRef:
+    let base = t.base.skipTypes(abstractInst - {tyTypeDesc})
+    if base.kind == tyObject and not isFinal(base):
+      # a polymorphic ``ref``. We have to assume that the dynamic type contains
+      # the `startId` type somewhere
+      result = tfAcyclic notin base.flags
+    else:
+      # a static ``ref``. Remember that we're analysing a type reachable
+      # through a ref indirection and check the element type
```
```suggestion
      # a static ``ref``. Check the elements' types as we're analysing for type
      # reachability through a ref indirection
```

Hopefully I preserved the meaning, but I believe the point is we have a static ref, so we need to check the elements in case anything can come back to `search`.

compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
...
+  of tyProc:
+    assert t.callConv != ccClosure
+    result = false
+  of tyOpenArray, tyVarargs:
+    # these are views; they don't own their items and thus can't keep them
+    # alive
+    result = false
+  of tyRef:
+    assert false
+  else:
```
saem: I'd recommend making these `case` checks exhaustive, withtout a catch all `else`, in order to avoid errors due to changes in `TTypeKind`.

compiler/ast/types.nim
```diff
@@ -1398,3 +1400,123 @@ proc isObjLackingTypeField*(typ: PType): bool {.inline.} =
...
+  ## Traverses the fields of the given record `n` and checks if the type
+  ## identified by `startId` is reachable through one of them. Returns the
+  ## result.
+  result = false
+  if n != nil:
+    case n.kind
+    of nkSym:
+      # cursor fields are non-owning -- they can't keep references alive and
+      # thus can't cause a reference cycle
+      if sfCursor notin n.sym.flags:
```
saem: Note for future discussion: cursor might be better as type information instead of symbol.

saem (commented):
Did one pass, no major blockers, did note that we should keep `is "cyclic"` quiet at first.

I didn't get a chance to stare at `TTypeKind` and come up with particularly interesting combinations; but the coverage in the tests is pretty good.

compiler/sem/semexprs.nim
beef331: Is there any reason this is not just implemented akin to `supportsCopyMem`? Seems like a `isCyclical` would suffice. This then allows:
```nim
type Cyclical = concept type C
  isCyclical(C)
```

compiler/ast/types.nim
zerbina: Yes, and it's going to be a problem should we make `typeDesc` a valid location type. Ideally, `tyTypeDesc` wouldn't be part of the `abstractInst` set (it is explicitly excluded in a lot of places).

compiler/sem/semexprs.nim
zerbina: I needed some way to test that the detection was working, and extending `is` was the most simple way of achieving it without feeling completely off.

Nonetheless, adding a type-trait query procedure is indeed the better and more correct thing to do.

compiler/ast/types.nim
zerbina: Yep, you're right. I accidentally removed the short-circuiting logic there.

tests/typerel/tcyclic.nim
zerbina: Yes, to both. The test is meant to make sure that structural types defined in-place are properly analysed (`canFormAcycle` doesn't correctly consider them).

compiler/ast/types.nim
zerbina: Yep, it's the better thing to do, thanks.

compiler/ast/types.nim
zerbina: `search` was originally named `startId`, and I forgot to update the documentation when changing it.

compiler/ast/types.nim
zerbina: I am/was a bit undecided in this case, as the `ModuleGraph` only provides additional information to the analysis, but is not central to it.

compiler/ast/types.nim
zerbina: Yes, I like it. Though I think `travNode` should also be named `productReachable` -- it's doing the same thing as `trav`, just with a record type's body instead of a type.

tests/typerel/tcyclic.nim
saem: OK, in which case, I have a suggestion.

```suggestion
# a container tuple/seq/etc with element of `ref` type matching the container
# can be part of a reference cycle
```

I'd consider changing the example to contain two elements and the first one forming a cycle to catch the short circuit issue above.

compiler/ast/types.nim
saem:
```suggestion
  ## identified by `search` is reachable through one of them. Returns the
```

compiler/ast/types.nim
saem: NP, it's not a huge preference.

compiler/ast/types.nim
saem: I used `break` instead of `return` to favour fall through/linear control flow.

```suggestion
      result = check(marker, g, t[i], search, isInd)
      if result:
        break
```

compiler/sem/semexprs.nim
saem: Thanks for the suggestion, @beef331.

compiler/ast/types.nim
saem: 🤔 OK, it's a "future" problem.

It's going to be an issue at `comptime` as that's exactly what's going to happen, when we want to do things like compute types. 😅

zerbina:
I applied the suggestions and added the `isCyclical` type-trait query procedure (replacing the `is "cyclic"` extension).

@saem If you think the PR is good-to-go, feel free to bors it already.

saem:
bors r+

This is super exciting!

PR #557, opened 2023-02-28
lexer: more specific diags, fewer strings

saem:
Introduced new diagnostircs for various specific circumstances. This reduces the need to allocate strings for various messages in the lexer. Pushed this through into legacy `reports` so most string creation happens in `cli_reporter` when required.

A few messages were changed, mostly punctuation or became slightly more specific.

---

## Notes for Reviewers
* nothing major, just saw things while examining the `lexer`

### DIFF
--- a/compiler/ast/report_enums.nim
+++ b/compiler/ast/report_enums.nim
@@ -188,10 +188,11 @@ type
     rlexMalformedUnderscores
     rlexMalformedTrailingUnderscre
     rlexInvalidToken
+    rlexInvalidTokenSpaceBetweenNumAndIdent
     rlexNoTabs
 
     # numbers
-    rlexInvalidIntegerPrefix
+    rlexInvalidIntegerLiteralOctalPrefix
     rlexInvalidIntegerSuffix
     rlexNumberNotInRange
     rlexExpectedHex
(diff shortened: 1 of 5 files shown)


zerbina (approved):
Looks good to me, thanks.

I left a small suggestion regarding the naming of a diagnostic, but it's nothing blocking.

compiler/ast/report_enums.nim
```diff
@@ -188,19 +188,24 @@ type
...
 
     # char
     rlexInvalidCharLiteral
+    rlexInvalidCharLiteralConstant
+    rlexInvalidCharLiteralPlatformNewline
+    rlexInvalidCharLiteralUnicodeCodepoint
     rlexMissingClosingApostrophe
-    rlexInvalidUnicodeCodepoint
+    rlexInvalidUnicodeCodepointEmpty
+    rlexInvalidUnicodeCodepointGreaterThan0x10FFFF
```
zerbina, line 208:
```suggestion
    rlexInvalidUnicodeNotACodepoint
```
I think the problem is more clear this way.

saem: I'll revise this separately, I had something like that before and it felt vague... the current one is possibly too in the weeds. Either way, not a blocker as you indicated and eaisly fixiable.

PR #558, opened 2023-02-28
compiler: replace `Rope`s with plain strings

zerbina:
## Summary

The usage of `Rope`s is blocking the switch to `--gc:orc` for the compiler, as very long chains that can happen in some cases (such as `const` arrays with a very large number of items) lead to stack overflows on `Rope` destruction.

In addition, replacing `Rope`s with strings leads to a decrease in peak memory consumption (when building the compiler itself) by 400 MiB -- execution time is not affected for a compiler built with `refc`.

## Details

- make `Rope` an alias for `string`
- remove the now-unused routines from `ropes.nim`
- replace `nil` with `"" where `Rope` where previously used

### Misc

- fix the "Unused import" warnings with regarding `debugutils`

---

## Notes for Reviewers
* a refactoring without any functional change

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/backend/cgen.nim
+++ b/compiler/backend/cgen.nim
@@ -48,8 +48,7 @@ import
     nversion,
     bitsets,
     ropes,
-    pathutils,
-    debugutils
+    pathutils
   ],
   compiler/sem/[
     passes,
@@ -80,10 +79,13 @@ from compiler/ast/report_enums import ReportKind
 #      destructor injections (or destructors, for that matter)
 from compiler/sem/injectdestructors import deferGlobalDestructor
 
-import std/strutils except `%` # collides with ropes.`%`
+import std/strutils except `%`, addf # collides with ropes.`%`
 
 from compiler/ic/ic import ModuleBackendFlag
 import dynlib
+
+when defined(nimCompilerStacktraceHints):
+  import compiler/utils/debugutils
 
 when not declared(dynlib.libCandidates):
   proc libCandidates(s: string, dest: var seq[string]) =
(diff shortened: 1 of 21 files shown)


compiler/backend/cgen.nim
```diff
@@ -80,11 +79,14 @@ from compiler/ast/report_enums import ReportKind
 #      destructor injections (or destructors, for that matter)
 from compiler/sem/injectdestructors import deferGlobalDestructor
 
-import std/strutils except `%` # collides with ropes.`%`
+import std/strutils except `%`, addf # collides with ropes.`%`
 
 from compiler/ic/ic import ModuleBackendFlag
 import dynlib
 
+when defined(nimCompilerStacktraceHints):
```
saem, line 87: thanks

saem:
@zerbina did you do a scan of the `when` statements, as well?

Not super concerned about `when false:` more the other ones that might not trigger locally or CI.

compiler/backend/cgen.nim
```diff
@@ -1798,7 +1796,7 @@ proc writeModule(m: BModule, pending: bool) =
   var cf = Cfile(nimname: m.module.name.s, cname: cfile,
                   obj: completeCfilePath(m.config, toObjFile(m.config, cfile)), flags: {})
   var code = genModule(m, cf)
-  if code != nil or m.config.symbolFiles != disabledSf:
+  if code != "" or m.config.symbolFiles != disabledSf:
```
saem, line 1799: `code` is a string, so the `$` conversion in the `when hasTinyCBackend` is unnecessary.

saem (approved):
Had a look through, I'm hoping this makes things run faster too! :D 

Also scanned the `when` clauses, looks good to merge! 🎉

compiler/backend/cgen.nim
saem: This is not a big deal, I'm totally OK ignoring this.

saem:
bors r+

(feel free to cancel it if you want to bring in changes)

PR #562, opened 2023-03-01
vmgen: support the return-value optimization for magics

zerbina:
## Summary
Allow *magic* procedures to modify the destination location directly if the result of the call is immediately assigned to something. The optimization is never actually made use of right now.

This is another step towards removing location setup (outside of `LdNull`) from the VM -- the only usage sites remaining are `opcConv` and the logic for closure calls.

## Details
- support in-place object construction without going through a temporary (optimization is not made use of right now)
- don't allocate unnecessary temporaries for `mNccInc`, `mNcsAdd`, and `mNcsIncl`
- `withTemp` creates full temporaries now

---

## Notes for Reviewers
Besides working towards removing location setup logic from the VM, this change is also meant to make it easier to separate the core VM from the macro evaluation support / compiler interfacing.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -324,19 +324,12 @@ func initLocReg*(r: var TFullReg, typ: PVmType, mm: var VmMemoryManager) =
 func initIntReg(r: var TFullReg, i: BiggestInt) =
   r = TFullReg(kind: rkInt, intVal: i)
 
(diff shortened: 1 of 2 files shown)


saem (approved):
bors r+

PR #564, opened 2023-03-02
vm: better `LocHandle` design

zerbina:
Use a significantly more efficient design around handles, made possible by improvements to the VM's allocator:
1) assign allocated cells a stable ID that doesn't change until they're freed
2) store the ID of the underlying cell in `LocHandle` 3) when creating a `LocHandle` from a random address, the address is mapped to a cell ID
4) seqs and strings only store a `CellPtr` for their data, reducing their in-memory from 24 to 16 byte

As a consequence:
- the access check logic knows directly in which cell the location is located (or if at all) -- mapping the pointer to a cell via a linear search is no longer required
- freeing a cell identified by a `LocHandle` has O(1) complexity instead of O(n), where `n` is the number of in-use cells. Right now, all cells are identified by `LocHandle`s

This results in a significant speed up for code running in the VM, especially if it uses a large number of long-living cells (both stack and heap). For example, with a `-d:release` compiler, the `tunidecode.nim` test now takes ~14 seconds to compile, compared to the ~40 seconds it took previously.

### General changes

- remove `MemRegionPtr` and associated routines. It's usages are replaced with either `VmMemoryRegion` or `VmMemPointer`. Since the former is an `openArray`, this also make it easier for the compiler to reason about the code
- use more consistent naming and terminology

### Changes to `vmmemory`

- deduplicate logic across the `allocX` procedures
- remove the unused `allocAtomLocation`
- disable stack-traces for the allocator related routines
- use an intrusive linked-list to efficiently remember free cells, replacing the usage of `seq.del` and making cell IDs stable

### Access checks

- remove the never taken `count == 0` branch
- remove all pointer-to-cell-id mapping logic from `vmchecks`

### Other

- introduce `VmSlice`. While also an important preparation for full `openArray` support, it's current use is to simplify some logic

---

## Notes for Reviewers
* this puts the VM on a good trajectory towards reaching the performance it had before the VM rework
* #548 will further speed up the VM, as with the PR, handles in the context of `var|lent T` stay as handles and are not unnecessarily turned into raw addresses.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/vm/vmchecks.nim
+++ b/compiler/vm/vmchecks.nim
@@ -1,10 +1,9 @@
 ## This module implements the handle validation logic.
 ##
-## Handles are validate by first making sure that they point to VM owned
-## memory. After finding the allocation the handle points into, the
-## allocation's layout is walked in order to verify that there exists a valid
-## location of the handle's type at the offset relative to the start of
-## the allocation.
+## After making making sure that the handle references a valid cell (it's an
+## error otherwise), the layout of cell is traversed in order to verify that
+## there exists a valid location of the handle's type at the offset relative to
+## the start of the cell.
 ##
 ## Special handling is required for variant objects, since whether a
 ## sub-location is valid depends on the run-time value of the respective
@@ -15,14 +14,13 @@ import
     idioms
   ],
   compiler/vm/[
-    vmdef,
-    vmmemory
+    vmdef
   ]
 
 from compiler/vm/vmobjects import variantFieldIndices
 from compiler/vm/vmtypes import alignedSize
 
-iterator fields(p: MemRegionPtr, t: PVmType): FieldIndex =
+iterator fields(p: VmMemoryRegion, t: PVmType): FieldIndex =
   ## Iterates and yields the indices of all active fields in the object at the
   ## given location
   if t.branches.len == 0:
@@ -33,10 +31,10 @@ iterator fields(p: MemRegionPtr, t: PVmType): FieldIndex =
       yield i
 
 
-func searchInObject(h: MemRegionPtr, otyp, t: PVmType, o: int): AccessViolationReason
+func searchInObject(h: VmMemoryRegion, otyp, t: PVmType, o: int): AccessViolationReason
 func testLocationType(locType: PVmType, t: PVmType): AccessViolationReason
 
-func searchInArray(h: MemRegionPtr, etyp: PVmType, t: PVmType, len, stride, o: int): AccessViolationReason =
+func searchInArray(h: VmMemoryRegion, etyp: PVmType, t: PVmType, len, stride, o: int): AccessViolationReason =
   let i = o /% stride
   assert i < len
 
@@ -47,12 +45,11 @@ func searchInArray(h: MemRegionPtr, etyp: PVmType, t: PVmType, len, stride, o: i
     of realAtomKinds:
       avrNoLocation
     of akObject:
-      searchInObject(h.getSubHandle(i*stride), etyp, t, o - (i*stride))
+      searchInObject(h.subView(i*stride), etyp, t, o - (i*stride))
     of akArray:
-      searchInArray(h.getSubHandle(i*stride), etyp.elementType, t, etyp.elementCount, etyp.elementStride, o - (i*stride))
+      searchInArray(h.subView(i*stride), etyp.elementType, t, etyp.elementCount, etyp.elementStride, o - (i*stride))
 
-
-func searchInObject(h: MemRegionPtr, otyp, t: PVmType, o: int): AccessViolationReason =
+func searchInObject(h: VmMemoryRegion, otyp, t: PVmType, o: int): AccessViolationReason =
   assert uint(o) < otyp.sizeInBytes
   assert otyp.kind == akObject
   for i in fields(h, otyp):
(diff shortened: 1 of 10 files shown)


compiler/vm/vmchecks.nim
```diff
@@ -1,10 +1,9 @@
...
 ##
-## Handles are validate by first making sure that they point to VM owned
-## memory. After finding the allocation the handle points into, the
-## allocation's layout is walked in order to verify that there exists a valid
-## location of the handle's type at the offset relative to the start of
-## the allocation.
+## After making making sure the that the handle references a valid cell (it's
+## an error otherwise), the layout of cell is traversed in order to verify that
+## there exists a valid location of the handle's type at the offset relative to
+## the start of the cell.
```
saem:
```suggestion
## After making making sure that the handle references a valid cell (it's an
## error otherwise), the layout of cell is traversed in order to verify that
## there exists a valid location of the handle's type at the offset relative to
## the start of the cell.
```

compiler/vm/vmdef.nim
```diff
@@ -269,22 +269,41 @@ type
...
+  CellPtr* = distinct ptr UncheckedArray[byte]
+    ## A pointer that either is nil or points to the start of a VM memory cell
+
   LocHandle* = object
-    ## A handle to a ``location``
-    h*: MemRegionPtr
-    typ* {.cursor.}: PVmType # ownership of `typ` is not needed
+    ## A handle to a VM memory location. Since looking up the cell the location
+    ## is part of is a very common operation (due to access checks), the cell's
+    ## ID is also stored here.
```
saem:
```suggestion
    ## A handle to a VM memory location. Since looking up the cell a location
    ## is part of is a very common operation (due to access checks), the cell's
    ## ID is also stored here.
```

compiler/vm/vmdef.nim
```diff
@@ -269,22 +269,41 @@ type
...
+    cell*: CellId
```
saem, line 284: At first I was a little worried this might go out of sync, but since these are stable id we're safe from that. 👍🏽

compiler/vm/vmdef.nim
```diff
@@ -269,22 +269,41 @@ type
...
+      ## the ID of the cell that owns the location
+    p*: VmMemPointer
+      ## the memory address of the reference location location, in host address
+      ## space
```
saem:
```suggestion
      ## the memory address of the referenced location, in host address space
```
If I understand things correctly that second location is redundant, because we're pointing directly into VM Managed Memory from the host.

compiler/vm/vmdef.nim
```diff
@@ -269,22 +269,41 @@ type
...
+    typ* {.cursor.}: PVmType
+      ## the type that the handles views the location as
+
+  VmSlice* = object
```
saem, line 291: I wouldn't change it now, and I'm not entirely sold on the name, but `VmMemHostSlice` might be indicative that this is host only.

compiler/vm/vmdef.nim
```diff
@@ -410,13 +429,44 @@ type
                                   ## symbol identifier
   VmCallback* = proc (args: VmArgs) {.closure.}
 
+  VmCell* = object
+    ## Stores information about a memory cell allocated by the VM's allocator
+    count*: int
```
saem, line 433: Not necessary now, just a thought; I'd favour hiding the field behind accessors.

```suggestion
    countOrNextPtr*: int
```

compiler/vm/vmdef.nim
```diff
@@ -410,13 +429,44 @@ type
...
   VmAllocator* = object
-    # XXX: the current implementation of the allocator is overly simple and
-    #      only temporary
     # XXX: support for memory regions going past the 0..2^63 (or 0..2^31)
     #      region is very fuzzy at the moment due to the mixing of int
     #      and uint
-    regions*: seq[tuple[typ: PVmType, count: int, start: int, len: int]]
+    cells*: seq[VmCell]
+      ## the underlying storage for the cell slots. Both stack and
+      ## (non-recounted) heap slots are included here
```
saem:
```suggestion
      ## (non-refcounted) heap slots are included here
```

saem (approved):
I didn't notice the notification earlier, would have reviewed it sooner.

I suggested a few minor things, no blockers, I'm really curious to try it out, especially with some of the dispatch streamlining I'm doing. 🎉

compiler/vm/vmdef.nim
zerbina: Your understanding is correct. The second "location" is a leftover from rewording :P .

compiler/vm/vmdef.nim
zerbina: I'm also not happy with `VmSlice` -- the name should be reserved for the soon-to-be in-memory representation for guest `openArray`s (similar to how `VmSeq` is used as the name for the guest `seq` representation). 

To me `VmMemHostSlice` sounds like it's a slice into host memory, which, while technically correct (the host and guest share the same address space), doesn't match the intention, i.e. that it's a guest memory slice.

compiler/vm/vmdef.nim
zerbina: Yep, it's indeed better. Since the accessors would have to be defined in `vmdef` for now (where they don't really belong), I left a to-do in the code for adding the accessor routines once `VmCell` is owned by the `vmmemory` module.

zerbina:
Thanks for taking a look, @saem.

bors r+

PR #565, opened 2023-03-03
sem: restructure call application handling

saem:
Summary
=======

- call operator now tries after failed overload resolution
- lower call operator `()` precedence

Major Fixes:
- overload resolution before call operator dispatch
- `()` overloads unable to receive untyped params
- `semIndirectOps` NPE (nil pointer exception)

Details
=======

Call Operator Precedence
------------------------

Lowered the `()` call operators precedence below the dot operators (`.`
and `.()`). Additionally, a number of associated changes resulted in
minor changes to error msg output. Along with additional tests, docs,
and the like. Many other fixes we

As of now the `()` will match last, and will also be a catchall for
unmatched overload resolutions -- previously a short coming.

This makes far more sense given the simple example `foo.bar()`. Here the
expectation is that `foo.bar` is:
1. a fully qualified identifier to a callable
2. a field access to a callable
3. method call syntax, requiring a dot transform
4. finally a fall back case via special operators

*Design Thinking*

The special operators are analoguous to Scala's `apply`/`unapply`. Long-
term this is the design direction, to enable fall backs for callables.

Presently, `()` receives `foo.bar`, implying that the expression
`foo.bar` should have already been evaluated. In which case, we should
have already attempted dot operator fallback. This is why the precedence
has been lowered. Alternatively, and not in keeping with the language,
the call operator should receive `bar`instead.

NB: The current behaviour is more correct, but unlikely "final".

Fixes
-----

*lower call operator precedence*
In some cases, where a call symbol was ambiguous, call operator dispath
would happen before overload resoltuion. This change now fixes that and
adds tests see `tests/specialops/tcallops`.

*untyped params for `()` overloads*
In addition, call operator (`()`) overloading couldn't receive untyped
arguments as they would be prematurely semantically analysed. This has
since been fixed, and the `tcallops` test been expanded to cover it.

*`semIndirectOps` NPE*
There was an NPE in `semDirectOps` logic, it's long standing, which is
indicative of this being a "pile of conditions" that sorta work, the
logic is entirely unsound.

Misc
====

Minor Improvements
------------------

- logic error `semOpAux`, some semantic errors in call operands would
  get overwrriten -- not sure how to reproduce this one
- `semExpr` now detects call errors sooner, slightly faster compiler
- clean-up misleading code in `maybeResemArgs` in `semcall`
- added additional call operator tests
- variety of docs for some of the procs, removal of dead code, and
  necessary stylistic clean-ups

"Fun Facts" Learned
-------------------

`nfDotField` and `nkSymChoices` are mutually exclusive for
`dotTransformations` (`nkDotExpr -> nkDotCall`)

Generic & template analysis both mangle `nkBranchExpr`. The current
implementation foolishly mangles `nkBracketExpr` to `nkCall` with a `[]`
callee symbol. This simply creates greater complexity within `semexprs`
module as it has to hackily reconstruct information and compensate for
this.

Closing Remark
--------------

As usual a big takeaway is that the existing implementation and
resulting language lack taste. By ~solving~ half-assing solutions in the
wrong places complexity and mistakes have spread throughout the code
base. This is just one small clean-up of many.

---

## Notes for Reviewers

* largely speaking, the old code, design, ideas, are all objectively bad
* changes to behaviour are pretty much guarantees
* with that in mind, the new code should be:
  * easier to reason about
  * preserve many/most things, at least in spirit

### Personal Note:

The existing code genuinely upsets me, this rework isn't a "fun" process, though
I'm looking forward to the outcome. I don't intend to mollify anyone upset about
some esoteric, half-baked, unsound, or objectively bad thing being changed or
removed.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -354,7 +354,6 @@ proc semConv(c: PContext, n: PNode): PNode =
       result = c.config.newError(n, PAstDiag(kind: adSemCannotBeConvertedTo,
                                              inputVal: op,
                                              targetTyp: result.typ))
-
   else:
     for it in op:
       let status = checkConvertible(c, result.typ, it)
@@ -366,7 +365,16 @@ proc semConv(c: PContext, n: PNode): PNode =
             handleError(c, result, hasError)
           else:
             it
-    errorUseQualifier(c, n.info, op[0].sym)
+    let s = errorUseQualifier(c, op, op[0].sym)
+    if s.isError:
+      result[1] = s.ast
+      result = c.config.wrapError(result)
+    else:
+      # `errorUseQualifier` won't return an error sym if it's unambiguous, but
+      # we already know that it can't be used for conversion via the loop above
+      result = c.config.newError(n, PAstDiag(kind: adSemCannotBeConvertedTo,
+                                             inputVal: op,
+                                             targetTyp: result.typ))
 
 proc semCast(c: PContext, n: PNode): PNode =
   ## Semantically analyze a casting ("cast[type](param)")
@@ -571,11 +579,12 @@ proc semOpAux(c: PContext, n: PNode): bool =
       a.typ = a[1].typ
     else:
       n[i] = semExprWithType(c, a, flags)
-      if n[1].isError:
+      if n[i].isError:
         result = true
 
 proc overloadedCallOpr(c: PContext, n: PNode): PNode =
-  # quick check if there is *any* () operator overloaded:
+  ## search for an overloaded call operator (`()`), if it exists create the
+  ## call tree, else return `nil`.
   var par = getIdent(c.cache, "()")
   var amb = false
   if searchInScopes(c, par, amb) == nil:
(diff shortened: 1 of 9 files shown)


compiler/sem/semexprs.nim
```diff
@@ -571,20 +579,22 @@ proc semOpAux(c: PContext, n: PNode): bool =
...
   if searchInScopes(c, par, amb) == nil:
     result = nil
   else:
     result = newNodeI(nkCall, n.info)
     result.add newIdentNode(par, n.info)
-    for i in 0..<n.len: result.add n[i]
-    result = semExpr(c, result)
+    for i in 0..<n.len:
+      result.add n[i]
+    # result = semExpr(c, result)
```
```suggestion
```

compiler/sem/semexprs.nim
```diff
@@ -1203,84 +1228,122 @@ proc semIndirectOp(c: PContext, n: PNode, flags: TExprFlags): PNode =
...
-        var hasErrorType = false
-        for i in 1..<n.len:
-          if n[i].typ.kind == tyError:
-            hasErrorType = true
-            break
+    elif n[0].kind == nkBracketExpr: # `someMacroOrTemplate[bar]()`
+      if n[0].len >= 1:
+        if n[0][0].kind == nkSym:
+          if n[0][0].sym.kind in {skMacro, skTemplate}:
+            setGenericParams(c, n[0])
```
zerbina, line 1243: It's a pre-existing thing, and no change is expected, but the `setGenericParams` name is really off -- it should be named something like `semGenericArgs`.

zerbina (approved):
I like the direction you're taking with the call operators, thank you for working on this!

I suggested one very small removal of commented code, but feel free to ignore it.

compiler/sem/semexprs.nim
saem: Thanks, done.

compiler/sem/semexprs.nim
saem: I didn't change it because it's mutating it in place rather than resulting in a new production. Not entirely sure that's the best call -- but it's an easy change later.

PR #566, opened 2023-03-03
vmgen/vm: conversion-related improvements

zerbina:
## Summary

Refactor the conversion-related parts in `vmgen`, which fixes multiple bugs and issues with code running in the VM:
- conversions where the source was of `sink` or `static` type always created a full copy
- lvalue-ness was not preserved for conversions to or from `distinct T` types was not always preserved
- assignments where the left-hand side is an lvalue conversion didn't update the destination location in some cases (such as for `ref`s or field assignments)

In addition, the VM access checks are now less strict: accessing a `ref|ptr T` location with a handle of `ref|ptr X` is allowed, as long as `X` is either a sub- or super-type of `T`.

## Details

Move the conversion logic for `ref`s and `objects` to a new instruction (`opcObjConv`). Compared to the previously used `opcConv`, using a dedicated conversion operator has the benefits that:
- only one type argument is required instead of two, meaning that the instruction can be encoded with two instead of three instruction words
- no full RTTI is required

For `opConv`:
- remove the `initLocReg` calls -- making sure that the destination location is set-up is now the responsibility of the code generator
- remove the catch-all branch

### Other

- fold `PointerLike(nil)` expressions into correctly typed `nkNilLit`s during `transf`

### Misc

- move the rendering of VM type conversion diagnostics to `cli_reporter` by introducing the dedicated `vmEvtIllegalConvFromXToY` event/diagnostic
- store the formal and actual type for VM type conversion reports directly, instead of as a `seq[TypeMismatch]`

---

## Notes for Reviewers
* this is a split-off from #548, with additional improvements and fixes 

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/ast/reports_vm.nim
+++ b/compiler/ast/reports_vm.nim
@@ -24,8 +24,8 @@ type
         stacktrace*: seq[tuple[sym: PSym, location: TLineInfo]]
         skipped*: int
 
-      of rvmCannotCast:
-        typeMismatch*: seq[SemTypeMismatch]
+      of rvmCannotCast, rvmIllegalConvFromXToY:
+        actualType*, formalType*: PType
 
       of rvmIndexError:
         indexSpec*: tuple[usedIdx, minIdx, maxIdx: Int128]
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -615,11 +615,11 @@ proc opConv(c: var TCtx; dest: var TFullReg, src: TFullReg, dt, st: (PType, PVmT
   # TODO: conversions are rather fuzzy right now. `opConv` needs to be revisited
   if desttyp.kind == tyString:
     # TODO: this part needs some refactoring...
-    initLocReg(dest, c.typeInfoCache.stringType, c.memory)
-
     let styp = srctyp.skipTypes(abstractRange)
     case styp.kind
     of tyEnum:
+      # TODO: don't use ``opcConv`` for string conversions -- make ``vmgen``
+      #       insert calls to the enum-to-string procedures instead
       let n = styp.n
       let x = src.intVal.int
       if x <% n.len and (let f = n[x].sym; f.position == x):
@@ -648,7 +648,7 @@ proc opConv(c: var TCtx; dest: var TFullReg, src: TFullReg, dt, st: (PType, PVmT
     of tyChar:
       dest.strVal = $chr(src.intVal)
     else:
-      internalError(c.config, "cannot convert to string " & desttyp.typeToString)
+      unreachable(styp.kind)
   else:
     let desttyp = skipTypes(desttyp, abstractVarRange)
     case desttyp.kind
@@ -661,7 +661,7 @@ proc opConv(c: var TCtx; dest: var TFullReg, src: TFullReg, dt, st: (PType, PVmT
         dest.intVal = src.intVal
       if toInt128(dest.intVal) < firstOrd(c.config, desttyp) or toInt128(dest.intVal) > lastOrd(c.config, desttyp):
         return true
-    of tyUInt..tyUInt64:
+    of tyUInt..tyUInt64, tyChar: # char is also an unsigned type
       dest = TFullReg(kind: rkInt)
       let styp = srctyp.skipTypes(abstractRange) # skip distinct types(dest type could do this too if needed)
       case styp.kind
@@ -697,34 +697,6 @@ proc opConv(c: var TCtx; dest: var TFullReg, src: TFullReg, dt, st: (PType, PVmT
         dest.floatVal = src.floatVal.float32.BiggestFloat
       else:
         dest.floatVal = src.floatVal
-    of tyObject:
-      c.config.internalAssert(srctyp.skipTypes(abstractVarRange).kind == tyObject, "invalid object-to-object conversion")
-      # XXX: fastAsgnComplex is wrong. A handle conversion should be performed
-      #      instead
-      fastAsgnComplex(dest, src)
-    of tyRef:
-      assert getTypeRel(st[1], dt[1]) != vtrUnrelated, "invalid ref-to-ref conversion" # vmgen issue
-      assert src.handle.typ.kind == akRef # vmgen issue
-
-      let refVal = deref(src.handle).refVal
-
-      if refVal.isNil:
-        discard "always allow conversion for nil ref"
-      else:
-        let loc = c.heap.tryDeref(refVal, noneType).value()
-        if getTypeRel(loc.typ, dt[1].targetType) notin {vtrSame, vtrSub}:
-          return true
-
-      block:
-        # src might be dependent on dest so copy first before cleaning up
-        var tmp: TFullReg
-        swap(tmp, dest)
-
-        initLocReg(dest, dt[1], c.memory)
-        asgnRef(deref(dest.handle).refVal, refVal, c.memory, reset=false)
-
-        cleanUpReg(tmp, c.memory)
-
     of tyOpenArray:
       assert dt[1].kind == akSeq
 
@@ -737,7 +709,6 @@ proc opConv(c: var TCtx; dest: var TFullReg, src: TFullReg, dt, st: (PType, PVmT
         #      it's okay for now.
         asgnValue(c.memory, dest, src)
       of tyArray, tyString:
-        dest.initLocReg(dt[1], c.memory)
         let
           t = dt[1].seqElemType
           L = arrayLen(src.handle)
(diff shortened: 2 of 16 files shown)


compiler/vm/vm.nim
```diff
@@ -2852,19 +2823,51 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): YieldReason
...
+      inc pc
+      let desttyp = c.types[c.code[pc].regBx - wordExcess]
+
+      assert src.typ.kind == desttyp.kind
+      # perform a conversion check, if necessary
+      case desttyp.kind
+      of akRef:
+        # a ref up- or down conversion
+        assert src.typ.kind == akRef # vmgen issue
+        checkHandle(regs[instr.regB])
```
saem, line 2851: Adding a `checkHandle` that takes the handle itself, and then this could become `checkHandle(src)` -- saves at least one compare.

this comment with the caveats:
- it shouldn't be in this PR
- I'm assuming the pattern is a common occurance

compiler/vm/vm.nim
```diff
@@ -2852,19 +2823,51 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): YieldReason
...
+
+        let refVal = deref(src).refVal
+
+        if not refVal.isNil:
```
saem: note to self: add `isNotNil` to `system`.

It's slightly less declarative but `refVal != nil` flows better when read, no change required here.

compiler/vm/vm.nim
```diff
@@ -2852,19 +2823,51 @@ proc rawExecute(c: var TCtx, pc: var int, tos: var StackFrameIndex): YieldReason
...
-          kind: vmEvtIllegalConv,
-          msg: errIllegalConvFromXtoY % [
-            typeToString(srctyp.nimType),
-            typeToString(desttyp.nimType)
-          ]))
+          kind: vmEvtIllegalConvFromXToY,
+          typeMismatch: VmTypeMismatch(
+            actualType: srctyp.nimType,
+            formalType: desttyp.nimType)))
+    of opcObjConv:
```
saem, line 2838: This and the `transf` logic (two major pieces I've read so far), are much improved! It's like a logical narrative that one can easily follow and reason about. 🎉

tests/lang_exprs/tlvalue_conv_assignment.nim
```diff
@@ -0,0 +1,98 @@
...
+      C(B(x)) = nil
+      doAssert x == nil
+
+    block:
+      # assignment to down-converted field:
+      var x = Obj[C](field: def)
+      B(x.field) = nil
+      doAssert x.field == nil
+
+      # assignemnet to down-converted field (nested):
```
saem:
```suggestion
      # assignment to down-converted field (nested):
```

tests/lang_exprs/tlvalue_conv_assignment.nim
```diff
@@ -0,0 +1,98 @@
...
+      A(B(x.field)) = nil
+      doAssert x.field == nil
+
+    block:
+      # assignment to up-converted field:
+      var x = Obj[C](field: def)
+      B(x.field) = nil
+      doAssert x.field == nil
+
+      # assignemnet to up-converted field (nested):
```
saem:
```suggestion
      # assignment to up-converted field (nested):
```

tests/lang_exprs/tlvalue_conv_assignment.nim
```diff
@@ -0,0 +1,98 @@
...
+  Base = object of RootObj
+  A = object of Base
+  B = object of A
+
+  Obj[T] = object
+    field: T
+
+block ref_and_ptr_conversion:
+  proc test[A, B, C](def: C) =
+    block down_conv:
```
saem:
```suggestion
    block down_conv_local:
```

tests/lang_exprs/tlvalue_conv_assignment.nim
```diff
@@ -0,0 +1,98 @@
...
+      B(x) = nil
+      doAssert x == nil
+
+      x = def
+
+      # assignment to down-converted local (nested):
+      A(B(x)) = nil
+      doAssert x == nil
+
+    block up_conv:
```
saem:
```suggestion
    block up_conv_local:
```

tests/lang_exprs/tlvalue_conv_assignment.nim
```diff
@@ -0,0 +1,98 @@
...
+      B(x) = nil
+      doAssert x == nil
+
+      x = A(def)
+
+      # assignment to up-converted local (nested):
+      C(B(x)) = nil
+      doAssert x == nil
+
+    block:
```
saem:
```suggestion
    block down_conv_field:
```

tests/lang_exprs/tlvalue_conv_assignment.nim
```diff
@@ -0,0 +1,98 @@
...
+      var x = Obj[C](field: def)
+      B(x.field) = nil
+      doAssert x.field == nil
+
+      # assignemnet to down-converted field (nested):
+      x = Obj[C](field: def)
+      A(B(x.field)) = nil
+      doAssert x.field == nil
+
+    block:
```
saem:
```suggestion
    block up_conv_field:
```

compiler/ast/reports_vm.nim
```diff
@@ -24,8 +24,8 @@ type
         stacktrace*: seq[tuple[sym: PSym, location: TLineInfo]]
         skipped*: int
 
-      of rvmCannotCast:
-        typeMismatch*: seq[SemTypeMismatch]
+      of rvmCannotCast, rvmIllegalConvFromXToY:
```
saem, line 27: Less is more, I like it.

compiler/vm/vmgen.nim
```diff
@@ -79,6 +79,10 @@ type
   VmGenError = object of CatchableError
     diag: VmGenDiag
 
+const SkipSet = abstractInst + {tyStatic} - {tyTypeDesc}
+  ## the set of types that are not relevant the VM. ``tyTypeDesc``, while not
+  ## relevant right now, is likely going to be in the future.
```
```suggestion
  ## the set of types that are not relevant to the VM. ``tyTypeDesc``, while
  ## not relevant right now, is likely going to be in the future.
```

compiler/vm/vmgen.nim
```diff
@@ -79,6 +79,10 @@ type
   VmGenError = object of CatchableError
     diag: VmGenDiag
 
+const SkipSet = abstractInst + {tyStatic} - {tyTypeDesc}
```
```suggestion
const IrrelevantTypes = abstractInst + {tyStatic} - {tyTypeDesc}
```

Mild suggestion based on reading the excellent doc comment. :D

compiler/vm/vmgen.nim
```diff
@@ -1201,25 +1205,50 @@ proc genAddSubInt(c: var TCtx; n: PNode; dest: var TDest; opc: TOpcode) =
...
-    if n.typ.kind == arg.typ.kind and arg.typ.kind == tyProc:
-      # don't do anything for lambda lifting conversions:
-      return true
+  let
+    a = skipTypes(n.typ, SkipSet)
+    b = skipTypes(arg.typ, SkipSet)
 
-  if implicitConv():
+  if sameLocationType(a, b) or
+     (a.kind == b.kind and a.kind == tyProc):
```
```suggestion
     {a.kind, b.kind} == {tyProc}:
```

Not sure if that's necessarily better, but you also get to make the `if` condition a one liner with it. 😆

saem (approved):
Looks really good. I have a few suggestions, nothing major IIRC.

compiler/vm/vm.nim
zerbina: A `checkHandle` that takes the handle itself already exists, but I didn't use it for efficiency reasons. The overload taking a register doesn't perform access checks for owning handles (i.e. `rkLocation` -- they're always valid), while the one taking a handle always does.

Further down the road, both the `rkLocation` and all the `checkHandle` calls will go away (the latter will be replaced with a dedicated operation, so that the code generator can decide on the placement).

compiler/vm/vm.nim
zerbina: It was a quick change, so I added an `isNotNil` routine for `HeapSlotHandle` and used it here.

compiler/vm/vmgen.nim
zerbina: It's much more descriptive, thanks.

compiler/vm/vmgen.nim
zerbina: I changed it -- having the condition on one line is indeed very nice :D

compiler/vm/vm.nim
saem: Sorry, I should have read the implementation a bit more closely, they looked equivalent except for the handle fetch. Thanks for the explanation.

saem:
Bors r+

Awesome.

PR #570, opened 2023-03-07
vm: make `opcExpandToAst` more flexible

zerbina:
## Summary

The `ExpandToAst` operation was previously responsible for turning each arguments' value into `PNode` AST in order for them to be passed to `evalTemplate`. For providing the `PType` of each argument, the AST of the original template call expression was passed as the first operand.

Since this approach would become a problem once `vmgen` no longer operates on `PNode` AST, a different one is required: the deserialization logic is moved to a dedicated operation (`opcDataToAst`) taking the value and the destination type as input, with `vmgen` emitting the instruction for arguments to template calls.

## Details

- add the `opcDataToAst` operation
- clean up the `opcExpandToAst` implementation
- adjust the logic for `mExpandToAst` and clean up / improve the code

### Misc

- fix `regToNode` not initializing the resulting node's line information

---

## Notes for Reviewers
* split off from #551

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -471,14 +471,14 @@ proc regToNode*(c: TCtx, x: TFullReg; typ: PType, info: TLineInfo): PNode =
   case x.kind
   of rkNone: unreachable()
   of rkInt:
-    result = newNodeIT(nkIntLit, TLineInfo(), typ)
+    result = newNodeIT(nkIntLit, info, typ)
     result.intVal = x.intVal
   of rkFloat:
-    result = newNodeIT(nkFloatLit, TLineInfo(), typ)
+    result = newNodeIT(nkFloatLit, info, typ)
     result.floatVal = x.floatVal
   of rkAddress:
     if x.addrVal.isNil:
-      result = newNode(nkNilLit)
+      result = newNodeI(nkNilLit, info)
       result.typ = typ
     else:
       # TODO: validate the address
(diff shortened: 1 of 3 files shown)


saem (commented):
I really like the change overall.

I do think the lineinfo results in errors that are real headscratchers. I'd keep that as is -- although I don't like `regToNode` modifying the info either.

----

More bits about lineinfo:

The "real fix" doesn't seem entirely obvious. From `opcDataToAst`'s perspective, it's simply generating AST and that's all that it should care about. From `vmgen`'s perspective it got told to do so. I think `semexprs` is where we have enough context to know what the lineinfo should be. Which in this case we're expanding the AST at a known source location and should set it accordingly. `semexprs` needs to leave info behind for `vmgen` to do the right thing wrt info. Alternatively, I can also see the argument being made that `vmgen` knows enough as this is behaviour that's intrinsic to the expansion operation, as such it just needs to direct the VM appropriately.

Regardless of what settles out, I do agree that `regToNode` doing this does not seem correct, perhaps it's quite a bit faster to do it this way.

saem:
Bors r+

Awesome, thanks for the day update.

saem:
Wat?

zerbina:
I've reverted the removal of the line information override in `opcExpandToAst`, but left a note in the code regarding the issue. In practice, this means that the following behaviour is restored:

```nim
macro m(x: untyped): untyped =
  template identity(i: untyped): untyped =
    i

  # at this, `x` has its original line information
  result = getAst(identity(x))
  # past this point, the line information of `x` points to the `getAst` call

discard m(1)
```

It's nothing too critical and can indeed be fixed separately.

PR #571, opened 2023-03-07
Remove the `.deprecated` redirection feature

zerbina:
## Summary

The `{.deprecated: [old, new].}` feature, originally introduced for providing information to `nimfix`, can be implemented as a library solution and doesn't need compiler support.

Since `skAlias` was only used for these redirections, the symbol kind is also removed, which simplifies lookup a bit, as alias symbols no longer need to be considered there.

## Details

The only user of the redirection feature was the `mysql` module; all usages there are now removed, meaning that code still using the deprecated enum names needs to be adjusted.

- remove the now unused `adSemBadDeprecatedArgs` and `rsemBadDeprecatedArgs` diagnostics
- update the `rsemBadDeprecatedArg` message

---

## Notes for Reviewers
* the removal was discussed [here](https://matrix.to/#/!MAXxvfCYnfYgZqSizw:matrix.org/$eig05VobpBjyZUrpNO5PN1N5a8Bs7hv88C1hxLV_clg?via=libera.chat&via=matrix.org&via=envs.net)

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -700,7 +700,6 @@ type
                           ## file (it is loaded on demand, which may
                           ## mean: never)
     skPackage             ## symbol is a package (used for canonicalization)
-    skAlias               ## an alias (needs to be resolved immediately)
 
   TSymKinds* = set[TSymKind]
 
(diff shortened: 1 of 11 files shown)


saem (approved):
Awesome, thanks

You'll see it if you check the CI output. `tdeprecated` is grumpy because line 5 in this test needs to be removed (now removed deprecation message) and add a new line after the testament spec to keep the line numbers in sync. Should be all clear afterwards.

saem:
bors r+

(previous run failed due to a homebrew timeout)

PR #573, opened 2023-03-08
pragmas: rework processing and application

zerbina:
## Summary

Processing and application of pragmas during the first semantic
analysis pass was all done by a single routine used for all three
contexts (declarations, statements, blocks), leading to complex
code with lots of conditional logic.

The rework is meant to make future changes to pragmas easier and the
code better to understand, while also adjusting or streamlining the
semantics of some pragmas.

### General changes

- implicit pragmas (i.e. those passed to the `.push` pragma) no longer
  affect pragma statements and pragma blocks
- pragma lists in on identdefs that become empty after macro/template
  pragma application are kept, instead of being replaced with an
  `nkEmpty` node
- allow custom pragmas on iterators (both `.closure` and `.inline`)

### Pragma changes

- the `.linksys` (unused) and `.unchecked` (redundant) pragmas are
  removed
- the `.deprecated` statement's message can now be supplied by a
  constant expression
- `.cast` pragmas cannot be `push`ed anymore
- note-setting pragmas (e.g. `.hint[X]: on`) are now only allowed in
  pragma statements
- `.pure` is now disallowed in pragma statements. It could previously
  be used to mark the enclosing entity (routine or module) as `pure`
- a pushed `.dynlib` now only applies to routines (except templates)
- allow `.stackTrace` and `.lineTrace` on iterators

## Details

For pragmas on declarations and for blocks, analysis and application
are separated into two sequential steps, where first all pragmas are
analysed (`.pragma` pragmas expanded, custom pragmas resolved, early
rejection of unapplicable pragmas) and then all non-erroneous ones
applied.

Since a single pragma in statement can alter note state (which could
affect analysis of the pragmas following it), analysis and application
for statement pragmas happens in an intertwined manner.

Handling of the three special pragmas `.pragma`, `.push`, and `.pop`
no longer relies on a node flag (`nfImplicitPragma`, which is now
removed), but instead happens separately from the other pragmas.

In order to remove the extra arguments previously passed to the
`pragma` procedure, each context where pragmas are applied now uses its
own dedicated procedure.

In general, `nkError` node propagation is improved and the amount of
input AST mutations reduced.

---

## To-do
- [x] apply implicit pragmas to all declarations again
- [x] wrap errors occurring during application of pushed pragmas (i.e. implicit pragmas) in `adSemImplicitPragmaError`s
- [x] use more consistent naming
- [x] write a commit message

## Notes for reviewers
- the goal is to make further changes/improvements to pragma easier and the code easier to navigate

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -591,13 +591,6 @@ type
     nfDefaultRefsParam ## a default param value references another parameter
                        ## the flag is applied to proc default values and to calls
     nfHasComment ## node has a comment
-    nfImplicitPragma ## node is a "singlePragma" this is a transition flag
-                  ## created as part of nkError refactoring for the pragmas
-                  ## module. an old proc, `singlePragma` did a lot of side-
-                  ## effects and returned a bool signal to callers typically to
-                  ## either break a loop and raise an error in
-                  ## `pragmas.implicitPragmas` or simply break a loop in
-                  ## `pragmas.pragmaRec`.
 
   TNodeFlags* = set[TNodeFlag]
   TTypeFlag* = enum   ## keep below 32 for efficiency reasons (now: 43)
@@ -1073,7 +1066,6 @@ type
     adSemLocksPragmaBadLevelString
     adSemBorrowPragmaNonDot
     adSemInvalidExtern
-    adSemBadDeprecatedArg
     adSemMisplacedEffectsOf
     adSemMissingPragmaArg
     adSemCannotPushCast
@@ -1255,7 +1247,6 @@ type
         adSemLocksPragmaBadLevelRange,
         adSemLocksPragmaBadLevelString,
         adSemBorrowPragmaNonDot,
-        adSemBadDeprecatedArg,
         adSemMisplacedEffectsOf,
         adSemMissingPragmaArg,
         adSemCannotPushCast,
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -933,9 +933,6 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemInvalidExtern:
       result = "invalid extern name: '" & r.externName & "'. (Forgot to escape '$'?)"
 
-    of rsemBadDeprecatedArg:
-      result = "string literal expected"
-
     of rsemInvalidPragma:
       result = "invalid pragma: " & r.ast.render
 
@@ -1640,7 +1637,7 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       result = r.str
 
     of rsemImplicitPragmaError:
-      result = "???"
+      result = "application of implicit pragma failed"
 
     of rsemInvalidModulePath:
       result = "invalid path: " & r.str
@@ -3821,7 +3818,6 @@ func astDiagToLegacyReport(conf: ConfigRef, diag: PAstDiag): Report {.inline.} =
       adSemLinePragmaExpectsTuple,
       adSemLocksPragmaExpectsList,
       adSemBorrowPragmaNonDot,
-      adSemBadDeprecatedArg,
       adSemMisplacedEffectsOf,
       adSemMissingPragmaArg,
       adSemCannotPushCast,
--- a/compiler/sem/pragmas.nim
+++ b/compiler/sem/pragmas.nim
@@ -34,9 +34,9 @@ import
     options
   ],
   compiler/utils/[
-    ropes,
     pathutils,
-    debugutils
+    debugutils,
+    idioms
   ],
   compiler/sem/[
     semdata,
@@ -87,6 +87,7 @@ const
   iteratorPragmas* = declPragmas + {FirstCallConv..LastCallConv, wNoSideEffect, wSideEffect,
     wMagic, wBorrow,
     wDiscardable, wGensym, wInject, wRaises, wEffectsOf,
+    wStackTrace, wLineTrace,
     wTags, wLocks, wGcSafe}
   exprPragmas* = {wLine, wLocks, wNoRewrite, wGcSafe, wNoSideEffect}
   stmtPragmas* = {wChecks, wObjChecks, wFieldChecks, wRangeChecks,
@@ -94,13 +95,14 @@ const
     wStyleChecks, wAssertions,
     wWarnings, wHints,
     wLineDir, wStackTrace, wLineTrace, wOptimization, wHint, wWarning, wError,
-    wFatal, wDefine, wUndef, wCompile, wLink, wLinksys, wPure, wPush, wPop,
+    wFatal, wDefine, wUndef, wCompile, wLink, wPush, wPop,
     wPassl, wPassc, wLocalPassc,
     wDeadCodeElimUnused,  # deprecated, always on
+    wDebugger, wProfiler,
     wDeprecated,
     wFloatChecks, wInfChecks, wNanChecks, wPragma, wEmit,
     wLinearScanEnd, wTrMacros, wEffects, wComputedGoto,
-    wExperimental, wUsed, wAssert}
+    wExperimental, wUsed, wByRef, wCallconv}
   lambdaPragmas* = {FirstCallConv..LastCallConv,
     wNoSideEffect, wSideEffect, wNoreturn, wNosinks, wDynlib, wHeader,
     wThread, wAsmNoStackFrame,
@@ -109,7 +111,7 @@ const
   typePragmas* = declPragmas + {wMagic, wAcyclic,
     wPure, wHeader, wCompilerProc, wCore, wFinal, wSize, wShallow,
     wIncompleteStruct, wCompleteStruct, wByCopy, wByRef,
-    wInheritable, wGensym, wInject, wRequiresInit, wUnchecked, wUnion, wPacked,
+    wInheritable, wGensym, wInject, wRequiresInit, wUnion, wPacked,
     wBorrow, wGcSafe, wExplain, wPackage}
   fieldPragmas* = declPragmas + {wGuard, wBitsize, wCursor,
     wRequiresInit, wNoalias, wAlign} - {wExportNims, wNodecl} # why exclude these?
@@ -129,6 +131,10 @@ const
   allRoutinePragmas* = methodPragmas + iteratorPragmas + lambdaPragmas
   enumFieldPragmas* = {wDeprecated}
 
+  allowsCustomPragma = {skVar, skLet, skParam, skField, skProc, skFunc,
+                        skConverter, skMethod, skIterator, skType}
+    ## the set of symbol kinds that allow attached custom pragmas
+
 proc getPragmaVal*(procAst: PNode; name: TSpecialWord): PNode =
   let p = procAst[pragmasPos]
   if p.kind == nkEmpty: return nil
@@ -136,9 +142,6 @@ proc getPragmaVal*(procAst: PNode; name: TSpecialWord): PNode =
     if it.kind in nkPragmaCallKinds and it.len == 2 and it[0].kind == nkIdent and
         it[0].ident.id == ord(name):
       return it[1]
-
-proc pragma*(c: PContext, sym: PSym, n: PNode, validPragmas: TSpecialWords;
-            isStatement: bool = false): PNode {.discardable.}
 
 proc recordPragma(c: PContext; n: PNode; args: varargs[string]) =
   var recorded = newNodeI(nkReplayAction, n.info)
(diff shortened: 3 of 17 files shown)


compiler/sem/pragmas.nim
```diff
@@ -129,6 +129,10 @@ const
   allRoutinePragmas* = methodPragmas + iteratorPragmas + lambdaPragmas
   enumFieldPragmas* = {wDeprecated}
 
+  allowsCustomPragma = {skVar, skLet, skParam, skField, skProc, skFunc,
```
saem, line 134: shouldn't this include iterator?

zerbina: I don't see a reason why not. The set was just copied from the previous the code, without me adjusting it.

compiler/sem/pragmas.nim
```diff
@@ -1563,90 +1377,56 @@ proc prepareSinglePragma(
...
+        assert sym.kind in {skVar, skLet}
+        sym.flags.incl sfGoto
       of wExportNims:
-        result = it
-        if sym == nil:
-          result = invalidPragma(c, it)
-        else:
+          # XXX: modifying the module graph during application of a symbol
+          #      operator doesn't seem like a good idea...
           result = magicsys.registerNimScriptSymbol2(c.graph, sym)
```
saem, line 1397: This is clean-up I forgot, we can drop the old `registerNimScriptSymbol` and take over the name. Doesn't have to be tackled in this PR, I can do it after.

compiler/sem/pragmas.nim
```diff
@@ -1563,90 +1377,56 @@ proc prepareSinglePragma(
...
-          sym.flags.incl sfGoto
+        result = noVal(c, it)
+        assert sym.kind in {skVar, skLet}
+        sym.flags.incl sfGoto
       of wExportNims:
-        result = it
-        if sym == nil:
-          result = invalidPragma(c, it)
-        else:
+          # XXX: modifying the module graph during application of a symbol
```
saem, line 1395: At first I thought it was in effect export it, but nope, it's adding it to `exposed`, which I believe amounts to export it at the codegen level, but not the semantic level. I can't seem to find any usage of `exportnims` either -- odd.

compiler/sem/semtypes.nim
```diff
@@ -56,7 +56,7 @@ proc semEnum(c: PContext, n: PNode, prev: PType): PType =
       if n[i][0].kind == nkPragmaExpr:
         e = newSymS(skEnumField, n[i][0][0], c)
         identToReplace = addr n[i][0][0]
-        n[i][0][1] = pragma(c, e, n[i][0][1], enumFieldPragmas)
+        n[i][0][1] = pragmaDecl(c, e, n[i][0][1], enumFieldPragmas)
         # check if we got any errors and if so report them
         for e in ifErrorWalkErrors(c.config, n[i][0][1]):
```
saem, line 61: I reallly like the direction of the PR. I think once these changes are in, I can swap out these walk errors with propagating the errors instead. That should clean-up a bunch of downstream code. 🎉

compiler/sem/pragmas.nim
zerbina: I believe it's meant for registering a procedure to later be able to look it up by name, when embedding NimScript.

saem: Yup, at the time I made the comment I did some searches trying to find the mechanism where things were actually stitched together. It was really difficult to find and couldn't get a complete picture. With that said, if that specific bit is revisited, that whole mechanism likely needs to be revised and perhaps this pragma will be shown to be unnecessary.

Not expecting any additional action.

compiler/sem/pragmas.nim
zerbina: Custom pragmas are now also allowed on iterators.

compiler/ast/ast_types.nim
```diff
@@ -591,13 +591,6 @@ type
     nfDefaultRefsParam ## a default param value references another parameter
                        ## the flag is applied to proc default values and to calls
     nfHasComment ## node has a comment
-    nfImplicitPragma ## node is a "singlePragma" this is a transition flag
```
saem, line 594: YES!

compiler/front/cli_reporter.nim
```diff
@@ -1640,7 +1637,7 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       result = r.str
 
     of rsemImplicitPragmaError:
-      result = "???"
+      result = "application of implicit pragma failed"
```
saem, line 1640: nice catch!

compiler/sem/pragmas.nim
```diff
@@ -572,37 +574,42 @@ proc processOption(c: PContext, n: PNode, resOptions: var TOptions): PNode =
...
 
-proc processPush(c: PContext, n: PNode, start: int): PNode =
-  ## produces (mutates) `n`, or an error, `start` indicates which of the
-  ## current pushed options (child of n) are being produced. will wrap the
-  ## child and `n` each in errors.
-  result = n
-  if n[start-1].kind in nkPragmaCallKinds:
-    result = c.config.newError(n, PAstDiag(kind: adSemUnexpectedPushArgument))
-    return
+proc processPush(c: PContext, info: TLineInfo, pragmas: openArray[PNode], r: PNode): bool =
```
saem, line 579: Might be too big a change, but return `true` on success "feels" better.

compiler/sem/pragmas.nim
```diff
@@ -572,37 +574,42 @@ proc processOption(c: PContext, n: PNode, resOptions: var TOptions): PNode =
...
+          x.otherPragmas = newNodeI(nkPragma, info)
+        x.otherPragmas.add it
+
     else:
-      n[i] = err
-      result = wrapError(c.config, n)
-      assert not cyclicTree(result)
+      r.add err
+      result = true
       return
```
saem: This is likely a bug on my part, should probably continue even if there are errors (for check, suggest, etc). Don't have to fix now.

compiler/sem/pragmas.nim
```diff
@@ -835,18 +842,20 @@ proc pragmaLine(c: PContext, n: PNode): PNode =
...
+  ## expands to on use, and returns the first item from `pragmas` (that is, the
+  ## name part). On failure, an error is returned.
+  let it = pragmas[0]
   result = it
   if it.kind notin nkPragmaCallKinds and it.safeLen == 2 or
      it.safeLen != 2 or it[0].kind != nkIdent or it[1].kind != nkIdent:
-    n[i] = invalidPragma(c, it)
-    result = n[i]
-    return
+    return invalidPragma(c, it)
```
saem, line 858: This was early `nkError` conversion, I could have done this better; this doesn't need to be changed now.

What should happen is the error should encompass `it` and the rest of the nodes -- not just `it`. Doing it that way means we can recover/restore from `nkError` by grabbing `wrongNode` from the diag.

compiler/sem/pragmas.nim
```diff
@@ -1044,179 +1026,94 @@ proc processEffectsOf(c: PContext, n: PNode; owner: PSym): PNode =
...
           else:
-            invalidPragma(c, it)
+            setExternName(c, sym, name.strVal)
+            it
+      of wDirty:
+        result = noVal(c, it)
+        assert sym.kind == skTemplate
+        incl(sym.flags, sfDirty)
       of wImportJs:
+        # note: don't use ``semExternName`` here. The operand is *not* an
```
saem, line 1105: Hmm, at a later point we likely need separate `importjs` pragmas, because we have the whole calling/emit pattern language hidding in here.

compiler/sem/pragmas.nim
```diff
@@ -1269,10 +1162,11 @@ proc prepareSinglePragma(
         incl(sym.flags, sfNoalias)
       of wEffectsOf:
         result = processEffectsOf(c, it, sym)
+        if result.isNil: # was there no error?
```
saem, line 1169: `processEffectsOf` and `processParam` therein are messing this up by silently dropping `semExpr` errors, and they really should create a full production. That doesn't have to be done now; pre-existing issues.

compiler/sem/pragmas.nim
```diff
@@ -1361,29 +1249,27 @@ proc prepareSinglePragma(
           incl(sym.flags, sfBorrow)
       of wFinal:
         result = noVal(c, it)
-        if sym.typ == nil:
-          result = invalidPragma(c, it)
-        else: incl(sym.typ.flags, tfFinal)
+        assert sym.typ != nil
```
saem, line 1256: (pre-existing issue) This should do a reciprocal check of `wInheritable` in case one pragma is applied before the other.

compiler/sem/pragmas.nim
```diff
@@ -1438,119 +1299,53 @@ proc prepareSinglePragma(
...
-            it
-          else:
-            err
-      of wPush:
-        result = processPush(c, n, i + 1)
-        result.flags.incl nfImplicitPragma # xxx: legacy singlepragma=true
-      of wPop:
-        result = processPop(c, it)
-        result.flags.incl nfImplicitPragma # xxx: legacy singlepragma=true
+          # XXX: disallow for the other declarations (e.g. ``var``)?
```
saem: yeah, that's weird, because `wError` is in `declPragmas` whcih is under `varPragmas`. 🤔 I bet the rest of the compiler can't handle it.

compiler/sem/pragmas.nim
```diff
@@ -1438,119 +1299,53 @@ proc prepareSinglePragma(
...
-        c.config.options = tmp
       of wStackTrace, wLineTrace:
-        if sym.kind in {skProc, skMethod, skConverter}:
-          result = processOption(c, it, sym.options)
-        else:
-          var tmp = c.config.options
-          result = processOption(c, it, tmp)
-          c.config.options = tmp
+        # TODO: also support for funcs and (closure) iterators -- there's no
+        #       reason not to
```
saem: Strongly agree.

compiler/sem/pragmas.nim
```diff
@@ -1438,119 +1299,53 @@ proc prepareSinglePragma(
...
       of wCompleteStruct:
         result = noVal(c, it)
-        if sym.typ == nil:
-          result = invalidPragma(c, it)
-        else:
-          incl(sym.typ.flags, tfCompleteStruct)
+        assert sym.typ != nil
+        incl(sym.typ.flags, tfCompleteStruct)
       of wUnchecked:
+        # XXX: seems redundant with ``tyUncheckedArray``. Remove it?
```
saem: Yup, it was supposedly removed in Nim 0.20.0 (per the change log).

saem (commented):
I'm part way done, going to `processBlockPragma` next.

compiler/sem/pragmas.nim
```diff
@@ -1657,65 +1418,324 @@ proc prepareSinglePragma(
...
+proc applyStmtPragma(c: PContext, owner: PSym, it: PNode, k: TSpecialWord): PNode =
+  ## Evaluates the single built-in pragma statement `it` for its side-effects.
+  ## Depending on the pragma, this might mutate `owner`. If the statement
+  ## is ill-formed (e.g. missing arguments) or the evaluation failed, an error
+  ## is returned.
+  case k
+  of wUsed:
+    result = noVal(c, it)
+    # XXX: an escaping mutation if the the pragma is used inside a ``compiles``
+    #      block
```
saem, line 1465:
```suggestion
    # XXX: an escaping mutation if the pragma is used inside a ``compiles``
    #      block
```

compiler/sem/pragmas.nim
```diff
@@ -1657,65 +1418,324 @@ proc prepareSinglePragma(
...
+    owner.flags.incl sfUsed
+  of wDeprecated:
+    result = it
+    # TODO: disallow when the owner is not a module
```
saem, line 1469: Wouldn't this preclude using deprecated for example on inner routines? It's a handy trick when working with really large proc refactoring.

compiler/sem/pragmas.nim
saem: 🤔 Future thought, if we had a more general `recordPragma` facility we could record all pragma application events, then unwind them when leaving compile or other such speculative/sandboxed execution contexts.

compiler/sem/pragmas.nim
```diff
@@ -1657,65 +1418,324 @@ proc prepareSinglePragma(
...
+    # XXX: the special word is not in any statement-related allow list
+    result = noVal(c, it)
+    var tmp = c.config.options
+    result = processOption(c, it, tmp)
+    c.config.options = tmp
+  of wExperimental:
+    if not isTopLevel(c):
+      result = c.config.newError(it,
+                  PAstDiag(kind: adSemExperimentalRequiresToplevel))
+    result = processExperimental(c, it)
```
saem: Not sure what was going on with the old code, but this should be if/else or maybe:

```suggestion
    result = processExperimental(c, it)
    if not isTopLevel(c):
      result = c.config.newError(result,
                  PAstDiag(kind: adSemExperimentalRequiresToplevel))

```

compiler/sem/pragmas.nim
```diff
@@ -1657,65 +1418,324 @@ proc prepareSinglePragma(
...
+  else:
```
saem: I was about to say that this switch should be exhaustive, but then I did an inventory of pragmas and here is what I got:

The constant `stmtPragmas` is missing the following pragmas from this case statement:
- `wCallConv` (same as pushing a calling convention, cooooool)
- `wDebugger` (dead with the embedded nim debugger, noop explicitly in the case?)
- `wProfiler`
- `wByRef`

The constant `stmtPragmas` has the following that are missing from this case statement:
- `wLinksys`
- `wPure` 
- `wDeadCodeElimUnused` (this one is deprecated per the comment)
- `wEffects` (pragma remains and `sempass2` sees it and emits effects at that point)
- `wAssert` (does nothing, might have been an old way to force an assertion)

The first set can be safely added to `stmtPragmas`. For the second set I would add `wEffects` to the case, I don't think it belongs in the symbol pragmas, the rest should be safe to drop from the constant. At which point this case can be exhaustive.

compiler/sem/pragmas.nim
```diff
@@ -1438,119 +1299,53 @@ proc prepareSinglePragma(
...
-        else:
-          sym.typ.callConv = wordToCallConv(k)
-          sym.typ.flags.incl tfExplicitCallConv
-      of wEmit:
-        result = pragmaEmit(c, it)
-      of wLinearScanEnd, wComputedGoto:
-        result = noVal(c, it)
+        sym.typ.callConv = wordToCallConv(k)
+        sym.typ.flags.incl tfExplicitCallConv
       of wEffects:
```
saem: I think this needs to move to the `applyStmtPragma`, I don't believe it's a symbol pragma.

compiler/sem/pragmas.nim
```diff
@@ -1657,65 +1418,324 @@ proc prepareSinglePragma(
...
+
+      r.add applyStmtPragma(c, owner, it, result)
+    else:
+      # it might still be a custom pragma
+      result = wInvalid
+      r.add semCustomPragma(c, it)
+
+proc semSinglePragmaInStmt(
+    c: PContext; owner: PSym, it: PNode, validPragmas: TSpecialWords, r: var seq[PNode]): TSpecialWord =
+  ## Processes a the single pragma expression `it`. If the pragma is a user-
```
saem:
```suggestion
  ## Processes the single pragma expression `it`. If the pragma is a user-
```

compiler/sem/pragmas.nim
```diff
@@ -1728,60 +1748,171 @@ proc hasPragma*(n: PNode, pragma: TSpecialWord): bool =
...
+      return true
+
+  result = false
+
+proc pragmaExpr*(c: PContext, n: PNode): PNode =
+  ## Pre-processes the pragma list `n` for the purpose of further processing
+  ## by sem in the context of applying pragmas to an expression.
+  assert n.kind == nkPragma
+  result = copyNode(n)
+  for i in 0..<n.len:
```
saem: Just curious why favour indexing over item iteration?

compiler/sem/pragmas.nim
```diff
@@ -1728,60 +1748,171 @@ proc hasPragma*(n: PNode, pragma: TSpecialWord): bool =
...
+    let k = semSinglePragmaInStmt(c, owner, n[i], stmtPragmas, result.sons)
+
+    case k
+    of wPush:
+      hasError = processPush(c, n.info, n.sons.toOpenArray(i+1, n.len-1), result)
+      break
+    of wPop:
+      result.add processPop(c, n[i])
+      # XXX: maybe it'd be better to either report an error if there are items
```
saem, line 1906: Hmm, on module close if there are unpopped pragmas that's an error? I think that's a good idea. Basically someone forgot to "clean-up" and this would be especially helpful if someone is stuck with push/pop across conditional compilation.

saem (commented):
All done; I think `wEffects` handling is the only error I noticed.

Rest are not blockers, but the exhaustiveness on case statements would be nice if feasible. If not, it's already such a big improvement. 🤩

compiler/sem/pragmas.nim
zerbina: Yep, I agree. Specifically for the `wUsed` pragma in *statement* pragmas, disallowing it might be better, I think. The following is currently allowed and works:
```nim
proc p() =
  var x = 0
  {.used.} # `p` is marked as used
```
which I'm not sure is a good idea.

compiler/sem/pragmas.nim
zerbina: It wouldn't, no. The `wDeprecated` handling here is only for pragma statements, which I believe were only intended to be used with modules, but due to the previous application architecture were (and currently still are) also usable inside procedures, like so:
```nim
proc p() =
  var x = 0
  {.deprecated.} # previously applied to `p`, but now to the module `p` is part of as deprecated
```

I think this usage should be disallowed, hence the todo.

**Edit**: corrected the description of how `.deprecated` works when in a statement.

compiler/sem/pragmas.nim
zerbina: You're very much correct. I'm also going to add a test for the pragma, as the testsuite doesn't seem to have caught this mistake.

compiler/sem/pragmas.nim
zerbina: I don't know :P. The body likely required access to the index at one point, but there's no reason to not use item iteration now.

compiler/sem/pragmas.nim
zerbina: The comment refers to all pragmas following a `pop` being silently dropped (e.g. `{.pop, error.}` doesn't abort compilation with a user error). This mirrors the previous behaviour, but I think in the future, the compiler should produce an error (or at least a warning) instead.

Thinking about it more, the three special pragmas should probably also be enforced to also come first (but not as part of this PR).

compiler/sem/pragmas.nim
zerbina: Applying `wFinal` when the type is already marked as `.inheritable` is now an error.

compiler/sem/pragmas.nim
zerbina: I restored the previous behaviour (`.error`s on, for example, a `var` being treated like they are in statements) for now.

compiler/sem/pragmas.nim
zerbina: Removed the pragma and special word.

compiler/sem/pragmas.nim
zerbina: An `if`/`else` is indeed better, thanks.

compiler/sem/pragmas.nim
zerbina: I made the case statement exhaustive, although with two differences:
- `wDeadCodeElimUnused` is kept as a no-op, with a comment suggesting its removal
- `wAssert` is kept -- it can be used to enable/disable assertions for a section of code

zerbina: Oh, I misread the code, `wAssert` is indeed unused.

compiler/sem/pragmas.nim
saem: Got it, thanks for the explanation.

compiler/sem/pragmas.nim
saem: Yup, that's not a good idea, it'll be easier for something to remove after this refactor! 🚀

compiler/sem/pragmas.nim
saem: I think the idea at some point was to allow "filtered" pop, so a pop would only pop the items that followed it. This unstructured popping, seems... dangerous.

I agree we should produce an error in the future.

saem:
Did a scan of the comments and recent changes, seems like anything that would remotely block this PR are done -- cancel if not.

Take it away,
bors r+

PR #578, opened 2023-03-13
front: rework `--eval`

zerbina:
## Summary

The `--eval` option was previously responsible for two things:
- if no command was specified yet, changing the command to `e`
- overriding the content of the main module with the given argument

Its implementation was scattered and the behaviour not very intuitive, so the option is simplified:
- it no longer accepts an operand, but instead uses the second command- line argument as the main modules content
- it doesn't set the active main command anymore

In addition, the option is renamed to `fromcmd` in order to better match its new behaviour. Example usage: `nim e --fromcmd "echo 1"`

One thing that is now no longer possible is overriding the contents of the main module while still using the provided project path.

## Details

- instead of multiple booleans, the source mode for the main module is now provided via an enum
- remove the now unused state previously required by the `--eval` implementation
- perform project name and path initialization in a single place (`processCmdLineAndProjectPath`)
- adjust the `truner.nim` test, which made use of `--eval`

### NimScript execution

When executing a NimScript file by either using the `e` command or only providing the path of a NimScript file, running the script now happens when during main command processing, instead of during configuration loading.

### Misc

Remove the `IdGenerator` parameter from `runNimScript`. The `IdGenerator` of the `ModuleGraph` used for executing the NimScript is passed to the VM context setup instead, which is a step towards sandboxing/isolating NimScript execution.

---

## Notes for Reviewers
* being able to provide the code to compile on the command-line (without using the standard input) seemed like a useful feature, so I opted to keep it.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/front/main.nim
+++ b/compiler/front/main.nim
@@ -72,6 +72,8 @@ from compiler/ast/report_enums import ReportKind,
   repWarningKinds,
   rstWarnings
 
+from compiler/front/scriptconfig import runNimScript
+
 when not defined(leanCompiler):
   import
     compiler/backend/jsgen,
@@ -344,7 +346,7 @@ proc mainCommand*(graph: ModuleGraph) =
     else:
       wantMainModule(conf)
       let docConf = if conf.cmd == cmdDoc2tex: DocTexConfig else: DocConfig
-      loadConfigs(docConf, cache, conf, graph.idgen)
+      loadConfigs(docConf, cache, conf)
       defineSymbol(conf, "nimdoc")
       body
 
@@ -396,7 +398,7 @@ proc mainCommand*(graph: ModuleGraph) =
     when defined(leanCompiler):
       conf.quitOrRaise "compiler wasn't built with documentation generator"
     else:
-      loadConfigs(DocConfig, cache, conf, graph.idgen)
+      loadConfigs(DocConfig, cache, conf)
       commandRst2Html(cache, conf)
   of cmdRst2tex, cmdDoc2tex:
     for warn in rstWarnings:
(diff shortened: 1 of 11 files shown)


saem:
> One thing that is now no longer possible is overriding the contents of the main module while still using the provided project path.

Not being able to do this sounds like a feature. 😆

compiler/front/options.nim
```diff
@@ -221,11 +227,6 @@ type
     ## against infinite macro expansion recursion
     exitcode*: int8
 
-    # `--eval` flag handling
-    cmdInput*: string    ## Code to evaluate from `--eval` switch
-    projectIsCmd*: bool  ## whether we're compiling from a command input (`--eval` switch)
-    implicitCmd*: bool   ## whether some flag triggered an implicit `command` (`--eval`)
-
```
saem, line 228: yay, no more mess of bools

compiler/front/main.nim
```diff
@@ -462,12 +464,20 @@ proc mainCommand*(graph: ModuleGraph) =
...
-    # main NimScript logic handled in `loadConfigs`.
+    let s =
+      case conf.inputMode
+      of pimStdin: llStreamOpenStdIn()
+      of pimCmd:   llStreamOpen(conf.commandArgs[0])
+      of pimFile:  llStreamOpen(conf.projectFull, fmRead)
+
+    # XXX: the ``runNimScript`` from ``scriptconfig`` is used, but the script
+    #      is not meant for configuration. While this has no practical
+    #      consequences right now, it's still a domain violation
```
saem, line 479: Fair compromise, this still gets things closer to where they should be.

saem (approved):
bors r+

Much cleaner. 🧹 ✨

PR #580, opened 2023-03-13
adjust the `.compileTime` pragma's semantics

zerbina:
## Summary

Adjust the semantics of the `.compileTime` pragma regarding locals in order to make it compose better and to slightly simplify/clean-up the implementation. The pragma is now:
- disallowed for locals not defined inside a compile-time-only routine
- treated as a no-op when used with locals defined inside a compile- time-only routine

### General changes

- add a dedicated error diagnostic for reporting illegal usage of the `.compileTime` pragma
- update the "compileTime pragma" section in the manual

### Semantic analysis changes

- apply the `.compileTime` pragma to all symbols (on which it is usable) during pragma application

---

## Notes for Reviewers
* this fixes a blocker of #573

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -373,7 +373,8 @@ type
     sfShadowed        ## a symbol that was shadowed in some inner scope
     sfThread          ## proc will run as a thread
                       ## variable is a thread variable
-    sfCompileTime     ## proc can be evaluated at compile time
+    sfCompileTime     ## proc can only be used in compile time contexts;
+                      ## global is accessible in compile time contexts
     sfDispatcher      ## copied method symbol is the dispatcher
                       ## deprecated and unused, except for the con
     sfBorrow          ## proc is borrowed
@@ -1134,6 +1135,7 @@ type
     adSemCannotInferTypeOfLiteral
     adSemProcHasNoConcreteType
     adSemPragmaDisallowedForTupleUnpacking
+    adSemIllegalCompileTime
     adSemDifferentTypeForReintroducedSymbol
     adSemThreadvarCannotInit
     adSemWrongNumberOfVariables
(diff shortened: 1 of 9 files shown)


saem (approved):
I didn't get a chance to pull down the code and see where the NPE is coming from, but the change makes a lot of sense.

zerbina:
Should I move the new logic to the pragma application phase? The logic being located there feels more correct, and the error could also be associated with the `.compileTime` pragma's location in the code then (with an additional hint, if the pragma is implicit).

The downside is that, until `var` and `let` entities defined directly inside `static` or `const` are correctly treated as globals, the application logic would need to query `inStaticContext`.

saem:
Hmm, the issue is that `let` and `var` handling need to be revised or we won't get a resolution, where we promote them to globals. If the issue is made apparent in the `pragma` code then the "pain" won't be felt in the "right place" as to where the solution needs to start.

In the above I care about "messy code should be messy" principle, or insure we "feel" the downside in the location where the majority of remedy needs to be applied or at least start. My reasoning about where to actually place that "pain" in this particular circumstance might be wrong.

saem:
Non-blocking, for reference: filed issue for broader `compiletime` discussion so we can figure out what should actually happen.

https://github.com/nim-works/nimskull/issues/585

zerbina:
I scaled down the impact of the PR a bit:
1) routines are not inferred as being compile-time-only when they're defined in a compile-time-only context
2) no change is made to the JS code generator

Implementing 1) later is very easy, as only the condition of the `if` that guards the `sfCompileTime` flag in `semProcAux` needs to be adjusted. As documented in the issues linked by @saem above, inheriting `sfCompileTime` would make it impossible for procedural values derived from inner procedures of `.compileTime` procedures to cross the compile-time/run-time boundary.

For 2), only changing the location of the assertion is not enough -- a more elaborate fix is required. Since that'd be unrelated to the main part of this PR, I decided to not implement it here.

zerbina:
I didn't consider that whether the symbol is that of a global is not yet known when applying the declaration's pragmas, so the `sfCompileTime` has to continue being located in `semNormalizedLetOrVar`, for now.

bors r+

PR #582, opened 2023-03-14
removed `useversion` compiler switch

saem:
## Summary

Removed `useversion` compiler switch and associated v1 and v1.2 emulation.

## Details

No behavioural changes, this just explains some of the "emulation" that was dropped.

Removed `nimDoesntTrackDefects` define in various gc modules. It was used to emulate old nim 1.0 behaviour where thread destruction callbacks weren't `raise: []` and the compiler tracked defects in `raise`s.

Remove define `nimOldRelativePathBehaviour`. `std/os.relativePath` will always return `"."` if the path would otherwise be empty.

Misc:
- dropped v1 type equality overrides in `eqTypeFlags` (made `const`)
- removed enum fields: `optNimV1Emulation` and `optNimV12Emulation`
- updated code accordingly, eliminating associated conditionals
- removed stdlib conditional checks for version

---

## Notes for Reviewers
* less is more


saem:
Bors r+

(feel free to cancel if reviewing)

compiler/ast/ast_types.nim
```diff
@@ -715,16 +715,12 @@ const
   tfReturnsNew* = tfInheritable
   skError* = skUnknown
 
-var
```
zerbina, line 718: If I remember correctly, `eqTypeFlags` being a `const` should make `sameType` and friends side-effect free :tada:.

zerbina (approved):
Looks good to me.

PR #584, opened 2023-03-14
sem: fix the compile-time-only detection

zerbina:
## Summary

The `inStaticContext` state that is, among other things, used to track whether the currently analysed code is only available at compile-time was part of the shared context (`PContext`), which is incorrect, as it means that the bodies of generic procedures instantiated in, for example, a `static` block are treated as being inside a compile-time-only context (which they are not).

Instead, the `inStaticContext` state is now part of `TProcCon`, and thus local to the analysis of a routine.

## Details

- always mark macros as being available compile-time only. This makes checking for compile-time-only routines easier, as macros no longer need to be tested for separately

---

## Notes for Reviewers
* I found the issue during the `.compileTime` pragma related work
* except for macros, this change doesn't affect which symbols are marked with `sfCompileTime` 

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semdata.nim
+++ b/compiler/sem/semdata.nim
@@ -78,6 +78,17 @@ type
     caseContext*: seq[tuple[n: PNode, idx: int]]
     localBindStmts*: seq[PNode]
 
+    inStaticContext*: int
+      ## > 0 if we are inside a ``static`` block/expression or initializer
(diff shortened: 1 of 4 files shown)


saem (approved):
Awesome!

bors r+

PR #586, opened 2023-03-14
enable lambda-lifting for the JavaScript target

zerbina:
## Summary

Previously, inner closure procedures were not lifted into free-standing procedures during `transf`, but rather kept as-is and mapped to nested JavaScript functions by `jsgen`. Besides requiring the JavaScript target to be special-cased in `transf` and `lambdalifting`, this also caused observable semantic issues: the destructors for captured variables were called when leaving the scope in which they were defined, not when the closure got destroyed.

In order to bring the closure semantics in line with the other targets and as a prerequisite for full ORC support, lambda-lifting is now also enabled when using the JS target -- the code generator is adjusted to support the lifted procedures. The code generator changes are also a
preparation for full closure iterator support.

As a side-effect, closures now also work as expected when using the JS target and executing non-compile-time-only procedures that have inner closure procedures at compile-time.

## Details

- remove the dispatching based on the selected backend from `transf` and `lambdalifting`
- remove all code related to the previous closure support from `jsgen`
- implement support for lifted procedures in `jsgen`

The hidden environment parameter of routines using the `.closure` calling convention is mapped to the `this` parameter at the JavaScript level.

When a NimSkull closure object is created, a new function instance is created via `bind` and the provided environment bound to the `this` parameter. In addition, the both the function and environment are also assigned to the respective `prc` and `env` properties on the function object -- this allows for later retrieval.

Compared to using an object, this approach has the benefit that closures can still be passed to an imported procedure that expects a JavaScript function object, without requiring any changes to the user code.

---

## Notes for Reviewers
* besides being a prerequisite for full ORC support, this change also simplifies #550

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/backend/jsgen.nim
+++ b/compiler/backend/jsgen.nim
@@ -143,25 +143,18 @@ type
     options: TOptions
     module: BModule
     g: PGlobals
-    generatedParamCopies: IntSet
     beforeRetNeeded: bool
     unique: int    # for temp identifier generation
     blocks: seq[TBlock]
     extraIndent: int
-    up: PProc     # up the call chain; required for closure support
     declaredGlobals: IntSet
 
 template config*(p: PProc): ConfigRef = p.module.config
 
 proc indentLine(p: PProc, r: Rope): Rope =
   result = r
-  var p = p
-  while true:
-    for i in 0..<p.blocks.len + p.extraIndent:
-      prepend(result, rope"  ")
-    if p.up == nil or p.up.prc != p.prc.owner:
-      break
-    p = p.up
+  for i in 0..<p.blocks.len + p.extraIndent:
+    prepend(result, rope"  ")
 
 template line(p: PProc, added: string) =
   p.body.add(indentLine(p, rope(added)))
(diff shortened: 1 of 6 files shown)


compiler/backend/jsgen.nim
```diff
@@ -1041,6 +1044,14 @@ proc generateHeader(p: PProc, typ: PType): Rope =
       result.add(name)
       result.add("_Idx")
 
+  if typ.callConv == ccClosure:
+    # generate the name for the hidden environment parameter. When creating a
+    # closure object, the environment is bound the ``this`` argument of the
```
```suggestion
    # closure object, the environment is bound to the ``this`` argument of the
```

saem (approved):
This is awesome, good to merge! 🚀 

I thought of two follow-ups. I'll start on the first one ~tomorrow and see about the second one based on the outcome:
1. have as much of `tests/lang_callable/iter` run for js/all platforms see where things land
2. quick update of the manual to remove the "big" mentions of can't use closure iterators for JS (rest can be long tail)

zerbina:
> 1. have as much of `tests/lang_callable/iter` run for js/all platforms see where things land

Yep, that's a good idea. Having the `iter` tests work, at least in theory, on the JS and VM would be a big help when implementing closure iterator in the respective backends.

compiler/backend/jsgen.nim
zerbina: Done, thanks.

PR #587, opened 2023-03-17
sempass2: fix illegal capture detection

zerbina:
## Summary

The illegal capture detection treated all routines enclosed by a `.compileTime` routine as also being being compile-time-only. This to illegal code (according to the current language rules) not being rejected.

The rules are updated and the detection adjusted to work according to them; see the changes to the related tests for details.

---

## Notes for Reviewers
* this doesn't block off inferring routines defined inside compile-time-only contexts as `.compileTime` -- it only removes the decision making from `sempass2`

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/sempass2.nim
+++ b/compiler/sem/sempass2.nim
@@ -1621,29 +1621,27 @@ proc detectCapture(owner, top: PSym, n: PNode, marker: var IntSet): PNode =
         # unwind
         return
 
-proc isCompileTimeProc2*(s: PSym): bool =
-  ## Tests if `s` is a compile-time procedure, but compared to
-  ## ``isCompileTimeProc``, also considers where `s` is defined. In the
-  ## following example:
-  ##
-  ## .. code-block:: nim
-  ##
-  ##   proc a() {.compileTime.} =
-  ##     proc b() =
-  ##       discard
-  ##
-  ## ``b`` is also a procedure that can only be run at compile-time, but it
-  ## doesn't have the ``sfCompileTime`` flag set
-  # TODO: propagate the ``sfCompileTime`` flag downwards instead (likely in
-  #       sempass2) and use ``isCompileTimeProc`` directly
-  # FIXME: procedures defined inside ``static`` blocks are not detected
-  var s = s
-  while s.kind != skModule:
-    if isCompileTimeProc(s):
-      return true
-    s = s.skipGenericOwner
+proc canCaptureFrom*(captor, target: PSym): bool =
+  ## Tests if the `captor` routine is allowed to capture a local from `target`,
+  ## taking the compile-time/run-time boundary into account:
+  ## 1) attempting to capture a local defined outside an inner macro from
+  ##   inside the macro is illegal
+  ## 2) closing over a local defined inside a compile-time-only routine from a
+  ##   routine than can also be used at run-time is only valid if the chain of
+  ##   enclosing routines leading up to `target` are all compile-time-only
+  ## 3) a compile-time-only routine closing over a run-time location is illegal
+  template isCompTimeOnly(s: PSym): bool =
+    sfCompileTime in s.flags
 
-  result = false
+  result = not captor.isCompTimeOnly or target.isCompTimeOnly    # rule #3
+  # check `captor` and all enclosing routines up to, but not including,
+  # `target` for rule violations
+  var s = captor
+  while result and s != target:
+    result =
+      s.kind != skMacro and                                      # rule #1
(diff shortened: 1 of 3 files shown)


zerbina:
Looking at the longe-term development, I don't think the approach of detecting illegal captures across the compile-/run-time boundary like in `sempass2` is the right one.

I think there are two (at least) things conflated with the current scoping mechanism:
- visibility: whether the existence of a symbol is known and it can be used in a declarative context (e.g. `typeof(x)`)
- accessibility: whether the symbol can be used in an imperative context (e.g. calling a procedure, reading from a location, etc.)

Separating them on a logical, and maybe even physical, level would allow for already detecting during the first semantic pass what the logic I adjusted here detects, as well as help with moving semantic analysis out of `vmgen`.

compiler/sem/sempass2.nim
```diff
@@ -1621,29 +1621,27 @@ proc detectCapture(owner, top: PSym, n: PNode, marker: var IntSet): PNode =
...
-  ## ``b`` is also a procedure that can only be run at compile-time, but it
-  ## doesn't have the ``sfCompileTime`` flag set
-  # TODO: propagate the ``sfCompileTime`` flag downwards instead (likely in
-  #       sempass2) and use ``isCompileTimeProc`` directly
-  # FIXME: procedures defined inside ``static`` blocks are not detected
-  var s = s
-  while s.kind != skModule:
-    if isCompileTimeProc(s):
-      return true
+proc canCaptureFrom*(source, dest: PSym): bool =
```
saem:
```suggestion
proc canCaptureFrom*(captor, trgt: PSym): bool =
```

Suggesting this because I kept finding myself reconfirming if `source` was the routine trying to capture or the source of potential locals.

zerbina: I adjusted the names. Thanks for the suggestion.

compiler/sem/sempass2.nim
```diff
@@ -1621,29 +1621,27 @@ proc detectCapture(owner, top: PSym, n: PNode, marker: var IntSet): PNode =
...
+  while s != target:
+    if s.kind == skMacro or (s != captor and
+       isCompTimeOnly(s) != isCompTimeOnly(target)):
+      return false
     s = s.skipGenericOwner
 
-  result = false
+  # it's not allowed for a compile-time-only routine to close over a run-time
+  # location
+  result = not (isCompTimeOnly(captor) and not isCompTimeOnly(target))
```
saem: I had a bit too much fun, but does this flow better?

```nim
  result = not captor.isCompTimeOnly or target.isCompTimeOnly
  var s = captor
  while result and s != target: # does this need to check for nil?
    result =
      s.kind != skMacro and
      (s == captor or        # for captor ensure not macro
       s.isCompTimeOnly == target.isCompTimeOnly) # ensure same time
    s = s.skipGenericOwner
```

compiler/sem/sempass2.nim
```diff
@@ -1684,8 +1682,7 @@ proc trackProc*(c: PContext; s: PSym, body: PNode) =
   if cap != nil:
     # the procedure captures something and thus requires a hidden environment
     # parameter
-    if s.kind == skMacro or
-       (isCompileTimeProc2(s) and not isCompileTimeProc2(cap.sym.owner)):
+    if not canCaptureFrom(s, cap.sym.owner):
```
saem, line 1685: I _believe_ it's the same logic if we drop the negation and swap the body of this `if` branch with that of the `else`.

I prefer avoiding negations, but by no means a hard requirement.

saem (approved):
Looks good to me, I have two more suggestions; but they're not hard blockers.

compiler/sem/sempass2.nim
saem: Pretty sure I did all the De Morgan's correctly. 🤞🏽

compiler/sem/sempass2.nim
zerbina: Not quite, the explicit calling convention check has to happen before setting the convention. The `elif` branch could of course be moved into the `if` branch as a nested `if` statement, but I think doing so would make the logic harder to follow.

saem: D'oh, didn't pick-up on that, thanks for clarifying.

saem:
bors r+

(assuming everything will pass; looks great, btw)

compiler/sem/sempass2.nim
zerbina: It does flow a lot better -- thanks for the suggestion.

PR #590, opened 2023-03-17
sem: decouple `semMacroDef` from `semProcAux`

zerbina:
## Summary

The `semMacroDef` procedure (semantic analysis of a macro definition) no longer uses `semProcAux`, which allows for removing conditional logic from `semProcAux`, makes it easier to extend macro definition analysis, and is meant to help with the planned `semProcAux` refactoring.

As part of the refactoring, multiple issues were found and fixed. The changes:

#### General fixes

- `nkPragmaExpr` nodes are now disallowed in routine name slots (previously possible via macros)
- template with gensym'ed names have the correct owner now
- it's now enforced that symbols have the expected kind (e.g. that a symbol in an `nkProcDef` name slot is a `skProc`)
- anonymous macros, templates, converters, and methods are disallowed (previously possible via macros)
- dot operators can no longer be enabled with `{.experimental: "destructor".}`
- template symbol suggestions now have the correct relative order to other routine suggestions

#### Term-rewriting macros

- patterns are only registered *once* now (previously each one was redundantly registered twice), improving performance
- term-rewriting macros and templates no longer apply to their own bodies
- patterns are now disallowed on routines outside of macros and templates
- term-rewriting *macros* are now only registered when the definition doesn't have errors

## Details

In general, the commit unifies multiple parts of routine definition analysis by merging the ones that are the same for all of them into dedicated procedures.

All definitions of callables except labmdas are now passed to `semRoutineDef` first. The procedure dispatches to the respective `semXDef` procedure after applying operations common to the all routines, such as applying macro pragmas and producing the name.

### Routine-def analysis

- remove the unused `validPragmas` parameter from `semProcAnnotation`
- add `semRoutineParams`, which replaces `setGenericParamsMisc` and reduces the code duplication between `semProcAux`, `semTemplateDef`, and (the new) `semMacroDef`
- move the check for dot and call operator names into a separate procedure and use AST diagnostics for the errors
- refactor `semProcAux` slightly in order for it to not mutate the input AST as much; this is necessary for `semRoutineParams` to be used
- remove all macro and template related conditional logic from `semProcAux`
- remove the redundant gensym-related check from `semIteratorDef`
- more `nkError` propagation

### Patterns

- remove pattern registration from `semPattern` (separation of concerns)
- use `addPattern` in `semTemplateDef` and `semMacroDef` when registering a pattern

### Other

- unify the lookup of line information for identifiers in name slots (`getIdentLineInfo`)
- replace `determineType` with an assertion; a non-magic routine symbol without type information reaching overload resolution is a logic error

---

### To-do
- [x] add regression tests for the fixed issues
- [x] resolve the remaining inline-questions

## Notes for Reviewers
* apart from the small to-dos above, I consider the PR done

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/hlo.nim
+++ b/compiler/sem/hlo.nim
@@ -29,19 +29,12 @@ proc evalPattern(c: PContext, n: PNode): PNode =
     original = n
 
   let s = n[0].sym
-  case s.kind
-  of skMacro:
-    result = semMacroExpr(c, n, s)
+  result =
+    case s.kind
+    of skMacro:    semMacroExpr(c, n, s)
+    of skTemplate: semTemplateExpr(c, n, s, {efFromHlo})
+    else:          unreachable("not a macro/template: " & $s.kind)
 
-  of skTemplate:
-    result = semTemplateExpr(c, n, s, {efFromHlo})
-
-  else:
-    if s.isError:
-      result = s.ast
-    else:
-      result = semDirectOp(c, n, {})
-    
   if c.config.hasHint(rsemPattern):
     c.config.localReport(n.info, SemReport(
       kind: rsemPattern,
--- a/compiler/sem/semcall.nim
+++ b/compiler/sem/semcall.nim
@@ -33,8 +33,6 @@ proc sameMethodDispatcher(a, b: PSym): bool =
       # be disambiguated by the programmer; this way the right generic is
       # instantiated.
 
-proc determineType(c: PContext, s: PSym)
-
 proc initCandidateSymbols(c: PContext, headSymbol: PNode,
                           initialBinding: PNode,
                           filter: TSymKinds,
@@ -96,7 +94,8 @@ proc pickBestCandidate(c: PContext,
       sym = nextOverloadIter(o, c, headSymbol)
       scope = o.lastOverloadScope
       continue
-    determineType(c, sym)
+    c.config.internalAssert(sym.typ != nil or sym.magic != mNone,
+                            "missing type information")
     initCandidate(c, z, sym, initialBinding, scope)
     if c.currentScope.symbols.counter == counterInitial or syms.len != 0:
       matches(c, n, z)
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -324,6 +324,23 @@ proc identWithin(n: PNode, s: PIdent): bool =
     if identWithin(n[i], s): return true
   result = n.kind == nkSym and n.sym.name.id == s.id
 
+proc getIdentLineInfo(n: PNode): TLineInfo =
+  ## Returns the line information of the identifier-like node in the
+  ## (semantically valid) AST `n` appearing in a name slot.
+  var n {.cursor.} = n
+  # unpack the node until we reach the identifier or symbol
+  if n.kind == nkPragmaExpr:
+    n = n[0]
+  if n.kind == nkPostfix:
+    n = n[1]
+  if n.kind == nkAccQuoted:
+    n = n[0]
+
+  result =
+    case n.kind
+    of nkIdent, nkSym: n.info
+    else:              unreachable(n.kind)
+
 proc semIdentDef(c: PContext, n: PNode, kind: TSymKind): PSym =
   addInNimDebugUtils(c.config, "semIdentDef", n, result)
   if isTopLevel(c):
@@ -337,19 +354,7 @@ proc semIdentDef(c: PContext, n: PNode, kind: TSymKind): PSym =
       incl(result.flags, sfGlobal)
   result.options = c.config.options
 
-  proc getLineInfo(n: PNode): TLineInfo =
-    case n.kind
-    of nkPostfix:
-      if len(n) > 1:
-        return getLineInfo(n[1])
-    of nkAccQuoted, nkPragmaExpr:
-      if len(n) > 0:
-        return getLineInfo(n[0])
-    else:
-      discard
-    result = n.info
-
-  let info = getLineInfo(n)
+  let info = getIdentLineInfo(n)
   suggestSym(c.graph, info, result, c.graph.usageSym)
 
 proc checkNilableOrError(c: PContext; def: PNode): PNode =
@@ -2258,8 +2263,7 @@ proc addResult(c: PContext, n: PNode, t: PType, owner: TSymKind) =
     addParamOrResult(c, c.p.resultSym, owner)
 
 
-proc semProcAnnotation(c: PContext, prc: PNode;
-                       validPragmas: TSpecialWords): PNode =
+proc semProcAnnotation(c: PContext, prc: PNode): PNode =
   var n = prc[pragmasPos]
   if n == nil or n.kind == nkEmpty: return
   for i in 0..<n.len:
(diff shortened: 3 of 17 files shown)


compiler/sem/hlo.nim
```diff
@@ -29,19 +29,12 @@ proc evalPattern(c: PContext, n: PNode): PNode =
...
 
   let s = n[0].sym
-  case s.kind
-  of skMacro:
-    result = semMacroExpr(c, n, s)
+  result =
+    case s.kind
+    of skMacro:    semMacroExpr(c, n, s)
+    of skTemplate: semTemplateExpr(c, n, s, {efFromHlo})
+    else:          unreachable(s.kind)
```
saem:
```suggestion
    else:          unreachable("patterns only on macro/template " & $s.kind)
```

compiler/sem/semcall.nim
```diff
@@ -96,7 +94,8 @@ proc pickBestCandidate(c: PContext,
       sym = nextOverloadIter(o, c, headSymbol)
       scope = o.lastOverloadScope
       continue
-    determineType(c, sym)
+    c.config.internalAssert(sym.typ != nil or sym.magic != mNone,
+                            "missing type information")
```
saem, line 98: This is a good cue!

When there is something "weird" in the compiler, throw in asserts and flush out the issues.

compiler/sem/semstmts.nim
```diff
@@ -2587,44 +2591,46 @@ proc semMethodPrototype(c: PContext; s: PSym; n: PNode) =
...
+  case n.kind
   of nkEmpty:
-    s = newSym(kind, c.cache.idAnon, nextSymId c.idgen, c.getCurrOwner, n.info)
-    s.flags.incl sfUsed
-    n[namePos] = newSymNode(s)
+    if allowAnon:
+      s = newSym(kind, c.cache.idAnon, nextSymId c.idgen, c.getCurrOwner, n.info)
+      s.flags.incl sfUsed
+    else:
+      # QUESTION: create an error symbol here instead?
```
saem: Unless we start semming the rest of such a routine and try to make progress inspite of the error, I don't think we need the extra error/tainted value tracking by injecting the sym here and in the symbol table.

compiler/sem/semstmts.nim
```diff
@@ -2633,15 +2639,30 @@ proc semProcAux(c: PContext, n: PNode, kind: TSymKind,
...
+
+  # TODO: refactor ``semParamList`` to not modify the input parameter list
+  routine[paramsPos] = formal
+
+  result =
+    case formal.kind
+    of nkFormalParams:
+      semParamList(c, routine[paramsPos], routine[genericParamsPos], kind)
+    of nkEmpty:
+      nil # let the callee handle empty parameter lists
```
saem: since we have the `routine` we can produce the correct info, why not avoid the `nil` return by making the type here?

compiler/sem/semstmts.nim
```diff
@@ -2944,24 +2993,151 @@ proc semConverterDef(c: PContext, n: PNode): PNode =
...
-  var allUntyped = true
-  for i in 1..<t.n.len:
-    let param = t.n[i].sym
-    if param.typ.kind != tyUntyped: allUntyped = false
-  if allUntyped: incl(s.flags, sfAllUntyped)
-  if n[bodyPos].kind == nkEmpty:
-    localReport(c.config, n, reportSym(
-      rsemImplementationExpected, s))
+
+  result = shallowCopy(n)
```
saem, line 2990: Because we're using `result` in a few places where `n` was previously used, I vaguely recall there were node flag(s) that didn't get copied (not in `persistentNodeFlags`) and caused some breakages.

Sorry, memory is fuzzy at the moment.

compiler/sem/semstmts.nim
```diff
@@ -2944,24 +2993,151 @@ proc semConverterDef(c: PContext, n: PNode): PNode =
...
+
+    hasError = true
+
+  closeScope(c)
+  popOwner(c)
+
+  if hasError:
+    # shortcut pattern registration
+    # QUESTION: transition to error symbol?
+    return c.config.wrapError(result)
```
saem: I think this might be required because of the way some of the lookups happen during `sigmatch`, but it's more a defensive measure and I didn't get to ground on why that was the case. I say leave it for now -- worst case it'll surface some bugs.

saem (approved):
I've read it once and I really like it; however, I did run out of time tonight so I didn't get a chance to chase up some of the flag weirdness and double check the order of operations -- some of this CI might have already surfaced. I'll take another peek tomorrow.

compiler/sem/semstmts.nim
zerbina: It should have been `caller` instead of `callee`, sorry.

As for avoiding `nil`, I don't see how that would work outside of resorting to conditional logic in `semRoutineParams`. Templates treat a missing parameter list differently than a macro or other routine.

saem: My bad I missed the difference in the template callsite.

zerbina:
The CI failure is because of the sexp issue that I made a fix for already; I missed rebasing onto devel after it got merged.

compiler/sem/semstmts.nim
zerbina: I looked into it, and the only flags that I think might be an issues are `nfNoRewrite` and ` nfHasComment`. Given that `semTemplateDef` also doesn't copy the flags and, to my knowledge, there haven't been any issues so far, I think not copying non-persistent flags is safe here.

Nonetheless, comments are indeed no longer copied at the moment (neither for macros nor other routines), so that still needs to be fixed.

saem:
Soon, soon, mwahahaha.

compiler/sem/semstmts.nim
zerbina: I missed that `copyNodeImpl` copies the comment, so no additional action is required for them.

zerbina:
@saem Do you still want to do a second pass? I'm asking since I'd combine pushing my changes with squashing and rebasing, but don't want to erase the history if you still need it.

saem:
If you can rebase without the squash that'd be cool; if that's really difficult then I can do a quick pass now.

zerbina:
I'll rebase without squashing then.

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -0,0 +1,59 @@
+discard """
+  description: "Generated symbols not matching the definition must be rejected"
+  action: reject
+  cmd: "nim check --msgFormat:sexp --filenames:canonical $options $file"
+  nimoutFormat: sexp
+"""
+
+import std/macros
+
+macro replaceName(kind: static[NimSymKind], def: untyped): untyped =
```
saem, line 10: I'll drop the assert from the other PR and go back to the error.

------

In the future, a change like unique identifier, or some change to the macros API, would ensure things are correct by construction and then obviate the error.

zerbina: Yep, I very much agree. There should be now way for macros to generated and place a symbol with the wrong kind in the first place.

compiler/sem/semstmts.nim
```diff
@@ -324,6 +324,23 @@ proc identWithin(n: PNode, s: PIdent): bool =
     if identWithin(n[i], s): return true
   result = n.kind == nkSym and n.sym.name.id == s.id
 
+proc getIdentLineInfo(n: PNode): TLineInfo =
+  ## Returns the line information of the identifier-like node in the
+  ## (semantically valid) AST `n` appearing in a name slot.
+  var n {.cursor.} = n
+  while true:
```
saem:
```suggestion
  while n.kind notin {nkIdent, nkSym}:
```

Don't have to, I just get a bit nervous about infinite loops, at least this way it'll likely npe/assert pretty quick.

zerbina: I manually unrolled the loop. The `if` chain encodes the expectations much better than the previous `while` loop.

zerbina:
I added the remaining tests and merged the `devel` branch into the PR, which *should* fix the remaining test failures. The PR is ready for the final review round now.

saem:
> I added the remaining tests and merged the `devel` branch into the PR, which _should_ fix the remaining test failures. The PR is ready for the final review round now.

🎶  it's the final review down 🎶 

😆

saem:
Bors r+

Let's get this party started. 🥳

PR #591, opened 2023-03-19
internals: minor clean-up of `types` ordinal procs

saem:
## Summary

Minor changes to ordinal related routines in the `types` module. These
are in preparation for a larget `semCase` refactor to use `nkError`.

## Details

Minor changes `firstOrd` and `lastOrd`, largely removing vestigial code
for when enum inheritance was a feature. Replaced bare `localReports`
usage with internal asserts/errors.

Unfortunately, the `ConfigRef` requirement for `firstOrd`, `lastOrd`,
and `lengthOrd` couldn't be dropped as it needs to look at
`ConfigRef.intSize` for `tyInt`. Ordinal querying happens all over the
compiler, with a number of subtle assumptions. Reworking these to use
the extra range of the `Int128` return type wasn't possible, yet.

I also discovered that enums could inherit from other enums, there is
still some vestigial code for this; will address in a follow-up PR.

---

## Notes for Reviewers
I pushed the two float related procs out of the way, it was too annoying,
unfrotunately git diff is having a really hard time with it. No actual
changes to `lengthOrd` or the float routines.

### DIFF
--- a/compiler/ast/types.nim
+++ b/compiler/ast/types.nim
@@ -94,13 +94,17 @@ proc isPureObject*(typ: PType): bool =
     t = t[0].skipTypes(skipPtrs)
   result = t.sym != nil and sfPure in t.sym.flags
 
-proc isUnsigned*(t: PType): bool =
+func isUnsigned*(t: PType): bool {.inline.} =
   t.skipTypes(abstractInst).kind in {tyChar, tyUInt..tyUInt64}
 
 proc getOrdValue*(n: PNode; onError = high(Int128)): Int128 =
-  var k = n.kind
-  if n.typ != nil and n.typ.skipTypes(abstractInst).kind in {tyChar, tyUInt..tyUInt64}:
-    k = nkUIntLit
+  let k =
+    if n.typ.isNil:
+      n.kind
+    elif n.typ.isUnsigned:
+      nkUIntLit
+    else:
+      n.kind
 
   case k
   of nkCharLit, nkUIntLit..nkUInt64Lit:
@@ -423,6 +427,14 @@ template isResolvedUserTypeClass*(t: PType): bool =
   tfResolved in t.flags
 
 proc firstOrd*(conf: ConfigRef; t: PType): Int128 =
+  ## computes the first ordinal value of a concrete `t`ype, taking into account:
+  ## - `int` bit-width via the `conf` param (a `nil` `conf` assumes 64 bits)
+  ## - the ordinal value set of the first enum element
+  # xxx: consider how/whether to handle inband reporting for `tyError` (aka
+  #      `tyProxy`), and potentially empty vector types like `tyCstring` and
+  #      `tyUncheckedArray`
+  # Note: `tyInt` and `tyUint`'s platform specifics, and therefore the `conf`
+  #       param, are here because of things like `sizeof` evaluation during sem
   case t.kind
   of tyBool, tyChar, tySequence, tyOpenArray, tyString, tyVarargs, tyProxy:
     result = Zero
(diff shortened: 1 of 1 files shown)


compiler/ast/types.nim
```diff
@@ -443,49 +449,29 @@ proc firstOrd*(conf: ConfigRef; t: PType): Int128 =
...
      tyStatic, tyInferred, tyUserTypeClasses, tyLent:
     result = firstOrd(conf, lastSon(t))
   of tyOrdinal:
-    if t.len > 0: result = firstOrd(conf, lastSon(t))
-    else:
-      conf.localReport InternalReport(
-        kind: rintUnreachable,
-        msg: "invalid kind for firstOrd(" & $t.kind & ')')
-
+    # TODO: convert to an invalid return and have callee handle it
```
zerbina: Did you mean `caller` instead of `callee` here?

compiler/ast/types.nim
```diff
@@ -443,49 +449,29 @@ proc firstOrd*(conf: ConfigRef; t: PType): Int128 =
...
+    conf.internalAssert(t.len > 0, "firstOrd - abstract ordinal type given")
+    result = firstOrd(conf, lastSon(t))
   of tyUncheckedArray, tyCstring:
     result = Zero
   else:
-    conf.localReport InternalReport(
-      kind: rintUnreachable,
-      msg: "invalid kind for firstOrd(" & $t.kind & ')')
+    conf.internalError("firstOrd - non-ordinal type given: " & $t.kind)
```
zerbina: I think an `unreachable` would fit better here, but I'm also okay with leaving the `internalError`.

compiler/ast/types.nim
```diff
@@ -513,27 +499,52 @@ proc lastOrd*(conf: ConfigRef; t: PType): Int128 =
...
+        assert(t.n[^1].kind == nkSym)
+        toInt128(t.n[^1].sym.position)
+      else:
+        # currently empty enums aren't support as a type, but they should be
+        Zero
   of tyGenericInst, tyDistinct, tyTypeDesc, tyAlias, tySink,
      tyStatic, tyInferred, tyUserTypeClasses, tyLent:
     result = lastOrd(conf, lastSon(t))
-  of tyProxy: result = Zero
+  of tyProxy: result = Zero # xxx: this seems off; also not in `firstOrd`
```
zerbina: Given that `tyProxy` is only used for errors now, I think this should be handled as part of the `else` branch (i.e. treated as unreachable).

Adjusting it doesn't have to happen as part of this PR, though.

compiler/ast/types.nim
```diff
@@ -513,27 +499,52 @@ proc lastOrd*(conf: ConfigRef; t: PType): Int128 =
...
   of tyOrdinal:
-    if t.len > 0: result = lastOrd(conf, lastSon(t))
-    else:
-      conf.localReport InternalReport(
-        kind: rintUnreachable,
-        msg: "invalid kind for firstOrd(" & $t.kind & ')')
-  of tyUncheckedArray:
+    conf.internalAssert(t.len > 0, "lastOrd - abstract ordinal type given")
+    result = lastOrd(conf, lastSon(t))
+  of tyUncheckedArray: # xxx: what about `tyCstring`; see `firstOrd`
```
zerbina: Neither `cstring` nor `UncheckedArray` have a compile-time-known lower or upper bound, so I think both shouldn't be handled by `firstOrd`/`lastOrd`.

This is also not something I expect to happen now.

compiler/ast/types.nim
```diff
@@ -513,27 +499,52 @@ proc lastOrd*(conf: ConfigRef; t: PType): Int128 =
...
   of tyEnum:
-    if t.n.len > 0:
-      assert(t.n[^1].kind == nkSym)
-      result = toInt128(t.n[^1].sym.position)
+    result =
+      if t.n.len > 0:
+        assert(t.n[^1].kind == nkSym)
+        toInt128(t.n[^1].sym.position)
+      else:
+        # currently empty enums aren't support as a type, but they should be
```
zerbina:
```suggestion
        # currently empty enums aren't supported as a type, but they should be
```

zerbina (commented):
I left a few small suggestions and comments, but nothing that's a hard blocker.

Regarding using the return value to indicate errors, I don't think doing so is necessary. Calling `firstOrd` or `lastOrd` with a type that doesn't have compile-time-known limits hints at a logic error somewhere and, in my opinion, should abort compilation.

compiler/ast/types.nim
saem: I absolutely do, thanks

compiler/ast/types.nim
saem: I want to crash the compiler and force bug fixes + remove open reports code, I could use the idiom I suppose.

compiler/ast/types.nim
saem: I was thinking `tyProxy` branches are where we might add in forgiveness for suggest situations, while everything else is a programming error.

compiler/ast/types.nim
saem: Yeah they're pessimisticly `-1`, as in empty; unless we have some hints like it's not `nil`.

compiler/ast/types.nim
saem: Thank you, thank you.

compiler/ast/types.nim
saem: ultimately removed the comment as it should simply be a failure.

compiler/ast/types.nim
saem: Captured the above via comment.

compiler/ast/types.nim
saem: captured the above as a comment; even if somewhat oblique.

zerbina (approved):

PR #592, opened 2023-03-19
vm: fix multiple `raise` and `finally` issues

zerbina:
## Summary

Fix multiple issues with `raise`-related stack unwinding and resuming after `finally` clauses. In addition, prepare for further `finally` fixes.

## Details

When leaving a `finally` section, resuming the stack unwinding started by a `raise` is now handled directly instead of via the indirect back- and-forth mechanism used previously.

As a consequence, cactus-like stack shapes are now no longer possible, which simplifies all logic related to traversing the stack and also makes future VM changes that expect a "real" stack possible. The top-of- stack also no longer has to be tracked, but is now known to always be the last element int the stack-frame sequence.

In addition, the saved program counter (which is now only used during frame cleanup) is stored separately for each stack-frame. This is necessary for the saved counter to be preserved across calls to `rawExecute`, and is also a small step towards `finally` clauses working as expected in the VM: raising and handling an exception or entering a `finally` clause no longer abort clean-up in the frame above.

With both changes together, VM yields no longer interrupt unwinding or frame cleanup.

### Misc

- support traces with the `vmrunner` again

<!--
Pull Request(PR) Help

Tips that make PRs easier:

### DIFF
--- a/compiler/vm/compilerbridge.nim
+++ b/compiler/vm/compilerbridge.nim
@@ -204,11 +204,44 @@ proc unpackResult(res: sink ExecutionResult; config: ConfigRef, node: PNode): PN
 
     result = config.newError(node, astDiag, instLoc(-1))
 
+proc createStackTrace(c: TCtx, raw: VmRawStackTrace;
+                      recursionLimit: int = 100): VmStackTrace =
+  ## Creates a compiler-facing stacktrace from the internal stacktrace `raw`.
+  result = VmStackTrace(currentExceptionA: nil, currentExceptionB: nil)
+
+  var count = 0
+  # create the stacktrace entries:
+  for i in countdown(raw.high, 0):
+    let f {.cursor.} = raw[i]
+
+    if count < recursionLimit - 1 or i == 0:
+      # The `i == 0` is to make sure that we're always including the bottom of
+      # the stack (the entry procedure) in the trace
+
+      # Since we're walking the stack from top to bottom, the elements are
+      # added to the trace in reverse order (the most recent procedure is
+      # first in the list, not last). This needs to be accounted for by the
+      # actual reporting logic
+      result.stacktrace.add((sym: f.sym, location: c.debug[f.pc]))
+
+    inc count
+
+  if count > recursionLimit:
+    result.skipped = count - recursionLimit
+
+  assert result.stacktrace.len() <= recursionLimit # post condition check
+
 proc buildError(c: TCtx, thread: VmThread, event: sink VmEvent): ExecErrorReport  =
   ## Creates an `ExecErrorReport` with the `event` and a stack-trace for
   ## `thread`
+  let stackTrace =
+    if event.kind == vmEvtUnhandledException:
+      createStackTrace(c, event.trace)
+    else:
+      createStackTrace(c, thread)
+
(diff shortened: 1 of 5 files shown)


saem (approved):
Bummer to lose cactus stacks but that's just me romanticizing things; first priority is getting things working consistently.

With the change I'm guessing there are a number of tests that now likely pass, I'd recommend the following:
- rename `tests/exception` to `tests/lang_exceptions`; and update `testament.defaultTargets` to add the vm and js targets, as well
- `tests/excpetion/texception_name_vm_issue`  likely works and can be enabled (removing `knownIssue` from the spec)
- add any exclusions as required

zerbina:
> With the change I'm guessing there are a number of tests that now likely pass

For the largest part, the change affects code that causes the VM to yield. For example:
```nim
proc p() =
  try:
    return
  finally:
    echo "a"
  echo "b"
```
would previously echo "ab", but now correctly echoes only "a".

Running the `exception` category (with all eligible tests enabled for the VM) yields only two more tests that work now compared to before. In general, `finally` handling is still very much broken for the VM.

> * `tests/excpetion/texception_name_vm_issue`  likely works and can be enabled (removing `knownIssue` from the spec)

The failure of the `texception_name_vm_issue` test is not because of control-flow, so it's not affected by this change.

Given the only minor improvement in passing tests, I'd make a separate PR for the test category cleanup and merge it first, though, if preferred, I'm also okay with merging it after this PR. 

---

Looking over the changes again, I noticed that there's a regression with the PR as it is now: stack-traces for unhandled exception are incorrect in the presence of finalizers, as the frame and its ancestors from which the exception was originally raised are now already discarded once the VM exits.

To fix this, I think it makes sense to generate the stack-trace when raising the exception and then later attaching it to the "unhandled exception" event.

saem:
> Running the `exception` category (with all eligible tests enabled for the VM) yields only two more tests that work now compared to before. In general, `finally` handling is still very much broken for the VM..
>
> > * `tests/excpetion/texception_name_vm_issue`  likely works and can be enabled (removing `knownIssue` from the spec)
> 
> The failure of the `texception_name_vm_issue` test is not because of control-flow, so it's not affected by this change.
> 
> Given the only minor improvement in passing tests, I'd make a separate PR for the test category cleanup and merge it first, though, if preferred, I'm also okay with merging it after this PR.

Sorry, I was a bit optimistic and assumed with this and recent changes it was pretty close and it was a good time to enable it -- separate PR it is.

saem:
bors cancel

saem:
sorry forgot about the regression, in case you wanted to address that as part of this PR.

zerbina:
I noticed another regression with the change: an active exception took precedence over other unstructured control-flow at the end of a `finally`, which is incorrect. The original behaviour, while also not correct, is now restored.

As mentioned, the whole `finally` handling needs to be reworked at one point. My plan is to make it fully goto-based, with a register-based save/restore mechanism for efficient `finally` support.

---

In order for the unhandled exception to show the correct stack-trace, the VM now creates the stack-trace internally and passes it on the event when reporting an unhandled exception.

@saem: If you think the diagnostics-related adjustments are fine, feel free to merge.

saem (approved):
bors r+

PR #593, opened 2023-03-20
sem: integer literal to holey enum error detection

saem:
## Summary
Previously range error detection for integer literal to holey enums were in `semfold`, this is now done in `semConv`.

## Details

This simplifies `semfold` and moves this logic to the earliest point when we have enough information to make the correct judgement. Added some tests to avoid regressions.

### Misc

Aside, noticed `lang_def` vs `lang_types` and that doesn't make sense, will fix in a separate PR.

---

## Notes for Reviewers
* one of many issues found while working on `semfold`/`semCase`

### DIFF
--- a/compiler/sem/semfold.nim
+++ b/compiler/sem/semfold.nim
@@ -35,6 +35,7 @@ import
   ],
   compiler/utils/[
     platform,
+    idioms,
   ]
 
 from compiler/ast/reports_sem import SemReport,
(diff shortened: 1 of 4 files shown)


saem:
Hmm, took a quick look at the runnable examples failure. It seems that it took the buggy behavior (now fixed), wrote more buggy stuff on top of it, and then passed all that of as a good idea -- Imma going to rip that stuff out. About the only thing we could do is a runtime conversion for non-literals which should raise out of range errors.

compiler/sem/semfold.nim
```diff
@@ -109,29 +110,35 @@ proc foldMul(a, b: Int128, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNode
...
+            return field.ast.strVal
+      if n.len > 0:
+        unreachable("range checking `semConv` eliminates this case")
+    else:
+      result = $x
+  of nkError:
+    unreachable("an error cannot make it here")
   else:
-    result = $x
+    unreachable("non-ordinals never make it hear")
```
saem:
```suggestion
    unreachable("non-ordinals never make it here)
```

PR #594, opened 2023-03-20
semexprs: make `semAsgn` error node aware

zerbina:
## Summary

- make `asgnToResultVar` `nkError` aware
- remove input AST mutations from  `semAsgn`
- replace `localReport`s with `nkError` usage in `semAsgn`
- make `semAsgn` propagate `nkError`s

In addition, the `semIterator`, `semMethod`, and `semMacroDef` now return early when `semProcAux` returns an error, in which they would previously have crashed.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -2140,12 +2140,18 @@ proc takeImplicitAddr(c: PContext, formal: PType, n: PNode): PNode =
   else:
     result = newTreeIT(nkHiddenAddr, n.info, formal): n
 
-proc asgnToResultVar(c: PContext, n, le, ri: PNode) {.inline.} =
-  # refactor: to return a PNode instead and handle nkError
-  
-  if n.isError:
+proc asgnToResultVar(c: PContext, n: PNode): PNode {.inline.} =
+  ## If the assignment statement `n` is that of an assignmnet to a ``var``
+  ## result, rewrites it to be a view assignment. If an error occurs, an
+  ## ``nkError`` node is returned -- the rewritten `n` otherwise.
+  result = n
+  if result.kind == nkError:
     return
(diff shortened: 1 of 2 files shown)


compiler/sem/semexprs.nim
```diff
@@ -2193,26 +2211,26 @@ proc semAsgn(c: PContext, n: PNode; mode=asgnNormal): PNode =
...
-      n[1] = fitNode(c, le, rhs, goodLineInfo(n[1]))
-      when false: liftTypeBoundOps(c, lhs.typ, lhs.info)
-
-      result = fixAbstractType(c, n)
-      asgnToResultVar(c, result, n[0], n[1])
-  result = n
+            result[1] = typeMismatch(c.config, n.info, lhs.typ, rhsTyp, rhs)
+            hasError = true
+
+      if not hasError:
```
saem, line 2294: The old code and diff are a bit tricky to follow, but I _believe_ this has to guard on `if result[1].kind != nkError`, because otherwise we cut the rhs analysis short and that _might_ be why CI is failing. Just a guess.

saem (approved):
Looks good to me.

I read through the code to see if I could spot where the CI errors are originating from, I have one guess which I left as a comment.

zerbina:
Thanks for taking a look @saem. The tests are failing for three reasons:
1. I missed adjusting the `n.typ = c.enforceVoidContext` line, so assignments to `result` no longer enforce a void context
2. `nkError` nodes in routine bodies are reported twice, once by `trackProc` and once by `trackStmt` (I'll make a fix in a separate PR)
3. `semReturn` is not `nkError` aware (fixed by the PR that's already up)

PR #595, opened 2023-03-21
convert diag to legacy report in `cli_reporter`

saem:
## Summary

Moved conversion to `cli_reporter` from `msgs` to allow for querying
data during transforms without dependency cycles. This allows
diagnostics to remain minimal without inheriting report data type flaws.

## Details

The reports issues continue: now when creating a `ConfigRef` one must
set the `astDiagToLegacyReport` field to the conversion procedure. This is
required until we can replace reports.

Reused the `cli_reporter` module because I didn't want to create yet
another reports module that needs to be eventually removed.

---

## Notes for Reviewers
* noticed while working on removing `reports` from `semfold`, querying at transformation would simplify diagnostics


zerbina (approved):
I think this is definitely a step in the right direction, i.e. lean diagnostics and querying for additional context in the rendering or transformation layer.

The callback on the `ConfigRef` is also fine, I think. Callbacks are generally a helpful tool for the reporting refactoring and `localReport` removal, though we need to watch out that the architecture around diagnostic reporting doesn't become over-reliant on them.

bors r+

saem:
Thanks for the review, sharing your thoughts, and seconding the direction, @zerbina

PR #596, opened 2023-03-21
semexprs: make `semReturn` error node aware

zerbina:
## Summary

- refactor the `semReturn` procedure to not use `return` and to not modify the input AST
- propagate `nkError` in `semReturn`
- replace `localReport`s with `nkError` node usage

In addition, a `(ReturnStmt (Asgn (...) (...)))` where the assignment is not that of the `result` variable now results in an `invalid expression` error, instead of an `expression has no type` error.

---

## Notes for Reviewers
* this is requirement for #594
* the code is a bit longer now, though I think it reads better without the return

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


saem (approved):
It does read better than before!

bors r+

PR #597, opened 2023-03-21
`semfold` uses `nkError` dropping  'reports'

saem:
## Summary

The `semfold` module now issues `nkError` instead of reports, this
simplifies the code as everything is more expression oriented.

## Details

There remains one indirect dependency on reports, via the calls to
`msg.internalError`, though they should be easy to remove soon.

Additionally, `internalError` and `internalAssert` in the `msgs` module
now hide their reports dependency, ensuring that callers in modules no
longer incur a direct reports dependency.

---

## Notes for Reviewers
* should open up work on `semCase` refactor
* wondering if `getConstExpr` should return `nil` for errors, but that could be a separate PR

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1538,9 +1538,6 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemExpectedObjectForOf:
       result = "'of' takes object types"
 
-    of rsemSemfoldDivByZero:
-      result = "over- or underflow"
-
     of rsemRuntimeDiscriminantRequiresElif:
       result = "branch initialization with a runtime discriminator " &
         "is not supported inside of an `elif` branch."
@@ -1722,6 +1719,9 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemSemfoldOverflow:
       result = "over- or underflow"
 
+    of rsemSemfoldDivByZero:
+      result = "division by zero"
+
     of rsemCaseInUnion:
       result = "Illegal use of ``case`` in union type."
 
@@ -1764,7 +1764,7 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       result = "len($1) must be less than 32768" % r.symstr
 
     of rsemCantConvertLiteralToRange:
-      result = "cannot convert " & $r.str & " to " & r.typ.render
+      result = "cannot convert $# to $#" % [r.wrongNode.render, $r.ast.typ.render]
 
     of rsemCantComputeOffsetof:
       result = "can't compute offsetof on this ast"
@@ -3739,7 +3739,7 @@ proc rotatedTrace(conf: ConfigRef, r: Report) =
     conf.incl optUseColors
 
 
-func astDiagToLegacyReport(diag: PAstDiag): Report {.inline.} =
+func astDiagToLegacyReport(conf: ConfigRef, diag: PAstDiag): Report {.inline.} =
   ## legacy because it converts a diag to a Sem/VM report wrapped in a Report.
   ## Why, you might ask? Well that's because reports are a mess and it'll take
   ## many messy intermediate steps to get out of this quagmire. So at present
@@ -3874,12 +3874,23 @@ func astDiagToLegacyReport(diag: PAstDiag): Report {.inline.} =
       adSemFieldAssignmentInvalidNeedSpace,
       adSemFieldAssignmentInvalid,
       adSemObjectConstructorIncorrect,
-      adSemExpectedObjectType:
+      adSemExpectedObjectType,
+      adSemFoldOverflow,
+      adSemFoldDivByZero,
+      adSemFoldCannotComputeOffset:
     semRep = SemReport(
         location: some diag.location,
         reportInst: diag.instLoc.toReportLineInfo,
         kind: kind,
         ast: diag.wrongNode)
+  of adSemInvalidRangeConversion:
+    semRep = SemReport(
+        location: some diag.location,
+        reportInst: diag.instLoc.toReportLineInfo,
+        kind: rsemSemfoldInvalidConversion,
+        ast: diag.wrongNode,
+        typeMismatch: @[typeMismatch(diag.wrongNode[0].typ,
+                                     diag.wrongNode.typ)])
   of adSemUseOrDiscardExpr:
     semRep = SemReport(
         location: some diag.undiscarded.info, # xxx: location override
@@ -4179,6 +4190,25 @@ func astDiagToLegacyReport(diag: PAstDiag): Report {.inline.} =
       kind: rsemCannotCastToNonConcrete,
       ast: diag.wrongNode,
       typ: diag.wrongType)
+  of adSemFoldRangeCheckForLiteralConversionFailed:
+    semRep = SemReport(
+      location: some diag.location,
+      reportInst: diag.instLoc.toReportLineInfo,
+      kind: rsemCantConvertLiteralToRange,
+      ast: diag.wrongNode,              # xxx: ignores `diag.wrongNode`
+      wrongNode: diag.inputLit)
+  of adSemIndexOutOfBoundsStatic:
+    semRep = SemReport(
+      location: some diag.location,
+      reportInst: diag.instLoc.toReportLineInfo,
+      kind: rsemStaticOutOfBounds,
+      ast: diag.wrongNode,
+      indexSpec: (usedIdx: getInt(diag.staticIndex),
+                  minIdx: if diag.staticCollection.kind == nkBracket:
+                            firstOrd(conf, diag.staticCollection.typ)
+                          else:
+                            toInt128(0),
+                  maxIdx: toInt128(diag.staticCollection.len - 1)))
   of adSemMagicExpectTypeOrValue:
     semRep = SemReport(
       location: some diag.location,
@@ -4194,7 +4224,8 @@ func astDiagToLegacyReport(diag: PAstDiag): Report {.inline.} =
       ast: diag.wrongNode,
       typ: diag.invalidTyp,
       str: magicToStr(diag.highLow))
-  of adSemUnknownIdentifier:
+  of adSemUnknownIdentifier,
+      adSemStaticFieldNotFound:
     semRep = SemReport(
       location: some diag.location,
       reportInst: diag.instLoc.toReportLineInfo,
@@ -4321,6 +4352,14 @@ func astDiagToLegacyReport(diag: PAstDiag): Report {.inline.} =
       kind: kind,
       ast: diag.wrongNode,
       typ: diag.expectedObjTyp)
+  of adSemInvalidIntDefine,
+      adSemInvalidBoolDefine:
+    semRep = SemReport(
+      location: some diag.location,
+      reportInst: diag.instLoc.toReportLineInfo,
+      kind: kind,
+      ast: diag.wrongNode,
+      str: diag.invalidDef)
   of adVmError:
     let
       kind = diag.vmErr.kind.astDiagVmToLegacyReportKind()
@@ -4475,7 +4514,7 @@ proc legacyReportBridge*(conf: ConfigRef, diag: PAstDiag): Report =
   # Not the ideal place for this proc, but I'd rather keep the reports mess in
   # as few files as possible
   # TODO: will use conf soon
-  astDiagToLegacyReport(diag)
+  astDiagToLegacyReport(conf, diag)
 
 proc reportHook*(conf: ConfigRef, r: Report): TErrorHandling =
   ## Default implementation of the report hook. Dispatches into
--- a/compiler/sem/semfold.nim
+++ b/compiler/sem/semfold.nim
@@ -23,6 +23,7 @@ import
     ast,
     trees,
     lineinfos,
+    errorhandling,
   ],
   compiler/modules/[
     magicsys,
@@ -30,7 +31,6 @@ import
   ],
   compiler/front/[
     commands,
-    msgs,
     options,
   ],
   compiler/utils/[
@@ -38,11 +38,8 @@ import
     idioms,
   ]
 
-from compiler/ast/reports_sem import SemReport,
-  reportAst,
-  reportStr,
-  reportTyp
-from compiler/ast/report_enums import ReportKind
+from compiler/front/msgs import internalError # xxx: legacy reports cruft is
+                                              #      hiding in here
 
 from system/memory import nimCStrLen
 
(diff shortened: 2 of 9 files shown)


compiler/sem/semfold.nim
```diff
@@ -462,21 +470,18 @@ proc getArrayConstr(m: PSym, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNod
     if result == nil: result = n
 
 proc foldArrayAccess(m: PSym, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNode =
-  var x = getConstExpr(m, n[0], idgen, g)
-  if x == nil or x.typ.skipTypes({tyGenericInst, tyAlias, tySink}).kind == tyTypeDesc:
+  let x = getConstExpr(m, n[0], idgen, g)
+  if x.isNil or
+      x.typ.skipTypes({tyGenericInst, tyAlias, tySink}).kind == tyTypeDesc:
```
zerbina, line 506: It's not something for this PR, but `{tyGenericInst, tyAlias, tySink}` is a very common set -- it should probably be put into a constant (though I'm not yet sure what name would fit).

compiler/sem/semfold.nim
```diff
@@ -620,6 +627,7 @@ proc getConstExpr(m: PSym, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNode
       case s.magic
       of mNone:
         # If it has no sideEffect, it should be evaluated. But not here.
+        # QUESTION: the right place to run it is the VM?
```
zerbina: Yes, via the `evalAtCompileTime` machinery.

compiler/sem/semfold.nim
```diff
@@ -730,32 +739,29 @@ proc getConstExpr(m: PSym, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNode
...
-    var a = getConstExpr(m, n[1], idgen, g)
+    let a = getConstExpr(m, n[1], idgen, g)
     if a == nil: return
     result = foldConv(n, a, idgen, g, check=true)
   of nkDerefExpr, nkHiddenDeref:
     let a = getConstExpr(m, n[0], idgen, g)
     if a != nil and a.kind == nkNilLit:
-      result = nil
-      #localReport(g.config, n.info, "nil dereference is not allowed")
+      result = nil   # xxx: why not return a nil lit instead?
```
zerbina: That wouldn't work. Returning a `nkNilLit` would mean that `(nil)[]` is treated as valid, and it shouldn't.

Returning an error is also not possible, since the expression might never be evaluated at run-time.

compiler/sem/semfold.nim
```diff
@@ -774,4 +780,4 @@ proc getConstExpr(m: PSym, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNode
     if i == n.len - 1:
       result = getConstExpr(m, n[i], idgen, g)
   else:
-    discard
+    discard # xxx: should we return nil for nkError?
```
zerbina, line 809: `nil`, because it's only an expression or statement that can't be folded (which isn't an error in itself). E.g. an assignment.

compiler/sem/semfold.nim
```diff
@@ -562,30 +566,33 @@ proc getConstExpr(m: PSym, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNode
...
-        if isDefined(g.config, s.name.s):
-          try:
-            result = newIntNodeT(toInt128(g.config.symbols[s.name.s].parseInt), n, idgen, g)
-          except ValueError:
-            g.config.localReport reportStr(
-              rsemInvalidIntdefine, g.config.symbols[s.name.s])
-
-        else:
-          result = copyTree(s.ast)
+        result =
```
zerbina: Given the similarity of the three branches, I think it would make sense to consolidate them into a single one, like so:
```nim
of mIntDefine, mStrDefine, mBoolDefine:
  result =
    if isDefined(g.config, s.name.s):
      let define = g.config.symbols[s.name.s]
      case s.magic
      of mIntDefine: ...
```
That's one more level of nesting, however, which is a bit meh. No change is expected from my side.

As an aside (nothing for this PR), resolving the `define` constants during folding feels off. Replacing the constant's value in `semNormalizedConst` would be better, I think. Doing so, would make retrieving the original value impossible, although I don't think that'd would be problematic.

compiler/front/cli_reporter.nim
```diff
@@ -4179,6 +4190,25 @@ func astDiagToLegacyReport(diag: PAstDiag): Report {.inline.} =
       kind: rsemCannotCastToNonConcrete,
       ast: diag.wrongNode,
       typ: diag.wrongType)
+  of adSemFoldRangeCheckForLiteralConversionFailed:
+    semRep = SemReport(
+      location: some diag.location,
+      reportInst: diag.instLoc.toReportLineInfo,
+      kind: rsemCantConvertLiteralToRange,
+      ast: diag.wrongNode,              # xxx: ignores `diag.wrongNode`
```
zerbina, line 4198: What is the comment referring to?

zerbina (approved):
Very nice, thank you! Yet another step towards the removal of `localReport`.

I answered a few questions and left some suggestion, but none of them are blocking.

compiler/front/cli_reporter.nim
saem: Ah, the report ignores the `ast` field, which is set to `diag.wrongNode`.

zerbina:
Hmm, with `getConstExpr` now being able to return an `nkError`, the various places where `semfold` calls `getConstExpr` itself (such as `magicCall`) now need to short-circuit on encountering an error.

compiler/sem/semfold.nim
saem: Hmm, I guess what's weird is that if we got an `nkNilLit` back, then this should in fact return an error (or we shouldn't have received an `nkNilLit`). I'm ok to leave it as is for now, as you've also indicated, there are going to be lots of fixes in this module and especially around `getConstExpr` callsites.

Hilariously, the error would be something like, `adSemConstDerefNilPointerException` 😆; I suspect in many cases this would be something we would recover from by aborting folding, but not necessarily always.

saem:
> Hmm, with `getConstExpr` now being able to return an `nkError`, the various places where `semfold` calls `getConstExpr` itself (such as `magicCall`) now need to short-circuit on encountering an error.

I fixed one and (`sempass2`) and bootstrap, tests, tools, testTools, and docs all came out clean. 🤞🏽 I'm sure there are still gaps, but between the code base become more `nkError` + the refactorings happening which are reducing the number of places we do things (including where we fold), we should ferret the rest out if we encounter them.

I'd rather a more exhaustive approach, but there are _many_ callsites and as I looked through them, it seemed like various layering violations made it hard to tell if we should emit the errors (likely most cases) vs consider it folding failure (likely rare).

saem:
Casual observation, might be interesting to revisit later:

This refactoring surfaced a potential pattern, but I didn't want to act on it just yet. `semfold` as the name indicates is effectively a fold or reduction on AST, ie: `AST -> Result[Option[AST], Error]`. The intention of that type is: if there is no error, then we might have an AST, where presence suggests a reduction and absence meaning it's not a constant expression.

Anyhow, besides the fact that this overlaps very strongly with partial evaluation (topic for another day), the diagnostics are likely best modeled as: `(failedAction: ActionEnum, inputAst: PNode, outputAst: PNode)`, where:
- `failedAction` has a near 1-to-1 correspondence with folding procs in `semfold`
- `outputAst` is the fold accumulator
- `failedAction: ActionEnum` value we should know what parts of the `input/outputAst` to query and in what manner in for emitting diagnostics

tl;dr: there might be an entirely uniform data representation for `semfold` errors.

compiler/sem/semfold.nim
saem: Strongly agree. I also think it's worth figuring out if we can straight-up replace the first two with flags.

With `sink` I do think we're looking at a type/kind, `sink[T]/sink`, but I don't buy that with `alias` or `genericInst`.

compiler/sem/semfold.nim
saem: Thanks, I'll update in a future PR.

I wrote it partially in a, "seriously, did we have to leave that vague?" to the original comment; but it's too dry.

compiler/sem/semfold.nim
zerbina: Receiving an `nkNilLit` from folding is okay, because while a `nkNilLit` can't be folded itself (it's atomic), it's a node that the rest of folding can reason about and thus shouldn't block.

Having thought about the whole topic a bit more since my original comment, returning an error here would be fine, as long as it's later discarded when the folding is happening in the context of high-level optimization (i.e. invoked through `hlo -> commonOptimizations`).

compiler/sem/semfold.nim
```diff
@@ -376,19 +378,42 @@ proc leValueConv*(a, b: PNode): bool =
...
+    a = getConstExpr(m, n[1], idgen, g)
+    b = if n.len > 2: getConstExpr(m, n[2], idgen, g) else: nil
+    c = if n.len > 3: getConstExpr(m, n[3], idgen, g) else: nil
+
+  template makeError(toCopy, a: PNode; b, c: PNode = nil) =
+    result = copyNode(toCopy)
+    result.flags = toCopy.flags # make an "exact copy"
+    result.add a
+    if b != nil: result.add b
+    if c != nil: result.add c
```
zerbina:
```suggestion
    result = shallowCopy(toCopy)
    result.flags = toCopy.flags # make an "exact" copy
    result[0] = toCopy[0]
    result[1] = toCopy[1]
    if toCopy.len > 2:
      result[2] = if b != nil: b else: toCopy[2]
    if toCopy.len > 3:
      result[3] = if c != nil: c else: toCopy[3]
```
This keeps the original shape intact.

zerbina (commented):
There's a small issues with the `makeError` template.

compiler/sem/semfold.nim
saem: Yeah, the define registration feels a bit off here. I think one question to answer is what do these defines really mean? Are they truly the equivalent of the user writing something like:

```nim
const foo = if defined(foo): getDefine(foo).parseInt else: 10
```

If so, then I think it's fair to say that the if expression is a subroutine (template) and so:
- we set the value in `semNormalizedConst`
- remember the default via template expansion (or a specialized equivalent)

compiler/sem/semfold.nim
saem: Good catch, will fix now.

compiler/sem/semfold.nim
saem: OK, I think it's a bigger refactor (separate PR). I wrote out what I thought was needed, but it doesn't work all that well.

For the bigger refactor, I think we need two procs:
1. reduce some AST as much as possible and also tell me if it's constant at the end
2. reduce to a constant `PNode/nil`; or `Option[PNode]`

At that point I think all the various use cases can be served.

----

What follows was my original thinking, but it's easier said than done, especially around maintaining fold/copy guarantees.

I think I need to:
- create a proc like `foldConstExpr`, which does most of what `getConstExpr` does today
- `getConstExpr` will check input params, delegate to `foldConstExpr`, and check the output before returning

The new `getConstExpr` would roughly be:

```nim
proc foldConstExpr*(m: PSym, n: PNode, ...): PNode =
  ## old `getConstExpr`, but never returns `nil`
  ## copies nodes iff AST is folded (non-identity)
  ##   - corollary: copies due to folding in a sub-tree cascade copy up to the parent
  ## `nkError` pass-through, except if in an _input_ sub-tree position
  ##    - ie. `nkCallKinds` input args must never start as nkError, but may produce them when folding
  # ... implementation ...

proc getConstExpr*(m: PSym, n: PNode, ...): PNode =
  assert n != nil
  case n.kind
  of nkError: n
  else:
    result = foldConstExpr(m, n, ...)
    if result.id == n.id: # nothing changed; foldExpr only mutates iff required
      result = nil
    else:
      # check to see if it's an atomic value, could be done via
      # - an `nonConst: var bool` param that's passed along and set when any node is not atomic
      # - ensure all of results immediate children have ids > `ast.gNodeId` before the fold started (copies)
```

Now `foldConstExpr` can take arbitrary, error free, AST and reduce it wherever constant expressions might exist, and `getConstExpr` works to only

compiler/sem/semfold.nim
saem: Thinking about it a bit more, there are two situations involving `nkError`:
1. `nkError` is in the input, unless it's the root of the AST (base call of `getConstExpr`), this is probably an input violation because the caller failed to wrap the tree, but folding this shoud result in a `nil`
2. `nkError` is produced during a folding operation, these should be propagated and emitted

If a caller is speculatively folding via placeholder values and an error is generated, then the caller should discard these. Speaking of which, it's not entirely safe to do speculative folding over permuations of the same tree, because things like defines are effectful (practically it's rarely a problem, but hard to trust the code).

compiler/sem/semfold.nim
saem: Fixed in the latest -- it is squashed.

compiler/sem/semfold.nim
saem: I updated the code style-wise; leaving the bigger what is a an `intdefine` etc as out of scope for this PR.

Won't resolve conversation for visilbity.

compiler/sem/semfold.nim
saem: Given my above write-up, I think the current behaviour is correct, as all other occurances can be checked prior to dispatch to `getConstExpr` on parts of the sub-tree.

compiler/sem/semfold.nim
saem: Left a note in the code saying that we should return `nkNilLit` and fix the code that's creating the incorrect deref.

compiler/sem/semfold.nim
saem: updated the comment so it's less vague.

compiler/sem/semfold.nim
zerbina: Hm, why is the deref incorrect? In the following code:
```nim
const x = SomeRef(nil)[] # should fail with a nil dereference error, either during folding, or, like it does now, during VM execution

# `y` is replaced with `nkNilLit` during inlining:
if someCond:
  y
else:
  y[].other # should not lead to a compile-time error, except if executed at compile-time

```

there are no incorrect derefs, and folding either of the `DerefExpr (NilLit)` expressions to just `NilLit` would result in semantically invalid code.

saem: Sorry about that, I was too hasty and what I wrote is not even close to what I was trying to convey. I'm pretty sure we're on the same page.

In case of `nkNiLit`, we should lift (not return) it and then produce an error because it's incorrect to dereference a `nil` -- example 1. Where `y` is speculatively replaced with `nkNilLit`, we should somehow inform `getConstExpr` to ignore certain `nkNilLit` nodes, and that should cover example 2 (inlining).

The parts where I meant we're generating incorrect derefs is focused on `nkHiddenDeref`. IIRC the analysis is not 100% and so I jumped to the conclusion we're misapplying them and a reason we end up in that code path.

Sorry for communicating all that poorly, thanks for the feedback and examples -- I'll updated the comment in a future PR, need the rest of the changes to make further progress.

PR #598, opened 2023-03-22
sem: report an error for calls to erroneous routines

zerbina:
## Summary

If overload resolution picked the symbol of a routine whose definition (header or body) contains an error, the call expression was passed on, resulting in AST containing `nkError` nodes to possibly `transf` and `mirgen` (in the case of macro evaluation), crashing the compiler.

After overload resolution, `semOverloadedCall` now produces an `nkError` if the callee's definition contains an error (identified by the `TSym.ast` field containing an `nkError` node).

## Details

- update the `TSym.ast` of the produced symbol with the error AST in `semProcAux`, `semTemplateDef`, and `generateInstance`
- short-circuit `trackProc` if the body has an error. This fixes some compiler crashes due to errors reaching logic that they shouldn't, and also fixes errors in routine bodies being reported twice

There still exists logic that doesn't resolve macros and templates through `semOverloadedCall` call. In order to help with finding these, `nkError` nodes reaching `transf` or template expansion are now treated as an internal error.

### Misc

- fix a field-access defect in `semCustomPragma` that happened when `semOverloadedCall` returns an error

---

## Notes for Reviewers
* this change unblocks #590
* the short-circuiting in `trackProc` fixes an issue that blocks #594 

<!--
Pull Request(PR) Help

Before Merge Ensure:
* title reads like a short changelog line entry
* code includes tests and is documented
* leave the source better than before, but split out big reformats

See contributor (guide)[https://nim-works.github.io/nimskull/contributing.html]
for details, especially if you're new to this project.

Tips that make PRs easier:
* for big/impactful changes, start with chat/discussions to refine ideas
* refine the pull request message over time; don't have to nail it in one go
* handle the single commit message requirement at the end of review

PR #599, opened 2023-03-23
`semStaticStmt` now produces `nkError`

saem:
## Summary

Semantic analysis for static statement blocks now produce `nkError` upon
failure instead of relying on legacy reporting.

## Details

As part of this change the VM `compilerbridge` now returns the resultant
`PNode` with possible errors when calling `evalStaticStmt` instead of
directly reportting; further reducing dependence on legacy reports.

Last but not least, `semStaticStmt` previously always attempted static
evaluation of the input AST, even when initial semantic analysis of the
input produced errors. This is no longer the case and instead it returns
the produced error early. This should both further prevent `transf` from
receiving `nkError` nodes, but also speed up that code path.

---

## Notes for Reviewers
* kept it as small as possible

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -4502,6 +4502,24 @@ func astDiagToLegacyReport(conf: ConfigRef, diag: PAstDiag): Report {.inline.} =
   result =
     case kind
     of repVMKinds:
+      if diag.kind == adVmGenError and diag.duringJit or
+          diag.kind == adVmError:
+        let st =
+          case diag.kind
+          of adVmGenError: diag.vmGenTrace
+          of adVmError:    diag.vmTrace
+          else:            unreachable()
+
+        {.cast(noSideEffect).}: # side-effect analysis fails
+          vmRep.trace =
+            (ref VMReport)(kind: rvmStackTrace,
+                      currentExceptionA: st.currentExceptionA,
+                      currentExceptionB: st.currentExceptionB,
+                      stacktrace: st.stacktrace,
+                      skipped: st.skipped,
+                      location: some st.location,
+                      reportInst: toReportLineInfo(st.instLoc))
+
       Report(category: repVM, vmReport: vmRep)
     of repSemKinds:
       Report(category: repSem, semReport: semRep)
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -3119,14 +3119,22 @@ proc semStaticStmt(c: PContext, n: PNode): PNode =
   let a = semStmt(c, n[0], {})
   closeScope(c)
   dec c.p.inStaticContext
-  n[0] = a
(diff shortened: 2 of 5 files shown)


saem:
Weird, I thought I would have had to end up with a new location for the vmreport, and all the assignments from the stacktrace would be copies. Maybe one of them is a ref field or something and it's resulting in aliasing, or something. Oh well, problem for tomorrow.

compiler/front/cli_reporter.nim
```diff
@@ -4502,6 +4502,23 @@ func astDiagToLegacyReport(conf: ConfigRef, diag: PAstDiag): Report {.inline.} =
...
+          else:            unreachable()
+
+        vmRep.trace =
+            (ref VMReport)(kind: rvmStackTrace,
+                      currentExceptionA: st.currentExceptionA,
+                      currentExceptionB: st.currentExceptionB,
+                      stacktrace: st.stacktrace,
+                      skipped: st.skipped,
+                      location: some st.location,
+                      reportInst: toReportLineInfo(st.instLoc))
```
```suggestion
        {.cast(noSideEffect).}: # side-effect analysis fails
          vmRep.trace =
            (ref VMReport)(kind: rvmStackTrace,
                      currentExceptionA: st.currentExceptionA,
                      currentExceptionB: st.currentExceptionB,
                      stacktrace: st.stacktrace,
                      skipped: st.skipped,
                      location: some st.location,
                      reportInst: toReportLineInfo(st.instLoc))
```
The side-effect analysis is failing here. While I haven't tested the above, it *should* work.

compiler/sem/semstmts.nim
```diff
@@ -3119,14 +3119,19 @@ proc semStaticStmt(c: PContext, n: PNode): PNode =
...
+    result = c.config.wrapError(result)
   else:
-    result = newNodeI(nkDiscardStmt, n.info, 1)
-    result[0] = c.graph.emptyNode
+    result.transitionSonsKind(nkDiscardStmt)
+    result[0] = evalStaticStmt(c.module, c.idgen, c.graph, a, c.p.owner)
+    if result[0].kind == nkError:
+      result = c.config.wrapError(result)
+    else:
+      result[0] = c.graph.emptyNode # we presently ignore non-error output
```
```suggestion
    case result[0].kind
    of nkError:
      result = c.config.wrapError(result)
    of nkEmpty:
      result[0] = c.graph.emptyNode # we presently ignore non-error output
    else:
      unreachable()
```
It's a bit more strict and would catch unexpected output from `evalStaticStmt`.

zerbina (approved):
Thank you! This should fix one of the two test failures of #598.

compiler/front/cli_reporter.nim
saem: Thanks, I had gone around in circles trusting the analysis far too much.

compiler/sem/semstmts.nim
saem: The stricter check is much better, thanks.

Updated PR will be up shortly.

saem:
bors r+

(might have been a fluke)

PR #600, opened 2023-03-23
vmgen: separate dependency collection

zerbina:
## Summary

Move the gathering of dependencies and registering them in the link table out of code generation into a separate step. While less efficient, this greatly reduces the amount of state that code generation and serialization (`vmcompilerserdes`) need access to.

## Details

This change is a preparation for splitting up `TCtx` and for the introduction of the unified backend processing. Most of the code adjusted here will change again as part of implementing the aforementioned tasks.

### Access changes

- the code generator routines no longer mutate `TCtx.symToIndexTbl` and `TCtx.codegenInOut` (the latter is also not accessed anymore)
- `vmcompilerserdes.serialize` no longer queries nor mutates `TCtx.codegenInOut` and `TCtx.typeInfoCache`

### Changes

- add `gatherDependencies` and call it prior to all `vmgen` invocations
- manually register referenced routines in `putIntoReg`
- simplify the procedure processing in `vmbackend` -- collecting routines referenced from constants now happens as part of `gatherDependencies`, removing the need for the loop
- move `flags` from `CodegenInOut` into `TCtx`
- rename `CodegenInOut` to `LinkState`

### Misc

- remove dead code in `vmgen.genVarSection`

---

## Notes for Reviewers
* as mentioned above, this is a preparation for #550
* the change is also going to make splitting up `TCtx` easier
* flat `seq`s are eventually going to replace the usage of `Table` for the linker, which will likely offset the small slowdown incurred here

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/front/scriptconfig.nim
+++ b/compiler/front/scriptconfig.nim
@@ -71,7 +71,7 @@ proc setupVM*(module: PSym; cache: IdentCache; scriptName: string;
   result = newCtx(module, cache, graph, idgen, legacyReportsVmTracer)
   # for backwards compatibility, allow meta expressions in nimscript (this
   # matches the previous behaviour)
-  result.codegenInOut.flags = {cgfAllowMeta}
+  result.flags = {cgfAllowMeta}
   result.mode = emRepl
   registerBasicOps(result[])
   let conf = graph.config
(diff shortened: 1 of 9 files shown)


saem (approved):
bors r+

PR #602, opened 2023-03-24
sem: refactor and fix `semArrayConstr`

zerbina:
## Summary

Refactor the `semArrayConstr` procedure, make it propagate `nkError` nodes, and fix a type inference issue: if the first element in an array constructor was a `sink T` parameter, the array's element type was inferred to be `sink T`, which resulted in a `type not allowed` error.

## Details

- move array element analysis to a separate procedure
- wrap the result in `semArrayConstr` in an `nkError`, if necessary
- improve the documentation of `semArrayConstr`
- add `semRealConstExpr`, meant for expressions that *must* be constant
- remove the `TAstDiag.maxOrdIdx` field -- the value is now taken from the provided `ordRange`
- don't treat a `nkConstDef` as having no init part if the init part is erroneous. This fixes a `constant expression expected` error always replacing the actual error

### Known issue

The (pre-existing) logic for rejecting up- or down-conversion of objects in an array constructor doesn't skip `sink`, alias, and generic instance types, which is incorrect. Correctly skipping the types uncovered that a test is depending on the previous behaviour, and called into question whether disallowing object slicing in array constructors is the right semantics.

For the time being, the incorrect logic is kept as-is.

---

## Notes for reviewers
* this unblocks #598

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -696,85 +696,146 @@ proc arrayConstrType(c: PContext, n: PNode): PType =
   typ[0] = makeRangeType(c, 0, n.len - 1, n.info)
   result = typ
 
+proc semArrayElementIndex(c: PContext, n: PNode, formalIdx: PType
+                         ): tuple[n: PNode, val: Int128] =
+  ## Analyses the array element expression `n`, producing either the
+  ## semantically analysed index expression plus its value, an error, or an
+  ## ``nkEmpty`` if no index is specified.
+  ##
+  ## `formalIdx` specifies the type that the index value must fit. If
+  ## `formalIdx` is nil, the index specified by `n` is always allowed as long
+  ## as it's of a valid ordinal-like type.
+  addInNimDebugUtils(c.config, "semArrayElementIndex", n, result.n)
+
+  case n.kind
+  of nkExprColonExpr:
+    checkSonsLen(n, 2, c.config)
+    let idx = semRealConstExpr(c, n[0])
+
+    if idx.kind == nkError:
+      (idx, Zero)
+    elif formalIdx != nil:
+      # we already have a formal type, make sure the provided index fits
+      let r = fitNode(c, formalIdx, idx, idx.info)
+      if r.kind == nkError:  (r, Zero)
+      else:                  (r, getOrdValue(r))
+    elif isOrdinalType(idx.typ):
+      # no formal type is set yet and we got a valid ordinal
+      (idx, getOrdValue(idx))
+    else:
+      # a constant expression, but not a value usable as an ordinal
+      (c.config.newError(idx, PAstDiag(kind: adSemExpectedOrdinalArrayIdx,
+                                       nonOrdInput: idx,
+                                       indexExpr: n)),
+       Zero)
+  else:
+    (c.graph.emptyNode, Zero)
+
 proc semArrayConstr(c: PContext, n: PNode, flags: TExprFlags): PNode =
   result = newNodeI(nkBracket, n.info)
   result.typ = newTypeS(tyArray, c)
-  rawAddSon(result.typ, nil)     # index type
-  
-  var
-    firstIndex, lastIndex: Int128
-    indexType = getSysType(c.graph, n.info, tyInt)
-    lastValidIndex = lastOrd(c.config, indexType)
-  
+
+  template createRange(first, last: Int128, typ: PType): PType =
+    makeRangeType(c, toInt64(first), toInt64(last), n.info, typ)
+
   if n.len == 0:
+    # an empty array constructor
+    let indexType = getSysType(c.graph, n.info, tyInt)
+    rawAddSon(result.typ, createRange(Zero, toInt128(-1), indexType))
     rawAddSon(result.typ, newTypeS(tyEmpty, c)) # needs an empty basetype!
-    lastIndex = toInt128(-1)
   else:
-    var x = n[0]
-    if x.kind == nkExprColonExpr and x.len == 2:
-      var idx = semConstExpr(c, x[0])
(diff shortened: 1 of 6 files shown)


compiler/sem/semexprs.nim
```diff
@@ -696,85 +696,135 @@ proc arrayConstrType(c: PContext, n: PNode): PType =
   typ[0] = makeRangeType(c, 0, n.len - 1, n.info)
   result = typ
 
+proc semArrayElement(c: PContext, n: PNode, formalIdx: PType
```
saem: Much nicer having this separated out.

compiler/sem/semexprs.nim
```diff
@@ -696,85 +696,135 @@ proc arrayConstrType(c: PContext, n: PNode): PType =
...
+    # analyse the first element separately -- the other elements depend on its
+    # type
+    var
+      (first, firstIndex, indexType) = semArrayElement(c, n[0], nil)
+      lastIndex = firstIndex ## tracks the index of the previous element
+
+    if indexType == nil:
+      # if no type is provided or an error occurred, default to ``int`` for
+      # the index type
+      indexType = getSysType(c.graph, n.info, tyInt)
```
saem: If you'd like to make this a `let` instead of a `var`, you could rename `indexType` to `inferredIndexType`, and then follow that up with a single define/set of `indexType`, like so

```suggestion
    let
      (first, firstIndex, inferredIndexType) = semArrayElement(c, n[0], nil)
      indexType =
        # if no type is provided or an error occurred, default to ``int`` for
        # the index type
        if inferredIndexType.isNil: c.graph.getSysType(n.info, tyInt)
        else: inferredIndexType
                  
    var lastIndex = firstIndex ## tracks the index of the previous element
```

compiler/sem/semexprs.nim
```diff
@@ -696,85 +696,135 @@ proc arrayConstrType(c: PContext, n: PNode): PType =
...
-      if x.kind == nkExprColonExpr and x.len == 2:
-        var idx = semConstExpr(c, x[0])
-        idx = fitNode(c, indexType, idx, x.info)
-        if lastIndex + 1 != getOrdValue(idx):
-          result = n
-          result[i] = c.config.newError(x, 
-                          PAstDiag(kind: adSemInvalidOrderInArrayConstructor))
-          result = c.config.wrapError(result)
-          return
+    if first.kind != nkError:
```
saem: With the `inferredIndexType` as a separate definition the condition `first.kind != nkError and inferredIndexType.isNil` would let you choose the starting index for the for loop below (currently hardcode to `1`). Doing that would mean all elements are handled the same way and a slight reduction in duplicated logic.

saem (approved):
Few non-blocking suggestions; the new code is much improved! 👍🏽

compiler/sem/semexprs.nim
zerbina: I'm not sure I understand. In the case that `first.kind != nkError and inferredIndexType != nil`, separate handling is still required, otherwise the first element would not be analysed or added.

compiler/sem/semexprs.nim
zerbina: It's changed locally, thanks.

compiler/sem/semexprs.nim
```diff
@@ -696,85 +696,135 @@ proc arrayConstrType(c: PContext, n: PNode): PType =
...
+  of nkExprColonExpr:
+    checkSonsLen(n, 2, c.config)
+    let idx = semRealConstExpr(c, n[0])
+
+    if idx.kind == nkError:
+      # wrap in an error
+      let e = newNodeI(nkExprColonExpr, n.info, 2)
+      e[0] = idx
+      e[1] = n[1]
+      (c.config.wrapError(e), Zero, nil)
```
saem: Not a hard requirement, I just think it'd lead to a some simplifications. Here I'd return `idx` instead of wrapping the whole `nkExprColonExpr` in an error. This way one would only get: the index node, the index node error, or an empty node (changing the `else` branch).

compiler/sem/semexprs.nim
saem: I started writing it out and it won't simplify without reworking `semArrayElement` to be more consistent in its returned element. I only suggested on consistency scheme there, but others would also work, such as: returning two `PNode` one for the index and the other for the element, if only the element is present the index is empty; or whatever else.

It's not a must, only if you want to. I don't think it necessarily will be more efficient, it might execute extra analysis when errors have already been found, it would simply read more linearly.

compiler/sem/semexprs.nim
```diff
@@ -696,85 +696,135 @@ proc arrayConstrType(c: PContext, n: PNode): PType =
...
+      if e.kind == nkError:
+        result.add e
+        continue
+
+      e = semExprWithType(c, e, {})
+      if {e.typ.kind, typ.kind} == {tyObject} and typ != e.typ:
+        # XXX: the check is meant to disallow implicit up- or down-conversion
+        #      of object types in an array constructor, but it is incorrect:
+        #      ``sink``, aliases, and generic instance types are not skipped
+        #      and the comparision doens't use ``sameType``.
```
```suggestion
        #      and the comparision doesn't use ``sameType``.
```

saem (approved):
Looks good.

All the discussions/suggestions are very soft/non-blockers; and/or me getting carried away with polish. 😆

compiler/sem/semexprs.nim
zerbina: Done. `semArrayElement` is renamed to `semArrayElementIndex` and now does what you described.

saem:
Let's go! 🥳 

Bors r+

PR #605, opened 2023-03-28
template non-ast param limited to lookup substitution

saem:
## Summary

* non-ast params (excluding `typed` and `untyped`) are only substituted
  for usage positions
* where a usage positionis is any syntactic position where a symbol is
  not intended to be introduced
* in additition, the manual and semtempl module have some doc comments
  outlining the new direction for templates

## Details

Starting with a motivating example, this now works after the change:

```nim
template foo(data: int): untyped =
  proc bar(data: int): int =
    2 * data
  bar(data)
doAssert foo(2) == 4
```

The reason this works now is `data` the template arg is a non-ast type,
it's not substituted into `data` the proc param, and `data` the proc
param is instead converted into a symbol. This means that when `data`
the proc body usage is looked up, it's lexically resolved with `data`
the proc param. Finally, when `data` the call argument's usage is looked
up, the template param is found and substituted.

### Further Details

Template parameters that are neither `typed` or `untyped` will no
longer subsititute for definitional syntax positions, or syntax
positions that meant to introduce routines, variables, types and other
named definitions. For non-ast typed substitution lexical scope based
lookups are used. AST typed template parameter substitutions continue to
work as is, since they're arbitrary symbolic substitutions.

This makes more sense given that we're substituting syntactic symbol
usage as opposed to syntactic symbol introductions; but it also removes
a vector by which symbol pollution can take place.

As templates have lots of rough edges in implementation there are still
many bugs, but these are almost entirely pre-existing.

### Overview of Concepts Introduced:

This is a brief overview, see the module and docs for more info, also
this is developing, these are still mostly in docs:
* the concept of template substitution being policy being set by
  template output type has been introduced
* a template body is to be thought of as an out parameter with the same
  type as the template output
* additionally, a template body can be considered as a form of quasi
  quoting, where the type determines how we treat the substitution of
  various kinds of parameters (ast and non-ast types).

---

## Notes for Reviewers
* first step in a series of many to completely rework templates


compiler/sem/semtempl.nim
```diff
@@ -9,6 +9,51 @@
...
+## as an `untyped` AST fragment, that substitution of parameters will evaluate
+## per the rules of `untyped` templates, and finally evaluation and insertion
+## of the template at the callsite will be hygeinic. The template parameter
+## `arg` will be captured as `untyped`, meaning no attempt will be made to
+## semantically analyse the parameter prior to substitution.
+## 
+## Template Taxonomy
+## =================
+## 
+## There are at least four types of templates across two categoriesare:
```
zerbina:
```suggestion
## There are at least four types of templates across two categories:
```

compiler/sem/semtempl.nim
```diff
@@ -9,6 +9,51 @@
...
+##   - `typed`
+## - expression templates (all types that are not `untyped` or `typed`)
+## 
+## Substitution Positions
+## ----------------------
+## Templates are ultimately AST level constructs regardless of output type,
+## even they follow the grammar. There are two types of positions in a template
+## body, one is `definition` and the other is `usage`. A `definition` are any
+## position where the grammar construct is intended to introduce a new symbol,
+## ie: the name of a routine, including its parameters; names of variables
```
zerbina:
```suggestion
## i.e.: the name of a routine, including its parameters; names of variables 
```

doc/manual.rst
```diff
@@ -5386,10 +5386,32 @@ templates:
...
+broader redesign of how templates work in |Nimskull|.
+
+Defining vs Using Positions
+---------------------------
+
+Substitution positions are places in the template body where template parameter
+substitution can take place. There are two substitution positions definition
+and usage, also referred to as definitional/defining/define or using/use,
+respectively. Definitional positions is any syntactic position intended to
+define new names (eg: routine, variable, parameter, type, field names), while
```
zerbina:
```suggestion
define new names (e.g.: routine, variable, parameter, type, field names), while
```

compiler/sem/sem.nim
```diff
@@ -350,18 +350,14 @@ proc newSymG*(kind: TSymKind, n: PNode, c: PContext): PSym =
...
-
-    when false:
-      if sfGenSym in result.flags and result.kind notin {skTemplate, skMacro, skParam}:
-        # declarative context, so produce a fresh gensym:
-        result = copySym(result)
-        result.ast = n.sym.ast
-        put(c.p, n.sym, result)
+    doAssert n.sym.kind in {kind, skTemp},
+      "there is a bug the callee or sanitizer, got: $#, expected: $#" %
+        [$n.sym.kind, $kind]
```
zerbina: It's definitely possible to leave the symbol check to the sanitizer, although I'm still a bit undecided on whether the check is something that fits better here.

In its current form, the assert would also conflict with the [tests/lang_callable/macros/tsymbol_must_match.nim](https://github.com/nim-works/nimskull/pull/590/commits/f22e1b69540ac2db7b61ed4c2683c50058ebf8a8#diff-128b474d331f2e1dcb7bfc6a78d5ece9433bcc828ed2c0559880cb26c7fe19bc) test added by #590.

Anyhow, I don't think the assertion is a blocker, and am okay with the removing the test mentioned above again.

zerbina (approved):
I very much agree with the direction. Thank you for working on this.

`typed` templates allowing `typed` parameters in definition positions is something that I'm still not entirely sure of, but I don't think it's blocking -- we can always exclude them later if in practice they turn out to be a problem.

The suggestions left are also non-blocking.

compiler/sem/sem.nim
saem: Thanks for the review. I'm ok with dropping the assertion, mostly I think the error generation shouldn't happen in this proc and we should have prevented/figured things out before hand. Where that figuring out happens very much depends upon how we got here.

saem: Went back to the report.

PR #606, opened 2023-03-29
`considerQuotedIdent` is no longer call recursive

saem:
## Summary

Rewrite `lookups.considerQuotedIdent` in preparation for further structuring AST diagnostics. The new version eliminates call recursion.

## Details

A minor improvement to `nimsuggest` was shipped as part of this to slightly reduce memory allocation spikes from reports.

To Revisit in the Future:
Macros can create grammatically invalid structures, including for identifiers, where should we check for identifier well formed-ness? The previous proc was slightly more forgiving but not by much.

---

## Notes for Reviewers
* tiny step, next is to further specialize and structure diagnostics it produces, see below

Future Next Steps:

The next step is to have a single primary diagnostic for "considering
an AST node as an identifier". Further "sub-diagnostics" (read: further
nested variant data) will go into detail as required. This will allow
for a sort of hierarchy without inheritance, and more importantly make
recovery from errors, specialization of errors, and querying of error
AST, far smoother.

```nim
type
  # assume a `adSemIdentifierNode` value in `AstDiagKind`
  ADSemIdentifierNodeKind = enum
    adSemIdentifierNodeExpected
    adSemIdentifierNodeAccQuotedEmpty
    adSemIdentifierNodeAccQuotedMalformed
    adSemIdentifierNodeSymChoiceEmpty
    adSemIdentifierNodeSymChoiceMalformed

  ADSemIdentifierNodeData = object
    # `TAstDiag`'s branch for `adSemIdentifierNode` will store this
    case kind*: ADSemIdentifierNodeKind:
      # variant fields
```

The above sketch will be tested as part of creating a variant of
the `newSymG` proc called `newSymGNode`, the difference being that it
will return a `PNode` guaranteed to be a symbolic node or error. In a
prior attempt this ran into roadblocks as `considerQuotedIdent`'s output
is not easy to recover from or disambiguate various identifier errors.
It's easy enough to add more kinds, but then there is a loose
collection of kinds and lots of name collisions for the same data when
storing it and fragmented names for the reader to unify.

### DIFF
--- a/compiler/sem/lookups.nim
+++ b/compiler/sem/lookups.nim
@@ -28,6 +28,7 @@ import
     modulegraphs
   ],
   compiler/utils/[
+    idioms,
     debugutils
   ],
   compiler/front/[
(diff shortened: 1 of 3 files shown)


compiler/sem/lookups.nim
```diff
@@ -56,64 +57,89 @@ type
...
+    const
+      renderableLiterals = nkLiterals - nkFloatLiterals
+      renderableKinds = atomicIdentKinds + renderableLiterals
+    var
+      id = ""
+      error = false
+    for q in quote.items:
+      case q.kind
+      of nkIdent, nkSym:     id.add(extractAtomicIdent(q).s)
+      of renderableLiterals: id.add(q.renderTree)
```
zerbina:
```suggestion
      of nkIntLiterals: id.addInt(x.intVal)
      of nkStrLiterals: id.add(x.strVal)
```
I think making the rendering explicit like this is better. The main reasons I think so is because:
* `renderTree` is meant for a different domain (rendering AST as source code)
* it's more strict; changes to the `nkLiterals` set (such as including `nkNilLit`) need to be explicitly handled here (or not)
* it's a tiny bit more efficient

It would be a breaking change, however: a `AccQuoted (Ident "a") (StrLit "b")` would now be rendered into `ab` instead of `a"b"`, and a `AccQuoted (Ident "a") (IntLit8 1)` into `a1` instead of `a1'i8`. I personally think the new behaviour better matches what is expected when using identifier construction.

compiler/sem/lookups.nim
```diff
@@ -56,64 +57,89 @@ type
...
-    case result[1].kind
-    of nkError:
-      discard    # nothing to do, it's already an error
+  var
+    currN {.cursor.} = n
+    ident: PIdent
+    descended = false
+      ## we need to descend once and only once for accent quoted and
+      ## open/closed sym choices; the logic upon descent changes
+  while ident.isNil:
```
zerbina: A soft suggestion of what I think might be easier to follow than the loop:
```nim
var
  currN {.cursor.} = n
  ident: PIdent

case currN.kind
of nkIdent, nkSym:
  ...
of nkOpenSymChoice, nkClosedSymChoice:
  ...
of nkAccQuoted:
  ...
of allNodeKinds - identKinds:
  ...

if ident == nil: # check the inner part
  ident =
    case currN.kind
    of ...

```

compiler/sem/lookups.nim
```diff
@@ -56,64 +57,89 @@ type
...
-        (ident: ic.getNotFoundIdent(), errNode: n)
-    else:
-      (ident: ic.getNotFoundIdent(), errNode: n)
+  let
+    ic = c.cache
+    notFoundIdent = ic.getNotFoundIdent()
+
+  const
+    atomicIdentKinds = {nkIdent, nkSym}
+    identKinds = atomicIdentKinds +
```
zerbina: You can use the already existing `nkIdentKinds` instead (it's the same set).

zerbina (commented):
I left a few suggestions regarding the code, but agree with the overall direction.

---

`considerQuotedIdent` being able to return an error introduces quite some complexity in the compiler. The `nkAccQuoted` nodes we get from the parser and that are outside the bodies templates are always valid -- the only possibly erroneous ones are those produced by template/macro.

Based on this observation, I think the following would make sense:
- the normal `considerQuotedIdent` only handles the `nkAccQuoted` shapes present in untyped AST (what the parser produces), which means that it never needs to produce an error
- for `typed` and non-AST typed templates, identifier construction expressions are evaluated during template evaluation, using similar logic to how `considerQuotedIdent` currently does it, but producing either an `nkIdent` or `nkError` node
- for macros and `untyped` templates, the sanitizer evaluates the `nkAccQuoted` nodes, producing either an `nkIdent` or `nkError` node for them (the current version of the sanitizer PR implements this already)

Together, this would allow for `considerQuotedIdent` to always return a valid `PIdent`, simplifying a lot of logic.

compiler/sem/lookups.nim
saem: I'll give it a shot and sheet what CI comes back with, I agree that it better matches what should happen for identifier construction -- if CI whines, I'll likely just change the tests.

compiler/sem/lookups.nim
saem: As I wrote the current loop approach I did feel I was being a little too clever, but got carried away. I think a part of me needed to hear "it's too clever" from another person. Let me update it.

zerbina (approved):
bors r+

saem:
Oh well, it timed out so I used the opportunity to add tests. There are a lot of pre-existing issues with the feature, they're cataloged in the test. 😓 

I'll merge it after bors passes again, the rest can be fixed later.

PR #607, opened 2023-03-29
closureiters: use `case`-based dispatcher

zerbina:
## Summary

Use a case statement for the dispatcher generated by the `closureiters` transformation, removing the need for both dedicated code generator support and the `nkState` node kind. This is a significant step towards closure iterator support the JavaScript and VM backend.

While neither officially supported nor documented, it was previously possible to emit `nkGotoState` and `nkState` nodes from a macro, which could be used for unsafe unstructured control-flow. With the removal of the `nkState` node kind and `nkGotoState` being disallowed in AST reaching semantic analysis, the aforementioned is no longer possible.

## Details

The AST generated by the `closureiters` transformation always looked as follows:

```nim
while true:
  block :stateLoop: # the block's body is wrapped in a try-except, if
		    # necessary
    goto env.:state # goto the label corresponding to the run-time value
                    # of `:state`
    # declarations
    body
```

where `body` defines the code blocks associated with each state value, like so:

```nim
state 0:
...
env.:state = 1
break :stateLoop
state 1:
...
state 2:
...
... # more labels (if used) follow
```

The control-flow described by the `goto` (`nkGotoState`) followed by the labels (`nkState`) is equally representable by the use of a case statement, which is what is now done after this commit.

To summarize the change:
- don't support `nkGotoState` nodes in AST reaching sem
- adjust `closureiters` to generate a case statement for the dispatcher and to not depend on `nkState`
- remove C code-generator support for `nkGotoState` and `nkState`
- remove the `nkState` node kind

Nodes with the `nkGotoState` kind now only exists *during* the transformation, but never before or after.

---

## Notes for Reviewers
* all code-generators support the closure-iterator transformation now, but `jsgen` and `vmgen` don't implement the `mFinished` magic yet, meaning that `.closure` iterators are still not supported by the respective backends
* I'm going to implement `mFinished` support as a follow-up PR

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/closureiters.nim
+++ b/compiler/sem/closureiters.nim
@@ -185,7 +185,7 @@ type
     unrollFinallySym: PSym # Indicates that we're unrolling finally states (either exception happened or premature return)
     curExcSym: PSym # Current exception
 
-    states: seq[PNode] # The resulting states. Every state is an nkState node.
+    states: seq[tuple[label: int, body: PNode]] # The resulting states
     blockLevel: int # Temp used to transform break and continue stmts
     stateLoopLabel: PSym # Label to break on, when jumping between states.
     exitStateIdx: int # index of the last state
@@ -266,9 +266,7 @@ proc newState(ctx: var Ctx, n, gotoOut: PNode): int =
   # Returns index of the newly created state
 
   result = ctx.states.len
-  let resLit = ctx.g.newIntLit(n.info, result)
-  let s = newTreeI(nkState, n.info): [resLit, n]
-  ctx.states.add(s)
+  ctx.states.add (result, n)
   ctx.exceptionTable.add(ctx.curExcHandlingState)
 
   if not gotoOut.isNil:
@@ -1128,7 +1126,7 @@ proc skipEmptyStates(ctx: Ctx, stateIdx: int): int =
     if label == -1:
       newLabel = ctx.exitStateIdx
     else:
-      let fs = skipStmtList(ctx, ctx.states[label][1])
+      let fs = skipStmtList(ctx, ctx.states[label].body)
       if fs.kind == nkGotoState:
         newLabel = fs[0].intVal.int
     if label == newLabel: break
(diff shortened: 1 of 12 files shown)


compiler/sem/closureiters.nim
```diff
@@ -185,7 +185,7 @@ type
     unrollFinallySym: PSym # Indicates that we're unrolling finally states (either exception happened or premature return)
     curExcSym: PSym # Current exception
 
-    states: seq[PNode] # The resulting states. Every state is an nkState node.
+    states: seq[tuple[label: int, body: PNode]] # The resulting states
```
saem, line 188: This type reads more nicely and the tuple fields make the access below easier to follow when skimming. 👍🏽

compiler/sem/closureiters.nim
```diff
@@ -1291,29 +1284,29 @@ proc deleteEmptyStates(ctx: var Ctx) =
...
       elif excHandlState != 0:
         ctx.exceptionTable[i] = ctx.skipEmptyStates(excHandlState)
 
-  var i = 0
+  # remove unused states except for the first (entry) and last one (exit)
+  var i = 1
   while i < ctx.states.len - 1:
-    let fs = skipStmtList(ctx, ctx.states[i][1])
-    if fs.kind == nkGotoState and i != 0:
+    if ctx.states[i].label == -1: # is it unused?
```
saem: Very soft suggestion: make `-1` a constant, such as `unusedStateIdx`, might read a bit better. It'd also be applicable to the setting of the label values above.

saem (approved):

saem:
I definitely like how this collapses the language into itself. Even before this change, I can imagine using case to explain how iterators work to a newcomer, and now the simple explanation and implementation are that much closer. Makes bridging the gap from language user to contributor that much smaller. 🚀

compiler/sem/closureiters.nim
zerbina: That's much better, thanks.

PR #608, opened 2023-03-29
vmgen: fix `set` elements with large offsets

zerbina:
## Summary

Using a non-static value as a `set` element operand to procedures like `in`, `incl`, etc. resulted in an internal compiler when the lower bound of the target `set` had an offset outside of the -127..127 range.

This was caused by the code generator incorrectly assuming that the offset would fit into the immediate operand of an instruction using the ABI encoding.

The code generator now selects the appropriate instruction for applying the offset, considering the offset value.

---

## Notes for Reviewers
* an issue that I noticed as part of implementing `mFinished` support for the VM
* the issues found by the `tsets.nim` test is pre-existing

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


saem (approved):
Nice find.

bors r+

PR #609, opened 2023-03-29
custom numeric literal parsing no longer crashes

saem:
## Summary

Parsing for custom numeric literals led to a compiler crash as the
logic resulted in a field access defect. That's been fixed and the
tests adjusted.

## Details

Updated `toPNode` routine to no longer attempt to access the
`isBlockArg` and `sons` fields for a parsed custom literal as those
fields do not exist on that node type.

Note, testing was finding the issue before, sort of. The reason why it
went unnoticed was because `tcustom_numeric_literals` reported a pass
because we first test it via a `static: ` call. This likely caused the
semfold/VM to not only suppress the errors but potentially modify the
AST such that runtime never executes the code. As a workaround for
this, until the real reason is found I've inverted the execution,
ensuring runtime is first.

---

## Notes for Reviewers
* found while adding tests for token construction

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -788,7 +788,6 @@ proc splitCustomLit*(n: ParsedNode): tuple[num, ident: ParsedNode] {.inline.} =
 
 proc toPNode*(parsed: ParsedNode): PNode =
   if parsed.isNil: return
-  let kind = parsed.kind.toTNodeKind
   result =
     case parsed.kind
     of pnkCustomLit:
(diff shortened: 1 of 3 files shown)


saem:
@zerbina added you as a reviewer just so you're aware of the issue. I'm going to look across the rest of the test cases which do the same `static: main()` trick and invert them. Hopefully not too much is hiding, it should also help indicate where in/around fold (likely culprit) things are going sideways.

saem:
I did a quick pass and didn't find others issues. I'll have to dig in a bit more to see how the error got swallowed, it was a clear field defect, so not sure what went on there.

zerbina (approved):
Ah, yeah, the custom literal issue. I've been carrying around a commit with the fix for some time now, but never got around to making a PR.

> I'll have to dig in a bit more to see how the error got swallowed, it was a clear field defect, so not sure what went on there.

The reason the issue hasn't surfaced in CI is because CI uses a `-d:danger` compiler, where field, overflow, and other run-time checks are disabled. I switched to always using a `-d:release` compiler locally, in order to be able to find issues not caught by CI.

Given the recent improvements to compiler performance, it might make sense to also switch to `-d:release` for the compiler used by the CI.

saem:
Bors r+

PR #611, opened 2023-03-30
`lang_callable` tests enabled for all backends

saem:
## Summary

Enabled all tests in the `lang_callable` test category for all backends
by default. Some tests might be a platform specific here and there, but
they generally run across the board. This work is to increase test
coverage when enabling closure iterators across all targets.

## Details

This resulted in a number of test modifications, such as dropping usage
of `system.stdout` for output. This was replaced with appending to
strings and/or `echo` as necessary. For disabling specific tests or
subsections comments were recorded with the keyword "knownIssue".

Replacing all `echo`ed output with `doAssert`s wasn't attempted, as the
scope of change would be too large.

Follow-Ups:
- false negatives due to VM floating point to string format differences
- certain imports json/stream/os seem cause VM target failures

---

## Notes for Reviewers
* focused primarily on getting the category running cleanly
* rest of the work can be incrementally approached

### DIFF
--- a/testament/testament.nim
+++ b/testament/testament.nim
@@ -797,10 +797,12 @@ proc generatedFile(test: TTest, target: TTarget): string =
     let ext = target.ext
     nimcacheDir(test.name, test.options, target) / "@m" & name.changeFileExt(ext)
 
-proc needsCodegenCheck(spec: TSpec): bool =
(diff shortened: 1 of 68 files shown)


saem:
Bors r+

PR #612, opened 2023-03-31
support `.closure` iterators with all backends

zerbina:
## Summary

`jsgen` and `vmgen` now implement the `finished` magic procedure, making `.closure` iterators available:
- at compile time
- for the NimScript target
- for the standalone VM target
- for the JavaScript target

## Details

The `finished` magic procedure is responsible for querying wether the provided iterator has finished, and is inserted by the compiler when transforming for-loops that use `.closure` iterators.

Implementing support for it in the JavaScript code-generator is straightforward, as the operation can be mapped to a simple field lookup + comparison.

Closures (which includes `.closure` iterator instances) are not represented as normal tuples in the VM, but rather via a dedicated type that uses dedicated operations for querying and modification. To be able to access the environment of a closure object, the new ``AccessEnv`` operation is introduced, which the `mFinished` implementation in `vmgen` then makes use of.

The compiler run-time procedure `closureIterExcSetup` is also implemented for both the JavaScript and VM target, which required the VM callback handling slightly adjusted in order to support setting the virtual machine's current exception.

### Tests

- enable the tests that make use of `.closure` iterators for the JavaScript and VM target
- reduce the test for a closure bug, removing the unnecessary dependency on the `os` module. Since the bug is unrelated to `.closure` iterators, it is moved to a different test file
- move a test unrelated to `hash` out of `thashes.nim`
- make a test in `ttemplate_issues.nim` work when located inside a `block`

---

Thank you again for preparing the `lang_callable` category, @saem.

## Notes for Reviewers
* this is a follow-up to #607 ; most of the work towards full `.closure` iterator support happened there

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


tests/lang_callable/closure/tclosure_nested_iterator.nim
```diff
@@ -1,5 +1,5 @@
 discard """
-  targets: "c !js !vm"
+  targets: "c js vm"
```
saem: Could delete this line outright.

saem (approved):
Awesome possum.

saem:
I'll kick off bors, worst case it can be cancelled, the incredibly minor suggestion can be tackled anytime someone sees it next.

bors r+

zerbina:
Ah, sorry. I missed the comment by one minute :D.

saem:
I laughed when I saw the push notification. 😆 

bors r+

PR #613, opened 2023-04-01
backtick identifier construction fully recursive

saem:
## Summary

Identifier construction using backticks (```) is now fully recrusive.
Allowing constructing identifiers from other `nkAccQuoted` identifier
strings.

## Details

Changed `considerQuotedIdent` to construct an identifier from child
node string values. In addition, a hardcoded limit on the number of
elements in a `nkAccQuoted` evaluation has been set to avoid run away 
computations.

It was always the case, but more so than ever, identifier construction
can be triggered from anywhere and is more capable than ever. See below
for initial thoughts about the discussion that needs to happen.

As a result of the vague nature of what should be, and how pervasively
we should allow identifier construction, which hasn't been detangled
from keyword escaping (much easier to do now), the code is a little
over-engineered.

#### Bigger Discussion/Open Question

`considerQuotedIdent` raises an interesting question around what is a
syntactically valid identifier vs a core language valid identifier.
That is to say what can a human provide as source that'll pass lexing
and parsing, and is generally known as Nimskull vs what our "lisp-y"
core language allows. Further to that, what do we afford
end-programmers via templates vs macros? Presently, `nkAccQuote` we
don't _really_ enforce identifier well-formed-ness in many places and
`nkAccQuoted` identifier construction let's us get away with a lot
more than keyword escaping.

---

## Notes for Reviewers
* admittedly feels over-engineered
* fun fact, we can do identifier construction pretty much everywhere, because hacks

### DIFF
--- a/compiler/sem/lookups.nim
+++ b/compiler/sem/lookups.nim
@@ -54,20 +54,148 @@ proc ensureNoMissingOrUnusedSymbols(c: PContext; scope: PScope)
 type
   PIdentResult* = tuple
     ident: PIdent      ## found ident, otherwise `IdentCache.notFoundIdent`
-    errNode: PNode     ## if ident is notFoundIdent, node where error occurred
-                       ## use with original PNode for better error reporting
+    errNode: PNode     ## if ident is notFoundIdent, node with the error, use
+                       ## with original PNode for better error reporting
 
 proc considerQuotedIdent*(c: PContext; n: PNode): PIdentResult =
-  ## Retrieve a PIdent from a PNode, taking into account accent nodes.
+  ## Retrieve a PIdent from a PNode, taking into account accent nodes for
+  ## keyword escaping (stropping) and identifier construction. If none found,
+  ## returns a `idents.IdentCache.identNotFound` and an error node (`nkError`).
   ##
-  ## If none found, returns a `idents.IdentCache.identNotFound`
+  ## Idents are derived as follows:
+  ## - nkIdent -> ident
+  ## - nkSym -> ident
+  ## - nkSymChoices -> ident of first child; assert if malformed
+  ## - nkAccQuoted:
+  ##    - if empty -> an error
+  ##    - if singleton with nkIdent|Sym|SymChoices -> handled as above
+  ##    - else -> construct identifier, see next
+  ## - anything else -> not found (see note )
+  ##
+  ## Ident Construction, creates an ident by concatenating child string values:
+  ## - nkIdent|Sym|SymChoices -> ident.s for string value
+  ## - nkAccQuoted:
+  ##   - if empty -> empty string
+  ##   - else -> recursively evaluate
+  ## - string literals -> string
+  ## - chararacter literal -> character as string
+  ## - integer literal -> int to string
+  ## - anything else -> error
+  ## To ensure we don't have a run away evaluation, identifier construction has
+  ## a hard limit on number of items.
+  # xxx: `considerQuotedIdent` raises an interesting question around what is a
+  #      syntactically valid identifier vs a core language valid identifier.
+  #      That is to say what can a human provide as source that'll pass lexing
+  #      and parsing, and is generally known as Nimskull vs what our "lisp-y"
+  #      core language allows. Further to that, what do we afford
+  #      end-programmers via templates vs macros? Presently, `nkAccQuote` we
+  #      don't _really_ enforce identifier well-formed-ness in many places and
+  #      `nkAccQuoted` identifier construction let's us get away with a lot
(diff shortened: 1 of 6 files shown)


compiler/sem/lookups.nim
```diff
@@ -54,50 +54,158 @@ proc ensureNoMissingOrUnusedSymbols(c: PContext; scope: PScope)
...
+      elif wellFormedAst:
+        var
+          id = ""
+          badNode: PNode = nil
+        for s in flattened.items:
+          case s.kind
+          of nkIdent:                     id.add(s.ident.s)
+          of nkSym:                       id.add(s.sym.name.s)
+          of nkStrLiterals:               id.add(s.strVal)
+          of nkCharLit:                   id.add(char(s.intVal))
```
zerbina, line 180: Nothing that has to happen now, but I think in the future, we should probably disallow characters outside of the 32..127 range.

tests/lang/s02_core/s05_templates/t01_templates.nim
```diff
@@ -89,19 +89,23 @@ block identifier_join:
...
-      getVar('\n') # produces "varName10" ... wtf
-      getVar("\n") # produces "varName<newLine>"
-
-  when defined(tryBrokenSpecification):
-    block works_with_stropping:
-      ## Stropping does not work with identifier join in templates
-      var `varName??` = 50
-      doAssert getVar(`??`) == 50
+  block works_with_stropping:
+    ## Stropping does not work with identifier join in templates
```
zerbina:
```suggestion
    ## Stropping works with identifier join in templates
```

compiler/sem/lookups.nim
```diff
@@ -54,50 +54,158 @@ proc ensureNoMissingOrUnusedSymbols(c: PContext; scope: PScope)
...
 
   const
     atomicIdentKinds = {nkIdent, nkSym}
+    renderableLiterals = nkLiterals - nkFloatLiterals
+    renderableKinds = atomicIdentKinds + renderableLiterals
     allNodeKinds = {low(TNodeKind)..high(TNodeKind)}
 
+  proc accQuotedConstrIdentNode(c: PContext, n: PNode): PNode =
+    ## construct and identifier the `nkAccQuoted` node `n`, returning an
+    ## `nkIdent` node, or `nkError` upon failure.
```
zerbina:
```suggestion
    ## construct an identifier from the `nkAccQuoted` node `n`, returning an
    ## `nkIdent` node, or `nkError` upon failure.
```

zerbina (commented):
Hmm, I get what you mean with the change feeling overengineered. Personally, I think one level of nesting, where the nested `nkAccQuoted` only contains either zero or a single identifier-like nodes, is enough (it allows for supporting the quoted-keywork-passed-to-template case).

As far as I can tell, more complex `nkAccQuoted` expression are only possible through macros, and I think it's okay to require them to construct identifiers themselves instead of pushing this into the compiler.

Anyhow, I'm okay with merging this for now, but think that we should revisit the logic in the future.

---

Regarding what kind of identifiers are allowed to be created via identifier construction, I think restricting it to the what the parser allows for quoted identifier makes sense.

compiler/sem/lookups.nim
saem: Yup, I was able to put new lines in identifier and all sorts of fun stuff; also negative numbers, and more.

saem:
Thanks, I'll update the grammar issues.

----

I agree we need to revise things and something simpler might be a better bet. I think the part that's tricky with macros/templates generating these is it is pretty handy; consider that our own analyses generates `:tmp` in the compiler today. So having a flexible core allowing for extension is important, but then the whole syntax roundtrip problem shows up. 😅 It needs a bigger think. Also, there are some very useful things one could do, here is an example from kotlin:

```kotlin
@Test
fun `adding 1 and 1 gives 2`() {
    assertEquals(myCalculator.add(1, 1), 2)
}
```

I need a bit of distance for a few days, but my guess at present is that for the human source code type stuff we implement it in a non-optional stylecheck, and for the rest we keep it more wild-west and have one level interpolation for `nkAccQuoted` expressions as you mentioned. Then the full recursive option can be in the macro API; the lack of ability to fold that in templates might be a bit of a bummer, but we can figure that out by experience.

Thanks again for the review!

tests/lang/s02_core/s05_templates/t01_templates.nim
saem: thanks, updated.

compiler/sem/lookups.nim
saem: thanks, updated.

PR #615, opened 2023-04-01
sem: fix default parameter handling for macros

zerbina:
## Summary

Default parameters did not work for macros that are invoked without overload resolution happening prior, resulting in `wrong number of arguments` errors. This is fixed now, meaning that invoking macros for which all parameters have default values now works even when not using the call or command syntax.

## Details

`semMacroExpr` now uses the `evalTemplateArgs` procedure for pre-processing and validating the arguments.

With the usage of `evalTemplateArgs`, the previous argument validation logic in `semMacroExpr` and `evalMacroCall` becomes redundant and is removed, reducing the amount of code duplication a bit.

In addition, the recursion check in `evalMacroCall` is removed. The procedure is not re-entrant, which made the guard unnecessary. Infinite nesting (i.e.: a macro evaluating to a call to itself) is already detected by `semAfterMacroCall`.

### Issues discovered

Default parameters are *always* analysed during routine header analysis, even for ``untyped`` parameters. In practice, this means that AST passed to ``untyped`` parameters with default values can not be relied on to really be untyped.

---

## Notes for Reviewers
* a small preparation for macros storing and using their internal signatures

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -652,14 +652,13 @@ proc semMacroExpr(c: PContext, n: PNode, sym: PSym,
   if sym == c.p.owner:
     globalReport(c.config, info, reportSym(rsemCyclicDependency, sym))
 
-  let
-    genericParams = sym.ast[genericParamsPos].safeLen
-    suppliedParams = max(n.safeLen - 1, 0)
-
-  if suppliedParams < genericParams:
-    globalReport(
-      c.config, info, reportAst(
-        rsemMissingGenericParamsForTemplate, n, sym = sym))
+  # XXX: also check for ``efFromHlo`` like ``semTemplateExpr`` does?
+  # XXX: using ``evalTemplateArgs`` (for both macros and templates) should not
+  #      be needed. If the routine doesn't match the provided arguments (both
+  #      generic and normal), invoking it should simply not be attempted.
+  let args = evalTemplateArgs(n, sym, c.config, fromHlo=false)
+  if args.kind == nkError:
+    return args
 
   let reportTraceExpand = c.config.macrosToExpand.hasKey(sym.name.s)
   var original: PNode
@@ -667,7 +666,7 @@ proc semMacroExpr(c: PContext, n: PNode, sym: PSym,
     original = n
 
   result = evalMacroCall(
-    c.module, c.idgen, c.graph, c.templInstCounter, n, sym)
+    c.module, c.idgen, c.graph, c.templInstCounter, n, args, sym)
 
   if efNoSemCheck notin flags:
     result = semAfterMacroCall(c, n, result, sym, flags)
--- a/compiler/vm/compilerbridge.nim
+++ b/compiler/vm/compilerbridge.nim
@@ -502,31 +502,18 @@ proc setupMacroParam(reg: var TFullReg, c: var TCtx, x: PNode, typ: PType) =
     reg = TFullReg(kind: rkNimNode, nimNode: n)
 
 proc evalMacroCall*(module: PSym; idgen: IdGenerator; g: ModuleGraph; templInstCounter: ref int;
-                    n: PNode, sym: PSym): PNode =
-  #if g.config.errorCounter > 0: return errorNode(idgen, module, n)
-
-  # XXX globalReport() is ugly here, but I don't know a better solution for now
-  inc(g.config.evalMacroCounter)
-  if g.config.evalMacroCounter > evalMacroLimit:
-    globalReport(g.config, n.info, VMReport(
-      kind: rsemMacroInstantiationTooNested, ast: n))
-
-  # immediate macros can bypass any type and arity checking so we check the
-  # arity here too:
-  if sym.typ.len > n.safeLen and sym.typ.len > 1:
-    globalReport(g.config, n.info, SemReport(
-      kind: rsemWrongNumberOfArguments,
-      ast: n,
-      countMismatch: (
-        expected: sym.typ.len - 1,
-        got: n.safeLen - 1)))
-
+                    call, args: PNode, sym: PSym): PNode =
+  ## Evaluates a call to the macro `sym` with arguments `arg` with the VM.
+  ##
+  ## `call` is the original call expression, which is used as the ``wrongNode``
+  ## in case of an error, as the node returned by the ``callsite`` macro API
+  ## procedure, and for providing line information.
   setupGlobalCtx(module, g, idgen)
   let c = PCtx g.vm
   let oldMode = c.mode
   c.mode = emStaticStmt
   c.comesFromHeuristic.line = 0'u16
-  c.callsite = n
+  c.callsite = call
   c.templInstCounter = templInstCounter
 
   defer:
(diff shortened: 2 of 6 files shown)


compiler/vm/compilerbridge.nim
```diff
@@ -502,31 +502,18 @@ proc setupMacroParam(reg: var TFullReg, c: var TCtx, x: PNode, typ: PType) =
...
-      countMismatch: (
-        expected: sym.typ.len - 1,
-        got: n.safeLen - 1)))
-
+                    call, args: PNode, sym: PSym): PNode =
+  ## Evaluates a call to the macro `sym` with arguments `arg` with the VM.
+  ##
+  ## `call` is the original call expression, which is used as the ``wrongNode``
+  ## in case of an error during error and as the node returned by the
+  ## ``callsite`` macro API procedure, and for providing line information.
```
```suggestion
  ## Evaluates a call to the macro `sym` with arguments `arg` with the VM.
  ##
  ## `call` is the original call expression, which is used as the ``wrongNode``
  ## in case of an error, as the node returned by the ``callsite`` macro API
  ## procedure, and for providing line information.
```

tests/lang_callable/macros/tdefault_parameters.nim
```diff
@@ -0,0 +1,66 @@
+discard """
+  description: "Tests for macros with default parameters"
+  action: compiles
+"""
+
+import std/macros
+
+block mixed_explicit_call:
+  # the macro is invoked via call/command syntax and has parameters without
+  # defaults
```
saem:
```suggestion
  # the macro is invoked via call/command syntax and has parameters with and
  # without defaults
```

tests/lang_callable/macros/tdefault_parameters.nim
```diff
@@ -0,0 +1,66 @@
...
+  macro m[T](a: untyped = 1) =
+    # XXX: the ``T`` is not (yet) accesible as a ``NimNode`` in the body
+    #doAssert T.typeKind == ntyInt
+    doAssert a.intVal == 1
+
+  m[int]
+
+# XXX: this doesn't work yet, but it should! If the parameter type depends on
+#      a generic parameter, the default should not be analysed until
+#      instantiation
```
saem: The current approach of analysing the ident def `a: T = 1` is definitely not right, at most it can analyse `1` and even then there are issues. I don't believe deferring the analysis helps; however, although it's more correct than todays behaviour it's still not sound. Let's say we get `macro m[T](a: T, b: T = 1): untyped` as a signature. I think the only sensible thing to do would be to error out in that circumstance... or get fancy.

### Just Error Out/Why an Error

So first the reason why I think it's an error generally speaking is that `T` has a constraint of `any` (implicitly) and assignment to `a: T: any` is contravariant; that is to say even though `int` is a subtype of `any`, `m[any]` is not a subtype of `m[int]`, but the opposite is true (`m[int]` is a subtype of `m[any]`).

### Extra Fancy

If we wanted to get fancy, then we can check the arity of a routine with respect to each type parameter, and as long as there is only one or zero(*) direct or indirect reference to a type parameter then we can consider it covariant. The multiple references/paths matter because default params (or type declataions) can themselves be sneaky, such as: `m[T](a: T, b = a + 10): untyped`; here there are two references to `T` and it can no longer be covariant.

### Minor Request

We can continue to discuss, but if the above makes sense then could you update the comment with something like:

```nim
# xxx: this doesn't work yet, but it could if we:
#      check the arity of a routine with respect to each type parameter, and as
#      long as there is only one, or zero, direct or indirect reference to a
#      type parameter then we can consider it covariant. The multiple
#      references/paths matter because default params (or type declarations)
#      can themselves be sneaky, such as: `m[T](a: T, b = a + 10): untyped`;
#      here there are two references to `T` and it can no longer be covariant.
#
#      Alternatively, this should be an explicit error, as `T` has a constraint
#      of `any` (implicitly) and assignment to `a: T: any` is contravariant;
#      that is to say even though `int` is a subtype of `any`, `m[any]` is not
#      a subtype of `m[int]`, but the opposite is true (`m[int]` is a subtype
#      of `m[any]`).
```

---------------------

(*): zero because of "generic alaises"... which is probably OK as type mapping, although I do wonder if they should be phantom/branded/flavoured types instead as instantiating a generic (i.e.: `SomeGeneric[int]`) is an instantiation of a type.

PS: I considered back propagation, but all that meta-programming only lets us "unify forward"; also that's an easy way to summon Cthulhu (🦑); we don't want to summon Cthulhu.

compiler/sem/sem.nim
```diff
@@ -652,22 +652,21 @@ proc semMacroExpr(c: PContext, n: PNode, sym: PSym,
...
 
-  let
-    genericParams = sym.ast[genericParamsPos].safeLen
-    suppliedParams = max(n.safeLen - 1, 0)
-
-  if suppliedParams < genericParams:
-    globalReport(
-      c.config, info, reportAst(
-        rsemMissingGenericParamsForTemplate, n, sym = sym))
+  # XXX: also check for ``efFromHlo`` like ``semTemplateExpr`` does?
```
saem, line 655: (Good intuition, pretty sure you found a bug)

Huh, yeah it never happened before, but HLO does definitely end up calling `semMacroExpr`. I quickly pulled down your branch and duplicated the last test block in `tests/lang_experimental/trmacros/trmacros_various.nim`, called `tstatic_t_bug` (see below for snippet).

Trying it as is gives a `Error: wrong number of arguments`. Then I add the `efFromHlo` flag check and pass the flag in from `hlo` to the `semMacroExpr` call, resulting in an `IndexDefect` in the `compilerbridge` (`compiler/vm/compilerbridge.nim(544) evalMacroCall`).

I'm guessing this code path never worked/was really tested before hand. Anyhow, if I'm following things correctly (I may not be), then the hlo flag setting is correct, and `evalMacroCall` needs to make similar adaptations as `evalTemplate`, as the macro isn't going to have generic param info either.

So I did that, added a quick `fromHlo = false` param to `evalMacroCall` and wired it up in `sem.semMacroExpr` based on `efFromHlo presence`. After that I threw a quick guard around the generic parameter setting and boom, the test passed.

For fun, I ran the same compiler against the whole `lang_experimental` category to cover everything in `trmacros` and that came out green. OK _technically_ the one test I modified failed because I didn't update the spec for the extra `"optimized"` that got output, but I'm gonna call that acceptable. ;) 

If you want to incorporate it into this PR go ahead, or I can do the above in a follow-up PR after this is merged.

---

Test snippet

```nim
block tstatic_m_bug:
  # bug #4227
  type Vector64[N: static[int]] = array[N, int]

  proc `*`[N: static[int]](a: Vector64[N]; b: float64): Vector64[N] =
    result = a

  proc `+=`[N: static[int]](a: var Vector64[N]; b: Vector64[N]) =
    echo "regular"

  proc linearCombinationMut[N: static[int]](a: float64, v: var Vector64[N], w: Vector64[N])  {. inline .} =
    echo "optimized"

  macro rewriteLinearCombinationMut{v += `*`(w, a)}(a: float64, v: var Vector64, w: Vector64): auto =
    quote: linearCombinationMut(`a`, `v`, `w`)

  proc main() =
    const scaleVal = 9.0
    var a, b: Vector64[7]
    a += b * scaleval

  main()
```

saem: FWIW, I think the whole "fromHlo" bit is a little unfortunate. I'm sure there is something better that could be done, I don't know quite what that is just yet. 😅

Given that we're operating on any `Vector64` (I think that's a kind(?), because we're saying "for all Vector64", instead of "for all Vector64[N], where N is a know integer") perhaps something further up the chain is off?, Such as the whole implicit generic parameter promotion and/or the compiler needs to start understanding kinds (or something) so implicit generics don't get confused in the first place.

One thing at a time getting this uniform isn't without merit and it'll probably help guide the way.

saem (changes requested):
One small request for change wrt the comment in the tests, but we can discuss that further if you think what I'm saying doesn't makes sense. Otherwise, it's good to go IMO.

This is such a great PR: small amount of code change, has tests, docs, good impact, leaves things better, list goes on. What makes it that much cooler is the quality of conversation around it. It was a lot of fun digging in, thinking through all the deeper problems. So good! 🎉

compiler/sem/sem.nim
zerbina: Thank you for looking into it, and for providing an elaborate summary/analysis!

> Anyhow, if I'm following things correctly (I may not be), then the hlo flag setting is correct, and `evalMacroCall` needs to make similar adaptations as `evalTemplate`, as the macro isn't going to have generic param info either.

`evalTemplate` doesn't handle the HLO flag itself, it only passes it to `evalTemplateArgs`. The reason that using a template works in the example while a macro causes a crash, is that compared to `evalTemplate`, `evalMacroCall` attempts to load all arguments (both normal and generic) the macro expects.

If the template in the example would have an explicit generic parameter that it uses somewhere in the body, the `evalTemplateAux` procedure would crash with an `IndexDefect` too.

---

`fromHlo` and `efFromHlo` are workarounds that I don't think are acceptable. I've thought about term-rewriting macros a bit more, and think that clarification regarding their behaviour is required first. The two possibilities I see:

1) term-rewriting macros/templates get typed AST and perform further checks themselves. This is easy to implement:
   - disallow explicit generic parameters on term-rewriting macros/templates
   - all their parameters must use `typed` (with optional AST/expression constraints), no other types are allowed

2) in addition to the other matching, the types of the captured sub-expression must also match the ones specified by the parameters. This is slightly more complex to implement, but would, in theory, allow for things like term-rewriting macro/template overloading:
   - disallow explicit generic parameters that are impossible to infer (e.g.: `template t{a - b}[T](a: int, b: int)`)
   - disallow `untyped` parameters -- these cannot work (well, they could, but I don't think that'd be a good idea)
   - after a pattern matches, don't dispatch to `semMacroExpr`/`semTemplateExpr`  right away, but first use `sigmatch.matches` + `semResolvedCall` in order to check if the arguments types match and to prepare the expression for macro/template evaluation

Both solutions would clarify how term-rewriting are supposed to work, and would also make both `efFromHlo` and `fromHlo` obsolete, as well as working towards making `evalTemplateArgs` obsolete.

saem: (Typed on my phone so might be a bit sloppy)

> evalTemplate doesn't handle the HLO flag itself, it only passes it to evalTemplateArgs. The reason that using a template works in the example while a macro causes a crash, is that compared to evalTemplate, evalMacroCall attempts to load all arguments (both normal and generic) the macro expects.

Yup, I was typing quickly last night, but that's what I meant by putting a guard in `evalMacroCall` for generic params to get it to work. But I left a bit out, see next paragraph.

> If the template in the example would have an explicit generic parameter that it uses somewhere in the body, the evalTemplateAux procedure would crash with an IndexDefect too.

Moreover, I didn't mention it, but I put in some debug prints for the `evalMacroCall` for the `call`, `args`, and `sym`and I discovered that the term rewriting macro def, symbol and type, that we got had generic params populated. Which is why it attempted to populate generic parameters and since `evalTemplateArgs` uses the hlo hint to mark their count as `0` length it never attempts to populates them, that's the guard I replicated.

So I think a bunch of things are wrong, at least:
1. Term rewriting macros need better checking to ensure their definition makes sense as it's pretty hacky, you've outlined some approaches
2. Term rewriting macro signatures don't really make sense as written and are an interface/usability issue
3. The definition processing with implicit generic parameter promotion is off, they're not generic parameters requiring substitution and in turn leading to instantiation/passing to support parametric polymorphism. Rather the term rewriting parameters are operating "over all such types" (aka `kind`, if you're not familiar with the term `kind`s are to `type`s what `type`s are to `value`s; assuming I understand then correctly myself).
4. The issue in definition processing is being solved with an approach in dispatch processing.

That last point put another way, the `efFromHlo` flag, fakes `kind` behaviour by papering over implicit generic promotion and circumvents generic parameter passing (metaprogramming parametric polymorphism is "runtime" from the meta function perspective, which is equivalent to the instantiation approach for non-meta routines).

---

I can at least make a follow-up PR, unless you want to make further changes to this one, such that template and macro term rewriting are consistent wrt the hlo hint. Then we can pull things back apart between standard metaprogramming and term rewriting based metaprogramming, but that teasing apart should happen at the definition processing level, I think.

I think what will happen is we'll fix definition processing then the hlo hint will simply disappear leaving a single more linear code path for meta routine argument setup. The above overlaps and compliments with my understanding of what you've described as improvements to term rewriting.

zerbina: > I can at least make a follow-up PR, unless you want to make further changes to this one, such that template and macro term rewriting are consistent wrt the hlo hint

A follow-up PR is likely better -- I'll leave the `fromHlo=false` as is.

saem (approved):

zerbina:
Thanks a lot for the review @saem!

I fixed the text issues and also adjusted the comment regarding the test case. It now elaborates a bit more on the "why" part, with a link to your comment that it might be better to make this an error instead.

saem (approved):
Awesome, thanks,

(d'oh, I didn't notice the CI errors)

Bors r+

PR #616, opened 2023-04-02
sem: fix a common crash caused by `semSym`

zerbina:
## Summary

This is a quick-fix for the long-standing issue of `semSym` crashing with a `FieldDefect` when a non-generic macro or template reaches it in a `efNoEvaluateGeneric` context (e.g. the `m` in `m[int]()`).

The logic assumed that a macro/template always has a node with a `sons` field in the generic parameter list slot, which is incorrect. `safeLen` is used instead of `len` now, fixing the defect.

In addition, the macro and template symbol is now checked for errors, with an error being produced if it does.

---

## Notes for Reviewers
* this issue is present since `nkEmpty` was made a node kind without access to the `TNode.sons` field
* it so far never surfaced in the CI due to the usage of a `-d:danger` compiler there, but I think it might be the cause of the CI failure in #615

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -1602,14 +1602,20 @@ proc semSym(c: PContext, n: PNode, sym: PSym, flags: TExprFlags): PNode =
     else:
       result = newSymNode(s, n.info)
   of skMacro:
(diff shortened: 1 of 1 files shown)


saem (approved):
Nice catch.

Bors r+

PR #617, opened 2023-04-02
instanced generic routine errors: fix raise defect

saem:
## Summary

Generics instantiated for runtime routines (proc, func, method,
converter) no longer raise field defect if their ast is an error,
instead they are appeneded to the module definition. This prevents field
access defects that were being silently ignored in `danger` builds.

## Details

`sem` previously internally asserted on all non-magic runtime instanced
generic routine symbols, checking the presence of an AST 
(definition) and subsequently the routine def's `bodyPos` field, without
accounting for the AST being an `nkError`. This has now been changed to
check the AST for presence and acceptable kinds (routine def or error).

As the code is simply more strict about nils and kinds, no additional
tests should be required.

This is the start of a series of commits fixing silently ignored defects
due to danger builds. Once the compiler is faster, we can enable release
build based testing to avoid regressions.

---

## Notes for Reviewers
* doing a series of quick fixes as I run the full testsuite locally with release builds
* can incorporate feedback asynchronously as this is mostly defensive code

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -847,13 +847,28 @@ proc reportUnusedModules(c: PContext) =
       localReport(c.config, c.unusedImports[i][1], reportSym(
         rsemUnusedImport, c.unusedImports[i][0]))
 
-proc addCodeForGenerics(c: PContext, n: PNode) =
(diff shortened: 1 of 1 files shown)


saem:
Bors r+

PR #618, opened 2023-04-05
internal: begin moving `newSymG` to `newSymGNode`

saem:
## Summary

Internal refactoring for how definition position `nkIdentKinds` are
handled in semantic analysis. Added `newSymGNode` as a replacement of
`newSymG` which returns a canonical symbol node or an error and no
longer does any legacy reporting. As part of this change also introduced
an approach that consolidates diagnostic kinds and let us do more error
recovery (for `check`, `suggest`, etc use cases).

## Details

As mentioned added `newSymGNode`, `semRoutine` now uses it instead of
`newSymG`, in follow-up changes we'll be able to drop a number of legacy
reports.

Unlike `newSymG` which produces a `PSym` along with report side-effects,
`newSymGNode` only returns a `PNode`, which might be an error. What's
different about `newSymGNode` is that it only ever returns a single
diagnostic kind `adSemDefNameSym`. This singular kind describes all
possible errors as subvariants, which no longer run into all sorts of
naming colisions or giant `TAstDiag` variant of doom. With this
approach, common fields were easily setup and a singular recovery symbol
field was provided, along callees to trivially continue to make progress
and have easy access to the original input.

As part of error generation, `newSymGNode` will not modify existing
symbols and instead will create a copy of the original to avoid mutating
pre-existing symbols.

With the above change, lambda kinds which previously assumed that all
lambdas are to be treated as `skProc` triggered the error condition,
resulting in a copy being made. Copies have a "desemming" affect where
any previously attached AST is dropped. Instead of copying over the AST
on error, the `nkLambdaKinds` branch in `semexprs` was modified with a
heuristic to determine whether an `skFunc`, `skIterator` or `skProc`
should be asked for when calling `newSymG`.

Additional improvements made while working on this change:
- `semOverride`: replace node param with info
- `semOverride`: move `prevDestructor` & `bindTypeHook` helpers inside
- `semstmts.swapResult`: single use; now inner proc
- `errorreporting.localReport` add `line` pragma for better stacktrace

### Follow-Up Work

In no particular order:
- update all other `newSymG` callsites
- update `considerQuotedIdent` to behave similarly
- similarily update `semIdentVis` and other definition name analyses

### Longer-Term

Consolidated diagnostic kinds should allow for far more error recovery
and better encode the expected ordering of analyses/diagnostics.

### Observations

`checkSpecialOperators` is `adSemDefinitionNameSym` aware. When doing
this I realized there is likely two dimensions along which we want to
organize diagnostic kinds. One is grouping based on AST position, or the
kind of analysis, the other is within that the fidelity, or timing, of
the analysis. A higher fidelity analysis should result in diagnostics
that are aware of the lower fidelidty ones in the same position. So in
this case, `checkSpecialOperators` like `newSymGNode` both operate over
'definition name' positions, so their diagnostic kinds should fall under
the same branch of the hierarchy, but within that branch, the analysis
carried out by, and resulting diagnostics, `checkSpecialOperators` must
be `newSymGNode` or another other lower fidelity analysis/diagnostic.
By having an ordering of diagnostic kinds we easily check and wrap
error nodes, which will let us compose them.

---

## Notes for Reviewers
* will refine diagnostic kind consolidation and naming further in future PRs -- need more examples

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -351,6 +351,122 @@ proc newSymS(kind: TSymKind, n: PNode, c: PContext): PSym =
   when defined(nimsuggest):
     suggestDecl(c, n, result)
 
+func defNameErrorNodeAllowsSymUpdate*(n: PNode): bool {.inline.} =
+  ## true if `n` is an `nkError` of kind `adSemSymNameSym` and the error is
+  ## minor enough to allow the recovery sym to be updated. Used to see if
+  ## progress is allowed inspite of errors.
+  n.kind == nkError and n.diag.kind == adSemDefNameSym and
+    n.diag.defNameSymData.kind == adSemDefNameSymExpectedKindMismatch
+
+func getDefNameSymOrRecover*(n: PNode): PSym {.inline.} =
+  ## extracts the symbol from `n`, which must be an `nkSym` or `nkError` with
+  ## diagnostic kind of `adSemDefNameSym`. If `n` is an error, a
+  ## recovery symbol is extracted, allowing progress to be made.
+  case n.kind
+  of nkSym: n.sym
+  of nkError:
+    case n.diag.kind
+    of adSemDefNameSym: n.diag.defNameSym
+    else: unreachable("all error cases must be covered")
+  else:
+    unreachable("no other cases supported")
+
+proc newSymGNode*(kind: TSymKind, n: PNode, c: PContext): PNode =
+  ## like newSymS, but considers gensym'ed symbols, analyses `n` producing a
+  ## canonical symbol node (`nkSym`), or if unsuccessful an `nkError` with an
+  ## `adSemDefNameSym` diagnostic.
+  ## 
+  ## Symbol canonicalization is is as follows:
+  ## 1. `nkSym` -> match `kind` (or `skTemp`) and updates the sym owner to the
+  ##               current context (gensym specific)
+  ## 2. `nkIdentKinds` -> instantiate a symbol considering quoted identifiers
+  ## 3. `nkError` -> simply returns the same error
+  ## 4. other kinds -> new error
+  # TODO: come up with a better name for this proc
+  let
+    info = n.info
+    currOwner = getCurrOwner(c)
+
+  template makeError(config: ConfigRef, n: PNode, sym: PSym,
+                     dataKindSuffix: untyped): PNode =
+    let
+      dataKind = `adSemDefNameSym dataKindSuffix`
+      (wrongNode, defData) =
+        case dataKind
+        of adSemDefNameSymExpectedKindMismatch:
+          (n, AdSemDefNameSym(kind: dataKind, expectedKind: kind))
+        of adSemDefNameSymIdentGenFailed:
+          (n.diag.wrongNode,
+          AdSemDefNameSym(kind: dataKind, identGenErr: n))
+        of adSemDefNameSymExistingError,
+            adSemDefNameSymIllformedAst:
+          (n, AdSemDefNameSym(kind: dataKind))
+
+    config.newError(wrongNode, PAstDiag(kind: adSemDefNameSym,
+                                        defNameSym: sym,
+                                        defNameSymData: defData))
+
+  case n.kind
+  of nkSym:
+    result =
+      # xxx: we really should guard on `sfGenSym`; but macros can transplant
+      #      symbols from pretty much anywhere, so we don't know where gensym
+      #      really came from.
+      if n.sym.kind in {kind, skTemp}:
+        n.sym.owner = currOwner # xxx: modifying the sym owner is suss
+        n
+      else:
+        let recoverySym = copySym(n.sym, nextSymId c.idgen)
+        recoverySym.transitionRoutineSymKind(kind)
+        recoverySym.owner = currOwner
+        c.config.makeError(n, recoverySym, ExpectedKindMismatch)
+  of nkIdent, nkAccQuoted:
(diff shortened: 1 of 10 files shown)


compiler/sem/sem.nim
```diff
@@ -351,6 +351,120 @@ proc newSymS(kind: TSymKind, n: PNode, c: PContext): PSym =
...
+    config.newError(wrongNode, PAstDiag(kind: adSemDefNameSym,
+                                        defNameSym: sym,
+                                        defNameSymData: defData))
+
+  case n.kind
+  of nkSym:
+    result =
+      # xxx: why don't we guard on `sfGenSym`?
+      if n.sym.kind in {kind, skTemp}:
+        n.sym.owner = currOwner # xxx: modifying the sym owner is suss
```
zerbina, line 416: That's unfortunately required because of gensyms. When they're created either during first-phase lookup or by a macro via `genSym`, the owner is not yet know, so it has to be patched here.

compiler/sem/sem.nim
```diff
@@ -351,6 +351,120 @@ proc newSymS(kind: TSymKind, n: PNode, c: PContext): PSym =
...
+          (n, AdSemDefNameSym(kind: dataKind))
+
+    config.newError(wrongNode, PAstDiag(kind: adSemDefNameSym,
+                                        defNameSym: sym,
+                                        defNameSymData: defData))
+
+  case n.kind
+  of nkSym:
+    result =
+      # xxx: why don't we guard on `sfGenSym`?
```
zerbina: We should, but currently can't. Macros can "transplant" mismatching non-gensyms into definition positions.

compiler/sem/sem.nim
```diff
@@ -351,6 +351,120 @@ proc newSymS(kind: TSymKind, n: PNode, c: PContext): PSym =
...
+        n.sym.owner = currOwner # xxx: modifying the sym owner is suss
+        n
+      else:
+        let recoverySym = copySym(n.sym, nextSymId c.idgen)
+        recoverySym.transitionRoutineSymKind(kind)
+        recoverySym.owner = currOwner
+        c.config.makeError(n, recoverySym, ExpectedKindMismatch)
+  of nkIdentKinds - {nkSym}:
+    # xxx: sym choices qualify here, but shouldn't those be errors in
+    #      definition positions?
```
zerbina, line 425: Yep, only `nkIdent` and `nkAccQuoted` should be allowed in this branch.

compiler/sem/semstmts.nim
```diff
@@ -2400,58 +2405,59 @@ proc canonType(c: PContext, t: PType): PType =
   else:
     result = t
 
-proc prevDestructor(c: PContext; prevOp: PSym; obj: PType; info: TLineInfo) =
-  if sfOverriden notin prevOp.flags:
-    localReport(c.config, info, reportSym(
-      rsemRebidingImplicitDestructor, prevOp, typ = obj))
+proc semOverride(c: PContext, s: PSym, info: TLineInfo) =
+  ## processes the type operation override defined by `s`ymbol, binding the
+  ## procedure to appropriate type, and reporting legacy errors.
```
```suggestion
  ## processes the type operation override defined by `s`ymbol, and either
  ## binds the procedure to the appropriate type or reports legacy errors.
```
It's slightly more clear this way, I think.

compiler/sem/semstmts.nim
```diff
@@ -3114,8 +3125,11 @@ proc semRoutineDef(c: PContext, n: PNode): PNode =
   result[namePos] =
     semRoutineName(c, n[namePos], kind, allowAnon = kind in AllowAnon)
 
-  if result[namePos].isError:
-    return c.config.wrapError(result) # early out
+  if result[namePos].kind == nkError:
+    if result[namePos].diag.kind == adSemDefNameSym:
+      discard "don't leave early, those methods are updated"
```
```suggestion
      discard "don't leave early, we can still make progress"
```
A very soft suggestion.

zerbina (commented):
Looks good to me. Adding routines with erroneous name slots to the symbol table feels like it's going to become a problem, but that's something that can be fixed later on.

I left a few comments, non-blocking suggestion that I think would make sense to have from the get-go.

compiler/sem/sem.nim
saem: Yeah, there are a few things going on, but we can't solve them here. It's partly being able to differentiate between first phase lookup vs macro `genSym` and guard/validate those to be true, and in the future changing macro `genSym` to unique identifier, or something.

compiler/sem/sem.nim
saem: Yup, which works in the other direction too, where they could transplant an otherwise valid definitional `nkSym` into another definition spot and it's _really_ hard to tell that that happened.

compiler/sem/sem.nim
saem: Hmm, I'll check what happens with tests when I tighten up the condition and generate an error instead.

compiler/sem/semstmts.nim
saem: Will do, thanks.

compiler/sem/semstmts.nim
saem: oh yeah, that's out of date, will fix.

compiler/sem/semstmts.nim
saem: update

compiler/sem/semstmts.nim
saem: update

saem:
One final run through CI and it should be good to go.

PR #619, opened 2023-04-06
semtypinst: fix `cannot instantiate` issues

zerbina:
## Summary

Fix an issue with the type-variable replacing logic that caused unrelated routines and types to be instantiated, leading to `cannot instantiate` errors with default parameter expressions and type- parameter-dependent `when`-condition expression.

## Details

During instantiation of a type or routine, and replacing references to unresolved types with the corresponding bound ones, bound symbols (`nkSym`) referring to unrelated entities like routines or types were not ignored (except for symbols in callee slots -- those were special cased), which lead to the type variable replacing logic traversing into unrelated types.

Besides being unnecessary work, this also caused spurious `cannot instantiate` errors.

`prepareNode` is adjusted to ignore all symbols not belonging to generic or non-generic parameters, preventing unrelated types from being instantiated -- a temporary workaround is required in order to still support accessing parameters of generic types.

In addition, type variable replacement of default parameter expression was incorrectly done via `replaceTypeVarsN` instead of via `prepareNode`. This is now also fixed, and the documentation of the respective procedures improved.

---

## Notes for Reviewers
* this PR removes some pre-existing workarounds, but unfortunately introduces some new ones -- fixing the underlying issues is a much more involved task, however  
* the focus is on unblocking both the `nkBracketExpr` rewriting removal and the generic call rework

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semtypinst.nim
+++ b/compiler/sem/semtypinst.nim
@@ -148,19 +148,73 @@ proc replaceTypeVarsT*(cl: var TReplTypeVars, t: PType): PType =
   result = replaceTypeVarsTAux(cl, t)
   checkMetaInvariants(cl, result)
 
-proc prepareNode(cl: var TReplTypeVars, n: PNode): PNode =
-  let t = replaceTypeVarsT(cl, n.typ)
-  if t != nil and t.kind == tyStatic and t.n != nil:
-    return if tfUnresolved in t.flags: prepareNode(cl, t.n)
-           else: t.n
-  result = copyNode(n)
-  result.typ = t
-  if result.kind == nkSym: result.sym = replaceTypeVarsS(cl, n.sym)
-  let isCall = result.kind in nkCallKinds
-  for i in 0..<n.safeLen:
-    # XXX HACK: ``f(a, b)``, avoid to instantiate `f`
-    if isCall and i == 0: result.add(n[i])
-    else: result.add(prepareNode(cl, n[i]))
+proc isTypeParameterAccess(cl: TReplTypeVars, s: PSym): bool =
+  # XXX: this is a workaround used for detecting whether the ``skType``
+  #      symbols represents a generic type parameter access
+  s.kind == skType and s.typ.kind == tyTypeDesc and
+    cl.typeMap.lookup(s.typ[0]) != nil
+
+proc prepareNode*(cl: var TReplTypeVars, n: PNode): PNode =
+  ## Replaces unresolved types referenced by the nodes of expression/statement
+  ## `n` with the corresponding bound ones.
+  ##
+  ## Because of how late-bound static parameters are currently implemented,
+  ## ``prepareNode`` is also responsible for replacing sub-expression of type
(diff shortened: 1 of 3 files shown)


compiler/sem/semtypinst.nim
```diff
@@ -148,19 +148,73 @@ proc replaceTypeVarsT*(cl: var TReplTypeVars, t: PType): PType =
...
+        {tfGenericTypeParam, tfImplicitTypeParam} * s.typ.flags != {}) or
+       isTypeParameterAccess(cl, s):
+      resolveStatic()
+      result = copyNode(n)
+      result.typ = t
+      result.sym = replaceTypeVarsS(cl, s)
+    else:
+      # important: don't use ``resolveStatic`` here, as the ``replaceTypeVarsT``
+      # call would traverse into typestypes not depending on any of the type
+      # variables we're replacing here
```
```suggestion
      # important: don't use ``resolveStatic`` here, as the ``replaceTypeVarsT``
      # call would traverse into types not depending on any of the type
      # variables we're replacing here
```

tests/lang_callable/generics/tgenerics_issues.nim
```diff
@@ -872,3 +872,46 @@ block: # Ensure no segfault from constraint
...
+
+block uninstantiated_symbol_in_default_value_ast:
+  # `prc` must have a parameter with a non-builtin container type where the
+  # relevant type parameter is not phantom type information
+  proc prc[T](x: Slice[T]) =
+    discard
+
+  template imm() =
+    # `imm` must be an immediate template, so that it is instantiated during
+    # when analysing the default value expression
```
saem:
```suggestion
    # `imm` must be an immediate template, so that it is instantiated when
    # analysing the default value expression
```

tests/lang_callable/generics/tgenerics_issues.nim
```diff
@@ -872,3 +872,46 @@ block: # Ensure no segfault from constraint
...
+    bind prc # <-- this is the imporant part. The symbol of an uninstantiated
+             # generic routine exists in a non-callee AST slot
+
+  # `param` must have a type that depends on a generic parameter
+  proc p[T](param = (imm(); default(T))) =
+    doAssert param == 0
+
+  # this failed with a "cannot instantiate; Maybe generic arguments are
+  # missing" error
```
saem:
```suggestion
  # this used to fail with a "cannot instantiate; Maybe generic arguments are
  # missing" error
```

saem (approved):
This makes sense, I did a quick read through for `replaceTypeVarAuxTAux` and then checked the callsites for `makeTypeFromExpr` and definitely see how things would be simplified with some consistency wrt static param handling. Without that in place, I think this heuristic is better than what came before, and the test coverage has improved so any proper/further fixes will have more of a safety net.

Nice work!

zerbina:
Thank you for the review @saem. I updated the comments with the suggested fixes.

saem:
@zerbina Thanks!

bors r+

🚀

PR #620, opened 2023-04-06
compiler: improve macro parameter handling

zerbina:
## Summary

This commit introduces internal signatures for macros, removing conditional logic from parameter symbol handling and generally moving the details of how macro parameters work out of `vmgen` and more into a single place in `sem`.

As two side-effect of the above:
- inspecting parameter types from a macro's header (via, for example, `typeof`) now yields the types as specified, instead of as `NimNode`
- the types (or values, in case of `static[T]`) passed to a macro's generic parameters can now be inspected by the macro

## Details

The symbols belonging to macro parameters were previously replaced with new ones that used either `NimNode` or, in the case of `static[T]`, the base type as the type, prior to adding them to the symbol table.

Replacing the symbols at some point is necessary in order for semantic analysis inside the macro body to view them as the correct type, but the previous implementation had two issues:
1. the parameters were only visible as `NimNode` to other types or expressions referencing them
2. the symbols were freestanding. They weren't registered anywhere, meaning that it was not possible to, for example, map all parameter symbols via their ID to some other entity

To fix both issues and in order for the implementation to be less scattered, a different approach is now used. Macro parameter handling is moved out of routine header analysis into `semMacroDef` (fixing issue 1). This allows for a more flexible implementation (more contextual information is available) and also reduces the amount of context required by `addParam`, etc.

So that they can be later queried, the created internal symbols are added as parameters to a dedicated `tyProc` type that represents the macro's internal signature, which is then stored as part of the macro's symbol.

### Changes

- remove macro parameter handling from `addParamOrResult` and `addResult`
- add a dedicated branch for `skMacro` to `TSym` and adjust the places depending on the previous layout
- create the internal symbols for macro parameters in `semMacroDef`, where they're also used to populate the internal signature
- add a `transformBody` variant that doesn't query nor updates the symbols `transformBody` field
- change `vmgen` to consider the macros internal signature, making the previously used workarounds obsolete
- rewrite `getAst(macro(a, b, c))` during `mirgen`, slightly simplifying the related logic in `vmgen`
- remove the now unused `adVmGenBadExpandToAstCallExprRequired` diagnostic

---

## Notes for Reviewers
* this removes the last remaining blocker for #548

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -981,7 +981,7 @@ type
   AstDiagVmGenKind* = enum
     ## Kinds for errors produced by `vmgen`
     adVmGenBadExpandToAstArgRequired      # | TODO: these enum values duplicate
-    adVmGenBadExpandToAstCallExprRequired # |       `VmGenDiagKind` vmgen enum
+                                          # |       `VmGenDiagKind` vmgen enum
     adVmGenTooManyRegistersRequired       # |       defined in the `vmdef`
     adVmGenCannotFindBreakTarget          # |       module. There should be a
     adVmGenNotUnused                      # |       way to cross-reference data
@@ -1001,7 +1001,6 @@ type
   AstDiagVmGenError* = object
     case kind*: AstDiagVmGenKind:
       of adVmGenBadExpandToAstArgRequired,
-          adVmGenBadExpandToAstCallExprRequired,
           adVmGenTooManyRegistersRequired,
           adVmGenCannotFindBreakTarget:
         discard
@@ -1595,10 +1594,15 @@ type
   TSym* {.acyclic.} = object of TIdObj # Keep in sync with PackedSym
     ## proc and type instantiations are cached in the generic symbol
     case kind*: TSymKind
-    of routineKinds:
+    of routineKinds - {skMacro}:
       #procInstCache*: seq[PInstantiation]
       gcUnsafetyReason*: PSym  ## for better error messages regarding gcsafe
       transformedBody*: PNode  ## cached body after transf pass
+    of skMacro:
+      internal*: PType ## the internal signature that the macro has in a
+                       ## compile-time evaluation context. Can be used to
+                       ## query the symbols the parameters use in the macro's
+                       ## body
     of skLet, skVar, skField, skForVar:
       guard*: PSym
       bitsize*: int
--- a/compiler/sem/sempass2.nim
+++ b/compiler/sem/sempass2.nim
@@ -273,12 +273,13 @@ proc warnAboutGcUnsafe(n: PNode; conf: ConfigRef) =
 proc markGcUnsafe(a: PEffects; reason: PSym) =
   if not a.inEnforcedGcSafe:
     a.gcUnsafe = true
-    if a.owner.kind in routineKinds: a.owner.gcUnsafetyReason = reason
+    if a.owner.kind in routineKinds - {skMacro}:
+      a.owner.gcUnsafetyReason = reason
 
 proc markGcUnsafe(a: PEffects; reason: PNode) =
   if not a.inEnforcedGcSafe:
     a.gcUnsafe = true
-    if a.owner.kind in routineKinds:
+    if a.owner.kind in routineKinds - {skMacro}:
       if reason.kind == nkSym:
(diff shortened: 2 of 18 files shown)


compiler/ast/ast_types.nim
```diff
@@ -981,7 +981,7 @@ type
   AstDiagVmGenKind* = enum
     ## Kinds for errors produced by `vmgen`
     adVmGenBadExpandToAstArgRequired      # | TODO: these enum values duplicate
-    adVmGenBadExpandToAstCallExprRequired # |       `VmGenDiagKind` vmgen enum
+                                          # |       `VmGenDiagKind` vmgen enum
```
saem, line 984: 🎉 less code!

compiler/ast/ast_types.nim
```diff
@@ -1595,10 +1594,15 @@ type
   TSym* {.acyclic.} = object of TIdObj # Keep in sync with PackedSym
     ## proc and type instantiations are cached in the generic symbol
     case kind*: TSymKind
-    of routineKinds:
+    of routineKinds - {skMacro}:
       #procInstCache*: seq[PInstantiation]
       gcUnsafetyReason*: PSym  ## for better error messages regarding gcsafe
       transformedBody*: PNode  ## cached body after transf pass
+    of skMacro:
```
saem, line 1601: I really like this a lot -- it moves macros much closer to "just" a function/procedure.

I can easily imagine building upon this incrementlaly in two very interesting ways:
1. return types of macros as "typed code values" (we know this is an `int` expression or `noreturn` stmt)
2. analysing macros for effects: does it only observe ast, does it change ast, does it read/write macrocache, etc
These could then be combined to determine whether we can cache macro results, trust a macro's output in sem, defer macro evaluation until generic instantiation, and more

_very cool_

compiler/sem/sempass2.nim
```diff
@@ -273,12 +273,13 @@ proc warnAboutGcUnsafe(n: PNode; conf: ConfigRef) =
 proc markGcUnsafe(a: PEffects; reason: PSym) =
   if not a.inEnforcedGcSafe:
     a.gcUnsafe = true
-    if a.owner.kind in routineKinds: a.owner.gcUnsafetyReason = reason
+    if a.owner.kind in routineKinds - {skMacro}:
```
saem, line 276: Should happen separately from this PR, but it might be time to start thinking of a new constant name, like `routineKindsWithGC` to absorbe some of the `- {skMacro}`.

tests/lang_callable/macros/tmacros_issues.nim
```diff
@@ -519,3 +519,16 @@ block double_sem_for_procs:
     result = 10.0
 
   discard exp(5.0)
+
+block orginal_parameter_types:
+  # inspecting a parameter type in the macro's header yields the
+  # original type, not ``NimNode``
+  macro m(x: int, y: typeof(x)) =
```
saem:
```suggestion
  template assertTypeOfInt(s): typedesc =
    doAssert typeof(s) is int
    typeof(s)
  # inspecting a parameter type in the macro's header yields the
  # original type, not ``NimNode``
  macro m(x: int, y: assertTypeOfInt(x)) =
```

It's a very narrow case, but this might help detect issues, if there were a cascade of bugs resulting in: `typeof(x)` is evaluated to `NimNode` and things get passed through for `y` regardless.

saem (approved):
This is really good, can't emphasize enough how good the direction is with how macros are treated (so much closer to "boring" functions/procedures). The amount of code that's deleted and fewer special cases also makes things much easier to understand.

🎉 

Left a few suggestions, all soft.

compiler/sem/sempass2.nim
zerbina: Yeah, I agree. When doing so, it might make sense to also exclude `skTemplate` from the set -- templates don't have a GC unsafety reason nor are they transformed.

tests/lang_callable/macros/tmacros_issues.nim
zerbina: Good call. I updated it.

saem:
Bors r+

PR #621, opened 2023-04-06
sem: rework explicit generic calls

zerbina:
## Summary

Rework and unify the handling of calls with explicitly provided type bindings (e.g.: `someRoutine[int, float]()`), significantly improving the usability of generic routines and fixing a large amount of related issues/bugs.

The core of the change is, that generic routines are now, except if explicitly requested otherwise, instantiated *after* overload resolution, instead of *before*. Forcing early instantiation is still possible via the `(someRoutine[int, float])()` syntax.

Since `void` parameter erasure only applies when when using early instantiation, code depending on the erasure needs to be adjusted.

Language changes/clarifications
-------------------------------

- default values for generic parameters on routines are now supported
- alias macros/templates are no longer considered for expansion in the callee position of the `someRoutine[A, B](c)` construct, making the behaviour consistent with that of normal call expressions
- the `someRoutine[]()` syntax is now valid. Instead of being treated as a dereference operation, it can now be used to force overload resolution to pick a generic routine (which includes nullary generics)
- using parenthesized callee expressions is now supported when the callee evaluates to a symbol choice
- the `T = x` (`nkExprEqExpr`) syntax is disallowed for type arguments

Fixed issues
------------

- fix passing values to type arguments with the `static[T]` not working when the callee is a template/macro, or the call located inside a template body / generic routine. A `type expected, but got` error was previously reported when attempting to pass the type argument
- all generic routines are now considered during overload resolution, regardless of which routine lookup finds first

### Previous implementation

Invocation of a routine where explicit generic type arguments where provided was handled by the compiler in two different ways:
1. explicit instantiation + overload resolution using the resulting instantiations (also referred to as early or eager instantiation here)
2. normal overload resolution (also referred to as late instantiation here)

How generic calls were treated was very dependent on the first symbol found by lookup. Early instantiation instantiation happened when:
- the first routine found by qualified lookup is not a template or macro
- the first non-template/macro routine in the candidate set had implicit generic parameters, and none of them were explicitly supplied
- the call expression was located in either the body of a template or generic routine (macros excluded)
while normal overload resolution was chosen in all other cases.

The early instantiation route worked by: gathering a set with all overloads; discarding all templates, macros, and routines with a different number of expected generic parameters (both explicit and implicit) than provided generic arguments; and generating an instantiation
for each remaining routine where the generic arguments matched the formal types.

If the resulting set was empty, a `cannot instantiate` error was produced. Otherwise, the closed set with the instantiated routines was passed on to normal overload resolution.

When taking the late instantiation route, the call expression with the callee being an `nkBracketExpr` node is passed directly to `resolveOverloads`, where the expressions in the bracket are analysed as normal operand expressions and the *types* of the analysed expressions are used as the initial bindings. No arity nor formal vs. argument type checking took place in this case, which allowed for semantically invalid programs to compile and possibly run.

There was another quirk with the late instantiation route: if the first found symbol is a macro or template, or the call expression is coming from a template/generic instantiation macros, each expression in the brackets was required to be a type expression (which prevents most valid `static` expressions), making it effectively impossible to use routines with explicit generic parameters using `static[T]` in these contexts.

### Now

* analyse and resolve all call expressions of the exact form `someRoutine[A, B](c)` through normal overload resolution
* extend `sigmatch` to perform the matching of generic arguments (if present) against the generic parameters -- a mismatch results in the overload to be rejected
* use the same analysis for generic arguments in the context of explicit instantiation (`explicitInstantiation`) and overload resolution (`resolveOverloads`)
* use dedicated processing for the generic call syntax in `semIndirectOp`, which allows for `someRoutine[]()` to work and also prevents alias templates/macros to be expanded in the callee symbols slot

### Other changes

* an implicit dereference is now inserted for `x.field()` where `field` is a view, making the behaviour of `nkDotExpr` consistent with other expressions in the context of indirect calls

### `void` parameter erasure

`sigmatch` could be adjusted to treat `void` parameters as effectively non-existing, though this wouldn't make `proc` types using generic parameters that are bound to `void` work.

### Error messages

When not explicitly requested, no instantiation of generic routines takes place prior to overload resolution, meaning that, for example, a type argument mismatch doesn't produce a `cannot instantiate` error anymore.

The `MismatchInfo` is extended to also include information about generic-parameter-related errors, but rendering of the error is still very basic.

---

## To-do
- [X] ~match default parameter values against the formal type~
- [X] add regression tests for the fixed issues

## Notes for Reviewers
* improving the error messages makes more sense as separate PR, I think
* the changeset is quite small compared to its impact
* I'm not sure if the language changes / additions should be reflected in the manual yet

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semcall.nim
+++ b/compiler/sem/semcall.nim
@@ -34,7 +34,6 @@ proc sameMethodDispatcher(a, b: PSym): bool =
       # instantiated.
 
 proc initCandidateSymbols(c: PContext, headSymbol: PNode,
-                          initialBinding: PNode,
                           filter: TSymKinds,
                           best, alt: var TCandidate,
                           o: var TOverloadIter): seq[tuple[s: PSym, scope: int]] =
@@ -47,16 +46,13 @@ proc initCandidateSymbols(c: PContext, headSymbol: PNode,
       localReport(c.config, symx.ast)
     symx = nextOverloadIter(o, c, headSymbol)
   if result.len > 0:
-    initCandidate(c, best, result[0].s, initialBinding,
-                  result[0].scope)
-    initCandidate(c, alt, result[0].s, initialBinding,
-                  result[0].scope)
+    initCallCandidate(c, best, result[0].s, result[0].scope)
+    initCallCandidate(c, alt, result[0].s, result[0].scope)
     best.state = csNoMatch
 
 proc pickBestCandidate(c: PContext,
                        headSymbol: PNode,
                        n: PNode,
-                       initialBinding: PNode,
                        filter: TSymKinds,
                        best, alt: var TCandidate,
                        errors: var seq[SemCallMismatch]) =
@@ -81,8 +77,8 @@ proc pickBestCandidate(c: PContext,
   while sym != nil:
     if sym.kind in filter:
       # Initialise 'best' and 'alt' with the first available symbol
-      initCandidate(c, best, sym, initialBinding, scope)
-      initCandidate(c, alt, sym, initialBinding, scope)
+      initCallCandidate(c, best, sym, scope)
+      initCallCandidate(c, alt, sym, scope)
       best.state = csNoMatch
       break
     else:
@@ -96,7 +92,7 @@ proc pickBestCandidate(c: PContext,
       continue
     c.config.internalAssert(sym.typ != nil or sym.magic != mNone,
                             "missing type information")
-    initCandidate(c, z, sym, initialBinding, scope)
+    initCallCandidate(c, z, sym, scope)
     if c.currentScope.symbols.counter == counterInitial or syms.len != 0:
       matches(c, n, z)
       if z.state == csMatch:
@@ -118,8 +114,7 @@ proc pickBestCandidate(c: PContext,
     else:
       # Symbol table has been modified. Restart and pre-calculate all syms
       # before any further candidate init and compare. SLOW, but rare case.
-      syms = initCandidateSymbols(c, headSymbol, initialBinding, filter,
-                                  best, alt, o)
+      syms = initCandidateSymbols(c, headSymbol, filter, best, alt, o)
       noSyms = false
     if noSyms:
       sym = nextOverloadIter(o, c, headSymbol)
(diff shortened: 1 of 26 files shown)


compiler/sem/semcall.nim
```diff
@@ -283,34 +279,82 @@ proc getMsgDiagnostic(
...
+  ## formal parameters), we fully analyse them here instead of deferring
+  ## that to sigmatch. This has the benefit that we only need to analyse/
+  ## evaluate each argument once, instead of for each matched against overload.
+  assert n.kind == nkBracketExpr
+  result = shallowCopy(n)
+  result[0] = n[0]
+
+  var hasError = false
+  for i in 1..<n.len:
+    # a generic argument must be either evaluate to a type (``tyTypedesc``) or
```
```suggestion
    # a generic argument must either evaluate to a type (``tyTypedesc``) or
```

compiler/sem/semcall.nim
```diff
@@ -283,34 +279,82 @@ proc getMsgDiagnostic(
...
-    initialBinding = f
-    if hasError:
-      f = c.config.wrapError(f)
+    # the call has explicit generic arguments specified
+    let args = semGenericArgs(c, f)
+    if args.kind == nkError:
+      f = args
     else:
-      f = f[0]
+      n[0] = args
```
saem, line 345: I wonder if manipulating `n[0]` like this may result in overloaded call operator macros/templates not being able to receive the callee name position as untyped?

Edit: Oh wait... I'm basing this off my older understanding, perhaps the paths into here are guarded for that.

compiler/sem/semcall.nim
```diff
@@ -429,19 +473,33 @@ proc inferWithMetatype(c: PContext, formal: PType,
...
     if nfDefaultParam in call[i].flags:
       let def = calleeParams[i].sym.ast
       if nfDefaultRefsParam in def.flags: call.flags.incl nfDefaultRefsParam
-      call[i] = def
+      call[i] =
+        case def.kind
+        of nkEmpty:
+          # ``instantiateProcType`` uses ``nkEmpty`` to signal an incompatible
+          # default expression
+          conf.newError(calleeParams[i],
```
saem, line 493:
```suggestion
          hasError = true
          conf.newError(calleeParams[i],
```

I might have missed it, but I think that's required.

saem (commented):
Overall, the intended semantics as laid out in the PR descriptin jives with the picture in my head of how things should work. The void parameter handling and template/macro expansion in callee symbol positions are also the right call for the purposes of this PR.

I did an initial pass of the code. There is a potential with some AST manipulation, but I might not be following the code correctly. This should surface in the call operator tests (`tests/specialops/tcallops` and friends).

I need to run through the code again to provide more insightful feedback.

compiler/sem/semcall.nim
zerbina: All paths that reach this mutation don't lead back to the call operator application, so these are not affected.

A callee bracket expression that represents the passed type arguments in the context of a generic routine call now redirects to `semDirectOp` in `semIndirectOp`.

Nonetheless, the mutation of input AST is a problem (although pre-existing; see `semOpAux`). Two possible solutions:
1. `copyTree` the call expression at the start of `resolveOverloads`
2. store the analysed operands in a separate list (or `PNode`) that is passed to `sigmatch`

compiler/sem/semcall.nim
zerbina: Yep, it's required.

compiler/sem/semcall.nim
saem: No action for this comment, thanks for clarifying.

---

I wish I had an easier time teaching that/keeping the fan in/out of these recursive calls straight in my head. There is some sort of memory technique I'm missing about how to easily build up/hold that picture in my head.

---

Tangent: there is something pretty interesting about option 2 (separate analysed operand storage), but I think it needs an analysis level marker on the node. It would tells us that certain parts of `sons` aren't part of the AST if we raise the node level back up. At which point we could have "metadata" in certain node positions. 

Alternatively, we have a new node kind (maybe `nkArgument`) that can hold this stuff in various positions. It could be nicely sanitized by extracting the nodes from a dedicated untyped sons position.

saem:
Marking `tparamsindefault` as a known issue is a reasonable approach.

Alternatively, a fix is also possible, in `semtypes`, in the `semProcTypeNode -> semParType -> semTypeNode -> nkSym branch` chain. In there, when the symbol is one of `skParam/Local/Let/Var/etc`, basically any storage, then we're taking a compile time type value and moving it to type land. Here we can unwrap one level of `typedesc` to determine the type. This logic also applies to the `nkIdent/nkAccQuoted` branch.

compiler/sem/semexprs.nim
```diff
@@ -2048,6 +2048,9 @@ proc semSubscript(c: PContext, n: PNode, flags: TExprFlags): PNode =
             elif n[0].kind in nkSymChoices: n[0][0].sym
             else: nil
     if s != nil:
+      # XXX: how overloadable symbols are handled here is problematic, as what
```
saem, line 2051: thanks for calling this out, it's a good reminder.

tests/lang_callable/generics/tlate_bound_typedesc.nim
```diff
@@ -0,0 +1,15 @@
+discard """
+  errmsg: "The default parameter 'arg' has incompatible type with the explicitly requested proc instantiation"
+  line: 12
+"""
+
+# original issue https://github.com/nim-lang/nim/issues/11660
+
+# XXX: this semantics of ``typedesc`` as the parameter type are very likely
```
saem, line 8: This is a good description of the issue, and when the behaviour is fixed the test will be a good reminder.

saem (approved):
Outside of any quibbles from CI, this is good to go: 🚀

zerbina:
The `tv2_cast.nim` test previously failed because the order of the definitions changed. This is because `injectdestructors` iterates over an unordered table that uses symbol-IDs as part of the key, when moving `def` statements. With less instantiations (and thus symbols) being generated now, the order of the table items changed. 

I adjusted the test for now, but a follow-up PR making the pass' output symbol-ID-independent is required. Note that only the order of the definition statements is affected, but not the order in which destructors are injected, meaning that the issue is (currently) purely cosmetic.

zerbina:
The PR is squashed and rebased -- I also added two more regression tests. If CI doesn't report any errors, I'm going to go ahead and merge it.

saem:
Looks to be spurious, re-running the job.

PR #623, opened 2023-04-10
vm: fix cell access checks

zerbina:
## Summary

Change the cell access validation logic to properly check whether the provided pointer points inside the cell. If the VM was not built with assertions enabled, some use-after-free accesses would either crash the VM (host) or remain undetected.

The pointer not pointing inside the cell is a situation that is possible during correct VM operation (e.g., when a cell is reused), but it was incorrectly treated as an internal programmer error (assertions were used) instead of a guest error.

### DIFF
--- a/compiler/vm/vmchecks.nim
+++ b/compiler/vm/vmchecks.nim
@@ -97,12 +97,20 @@ func testLocationType(locType: PVmType, t: PVmType): AccessViolationReason =
 
 
 func checkValid(p: VmMemPointer, typ: PVmType, cell: VmCell): AccessViolationReason =
-  assert p.rawPointer >= cell.p, "memory not part of `cell`"
-
-  let off = cast[int](p) - cast[int](cell.p)
-  if off > 0:
+  ## Tests whether `p` points to a valid memory location inside `cell` that is
+  ## accesible as type `typ`, and returns either a success or the reason for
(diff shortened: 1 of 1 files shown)


compiler/vm/vmchecks.nim
```diff
@@ -97,22 +97,29 @@ func testLocationType(locType: PVmType, t: PVmType): AccessViolationReason =
...
+      result = avrOutOfBounds
+    elif typ != nil:
       # the cell stores either a single- or a contiguous sequence of locations
       let stride = cell.typ.alignedSize.int
       result = searchInArray(toOpenArray(cell.p, 0, int(cell.sizeInBytes-1)), cell.typ, typ, cell.count, stride, off)
     else:
       # untyped memory; not yet supported
       unreachable()
-  else:
+  elif p.rawPointer == cell.p:
```
saem: I'd personally lead with this branch as it's the most straightforward case and for the reader eliminates that, then they're primed to think about edge cases (greater and less than).

(very soft suggestion. for example: if you need this change in sooner and don't want to bother with the CI time, then I'd ignore it)

saem (approved):

compiler/vm/vmchecks.nim
zerbina: I reordered the branches -- it definitely reads better now.

PR #624, opened 2023-04-10
vm: improve and fix numeric conversion support

zerbina:
## Summary

Rework the handling of conversion between two numeric values (integers, floats, enums, etc.) in the VM and `vmgen`, fixing two severe issues:
- unsigned integers greater than 2^63 being treated as signed values during unsigned-to-float conversions
- range checks being erroneously inserted for X-to-unsigned conversions

Numeric conversions now work the same for the VM target as they do with the C target. In addition, they no longer use the high-quality RTTI, thus making them more efficient.

## Details

One problem with the previous `Conv` operation was, that the semantics regarding the destination register were dependent on the target type. For to-`string` and to-`openArray` conversions, the destination register was expected to store a handle, while for the numeric conversions it was an "out" register (meaning that its content is overwritten).

Because the operation unconditionally validated the handle stored in destination (even it was only meant as an "out" register), this would lead to spurious access violation errors if the destination register stores a stale handle.

In addition, numeric conversions didn't properly clean up the destination register, which was a cause of memory leaks -- these are fixed now.

- introduce the new `NumConv` operation, which is used for conversion involving floats, and for the to-bool conversion
- adjust numerical conversion handling in `vmgen` to emit `NumConv` where applicable
- remove the numeric-conversion-related logic from `opConv`
- change the `Conv` and `Cast` operation to use two instruction words instead of three
- only emit `RangeCheck` instructions for `nkChckRange` expression if the destination is not an unsigned integer, mirroring the C code- generator's logic
- treat `range`-to-base-type conversions as a no-op in `vmgen`
- split the `cast` generation logic into a separate procedure
- unsupported `cast`s are now reported during code generation instead of at run-time

---

## Notes for Reviewers
* blocks #548, as the changes there trigger the "Conv destination register is stale handle" case 
* this change makes further progress towards the eventual removal of `opcConv`
* changing the encoding of `opcConv` and `opcCast` instructions was also possible prior to separating the numeric conversion

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -84,7 +84,7 @@ when defined(nimVMDebugGenerate):
 #      executed instruction
 
 import std/options as stdoptions
-from std/math import round
+from std/math import round, copySign
 
 type
   VmThread* = object
@@ -686,51 +686,6 @@ proc opConv(c: var TCtx; dest: var TFullReg, src: TFullReg, dt, st: (PType, PVmT
   else:
     let desttyp = skipTypes(desttyp, abstractVarRange)
     case desttyp.kind
-    of tyInt..tyInt64:
-      dest = TFullReg(kind: rkInt)
-      case skipTypes(srctyp, abstractRange).kind
-      of tyFloat..tyFloat64:
-        dest.intVal = int(src.floatVal)
-      else:
-        dest.intVal = src.intVal
-      if toInt128(dest.intVal) < firstOrd(c.config, desttyp) or toInt128(dest.intVal) > lastOrd(c.config, desttyp):
-        return true
-    of tyUInt..tyUInt64, tyChar: # char is also an unsigned type
-      dest = TFullReg(kind: rkInt)
-      let styp = srctyp.skipTypes(abstractRange) # skip distinct types(dest type could do this too if needed)
-      case styp.kind
-      of tyFloat..tyFloat64:
-        dest.intVal = int(src.floatVal)
-      else:
-        let srcDist = (sizeof(src.intVal) - styp.size) * 8
-        let destDist = (sizeof(dest.intVal) - desttyp.size) * 8
-        var value = cast[BiggestUInt](src.intVal)
-        value = (value shl srcDist) shr srcDist
-        value = (value shl destDist) shr destDist
-        dest.intVal = cast[BiggestInt](value)
-    of tyBool:
-      dest = TFullReg(kind: rkInt)
-      dest.intVal =
-        case skipTypes(srctyp, abstractRange).kind
-          of tyFloat..tyFloat64: int(src.floatVal != 0.0)
-          else: int(src.intVal != 0)
-    of tyFloat, tyFloat64:
-      dest = TFullReg(kind: rkFloat)
-      case skipTypes(srctyp, abstractRange).kind
-      of tyInt..tyInt64, tyUInt..tyUInt64, tyEnum, tyBool, tyChar:
-        dest.floatVal = toBiggestFloat(src.intVal)
-      else:
-        dest.floatVal = src.floatVal
-    of tyFloat32:
-      # convert to `float32` first and then to `BiggestFloat`
-      dest = TFullReg(kind: rkFloat)
-      case skipTypes(srctyp, abstractRange).kind
-      of tyInt..tyInt64, tyUInt..tyUInt64, tyEnum, tyBool, tyChar:
-        dest.floatVal = src.intVal.float32.BiggestFloat
-      of tyFloat, tyFloat64:
-        dest.floatVal = src.floatVal.float32.BiggestFloat
-      else:
-        dest.floatVal = src.floatVal
     of tyOpenArray:
       assert dt[1].kind == akSeq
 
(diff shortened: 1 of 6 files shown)


compiler/vm/vm.nim
```diff
@@ -763,6 +718,44 @@ proc opConv(c: var TCtx; dest: var TFullReg, src: TFullReg, dt, st: (PType, PVmT
...
+                    floatVal: toFloat(cast[BiggestUInt](src.intVal), destbytes))
+  of nckFToF: # float-to-float
+    dest = TFullReg(kind: rkFloat,
+                    floatVal: toFloat(src.floatVal, destbytes))
+  of nckToB:  # float or int to bool
+    dest = TFullReg(kind: rkInt)
+    dest.intVal =
+      case src.kind
+      of rkInt:   ord(src.intVal   != 0)
+      of rkFloat: ord(src.floatVal != 0)
```
saem: (not the end of the world)

IIUC, this leaves treatment of negative zero upto the particular compiler that compiled the VM, so there is some UB. Said compiler might not be the same as what will be used to compile the user program, if my understanding is correct, we may want to force it to work one way (my instinct: positive zero is false, negative zero is true).

compiler/vm/vmdef.nim
```diff
@@ -1052,7 +1059,23 @@ template `currentException=`*(a: VmArgs, h: HeapSlotHandle) =
...
+                      ): tuple[op: NumericConvKind, dstbytes, srcbytes: int] =
+  ## Unpacks the numeric conversion description from `info`.
+  result.op = NumericConvKind(info and 0x7)
+  # note: add 1 to undo the shifting done during packing
+  result.dstbytes = 1 + int((info shr 3) and 0x7)
+  result.srcbytes = 1 + int((info shr 6) and 0x7)
+
+func packedConvDesc*(op: NumericConvKind, dstbytes, srcbytes: range[1..8]): uint16 =
+  ## Packs the numeric conversion description into a single ``uint16``.
+  debugEcho op, "; ", dstbytes, "; ", srcbytes
```
saem: leftover debug statement?

compiler/vm/vmgen.nim
```diff
@@ -2768,18 +2852,32 @@ proc gen(c: var TCtx; n: PNode; dest: var TDest; flags: TGenFlags = {}) =
...
-      tmp1 = c.genx(n[1])
-      tmp2 = c.genx(n[2])
-    c.gABC(n, opcRangeChck, tmp0, tmp1, tmp2)
-    c.freeTemp(tmp1)
-    c.freeTemp(tmp2)
-    if dest >= 0:
-      gABC(c, n, whichAsgnOpc(n), dest, tmp0)
+    let tmp0 = c.genx(n[0])
+    # XXX: range checks currently always happen, even if disabled by the user.
+    #      Once the range check injection logic is a MIR pass, this should be
```
```suggestion
    #      Once the range check injection logic is an MIR pass, this should be
```

saem (commented):
The design is good.

I left a minor suggestion (typo), possible leftover debug stmt, and there is a potential undefined behaviour which might merit tightening.

compiler/vm/vmdef.nim
zerbina: Good catch -- it is.

compiler/vm/vmgen.nim
zerbina: Hm, but "MIR" doesn't start with a vowel sound, no? (or at least, I haven't pronounced it like that)

saem: I say 'MIR' like one would 'IBM', sounding out each letter, so 'M' -> 'em'. It's not a big deal, that was an autopilot thing.

zerbina: Ah, okay, that I didn't consider. I'll change the `a` to an `an` then.

compiler/vm/vm.nim
saem: Did I totally miss the `ord` on my last read, or did that change? If I missed it, very sorry for misunderstanding it. 🤦‍♂️

saem:
I like it, please merge whenever you're ready.

saem:
Looks like someone actually tested it.

Some background: test was added due to C++ backend -- msvc warning

I'm going to double check the spec for C on this, but IIRC it was unspecified. If the spec disagrees, then `ord` is broken and this PR itself is fine.

zerbina:
The C standard seems to say that `-0.0 == 0.0` equals true, and I believe for JavaScript it's the same.

For now, I think it'd make sense to keep converting `-0.0` to false, but change the logic in `opNumConv` to check for `floatVal != 0.0 and floatVal != -0.0`.

compiler/vm/vm.nim
zerbina: The was `ord` already there previously -- I re-used the pre-existing float-to-conversion logic from the `opConv` procedure. No worries :)

saem:
> The C standard seems to say that `-0.0 == 0.0` equals true, and I believe for JavaScript it's the same.

By chance do you have a link, I'm looking through [89 draft](http://port70.net/~nsz/c/c89/c89-draft.html) and haven't found anything definitive yet.

> For now, I think it'd make sense to keep converting `-0.0` to false, but change the logic in `opNumConv` to check for `floatVal != 0.0 and floatVal != -0.0`.

Works for me.

zerbina:
> By chance do you have a link, I'm looking through [89 draft](http://port70.net/~nsz/c/c89/c89-draft.html) and haven't found anything definitive yet.

The best I've found is the section from [cpprference](https://en.cppreference.com/w/c/language/arithmetic_types), but it doesn't cite any sources.

So far, I also wasn't able to find any concrete mentions of `float` signed-zero comparison behaviour in C.

saem:
tl;dr: we should treat `-0.0` as `false`.

Revisiting `ord` is a separate matter (my guess is it should also treat `-0.0` as `0`).

---

A search proved fruitful, I think this [Stackoverflow Q/A](https://stackoverflow.com/questions/73598674/is-negative-floating-point-zero-guaranteed-to-evaluate-to-false-in-c) is exactly what we're asking and the answer is well reasoned with citations.

I checked the citations vs the C89 draft, for `if` see [3.6.4.1](http://port70.net/~nsz/c/c89/c89-draft.html#3.6.4.1):

> In both forms, the first substatement is executed if the expression compares unequal to 0. In the else form, the second substatement is executed if the expression compares equal to 0 ...

For the floating point model in the C89 draft is incredibly spare, but the reasonable interpretation is as the answer states while referencing the 2018 spec. Long story short, per ieee the mathematical value of `-0.0` is `0` (for the decimal part). A C compiler/CPU not conforming to this is buggy IMO.

zerbina:
Thank you very much for looking into the matter! The most recent commit is correct then.

bors r+

PR #627, opened 2023-04-13
semcall: isolate overload resolution more

zerbina:
## Summary

Prevent symbols introduced by sem-checking the operands from affecting overload resolution. In addition, only commit these symbols to the symbol table if a match is found, which reduces the amount of side-effects `resolveOverloads` has in case of an error.

## Details

- add an `initOverloadIter` overload that allows specifying the scope to start traversal in
- wrap the call(s) to `pickBestCandidate` in a shadow scope that is only merged if a match was found
- set the `currentScope` when initializing an overload iterator with a symbol choice, which prevents changes to the `PContext`'s `currentScope` from affecting the iterator
- remove the logic for detecting symbol table changes from `pickBestCandidate`

---

## Notes for Reviewers
* a small simplification / fix I noticed as part of the explicit generic call rework
* none of my other tasks depend on this PR, so there's no urgency to merging it

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semcall.nim
+++ b/compiler/sem/semcall.nim
@@ -33,41 +33,17 @@ proc sameMethodDispatcher(a, b: PSym): bool =
       # be disambiguated by the programmer; this way the right generic is
       # instantiated.
 
-proc initCandidateSymbols(c: PContext, headSymbol: PNode,
-                          filter: TSymKinds,
-                          best, alt: var TCandidate,
-                          o: var TOverloadIter): seq[tuple[s: PSym, scope: int]] =
-  result = @[]
-  var symx = initOverloadIter(o, c, headSymbol)
-  while symx != nil:
-    if symx.kind in filter:
-      result.add((symx, o.lastOverloadScope))
-    elif symx.isError:
-      localReport(c.config, symx.ast)
-    symx = nextOverloadIter(o, c, headSymbol)
-  if result.len > 0:
-    initCallCandidate(c, best, result[0].s, result[0].scope)
-    initCallCandidate(c, alt, result[0].s, result[0].scope)
-    best.state = csNoMatch
-
 proc pickBestCandidate(c: PContext,
                        headSymbol: PNode,
                        n: PNode,
+                       startScope: PScope,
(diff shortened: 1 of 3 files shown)


compiler/sem/semcall.nim
```diff
@@ -352,8 +315,14 @@ proc resolveOverloads(c: PContext, n: PNode,
     result.call = c.config.wrapError(n)
     return
 
+  let scope = c.currentScope
+  # all symbols created during lazy semchecking of operands first get
+  # committed into a shadow scope, and only iff there was a match are merged
+  # into the original scope
```
```suggestion
  # all symbols created during lazy semchecking of operands first get
  # committed into a shadow scope, and iff there was a match are merged into
  # the original scope
```

saem (approved):
The PR is nice and compact, just needed a quick glance to look over. 😆 

I left an entirely optional change suggestion wrt a doc comment -- not necessary. Merge away if you'd like.

saem:
Discussed in chat, this is good to merge.

---

@Clyybber did point out that an equilvance that is no longer the case is:

```nim
call(a, b, c)
```
is not the same as
```
let arg1 = a
let arg2 = b
let arg3 = c
call(arg1, arg2, arg3)
```
when `call` is a `proc`, specifically.

In particular, if one of the args is `(proc call(a, b, c) = ...; a)`, then previously the newly defined `call` would participate in overload resolution (modifying the symbols to iterate over). This is a neat trick (assuming that we even trust it to work) that in conjunction with other improvements could allow for late static multiple-dispatch... but this is rather fringe and it needs more than just the above.

For now, this change makes a lot of sense, in that the overload list is set once at the beginning, roughly equivalent to:

```nim
block: # extract the call afterwards
  let
    call = foo  # effectively a symchoie of overloads
    arg1 = a
    arg2 = b
    arg3 = c
  call(arg1, arg2, arg3)
```

---

bors r+

saem:
bors r+

(spurious; I think there is a race in megatest, likely around batching/nimcache)

PR #628, opened 2023-04-14
transf: de-duplicate non-routine symbols

zerbina:
## Summary

Rewrite var, let, and const definitions to use a copied symbol if the symbol already appeared in a definition position once, and redirect following usages of the original symbol to said copies.

Until semantic analysis properly sanitizes the AST after template or macro expansion, this works around a common internal compiler error, and is a preparation for switching to ORC as the default memory management strategy.

## Details

Substituting template parameters with typed AST effectively leads to symbol redefinition (although it is not reported as an error).

This causes issues for the `injectdestructors` pass and the code generators, as both operate under the assumption that a symbol only appears exactly once in a definition context.

`transf` is chosen for fixing-up the symbols because:
- it has access to the owning module's `IdGenerator` (`mirgen` doesn't)
- it already has a facility for rewriting symbol nodes in usage positions

Only globals, constants, and locals not processed by lambda-lifting are de-duplicated however -- other redefined symbols (like those of routines) will still cause issues.

---

## Notes for Reviewers
* since it's only meant as temporary work-around, the added code is not very polished
* this blocks making `--gc:orc` the default, but progress on the task is still possible without this PR

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


compiler/sem/injectdestructors.nim
```diff
@@ -408,13 +408,7 @@ func initEntityDict(tree: MirTree, owner: PSym): EntityDict =
...
       if t != nil and hasDestructor(t):
         let re = toName(entity)
-        # XXX: because of an issue with semantic analysis (see
-        #      ``tests/arc/tstrformat.nim``) it can happen that the same entity
-        #      is defined multiple times. In order to not silently generate
-        #      incorrect code, ``doAssert`` is misused to report an internal
-        #      compiler error here. Using the proper facility would require
-        #      access to a ``ConfigRef`` (which we neither do nor should have)
-        doAssert re notin result, "entity appears in a 'def' multiple times"
+        assert re notin result, "entity appears in a 'def' multiple times"
```
saem: I would consider leaving it as a `doAssert` until we can get back to `release` builds in CI.

compiler/sem/transf.nim
```diff
@@ -174,6 +188,31 @@ proc freshVar(c: PTransf; v: PSym): PNode =
     newVar.owner = owner
     result = newSymNode(newVar)
 
+proc transformDefSym(c: PTransf, n: PNode): PNode =
```
saem: It's probably worth marking this as `{.deprecated: "workaround sem not sanitizing AST".}`

compiler/sem/transf.nim
```diff
@@ -174,6 +188,31 @@ proc freshVar(c: PTransf; v: PSym): PNode =
...
+  ## Transforms a node that appears in a definition position. If the same
+  ## symbol already appeared in a definition position, a copy is created, and
+  ## a mapping registered so that all further usages of the symbol use the
+  ## copy.
+  # XXX: this is only intended to fix the most simple cases of duplicate symbols
```
saem, line 196: Ah yes, this would be dangerous to use for routines, for example: forward declaration(s) AST would define the symbol and then it shows up again later in the definition.

saem (commented):
I just encountered this issue in CLI report refactoring.

Left a few small suggestions.

---

Part of me wonders if it'd be better to instead raise an error in sem within `semNormalizedLetOrVar` and `semNormalizedConst` if it detects a pre-existing definition. Not sure which way is necessarily "better", they both feel short-term. Given all that, they both sound reasonable and about the only thing the sem approach has over this one is that it's more "vocal" when things are wrong.

I suggested the `doAssert` and `deprecated` as it has some of the "hard to ignore" aspect of the sem based approach while not being quite as harsh.

compiler/sem/injectdestructors.nim
zerbina: I changed it back to `doAssert`.

PR #629, opened 2023-04-15
stdlib: fix `range` types for `repr_v2`

zerbina:
## Summary

Fix an infinite template expansion error when passing a range type with a `distinct T` base type to `repr` while using `repr_v2`.

In order to support to fix the `repr` the implementation, the new type-trait routine `rangeBase` is introduced. As the name implies, it used for querying a `range` type's base type.

## Details

For an expression of type `range T` where `T` is a `distinct`, the template `repr(x: distinct)` is chosen by overload resolution.

But since `range T` is not a `distinct` type, `distinctBase` returns the same type as is passed in, which causes the type to be left unchanged and the same `repr` overload to be selected again.

---

## Notes for Reviewers
* blocks #630

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/lib/pure/typetraits.nim
+++ b/lib/pure/typetraits.nim
@@ -184,6 +184,19 @@ since (1, 3, 5):
 
     typeof(block: (for ai in a: ai))
 
+proc rangeBase*(t: typedesc): typedesc {.magic: "TypeTrait".} =
+  ## Returns the base type of the ``range`` type `t`. Only a single level is
+  ## skipped, that is, if the range type's base type is also a range type,
+  ## the base type is not skipped.
+  runnableExamples:
+    type
+      Range = range[1..3]
+      Nested = range[Range(1)..Range(3)]
+
+    doAssert rangeBase(Range) is int
+    doAssert rangeBase(Range) isnot range
+    doAssert rangeBase(Nested) is Range
+
 proc isCyclical*(t: typedesc): bool {.magic: "TypeTrait".} =
   ## Returns whether the type `t` is *potentially* able to be part of a
   ## reference cycle when used as the type of a managed heap location.
--- a/lib/system/repr_v2.nim
+++ b/lib/system/repr_v2.nim
@@ -6,6 +6,9 @@ proc isNamedTuple(T: typedesc): bool {.magic: "TypeTrait".}
 proc distinctBase(T: typedesc, recursive: static bool = true): typedesc {.magic: "TypeTrait".}
   ## imported from typetraits
 
+proc rangeBase(T: typedesc): typedesc {.magic: "TypeTrait".}
+  ## imported from typetraits
+
 proc repr*(x: NimNode): string {.magic: "Repr", noSideEffect.}
 
 proc repr*(x: int): string =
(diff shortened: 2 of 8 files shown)


lib/pure/typetraits.nim
```diff
@@ -184,6 +184,19 @@ since (1, 3, 5):
 
     typeof(block: (for ai in a: ai))
 
+proc rangeBase*(t: typedesc): typedesc {.magic: "TypeTrait".} =
+  ## Returns the base type for ``range`` types. Only a single level is skipped,
+  ## that is, if the base type the range type `t` is also a range type, it is
+  ## not skipped.
```
saem:
```suggestion
  ## Returns the base type for ``range`` types. Only a single level is skipped,
  ## that is, if the base type of range type `t` is also a range type, it is
  ## not skipped.
```

tests/stdlib/algorithm/trepr.nim
```diff
@@ -67,6 +67,20 @@ template main() =
...
+      Nested = distinct range[DistinctRange(1)..DistinctRange(2)]
+
+    # simple case: only a range
+    doAssert repr(Range(1)) == "1"
+    # more complex: a range wrapped in a ``distinct``
+    doAssert repr(DistinctRange(1)) == "1"
+    # complex case: a distinct range where the base type is also a distinct
+    # type
+    doAssert repr(Nested(DistinctRange(1))) == "1"
+
```
saem, line 83: This will be caught via runnable examples in CI, although much later..

lib/system/repr_v2.nim
```diff
@@ -92,8 +95,15 @@ proc repr*(p: proc): string =
   ## repr of a proc as its address
   repr(cast[ptr pointer](unsafeAddr p)[])
 
-template repr*(x: distinct): string =
-  repr(distinctBase(typeof(x))(x))
+template repr*[T: distinct|range](x: T): string =
+  # note: distinct and range match equally well, so it's not possible to use
+  # two templates -- doing so would lead to ambiguity errors
```
saem, line 100: Maybe this should become a generics known issue, `range` is more precise than `distinct` at the very least. Also, I'm pretty sure `distinct` should not mean `distinct|range[distinct]`... that seems like madness.

saem (commented):
Nothing major, found one typo and the other item is simply highlighting possible unintended redundancy in testing.

Interesting tidbit learning that `distinct` and `range` get muddled; that seems unfortunate. It's more  tangential so we don't need to figure that out as part of this PR, I think.

lib/system/repr_v2.nim
zerbina: Yep, I agree -- I'll add a `knownIssue` test.

tests/stdlib/algorithm/trepr.nim
zerbina: Could you point me to the `runnableExamples`? I can't seem to find the one that tests this case.

saem: Sorry, I should have confirmed it, but I got the false impression that the rangebase runnable examples were equivalent. Brain was doing fast and loose pattern matching based on the setup (wasn't conscious of it).

You're right, these are different sorry about that.

saem:
Bors r+

PR #630, opened 2023-04-15
compiler: default to `--gc:orc` for the C target

zerbina:
## Summary

After bootstrapping the compiler itself with `--gc:orc`, this is the
second step towards removing the old runtime.

In order to make to make sure that ORC is stable enough, and that the
standard library works with the new runtime, `--gc:orc` is now implied
for the C target when no memory management strategy is explicitly
selected.

### The New Runtime

Using either `--gc:orc` or `--gc:arc` implies usage of the new runtime,
meaning that the new runtime is now the default. The key differences
compared to the old runtime are:
* usage of `deepcopy` in alive code requires the `--deepcopy` option to
  be enabled
* modifying a string through unsafe facilities (pointers and `cstring`)
  requires preparing the string for mutation first (via
  `prepareMutation`)
* `GC_unref` and `GC_ref` are no longer available for `string` and `seq`
  types
* `repr` version 2 is used

With `--gc:orc` and `--gc:arc`, `shallowCopy` is a no-op.

## Details

* support `prepareMutation` for code running in the VM (it's currently a
  no-op)
* fix a crash with `marshal.loadAny` and empty strings, when using the
  new runtime
* disable tests depending on `nimrtl` -- `nimrtl` doesn't yet work with
  the new runtime
* the IC implementation doesn't support type-bound operators (and thus
  `--gc:arc|orc`) at this time, so change its tests to explicitly use
  `--gc:refc`
* adjust the test suite where necessary


compiler/nim.nim
```diff
@@ -89,6 +90,18 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef): CmdLineHandlingResult =
       conf.errorCounter != 0:
     return
 
+  # if no GC was selected by either the command line or the config files,
```
saem: This needs to be true for suggests as well, I'd put it in `cmdlinehelper` for now.

tests/stdlib/metaprogramming/mgenast.nim
```diff
@@ -31,6 +31,7 @@ macro bindme6UseExpose*(): untyped =
   genAst:
     var tst = "sometext"
     var ss = newStringStream("anothertext")
+    prepareMutation(tst)
```
saem, line 34: Promise me that if we do something with copy on write it has to be named `prepareMootation`. 🐄

tests/stdlib/types/tpackedsets.nim
```diff
@@ -235,15 +235,15 @@ block legacyMainModuleTests:
 
       proc bug12366 =
         var
-          x = initPackedSet[A]()
+          a = initPackedSet[A]()
```
saem, line 238: No need to change it back, just curious why the rename?

saem (commented):
I spotted one thing that needs fixing, selecting the gc if none is set as it impacts what's defined or not in stdlib.

Otherwise this is great, very much looking forward to getting this merged in. Thank you very much for persevering. ♥️🚀

compiler/nim.nim
zerbina: Ah, thanks for catching this. I forgot about that part.

tests/stdlib/types/tpackedsets.nim
zerbina: It's the symbol re-definition issue again (although in a much more subtle way).

`x` is also the name of a `typed` argument of the template this procedure is defined inside. Below, the template is invoked with non-top-level globals passed to `x`.

This leads to the `var substitutedGlobalSymbolGoesHere` to be treated as a procedure-level global. If those have a destructor (`legacyInit`, `instGenericInit`, etc. do have one), they're registered in the global destructor list, which in this case caused the globals to be destroyed twice (a double free that resulted in a crash).

PR #632, opened 2023-04-16
drop deprecated CLI options and arg

saem:
## Summary

These will produce a CLI error instead of warn deprecated: `g`,
`nilseqs`, `nilchecks`, `mainmodule`, `m`, `symbol`, `taintmode`, `cs`,
`deadcodeelim`, and the `v2` `gc` arg.

This is part of a broader refactor of the CLI and replacing legacy reports.

## Details

Besides dropping these from the `commands` module, some updates to the
manual, and other docs were required, and the occasional test
adjustment. Most things were not documented or tested.

---

## Notes for Reviewers
* extracted a small part of the bigger cli refactor


saem:
Bors r+

PR #634, opened 2023-04-16
ast: introduce the `CheckedAst` type

zerbina:
## Summary

It is intended to make error propagation and handling easier and less manual. This commit only adds the type and some basic routines for interacting with it -- no other code is adjusted to make use of it.

So far, the API only covers the very basics, and should be extended on an as-needed basis.

## Details

The plan is to incrementally adjust all semantic analysis procedures to return a `CheckedAst` instance, which should eventually allow for the removal of generic wrapper errors (i.e. `adWrappedError`).

In addition, it could also become useful for storing warnings, or for encapsulating other side-effects produced by sem-checking a piece of AST.

### DIFF
--- a/compiler/ast/checked_ast.nim
+++ b/compiler/ast/checked_ast.nim
@@ -1,0 +1,114 @@
+## The ``CheckedAst`` type is implemented here, along with its accessor
+## routines and routines for creating and manipulating ``CheckedAst``
+## instances.
+##
+## ``CheckedAst`` is a wrapper around ``PNode``, and is intended to make
+## propagation less manual, by keeping track of whether there is an error
+## *somewhere* in the managed AST.
+##
+## The idea is to use ``CheckedAst`` as the return type of all semantic
+## analysis procedures, eventually removing the need for wrapper errors
+## (``adWrappedError``).
+
+import
+  compiler/ast/[
+    ast,
+    ast_types,
+    errorhandling,
+    lineinfos
+  ],
+  compiler/front/[
+    options
+  ]
+
+type
+  CheckedAst* = object
+    ## Represents either in-progress or fully semanticized AST. It is meant
+    ## as a wrapper for semantically *checked* AST, hence the name.
+    n: PNode       ## stores the AST. Never nil
+    hasError: bool ## tracks whether `n` contains an ``nkError`` node
+                   ## somewhere
+
+proc initForMappingChildren*(x: var CheckedAst, n: PNode) =
+  ## Initialize `x` as a node of kind ``n.kind``. Space for all children of
+  ## `n` is allocated, but the child nodes are not copied. For the flags, only
+  ## the persistent ones are copied.
+  assert x.n == nil, "already initialized"
+  assert n.kind in nkWithSons
+  assert n.len > 0
+  x.n = shallowCopy(n)
+
+func initWith*(x: var CheckedAst, n: sink PNode) =
+  ## Initializes `x`, using `n` as the tree. This is only meant to be used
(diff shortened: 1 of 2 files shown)


zerbina:
### Notes for reviewers

- I believe the type's name is okay, but that doesn't mean it's final -- suggestions for better names are welcome
- the earlier discussion surrounding the type's introduction can be found [here](https://matrix.to/#/!MAXxvfCYnfYgZqSizw:matrix.org/$Lp8xCcz9MwHrGWJMY5cGLdkW2MasvcOzW0muucVw9zs?via=libera.chat&via=matrix.org&via=envs.net)

compiler/ast/checked_ast.nim
```diff
@@ -0,0 +1,112 @@
...
+
+func `[]=`*(x: var CheckedAst, i: int|BackwardsIndex, n: sink PNode) =
+  x.hasError = x.hasError or n.kind == nkError
+  x.n[idx(x.n.len, i)] = n
+
+func `[]=`*(x: var CheckedAst, i: int|BackwardsIndex, n: sink CheckedAst) =
+  x.hasError = x.hasError or n.hasError
+  x.n[idx(x.n.len, i)] = n.n
+
+proc wrap*(c: ConfigRef, n: sink CheckedAst): PNode =
```
saem:
```suggestion
proc extract*(c: ConfigRef, n: sink CheckedAst): PNode =
```

Suggesting because `CheckedAst` is itself a wrapper, so the current name is a bit confusing. I was thinking of an antonym for `prepare` that might work, but that's about the best I could come up with.

compiler/ast/checked_ast.nim
```diff
@@ -0,0 +1,112 @@
...
+  ]
+
+type
+  CheckedAst* = object
+    ## Represents either in-progress or fully semanticized AST.
+    n: PNode       ## stores the AST. Never nil
+    hasError: bool ## tracks whether `n` contains an ``nkError`` node
+                   ## somewhere
+
+proc prepare*(x: var CheckedAst, n: PNode) =
```
saem:
```suggestion
proc initForMappingChildren*(x: var CheckedAst, n: PNode) =
```

The name is _huge_, but I'm hoping it can help inspire some ideas and we can land on something better.

Additionally, I think this should ensure that `n` has children, otherwise, I don't think it really makes sense that we `shallowCopy`; feels like the purpose is off.

compiler/ast/checked_ast.nim
```diff
@@ -0,0 +1,112 @@
...
+    ast_types,
+    errorhandling,
+    lineinfos
+  ],
+  compiler/front/[
+    options
+  ]
+
+type
+  CheckedAst* = object
```
saem, line 25:
```suggestion
  TrackedAst* = object
```

We're more tracking than checking.  I think checking would be something like assertions on replacing an error node with something else.

saem (commented):
Some light suggesitons, these can also be implemented after the fact, just a bit of compiler bashing for renames.

compiler/ast/checked_ast.nim
zerbina: The idea was to highlight that the type is meant for sem-checked only (although that is currently not always true). I don't have a strong preference, however, and am okay with changing it.

saem: Adding the bit about `sem-checked` as the etymology of the name in a doc comment is also workable, sorta recontextualized it for me. Your call, this is one of those things where can easily change our minds later.

zerbina: Okay, given that the module already uses the name, I'd leave it as-is for the time being (but with an added explained for the name in the doc comment).

PR #635, opened 2023-04-16
sem: don't create improper type instantiations

zerbina:
## Summary

Fix conversions to composite type-classes (e.g. `GenericType(x)`) leading to the creation of improper type instantiations. These would lead to invalid C code being generated and thus, in some cases, C compiler errors.

## Details

An improper type instantiations is a `tyGenericInst` type that wasn't created as the result of resolving/evaluating a `tyGenericInvocation`. Their carried concrete object type (if they carry one, that is) doesn't have the `typeInst` field set, and the concrete type is also possibly missing a `=deepcopy` type-bound operator.

They were the result of a non-concrete `tyGenericInst` (that is, one containing meta types) being subjected to type-var replacing, where all referenced meta types had bindings to concrete type.

Since signature hashing takes an object type's `typeInst` field into account, this lead to the C code generator treating improper instantiations as standalone types, emitting and using a separate C type for them, and thus generating illegal C code.

The logic in `inferWithMetatype` is adjusted to not pass a `tyGenericInst` that contains meta types to `generateTypeInstance`, but to instead create a generic invocation and evaluate that.

### Type-var replacing

As an experiment, `tyGenericInst` types are no longer modified nor copied by `replaceTypeVarsT`, and the workarounds introduced by the the commits f915b3aa86929d87e162b6ddd7589e6337a30397 and 3f3e0fa303a2729d0d6a9ffe54936c1c830e2d7e removed.

For concrete `tyGenericInst` types, this is always correct, as they don't contain any type variables. Copying and processing them is thus unnecessary.

`tyGenericInst` instances that contain meta types *should* only reach through when generic invocations are passed to `prepareMetatypeForSigmatch`, in which case copying and modifying them is wrong.

### DIFF
--- a/compiler/sem/semcall.nim
+++ b/compiler/sem/semcall.nim
@@ -424,6 +424,8 @@ proc indexTypesMatch(c: PContext, f, a: PType, arg: PNode): PNode =
 
 proc inferWithMetatype(c: PContext, formal: PType,
                        arg: PNode, coerceDistincts = false): PNode =
+  addInNimDebugUtils(c.config, "inferWithMetatype", arg, result)
+
   var m = newCandidate(c, formal)
   m.coerceDistincts = coerceDistincts
   
@@ -440,8 +442,30 @@ proc inferWithMetatype(c: PContext, formal: PType,
     # This almost exactly replicates the steps taken by the compiler during
     # param matching. It performs an embarrassing amount of back-and-forth
     # type jugling, but it's the price to pay for consistency and correctness
-    result.typ = generateTypeInstance(c, m.bindings, arg.info,
-                                      formal.skipTypes({tyCompositeTypeClass}))
+    # XXX: overwriting the type of `result` like it's done here is not correct.
+    #      At least for the ``coerceDistincts == true`` case (currently only
+    #      used by ``semConv``), it makes more sense to return both the fitted
+    #      node *and* the inferred formal type, and let the callsite handle it
+    #      from there
+    if formal.kind == tyCompositeTypeClass:
+      # passing the composite type-class to ``generateTypeInstance`` would
+      # get us the matched source type, which is not what we want here
+      # (especially in the presense of ``distinct``s). We want the instantiated
+      # base type.
+      doAssert formal[0].kind == tyGenericBody
+      let inst = formal[1] ## the fully instantiated meta type
+      assert inst.kind == tyGenericInst
+
+      var invocation = newTypeS(tyGenericInvocation, c)
+      invocation.sons = @[formal[0]] # don't propagte the flags
+      # add the instance arguments as the invocation parameters:
+      for i in 1..<inst.len-1:
+        invocation.rawAddSon(inst[i])
+
+      # evaluate the invocation:
+      result.typ = generateTypeInstance(c, m.bindings, arg.info, invocation)
+    else:
+      result.typ = generateTypeInstance(c, m.bindings, arg.info, formal)
   else:
     result = typeMismatch(c.config, arg.info, formal, arg.typ, arg)
     if result.kind != nkError:
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -319,6 +319,8 @@ proc semConv(c: PContext, n: PNode): PNode =
   var op = semExprWithType(c, n[1])
   if targetType.kind != tyGenericParam and targetType.isMetaType:
     let final = inferWithMetatype(c, targetType, op, true)
+    # XXX: this makes little sense -- the source and target type of the
+    #      resulting conversion are the same
     result.add final
     result.typ = final.typ
     return handleError(c, result, hasError)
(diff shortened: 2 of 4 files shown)


zerbina:
I believe that this is the same issue that caused the `t01_generic_converter.nim` test to fail with the C++ target. It now surfaced again when building the test with `--gc:orc` (which makes this PR a blocker for #630).

The experiment with the `replaceTypeVarsT` adjustment is meant to untangle the type instantiation logic a bit.

compiler/sem/semcall.nim
```diff
@@ -440,8 +442,34 @@ proc inferWithMetatype(c: PContext, formal: PType,
...
+    if formal.kind == tyCompositeTypeClass:
+      # passing the composite type-class to ``generateTypeInstance`` would
+      # get us the matched source type, which is not what we want here
+      # (especially in the presense of ``distinct``s). We want the instantiated
+      # base type.
+      if formal[0].kind == tyGenericBody:
+        let inst = formal[1] ## the fully instantiated meta type
+        assert inst.kind == tyGenericInst
+
+        var invocation = newTypeS(tyGenericInvocation, c)
```
saem: I think this is more correct as we are in fact doing type application here (`tyGenericInvocation`).

compiler/sem/semcall.nim
```diff
@@ -440,8 +442,34 @@ proc inferWithMetatype(c: PContext, formal: PType,
...
+        invocation.sons = @[formal[0]] # don't propagte the flags
+        # add the instance arguments as the invocation parameters:
+        for i in 1..<inst.len-1:
+          invocation.rawAddSon(inst[i])
+
+        # evaluate the invocation:
+        result.typ = generateTypeInstance(c, m.bindings, arg.info, invocation)
+      else:
+        # XXX: is this case ever reached? A ``tyCompositeTypeClass`` where the
+        #      body is not a ``tyGenericBody`` doesn't look possible...
```
saem: I don't see how this is possible, either. I'd perhaps put in a doAssert, as either it's a data input issue (bug elsewhere in the compiler) or misunderstanding here.

compiler/sem/semexprs.nim
```diff
@@ -319,6 +319,8 @@ proc semConv(c: PContext, n: PNode): PNode =
   var op = semExprWithType(c, n[1])
   if targetType.kind != tyGenericParam and targetType.isMetaType:
     let final = inferWithMetatype(c, targetType, op, true)
+    # XXX: this makes little sense -- the source and target type of the
+    #      resulting conversion are the same
```
saem, line 323: My guess is that this _might_ be when going from a given type to `untyped` or `typed`. See this test case: https://github.com/nim-works/nimskull/blob/devel/tests/lang_stmts/bind/tinvalidbindtypedesc.nim

I haven't traced it in detail, but `echo`'s params are `varargs[typed, $]`, that would explain why they had to make it act like an identity here to get the code to "work".

saem (approved):
Looks good!

I didn't spot anything that wasn't pre-existing, and the new code is chipping away at the redundancy/convoluted nature of the existing.

compiler/sem/semcall.nim
zerbina: This case is now guarded against with an assertion.

compiler/sem/semexprs.nim
zerbina: It's something arising from `inferWithMetatype` always overwriting the result node's type. Consider the follow (which takes the `inferWithMetatype` path):
```nim
type ArrayTy[T] = distinct array[1, T]

discard ArrayTy(["str"])
```

`inferWithMetatype` sets the type of the expression `["str"]` to `ArrayTy[string]`, but here in `semConv`, it is then wrapped with `nkConv` node also using `ArrayTy[string]` as the type.

The more correct thing to do would be leaving the type of the processed expression in `inferWithMetatype` as is (at least if `coerceDistincts == true`), and only use the inferred type for the `nkConv` expression.

saem: I strongly agree, and would perhaps go a bit further; I think this PR is all good so going to bors.

---

### Aside: Going a Bit Further

The key points for me, and I believe entirely line up with yours are:
1. the type of expression `["str"]` should always be the same, something like: `array[1, static[string]]`
2. applying `ArrayTy` is a few steps, but the key conversion expression result's type is `ArrayTy[static[string]]`

Coercion of distincts should determine whether the conversion expression result's type is an error (because distinct coercion is disallowed) or successful and therefore `ArrayTy[static[string]]`.

Definitely agree that `inferWithMetatype` is applying the wrong type on `["str"]`, I think this is a generally unhygienic habit of not creating intermediate nodes and apply appropriate types and conversions all over the compiler.

saem:
Seeing if I can kick it into working with a workflow rerun.

compiler/sem/semexprs.nim
zerbina: Yep, they entirely line up.

PR #637, opened 2023-04-18
remove nimscript as config

saem:
## Summary

Scripting in configuration led to lots of complexity, and many usability
and security issues. It now is works only for scripting, and config is
handled by `.cfg` files.

This is a fairly far reaching change:

- removed the ability to use `nimscript` for config
- dropped all associated APIs in `nimscript` module
- removed all nimscript config from this codebase
- updated `nimscript` docs to focus on scripting
- added/enabled many tests, and a few related bug fixes

bug fixes along the way:
- index out of range for `opcWrArr` was off (over) by one
- legacy reports silently dropped `@write` directive output

## Details

- renamed `scriptconfig` module to `scripting`
- fixed legacy code that disabled nimscript execution
- removed old code related to resetting things for config usecases

`nimscript` API:
- removed all config related API, including nimble things
- added `writeFile(filename, content: string): void`
- `-d:scriptmode` now allows CLI override of a script's mode, see docs
- `exec(cmd, intput: string)`, now returns a string

Bug Fix: Config `@write` Directive:
Fixed a bug where legacy reports simply dropped output from the `@write`
directive in config files.

Bug Fix: VM `opcWrArr` correctly reports max array index
- `opcWrArr` used the array len as the max for the `rvmIndexError` msg
- fixed this in order to get the test passing again
- previously the length of the array was reported as max (over by one)

#### Tests:

Testament no longer does special handling for `nims` config. Testament
currently can't run `nims` directly, this is something to improve in the
future.

In order to get the nimscript tests to work without testament support,
used `tests/misc/trunner` and wrote the occasional runner `.nim` file to
get them going for now.

Also reenabled some nimscript api testing in CI, for now added it to
koch's `testTools`, need a better place in the future.

Added and reenabled a number of tests for scripting and the `nimscript`
API. Exposed some latent `ic` bugs that relied on nimscript for config
processing `system` modules. These tests were disabled as `ic` is on the
way out, this should reduce CI cycles wasted on this stuff.

#### Minor:
- dropped unused imports from `scriptconfig`
- lots of little formatting fixes along the way
- noted strtabs `useEnvironment` will drop, dependency on `os` is daft

#### Lessons Learned/Future Work:

Dropping nimscript from config meant facing the silly standard library
version mess. `NimMajor`, and the minor/patch variants were set
awkwardly in the standard library, then used to construct the
`NimVersion` string, which was then reinjected into the build. There are
so many things wrong with this, for starters:

- conflates standard library version and compiler version
- introduces a recursive dependency
- doesn't act like a version, you can only have one
- sets a bad example for others

Introduced some initial separation of compiler and standard library
versions. This work will be carried forward separately.

Separately `trunner` is used to run a bunch of "weird" tests, this
screams for significant revision of testament's spec and category
handling so we can run these different variants without this janky code.

Thanks to Zerbina for review and suggestions.


saem:
## Notes for Reviewers
* need to do some clean-up still
* IC tests broke, I'm simply going to disable the ones that did
  * looks like latent bug around module graph pollution
  * replicating is possible with the current compiler, simply remove `tests/config.nims`

Task list:

- [x] clean-up existing changes (TODOs/comments)
- [x] drop the `doAssert` if a nimscript config is found
- [x] add tests for items that never had them
- [x] remove `dump` command hack
- [x] fix testament bug related to config file/joinable test detection
- [x] throw out IC tests that are flailing
- [x] typos/suggestions
- [x] fix tests

saem:
Just leaving some notes here for myself and anyone else interested, my current hypothesis is that I've exposed a bug in IC. Presently, my thinking is that IC is just another source of bugs and nearing retirement. I'm likely going to just disable/remove those tests and call it a day.

---
### hypothesis/investigation

**updated:** The biggest difference I can think of between the `config.nims` vs `nim.cfg` version is that the former ends up building up a module graph and associated rod files for all of system. This doesn't happen the same way when we get to the `tgenerics` test. It crashes in `cgen` after a few test edits, as it can't `cgen.findPendingModule` the `raiseOverflow` `compilerproc` symbol's module (`system` module, id: `3`) in the `BModule.g.modules`. This all occurs in the `hashcommon` module, triggered off of line 32 (guarding an overflow).

compiler/ast/reports_internal.nim
```diff
@@ -44,7 +44,7 @@ type
     prefixdir*: string
     libpath*: string
     projectPath*: string
-    definedSymbols*: seq[string]
+    definedSymbols*: seq[tuple[key: string, val: string]]
```
saem: This was some quick and dirty debugging, although a cleaned up version of this would be good for introspection (separate PR).

compiler/front/cli_reporter.nim
```diff
@@ -2745,14 +2745,18 @@ To create a stacktrace, rerun compilation with './koch temp $1 <file>'
       assert false, "is a configuration hint, should not be reported manually"
 
     of rintNimconfWrite:
-      result = ""
+      result = r.msg
```
saem, line 2745: Bug fix.

compiler/front/cli_reporter.nim
```diff
@@ -2745,14 +2745,18 @@ To create a stacktrace, rerun compilation with './koch temp $1 <file>'
...
         let s = r.stateDump
-        var definedSymbols = newJArray()
-        for s in s.definedSymbols:
+        var
+          definedSymbols = newJArray()
+          symsWithValues = newJObject()
+        for s, val in s.definedSymbols.items:
           definedSymbols.elems.add(%s)
+          if val notin ["true", ""]:
+            symsWithValues.add(s, %val)
```
saem: More quick and dirty debugging

compiler/front/main.nim
```diff
@@ -409,8 +409,8 @@ proc mainCommand*(graph: ModuleGraph) =
   of cmdDump:
     wantMainModule(conf)
     var state = InternalStateDump()
-    for s in definedSymbolNames(conf):
-      state.definedSymbols.add $s
+    for s, val in conf.active.symbols.pairs:
+      state.definedSymbols.add (key: $s, val: val)
```
saem: and the last bit of said debugging

compiler/front/nimconf.nim
```diff
@@ -438,9 +435,7 @@ proc loadConfigs(
 
   proc runNimScriptIfExists(N: var NimConfParser, path: AbsoluteFile) =
     if fileExists(path):
-      var s = llStreamOpen(path, fmRead)
-      N.config.configFiles.add(path)
-      runNimScript(cache, path, freshDefines = false, N.config, s)
+      doAssert false, "nimscript for config no longer supported: " & $path
```
saem: way to detect `config.nims` and what not, going to simply drop this whole proc and all calls to it.

compiler/front/scriptconfig.nim
```diff
@@ -7,8 +7,7 @@
 #    distribution, for details about the copyright.
 #
 
-## Implements the new configuration system for Nim. Uses Nim as a scripting
-## language.
+## Use Nimskull as a scripting language.
```
saem, line 10: It's only for scripting, done.

doc/nims.rst
```diff
@@ -9,31 +9,60 @@ Strictly speaking, `NimScript` is the subset of Nim that can be evaluated
...
+
+
+Dry-Run and Debugging
+=====================
+
+Defining `scriptmode` on the command-line like so:
+`nim -d:scriptmode foo.nims`:cmd: ensures that it's run in `WhatIf` mode, where
+effects such as shell execution, file IO, and so on are logged to the console
+and never ran. There are additional options, see `nimscript <nimscript.html>`_
+for details.
```
saem, line 55: This is new functionality, felt like the old way was just such a big security issue and missed opportunity.

lib/system.nim
```diff
@@ -2050,20 +2050,56 @@ template unlikely*(val: bool): bool =
       unlikelyProc(val)
 
 const
+  ## TODO: in the next csources release:
```
saem, line 2053: Sketched out what I think we should do with the compiler versions etc being exposed via the standard library/programmatically. Also need to dial the numbers back when we cut the next csources.

tl;dr: compiler version != standard library versions, also I'm completely uninterested in keep nim version compatibility, people can just edit their code.

lib/system.nim
```diff
@@ -2050,20 +2050,56 @@ template unlikely*(val: bool): bool =
...
+    ## .. code-block:: Nim
+    ##   when (StdlibMajor, StdlibMinor, StdlibPatch) >= (1, 3, 1): discard
+  StdlibMinor*: int = 6
+    ## standard library minor version
+    ## TODO: change to `1` in next csources
+  StdlibPatch*: int = 0
+    ## standard library patch version
+    ## TODO: change to `0` in next csources
+
+  # hardcoded in config/config.nims
```
saem: I dropped the code that hardcodes them, but might bring that back in a subsequent commit.

lib/system/nimscript.nim
```diff
@@ -156,35 +103,40 @@ proc strip(s: string): string =
...
+      echo "[NimScript] ", msg
+    if actualMode != ScriptMode.Whatif:
+      body
+else:
+  template log(msg: string, body: untyped) =
+    if mode in {ScriptMode.Verbose, ScriptMode.Whatif}:
+      echo "[NimScript] ", msg
+    if mode != ScriptMode.Whatif:
+      body
+
```
saem, line 139: new functionality

lib/system/nimscript.nim
```diff
@@ -256,39 +202,31 @@ proc exec*(command: string) {.
...
+proc exec*(command: string, input: string, cache = ""): string {.
   raises: [OSError], tags: [ExecIOEffect, WriteIOEffect].} =
   ## Executes an external process. If the external process terminates with
   ## a non-zero exit code, an OSError exception is raised.
   log "exec: " & command:
     let (output, exitCode) = gorgeEx(command, input, cache)
     if exitCode != 0:
       raise newException(OSError, "FAILED: " & command)
-    echo output
+    result = output
```
saem, line 218: minor change in API, seems silly not to do it this way.

tests/misc/trunner.nim
```diff
@@ -191,30 +191,31 @@ sub/mmain.idx""", context
...
-tests/newconfig/bar/config.nims
-tests/newconfig/bar/mfoo.nim.cfg
-tests/newconfig/bar/mfoo.nims""".splitLines
-    var expected = ""
-    for a in files:
-      let b = dir / a
-      expected.add &"Hint: used config file '{b}' [Conf]\n"
+    let
+      dir = getCurrentDir()
+      files = ["tests/misc/config/bar/nim.cfg", "tests/misc/config/bar/mfoo.nim.cfg"]
```
saem, line 201: Ah, I think I missed a file.

saem:
@zerbina here are the changes dropping nimscript config and switching to all config files, if you wanted to take a quick peek. Gonna run through clean-ups tomorrow.

compiler/front/cli_reporter.nim
```diff
@@ -2784,6 +2788,7 @@ To create a stacktrace, rerun compilation with './koch temp $1 <file>'
           (key: "nimcache",        val: %s.nimcache),
           (key: "hints",           val: hints),
           (key: "warnings",        val: warnings),
+          (key: "symsWIthValues",  val: symsWithValues),
```
```suggestion
          (key: "symsWithValues",  val: symsWithValues),
```

compiler/front/scriptconfig.nim
```diff
@@ -191,53 +170,22 @@ proc setupVM*(module: PSym; cache: IdentCache; scriptName: string;
...
 proc runNimScript*(cache: IdentCache; scriptName: AbsoluteFile;
                    freshDefines=true; conf: ConfigRef, stream: PLLStream) =
-
   conf.localReport DebugReport(
     kind: rdbgStartingConfRead,
-    filename: scriptName.string
-  )
+    filename: scriptName.string)
 
   let oldSymbolFiles = conf.symbolFiles
```
zerbina: It's nothing that has to happen as part of this PR, but with NimScript no longer used for configuration, some of the save/restore/cleanup logic in this procedure (e.g. `oldSymbolFiles`, `oldSelectedGC`, `resetSystemArtifacts`, etc.) can be removed too.

doc/nims.rst
```diff
@@ -267,8 +194,9 @@ that means there is nothing new to learn, no context switch for developers.
 Powerful Metaprogramming
 ------------------------
 
-NimScript can use Nim's templates, macros, types, concepts, effect tracking system, and more,
-you can create modules that work on compiled Nim and also on interpreted NimScript.
+NimScript can use Nim's templates, macros, types, concepts, effect tracking
+system, and more, you can create modules that work in comipled or scripting
+context.
```
zerbina:
```suggestion
system, and more, you can create modules that work in both a compiled or
scripting context.
```
I think it reads slightly better.

lib/system/nimscript.nim
```diff
@@ -353,18 +279,12 @@ template withDir*(dir: string; body: untyped): untyped =
   finally:
     cd(curDir)
 
-
 proc writeTask(name, desc: string) =
   if desc.len > 0:
     var spaces = " "
     for i in 0 ..< 20 - name.len: spaces.add ' '
     echo name, spaces, desc
```
zerbina:
```suggestion
```
It's dead code now.

tests/misc/script/tscript.nims
```diff
@@ -106,3 +65,14 @@ when false:
 
 rmDir("tempXYZ")
 doAssert dirExists("tempXYZ") == false
+
+import std/strtabs
+block ensure_strtab_interpolation_works_in_scripting_env:
+  # the strtab module relied on `os.getEnv` explicitly, now it's not longer
```
zerbina: `std/strtabs` depending on the `os` module feels ridiculous. Given that only the compiler seems to use the "default to environment variable" feature, maybe the `flags` should be replaced with a callback.

lib/system/nimscript.nim
```diff
@@ -256,39 +202,31 @@ proc exec*(command: string) {.
   raises: [OSError], tags: [ExecIOEffect, WriteIOEffect].} =
   ## Executes an external process. If the external process terminates with
   ## a non-zero exit code, an OSError exception is raised.
-  ##
-  ## **Note:** If you need a version of `exec` that returns the exit code
-  ## and text output of the command, you can use `system.gorgeEx
```
zerbina, line 261: It's also nothing that has to be changed as part of the PR, but in the future, `gorgeEx` and `writeFile` also need to be restricted, both a compile- and execution-time.

zerbina (commented):
I did a pass over the current state, and I like it. I left some small suggestions, typo fixes, and remarks about the future / follow-ups, but nothing that is blocking.

As an aside: one thing I personally used NimScript configuration files for was implementing compiler version requirements without relying on Nimble, but that's something that is indeed better implemented via a different approach to the build system.

compiler/vm/vm.nim
```diff
@@ -1133,7 +1133,7 @@ proc rawExecute(c: var TCtx, pc: var int): YieldReason =
         checkHandle(regs[rc])
         writeLoc(slice[idx], regs[rc], c.memory)
       else:
-        raiseVmError(reportVmIdx(idx, slice.len))
+        raiseVmError(reportVmIdx(idx, slice.len - 1))
```
saem, line 1136: bug fix for wrong max index for `opcWrArr`

lib/system/nimscript.nim
saem: deleted locally.

compiler/front/scriptconfig.nim
saem: Done.

tests/misc/trunner.nim
saem: nope, this is correct.

saem:
Time to start testing and debug any straggler issues tomorrow.

tests/misc/script/tscript.nims
saem: Put in a comment indicating that this is on the way out.

compiler/front/nimconf.nim
```diff
@@ -481,11 +461,6 @@ proc loadConfigs(
     scriptFile = N.config.projectFull.changeFileExt("nims")
     scriptIsProj = scriptFile == N.config.projectFull
```
zerbina, line 462: These are obsolete now.

compiler/front/scripting.nim
```diff
@@ -74,86 +62,92 @@ proc setupVM*(module: PSym; cache: IdentCache; scriptName: string;
...
+    cbeff(name, exc, "system", body)
+
+  template cbio(name, body) {.dirty.} =
+    cbeff(name, IOError, "io", body)
+
   template cbos(name, body) {.dirty.} =
     cbexc(name, OSError, body)
 
+  template guardEffect(body) {.dirty.} =
+    # might not be needed once this is no longer allowed for configs
```
zerbina, line 102: Yep, I think the template is obsolete now. Removing it can happen as a follow-up, however.

zerbina (approved):
Thank you for working on this! I think it's good to go -- the suggestion are only related to now-dead code, and can all be applied in a follow-up PR.

saem:
Thanks, @zerbina -- d'oh, I missed those, I'll get them in a follow-up.

bors r+

PR #638, opened 2023-04-18
sem: disallow `lent|sink T` in non-type contexts

zerbina:
## Summary

A `lent|sink T` expression is represented as a `Prefix (Ident ...) (...)` in untyped AST, and was analyzed in `semConv`. Besides this being unintuitive, inefficient, and also inconsistent with other modifiers like `var`, it also allowed `lent|sink T` type expressions in non-type contexts.

Instead, `lent T` and `sink T` type expression are now analyzed as part of `semTypeNode`, which makes their handling more consistent with that of `var T`.

## Details

The `lent2` helper template in the `iterators_1.nim` module is adjusted to use `typedesc` templates. This is now required because `lent T` can no longer be used in a non-type context (which currently includes `untyped` templates being expanded in type positions).

### DIFF
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -2010,6 +2010,12 @@ proc semTypeNode(c: PContext, n: PNode, prev: PType): PType =
         # skip 'owned' in type expressions and produce a warning
         localReport(c.config, n, reportSem rsemOwnedTypeDeprecated)
         result = semTypeExpr(c, n[1], prev)
+      elif (op.s == "sink" or op.s == "lent") and n.len == 2:
+        # it's a 'sink T' or 'lent T' expression
+        # XXX: consider introducing special words for both (i.e., ``wLent``,
+        #      ``wSink``)
+        result = newOrPrevType((if op.s == "sink": tySink else: tyLent), nil, c)
+        result.rawAddSonNoPropagationOfTypeFlags semTypeNode(c, n[1], nil)
       else:
         if c.inGenericContext > 0 and n.kind == nkCall:
           result = makeTypeFromExpr(c, n.copyTree)
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -281,20 +281,7 @@ proc semConv(c: PContext, n: PNode): PNode =
 
   var hasError = false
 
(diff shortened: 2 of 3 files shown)


zerbina:
I noticed this as part of the ORC-related preparations, and decided to fix it.

---

While using a `typedesc` template is correct, an `untyped` template should have worked too. Making this work is a bit more involved, however, as it requires delaying expansion (or at least analysis of the result) of until control-flow is back in `semTypeExpr`.

`efNoSemcheck` cannot be used for this, as it applies to all invoked templates and macros part of a complex expression, not only the ones used in the expression slot of, for example, an `nkStmtListExpr`. In addition, the flag doesn't discriminate between the template types -- `typed` should likely still be eagerly analyzed.

compiler/sem/semtypes.nim
```diff
@@ -2010,6 +2010,12 @@ proc semTypeNode(c: PContext, n: PNode, prev: PType): PType =
         # skip 'owned' in type expressions and produce a warning
         localReport(c.config, n, reportSem rsemOwnedTypeDeprecated)
         result = semTypeExpr(c, n[1], prev)
+      elif (op.s == "sink" or op.s == "lent") and n.len == 2:
```
saem, line 2013: This means we can't have overloaded single airity `sink` or `lent` routines?

(just an observation, I don't think it needs to change or if that's something we should even allow).

saem (approved):
bors r+

Just an observation, but it's good to go! I very much like the `typedesc` output type for the template, it's definitely the right direction.

----

Aside: wtih respect to allowing for `untyped`, I get quesy feelings about that, we've got an `untyped -> untyped`, or `typeCtx -> untyped -> untyped` depending upon how we want to see it, and there is nothing stopping the template from being an identity. We do have a lmiter but that still doesn't sit right. I think there might be a missing capability, like one shot/singleton/staged templates, meaning once applied to a position it can never apply there again (affine template?).

compiler/sem/semtypes.nim
zerbina: They are generally possible, but are not considered in type contexts.

I think that this is an improvement over how it worked previously, where the presence of a `sink` routine would make all following usages of the `sink` type modifier an error (except when the `sink` routine is a single arity template with a `typedesc`/`untyped`/`typed` parameter).

PR #639, opened 2023-04-19
don't omit the goto-handler for `system` routines

zerbina:
## Summary

Routines defined in the `system` module were effectively treated as being to unable to raise defects. This is incorrect (most notably for `sysFatal`) and would lead to incorrect control-flow in some situations.

Only code built with using `goto` exceptions (implied by `--gc:arc|orc`) and panics disabled (the default) was affected.

## Details

- remove the exception regarding defects for routines defined in `system`
- adjust `nimTestErrorFlag` to clear out the exceptions flag
- test the `tfloat1.nim` with `--exceptions:goto`, as it was affected by the issue

Since a goto-handler is now also emitted when calling non-magic routines from the `system` module, this has small negative impact on execution speed of the generated C code.

### DIFF
--- a/compiler/backend/ccgcalls.nim
+++ b/compiler/backend/ccgcalls.nim
@@ -11,11 +11,13 @@
 ## included from cgen.nim
 
 proc canRaiseDisp(p: BProc; n: PNode): bool =
-  # we assume things like sysFatal cannot raise themselves
+  ## 'true' if calling the callee expression `n` can exit via exceptional
+  ## control-flow, otherwise 'false'. If panics are disabled, this also
+  ## includes all routines that are not certain magics, compiler procs, or
+  ## imported.
   if n.kind == nkSym and {sfNeverRaises, sfImportc, sfCompilerProc} * n.sym.flags != {}:
     result = false
-  elif optPanics in p.config.globalOptions or
-      (n.kind == nkSym and sfSystemModule in getModule(n.sym).flags):
+  elif optPanics in p.config.globalOptions:
     # we know we can be strict:
     result = canRaise(n)
   else:
--- a/lib/system/excpt.nim
+++ b/lib/system/excpt.nim
@@ -420,7 +420,12 @@ when gotoBasedExceptions:
     ## This proc must be called before `currException` is destroyed.
     ## It also must be called at the end of every thread to ensure no
     ## error is swallowed.
-    if nimInErrorMode and currException != nil:
+    # clear the error flag before doing anything else. Otherwise the first
+    # non-magic call would cause this procedure to exit
+    var wasInErrorMode = false
+    swap(nimInErrorMode, wasInErrorMode)
+
+    if wasInErrorMode and currException != nil:
       reportUnhandledError(currException)
       currException = nil
       quit(1)
--- a/tests/lang_types/float/tfloat1.nim
+++ b/tests/lang_types/float/tfloat1.nim
@@ -1,4 +1,5 @@
 discard """
+  matrix: "--exceptions:setjmp;--exceptions:goto"
   outputsub: "Error: unhandled exception: FPU operation caused an overflow [FloatOverflowDefect]"
   exitcode: "1"
 """


zerbina:
### Notes for reviewers
- this is a blocker for #630, and a preparation for making `--exceptions:goto` the default
- using `--panics:on` for the compiler might make sense in the future

compiler/backend/ccgcalls.nim
```diff
@@ -11,11 +11,12 @@
 ## included from cgen.nim
 
 proc canRaiseDisp(p: BProc; n: PNode): bool =
-  # we assume things like sysFatal cannot raise themselves
+  ## Returns whether the callee expression `n` evaluates to a procedural values
+  ## for which a call can exit via exceptional control-flow. When panics are
+  ## not enabled, raising a defect also to exceptional control-flow.
```
saem:
```suggestion
  ## True if calling the callee expression `n` can exit via exceptional
  ## control-flow, otherwise false. Defects are treated as exceptions (true)
  ## iff panics are disabled, otherwise false.
```

I think this preserves the point and might be a bit easier to parse.

saem (approved):
Entirely optional suggestion for the doc comment.

Looks good.

saem:
FYI: I changed the "where" to "were" in the first line of the PR description; might want to update the commit message if you're updating the PR.

compiler/backend/ccgcalls.nim
zerbina: Yep, it is. Thanks.

zerbina:
> FYI: I changed the "where" to "were" in the first line of the PR description; might want to update the commit message if you're updating the PR.

Will do, thanks.

alaviss:
`nimsuggest` is hanging up due to these locks not being released: 

https://github.com/nim-works/nimskull/blob/cddd2bc7d4a890928c869f88b1212013196e02d6/lib/system/alloc.nim#L1096-L1099

Now obviously we should change this to try-finally, but here's the issue: none of the allocator code were raising, so the release shouldn't be skipped in the first place!

Tracing it back shows that this happened within this exception handler (first line):

https://github.com/nim-works/nimskull/blob/cddd2bc7d4a890928c869f88b1212013196e02d6/compiler/sem/sem.nim#L947-L955

My thought was that maybe we didn't clear the error flag properly, but I couldn't build a small reproducible sample with that idea.

alaviss:
I figured out the problem: The compiler does not clear error flag before destroying refs.

```nim
proc r() =
  var o = (ref RootObj)()
  raise newException(ValueError, "")

proc main() =
  try:
    r()
  except:
    discard

main()
```

The corresponding C code around raise:

```c
	raiseExceptionEx((Exception*)T3_, "ValueError", "r", "test.nim", 3);
	goto LA1_;
	{
		LA1_:;
	}
	{
		eqdestroy___test_20(&o);
	}
	if (NIM_UNLIKELY(*nimErr_)) goto BeforeRet_;
```

If we replace the ref with an object that has destructors attached:

```nim
type
  O = object

proc `=destroy`(o: var O) =
  discard

proc r() =
  var o = O()
  raise newException(ValueError, "")

proc main() =
  try:
    r()
  except:
    discard

main()
```

Then the C code has flag clearing implemented:

```c
	raiseExceptionEx((Exception*)T2_, "ValueError", "r", "test.nim", 9);
	goto LA1_;
	{
		LA1_:;
	}
	{
		oldNimErrFin1_ = *nimErr_; *nimErr_ = NIM_FALSE;
		eqdestroy___test_2((&o));
		if (NIM_UNLIKELY(*nimErr_)) goto BeforeRet_;
		*nimErr_ = oldNimErrFin1_;
	}
	if (NIM_UNLIKELY(*nimErr_)) goto BeforeRet_;
```

Either this was an "optimization" or simply an oversight that got masked by the fact that `system/alloc` never had `goto` branches.

compiler/backend/ccgcalls.nim
```diff
@@ -11,11 +11,12 @@
...
   if n.kind == nkSym and {sfNeverRaises, sfImportc, sfCompilerProc} * n.sym.flags != {}:
```
alaviss, line 18: So this is the problematic line.

All ref-based ops are compiler procs, so error flag clearing for ref destructors will never be done as the system assumes no raises.

saem:
Awesome sleuthing, @alaviss.

Can't recall the flags off the top of my head, but I think there is one for type ops, can use that to guard. Heaping on yet more conditionals doesn't feel great, but it'll get things unstuck.

zerbina:
Thanks for investigating the issue, @alaviss.

The root cause is not directly related to `ref`-destructors. Rather, as an (incorrect) optimization, the C code generator omits the save/clear/restore error flag logic for a `finally` clause if its body doesn't raise any uncaught exceptions or defects.

Auto-generated lifetime tracking hooks have the `sfNeverRaises` flag included when they either don't call any other type-bound operators, or all of the used ones are also marked as `sfNeverRaises`. The latter was the case for the `Sugggest` type's lifted `=destroy`, which lead to the compiler-generated `finally` in `suggestEnum` not clearing the error flag after being entered as part of the unwinding started by `suggestQuit`, triggering the issue you described (calling `deallocSharedImpl` while the error mode is still active, thus skipping the release of the lock).

I made a separate PR with the fix.

compiler/backend/ccgcalls.nim
zerbina: Not quite. Auto-generated type-bound operations don't have the `sfCompilerProc` flag included. They do have the `sfNeverRaises` flag included, if the analysis in `liftdestructors` yields that they never call a user-provided operation.

In your test snippet from above, replacing `var o = O()` with `var o = (ref O)()` would make the code generator emit the error flag clear/restore properly, as the auto-generated `ref`-destructor doesn't have the `sfNeverRaises` flag set.

It's important to note that enabling `--panics:on` would break this again, as then the `canRaise` analysis is used, which treats all routines marked with `sfGeneratedOp` (which is set for all lifted type-bound operations) as never raising. Even if this wouldn't be the case, lifted operations don't inherit the effects from the operations they use, meaning that the effect specification for the lifted `=destroy` hook from the example is currently `.raises: []`.

alaviss: Yea I skipped some steps here, but what I meant is that the generated `=destroy` calls `nimDecRef*` which is `compilerproc` and is assumed to not raise and/or interact with error flags.

This is really a pain to test and I believe there will be more code that rely on these assumptions in the codebase.

zerbina: > the generated `=destroy` calls `nimDecRef*` which is `compilerproc` and is assumed to not raise and/or intera

Yep, but even them not being `.compilerproc`s wouldn't have helped here. Well, it would have prevented the dead-lock, but only because the `deallocShared` call would have never been reached then.

compiler/backend/ccgcalls.nim
zerbina: I expanded a little bit on the defects as exceptions bit, but used your suggestion otherwise.

zerbina:
Nice, that should do. Thanks again for looking into the issue, @alaviss. Next up is making `--exceptions:goto` the default and removing the other modes.

bors r+

PR #640, opened 2023-04-20
cgen: fix a severe issue with `finally`

zerbina:
## Summary

When using `--exceptions:goto`, `try` blocks inside a `finally` clause now work properly when the `finally` clause doesn't raise an *uncaught* exception or defect. Previously, the first call executed as part of the nested `try` block that can potentially raise an exception would be treated as having raised an exception.

Since it is obsolete now, remove the undocumented `enforceNoRaises` pragma introduced as a workaround for the underlying issues.

## Details

The root cause of the issues is an incorrect optimization performed by the code generator: if a `finally` clause doesn't raise an *uncaught* exception or defect, the error flag (which indicates that unwinding is in progress) was not cleared.

This is wrong. No exceptional control-flow leaving the `finally` clause does not imply that the error flag is not tested within (both directly and indirectly). If it is tested as part of calling a routine that can potentially raise, and the `finally` clause was entered as part of unwinding caused by a raised exception, unwinding would start again, regardless of whether the called routine actually raised an exception.

### The Fix

The error flag is now only left unchanged if neither the `try`'s body nor any `except` branches have exception-related control-flow effects (i.e., they must not raise an uncaught exception for the optimization to be applied).

In the future, the code generator should track for each procedure whether it observes the error flag (both directly and transitively), and use this information to safely apply the partially removed optimization.

### DIFF
--- a/compiler/ast/wordrecg.nim
+++ b/compiler/ast/wordrecg.nim
@@ -80,7 +80,6 @@ type
     wAsmNoStackFrame = "asmNoStackFrame", wImplicitStatic = "implicitStatic",
     wGlobal = "global", wCodegenDecl = "codegenDecl", wUnchecked = "unchecked",
     wGuard = "guard", wLocks = "locks", wExplain = "explain",
-    wEnforceNoRaises = "enforceNoRaises",
 
     wAuto = "auto", wBool = "bool", wCatch = "catch", wChar = "char",
     wClass = "class", wCompl = "compl", wConst_cast = "const_cast", wDefault = "default",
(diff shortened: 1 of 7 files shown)


zerbina:
This, at least partially, fixes `nimsuggest` running into a dead-lock with #639. See @alaviss [investigation](https://github.com/nim-works/nimskull/pull/639#issuecomment-1515735825) plus this [comment](https://github.com/nim-works/nimskull/pull/639#issuecomment-1516288770) for the reason why `nimsuggest` was affected.

saem:
Bors r+

Nicely done.

PR #642, opened 2023-04-22
sem: propagate `nkError` in `semCast`

zerbina:
## Summary

- make `semCast` propagate error nodes
- introduce `semTypeNode2`, which is similar to `semTypeNode`, but
  returns a node instead of a type
- handle error types in `isCastable`, `productReachable`, and
  `iterOverTypeAux`

### AST types

- extend the `CheckedAst` API with `typ` accessors and more ways to
  initialize
- disallow and prevent `CheckedAst` from storing an error node directly
- add the experimental ``ElaborateAst`` type (name is going to change)

## Details

The idea with `semTypeNode2` is that it should eventually replace
`semTypeNode`, with all of `semtypes` being refactored to return nodes
instead of types. This would allow the type-node usage sites to store
typed AST, would reduce the amount of input AST modifications, and make
error propagation in type node contexts a bit simpler.

A lesson learned from the `tsequtils.nim` test failure is that error
types (`tyError`) are sometimes either dropped or reported too late.
The direction that this commit goes into in order to prevent this, is
to produce an error node for all erroneous type-expressions appearing
in the AST.

### DIFF
--- a/compiler/ast/checked_ast.nim
+++ b/compiler/ast/checked_ast.nim
@@ -9,6 +9,10 @@
 ## The idea is to use ``CheckedAst`` as the return type of all semantic
 ## analysis procedures, eventually removing the need for wrapper errors
 ## (``adWrappedError``).
+##
+## A second experiment, ``ElaborateAst``, is also implemented here. Rather
+## than being a wrapper around ``PNode``, it acts as sort of "builder"
+## during piece-by-piece production of an AST node.
 
 import
   compiler/ast/[
@@ -25,9 +29,70 @@ type
   CheckedAst* = object
     ## Represents either in-progress or fully semanticized AST. It is meant
     ## as a wrapper for semantically *checked* AST, hence the name.
-    n: PNode       ## stores the AST. Never nil
+    n: PNode       ## stores the AST. Never nil nor an error
     hasError: bool ## tracks whether `n` contains an ``nkError`` node
                    ## somewhere
+
+  ElaborateAst* = object
+    ## *Note*: the name is not final and going to change.
+    ## Represents an in-progress or finished production of either an AST of the
+    ## same kind and number of children as the source, or an error.
+    orig: PNode ## the source node
+
+    diag*: PAstDiag       ## the error diagnostic, or nil. If the production
+                          ## finishes and the diagnostic is non-nil, an error
+                          ## is produced instead
+    hasError: bool        ## whether one of the produced children contains an
+                          ## error somewhere
+
+    children: seq[PNode]  ## the produced child nodes; has the number of
+                          ## elements as `n` has children
+    typ*: PType           ## the type of the produced expression
+
+# --------------- ElaborateAst API ------------------
+
+proc initWith*(x: var ElaborateAst, n: PNode) =
+  ## Setup the production of a new AST of the kind and with the number of
+  ## children that `n` has.
+  assert x.orig == nil, "already initialized"
+  assert n.kind in nkWithSons
+  assert n.len > 0
+  x.orig = n
+  x.children.setLen(n.len)
+
+template `[]`*(x: var ElaborateAst, i: int|BackwardsIndex): PNode =
+  x.children[i]
+
+func `[]=`*(x: var ElaborateAst, i: int|BackwardsIndex, n: PNode) {.inline.} =
+  ## Sets the child node at `i`
+  x.hasError = x.hasError or n.kind == nkError
+  x.children[i] = n
+
+func `[]=`*(x: var ElaborateAst, i: int|BackwardsIndex,
+            n: sink CheckedAst) {.inline.} =
+  ## Sets the child node at `i`
+  x.hasError = x.hasError or n.hasError
+  x.children[i] = n.n
+
+proc extract*(c: ConfigRef, n: sink ElaborateAst): PNode =
+  ## Finishes the production and returns the result as a ``PNode``.
+  result = newNodeI(n.orig.kind, n.orig.info)
+  result.flags = n.orig.flags * PersistentNodeFlags
+  result.typ   = n.typ # may be nil
+  result.sons  = move n.children
+
+  # XXX: the original node could be re-used when the children are the same as
+  #      the original ones
+  if n.diag != nil:
+    result = c.newError(result, n.diag)
+  elif n.hasError:
+    # at least one of the child nodes contains an error somewhere; use the
+    # generic wrapper error
+    result = c.wrapError(result)
+  else:
+    discard "leave as is"
+
+# -------------- CheckedAst API ------------------
 
 proc initForMappingChildren*(x: var CheckedAst, n: PNode) =
   ## Initialize `x` as a node of kind ``n.kind``. Space for all children of
@@ -45,10 +110,12 @@ func initWith*(x: var CheckedAst, n: sink PNode) =
   assert n.kind != nkError
   x.n = n
 
-proc newError*(c: ConfigRef, n: sink CheckedAst, diag: PAstDiag): CheckedAst =
-  ## Creates an ``CheckedAst`` error node.
-  result.n = c.newError(n.n, diag)
-  result.hasError = true
+func assign(x: var CheckedAst, n: sink PNode) {.inline.} =
+  ## Assigns the tree `n` to `x`. The behaviour is undefined if `n` is either
+  ## an error error or contains one.
+  assert n != nil
+  assert n.kind != nkError
+  x = CheckedAst(n: n, hasError: false)
 
 # speed up debug builds by not generating stack-trace information for the
 # accessors procedures
@@ -62,6 +129,12 @@ func len*(x: CheckedAst): int {.inline.} =
 
 func info*(x: CheckedAst): TLineInfo {.inline.} =
   x.n.info
+
+func typ*(x: CheckedAst): PType {.inline.} =
+  x.n.typ
+
+func `typ=`*(x: var CheckedAst, typ: PType) {.inline.} =
+  x.n.typ = typ
 
 proc add*(x: var CheckedAst, n: sink PNode) {.inline.} =
   x.hasError = x.hasError or n.kind == nkError
@@ -96,7 +169,7 @@ proc extract*(c: ConfigRef, n: sink CheckedAst): PNode =
   ##
   ## This procedure is only meant to be used at the edge to ``CheckedAst``-
   ## unaware code.
-  if n.hasError and n.n.kind != nkError:
+  if n.hasError:
     result = c.wrapError(n.n)
   else:
     result = n.n
--- a/compiler/ast/types.nim
+++ b/compiler/ast/types.nim
@@ -182,6 +182,9 @@ proc iterOverTypeAux(marker: var IntSet, t: PType, iter: TTypeIter,
     case t.kind
     of tyGenericInst, tyGenericBody, tyAlias, tySink, tyInferred:
       result = iterOverTypeAux(marker, lastSon(t), iter, closure)
+    of tyError:
+      # the error is reported elsewhere
+      discard
     else:
       for i in 0..<t.len:
         result = iterOverTypeAux(marker, t[i], iter, closure)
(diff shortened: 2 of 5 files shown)


zerbina:
### Notes for reviewers
- this is a blocker for #630
- not producing an error node for the type was not the only reason the test failed when using `--gc:orc`, but it nonetheless delayed error reporting

compiler/ast/checked_ast.nim
```diff
@@ -63,6 +71,12 @@ func len*(x: CheckedAst): int {.inline.} =
 func info*(x: CheckedAst): TLineInfo {.inline.} =
   x.n.info
 
+func typ*(x: CheckedAst): PType {.inline.} =
+  x.n.typ
+
+func `typ=`*(x: var CheckedAst, typ: PType) {.inline.} =
+  x.n.typ = typ
```
zerbina, line 137: There was an assertion here that disallowed setting the type of an AST that contained an error somewhere to anything else than a `tyError`, but I removed it again, as it would prevent using a non-error type for, e.g. a `(error; validExpr)` expression, which doesn't seem like a good idea.

saem: Hmm, it should probably assert if `n` itself is an error, but otherwise let it through. Previously, `n` could never be an error, but `newError2` changes that. Alternative to blocking would be to check if `n` is an error and if so, set it on `n.diag.wrongNode` (assuming that's not already an error, would require a linear traversal `var okNode = n.diag.wrongNode; while okNode.kind == nkError: okNode = okNode.diag.wrongNode`).

-----

Why I think this is the case:

The reasoning being that `(error; validExpr)` itself should be type of `validExpr` as the rule is/should be that a multi-expression-statement expression `n` takes the type and value of the last expression-statement. Formulating the rule like that keeps it neutral of errors prior to the last expression-statement.

Then a more general rule of, "any list of statements with one or more errors is itself an error" is ncie and orthogonal, and ends up with the wrapper `error(error; validExpr)`, and that wrapping error's `diag.wrongNode` will contain `(error; validExpr)`. That creates a nice cascade for detailed error diagnostics and still allows for recovery for check/suggest.

compiler/ast/checked_ast.nim
```diff
@@ -50,6 +50,14 @@ proc newError*(c: ConfigRef, n: sink CheckedAst, diag: PAstDiag): CheckedAst =
   result.n = c.newError(n.n, diag)
   result.hasError = true
 
+proc newError2*(c: ConfigRef, n: sink PNode, diag: PAstDiag): CheckedAst =
```
saem:
```suggestion
proc initAsError*(c: ConfigRef, n: sink PNode, diag: PAstDiag): CheckedAst =
```

saem: Hmm, what about storing the `PAstDiag` as a field in `CheckAst`, then `n` is kept as is and can be enforced to not be an error itself at init time and all operations can happen on `n` as today? Then when mapping `CheckedAst` to `PNode`, it can check `hasError` and:
1. if `diag` is `nil` wrap it, as today
2. otherwise, `diag` is not `nil` and the more specific error is used instead of the generic wrapper

I'm treating `CheckedAst` sorta like a lens here, where its allowing the user to ignore where the "real" `n` is and that might not be the right use case to facilitate.

compiler/ast/types.nim
```diff
@@ -1551,10 +1554,15 @@ proc productReachable(marker: var IntSet, g: ModuleGraph, t: PType, search: PTyp
     result = false
   of tyDistinct, tyGenericInst, tyAlias, tyUserTypeClassInst, tyInferred:
     result = productReachable(marker, g, t.lastSon, search, isInd)
+  of tyError:
+    # ``productReachable`` returning true usually means more work for the
+    # compiler, so we treat the error type as not being able to introduce
+    # reference cycles -- the type doesn't reach the code generators anyway
```
saem, line 1560: Makes a lot of sense, if it can't reach code gen it can't refer to memory, if it can't refer to memory it can't create cycles. :D

compiler/sem/semexprs.nim
```diff
@@ -197,6 +197,8 @@ proc isCastable(c: PContext; dst, src: PType): bool =
   #                       tySequence, tyPointer, tyNil, tyOpenArray,
   #                       tyProc, tySet, tyEnum, tyBool, tyChar}
   let src = src.skipTypes(tyUserTypeClasses)
+  if tyError in {src.kind, dst.kind}:
+    return true # error correction
```
```suggestion
    return true # error correction for check/suggest/etc
```

I remember finding it confusing to read "error correction" when I first encountered it.

compiler/ast/checked_ast.nim
```diff
@@ -50,6 +50,14 @@ proc newError*(c: ConfigRef, n: sink CheckedAst, diag: PAstDiag): CheckedAst =
...
+  result.n = c.newError(n, diag)
+  result.hasError = true
+
+func assign(x: var CheckedAst, n: sink PNode) {.inline.} =
```
saem, line 113: unused?

compiler/sem/semexprs.nim
```diff
@@ -366,29 +368,30 @@ proc semConv(c: PContext, n: PNode): PNode =
...
   checkSonsLen(n, 2, c.config)
   let
-    targetType = semTypeNode(c, n[0], nil)
+    typeExpr   = semTypeNode2(c, n[0], nil)
+    targetType = typeExpr.typ
     castedExpr = semExprWithType(c, n[1])
   
   if tfHasMeta in targetType.flags:
-    result = c.config.newError(n,
+    result = c.config.newError2(n,
```
saem: Looking at this and the example below, maybe a proc like:

`proc initWith(x: var CheckedAst, n: PNode, diag: PAstDiag = nil)`

If `diag` is `nil` then `hasError` would be `false`. Then maybe something like this would be nice:

```nim
func isError*(x: CheckedAst): bool {.inlne.} = x.n != nil and x.n.kind == nkError
func isInitialized*(x: CheckedAst): bool {.inline.} = x.n != nil

template initIfNotError*(x: var CheckedAst, n: PNode, body: typed) =
  assert x.isError or not x.isInitialized
  if not x.isError:
    x.initWith(n)
    body

# also/or this would allow a `nil` `diag` to keep on the happy path
template initIfNotError*(x: var CheckedAst, n: PNode, diag: PAstDiag, body: typed) =
  assert x.isError or not x.isInitialized
  if diag != nil:
    x.initAsError(n, diag)
  if not x.isError:
    x.initWith(n)
    body
```

which would allow things like:

```nim
  let
    typeExpr   = semTypeNode2(c, n[0], nil)
    targetType = typeExpr.typ
    castedExpr = semExprWithType(c, n[1])
    diag =
      if tfHasMeta in targetType.flags:
        PAstDiag(kind: adSemCannotCastToNonConcrete, wrongType: targeType)
      elif not isCastable(c, targetType, castedExpr.typ):
        PAstDiag(kind: adSemCannotCastTypes, ...)

  result.initIfNotError(n, diag):
    result.typ = targeType
    ...
```

saem (commented):
Mostly soft suggestions.

Curious to hear your thoughts about storing the `diag` on the side instead and also some of the API musings.

This was a really awesome PR to review, made me think of a lot of things. Perhaps one thing to experiment with would be something similar to `CheckedAst`, working title of `ElaborateAst`.

Its purpose is to track an existing node `n` of some kind `k` during some analysis, that will produce a node of the same `k` and shape, or error. Basically the `semCast` scenario, where we do piece-wise child analysis and then construct a production (same shape or error). Besides tracking errors, internally it would store an `n` and a `productionChildren: seq[PNode]`, expecting a 1 to 1 correspondence of children. Typical piece-wise analysis of each child is done and assigned to `productionChildren` positions. Then at extraction time `ElaborateAst` can check the errors, but also detect the case where if all piece-wise analysis produced the exact same nodeId no copy is requried. It would likely have a few other `PNode` fields (`flags`, `typ`, etc), but that's not a big deal.

compiler/ast/checked_ast.nim
zerbina: Hm, yeah, good call -- I hadn't considered that case. An error node should indeed not be assigned a type other than `tyError`.

I think blocking is the way to go here. Allowing one to cross the the diagnostic-to-AST border when assigning the type feels unintuitive and likely to cause issues (e.g. accidentally modifying the type of the stored `wrongNode`).

compiler/ast/checked_ast.nim
zerbina: Yep, it currently is. `assign` is/was intended to make the API a bit more ergonomic (e.g. `result.assign n` where `n` is a `PNode`).

compiler/sem/semexprs.nim
zerbina: That does collapse the code quite a bit, I like it. Each template parameter appearing in the body multiple times does seem like it could easily become a problem, however.

zerbina:
Thanks for the review, @saem. I like the idea with `ElaborateAst`! It's similar to something I tried as part of the macro-from-semProcAux separation.

Essentially, there was a `RoutineAst` type that stored relevant information about the analysed routine (e.g. the symbol, the original node, the produced nodes, etc.) and was set-up at the start of `semRoutineDef`. It removed a bit of boilerplate code from `semProcAux` and friend, but since this kind of refactoring was outside the PR's scope, I eventually dropped it again.

Anyhow, I'm going to experiment with using an `ElaborateAst` (maybe it should be named something like `Production`?) type like you described it.

compiler/ast/checked_ast.nim
saem: As discussed below, treating it like a lens will likely lead to more cleverness and likely plenty of hard to reason about bugs. We need to keep clearer delineation of error and happy paths.

compiler/ast/checked_ast.nim
saem: I would defend against error assignment, me thinks.

saem:
> Thanks for the review, @saem.

You're very welcome.

> I like the idea with `ElaborateAst`! It's similar to something I tried as part of the macro-from-semProcAux separation.

Cool beans!

> Anyhow, I'm going to experiment with using an `ElaborateAst` (maybe it should be named something like `Production`?) type like you described it.

I'm not a fan of `Elaborate`, I was trying to make it a bit specific to indicate the structure preserving change, as opposed to collapse (`dotExpr` -> qualified lookup `nkSym`, const expr eval, etc), swapping (macro pragma), etc... this way we can eventually build towards each AST transform for module AST keeping track of lineage across various transformations.

zerbina:
I added the `ElaborateAst` type, but with the note that the name is subject to change. The API is only very basic at the moment, having only the things needed for `semCast` to make use of it.

Due to `newError` requiring a `ConfigRef`, a simple subscript assignment operator involving two `ElaborateAst` instances is unfortunately not possible. `a[i, config] = b` or `assign(config, a, i, b)` would be possible, but I haven't added them yet. Alternatively, we could look into adding an overload for `newError` that doesn't require a `ConfigRef`. A `ConfigRef` is no longer strictly needed for error construction, and is only used for debugging/tracing related things.

In addition, following the discussion above, `CheckedAst` is now disallowed from storing an error node directly -- both the pre-existing `newError` and newly added `newError2` are removed.

compiler/ast/checked_ast.nim
```diff
@@ -25,10 +29,71 @@ type
...
+
+    diag*: PAstDiag       ## the error diagnostic, or nil. If the production
+                          ## finishes and the diagnostic is non-nil, an error
+                          ## is produced instead
+    hasError: bool        ## whether one of the produced children contains an
+                          ## error somewhere
+
+    children: seq[PNode]  ## the produced child nodes; has the number of
+                          ## elements as `n` has children
+    typ*: PType           ## the type of the produced expression
```
saem, line 50: (no action required) In the future, if we end up replicating a lot we can simply store a `TNode`.

saem:
> I added the `ElaborateAst` type, but with the note that the name is subject to change. The API is only very basic at the moment, having only the things needed for `semCast` to make use of it.

Looks good already, thanks!

> Due to `newError` requiring a `ConfigRef`, a simple subscript assignment operator involving two `ElaborateAst` instances is unfortunately not possible. `a[i, config] = b` or `assign(config, a, i, b)` would be possible, but I haven't added them yet. Alternatively, we could look into adding an overload for `newError` that doesn't require a `ConfigRef`. A `ConfigRef` is no longer strictly needed for error construction, and is only used for debugging/tracing related things.

I like the idea of having a new `newError` which doesn't require `ConfigRef`, seems like the way to go. Just need to think a tiny bit about the API, as direct creation of errors in that way should be used in very few spots. Could be as simple as `ElaborateAst` stores a seq of every newly generated error it sees, then when `extract`ing adds them to the `ConfigRef`. All behind conditional compilation, of course. I believe this is all outside the scope of this PR.

> In addition, following the discussion above, `CheckedAst` is now disallowed from storing an error node directly -- both the pre-existing `newError` and newly added `newError2` are removed.

🎉

compiler/ast/checked_ast.nim
```diff
@@ -25,10 +29,71 @@ type
...
     ## as a wrapper for semantically *checked* AST, hence the name.
-    n: PNode       ## stores the AST. Never nil
+    n: PNode       ## stores the AST. Never nil nor an error
     hasError: bool ## tracks whether `n` contains an ``nkError`` node
                    ## somewhere
 
+  ElaborateAst* = object
+    ## *Note*: the name is not final and going to change.
+    ## Represents an in-progress or finished production of either an AST of the
+    ## same kind as the source, or an error.
```
saem:
```suggestion
    ## Represents an in-progress or finished production of either an AST of the
    ## same kind and number of children as the source, or an error.
```

saem (approved):
That looks really good and can't wait to use it.

One suggestion, might be a bit redundant; feel free to ignore it.

compiler/ast/checked_ast.nim
zerbina: Yup, being clear here is important -- I changed it.

PR #643, opened 2023-04-22
partially refactor `semtypes`

zerbina:
## Summary

The idea/goal is to produce nodes instead of types for `semTypeNode` and most of the routines in `semtypes`. In short, this means that analysis of type AST stays at the AST level, instead of switching back and forth between types and AST.

### DIFF
--- a/compiler/ast/reports_sem.nim
+++ b/compiler/ast/reports_sem.nim
@@ -217,6 +217,7 @@ type
     stepSymNodeToNode
     stepNodeFlagsToNode
     stepNodeTypeToNode
+    stepNodeToNodeType
     stepTypeTypeToType
     stepResolveOverload
     stepNodeSigMatch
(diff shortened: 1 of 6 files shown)


zerbina:
### Notes for reviewers
- it's a very early work-in-progress. The compiler isn't able to bootstrap itself yet
- the idea originated from #642
- `CheckedAst` isn't used so far because the `CheckedAst` PR hadn't been merged yet when I started working on the refactoring
- changing routines like `semObjectNode`, `semTypeClass`, and `semProcTypeNode` to produce AST is out of the scope of this PR. For bridging to AST that produces nodes, the type they produce is assigned to a `nkType` node

zerbina:
The PR has become stale, and I'm thus closing it. Refactoring the type AST analysis procedures to return AST is still something that needs to be done at some point, but I believe it's easier to do so this on a procedure-by-procedure basis.

PR #645, opened 2023-04-24
make `goto` exceptions the default

zerbina:
## Summary

This is a preparation for the removal of the `setjmp` and `quirky` exceptions modes. `goto` exceptions were already the default when using `--gc:arc|orc`.

* fix the programs crashing when a thread was created and both thread- local storage emulation and `--exceptions:goto` were enabled
* fix execution continuing after a call to an RVO-using closure when using `--exceptions:goto`
* fix an issue where assignments were observable if the right-hand expression is a call that raised an exception
* remove the `segfaults` standard library module; its current implementation is incompatible with anything besides setjmp-exceptions
* `--exceptions:native` now means `--exceptions:goto` when using the C target

There's currently an issue with the ordering of routines in the `system` module, which prevents `nimrtl` from compiling when using `--exceptions:goto`.

## Details

The infinite recursion error that occurred when using emulated thread- local storage is caused by the procedure for accessing emulated threadvars (`GetThreadLocalVars`) getting an error flag query (read, access of a threadvar) injected, now that `system` procedures are no longer treated as never raising a defect.

The intermediate solution: disable error flag injection for all `.compilerproc`s. As a consequence, exceptional control-flow inside their  immediate bodies is no longer possible. In order to still support raise-dispatcher-like compilerprocs (e.g. `raiseFloatOverflow`), an unconditional return is inserted after the call to a `noreturn` procedure when inside a procedure that has the error flag disabled.

In addition, special-case the `threadProcWrapper` system procedure to also have the the error flag (and thus exceptional control-flow) disabled, as it is responsible for setting up the emulated thread-local storage.

### Standard Library

The `segfaults` module implicitly relied on `setjmp`-exceptions, as only `longjmp` is able to leave the signal handler and enter the NimSkull exception handler without additional support from the code generator.

Since it would become non-functional with the planned removal of `setjmp`-exceptions, it is removed already.

---

In order to simplify the code generators and to have uniform behaviour across all backends, the plan is to have the return-value optimization related logic eventually be an MIR pass.

Besides being only usable with `setjmp` exceptions, the `segfaults` module also relied on C undefined behaviour, as it used `longjmp` to exit the signal handler. If we want nil access errors to be catchable (or at least processable) again in the future, it is likely better implemented via `if` checks around each dereference of pointers that the compiler cannot statically prove to be non-nil (which could take the `not nil` annotation into account). 

## Notes for Reviewers
* blocks #630

Regarding the workaround with the compilerprocs and `threadProcWrapper`: I think it's okay for now, although I'm open to suggestion for better temporary solutions.

Eventually, we need some way to designate procedures as being part of a minimal base runtime that is exempt from the various injections (like exceptional control-flow, overflow checks, stack traces, etc.).
<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


lib/system/chcks.nim
```diff
@@ -32,7 +32,7 @@ proc raiseFieldError(f: string) {.compilerproc, noinline.} =
 when defined(nimV2):
   proc raiseFieldError2(f: string, discVal: int) {.compilerproc, noinline.} =
     ## raised when field is inaccessible given runtime value of discriminant
-    sysFatal(FieldError, f & $discVal & "'")
+    sysFatal(FieldError, formatFieldDefect(f, $discVal))
```
saem, line 35: little clean-ups like these are surprisingly uplifting... it's the opposite of tragedy of the commons.

tests/ccgbugs/tgotoexceptions.nim
```diff
@@ -0,0 +1,24 @@
...
+      discard p # capture something
+      raise CatchableError.newException("")
+
+    result = inner()
+    p[] = 1 # `p` is nil; the assignment must never be reached
+
+  try:
+    discard outer(nil)
+  except CatchableError:
+    discard
```
saem:
```suggestion
  var caught = false
  try:
    discard outer(nil)
  except CatchableError:
    caught = true
  doAssert caught
```

Soft suggestion, mostly there to catch errors introduced by someone editing the test. Comes from the philosophy that a tests should always assert/check something in case the subject under test or the test itself has a bug.

tests/exception/tassignment.nim
```diff
@@ -0,0 +1,26 @@
...
+    raise CatchableError.newException("")
+
+  proc test() =
+    var x: array[4, int]
+    try:
+      x = raiseEx(x)
+    except CatchableError:
+      discard
+
+    doAssert x[0] == 0
```
saem:
```suggestion
    except CatchableError:
      doAssert x[0] == 0, "except observed changed `x[0]`"

    doAssert x[0] == 0, "following statement observed changed `x[0]`"
```

Again, just covering bases.

compiler/backend/cgen.nim
```diff
@@ -1063,6 +1063,18 @@ proc genProcAux(m: BModule, prc: PSym) =
         #incl(res.loc.flags, lfIndirect)
         res.loc.storage = OnUnknown
 
+  # for now, we treat all compilerprocs as being able to run in a boot
+  # environment where the error flag is not yet accessible. This is not quite
+  # correct, and a dedicated facility for designating runtime procedures as
+  # usable in a boot environment is eventually required
```
saem, line 1069: One thing I can think of is that we check to see if some symbol (e.g. `system.bootRTErrorFlag`) is defined, that symbol acts as a sentinel indicating that we're far enough along in runtime environment setup that exceptions can now be used. When we analyse any code relying on try/except/raise, has raise in its effects, etc... we check for the symbol and if not present can handle it later in codegen and or raise it as a sem error as appropriate. I'm sure I've missed an important detail, but that's what currently comes to mind.

Writing it out and thinking about it a little, I'm not sure if the above is a contender for temporary solution, I suspect plenty of rearranging of `system` and related modules would be required in order to get this working.

saem (approved):
Left a few recommendations for test changes, adding asserts.

The sketch of an alternative temporary solution I did come up with seems, not all that temporary wrt implementation time. 😅

tests/ccgbugs/tgotoexceptions.nim
zerbina: Yep, good point. I was a bit too hasty here.

compiler/backend/cgen.nim
zerbina: Yeah, I think checking for the presence of a symbol or for a procedure's owner is the way to go.

As you noted, the `system` has to be reordered quite a bit for that to work. That has to happen eventually anyway, but at this point, I'd wait with these efforts until the `system` and related modules are no longer shared between the normal compiler and csources compiler, as we don't need to consider compatibility as much then.

saem:
bors r+

let's go!!!! 🚀

saem:
There might be a bit of a race here with bors vs merge queues. 😆

PR #647, opened 2023-04-26
sigmatch: simpler procedural type matching

zerbina:
## Summary

* simplify the logic for handling generic routines using `auto` as the
  return type
* reject un-instantiated routines earlier when used in branch and array
  constructor expressions, resulting in clearer error messages
  (`has no concrete type` instead of `cannot instantiate`)
* fix spurious `cannot instantiate` errors when passing generic routines
  as arguments to parameters that are also generic
* support lambda expressions coming from template expansions

## Details

Introduced by https://github.com/nim-lang/Nim/pull/3234, when both the
formal and actual type are unresolved meta types, the relation is
considered to be `isBothMetaConvertible`. The intention seems to have
been supporting passing `auto`-returning routines or lambda-expressions
to generic procedure-type parameters, but the logic introduced by the
aforementioned PR is not necessary for achieving that.

`auto` return types now use dedicated handling in `procTypeRel`, both
the formal and actual type being unresolved generics results in a
mismatch again, and `typeRel` is no longer applied repeatedly for
`isBothMetaConvertible` matches.

There was also the issue of `generateTypeInstance` being misused, which
could result in spurious `cannot instantiate` or internal compiler
errors. Until type instantiation properly supports the case where it's
not certain whether all used generic parameters have been inferred
already, the `tryGenerateInstance` routine is introduced as a temporary
solution.

Finally, both `nkStmtListExpr` and `nkBlockExpr` nodes are now
considered when instantiating a generic routine as part of parameter
matching, which allows for complex expressions yielding un-instantiated
generic routines (e.g.: `call((discard x; genericRoutine))`).

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1785,10 +1785,9 @@ type
     isSubtype
     isSubrange               ## subrange of the wanted type; no type conversion
                              ## but apart from that counts as ``isSubtype``
-    isBothMetaConvertible    ## generic proc parameter was matched against
-                             ## generic type, e.g., map(mySeq, x=>x+1),
-                             ## maybe recoverable by rerun if the parameter is
-                             ## the proc's return value
+    isBothMetaConvertible    ## a generic procedure that otherwise matched has
+                             ## an 'auto' return type; it needs to be instantiated
+                             ## first
     isInferred               ## generic proc was matched against a concrete type
     isInferredConvertible    ## same as above, but requiring proc CC conversion
     isGeneric
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1785,9 +1785,9 @@ type
     isSubtype
     isSubrange               ## subrange of the wanted type; no type conversion
                              ## but apart from that counts as ``isSubtype``
-    isBothMetaConvertible    ## a generic procedure that otherwise matched has
-                             ## an 'auto' return type; it needs to be instantiated
-                             ## first
+    isBothMetaConvertible    ## a generic procedure with an 'auto' return type
+                             ## that otherwise matched; it needs to be
+                             ## instantiated first
     isInferred               ## generic proc was matched against a concrete type
     isInferredConvertible    ## same as above, but requiring proc CC conversion
     isGeneric
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -243,9 +243,10 @@ proc fitNodeConsiderViewType(c: PContext, formal: PType, arg: PNode; info: TLine
   else:
    result = a
 
-proc genericProcCheck(c: PContext, n: PNode): PNode =
-  ## Checks that the analysed expression `n` is of generic procedural type.
-  ## Returns either `n` or an error.
+proc exprNotGenericRoutine(c: PContext, n: PNode): PNode =
+  ## Checks that the analysed expression `n` is *not* an uninstantiated generic
+  ## routine, and returns an error node if it is one. In the case of no error,
+  ## `n` is returned.
   if n.typ != nil and n.typ.kind == tyError:
     return n
 
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -243,6 +243,23 @@ proc fitNodeConsiderViewType(c: PContext, formal: PType, arg: PNode; info: TLine
   else:
    result = a
 
+proc genericProcCheck(c: PContext, n: PNode): PNode =
+  ## Checks that the analysed expression `n` is of generic procedural type.
+  ## Returns either `n` or an error.
+  if n.typ != nil and n.typ.kind == tyError:
+    return n
+
+  # skip all statement list wrappers:
+  var it {.cursor.} = n
+  while it.kind == nkStmtListExpr:
+    it = it.lastSon
+
+  if (it.kind == nkSym and it.sym.isGenericRoutineStrict) or
+     it.isGenericRoutine:
+    c.config.newError(n, PAstDiag(kind: adSemProcHasNoConcreteType))
+  else:
+    n
+
 proc inferWithMetatype(c: PContext, formal: PType,
                        arg: PNode, coerceDistincts = false): PNode
 
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -252,7 +252,7 @@ proc exprNotGenericRoutine(c: PContext, n: PNode): PNode =
 
   # skip all statement list wrappers:
   var it {.cursor.} = n
-  while it.kind == nkStmtListExpr:
+  while it.kind in {nkStmtListExpr, nkBlockExpr}:
     it = it.lastSon
 
   if (it.kind == nkSym and it.sym.isGenericRoutineStrict) or
--- a/compiler/sem/semtypinst.nim
+++ b/compiler/sem/semtypinst.nim
@@ -823,3 +823,34 @@ proc prepareMetatypeForSigmatch*(p: PContext, pt: TIdTable, info: TLineInfo,
 template generateTypeInstance*(p: PContext, pt: TIdTable, arg: PNode,
                                t: PType): untyped =
   generateTypeInstance(p, pt, arg.info, t)
+
+proc tryGenerateInstance*(c: PContext, pt: TIdTable, info: TLineInfo, t: PType): PType =
+  ## Tries to resolve the generic type `t` to a concrete type. Returns `nil` on
+  ## failure, and the resolved type on success.
+  ##
+  ## XXX: this procedure is only a workaround. ``replaceTypeVarsT`` should
+  ##      properly support the case where it's not certain whether all
+  ##      referenced type parameters are resolved already
+  assert containsGenericType(t)
+  result = prepareMetatypeForSigmatch(c, pt, info, t)
+
+  proc containsGenericType2(t: PType): bool =
+    if t.kind == tyGenericInst:
+      # check the parameters:
+      for i in 1..<t.len-1:
+        if t[i].isMetaType:
+          return true
+
+    result = containsGenericType(t)
+
+  # ``handleGenericInvocation`` doesn't properly propagate the ``tfHasMeta``
+  # flag, so we don't rely on it here. We also need to use a special-purpose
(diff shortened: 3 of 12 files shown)


zerbina:
Regarding the effect on overload quality, the only difference is for generic procedural types with an `auto` return type. If after instantiation, the concrete procedural type matches exactly, the match is now considered as 1 conversion + 1 generic instead of 1 conversion + 1 exact. Example case:
```nim
proc call(x: proc(x: int): int {.nimcall.}) = discard

proc generic[T](x: T): auto {.nimcall.} =
  result = x

call(generic) # after instantiation, the types match exactly
```

I'm unsure whether this is a good idea or not.

compiler/sem/sem.nim
```diff
@@ -243,6 +243,23 @@ proc fitNodeConsiderViewType(c: PContext, formal: PType, arg: PNode; info: TLine
   else:
    result = a
 
+proc genericProcCheck(c: PContext, n: PNode): PNode =
```
saem: It only generiates an error if the expression is a generic routine.

Would this be a better name: `exprNotGenericRoutine`?

compiler/ast/ast_types.nim
```diff
@@ -1768,10 +1768,9 @@ type
     isSubtype
     isSubrange               ## subrange of the wanted type; no type conversion
                              ## but apart from that counts as ``isSubtype``
-    isBothMetaConvertible    ## generic proc parameter was matched against
-                             ## generic type, e.g., map(mySeq, x=>x+1),
-                             ## maybe recoverable by rerun if the parameter is
-                             ## the proc's return value
+    isBothMetaConvertible    ## a generic procedure that otherwise matched has
+                             ## an 'auto' return type; it needs to be instantiated
+                             ## first
```
```suggestion
    isBothMetaConvertible    ## a generic procedure with an 'auto' return type
                             ## that otherwise matched; it needs to be
                             ## instantiated first
```

I thought separating the exceptional aspect (auto return type) from the description of the subject of the match (generic routine with auto return) makes it easier to follow.

compiler/sem/semtypinst.nim
```diff
@@ -793,6 +793,10 @@ proc recomputeFieldPositions*(t: PType; obj: PNode; currPosition: var int) =
 
 proc generateTypeInstance*(p: PContext, pt: TIdTable, info: TLineInfo,
                            t: PType): PType =
+  ## Produces the instantiated type for the generic type `t`, using the
+  ## bindings provided by `pt`. All type variables used by `t` must have
+  ## *concrete* type bounds to them -- both meta types and missing bindings
```
```suggestion
  ## *concrete* types bound to them -- both meta types and missing bindings
```

compiler/sem/sigmatch.nim
```diff
@@ -509,20 +510,20 @@ proc procParamTypeRel(c: var TCandidate, f, a: PType): TTypeRelation =
...
-  ##    Q=int mapping will be saved to c.bindings.
-  ## - matches f=S to a=W. Both of these metatypes are unknown, so we
-  ##    return with isBothMetaConvertible to ask for rerun.
-  ## - matches f=S to a=W. At this point the return type of innerProc
-  ##    is known (we get it from c.bindings). We can use that value
-  ##    to match with f, and save back to c.bindings.
+  ## two different calls:
+  ## - matches `f`=T to `a`=Q. `f` is a resolved metatype ('int' is bound to
+  ##   it already), so `a` can be inferred; a binding for Q=int is saved
+  ## - matces `f`=U to `a`=Q. `f` is an unresolved metatype, but since Q was
```
```suggestion
  ## - matches `f`=U to `a`=Q. `f` is an unresolved metatype, but since Q was
```

compiler/sem/sigmatch.nim
```diff
@@ -539,12 +541,14 @@ proc procParamTypeRel(c: var TCandidate, f, a: PType): TTypeRelation =
...
-        f = generateTypeInstance(c.c, c.bindings, c.call.info, f)
+        f = prepareMetatypeForSigmatch(c.c, c.bindings, c.call.info, f)
       else:
         f = nil
-      if f.isNil() or f.isMetaType:
-        # no luck resolving the type, so the inference fails
-        return isBothMetaConvertible
+      if f.isNil() or containsGenericType(f):
+        # no luck resolving the type, so the inference fails. Note that using
+        # ``isMetaType`` wont work, as it doesn't check for nested
```
```suggestion
        # ``isMetaType`` won't work, as it doesn't check for nested
```

compiler/sem/sigmatch.nim
```diff
@@ -2078,6 +2086,62 @@ template matchesVoidProc(t: PType): bool =
...
+    result =
+      if inferred.isError:
+        inferred.ast
+      else:
+        newSymNode(inferred, n.info)
+  of nkProcTy, nkIteratorTy:
+    # possible in a concept context. There's nothing to instantiate
+    return nil
+  else:
+    # nothing else is able to provide uninstantiated generic routines
```
saem, line 2121: (just thinking outloud; entirely ignorable)

This isn't a bug, just an observation, sigmatch has an assumption that all name lookups have been complete and we're dealing with symbols only (not counting untyped AST parameter values, but sigmatch just passes those around like opaque values).  This is one of those invariants of the module that I'm figuring out how to capture in the long run (entirely outside this PR).

compiler/sem/sem.nim
```diff
@@ -243,6 +243,23 @@ proc fitNodeConsiderViewType(c: PContext, formal: PType, arg: PNode; info: TLine
...
+  ## Checks that the analysed expression `n` is of generic procedural type.
+  ## Returns either `n` or an error.
+  if n.typ != nil and n.typ.kind == tyError:
+    return n
+
+  # skip all statement list wrappers:
+  var it {.cursor.} = n
+  while it.kind == nkStmtListExpr:
```
saem: Do we also need to consider block expressions?

tests/lang_callable/overload/tproc_type_inference.nim
```diff
@@ -0,0 +1,95 @@
...
+  # complex expressions yielding generic routines also work
+  proc call(val: int, x: proc(x: int): int) =
+    doAssert x(val) == val
+
+  proc generic[T](x: T): auto =
+    result = x
+
+  call(1, (discard ""; generic))
+  # also works with lambda expressions:
+  call(2, (discard ""; (proc(x: auto): auto = x)))
```
saem:
```suggestion
  call(1, (discard ""; generic))
  call(2):
    block:
      discard ""
      generic
  # also works with lambda expressions:
  call(3, (discard ""; (proc(x: auto): auto = x)))
  call(4):
    block:
      discard ""
      (proc(x: auto): auto = x)
```

Just to illustrate the block expression case.

tests/lang_callable/overload/tproc_type_inference.nim
```diff
@@ -0,0 +1,95 @@
...
+  # was inferred as
+  proc call[A, B](val: A, x: proc(a: A, b: B): bool) =
+    doAssert x(val, val) == true
+
+  proc generic[A](a: A, b: A): bool =
+    result = true
+
+  call(1, generic)
+  # a lambda expression with two parameters using the same type variable
+  # (i.e., generic parameter) doesn't work
```
saem: (no action needed) I'm still tracing through figuring out why this is the case, but I'm guessing this is due to a pre-existing issue that begins with the `semGenerateInstance` call within `instantiateRoutineExpr`? If you don't know that's ok, just want to make sure I'm "digging in the right spot".

saem (approved):
This is far more sensible than before. 🎉 

I made some doc suggestions and possibly discovered a case that needs covering (`nkBlockExpr`).

tests/lang_callable/overload/tproc_type_inference.nim
zerbina: It's because a lambda expression with explicit generic parameters is not supported by the parser.

```nim
call(1, (proc[T](x: T, y: T) = discard))
```
is not valid code.

compiler/sem/sem.nim
zerbina: Hmm, good question. From the top of my head, I can't think of a reason why we don't want to support that case.

tests/lang_callable/overload/tproc_type_inference.nim
saem: Thanks -- it's always a bit surprising what isn't supported. Although it might be tricky in ways I don't yet realize.

saem:
> Regarding the effect on overload quality, the only difference is for generic procedural types with an `auto` return type. If after instantiation, the concrete procedural type matches exactly, the match is now considered as 1 conversion + 1 generic instead of 1 conversion + 1 exact. Example case:
> 
> ```nim
> proc call(x: proc(x: int): int {.nimcall.}) = discard
> 
> proc generic[T](x: T): auto {.nimcall.} =
>   result = x
> 
> call(generic) # after instantiation, the types match exactly
> ```
> 
> I'm unsure whether this is a good idea or not.

It's not ideal, but I'm not quite sure if +1 generic and exact is either. If it's cumbersome to fix right now I would note it in a comment as an issue and move on.

compiler/sem/sem.nim
zerbina: They're now also considered. Thanks for raising this inconsistency!

zerbina:
I have reverted the effects an instantiated `auto`-returning generic has on an overloads quality back to how they were before this PR.

The change with `prepareMetatypeForSigmatch` was also not quite correct:
- it doesn't always produce proper instances (because the instance cache is partially disabled), which causes subtle issues and can lead to C compiler errors
- the `isMetaType` + `containsGenericType` check isn't able to detect generic phantom types like, for example, `Type[T] = object`, which would lead to compiler assertions (rightfully) failing

To more properly fix the misuse of `generateTypeInstance` for the time being, I introduced the `tryGenerateInstance` routine. It uses `prepareMetatypeForSigmatch` as a way to check if all referenced generic parameters have been inferred already, and only if that is the case creates a proper instance via `generateTypeInstance`. This is rather inefficient, but it's the most robust and low-impact temporary solution that I could come up with.

zerbina:
That should fix the test failures. I had forgotten that the `tfHasMeta` flag is also not *excluded* properly, meaning that it cannot be relied on at all for types produced by `replaceTypeVarsT` when meta-types are allowed.

saem (approved):

saem:
This is a good step forward in making this bit of sem make more sense.

PR #648, opened 2023-04-26
cgen: handle uncaught exceptions at module exit

zerbina:
## Summary

For the C target, uncaught exceptions were only detected and reported (via `nimTestErrorFlag`) when raised from inside the main module, but not when raised from inside others.

When starting to execute the next module, the error mode was still active, which, depending on the contents of the module's module-level code, could lead to a cascade of failures eventually resulting in a crash unrelated to the original exception.

A call to `nimTestErrorFlag` is now inserted at the exit of every module's init procedure.

## Misc

- fix a `FieldDefect` error in `vmrunner` that ocurred when reporting an unhandled exception

---

## Notes for Reviewers
* the issues surfaced during megatest execution in #630

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/backend/cgen.nim
+++ b/compiler/backend/cgen.nim
@@ -1588,9 +1588,10 @@ proc genInitCode(m: BModule) =
     if beforeRetNeeded in m.initProc.flags:
       prc.add(~"\tBeforeRet_: ;$n")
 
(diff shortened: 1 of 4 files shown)


saem (approved):
nice catch.

bors r+

PR #654, opened 2023-04-27
cli: dump output moved out of legacy reports

saem:
## Summary
The `dump` sub-commands' output generation is back in `main`.

## Details
This moved the `InternalStateDump` data structure out of legacy reports
and adding some output helpers to `commands` (temporary home). This is
part of a long term migration away from legacy reports.

### DIFF
--- a/compiler/front/commands.nim
+++ b/compiler/front/commands.nim
@@ -28,11 +28,9 @@ import
   std/[
     os,
     strutils,
-    parseutils,
     parseopt,
     sequtils,
     strtabs,
-    pathnorm
   ],
   compiler/modules/[
     nimblecmd,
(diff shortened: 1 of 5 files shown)


compiler/front/commands.nim
```diff
@@ -80,6 +78,27 @@ type
     passCmd2,                 # second pass over the command line
     passPP                    # preprocessor called processCommand()
 
+# TODO: temporary, move into `msgs` or `commands`
```
zerbina, line 81:
```suggestion
```
I believe the TODO doesn't apply anymore?

zerbina (approved):
There's a leftover todo comment I believe, but it looks good to go otherwise.

compiler/front/commands.nim
saem: I'm not quite sure about the current home. 😕

PR #656, opened 2023-04-27
system: remove the `shallow` procedure

zerbina:
## Summary

The `shallow` procedure was used to toggle whether a `seq` or `string`
uses a shallow copy when copying it. Given that it only has an effect
for the C target and when using the sequence implementation from the old
runtime, it is removed as a preparation for ORC becoming the default and
the old runtime being eventually removed.

## Details

* remove all usages of `shallow` from the compiler and tests
* remove the compiler-internal and experimental `forceCopy` procedure
  (it no longer has a purpose)
* remove mentions of the `shallow` procedure from documentation

The compiler-internal and experimental `setLen` procedure from the
`strutils2` is also removed, as it is not compatible with the string
implementation from the new runtime.

### DIFF
--- a/compiler/backend/ccgtypes.nim
+++ b/compiler/backend/ccgtypes.nim
@@ -84,7 +84,6 @@ proc scopeMangledParam(p: BProc; param: PSym) =
   ## generate unique identifiers reliably (consider that ``var a = a`` is
   ## even an idiom in Nim).
   var key = param.name.s.mangle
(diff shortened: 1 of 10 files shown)


saem (approved):
/merge

Yay, less silly code.

alaviss:
/merge

@saem The weak point of this new stuff is that you have to write the command in a regular comment, not an approval one

PR #657, opened 2023-04-27
system: use a `seq` with `WideCStringObj`

zerbina:
## Summary

When using the old runtime, the implementation was `ref`-based
(requiring the use of `unsafeNew`), while with the new runtime it was
destructor-based.

Change `WideCStringObj` to store the character buffer as a
`seq[Utf16Char]`, which has the benefit that:
- only a single implementation is required
- the implementation becomes simpler
- the type is usable at compile-time and with the JS and VM target

The downside is that each `WideCStringObj` requires 4 or 8 byte
(depending on the target's integer size) more heap memory because of the
capacity stored by the `seq` as part of the payload. The public API
stays the same.

### DIFF
--- a/lib/system/widestrs.nim
+++ b/lib/system/widestrs.nim
@@ -16,52 +16,30 @@
 type
   Utf16Char* = distinct int16
 
-when defined(nimv2):
-
+when true:
   type
     WideCString* = ptr UncheckedArray[Utf16Char]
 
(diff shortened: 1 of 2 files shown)


saem (approved):
/merge

Much better; it's pure library code, so if not easier to maintain it gets deleted wholesale. 😁

PR #583, opened 2023-03-14
Saem cli remove legacy reports

saem:
## Summary
* remove legacy reports from various CLI

## Details
* tbd

---

## Notes for Reviewers
* still heavily reworking this and pulling bits out into smaller PRs

quick win list (doesn't include already completed):

- [x] `main.commandGenDepend` and `nim.handleCmdLine` simply calls `execCmd` dropping reports and `extccomp.execExternalProgram` dep
- [x] drop `gcstats` "hint"
- [x] remove `@trace` and other debug junk introduced in cfg file processing
- [x] add missing names for report enums (`rsemHintLibDependency = "Dependency"` and `rcmdRunnableExamplesSuccess = "Success"`)
- [x] move cli/sexp_reporting changing out of `commands`
- [x] drop reports from `nimblecmd`
- [x] make comptime compiler option/option arg testing functional
- [x] rework `commands.processSpecificNote` in a functional style

Then two bigger changes, which might need to be combined:

- [x] make option processing and friends return a result
- [x] use command line events for messages
- [x] remove unused legacy reports
- [x] clean-up warnings

PR #658, opened 2023-04-27
cfg: drop project ".nimcfg" support

saem:
## Summary

It was possible to defined a project config with a `.nimcfg` extension,
this is no longer supported. Now config file discovery is a bit simpler
and this half-baked bug-in-waiting is removed.

## Details

`nimconf` now uses a simple iterator to produce a list of config files
to attempt processing, making its implementation easier to follow.
Removed ".nimcfg" from `options.findProjectNimFile`'s heuristic.

### DIFF
--- a/compiler/front/nimconf.nim
+++ b/compiler/front/nimconf.nim
@@ -422,44 +422,36 @@ proc getSystemConfigPath*(conf: ConfigRef; filename: RelativeFile): AbsoluteFile
     if not fileExists(result): result = p / RelativeDir"etc/nim" / filename
     if not fileExists(result): result = AbsoluteDir"/etc/nim" / filename
 
+iterator configFiles(N: NimConfParser, cfg: RelativeFile): AbsoluteFile =
+  # xxx: maybe this should open lexers instead
+  if optSkipSystemConfigFile notin N.config.globalOptions:
+    yield getSystemConfigPath(N.config, cfg)
+
+  if optSkipUserConfigFile notin N.config.globalOptions:
+    yield getUserConfigPath(cfg)
+
+  let pd = if N.config.projectPath.isEmpty: AbsoluteDir(getCurrentDir())
+           else:                            N.config.projectPath
+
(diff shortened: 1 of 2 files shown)


compiler/front/nimconf.nim
```diff
@@ -422,44 +422,36 @@ proc getSystemConfigPath*(conf: ConfigRef; filename: RelativeFile): AbsoluteFile
...
-    N: var NimConfParser, cfg: RelativeFile, cache: IdentCache
-  ) =
-  setDefaultLibpath(N.config)
-
-  proc readConfigFile(N: var NimConfParser, path: AbsoluteFile) =
-    let configPath = path
-    if readConfigFile(N, configPath, cache):
-      N.config.configFiles.add(configPath)
-
+iterator configFiles(N: NimConfParser, cfg: RelativeFile): AbsoluteFile =
```
zerbina, line 425: Only a very soft suggestion, but given that only the `N.config` is used, a `ConfigRef` could be passed directly. If this conflicts with your plans for the iterator (e.g. opening lexers), feel free to disregard the suggestion.

zerbina (approved):
Nice, iterators for the win :D

Edit: I believe you meant to reference the PR #583 in the "notes for reviewers" section?

compiler/front/nimconf.nim
saem: I can do this after the merge if it makes sense. I'm trying to make all these little PRs close enough to the big one (`saem-cli-remove-legacy-reports`) so I can shrink the delta between it and `devel`.

I'm also not quite sure where things are going to land, because the `ConfigRef` we're passing around should probably be "just data" and no user visible actions be taken based on it. Then when we know it's good it should get merged into the "real" `ConfigRef`... but that approach might be off for reasons I don't quite understand as of yet.

saem:
> Nice, iterators for the win :D

ikr?! I compared it to the old code and was floored how much simpler the new stuff is.

> Edit: I believe you meant to reference the PR #583 in the "notes for reviewers" section?

I sure did! updated.

PR #661, opened 2023-04-27
streams: support copy-on-write string literals

zerbina:
## Summary

String literals use a copy-on-write mechanism when the new runtime is
enabled. Prior to mutating them through an unsafe facility like
pointers or `cstring`, the `prepareMutation` routine has to be
called on the string.

Adjust the routines modifying strings through unsafe facilities to use
the compatibility template `prepareStrMutation` (a no-op when the old
runtime is used).

### DIFF
--- a/lib/pure/streams.nim
+++ b/lib/pure/streams.nim
@@ -100,6 +100,13 @@ proc newEIO(msg: string): owned(ref IOError) =
   new(result)
   result.msg = msg
 
+template prepareStrMutation(x: var string) =
+  ## A compatibility template that is meant to be used until the
+  ## old runtime is removed.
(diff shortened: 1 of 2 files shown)


saem:
Change looks good, for the PR/commit message:

> Adjust the string the routines modifying strings through unsafe
> facilities to use the compatibility template `prepareStrMutation` (a
> no-op when the old runtime is used).

Did you mean?

```
Adjust the routines modifying strings through unsafe facilities to use
the compatibility template `prepareStrMutation` (a no-op when the old
runtime is used).
```

zerbina:
> Did you mean?

Yep, I indeed did. It's fixed now.

PR #664, opened 2023-04-27
improve `getAst` magic processing

zerbina:
## Summary

Move part of the processing into `transf` and transform the magic call
into something that matches its semantics better. This fixes a small
issues where the `getAst` magic appeared to return nothing when the
meta-routine to expand is a macro.

## Details

A `getAst` expression like `getAst(templ(a, b, c))` is now transformed
into `getAst(templ, a, b, c)` during `transf`. This is slightly more
correct at the MIR level, as `templ` is not "called" here, but is rather
an argument to the `getAst` magic.

* remove the `mirgen` special handling of `static` parameters for
  templates and macros in a `getAst` context. Attempting to work around
  the issues with `getAst` like this is not a good idea
* adjust `vmgen` to work to the new flat representation of `getAst`
  calls
* remove the `vmGenDiagBadExpandToAstArgRequired` diagnostic (semantic
  analysis already rejects ill-formed `getAst` argument expressions)

The regression was due to `vmgen.genCall` using the type of the
macro call expression (which incorrectly was `nil` when the macro had no
return type specified) for detecting whether it returns something.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -876,61 +876,31 @@ proc genCall(c: var TCtx, n: PNode): EValue =
 
   finishCall(c, n[0], fntyp)
 
-proc genMacroCall(c: var TCtx, n: PNode): EValue =
-  ## Generates the code for a macro/template call expression. These can only
-  ## occur in a ``getAst`` context.
-  ##
-  ## The argument expressions require special handling, so a dedicated
-  ## procedure is used in order to encapsulate the required workarounds in a
-  ## single place, making it easier to spot, document, and eventually fix them.
-  assert n[0].kind == nkSym and n[0].sym.kind in {skMacro, skTemplate}
+proc genMacroCallArgs(c: var TCtx, n: PNode, kind: TSymKind, fntyp: PType) =
+  ## Generates the arguments for a macro/template call expression. `n` is
+  ## expected to be a ``getAst`` expression that has been transformed to the
+  ## internal representation. `kind` is the meta-routine's kind, and `fntyp`
+  ## its signature.
+  assert kind in {skMacro, skTemplate}
+  chain: genCallee(c, n[1]) => arg(c)
 
-  # XXX: ``genAst(templ(x))`` (where `templ` is a template) should be turned
-  #      into:
-  #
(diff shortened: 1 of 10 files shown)


compiler/mir/mirgen.nim
```diff
@@ -1094,35 +1064,30 @@ proc genMagic(c: var TCtx, n: PNode; m: TMagic): EValue =
...
       argBlock(c.stmts):
-        for i in 1..<n.len:
-          assert n[i].typ == nil
-          # even the argument expressions don't have types in this case, so we
-          # can't use ``argExpr``
-          # FIXME: this happens in the context of ``quote`` (i.e.
-          #        ``semQuoteAst``), because the quoted expressions are not
-          #        semantically analysed.
-          #        The arguments (quoted expressions) should be analysed and
-          #        the call should also use a complete symbol.
```
zerbina, line 1108: The comment was and is incorrect. The quoted expressions are run through semantic analysis -- only the `getAst` symbol has no type information, as it is created via `createMagic` in `semQuoteAst`.

tests/vm/tvmgen_regressions.nim
```diff
@@ -16,4 +16,18 @@ block:
...
\ No newline at end of file
+  doAssert o == 1.0
+
+block wrong_getast:
+  macro m() = # <- no return type specified
+    result = quote: 1
+
+  macro m2() =
+    # the macro's public signature was used to detect whether it returns
+    # something, which lead to incorrect bytecode being generated in situations
```
saem:
```suggestion
    # something, which led to incorrect bytecode being generated in situations
```

saem (approved):
One minor typo (optional) in a comment, otherwise good to go.

We could probably simplify things a bunch here and elsehwere if we had a `nkAstLit`, it would nicely fence analysis, but I think that's been brought up before.

tests/vm/tvmgen_regressions.nim
zerbina: Thanks. Fixed it.

zerbina:
/merge

> We could probably simplify things a bunch here and elsehwere if we had a `nkAstLit`, it would nicely fence analysis, but I think that's been brought up before.

Maybe it's what you're referring to, but we do have a `nkNimNodeLit` node kind.

PR #665, opened 2023-04-28
astgen: fix issues with the MIR -> AST step

zerbina:
## Summary

Fix two issues in `astgen` where the semantics of the MIR were not
correctly mapped to `PNode` AST. This led to `wasMoved` operations only
being applied to intermediate temporary locations in some situations,
causing double frees.

## Details

The first issue was that an expression like `(a; lvalue).c` was not
detected to be an lvalue expression due to the first operand being a
statement list expression (`nkStmtListExpr`), causing a temporary
instead of an lvalue reference (view) to be created when using the
expression as a to-be-consumed operand to a MIR region.

If the operand was passed on to a `wasMoved` operation, only the
temporary location is reset, but not the expected source location. To
visualize the problem:
```nim
  x = callReturningASeq()[0] # x is something with a destructor
  # gets transformed to MIR code that, after translation back to AST,
  # looked roughly like:
  var tmp: T
  var tmp2: T = (tmp = callReturningASeq(); tmp)[0]
  x = tmp2
  wasMoved(tmp2) # `tmp[0]` is what needs to be reset here
  ...
```

For fixing the issue, the existing unwrapping transformation that is
applied prior to detecting whether an expression is a view is extended
to also "forward" nodes: an expression like `(a; b).c` gets transformed
into `(a; b.c)`. Besides fixing the issue, this also reduces nesting of
the generated AST a bit.

Using the example from above, it now looks like:
```nim
  var tmp: T
  tmp = callReturningASeq()
  var tmp2: var T = tmp[0]
  x = tmp2
  wasMoved(tmp2)
  ...
```
which is correct.

The second issue was with 'consume' arguments of numeric type not being
passed translated to lvalue references (if possible).

### DIFF
--- a/compiler/mir/astgen.nim
+++ b/compiler/mir/astgen.nim
@@ -248,14 +248,52 @@ proc wrapArg(stmts: seq[PNode], info: TLineInfo, val: PNode): PNode =
     result = newTreeIT(nkStmtListExpr, info, val.typ, stmts)
     result.add val
 
-func unwrap(expr: PNode, stmts: var seq[PNode]): PNode =
-  if expr.kind == nkStmtListExpr:
-    for i in 0..<expr.len-1:
-      stmts.add expr[i]
+proc flatten(expr: PNode, stmts: var seq[PNode]): PNode =
+  ## Applies the following transformations to the expression `expr`:
+  ## 1.) transform an AST like: ``(A (StmtListExpr (x) (y)) (z))`` into
+  ##     ``(StmtListExpr (x) (A (y) (z)))``
+  ## 2.) add all statements from a statement list expression to `stmts` and
+  ##     pass on the last node
+  ##
+  ## The result of #2 is passed back to step #1 and the process is repeated
+  ## until none of the two transformations apply anymore. Together, these
+  ## transformations make processing easier for the following analysis steps,
+  ## and the generated AST a bit less nested.
+  proc forward(n: PNode, p: int): PNode =
+    ## Performs transformation #1
+    if n[p].kind == nkStmtListExpr:
+      result = n[p]
+      n[p] = result[^1]
+      result[^1] = n
+    else:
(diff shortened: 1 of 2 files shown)


saem:
Just read the PR description alone and this sounds like a general transform that we want to have, where we normalize for this property (it's sorta the opposite of distributive).

Update: yup, and that's what the PR does with `flatten`, yay!

compiler/mir/astgen.nim
```diff
@@ -248,14 +248,52 @@ proc wrapArg(stmts: seq[PNode], info: TLineInfo, val: PNode): PNode =
...
+      nkDerefExpr, nkHiddenDeref, nkCStringToString, nkStringToCString,
+      nkObjDownConv, nkObjUpConv:
+      it = forward(it, 0)
+    of nkConv, nkHiddenStdConv, nkCast:
+      it = forward(it, 1)
+    else:
+      # no AST to which transform #1 applies
+      discard
+
+    if it.kind == nkStmtListExpr:
```
saem, line 286: (not sure if I'd make this change, you might have already thought of it)

technically this could join the `case` statement above? And assuming the consts existed, the while loop predicate could be `it.kind notin nkSideEffectFreeKinds + nkCastAndConvKinds + {nkStmtList}`?

Not sure if that's necessarily easier to reason about, though.

saem (approved):
The transforms are nice and clear, thanks.

saem:
If the tests pass, I say it's good to merge as is; unless you feel the suggestion improves readability.

compiler/mir/astgen.nim
zerbina: Yep, I have, and ultimately decided against joining both statements. The extra `if` does stand a bit out, but I think the step-by-step aspect makes it a tiny bit more clearer.

PR #672, opened 2023-04-29
mirgen: ignore `.compileTime` variables

zerbina:
## Summary

Don't emit a 'def' for `.compileTime` globals in `mirgen`. Until the
implementation of `.compileTime` variables gets an overhaul, this
restores their lazy-generation behaviour when destructors are used, but
also means that they're for now not destroyed anymore when going out of
scope.

## Details

Lazy-generation not working when destructors are used led to issues when
the `.compileTime` variable in question has an initializer expression
that makes use of compile-time-only procedures.

* ignore definitions of `.compileTime` variables outside of non-compile-
  time contexts in `mirgen`
* remove now obsolete `.compileTime` variable handling from `cgen` and
  `jsgen`
* run the `tunique_type.nim` test with `--gc:orc`

Registering destructors for `.compileTime` variables that require them
is possible, but skipped here because it would only introduce more
obsolete-on-arrival code.

### DIFF
--- a/compiler/backend/ccgstmts.nim
+++ b/compiler/backend/ccgstmts.nim
@@ -74,7 +74,6 @@ proc genVarTuple(p: BProc, n: PNode) =
   for i in 0..<n.len-2:
     let vn = n[i]
     let v = vn.sym
-    if sfCompileTime in v.flags: continue
     var traverseProc: Rope
     if sfGlobal in v.flags:
       assignGlobalVar(p, vn, "")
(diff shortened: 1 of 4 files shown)


saem (approved):
I don't see enough wins by working this change through properly, I don't doubt it'll improve a number of things, but those improvements will be forced anyways between the explicit revisit of these and/or further rework of semantic analysis for `const`, `let`, and `var`s.

So the temporary fix approach makes sense in the interim.

PR #673, opened 2023-04-29
internal refactor of hint/warn flag processing

saem:
## Summary

Hints and warnings can be enabled/disabled via `--hint[foo]:on` on the
CLI or in cfg files, this change separates the logic from legacy reports
in the internal implementation.

## Details

The `processSpecificNote` procedure within the `commands` module handles
this processing, introduced a revised version which returns a result
data type describing the outcome of processing. Then used a template to
bridge the result to legacy reports output. The core option processing
in `processSwitch` then uses the template making this change
transparent. Also fixed a minor issue where line info wasn't being
passed in when cfg files were involved.

### DIFF
--- a/compiler/front/commands.nim
+++ b/compiler/front/commands.nim
@@ -52,7 +52,8 @@ import
   compiler/utils/[
     nversion,
     pathutils,
-    platform
+    platform,
+    idioms,
   ]
 
 from compiler/front/in_options import MsgFormatKind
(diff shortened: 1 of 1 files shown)


compiler/front/commands.nim
```diff
@@ -214,7 +215,124 @@ proc expectNoArg(conf: ConfigRef; switch, arg: string, pass: TCmdLinePass, info:
...
+        else:
+          conf.flip(cnCurrent, n, isOn)
+          conf.flip(cnMainPackage, n, isOn)
+
+        if not isOn:
+          conf.excl(cnForeign, n)
+
+  result = ProcessNoteResult(kind: procNoteSuccess)
+
+proc deadprocessSpecificNote*(arg: string, state: TSpecialWord, pass: TCmdLinePass,
```
zerbina: Does the parent PR still use `deadprocessSpecificNote`, or can it be removed?

saem: crap... I thought I deleted that code!

saem: Thanks for spotting this, I'm going to push an updated removing this. 😅

PR #674, opened 2023-04-29
astgen: fix illegal C code being generated

zerbina:
## Summary

Fix the redundant-conversion elision not triggering due to `tySink`
not being skipped. The optimization is currently necessary in order to
work around a C code-generator issue where illegal C code is generated.

### DIFF
--- a/tests/lang_types/sink/tmove_from_lvalue_conversion.nim
+++ b/tests/lang_types/sink/tmove_from_lvalue_conversion.nim
@@ -1,0 +1,44 @@
+discard """
+  targets: "c !js !vm"
+  matrix: "--gc:orc"
+  description: '''
+    Destrutively moving from an up- or down converted `ref` value passed as a
+    `sink` parameter must work
+  '''
+"""
+
+# knownIssue: destructor-using refs (read, arc/orc support) are not yet
+#             implemented for the JS/VM target
+
+type
+  A = object of RootObj
+    val: int
+  B = object of A
+  C = object of B
+
+var instance = (ref C)(val: 1)
+
+template test(target: typedesc, explicit: static[bool]) =
+  block:
+    proc conv(x: sink(ref B)) =
+      when explicit:
+        let y = move(target(x))
+      else:
+        let y = target(x)
+
+      # force a destructive move by reassigning `x`
+      x = new(B)
+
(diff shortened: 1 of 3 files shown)


zerbina:
I first implemented the fix via preventing `sink` types from entering MIR code, and while that is a more general and better solution, it requires creating symbol nodes in the backend for which the node's type doesn't match the symbol's type, so I ultimately decided against that approach.

In the future, it might makes sense to prevent sink types from reaching into general semantic  analysis and elsewhere by omitting it for the parameter symbols and only keep it in procedure types. So for a procedure like `proc p(x: sink seq[int])`, the `x` parameter would have the type `tySeq[tyInt]` and `p` the type `tyProc[nil, tySink[tySeq[tyInt]]]`. This way, the places in the compiler that are not interested in `sink` (most of the compiler isn't) don't have to consider the type.

saem:
> In the future, it might makes sense to prevent sink types from reaching into general semantic analysis and elsewhere by omitting it for the parameter symbols and only keep it in procedure types. So for a procedure like proc p(x: sink seq[int]), the x parameter would have the type tySeq[tyInt] and p the type tyProc[nil, tySink[tySeq[tyInt]]]. This way, the places in the compiler that are not interested in sink (most of the compiler isn't) don't have to consider the type.

Maybe this is the difference between the parameter passing/calling contract (caller-side write) vs parameter usage/in-routine (in routine, read side of the parameter)?

I'm guessing the tricky bits will come up when considering:
- parameters that refer to past parameters
- checking if a call is equivalent to the same call but it's argument expressions are first assigned to locals

Anyhow, not a blocker for this change, just an interesting discussion topic.

tests/lang_types/sink/tmove_from_lvalue_conversion.nim
```diff
@@ -0,0 +1,44 @@
...
+        let y = move(target(x))
+      else:
+        let y = target(x)
+
+      # force a destructive move by reassigning `x`
+      x = new(B)
+
+      # test the value after moving the ref to make sure the move really
+      # worked at run-time
+      doAssert y.val == 1
```
saem, line 34: nice. :D

saem (approved):

PR #677, opened 2023-04-30
system: more correct `SysThread` bindings

zerbina:
## Summary

Change the `pthread_t` type binding (`SysThread`) to be treated as a
`distinct pointer` instead of an `object` for all targets besides 64-bit
Linux at the NimSkull side. The Posix standard doesn't require
`pthread_t` to be implemented as a pointer, but many implementations do
implement it as an opaque pointer.

## Details

Treating the type as an `object` led to the C code generator using
bracket initialization for initializing it in certain contexts, which is
incorrect if `pthread_t` is not implemented as a C struct and thus lead
to C compiler errors (MacOS was affected when using `--gc:orc`).

### DIFF
--- a/lib/system/threadlocalstorage.nim
+++ b/lib/system/threadlocalstorage.nim
@@ -88,14 +88,14 @@ else:
                     header: "<sys/types.h>".} = distinct cuint
   elif defined(openbsd) and defined(amd64):
     type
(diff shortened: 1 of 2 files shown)


saem (approved):
Pretty sure I've seen a few subtle bugs related to this very thing, nice.

PR #678, opened 2023-04-30
consider destructors of concrete types bound to concepts

zerbina:
## Summary

Turn the `hasDestructor` template into a procedure and skip wrapper-like
types (including `tyUserTypeClasses`) before testing for the
`tfHasAsgn`. This fixes:

* resolved concept types not being considered by cursor inference,
  destructor injection, or the move analyser
* `supportsCopyMem` reporting 'false' for concept types where the bound
  concrete type has a destructor

In addition, the JS backend failing with an internal compiler error for
variables using a resolve type-class as the type is also fixed.

## Details

Since the `tyUserTypeClasses` set is not included in `abstractInst`, the
logic part of the mid- and back-end needs to skip them manually, which
it more often than not does not. A `knownIssue` test is added for one
case where this leads to an issue.

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -557,7 +557,15 @@ when false:
       if n[i].containsNil: return true
 
 
-template hasDestructor*(t: PType): bool = tfHasAsgn in t.flags
+func hasDestructor*(t: PType): bool {.inline.} =
+  ## Returns whether the underlying concrete type of `t` has attached lifetime
+  ## tracking hooks (that is, is resource-like).
+  # skip all wrapper-like types that cannot have type-bound operations attached
+  # themeselves operations attached. Also skip user-type-classes here, as they
+  # should be resolved at this point
+  let t = t.skipTypes({tyGenericInst, tyOrdinal, tyAlias, tyInferred, tySink} +
+                       tyUserTypeClasses)
+  result = tfHasAsgn in t.flags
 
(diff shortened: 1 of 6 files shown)


compiler/ast/ast_query.nim
```diff
@@ -557,7 +557,15 @@ when false:
       if n[i].containsNil: return true
 
 
-template hasDestructor*(t: PType): bool = tfHasAsgn in t.flags
+func hasDestructor*(t: PType): bool {.inline.} =
```
saem, line 560: Separate from this PR/Tangent: I think it might be a good idea to think about moving queries that are only true after some amount of analysis to go into dedicated modules.

saem (approved):
Went from workarounds to bug fix. 🎉 

Time to strap in for `--gc:orc`!

zerbina:
An `nkObjConstr` node having a resolved user type-class as the type isn't expected by an assertion in `astgen` and also not by the C code generator (which crashes if such an object constructor reaches it).

The `var a: MatchAll = WithDestructor()` code I used in `ttypetraits.nim` triggered this issue, but since fixing it is out of this PR's scope, I changed the test code and added a `knownIssue`.

Eventually, wrapper types irrelevant for the mid- and back-end (e.g. `tyGenericInst`, `tyAlias`, `tyUserTypeClasses`, etc.) should be replaced prior with the underlying concrete types prior to entering the MIR (@saem mentioned this, and #424 had it implemented).

PR #679, opened 2023-04-30
move cli/cfg flag processing to `optionsprocessor`

saem:
## Summary

Created a new module `optionsprocessor` to take over handling flag
processing, command setting, and option testing. This functionality was
extracted from `commands`, which now handles CLI output.

## Details

This is part of a large scale refactor removing legacy reports from the
CLI aspects of the compiler. The key procedure, `processSwitch`, moving
resulted in a number of modules seeing subsequent updates due to changes
in imported files and that `processSwitch` now returns a result
value instead of using reports. Callers then convert these into
`CliEvent` types and output them all using the `commands` module.

In addition, some refactoring of `nimconf` was done to ensure path
events from `nimblecmd` were displayed if requested. Finally, both
`lexer` and `optionsprocessor` provide functions to generate core
diagnostic messages in a shift towards localizing message strings within
their owning module.

Finally, since `localReport` conflated messaging and control flow,
`nimconf` was adjusted to track errors and stop configuration
processing upon error to approximate this behaviour. Not all callers
want this, so `nimconf`'s primary API `loadConfigs` was adjusted in two
ways:
1. take a `stopOnError` param to indicate the desired behaviour
2. return true if processing was successful, otherwise false
This does a reasonable job simulating the old behaviour, and also moves
in the direction of treating the `ConfigRef` that `nimconf` is operating
on as not the canonical `ConfigRef`.

### DIFF
--- a/compiler/front/cmdlinehelper.nim
+++ b/compiler/front/cmdlinehelper.nim
@@ -16,7 +16,8 @@ import
   std/options as std_options,
   compiler/ast/[
     idents,
-    lineinfos
+    lineinfos,
+    lexer,
   ],
   compiler/modules/[
     modulegraphs
@@ -26,29 +27,24 @@ import
     commands,
     msgs,
     options,
+    optionsprocessor,
     condsyms,
     cli_reporter,
     sexp_reporter,
   ],
   compiler/utils/[
     pathutils,
-    idioms
   ],
   compiler/backend/[
     extccomp
   ]
 
+from experimental/colortext import ForegroundColor, toString
+
 from std/strutils import endsWith, `%`
 
 # xxx: reports are a code smell meaning data types are misplaced
-from compiler/ast/reports_lexer import LexerReport
-from compiler/ast/reports_parser import ParserReport
-from compiler/ast/reports_internal import InternalReport
-from compiler/ast/reports_external import ExternalReport
 from compiler/ast/report_enums import ReportKind
-from compiler/ast/reports import Report,
-  ReportCategory,
-  toReportLineInfo
 
 proc prependCurDir*(f: AbsoluteFile): AbsoluteFile =
   when defined(unix):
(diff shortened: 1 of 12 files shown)


compiler/front/nimconf.nim
```diff
@@ -30,101 +36,133 @@ import
...
+
   NimConfParser = object
     lexer: Lexer
     condStack: seq[bool]
     config: ConfigRef
-    cfgEvtHandler: NimConfEvtHandler
+    cfgEvtWriter: NimConfEvtWriter ## do not call, use `callEvtWriter`
+    stopOnError: bool              ## whether to continue if an error occurs
+    stopProcessing: bool           ## set if stopOnError and an error, or if
+                                   ##   there is a fatal error
```
zerbina:
```suggestion
    stopProcessing: bool           ## set if `stopOnError` and an error
                                   ## occurred, or if there is a fatal error
```
I think it's slightly easier to parse this way.

compiler/front/nimconf.nim
```diff
@@ -364,24 +423,35 @@ proc readConfigFile(N: var NimConfParser, filename: AbsoluteFile,
...
+    finally:
+      # restore to the previous source of command parameters
+      N.config.commandLineSrcIdx = oldCmdLineSrcIdx
 
-proc readConfigFile*(filename: AbsoluteFile, cache: IdentCache,
-                     config: ConfigRef, evtHandler: NimConfEvtHandler
+      closeLexer(N.lexer)
+
+proc readConfigFile(filename: AbsoluteFile, cache: IdentCache,
+                     config: ConfigRef, evtHandler: NimConfEvtWriter
```
zerbina: This procedure seems to not be used. Can it be removed?

compiler/front/nimconf.nim
```diff
@@ -30,101 +36,133 @@ import
...
 
-proc handleRead(N: NimConfParser,
+proc handleRead(N: var NimConfParser,
                 filename: string,
                 instLoc = instLoc(-1)) =
-    let e = ConfigFileEvent(kind: cekProgressConfStart,
-                            instLoc: instLoc,
-                            msg: filename)
-    N.cfgEvtHandler(N.config, e, instLoc)
+    if N.config.hasHint(rextConf):
```
zerbina: Given that `conf` might not be the canonical configuration, I think it might be better to leave this decision (the `hasHint` check) to the event writer.

compiler/front/nimconf.nim
```diff
@@ -346,16 +386,35 @@ proc parseAssignment(N: var NimConfParser, tok: var Token) =
...
+
+  case r.switch
+  of cmdSwitchNimblepath:
+    if N.config.hasHint(rextPath) and r.processedNimblePath.didProcess:
+      for res in r.processedNimblePath.nimblePathResult.addedPaths:
+        N.callEvtWriter(ConfigFileEvent(kind: cekProgressPathAdded,
+                                        msg: res.string,
+                                        location: info,
+                                        instLoc: instLoc()),
+                        instLoc())
```
zerbina: Maybe you have considered this already, but what are your thoughts about moving this to the event handler? That is, when switch processing was successful, dispatch a `cekSwitchProcessed` (or similar) event to the writer.

compiler/front/commands.nim
```diff
@@ -1412,6 +183,186 @@ proc processArgument*(pass: TCmdLinePass; p: OptParser;
...
+    of cliEvtErrFlagArgExpectedFromList:
+      "expected value for switch '$1'. Expected one of $2, but got nothing" %
+        [evt.givenFlg, allowedCliOptionsArgs(evt.flag).join(", ")]
+    of cliEvtErrFlagArgNotFromValidList:
+      "expected value for switch '$1'. Expected one of $2, but got nothing" %
+        [evt.givenFlg, allowedCliOptionsArgs(evt.flag).join(", ")]
+    of cliEvtErrFlagProcessing:
+      procResultToHumanStr(evt.procResult)
+    of cliEvtWarnings, cliEvtHints:
+      unreachable($evt.kind)
```
zerbina:
```suggestion
      unreachable(evt.kind)
```

compiler/front/commands.nim
```diff
@@ -1412,6 +183,186 @@ proc processArgument*(pass: TCmdLinePass; p: OptParser;
...
+      "execution of an external program '$1' failed with exit code '$2'" %
+        [evt.shellCmd, $evt.exitCode]
+    of cliEvtErrNoCliParamsProvided:
+      "no command-line parameters provided"
+    of cliEvtErrFlagArgForbidden:
+      "$1 expects no arguments, but '$2' found" %
+        [evt.givenFlg, evt.givenArg]
+    of cliEvtErrFlagArgExpectedFromList:
+      "expected value for switch '$1'. Expected one of $2, but got nothing" %
+        [evt.givenFlg, allowedCliOptionsArgs(evt.flag).join(", ")]
```
zerbina, line 320: If I'm reading the code correctly, this is going to cause a range defect for all flags except `cliFlagMsgFormat`.

compiler/front/commands.nim
```diff
@@ -1412,6 +183,186 @@ proc processArgument*(pass: TCmdLinePass; p: OptParser;
...
+      "added path: '$1'" % evt.pathAdded
+    of cliEvtErrors, cliEvtWarnings:
+      unreachable($evt.kind)
+
+  inc conf.hintCounter
+  if conf.verbosity > compVerbosityDefault:
+    conf.writeLog(msg, evt)
+
+proc cliEventLogger*(conf: ConfigRef, evt: CliEvent) =
+  ## a basic event logger that will write to standard err/out as apporpriate
```
zerbina:
```suggestion
  ## a basic event logger that will write to standard err/out as appropriate
```

compiler/front/commands.nim
```diff
@@ -1412,6 +183,186 @@ proc processArgument*(pass: TCmdLinePass; p: OptParser;
...
+  cliEvtWarnings = {cliEvtWarnSwitchValDeprecatedNoop}
+  cliEvtHints    = {cliEvtHintPathAdded}
+
+static:
+  const unaccountedForEvtKinds =
+    cliLogAllKinds - cliEvtErrors - cliEvtWarnings - cliEvtHints
+  doAssert unaccountedForEvtKinds == {}, "Uncategorized event kinds: " &
+                                            $unaccountedForEvtKinds
+
+proc procSwitchResultToEvents*(conf: ConfigRef, pass: TCmdLinePass,
```
zerbina: Given its current callsites, I think it might make sense to turn this into an iterator.

compiler/front/optionsprocessor.nim
```diff
@@ -0,0 +1,1698 @@
+## this modules process various options given as a flags and applies them to
+## a given `ConfigRef`. This allows for command line, configuration, comptime,
+## etc configuration building and updates through a unified interface
```
zerbina:
```suggestion
## this module processes various options given as flags and applies them to
## a given `ConfigRef`. This allows for command line, configuration, compile-time,
## etc configuration building and updates through a unified interface
```
'comptime' -> 'compile-time' is only a soft suggestion -- I think it makes sense to write it out in documentation.

compiler/front/optionsprocessor.nim
```diff
@@ -0,0 +1,1836 @@
...
+  of cmdSwitchApp         : @["gui", "console", "lib", "staticlib"]
+  of cmdSwitchOs          : platform.listOSnames()
+  of cmdSwitchCpu         : platform.listCPUnames()
+  of cmdSwitchVerbosity   : @["0", "1", "2", "3"]
+  of cmdSwitchIncremental : @["on", "off", "writeonly", "readonly", "v2", "stress"]
+  of cmdSwitchCc          : listCCnames()
+  of cmdSwitchFilenames   : @["abs", "canonical", "legacyRelProj"]
+  of cmdSwitchProcessing  : @["dots", "filenames", "off"]
+  of cmdSwitchExperimental: experimentalFeatures.toSeq.mapIt($it)
+  of cmdSwitchExceptions  : @["native", "setjmp", "quirky", "goto"]
```
zerbina:
```suggestion
  of cmdSwitchExceptions  : @["native", "goto"]
```

compiler/front/optionsprocessor.nim
```diff
@@ -0,0 +1,1836 @@
...
+    case arg.normalize
+    of "", "on": conf.incl op
+    of "off": conf.excl op
+    else:
+      result = ProcSwitchResult(kind: procSwitchErrArgExpectedOnOrOff,
+                                switch: result.switch,
+                                givenSwitch: s,
+                                givenArg: arg,
+                                srcCodeOrigin: instLoc())
+      return
```
zerbina:
```suggestion
    {.line.}:
      case switchOn(s, arg)
      of true:  conf.incl op
      of false: conf.excl op
```
Soft suggestion. It slightly reduces code duplication.

compiler/front/optionsprocessor.nim
```diff
@@ -0,0 +1,1836 @@
...
```
zerbina:
```suggestion
    {.line.}:
      case switchOn(s, arg)
      of true:  conf.incl op
      of false: conf.excl op
```
Same here.

zerbina (commented):
Finished the first pass of my review. I've mostly looked into `nimconf` so far; still need to take a closer look at `optionsprocessor`.

The direction with moving decision making out of the immediate option processing by returning a result is good, I think, and having the formatting logic in the event's owning module is nice.

compiler/front/commands.nim
saem: Yeah, not well documented, but that's the only thing that can raise that particular error, I should pick a lane general implementation useable by all or specific implementation tied to the single use case.

saem: Creating a dedicated range and putting that into the variant results in an NPE in the compiler, I'll trouble shoot that later/separately.

compiler/front/nimconf.nim
saem: I've moved the `hasHint` out, but I'm less certain about moving the rest of the logic. The module should process the config file: parsing, figuring out what the instructions are, then executing them against the given `ConfigRef`.

My thinking is that the caller wants to know that paths were added, but doesn't want to have to bother digging that information out or be sensitive to the changes in data type definition of `optionsprocessor.ProcSwitchResult`.

compiler/front/optionsprocessor.nim
saem: I'm aiming to make `comptime` part of the regular vernacular.

compiler/front/optionsprocessor.nim
saem: Thanks, I thought I had gotten the last of these.

compiler/front/nimconf.nim
saem: Removed locally.

compiler/front/nimconf.nim
saem: updated, thanks.

compiler/front/nimconf.nim
saem: It's the struggle between direct/straightforward code vs more indirection/callback. Longer term, heading towards iterators/coroutines/continuations the caller should react to these things, which is currently best represented by the callback.

I'll move this and `rextPath` into the callback.

compiler/front/commands.nim
saem: I forgot about the little optimization the template did; updated, thanks.

compiler/front/optionsprocessor.nim
saem: I like it!

compiler/front/optionsprocessor.nim
saem: Did this one too.

compiler/front/commands.nim
saem: Thanks, updated.

compiler/front/commands.nim
saem: Imma do this

saem: Done.

saem:
If this is good to go, I can push the magics one shortly thereafter.

In the meantime/following, I'm going side-quest to see if I can fix an issue or two I encountered in `sem` while working on these PRs.

compiler/front/cmdlinehelper.nim
```diff
@@ -194,22 +239,25 @@ proc loadConfigsAndProcessCmdLine*(self: NimProg, cache: IdentCache; conf: Confi
...
+    self.processCmdLine(passCmd2, "", conf)
+    if conf.cmd == cmdNone:
+      conf.logError(CliEvent(kind: cliEvtErrCmdMissing,
+                              srcCodeOrigin: instLoc(),
+                              pass: passCmd2))
+
+    graph.suggestMode = self.suggestMode
+    return true
+  else:
+    return false
```
zerbina:
```suggestion
    result = true
  else:
    result = false
```
Very soft suggestion -- feel free to disregard it.

compiler/front/commands.nim
```diff
@@ -1506,6 +455,16 @@ proc processCmdLine*(pass: TCmdLinePass, cmd: string; config: ConfigRef) =
...
 
+  template expectNoArg(f: CliFlagKind, flg, arg: string) =
+    if arg != "":
+      config.cliEventLogger:
+        CliEvent(kind: cliEvtErrFlagArgForbidden,
+                flag: f,
+                givenFlg: flg,
+                givenArg: arg,
+                pass: pass,
+                srcCodeOrigin: instLoc())
```
zerbina, line 467:
```suggestion
                srcCodeOrigin: instLoc())
      continue
```
A temporary solution in order to not print help, version, etc. messages when an argument is provided (the previous behaviour).

compiler/front/optionsprocessor.nim
```diff
@@ -0,0 +1,1824 @@
...
+      result = ProcSwitchResult(kind: procSwitchErrArgForbidden,
+                                switch: result.switch,
+                                givenSwitch: s,
+                                givenArg: arg,
+                                srcCodeOrigin: instLoc())
+      return
+
+  template expectArgValue(cond: bool, arg, s: string, list: seq[string]) =
+    if cond:
+      # TODO: deal with list
```
zerbina: Is this todo resolved as part of the main PR?

compiler/front/optionsprocessor.nim
```diff
@@ -0,0 +1,1824 @@
...
+      cmdSwitchDeepcopy           : {fullSwitchTxtDeepcopy},
+      cmdSwitchProjStdin          : {smolSwitchTxtProjStdin},
+      cmdSwitchCmdexitgcstats     : {fullSwitchTxtCmdexitgcstats},
+      cmdSwitchConfigVar          : {fullSwitchTxtConfigVar},
+    ]
+
+    cmdTxtToSwitch: array[CmdSwitchTextKind, CmdSwitchKind] =
+      block:
+        # for each CmdSwitchKind we might have one or two text flags, iterate
+        # through each switch and add one or two text -> switch entry
```
zerbina:
```suggestion
        # through each switch and add one or two text -> switch entries
```

compiler/front/commands.nim
```diff
@@ -1412,6 +183,184 @@ proc processArgument*(pass: TCmdLinePass; p: OptParser;
...
+    # warnings - general flags/options/switches
+    cliEvtWarnSwitchValDeprecatedNoop
+    # hints - general flag/options/switches and/or processing
+    cliEvtHintPathAdded
+      ## currently only triggered if nimble adds a path
+      # xxx: this doesn't feel like a hint, more like "info" or "trace"
+
+  CliEvent* = object
+    # xxx: should these be 'flat' log events? put another way is the 
+    #      `procResult` field a design smell?
```
zerbina: I personally think the nested approach (i.e., storing `ProcSwithResult` here) is fine. The subset-of relationship between switch processing error and CLI errors is made clear, and it also allows for re-using the `ProcSwitchResult` rendering facilities.

zerbina (approved):
I finished the review, and like the change. Further changes to option processing become easier, and the responsibilities clearer. The suggestions I left are non-blocking.

Event messages logged to the output (via `logError`, `logWarning`, and `logHint`) no longer have a colored prefix, but I think it's okay to restore this as a follow up.

compiler/front/commands.nim
saem: Yup, I figure that the current behaviour isn't right and not conventional wrt CLI in the large, where such syntax errors typically result in usage being printed, or a message alluding to it.

compiler/front/cmdlinehelper.nim
saem: Thanks, done.

compiler/front/commands.nim
saem: Thanks, I rewrote the comment based upon yours.

compiler/front/optionsprocessor.nim
saem: It wasn't, I had to drop the param and the argument from all the callsites, which I just did locally.

saem:
/merge

everything should be good to go, going to queue for merge.

saem:
/merge

maybe one day I'll learn to more consistently tests after a rebase -- thank goodness for CI.

PR #680, opened 2023-05-01
internal: clarify type-bound operator attachment

zerbina:
## Summary

Move to only attaching type-bound operators to unwrapped types, by
skipping to the underlying concrete type before calling
`createTypeBoundOps` in `sempass2`. This is small step towards cleaning
up the type-bound operator implementation.

In addition, this also fixes a regression caused by `hasDestructor` now
skipping the wrapper-like types, which caused alias types and generic
instantiations to not be detected as having a destructor attached.

## Details

* move the skip set used in `hasDestructor` to the `skipForHooks`
  constant
* skip types from the `skipForHooks` set before querying or creating the
  type-bound operators for a type in `sempass2`
* skip types from the `skipForHooks` set before querying type-bound
  operators in `injectdestructors`
* don't attempt to rewrite explicit type-bound operators call during
  `sem` -- that's the responsibility of the `sempass2`
* use `getOp` for querying the attached sink operator in
  `injectdestructors` (it should have been used there from the start)

Calls to `createTypeBoundOps` from `liftdestructors` still attach the
operators to the original (read, unskipped) type.

### The fixed problem

Both closure types (`tyProc` with the `ccClosure` calling convention)
and compound types where none of the element/field types has an
overridden type-bound operation at the time when the compound type is
defined don't have the `tfHasAsgn` flag included at definition time.

`createTypeBoundOps` only includes the flag on the type that was passed
to it, not the skipped one, meaning that if `createTypeBoundOps` is
never invoked directly with the skipped type, the skipped type doesn't
get the `tfHasAsgn` flag included, resulting in it being effectively
treated as having no attached type operation.

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -560,12 +560,7 @@ when false:
 func hasDestructor*(t: PType): bool {.inline.} =
   ## Returns whether the underlying concrete type of `t` has attached lifetime
   ## tracking hooks (that is, is resource-like).
-  # skip all wrapper-like types that cannot have type-bound operations attached
-  # themeselves operations attached. Also skip user-type-classes here, as they
-  # should be resolved at this point
-  let t = t.skipTypes({tyGenericInst, tyOrdinal, tyAlias, tyInferred, tySink} +
-                       tyUserTypeClasses)
-  result = tfHasAsgn in t.flags
+  result = tfHasAsgn in t.skipTypes(skipForHooks).flags
 
 template incompleteType*(t: PType): bool =
   t.sym != nil and {sfForward, sfNoForward} * t.sym.flags == {sfForward}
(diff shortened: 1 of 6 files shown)


saem (approved):
🤞🏽 hopefully this is the last blocker.

zerbina:
There were more places that expected the type-bound operators to be attached to the wrapper types, hence the test failures. I adjusted them, and also removed the `mDestroy`/`mTrace` patching from `magicsAfterOverloadResolution`.

The correct type-bound operator might not exist at that point yet, and if it doesn't, an uninstantiated generic symbol is sometimes inserted for types where the respective operator is generic. Until it's clear how an explicit call to a `=destroy` or `=trace` operator should work in the case where only the generic system overload matches, not patching the call during the first pass of semantic analysis seems like the more sensible thing to do.

PR #681, opened 2023-05-02
magics: move `compileroption` to `magicsafteroverloadresolution`

saem:
## Summary

Moved `compileOption` and `compileOptionArg` magics handling from the
`semfold` module to `semmagic`, specifically into the procedure
`magicsAfterOverloadResolution`. This avoids yet another place where
`semfold` can generate an `nkError` that would pollute `transf`. This
also fixes a bug where the compiler would crash if an compile option was
tested.

## Details

Beyond the procedure move, the implementation now uses
`sem.evalConstExpr` instead of `semfold.getConstExpr`, this not only
cleans up the call, but more accurately represents what should be
happening. What's happening: semantic analysis replaces at compile time
calls to `system.compileOption` with their value. Meaning that these
calls never reach runtime (only the results), but it "feels" like it.

Also added some tests to ensure the regression of the compiler crash
when providing invalid arguments doesn't happen again. As part of this
testing a small bug was found in `semDiscard` where the operand for a
discard statement wasn't checked for `nkError` (this has also been
fixed) and triggered an assert in `transf`.

Finally some unused imports were opportunistically removed in order to
quiet some build warnings.

### DIFF
--- a/compiler/front/scripting.nim
+++ b/compiler/front/scripting.nim
@@ -36,7 +36,6 @@ import
     vmops
   ],
   compiler/front/[
-    msgs,
     condsyms,
     options,
   ],
--- a/compiler/sem/semfold.nim
+++ b/compiler/sem/semfold.nim
@@ -31,7 +31,6 @@ import
   ],
   compiler/front/[
     options,
-    optionsprocessor,
   ],
   compiler/utils/[
     platform,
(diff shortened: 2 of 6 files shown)


zerbina (commented):
Nice, thank you. Since the logic is now located in `sem`, all compile-time compute-able expressions could be supported as the arguments (as long as they yield a string) by using `evalConstExpr` instead of `getConstExpr`. It's nothing that has to happen as part of this PR, however.

saem:
@zerbina FYI, I made the transition to `evalConstExpr`

PR #682, opened 2023-05-04
static `try` void-expression no longer crashes

saem:
## Summary

Fix compiler crash for `try` void expressions in `static` blocks;
moreover, this fixes a crash for any AST that produced bytecode but
didn't allocate any registers.

## Details

Static blocks presently are void-expressions (statements), `vmgen` and
the `compilerbridge` were not handling these correctly in all cases.
When `vmgen` generated bytecode that didn't allocate any registers the
`compilerbridge` would crash attempting to access a non-existent local
for a return value. One way to trigger this scenario is with some void-
expressions (see added test).

In particular it uses `opcRet` for expression evaluation, where a return
value is required from VM code execution, and `opcEof` when one is not
required. In the case of the former the present behaviour is largely
maintained except we now return the index of local storage in the main
VM stack-frame; and for the latter we don't return any index signaling
to the consumer that no value as a `TFullRegister` with kind `rkNone` is
to be provided.

In debugging this change, an issue with VM code gen listings was found
and resolved, as well.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -528,7 +528,7 @@ type
       ## instantiation and prior to this it has the potential to
       ## be any type.
 
-    tyVoid ## now different from tyEmpty, hurray!
+    tyVoid ## void type, lack of a value or unit
 
 static:
   # remind us when TTypeKind stops to fit in a single 64-bit word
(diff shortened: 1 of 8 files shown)


zerbina (commented):
Looks to good to me, thanks!

A small request: could you adjust the PR message to indicate that the problem affected all AST that produced bytecode but without allocating any registers? The reason that `static: discard` worked and works without error is that no bytecode is generated for static's body, which results in `evalConstExprAux` returning early.

An empty `try` statement does translate to bytecode, but allocates no registers by default, causing the frame to have an empty slot and thus triggering the out-of-bounds access.

saem:
/merge

updated message

PR #683, opened 2023-05-04
IC: fix the code generator crashing

zerbina:
## Summary

Always allocate space for all modules in the backend module list before
starting code generation, regardless of what parts of the list are
filled. This fixes the C code generator crashing with an `IndexDefect`
in certain situation where not all modules were recompiled.

## Details

If a module is cached there's no change in its set of alive symbols, no
backend module list entry is created for it. No entry being present is
not a problem for the code generator, but the module list still needs to
have an empty slot reserved for the entry.


zerbina:
@saem I believe this fixes the crash you encountered with the removal of NimScript configuration. If wanted, I can re-add the tests in order to improve test coverage for IC a bit.

saem (approved):
Nice, less clever code is better.

Yup, I read the PR description and quickly realized it's most likely the fix for issue I encountered earlier.

PR #684, opened 2023-05-04
IC: fix an internal loading bug

zerbina:
## Summary

Fix a severe internal bug with type and symbol loading. In the common
case it had no effect, but in many other cases it would either cause an
internal compiler error or, in the worst case, would cause the
wrong symbol or type being silently loaded.

## Details

Apart from the packed item ID, items that support on-demand loading
(symbols and types) also store an additional module ID. This ID
identifies the module as part of which the packed item ID was originally
encoded, and is required to correctly decode the packed ID.

There were two problems. First, the loader logic assumed that the
decoding module is the same as the module to which the to-be-loaded
symbol belongs, and tried to look up the item in the decoding module's
cache. If this resulted in a cache hit, a fully unrelated symbol or
type was returned instead of the requested one, and subsequently
persisted!

The second problem was that the logic for replaying various
`ModuleGraph` state changes (`replayGenericCacheInformation`) treated
foreign types and symbols as being encoded as part of the attached-to
module instead of the one to which the state change was recorded (which
is the module that the IDs were encoded with).

A concrete situation where this led to issues is the following: module
`A` imports module `B`, and both import `C`. `B` records an
procedure instantiation cache entry that uses type `T`, which is
attached to `C`. `A` queries the instantiation cache.

When both `A` and `B` are recompiled, everything works as expected.
When only `A` is recompiled, however, and the compiler resolves (i.e.
loads) the deferred types part of the instantiation cache entry, the ID
of `T` is decoded with the context of module `C`, which is incorrect --
it needs to be `B`, as `B` recorded the instantiation (and encoded the
packed ID).

### DIFF
--- a/compiler/ic/ic.nim
+++ b/compiler/ic/ic.nim
@@ -1071,32 +1071,24 @@ proc loadProcBody*(config: ConfigRef, cache: IdentCache;
 
 proc loadTypeFromId*(config: ConfigRef, cache: IdentCache;
                      g: var PackedModuleGraph; module: int; id: PackedItemId): PType =
-  if id.item < g[module].types.len:
+  if id.module == LitId(0) and g[module].typesInit:
+    # it's a type from `module`, but it doesn't have to be cached yet
     result = g[module].types[id.item]
-  else:
-    result = nil
+
   if result == nil:
-    var decoder = PackedDecoder(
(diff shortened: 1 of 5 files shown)


saem:
Thanks for the clear write-up and accompanying tests, it made it easy to follow the change.

PR #685, opened 2023-05-04
IC: support type-bound operators and `--gc:orc`

zerbina:
## Summary

Finish the implementation for recording type-bound operator into packed
modules, and adjust the IC backend and its DCE implementation to support
both lifetime hooks and parts of the new runtime.

This allows for simple programs to be compiled with `--incremental:on`
and `--gc:orc`, and is another step towards removing the old runtime.

## Details

Getting type-bound operators and `--gc:orc` supported consists of four
parts.

### Packed Modules

Attaching either a user-defined or generated type-bound operator to a
type is a change to the `ModuleGraph`, and thus has to be recorded in
order to allow for later replay of the state changes. The basic
facilities already existed, but storing and replaying the `attachedOps`
changes was missing.

* implement recording the attachment changes
* implement replaying the attachment changes
* split out replay of state relevant to the backend into a separate
  procedure
* fix the "trivial attached operator" check not taking on-demand
  loading of procedure bodies into account

### Backend setup

Restore backend-relevant `ModuleGraph` state discarded by
`resetForBackend` before running alive analysis -- no information
about attached operator would be present otherwise.

### Dead-code elimination

Compared to the other backends, the IC implementation uses a dedicated
alive analysis pass over the whole packed module-graph in order to
figure out the set of alive routines. Only routines present in this set
have C code generated for them -- the others are ignored.

The `injectdestructors` pass, which is responsible for injecting copy,
sink, and destroy hooks, is only run for all alive code reaching the
code generators, but at that point, the previous alive analysis would
have marked all not explicitly used hooks as dead code.

In order to get around this phase-ordering problem, the alive analysis
now conservatively marks all lifetime hooks for types used by alive:
* locals
* globals
* constants
* parameters

as being in the alive set. The same goes for lifetime hooks attached to
types used as the return type of alive routines.

The downside of this conservative approach is, that hooks ending up
not being actually used are still emitted and compiled, leading to
larger binaries.

### Code generator

Include the magics used by the seqsv2 implementation in the set of
magics to consider for code generation. Said magics are implemented
as normal (generic) procedures, and thus their bodies also need to
be compiled.

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -2945,7 +2945,7 @@ proc expr(p: BProc, n: PNode, d: var TLoc) =
     if n[genericParamsPos].kind == nkEmpty:
       var prc = n[namePos].sym
       if useAliveDataFromDce in p.module.flags:
-        if p.module.alive.contains(prc.itemId.item) and prc.magic in {mNone, mIsolate}:
+        if p.module.alive.contains(prc.itemId.item) and prc.magic in NonMagics:
           genProc(p.module, prc)
       elif prc.skipGenericOwner.kind == skModule and sfCompileTime notin prc.flags:
         if ({sfExportc, sfCompilerProc} * prc.flags == {sfExportc}) or
--- a/compiler/backend/cgen.nim
+++ b/compiler/backend/cgen.nim
@@ -103,6 +103,10 @@ when not declared(dynlib.libCandidates):
 when options.hasTinyCBackend:
   import backend/tccgen
 
+const NonMagics* = {mNone, mIsolate, mNewSeq, mSetLengthSeq, mAppendSeqElem}
+  ## magics that are treated like normal procedures by the code generator.
+  ## This set only applies when using the new runtime.
+
 proc addForwardedProc(m: BModule, prc: PSym) =
   m.g.forwardedProcs.add(prc)
 
--- a/compiler/ic/cbackend.nim
+++ b/compiler/ic/cbackend.nim
@@ -45,6 +45,7 @@ import
     packed_ast,
     ic,
     dce,
+    replayer,
     rodfiles
   ]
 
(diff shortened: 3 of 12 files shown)


zerbina:
The implementation as it is now has two problems.

### Indirect dependencies

A module can attach to a type a symbol that is not reachable through any of its import dependencies. If the module the symbol is part of is invalidated (i.e., it is recompiled), but the module that recorded the attachment is not, the unresolved symbol stored as part of the attachment entry will become stale, and attempting to resolve it will result in the compiler either crashing or producing incorrect code.

Example:
```nim
# --- a.nim
var x: seq[(int, int)]
# a destructor is lifted for `seq[(int, int)]`. If this is the first occurrence of
# `seq[(int, int)]` as part of the current compilation, this instance becomes the
# canonical one.

# --- b.nim
var z*: seq[(int, int)]
# if this module is compiled after `b`, no new destructor will be lifted. Instead, the
# one attached to the canonical type is also attached to the instance of `seq[(int, int)]`.
# `b` now implicitly depends on `a`.

# --- c.nim
import a, b

discard z
```

If `a` is invalidated while `b` is not, it will cause problems. A quick fix is to record the implicit dependency in the packed module's `import` sequence. I had this implemented at one point, but, given the effectively non-existent test coverage for IC, decided against including it here.

### Late-attached type bound operators

The types for closure environments are generated as part of `transf`, which phase-wise happens after semantic analysis and as part of the backend processing -- at this point, the packed modules must not be modified anymore. Lambda-lifting does so nonetheless (by creating type-bound operators for the generated types), which breaks IC and causes the compiler to either crash, or fail with a linker error due to missing definitions.

**Edit:** the backend *is* allowed to (and does) modify not-yet-written-to-disk packed modules. However, these kind of changes must only be made to specific backend-only state (e.g. `emittedTypeInfo`), which the state potentially modified by `createTypeBoundOps` is **not** part of.

---

Despite their shortcomings listed above, the additions implemented by this PR do improve on the situation, and I think merging them is fine.

saem:
Read through the change as it stands currently and the concerns above. I agree; the king and short of it is that it just needs to work with so we can move to `--gc:orc` and eventually this stuff will go away.

saem (approved):
Looks good, very little changed since my last read. :D

PR #686, opened 2023-05-05
sem: process `mEnumToStr` earlier

zerbina:
## Summary

Mark compiler-generated enum-to-string procedures as being `mEnumToStr`
instances, and replace the generic `mEnumToStr` magic symbols with the
compiler-generated ones during semantic analysis already.

This makes it easier for code generators to opt-in to using the
compiler-generated procedure instead of implementing it via custom
logic. As an intentional side-effect of this change, the default
`$` operator for enums can now be used when both the IC backend and
new runtime are enabled.

### DIFF
--- a/compiler/sem/semmagic.nim
+++ b/compiler/sem/semmagic.nim
@@ -401,6 +401,14 @@ proc magicsAfterOverloadResolution(c: PContext, n: PNode,
       result = semAsgnOpr(c, n)
     else:
       result = semShallowCopy(c, n, flags)
+  of mEnumToStr:
+    # overload resolution picked the generic enum-to-string magic.
+    # Replace the symbol with that of the auto-generated procedure.
+    let
+      info = n[0].info
+      prc = getToStringProc(c.graph, n[1].typ.skipTypes(abstractRange))
+    result = n
+    result[0] = newSymNode(prc, info)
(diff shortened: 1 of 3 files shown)


compiler/sem/semmagic.nim
```diff
@@ -401,6 +401,15 @@ proc magicsAfterOverloadResolution(c: PContext, n: PNode,
...
       result = semShallowCopy(c, n, flags)
+  of mEnumToStr:
+    # overload resolution picked the generic enum-to-string magic.
+    # Replace the symbol with that of the auto-generated procedure.
+    let
+      info = n[0].info
+      prc = getToStringProc(c.graph, n[1].typ.skipTypes(abstractRange))
+    result = n
+    result[0] = newSymNode(prc)
+    result[0].info = info
```
saem:
```suggestion
    result[0] = newSymNode(prc, info)
```

Soft suggesiotn, but it saves a a line. :D

saem:
This is a good simplification, and I wouldn't be surprised if we can use this approach in a few other places.

compiler/sem/semmagic.nim
zerbina: D'oh, I forgot about this overload existing. Thanks for the reminder :D

saem:
/merge

Let's go!

PR #687, opened 2023-05-05
front: centralize backend-specific customization

zerbina:
## Summary

Start moving backend-specific customization into a single procedure that
is called prior to executing a command (now including `nimsuggest`). In
addition, fix `nim e` configuring the compiler as if the C backend was
used -- defines like `c` are now no longer present when running
NimScript.

This is another step towards cleaning up the front-end, and fixes the
following issues:
- the backend defines (`c`, `js`, `vm`) not being present in
  `nimsuggest` mode
- no GC being selected by default in `nimsuggest` mode
- the incorrect exception mode being selected when using `nimsuggest` or
  `check`

## Details

* move `customizeForBackend` into a standalone procedure in currently
  located in `main` (until a better home for it is found) and call it
  before running `nimsuggest`'s `mainCommand`
* move the exception mode configuration for the JS and VM backend into
  `customizeForBackend`
* don't call `customizeForBackend` prior to executing a file in
  NimScript mode
* fix the GC staying unselected when using the JS or VM backend
* remove the now obsolete `selectDefaultGC` procedure

Due a side-effect of using the new runtime, the `tsigmatch.nim` test
fails failed because of an issue with type aliases in procedure
instantiations. A `knownIssue` test is added.

### DIFF
--- a/compiler/front/main.nim
+++ b/compiler/front/main.nim
@@ -73,6 +73,8 @@ from compiler/ast/report_enums import ReportKind,
   repWarningKinds,
   rstWarnings
 
+from compiler/front/optionsprocessor import processSwitch,
+  procSwitchSuccess
 from compiler/front/scripting import runNimScript
 
 when not defined(leanCompiler):
@@ -226,7 +228,6 @@ proc commandCompileToJS(graph: ModuleGraph) =
       kind: rintUsingLeanCompiler,
       msg: "Compiler was not build with js support"))
   else:
-    conf.exc = excNative
     conf.target =
       block:
         var t = conf.target
@@ -243,9 +244,6 @@ proc commandCompileToJS(graph: ModuleGraph) =
       writeDepsFile(graph)
 
 proc commandCompileToVM(graph: ModuleGraph) =
-  let conf = graph.config
-  conf.exc = excNative
-
   semanticPasses(graph)
   registerPass(graph, vmgenPass)
   compileProject(graph)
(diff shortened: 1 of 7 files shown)


saem:
Dropping the weird dependency from `nimsuggest` to `main` might just have to wait until we have a first class `nim lsp` subcommand. So overall I think things are heading in the right direction.

compiler/front/main.nim
```diff
@@ -428,6 +426,37 @@ proc `$`(params: UsedBuildParams): string =
...
+  case conf.backend
+  of backendC:
+    case conf.exc
+    of excNone, excNative: conf.exc = excGoto
+    of excGoto:            discard
+
+    if conf.selectedGC == gcUnselected:
+      # the default gc for the C backend is ORC. We can't just set it to
+      # ``gcOrc`` directly, however, as additional defines, etc. are required
+      # XXX: the dependency on ``optionsprocessor`` hints that a different
```
saem, line 447: I've held off on separating flag parsing vs flag "executing", the latter meaning actually acting on some strictly defined data describing the parsed out flag.

If those were separated then there could be a direct call to the executing part, skipping past this silly string business.

saem (approved):
Thanks, glad to see this is starting to unlock yet more cleanup.

zerbina:
The reason the `tsigmatch.nim` is now failing is only very indirectly related to the changes here. No default GC was previously selected for `nim check` but now is, triggering an issue that's also not really related to `--gc:orc` vs. `--gc:arc`.

The underlying problem is the following:
```nim
proc test[T](): T =
  discard

type Alias = string

echo $typeof(test[Alias]())
echo $typeof(test[string]()) # will also echo "Alias"
```

The `Alias` type in this instance is a `tyString` instance with the symbol "Alias" attached to it. When `test` is instantiated and added to the instance cache, the `Alias` instance is the type it is recorded with.

That's not a problem for cache lookup, as the symbol on a type doesn't affect equality between two instance, but it is problem for overload resolution, since the return type of the instantiation is used for the call expression type.

I'm inclined to move the failing part into a `knownIssue` test and fix the problem later.

saem:
`knownIssue` is fine.

zerbina:
/merge

I added a `knownIssue` test for the issue.

PR #688, opened 2023-05-05
tester: escape strings in EPC mode

zerbina:
## Summary

Escape the path strings part of the S-expression messages generated by
the `nimsuggest` tester. Because of the backslashes used by paths on
Windows, the strings not being escaped properly led to all EPC tests
failing there.

## Details

Since the strings are part of an S-expression, `escapeJson` is used
instead of `escape`.


saem (approved):
/merge

Funny the things that get missed

PR #689, opened 2023-05-05
reports: remove most CLI/cfg related legacy

saem:
## Summary

With the introduction of `commands.CliEvent` and compiler messages a
number of legacy reports related to thses are no longer required. This
change removes a large majority of these.

## Details

Removes a number of report kind enums, most of these are external
reports relating to command, flags, and arguments. This also removes a
few incorrectly categorized lexer reports, which relate to cfg
processing.

This change is entirely internal, and the messages are still produced as
before.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -2686,8 +2686,6 @@ To create a stacktrace, rerun compilation with './koch temp $1 <file>'
     of rintMsgOrigin, rintErrKind:
       assert false, "is a configuration hint, should not be reported manually"
 
-    of rintNimconfWrite:
-      result = r.msg
 
 proc reportFull*(conf: ConfigRef, r: InternalReport): string =
   assertKind r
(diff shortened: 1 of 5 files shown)


compiler/front/cli_reporter.nim
```diff
@@ -2807,89 +2799,13 @@ proc reportBody*(conf: ConfigRef, r: ExternalReport): string =
   assertKind r
   case ExternalReportKind(r.kind):
     of rextConf:
-      result.add("used config file '$1'" % r.msg, conf.suffix(r))
+      doAssert false, "this should never be reported"
```
```suggestion
      unreachable("this should never be reported")
```

compiler/front/cli_reporter.nim
```diff
@@ -2807,89 +2799,13 @@ proc reportBody*(conf: ConfigRef, r: ExternalReport): string =
...
-
-    of rextInvalidPackageName:
-      result = "invalid package name: " & r.packageName
-
-    of rextDeprecated:
-      result = r.msg
-
     of rextPath:
-      result = "added path: '$1'" % r.packagePath
+      doAssert false, "this should never be reported"
```
```suggestion
      unreachable("this should never be reported")
```

zerbina (approved):
Nice, one step closer to removing legacy report. I left two soft suggestions regarding `doAssert false` -> `unreachable`, but those are not blockers.

saem:
@zerbina thanks for the review and suggestions (I updated the code).

PR #690, opened 2023-05-06
cgen: unify field defect raising

zerbina:
## Summary

Simplify and streamline the logic for emitting the raising of field
defects in `ccgexprs`, which results in the message now properly
rendering enum and boolean discriminator values when using `--gc:orc`.

## Details

* use a common set of `raiseFieldDefectX` runtime procedures, and mark
  the old ones as obsolete
* `reprDiscriminant` is now no longer used for the C target
* when using the runtime, render enum discriminator values with the
  compiler-generated enum-to-string procedure
* remove the backwards-compatibility logic for supporting legacy stdlib
  versions

With the IC backend, using the compiler-generated enum-to-string for
rendering the discriminator value would require `dce` to analyzing
`nkCheckedFieldExpr` nodes in a similar fashion to the `nkChckRange`
analysis, but this is skipped for now. When both the IC backend and
the new runtime are used, enum discriminator values continue to be
rendered as integers.


saem (approved):
Thank you for doing this, it makes debugging these so much nicer.

PR #691, opened 2023-05-07
initial cut of the change

saem:
pretty sure there is a bug in `semtempl`, maybe `semgnrc`

## Summary
* what changed and how?
* why are we changing it?

## Details
* anything else

Fixes full_issue_url

---

## Notes for Reviewers
* first cut of the change
* need to ferret out the bugs
* finish updating `macros` API
* add in any quick wins

### DIFF
--- a/compiler/sem/semfields.nim
+++ b/compiler/sem/semfields.nim
@@ -50,6 +50,8 @@ proc instFieldLoopBody(c: TFieldInstCtx, n: PNode, forLoop: PNode): PNode =
           result.add(tupl)
           result.add(newSymNode(c.field, n.info))
         break
+  of nkSymChoices:
+    result = instFieldLoopBody(c, newSymNode(n.choices[0], n.info), forLoop)
   else:
     if n.kind == nkContinueStmt:
       localReport(c.c.config, n, reportSem rsemFieldsIteratorCannotContinue)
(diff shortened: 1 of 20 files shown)


compiler/sem/semfields.nim
```diff
@@ -50,6 +50,8 @@ proc instFieldLoopBody(c: TFieldInstCtx, n: PNode, forLoop: PNode): PNode =
           result.add(tupl)
           result.add(newSymNode(c.field, n.info))
         break
+  of nkSymChoices:
+    result = instFieldLoopBody(c, newSymNode(n.choices[0], n.info), forLoop)
```
zerbina:
```suggestion
  of nkSymChoices:
    result = copyNode(n)
```

The symbol choice needs to stay as one, otherwise a `fields` for-loop in a generic context is going to cause problems (I think this the reason the compiler fails to build).

---

(outside the scope of this PR) The implementation of the `fields` for-loop, especially `instFieldLoopBody`, seems to be needlessly complex while at the same time skipping over some edge cases. For example, the following fails with an "implicit object field construction requires a .partial object" error:
```nim
for a in fields(someObj):
  var a = 0 # <- the error happens here
  a = 1
```

I think a more sensible approach would be to internally treat the body of the loop as a template body, like so:
```nim
template loopBody(a: untyped): void =
  var a = 0
  a = 1
```
and then expand this template for each field. The benefits are that it's less logic duplication, `inUnrolledContext` becomes obsolete (it already is), and the above snippet would compile (although shadowing `a` would still not work).

If we want shadowing the for-vars to work, instead of treating the body as a template body, the symbol binding pre-pass (`semGenericStmt`) could be run first instead and the resulting AST then duplicated and processed by `semStmt` for each field. The only problem I can see with this approach, is that the symbol pre-pass (currently) unconditionally evaluates `untyped` macros and templates, which *could* violate user expectations.

PR #693, opened 2023-05-08
Rename the `native` gc option to `default`

ghost:
## Summary
Rename the new `native` gc option to `default` for improved clarity.
Made the `native` option internal only.

## Details
I raised a discussion about the naming of the new `native` gc option
due to it potentially being confusing depending on the backend.
We agreed to rename it to `default` to better fit what this option
actually means. I made the `native` option not choosable via
command-line switches as it should be an internal implementation detail.

---

## Notes for Reviewers
I mostly just renamed the changes from this pr https://github.com/nim-works/nimskull/pull/692.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:
* handle the single commit message requirement at the end of review.

### DIFF
--- a/compiler/front/main.nim
+++ b/compiler/front/main.nim
@@ -441,7 +441,7 @@ proc customizeForBackend*(graph: ModuleGraph, conf: ConfigRef,
     of excNone, excNative: conf.exc = excGoto
     of excGoto:            discard
 
-    if conf.selectedGC in {gcUnselected, gcNative}:
+    if conf.selectedGC == gcUnselected:
       # the default gc for the C backend is ORC. We can't just set it to
       # ``gcOrc`` directly, however, as additional defines, etc. are required
       # XXX: the dependency on ``optionsprocessor`` hints that a different
--- a/compiler/front/optionsprocessor.nim
+++ b/compiler/front/optionsprocessor.nim
@@ -589,7 +589,7 @@ proc processCompile(conf: ConfigRef; filename: string) =
 # multiple names for the same garbage collector.
 const
   gcNames = @[
-    "native", "boehm", "refc", "markandsweep", "destructors", "arc", "orc",
+    "default", "boehm", "refc", "markandsweep", "destructors", "arc", "orc",
     "hooks", "go", "none", "stack", "regions",]
 
   cmdNames = @[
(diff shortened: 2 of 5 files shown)


saem:
I don't think this is quite right, because the GC on the JS and VM targets is `native` to them, so simply replacing `native` with `default` now makes it more confusing for those.

compiler/front/optionsprocessor.nim
```diff
@@ -644,7 +644,7 @@ func testCompileOptionArg*(conf: ConfigRef; switch, arg: string): CompileOptArgC
   case switch.normalize
   of "gc":
     case arg.normalize
-    of "native": asResult conf.selectedGC == gcNative
+    of "default": asResult conf.selectedGC == gcDefault
```
zerbina:
```suggestion
    of "default":
      let default =
        case conf.backend
        of backendC:             gcOrc
        of backendJs, backendVm: gcNative
        of backendInvalid:       unreachable() 
      asResult conf.selectedGC == default
```

The `testCompilationOptionArg` is used to implement the `compileOption("gc", x)` magic, so a `gcDefault` cannot reach here.

Having said that, I currently can't come up with a situation where querying whether the default GC is selected makes sense, so I'm not sure if doing so should be allowed.

Is there a use-case you have in mind?

compiler/front/optionsprocessor.nim
```diff
@@ -1154,8 +1154,8 @@ proc processSwitch*(switch, arg: string, pass: TCmdLinePass,
     expectArg(switch, arg)
     if pass in {passCmd2, passPP}:
       case arg.normalize
-      of "native":
-        conf.selectedGC = gcNative
+      of "default":
+        conf.selectedGC = gcDefault
```
zerbina:
```suggestion
        conf.selectedGC = gcUnselected
```

compiler/front/main.nim
```diff
@@ -441,7 +441,7 @@ proc customizeForBackend*(graph: ModuleGraph, conf: ConfigRef,
     of excNone, excNative: conf.exc = excGoto
     of excGoto:            discard
 
-    if conf.selectedGC in {gcUnselected, gcNative}:
+    if conf.selectedGC in {gcUnselected, gcDefault}:
```
zerbina:
```suggestion
    if conf.selectedGC == gcUnselected:
```

zerbina (changes requested):
I think handling of the "default" GC should stay in the options processing / front-end of the compiler.

Please use `gcUnselected` to implement the "default" option. It reduces the amount of "unreachable" states that the rest of the compiler has to consider.

PR #694, opened 2023-05-08
cgen: remove leftover Genode support

zerbina:
## Summary

Much of the Genode OS framework support was removed as part of commit https://github.com/nim-works/nimskull/commit/ab1766e3444bbef25a52696e179bba62832b7ee9,
but some small parts of it remained. These are now removed too.


saem (approved):
yay, less stuff.

PR #695, opened 2023-05-09
system: declare `prepareMutation` for all targets

zerbina:
## Summary

The system procedure for preparing a string for mutation
(`prepareMutation`) was previously only available when targeting C,
requiring one to guard usages of it with `when` statements.

Make it available for all targets, with the ones not using copy-on-write
string literals treating it as a no-op. This makes it easier to write
code that works without change across all backend.

## Details

* declare `prepareMutation` for the JavaScript and VM target
* remove `when` guards around `prepareMutation` call in the standard
  library and tests

### DIFF
--- a/lib/pure/streams.nim
+++ b/lib/pure/streams.nim
@@ -100,13 +100,6 @@ proc newEIO(msg: string): owned(ref IOError) =
   new(result)
   result.msg = msg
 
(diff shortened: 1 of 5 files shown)


saem (approved):
That's much cleaner

PR #696, opened 2023-05-09
Make the new `native` gc internal only

ghost:
## Summary
Made the new `native` gc not accessible via command line arguments or
via `compileOption`.

## Details
The latest `native` gc was created to better represent the gc the JS and
VM backends use. But it has no meaning for the C backend and instead
redirected to orc. As there is no reason to explicitly choose the
`native` gc via command-line arguments I removed the option. I also
removed the option for the `compileOption` proc because one should
query the used backend instead.

### DIFF
--- a/compiler/front/main.nim
+++ b/compiler/front/main.nim
@@ -441,7 +441,7 @@ proc customizeForBackend*(graph: ModuleGraph, conf: ConfigRef,
     of excNone, excNative: conf.exc = excGoto
     of excGoto:            discard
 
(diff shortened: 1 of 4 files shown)


zerbina (approved):
Thank you. The `native` option being internal-only is indeed better.

**Edit:** I updated the PR message to fit the 72 characters line width. With the recent change to using the merge queue, the PR message, instead of the single commit's message, is now used as the message for the final commit, meaning that it should adhere to the commit message guidelines.

PR #697, opened 2023-05-09
lambdalifting: ignore AST literals for var capture

saem:
## Summary

`nkNimNodeLit` are no longer scanned during var capture detection,
ensuring that literals do not alter capture analysis of a routine
holding typed literal AST with symbols from the outer routine.

## Details

Capture detection is done in two modules `lambdalifting` and `sempass2`
in the `detectCapturedVars` and `detectCapture` procedures,
respectively. With this change both procedures do not traverse into
`nkNimNodeLit` (treating them as more opaque literal data) for literal
analysis. A regression tests has been added a test for this (thanks to
Zerbina).

### DIFF
--- a/compiler/sem/lambdalifting.nim
+++ b/compiler/sem/lambdalifting.nim
@@ -35,6 +35,7 @@ import
     transf,
     lowerings
   ]
+  , compiler/utils/astrepr
 
 # xxx: reports are a code smell meaning data types are misplaced
 from compiler/ast/reports_sem import SemReport,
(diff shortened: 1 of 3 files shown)


compiler/sem/lambdalifting.nim
```diff
@@ -35,6 +35,7 @@ import
     transf,
     lowerings
   ]
+  , compiler/utils/astrepr
```
zerbina:
```suggestion
```

zerbina (commented):
Thanks. `liftCapturedVars` also needs to skip `nkNimNodeLit` nodes, but otherwise looks good to me.

As for a test case, the following snippet could be used:
```nim
proc sym() =
  discard

proc outer() {.compileTime.} =
  var x = 0

  proc sym() {.compileTime.} =
    # close over an outer local in order to make `sym` a closure procedure:
    x = 0

  proc inner() {.compileTime.} =
    let x = bindSym"sym" # create a symbol choice that includes the closure
                         # procedure

  # `inner` doesn't close over a local and also doesn't access a closure
  # procedure, so it must not be a closure procedure
  doAssert inner isnot "closure"

static:
  # call the procedure to make sure it passes the lambda-lifting pass
  outer()
```

Unfortunately, `sempass2.detectCaptures` also doesn't skip `nkNimNodeLit` nodes, so the test would fail even with the changes to `lambdalifting`.

PR #702, opened 2023-05-12
sem: refactor `newSymGNode` replaces `newSymG`

saem:
## Summary

Internal refactor replacing `sem.newSymG` with `newSymGNode`, no change
in actual behaviour

## Details

This replaces the `newSymG`, which produced a symbol and effected
errors. `newSymGNode` will return an `nkSym` node on succees and
`nkError` if it failed. The `getDefNameSymOrRecover` proc can be used in
conjunction in order to ensure progress can be made in the presence of
errors.

All call sites updated in such a fashion are now more ready to be
converted to a more `nkError` aware style.

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -468,7 +468,7 @@ proc newSymGNode*(kind: TSymKind, n: PNode, c: PContext): PNode =
     else:
       unreachable("only produces `nkSym` or `nkError`")
 
-proc newSymG*(kind: TSymKind, n: PNode, c: PContext): PSym =
+proc newSymG*(kind: TSymKind, n: PNode, c: PContext): PSym {.deprecated: "replace with newSymGNode".} =
   # like newSymS, but considers gensym'ed symbols
   case n.kind
   of nkSym:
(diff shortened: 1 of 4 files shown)


compiler/sem/sem.nim
```diff
@@ -368,9 +368,9 @@ func getDefNameSymOrRecover*(n: PNode): PSym {.inline.} =
   of nkError:
     case n.diag.kind
     of adSemDefNameSym: n.diag.defNameSym
-    else: unreachable("all error cases must be covered")
+    else: unreachable("all error cases must be covered, got: " & $n.diag.kind)
```
zerbina, line 371: (no action required) If logging both a message plus an enum value in case of a failure is something that comes up often, then I think adding a dedicated `unreachable` overload would make sense.

compiler/sem/semtypes.nim
```diff
@@ -561,9 +574,9 @@ proc semIdentVis(c: PContext, kind: TSymKind, n: PNode,
     else:
       c.config.semReportIllformedAst(
         n, "Expected two nodes for postfix expression, but found " & $n.len)
-
   else:
-    result = newSymG(kind, n, c)
+    let identNode = newSymGNode(kind, n, c)
+    result = getDefNameSymOrRecover(identNode)
```
```suggestion
    result = getDefNameSymOrRecover(newSymGNode(kind, n, c))
```
Only a soft suggestion.

zerbina (approved):
Looks good to me. More progress towards a `localReport`-free semantic analysis :D

The one suggestion I left is non-blocking.

compiler/sem/sem.nim
saem: Smort!

saem:

PR #704, opened 2023-05-16
linter: remove nep-1

saem:
## Summary

Drop nep-1 as it's an awful standard, replaced with a more lax
implementation that can be tightened up over time.

## Details

This remove all mentions of nep-1, updates the style checker to be
closer to symbol kind --> naming style policy map. This policy in turn
is applied to the name string.

The styles remain an incredible mess and there really isn't a great way
to setup this rules based on the current level of information and style
control input. A major overhaul is required in the future, this is a
small stepping stone.

With that said, updated the internals, cleaned up the docs, and added a
few tests. Additionally, moved the `tlinter` test to the `stylecheck`
category along with the rest of the style related tests.

Lastly a few selective renames were applied to ensure things built.

### DIFF
--- a/compiler/ast/linter.nim
+++ b/compiler/ast/linter.nim
@@ -83,7 +83,6 @@ proc beautifyName(s: string, k: TSymKind): string =
     allUpper = allCharsInSet(s, {'A'..'Z', '0'..'9', '_'})
     hasUnderscore = '_' in s
   if allUpper and k in {skType, skConst, skEnumField}: return s
-  result = newStringOfCap(s.len)
 
   # fast tracked handling
   case k
@@ -112,6 +111,7 @@ proc beautifyName(s: string, k: TSymKind): string =
   if result != "":
     return # we got a fast tracked result
 
+  result = newStringOfCap(s.len)
   var style: NamingStyle
   case k
   of skContainers:
@@ -197,7 +197,8 @@ proc checkDefImpl(conf: ConfigRef; info: TLineInfo; s: PSym; k: TSymKind) =
   if s.typ != nil and s.typ.kind == tyTypeDesc: return
   if {sfImportc, sfExportc} * s.flags != {}: return
   if k == skParam and {sfImportc, sfExportc} * s.owner.flags != {}: return
-  if optStyleCheck notin conf.options: return
+  if optStyleCheck notin s.options: return # xxx: invert this option so the
+                                           #      default is to check
   let wanted = beautifyName(s.name.s, k)
   if s.name.s != wanted:
     conf.localReport(info, SemReport(
--- a/compiler/ast/linter.nim
+++ b/compiler/ast/linter.nim
@@ -200,7 +200,6 @@ proc checkDefImpl(conf: ConfigRef; info: TLineInfo; s: PSym; k: TSymKind) =
   if optStyleCheck notin conf.options: return
   let wanted = beautifyName(s.name.s, k)
   if s.name.s != wanted:
-    echo "name: ", s.name.s, " owner.name: ", s.owner.name.s, " owner.flags: ", s.owner.flags
     conf.localReport(info, SemReport(
       sym: s,
       kind: rsemLinterReport,
--- a/compiler/ast/linter.nim
+++ b/compiler/ast/linter.nim
@@ -8,6 +8,20 @@
 #
 
 ## This module implements the style checker.
+## 
+## Directionally, style will become part of the language, think `go fmt` where
+## violations are errors.
+## 
+## Currently focused on names, but should cover spacing, and more.
+## 
+## ## General rules of style:
+## - Saem decides style matters
+## - lower/upper camel case good; snake case bad
+## - lower camel case for most things
+## - upper camel case for non-runtime things (types, some consts)
+## - single letter names are ok, should be lower; tolerate upper
+## - all-caps snake case is occassionally tolerated
+
 import
   std/[
     strutils
@@ -21,9 +35,12 @@ import
   compiler/front/[
     options,
     msgs
+  ],
+  compiler/utils/[
+    idioms,
   ]
 
-# TODO: linter should have it's own diag/event/telemetry types
+# TODO: linter should have its own diag/event/telemetry types
 from compiler/ast/reports_sem import SemReport
 
 const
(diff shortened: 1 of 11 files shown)


saem:
I'm inclined to lower case the proc name `NSGetEnviron` to `nsGetEnviron` rather than introduce mixed camel casing for routine kinds.

saem:
It's going to be one of those PRs, where it has to be baby sat through CI. 🙄

compiler/ast/linter.nim
```diff
@@ -37,54 +54,119 @@ proc `=~`(s: string, a: openArray[string]): bool =
...
+  # xxx: these style rules are a mess, any changes to the actual rules require
+  #      Saem's approval in the PR.
 
-  let allUpper = allCharsInSet(s, {'A'..'Z', '0'..'9', '_'})
-  if allUpper and k in {skConst, skEnumField, skType}: return s
+  let
+    allUpper = allCharsInSet(s, {'A'..'Z', '0'..'9', '_'})
+    hasUnderscore = '_' in s
+  if allUpper and k in {skType, skConst, skEnumField}: return s
   result = newStringOfCap(s.len)
```
zerbina: Since the fast-track handling doesn't need the pre-allocated buffer, this line can be moved to after the early exit (line 113) -- it's a little bit more efficient.

compiler/ast/linter.nim
```diff
@@ -37,54 +54,119 @@ proc `=~`(s: string, a: openArray[string]): bool =
...
+      case s[i]
+      of '_':
+        case style
+        of upperCamelStyle, shoutingStyle, lowerCamelStyle, mixedCamelStyle:
+          result.add '_'
+          inc i
+          result.add s[i]
+        of lowerCamelStrict:
+          inc i
+          result.add toUpperAscii(s[i])
```
zerbina, line 158: It looks to be pre-existing, but doesn't the logic here mean that `a__B` gets turned into `a_B`?

compiler/ast/linter.nim
```diff
@@ -96,41 +178,44 @@ proc differ*(line: string, a, b: int, x: string): string =
...
   if k in {skResult, skTemp} or s.name.s[0] notin Letters: return
   if k in {skType, skGenericParam} and sfAnon in s.flags: return
   if s.typ != nil and s.typ.kind == tyTypeDesc: return
   if {sfImportc, sfExportc} * s.flags != {}: return
-  if optStyleCheck notin s.options: return
+  if k == skParam and {sfImportc, sfExportc} * s.owner.flags != {}: return
+  if optStyleCheck notin conf.options: return
   let wanted = beautifyName(s.name.s, k)
   if s.name.s != wanted:
+    echo "name: ", s.name.s, " owner.name: ", s.owner.name.s, " owner.flags: ", s.owner.flags
```
zerbina:
```suggestion
```

compiler/front/in_options.nim
```diff
@@ -23,8 +23,8 @@ type
     optGenMapping             ## generate a mapping file
     optRun                    ## run the compiled project
     optUseNimcache            ## save artifacts (including binary) in $nimcache
-    optStyleHint              ## check that the names adhere to NEP-1
-    optStyleError             ## enforce that the names adhere to NEP-1
+    optStyleHint              ## check that the names adhere to style guide
+    optStyleError             ## enforce that the names adhere to style guide
```
zerbina:
```suggestion
    optStyleHint              ## check that the names adhere to the style guide
    optStyleError             ## enforce that the names adhere to the style guide
```

compiler/ast/linter.nim
```diff
@@ -96,41 +178,44 @@ proc differ*(line: string, a, b: int, x: string): string =
...
+
+proc checkDefImpl(conf: ConfigRef; info: TLineInfo; s: PSym; k: TSymKind) =
   # operators stay as they are:
   if k in {skResult, skTemp} or s.name.s[0] notin Letters: return
   if k in {skType, skGenericParam} and sfAnon in s.flags: return
   if s.typ != nil and s.typ.kind == tyTypeDesc: return
   if {sfImportc, sfExportc} * s.flags != {}: return
-  if optStyleCheck notin s.options: return
+  if k == skParam and {sfImportc, sfExportc} * s.owner.flags != {}: return
+  if optStyleCheck notin conf.options: return
```
zerbina: I haven't tested it, but wouldn't the change here cause the following to fail (assuming that style checks are enabled globally):
```nim
{.push styleChecks: off.}

proc p[T]() =
  var Illegal = 1 # non-conforming name according to the style rules

{.pop.}

# instantiate:
p[int]()
```

zerbina (commented):
Some small suggestion and questions, but otherwise looks good to me. Making the style rules explicit via the enum is nice.

For the direction, is style insensitivity something that is going to be removed? Or are styles not adhering to the official style guide only an error if explicitly requested?

**Edit**: I removed a superfluous 'the' from the PR message.

compiler/ast/linter.nim
saem: Ah, good eye. I'll leave it as is for now.

My thinking is these style rules are super capricious. Nothing really makes sense and require a ground up rethink using examples to highlight the kind of code we're want to see and the kind we don't. I got a ton of failures and the use of mixed camel case, uppercase checks, etc is me hacking around them.

compiler/ast/linter.nim
saem: True, I found lots of symbols seem to lose style checks and didn't take the time to debug that. I'll revert this and might have to disable the test until it's fixed for good.

saem:

saem:
> For the direction, is style insensitivity something that is going to be removed?

I'm on the fence about style insensitivity, it's subtly good:
- it prevents minor casing changes from becoming breaking changes for APIs when doing named parameter calls
- it allows me to not write snake case, or whatever weird thing, just because someone thought it was a great idea
- it eliminates an entire class to typos (errors) out of existence

I don't care about it in the long run, because transpiling is not a goal, but right now it helps. The papering over c APIs to bind to imported c functions.

> Or are styles not adhering to the official style guide only an error if explicitly requested?

It partially depends upon where we land of the first question and I'm not sure what the right answer is there.

The highly probable outcome is that once a style is settled upon it's an error unless _disabled_. I'd rather the few programmers going against the grain pay the price than have the majority work with poor defaults.

PR #705, opened 2023-05-18
injectdestructors: keep the relative order of definitions

zerbina:
## Summary

Sort the list of all entities requiring destruction *before* moving
their definitions. This ensures that relative order of the *moved*
definitions is kept intact.

The relative order changing had no effect on program semantics at this
time, but might cause issues in the future. It was also observable via
`--expandArc`, which led to spurious test failures when the IDs assigned
to symbols changed as part of an otherwise unrelated change.

## Details

The order of the items in the list of all entities requiring destruction
is dependent on the order in which the `Table.pairs` iterator yields the
pairs, meaning that changes to the hash function or symbol IDs (which
are used as the key) led to changes in the order that definition
statement for locals had after the `injectdestructors` pass.

Sorting the list prior to iterating over it for the purpose of moving
definitions ensures that the moved definitions' relative order is kept
intact.

### DIFF
--- a/compiler/sem/injectdestructors.nim
+++ b/compiler/sem/injectdestructors.nim
@@ -1089,6 +1089,16 @@ proc injectDestructors(tree: MirTree, graph: ModuleGraph,
 
     entries.add (pos: pos, scope: scopeStart)
 
+  # sort the entries by scope (first-order) and position (second-order) in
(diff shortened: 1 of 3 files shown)


saem (approved):
yay, stable ordering!

PR #706, opened 2023-05-18
cgen: fix a `FieldDefect` crash

zerbina:
## Summary

Pass the AST representing a type's default value to `rawConstExpr`
instead of misusing the `nkType` node kind for this.

Using a `nkType` node only worked by accident, and started to fail after
the `sons` field became inaccessible for `nkType` nodes.


saem (approved):
One step closer to release testing. 👍🏽

PR #709, opened 2023-05-19
parser: graceful keyword in ident position error

saem:
## Summary

The parser now makes progress even if an keyword is present in an
identifier position.

## Details

Previously the `parser` module did not gracefully handle keywords
present in identifier positions, such as `add` in `var add = 10`, here
`add` needs to be escaped, via accent quotes. Instead it stopped
progressing and this led to spurious invalid indentation errors and
early termination of incremental parsing.

Now when a keyword is found in an identifier position, the keyword token
 is turned into an identifier node and processed as if it was
 successful. This means a diagnostic is generated for the error, but the
 token is still consumed (progress).

 After the change the `07_keywords_fail` language specification test now
 works correctly.

 As part of this change an opportunistic refactoring of the
 `lexer.getEscapedChar` was conducted to clean-up some code.


saem:
The parser still feels "weird" to me, so if either or both @zerbina and @Clyybber could look at this PR, that'd be appreciated.

compiler/ast/parser.nim
```diff
@@ -447,7 +447,9 @@ proc parseSymbol(p: var Parser, mode = smNormal): ParsedNode =
       p.getTok
     else:
       p.invalidExpectedIdent()
-      result = p.emptyNode
+      # result = p.emptyNode
```
zerbina:
```suggestion
      # error correction: consume the token and treat it as an identifier in
      # order to allow for further progress
```

compiler/ast/parser.nim
```diff
@@ -2111,6 +2116,8 @@ proc parseVariable(p: var Parser): ParsedNode =
     result.add parseExpr(p)
   else:
     result = p.parseIdentColonEquals({withPragma, withDot})
+  if result.len == 0:
+    echo p.tok
```
zerbina:
```suggestion
```

compiler/ast/parser.nim
```diff
@@ -1821,6 +1822,10 @@ proc parseSection(p: var Parser, kind: ParsedNodeKind,
   elif p.tok.tokType in {tkSymbol, tkAccent, tkParLe} and p.tok.indent < 0:
     # tkParLe is allowed for ``var (x, y) = ...`` tuple parsing
     result.add defparser(p)
+  elif p.tok.tokType.isKeyword() and p.tok.indent < 0:
+    # in order to make progress accept the unescaped keyword
+    # p.invalidExpectedIdent()
```
zerbina:
```suggestion
    # in order to make progress accept the unescaped keyword. The definition
    # parsing is responsible for emitting an error diagnostic, so none is added
    # here
```
For slightly improved clarity.

In addition, I think a branch for `tkKeyWords` also needs to be added to the loop above (line 1808) in order for the following to parse (albeit with errors):
```nim
var
  addr = 0
```

compiler/ast/parser.nim
```diff
@@ -996,6 +993,10 @@ proc parseIdentColonEquals(p: var Parser, flags: DeclaredIdentFlags): ParsedNode
       if withPragma in flags: a = p.identWithPragma(allowDot=withDot in flags)
       else: a = parseSymbol(p)
       if a.kind == pnkEmpty: return
+    of tkKeyWords:
```
zerbina: Given that they're doing the same thing, I think it makes sense to merge this with the other branch.

zerbina (commented):
To preface my review: I only have limited experience with the parser.

Having said that, I do think applying error correction for keywords in definition positions is a good idea.

compiler/ast/parser.nim
saem: Good call.

compiler/ast/parser.nim
saem: Yeah, those were different initially. Only reason I kept them apart was because it highlighted one is "the odd one out" even at this point.

saem: added some comments and combined the paths.

zerbina (approved):
Looks good to me!

saem:
Thanks, @zerbina.

Time willing, @Clyybber can look at it later; we can always change the code later.

Clyybber:
I'm quite late but LGTM :)

PR #712, opened 2023-05-19
use whole-program code-generation for all backends

zerbina:
## Summary

Change the C and JavaScript backends to use whole-program code-
generation in the same way that the VM and IC backends do. In short,
this means that code generation is now only run after *all* modules part
of the program were semantically analyzed.

This brings the architecture of the C and JS backends closer to that of
the VM backend, and is the first step towards unifying the backend
processing.

The final goal is to have all backends process code in the same way,
with as much as possible of the pre-processing currently performed by
the code generators being moved to a shared, backend-agnostic layer.

## Details

Generalize the `ModuleGraph` pass that the VM backend used for gathering
the AST of each alive module, and move it into its own module. Compared
to the original implementation, the generalized pass:
* doesn't drop declarative nodes (the VM backend doesn't care about
  them, but the C and JS backends do)
* remembers the `IdGenerator` associated with each module
* doesn't introduce its own module order (that's left to the backend)
* uses the more descriptive `SeqMap` instead of a raw `seq`

`vmbackend` is adjusted to work with the generalized collection pass:
the module list produced by the pass is translated into the structure
that the rest of the VM backend still expects, but a `Store` with a
dedicated ID type is now used instead of a raw `seq` (preparing for
future improvements).

Using the same naming scheme as `vmbackend`, the modules `cbackend`
and `jsbackend` are introduced -- they implement the code-generation
orchestrators for the C and JavaScript backends, respectively. The
`passes` integration is removed from `cgen` and `jsgen`, as invoking the
code generators is now the responsibility of the orchestrators.

The new orchestrators take the module list produced by the collector
pass and generate the code for it (with `jsbackend` also writing the
output to disk already). They're very basic at the moment, but in the
future will take on similar responsibilities as `vmbackend` does
for the VM backend (e.g., dead-code elimination, running `transf`,
etc.).

Since forwarded procedures don't reach the code generators anymore, the
special handling for them is removed.

### Known Issues

Changes to options performed by `.push` having an effect on top-level
code relied on semantic analysis and code generation happening in a
pipelined manner.

Since this is no longer the case, disabling or enabling checks for top-
level code stops working for now, but the plan is to bring this feature
back in the future.


compiler/backend/collectors.nim
```diff
@@ -0,0 +1,82 @@
+## Implements the "collect" pass. This pass gathers the full AST for each
+## module into a single structure, which is then meant to be consumed by the
+## code-generation orchestrators.
```
saem:
```suggestion
## code-generation orchestrators (`cbackend`, `jsbackend`, etc).
```

Orchestrator doesn't appear in any identifiers so I figured a bit of clarity here might send folks new to the code in the right direction.

compiler/modules/modulegraphs.nim
```diff
@@ -108,7 +108,9 @@ type
                                             # first module that included it
     importStack*: seq[FileIndex]  # The current import stack. Used for detecting recursive
                                   # module dependencies.
-    backend*: RootRef # minor hack so that a backend can extend this easily
+    backend*: RootRef # XXX: having this field is a hack, but it's still
+                      #      required by the current ``passes`` design. Remove
+                      #      the ``passes`` design is phased out
```
saem:
```suggestion
                      #      once the ``passes`` design is phased out
```

tests/overflw/toverflw.nim
```diff
@@ -1,24 +1,31 @@
...
+  r = (b - a)
 except OverflowDefect:
   overflowDetected = true
 
 {.pop.} # overflow check
 
-doAssert(overflowDetected)
+# XXX: overflow checks (and other checks) cannot be disabled for module-level
+#      code at the moment
+doAssert(overflowDetected == false)
```
saem:
```suggestion
doAssert(overflowDetected == false), "re-enable this test, module-level code bounds checking now works"
```

saem (approved):
Looks good.

I left a few minor suggestions mostly to do with a typo here or there in comments.

~I saw a few in the PR desc, I'll tackle those after I take a walk.~

saem:
I was going to suggest it but things are too different right now, but we can probably get away with a single `backend` module, that has a single public routine, something like: `generateCode*(g: ModuleGraph, mlist: sink ModuleList, backend: TBackend)`, this would evolve over time, likely become an iterator producing progress events. But that's all nice to have and that consolidation doesn't buy us enough as things are still not similar enough.

zerbina:
> I was going to suggest it but things are too different right now, but we can probably get away with a single `backend` module

Yep, that's what I initially tried as part of #550, but the differences in processing were too large for it to make sense at this time. With that said, having a single entry point into the backend *is* roughly the direction I'm going in.

PR #714, opened 2023-05-21
backend: lower modules into procedures

zerbina:
## Summary

- turn module-level imperative code into a procedure (a module's 'init'
  procedure)
- generate the program entry-point in a mostly backend-agnostic way
- change the destruction order of globals, and make it the same across
  backend
- make initialization behaviour of (module-level) globals the same
  across all backend
- make module initialization order the same across all backends (except
  the IC backend)

Almost all decision making regarding initialization order of globals and
modules is moved out of the code generators, and into a central place,
simplifying the code generators and making it much easier to adjust this
behaviour in the future.

Notable changes in behaviour:
- the correct module initialization order is now the one previously used
  with the C backend
- "lifted" globals (those marked with the `.global` pragma) for a module
  are now always initialized before the module-level ones 
- globals are destroyed on a per-module basis; modules are destroyed in
  the reverse order they were initialized (i.e. their module-level code
  run) in
- the `header` pragma is now respected for globals that don't have an
  initial value specified
- `asm` and `emit` statement cannot be used for emitting top-level
  JavaScript code anymore

### Lowering a module

Turning a module's imperative code into a procedure is integrated into
the collect pass (`setupModule`); once the full semanticized AST for a
module is present, the parts of the declarative AST that the code
generators still need is wrapped in a standalone statement list, while
the imperative AST is pre-processed and wrapped in a procedure.

Pre-processing means:
- scanning the module-level AST for definitions of globals and
  threadvars, and registering them with the `Module` instance
- changing the owner of all symbols defined in the imperative AST to be
  the wrapped-in procedure

The `var` and `let` sections with the globals cannot be turned into
assignments during the pre-processing. This is because:
- `nkForStmt`s weren't lowered yet
- the `injectdestructors` still needs to know about the definition
  points of globals in nested scopes
- the semantics of an "initializing assignment" cannot be encoded with
  `PNode` AST

So instead, transforming the definitions into assignment is done during
the MIR phase (more on that later). Before wrapping up setup of a
`Module` instance, symbols for all other "module-bound" operators are
created. These are:
- module-data initialization procedure
- module de-init (a.k.a. destructor) procedure
- module pre-init procedure

The module destructor is already filled with destructor calls for
globals defined in the outermost scope.

### MIR phase

Transforming the definitions of globals is implemented as two MIR passes
(`rewriteGlobalDefs`). They're run as follows:
- rewrite definitions of globals defined in the *outermost* scope
- `injectdestructors` pass
- rewrite definitions of globals defined in *nested* scopes

With handling of destruction for globals defined in the outermost scope
now being the responsibility of the earlier module processing,
`injectdestructors` only has to defer destructor calls (the
`globalDestructors` mechanism) for lifted globals. The only reason the
`globalDestructors` still needs to exists is that is that `jsgen`
doesn't yet extract lifted globals prior to running the MIR passes. 

After all MIR passes have run, no definitions for globals are left in
the AST that is output for the init procedures -- they were all replaced
with assignments.

### Code generation

The new `backends` module contains code-generator-agnostic routines that
are used by all orchestrators.

The changes to the code generators / orchestrators can be summarized as:
- emitting / handling the definitions for globals is now the
  responsibility of the orchestrators
- deciding on the content of the 'main' procedure is the responsibility
  of the orchestrators (which delegate this to `generateMain`,
  `generateTeardown`, and `generateMainProcedure`)
- the code generators are no longer responsible for handling destruction
  of globals

It's only possible to generate the code for destroying lifted globals
once all alive ones were discovered, but this is only the case after the
main part of code generation is done, as only then has all alive code
been discovered. For this reason, the orchestrators are currently
responsible for calling `finishDeinit` at the correct point in time.

Additional information and highlighted changes.

#### C backend

`cgen` still needs to append code to a module's init procedure from
outside of the init procedure, meaning that the module-init procedures
still have to be special cased.

The code generation logic around globals simplifies a bit, and
`cgenWriteModules` now no longer alters the C code for a module. Because
'main' procedure generation is handled elsewhere, the call to
`initThreadVarsEmulation` is appended to the end of the `system`
module's data-init procedure.

#### JS backend

A special stack-trace entry is inserted for module-level code when using
the JS backend. In order to keep this behaviour, the init procedure is
marked with a symbol flag (the repurposed `sfMainModule`) that signals
to `genProcBody` that the special stack-trace entry is to be used.

#### VM backend

A lot of the code in `vmbackend` was dedicated to module-level code and
entry-point generation, so the module shrinks down quite a bit.

- re-use the "module closed" order information from `ModuleList`
- use the correct module context when generating code
- remove dedicated support for module-level code generation
- replace the manual entry-point code generation with enqueuing the
  procedure produced by `generateMainProcedure`

#### IC backend

In order to make the backend unification process easier, the IC backend
is turned into a frontend for `backend/cbackend.nim`. It mostly acts as
a rodfile -> `ModuleList` adapter now, but still performs the alive
analysis.

A negative side-effect of this is, that when using the IC backend, the
code generator is now also invoked for modules of which neither the
content nor the set of alive symbols changed, significantly increasing
compile times when nothing changed.

### Other changes

- adjust the JavaScript implementation of `std/sysrand`, now that `emit`
  cannot be used to emit top-level JavaScript code anymore
- mark the `globalsSlot` global in `threadlocalstorage` as `.noInit`,
  which is necessary now that all globals without an initial value are
  default-initialized by the module's init procedure

### Changes to tests

- move the test from `tclosure_issues.nim` into a `knownIssue` test; it
  now fails with all backends
- change the `js/tclosure.nim` test to not rely on `asm` producing top-
  level JavaScript code
- don't import modules unavailable for the current platform in
  `tstyle_imports` and `tstrict_funcs_imports` (necessary now that the
  `header` pragma on globals works properly)
- enable some global-related tests for the VM target


saem:
OMG, this so good. Making it a proc would make it so so good.

zerbina:
I rebased the PR onto the `devel` branch, which fixed the closure-iterator-related issues. In addition, I've also rewritten the commit history, making the commits less broad and easier to follow along.

During the second bootstrap iteration, the compiler crashes with an access violation error inside one of procedures related to `refc` -- a `devel` compiler is able to build the compiler without any problems. An update of the csources compiler is probably needed now.

compiler/backend/ccgstmts.nim
```diff
@@ -156,40 +156,31 @@ proc genGotoVar(p: BProc; value: PNode) =
...
     # translate 'var state {.goto.} = X' into 'goto LX':
     genGotoVar(p, value)
     return
   var targetProc = p
-  var traverseProc: Rope
-  let valueAsRope = potentialValueInit(p, v, value)
   if sfGlobal in v.flags:
+    #assert {sfPure, sfThread} * v.flags != {}, "normal globals don't reach here"
+    # XXX: thanks to ``.compileTime`` globals, the assertion cannot be used
+    #      yet
```
zerbina, line 168: I'm going to try disallowing their usage outside of compile-time-only context once this PR is merged.

The naive approach of detecting illegal usages during the first semantic pass didn't work out -- the detection probably needs to happen during the second pass (`sempass2`).

compiler/backend/cgen.nim
```diff
@@ -1663,24 +1603,16 @@ proc shouldRecompile(m: BModule; code: Rope, cfile: Cfile): bool =
...
-# elimination and incremental compilation! Non pending modules need no
-# such logic and in fact the logic hurts for the main module at least;
-# it would generate multiple 'main' procs, for instance.
+proc finalizeModule*(m: BModule) =
+  # genInitCode(m)
+  finishTypeDescriptions(m)
 
-proc writeModule(m: BModule, pending: bool) =
+proc finalizeMainModule*(m: BModule) =
+  generateThreadVarsSize(m) # TODO: not the job of the code generator
```
zerbina, line 1597: It's still done in the code generator for now, but will eventually become the responsibility of the orchestrator.

compiler/backend/cgen.nim
```diff
@@ -1706,11 +1638,9 @@ proc finalCodegenActions*(graph: ModuleGraph; m: BModule; n: PNode) =
...
   if sfSystemModule in m.module.flags:
     discard cgsym(m, "nimTestErrorFlag")
-
-  if sfMainModule in m.module.flags:
-    if {optGenStaticLib, optGenDynLib, optNoMain} * m.config.globalOptions == {}:
-      for i in countdown(high(graph.globalDestructors), 0):
-        n.add graph.globalDestructors[i]
+    if emulatedThreadVars(m.config) and m.config.target.targetOS != osStandalone:
+      # raise the dependency on behalf of ``genDatInitCode``:
+      discard cgsym(m, "initThreadVarsEmulation")
```
zerbina, line 1629: This kind of dependency raising is all going to be phased out.

Ideally, the bits around generating the thread-local storage setup/teardown logic are all implemented in a backend-agnostic way, so that they can be re-used by the VM and JS backends (once they support threading, that is).

compiler/ic/cbackend.nim
```diff
@@ -183,22 +163,51 @@ proc generateCode*(g: ModuleGraph) =
...
+      continue
+
+    let
+      pos = m.module.position
+      c = pass.open(g, m.module, backend.modules[pos].idgen)
+    for p in allNodes(m.fromDisk.topLevel):
+      let n = unpackTree(g, pos, m.fromDisk.topLevel, p)
+      discard pass.process(c, n)
+
+    # XXX: the order in which the modules are closed is incorrect
```
zerbina, line 193: A pre-existing issue. The IC mechanism doesn't record the module closed order, so the order in which modules are registered with the file manager is used.

compiler/mir/mirbridge.nim
```diff
@@ -80,6 +82,89 @@ proc echoOutput(config: ConfigRef, owner: PSym, body: PNode) =
...
+        body[def].sym.owner.kind == skModule and
+        (not outermost or depth == 1):
+        let
+          sym = body[def].sym
+          typ = sym.typ
+        changes.seek(i)
+        if hasInput(body, Operation i):
+          # the definition has a starting value
+          # XXX: consider disallowing inputs to 'def's, they complicate things
+          #      quite a bit (as made evident here)
```
zerbina: I'm still a bit torn on whether or not to disallow this. While `def`s supporting input makes rewriting them harder (which is the reason for `mirgen` emitting a `def` + `init` for locations that have lifetime hooks), it, by construction, prevents the definition from being separated from its initial assignment.

compiler/mir/mirbridge.nim
```diff
@@ -80,6 +82,89 @@ proc echoOutput(config: ConfigRef, owner: PSym, body: PNode) =
...
+              buf.add MirNode(kind: mnkTag, effect: ekReassign, typ: typ)
+              buf.add MirNode(kind: mnkName, typ: typ)
+              buf.add MirNode(kind: mnkTemp, temp: tmp, typ: typ)
+              buf.add MirNode(kind: mnkConsume, typ: typ)
+            buf.add MirNode(kind: mnkInit)
+
+        elif sym.typ.kind in {tyVar, tyLent}:
+          # XXX: this works around an issue in ``transf``, where initialization
+          #      of the view is disjoint from its definition. Fix the bug and
+          #      then remove this special case
```
zerbina: The issue is fixed by #744.

compiler/sem/collectors.nim
```diff
@@ -1,17 +1,31 @@
...
+## For integration with the ``passes`` interface, a simple adapter pass is
+## provided. It sets up and populates a ``ModuleList`` instance, which can
+## then be retrieved from the ``ModuleGraph`` via ``takeModuleList`` after
+## all passes have run.
+##
+## The collection part is somewhat similar to the rodfile-based IC backend, but
+## instead of reading the modules' content from the rodfiles, it's collected
+## via the pass interface.
+
+# TODO: the module needs a better name
```
zerbina: Suggestions for better names are welcome -- I haven't been able to come up with one yet.

compiler/sem/collectors.nim
```diff
@@ -46,18 +99,310 @@ type
...
+    # procedure.
+    # XXX: ^^ this is limitation of the lambda-lifting pass
+    discard
+  of nkWithoutSons, nkTypeSection:
+    discard "ignore"
+  of nkWithSons - entityDefs - nkTypeSection:
+    for it in n.items:
+      changeOwner(it, newOwner)
+
+proc setupModule*(graph: ModuleGraph, idgen: IdGenerator, m: PSym,
```
zerbina: The procedure might need a more descriptive name. While it does set up a `Module` instance, it also does a lot more.

compiler/ast/ast_query.nim
```diff
@@ -103,6 +103,9 @@ const
   routineDefs* = declarativeDefs + {nkMacroDef, nkTemplateDef}
   procDefs* = nkLambdaKinds + declarativeDefs
   callableDefs* = nkLambdaKinds + routineDefs
+  entityDefs* = callableDefs + {nkIdentDefs, nkVarTuple, nkConstDef, nkForStmt}
```
saem, line 106: I like it.

tests/stylecheck/tstyle_imports.nim
saem, line 1: note to self: this changed because the `.headers` pragma now works correctly.

tests/misc/tlifted_lambda_issue.nim
```diff
@@ -0,0 +1,28 @@
...
+"""
+
+# https://github.com/nim-lang/nim/issues/7104
+
+var output: seq[int]
+
+proc sp(cb: proc())=
+  cb()
+
+sp:
```
saem, line 17: Ah, definite bug in sem, I think it might be easier to fix now that I dropped `nfBlockArgs`. Although, thinking about it a bunch of small fixes might be required before the behaviour is quite right. I'll see if I can poke at it after I've wrapped up some other sem changes.

lib/std/sysrand.nim
```diff
@@ -83,7 +83,9 @@ when defined(js):
   import std/private/jsutils
 
   when defined(nodejs):
-    {.emit: "const _nim_nodejs_crypto = require('crypto');".}
+    proc require(p: cstring): JsRoot {.importc.}
+
+    var crypto_module {.exportc: "_nim_nodejs_crypto".} = require("crypto")
```
saem:
```suggestion
    let crypto_module {.exportc: "_nim_nodejs_crypto".} = require("crypto")
```

Not that it should make too much of a difference, but it's the right idea as it _should_ be an immutable ref.

compiler/sem/collectors.nim
```diff
@@ -46,18 +99,306 @@ type
...
+                           name          = newSymNode(result),
+                           pattern       = graph.emptyNode,
+                           genericParams = graph.emptyNode,
+                           pragmas       = graph.emptyNode,
+                           exceptions    = graph.emptyNode)
+
+proc registerGlobals(stmts: seq[PNode], structs: var ModuleStructs) =
+  ## Create an entry in `structs` for each global or threadvar defined at the
+  ## the module level. `stmts` represents the imperative part of a module's
+  ## body.
```
saem:
```suggestion
  ## module level (within the module imperative body `stmts`).
```

compiler/sem/collectors.nim
```diff
@@ -46,18 +99,306 @@ type
...
+  for i in countdown(modules.modulesClosed.high, 0):
+    yield modules[modules.modulesClosed[i]]
+
 proc takeModuleList*(graph: ModuleGraph): ModuleList =
   ## Moves the ``ModuleList`` set up by the collector pass out of the
   ## `graph.backend` field and returns it.
   result = move ModuleListBackend(graph.backend).modules
   graph.backend = nil
 
+proc distribute(n: PNode, decl, imperative: var seq[PNode]) =
```
saem:
```suggestion
proc classify(n: PNode, decl, imperative: var seq[PNode]) =
```

An alternative name, might be better. Either that or `bucket` or `group`; `distribute` doesn't indicate the aggregation or collecting aspect.

compiler/sem/collectors.nim
```diff
@@ -20,23 +34,62 @@ import
...
   ]
 
+from compiler/sem/injectdestructors import genDestroy
+
 type
+  ModuleStructs* = object
+    globals*: seq[PSym]
+      ## all top-level globals part of the module
+
+    globals2*: seq[PSym]
```
saem:
```suggestion
    nestedGlobals*: seq[PSym]
```

Soft suggestion, might not be worth it consider the field might go away eventually.

compiler/sem/collectors.nim
```diff
@@ -20,23 +34,62 @@ import
...
+    sym*: PSym          ## module symbol
+    idgen*: IdGenerator ## the ID generator associated with the module
+
+    decls*: PNode
+      ## all declarative statements (type, routine, and constant
+      ## definitions)
+    structs*: ModuleStructs
+      ## the contents of the module's structs
+
+    # the module-bound procedures. Conceptually, these operate on the module
```
saem:
```suggestion
    # the module-bound procedures. Conceptually, these operate on the module's
```

compiler/sem/collectors.nim
saem: - `modulelowering`
- `moduleloweringpass`

compiler/mir/mirbridge.nim
saem: I would leave them together for now. There might be some value in having more specialized node kinds, such as with initialization, no-initialization, and default initiatlization (something like that). Encoding the decision like that upfront might take out some of the querying/conditionals mixed into the transformation logic.

compiler/mir/mirbridge.nim
```diff
@@ -80,6 +82,87 @@ proc echoOutput(config: ConfigRef, owner: PSym, body: PNode) =
...
+              buf.add MirNode(kind: mnkGlobal, sym: sym, typ: typ)
+              buf.add MirNode(kind: mnkTag, effect: ekReassign, typ: typ)
+              buf.add MirNode(kind: mnkName, typ: typ)
+              buf.add MirNode(kind: mnkTemp, temp: tmp, typ: typ)
+              buf.add MirNode(kind: mnkConsume, typ: typ)
+            buf.add MirNode(kind: mnkInit)
+        elif {sfImportc, sfNoInit} * sym.flags == {} and
+             {lfDynamicLib, lfNoDecl} * sym.loc.flags == {}:
+          # XXX: ^^ re-think this condition from first principles. Right now,
+          #      it's just meant to make some tests work
```
saem, line 133: These seem like "uncontrolled" locations, having said that, I suspect sem isn't quite doing enough here.

saem (approved):
This is awesome! All my comments are really minor, but there is lots of great consolation and things are quite a bit simpler.

We're on the cusp of a module type, too. Lots of exciting stuff. 🤩

compiler/sem/collectors.nim
zerbina: Yeah, `distribute` I was very unsure about -- I think I like `group` most.

tests/misc/tlifted_lambda_issue.nim
zerbina: The elif branch in `sigmatch.paramTypesMatchAux` that uses `matchesVoidProc` is where lifting a statement list into a lambda expression happens.

zerbina:

tests/misc/tlifted_lambda_issue.nim
saem: Crap, I lost my reply. :( Shortened version, of what I recall.

I think the issue is `prepareOperand` needs to store more than one analysis per argument, because semanticizing for a formal parameter that's lifted to a procedural isn't the same as say `typed`. So we should do the conversion earlier and store the conversion for the arg in its list of analysed possibilities. Then depending upon which one we choose to dispatch, we pick that analysis. Something like `seq[array[AnalysisKind, PNode]]`, where `seq.len == args.len` and `AnalysisKind` covers, concrete, iterator, meta, typed, procedural, untyped, etc.

The thing is shadow scopes juggling becomes bonkers, we might need to create them as nested based on unique sequence of analyses. So we can clearly discard things.

zerbina:
I just noticed that I had missed removing the TODO regarding the module renaming -- that's fixed now.

saem:
/merge

I think that's everything.

PR #715, opened 2023-05-22
refactor the `collectors` module

zerbina:
## Summary

Simplify and clean up the implementation, and move the module to the
`sem` directory. The latter is preparation for the pass taking on
responsibilities that are more part of semantic analysis than they are
of backend-related processing.

## Details

* rename the `Module` type in `vmbackend` to `BModule`, making the name
  adhere to the naming scheme used by the other backends
* rename the `FullModule` type to `Module`
* restructure the passes's implementation so that a `Module` instance is
  produced when a module is *closed*
* don't require `ModuleList` to inherit from `RootObj`

### DIFF
--- a/compiler/backend/cbackend.nim
+++ b/compiler/backend/cbackend.nim
@@ -12,7 +12,6 @@ import
   compiler/backend/[
     cgen,
     cgendata,
-    collectors,
     extccomp
   ],
   compiler/front/[
@@ -20,6 +19,9 @@ import
   ],
   compiler/modules/[
     modulegraphs
+  ],
+  compiler/sem/[
+    collectors
   ],
   compiler/utils/[
     containers,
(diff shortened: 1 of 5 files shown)


saem (approved):
Nice

saem (approved):
Nice

saem:
Hmm, I'll have to take a closer look at the error message, glanced quickly on my phone and thought it might have been spurious.

zerbina:
It seems to be the C megatest issue again (the test succeeds in my fork), although in a surprisingly persistent incarnation this time around.

I've already restarted the job once (it failed), and now did so again. Should it fail again, we probably need to look into the spurious C megatest failures on MacOS first.

saem:
Hmm, I might be able to get my hands on a mac, that should make it easier to debug locally, hopefully.

PR #716, opened 2023-05-23
error on distinct or tuple type recursion

### LINKED ISSUE #148: Self-referencing distinct will cause the compiler to hang
### Example

```nim
type Id = distinct Id
```

### Actual Output
The compiler stuck at 100% CPU usage

### Expected Output
Some errors about self-referencing

saem:
## Summary

Detect type definition recursion for value and container types and raise
an error, fixing an infinite loop in some cases such as `distinct` types.

## Details

Add `aliasesType` predicate in `semtypes` that can check types
for self-aliasing. Then various `sem*` procedures use this to generate
errors when they occur.

Generic and constrained generic overload disambiguation is broken,
marked failing test in `toverload_issues` as a knownIssue.

Fixes https://github.com/nim-works/nimskull/issues/148


compiler/sem/semtypes.nim
```diff
@@ -1969,6 +1992,10 @@ proc semTypeNode(c: PContext, n: PNode, prev: PType): PType =
         else:
           result = if op.id == ord(wAnd): makeAndType(c, t1, t2)
                    else: makeOrType(c, t1, t2)
+          if prev != nil and
```
saem: I gotta figure this out because the logic of this `if` doesn't make sense.

compiler/sem/semtypes.nim
```diff
@@ -232,12 +228,32 @@ proc semVarOutType(c: PContext, n: PNode, prev: PType; kind: TTypeKind): PType =
...
+    return false
+  if cycleDetector.containsOrIncl(t.id):
+    true
+  else:
+    case t.kind
+    of tyAlias, tyGenericInst, tyDistinct:
+      isRecursiveType(t.lastSon, cycleDetector)
+    else:
+      false
+
```
zerbina: Given how it's currently used, this can be simplified to something like:
```suggestion
proc isReachable(t: PType, from: PType) =
  var curr = from
  while curr != t and curr.kind in {tyAlias, tyGenericInst, tyDistinct}:
    curr = curr.lastSon

  result = curr == t
```
This does require that self-recursive alias and distinct types weren't introduced earlier, however.

compiler/sem/semtypes.nim
```diff
@@ -494,7 +504,13 @@ proc semAnonTuple(c: PContext, n: PNode, prev: PType): PType =
...
   result = newOrPrevType(tyTuple, prev, c)
   for it in n:
-    addSonSkipIntLit(result, semTypeNode(c, it, nil), c.idgen)
+    let t = semTypeNode(c, it, nil).skipIntLit(c.idgen)
+    var cycleDetector = initIntSet()
+    if isRecursiveType(t, cycleDetector):
+      result.add t
+      c.config.localReport(n.info, reportTyp(rsemIllegalRecursion, t))
+    else:
+      rawAddSon(result, t)
```
zerbina: The general direction of the suggestion above also applies here.

compiler/sem/semtypes.nim
```diff
@@ -1717,21 +1732,27 @@ template modifierTypeKindOfNode(n: PNode): TTypeKind =
...
+    cycleDetector = initIntSet()
   result.sons = @[candidateTypeSlot]
   result.n = n
 
   if inherited.kind != nkEmpty:
     for n in inherited.sons:
-      result.add semTypeNode(c, n, nil)
+      let t = semTypeNode(c, n, nil)
+      if isRecursiveType(t, cycleDetector):
+        c.config.localReport(n.info, reportTyp(rsemIllegalRecursion, t))
```
zerbina: It's not strictly necessary to check for `t == prev` here, as `t` is not added via `rawAddSon`.

compiler/sem/semtypes.nim
```diff
@@ -232,12 +228,32 @@ proc semVarOutType(c: PContext, n: PNode, prev: PType; kind: TTypeKind): PType =
...
   if n.len == 0: return newConstraint(c, tyDistinct)
   result = newOrPrevType(tyDistinct, prev, c)
-  addSonSkipIntLit(result, semTypeNode(c, n[0], nil), c.idgen)
+  let t = semTypeNode(c, n[0], nil).skipIntLit(c.idgen)
+  var cycleDetector = initIntSet()
+  if isRecursiveType(t, cycleDetector):
+    result.add t
+    c.config.localReport(n.info, reportTyp(rsemIllegalRecursion, t))
+  else:
+    rawAddSon(result, t)
```
```suggestion
  let t = semTypeNode(c, n[0], nil).skipIntLit(c.idgen)
  if prev != nil and isReachable(prev, t):
    result = newOrPrevType(tyError, prev, c)
    c.config.localReport(n.info, reportTyp(rsemIllegalRecursion, t))
  else:
    result = newOrPrevType(tyDistinct, prev, c)
    rawAddSon(result, t)
```
Assuming the introduction of the `isReachable` from above. The earlier `result` assignment needs to be removed for this to work. Creating a `tyError` makes sure that no cyclic type is introduced in the case of `nimsuggest` and `nim check`.

zerbina (commented):
Sorry that I haven't gotten around to taking a more in-depth look earlier.

In general, the full recursion check can only happen after all forwarded types are resolved, which is not yet the case during the various type node analysis routines. The third type section pass (`semstmts.typeSectionFinalPass`) already performs the recursion check, but it currently skips type classes, see `semtypinst.checkConstructedType`.

In `semDistinct`, etc. the only thing you *should* need to check for is that the added child type is not the same as `prev` (direct self-recursion). I *think* you could get around even that by moving the `propgateToOwner` call in ` rawAddSon` to before the `father.sons.add(son)`.

compiler/sem/semtypes.nim
saem: 'reachablility' is also what we're really testing here, and it's not always recursion (occurence of type application) because things like alias and forward are not instances of type application.

compiler/sem/semtypes.nim
saem: Couldn't one do `Foo = object of Foo` in which case we'd have the same `tyForward`?

zerbina: Yep, that case is possible, but the recursion isn't directly fatal and already rejected in `typeSectionFinalPass`.

Nonetheless, an `inheritance only work with non-final object` error is currently produced with your example, as `semObjectNode` doesn't properly consider `tyForward`.

compiler/sem/semtypes.nim
zerbina: Oops, sorry. The comments got reordered when submitting the review. [This](https://github.com/nim-works/nimskull/pull/716#discussion_r1207140928) is the suggestion I was referring to.

compiler/sem/semtypes.nim
saem: Updated, thanks again for the insight.

compiler/sem/semtypes.nim
saem: Updated

compiler/sem/semtypes.nim
saem: Updated

zerbina:
@saem Do you still want to update the PR message?

compiler/sem/semtypes.nim
zerbina: Naming it `aliasesType` like you did is probably the better choice.

saem:
> @saem Do you still want to update the PR message?

Whoops, I forgot to do that

PR #717, opened 2023-05-23
fix compiler debug tracing

saem:
Compiler debug tracing wasn't working because, a `DebugReport` instead
of a `SemTraceReport` was being submitted, meaning the compiler would
fail on an assert. This is yet another design flaw of legacy reports,
the API does not catch this type of error at compile time.

Added instrumentation to `semAllTypeSections`.


zerbina (approved):
Looks good to me.

Maybe mention that `semAllTypeSections` is now instrumented in the PR message, but that's not a blocker.

saem:
Thanks, @zerbina.

PR #719, opened 2023-05-24
fix(bootstrap/cli): `nim -v` shows git hash

iacore:
## Summary

Pass git hash and build date on build correctly. `nim -v` now
shows git hash.

You should see this now (if you build from `./koch.py boot`)
```
❯ bin/nim -v
Nimskull Compiler Version 0.1.0-dev.2+dirty [linux: amd64]

Source hash: dcfd2b92a8e26e40878d45cec50293eacbaff4d1
Source date: 2023-05-24

active boot switches:    ⏎ 
```

## Details

- Read `-d:sourceHash` and `-d:sourceDate` properly.
- `writeVersionInfo` and friends now doesn't quit the program
- Rename the `nimUseLinenoise` boot define to `useLinenoise`
- Remove the `useGnuReadline` boot define (it was a legacy
  alias for `useLinenoise`)
- Updated the flakey test `tthread_destroy_before_finish.nim`, blocked CI
- Clean up some code


compiler/front/commands.nim
```diff
@@ -371,19 +371,23 @@ type
     os*: TSystemOS ## Target OS
 
 const
-  sourceHash {.strdefine.} = "" # defined by koch
-  sourceDate {.strdefine.} = "" # defined by koch
+  nimSourceHash {.strdefine.} = "" # defined by koch
+  nimSourceDate {.strdefine.} = "" # defined by koch
```
saem:
```suggestion
  nimSourceHash {.strdefine.} = "" # defined by koch
  nimSourceDate {.strdefine.} = "" # defined by koch
  sourceHash = nimSourceHash # HACK: until we can specify define names
  sourceDate = nimSourceDate # HACK: until we can specify define names
```

This way it's a smaller change. Also we're trying to remove `nim` from the source code as part of prep to rename so this way it's less of a regression.

tools/koch/koch.nim
```diff
@@ -125,7 +125,7 @@ proc defineSourceMetadata(): string =
   if hash != "" and date != "":
     result = quoteShellCommand(["-d:nimSourceHash=" & hash, "-d:nimSourceDate=" & date])
   if versionSuffix != "":
-    result &= " -d:CompilerVersionSuffix=" & quoteShell(versionSuffix)
+    result &= " -d:nimCompilerVersionSuffix=" & quoteShell(versionSuffix)
```
saem: Do we need this change, the `nim` prefix should be unnecessary in  the `nversion` module too if we leave it as is?

saem (changes requested):
Thank you very much for taking the time to put together the PR, overall I like the direction.

I've made a few suggestions that can help minimize the change overall and avoid some minor regressions wrt naming.

I had a bit of an emergency today but I'll be able to work with you to get things merged tomorrow (12 hours or so).

tools/koch/koch.nim
iacore: for consistency
I have OCD

compiler/front/commands.nim
```diff
@@ -480,18 +462,22 @@ proc processCmdLine*(pass: TCmdLinePass, cmd: string; config: ConfigRef) =
         # only kept because of user expectations
         expectNoArg(cliFlagVersion, p.key, p.val)
         writeVersionInfo(config)
+        msgQuit(0)
```
zerbina, line 462: It's a small change (i.e., quitting the compiler processing a help or version flag) that doesn't affect much, but please list it in the PR message.

zerbina (changes requested):
Thank you, and welcome!

Please move the changes to `cbackend`, `ccgtypes`, and `vm` into a separate PR, as these are refactorings unrelated to the main change (small unrelated refactorings in modules that are changed anyway are okay, I think).

iacore:
> Thank you, and welcome!
> 
> Please move the changes to `cbackend`, `ccgtypes`, and `vm` into a separate PR, as these are refactorings unrelated to the main change (small unrelated refactorings in modules that are changed anyway are okay, I think).

Those are only code clean up. No change in semantic.

zerbina:
> Those are only code clean up. No change in semantic.

I know, but they're still in modules wholly unrelated to the main part of this PR.

In addition, completely removing the import of `InternalReport` is not correct (the `tcompilerapi.nim` test fails because of it) -- it needs to be wrapped in a `when defined(nimcore)` statement instead.

> * `writeVersionInfo` and friends now doesn't quit the program

Isn't the inverse the case? That is, writing out the version information, etc. now *do* quit the program.

iacore:
> Isn't the inverse the case? That is, writing out the version information, etc. now do quit the program.

It is originally like that.

tools/koch/koch.nim
saem: We're leaving more and more of the `nim` wording behind, IIRC the names were purposefully changed to drop it. At the same time I appreciate the desire for consistency, and that some folks might experience that urge more strongly than others. Reverting these occurances of `nim` and dropping a few additional ones would have it trend in the other direction, and perhaps satisfy your desire for consistency while also contributing to the longer term goal?

iacore: Yes. So I should change `nimSourceHash` to `sourceHash`?

iacore:
I fixed this PR

tools/koch/koch.nim
saem: Yes, please.

compiler/ast/llstream.nim
```diff
@@ -13,7 +13,7 @@ import
   compiler/utils/pathutils
 
 # support `useGnuReadline`, `useLinenoise` for backwards compatibility
-const hasRstdin = (defined(nimUseLinenoise) or defined(useLinenoise) or defined(useGnuReadline)) and
+const hasRstdin = (defined(useLinenoise) or defined(useLinenoise) or defined(useGnuReadline)) and
```
saem: Awesome, thanks.

zerbina:
> > Isn't the inverse the case? That is, writing out the version information, etc. now do quit the program.
> 
> It is originally like that.

Ah, you're right, sorry. I missed that there previously were `msgQuit` calls in `writeVersionInfo` and friends.

---

I've slightly adjusted the PR message to make it adhere to the guidelines, and also added a mention of the `nimUseLinenoise` renaming.

Only the text coming before the first topical break (the three dashes) is makes up the final commit message (the commits from a PR are automatically rebased and squashed into a single one prior to merging), so the summary and details of the PR should go there.

compiler/nim.nim
```diff
@@ -149,7 +149,7 @@ when not defined(selftest):
   of cliErrNoParamsProvided:
     inc conf.errorCounter # causes a non-0 exit, will be replaced soon
     conf.msgWrite("no command-line parameters provided\n", {msgNoUnitSep})
-    conf.showMsg(helpOnErrorMsg(conf))
+    writeUsage(conf)
```
saem, line 152: Nice, now that it no longer quits

saem (approved):
Looks good to me!

compiler/front/commands.nim
```diff
@@ -8,18 +8,26 @@
...
-template bootSwitch(name, expr, userString) =
-  # Helper to build boot constants, for debugging you can 'echo' the else part.
-  const name = if expr: " " & userString else: ""
-
-bootSwitch(usedRelease, defined(release), "-d:release")
-bootSwitch(usedDanger, defined(danger), "-d:danger")
-# `useLinenoise` deprecated in favor of `nimUseLinenoise`, kept for backward compatibility
-bootSwitch(useLinenoise, defined(nimUseLinenoise) or defined(useLinenoise), "-d:nimUseLinenoise")
+#
+# Don't used this other than printing.
```
zerbina:
```suggestion
# Don't use the constant for anything other than printing.
```

compiler/front/commands.nim
```diff
@@ -8,18 +8,26 @@
...
+const bootSwitchEnabled: seq[string] = block:
+  var result: seq[string]
+  
+  template testSwitch(expr, userString) =
+    ## Helper to build boot constants
+    if expr:
+      result &= userString
```
zerbina:
```suggestion
      result.add(userString)
```
Please prefer `add` when appending to sequences.

compiler/front/commands.nim
```diff
@@ -8,18 +8,26 @@
 #
 
 # This module handles the parsing of command line arguments.
-
```
zerbina, line 11: The break introduce by the new line makes sense (the top one is documentation for the module, while the other is for the `bootSwitchEnabled` facility), please restore it.

For improved clarity, you can change line 10 to use `##` (turning it into a doc comment).

iacore:
I cleaned up more stuff

compiler/front/commands.nim
```diff
@@ -7,19 +7,27 @@
...
-  # Helper to build boot constants, for debugging you can 'echo' the else part.
-  const name = if expr: " " & userString else: ""
 
-bootSwitch(usedRelease, defined(release), "-d:release")
-bootSwitch(usedDanger, defined(danger), "-d:danger")
-# `useLinenoise` deprecated in favor of `nimUseLinenoise`, kept for backward compatibility
-bootSwitch(useLinenoise, defined(nimUseLinenoise) or defined(useLinenoise), "-d:nimUseLinenoise")
+## Switches specified when the compiler is built (-d:xxx)
+## 
+## Don't use the constant for anything other than printing.
```
zerbina: As it is now, this is treated as a doc comment attached to the module. For attaching doc comments to constants (or other symbol definitions), it needs to be placed in the line following the definition. Example:

```nim
const c = 1
  ## a doc comment. Notice the indentation

const withBlock = block:
    1
  ## the comment must be at one indentation level higher than the block's content

```

zerbina (commented):
I noticed one last issue with the documentation comment for `bootSwitchEnabled`, but after that, I think this PR is good to go.

I have also updated the PR message with the removal of `useGnuReadline` as an alias for `useLinenoise`.

compiler/front/commands.nim
iacore: ⍤ ⍥ It feels janky placing it at the end of a block

saem: Agreed, it does feel janky, a proper solution likely requires some marker to indicate the associativity of the comment and/or revise the comment association heruistic... but it works this way now so that's the constraint. 😬

I'm down for revising how comment association works (outisde this PR) so this jank can be avoided in the future.

iacore:
@saem you can merge this now

compiler/front/commands.nim
saem: @iacore are you going to change this in a follow-up PR because this will mess with the compiler docs: https://nim-works.github.io/nimskull/compiler/front/commands.html

iacore: I fixed it now. I remembered fixing it.

saem: Thank you very much.

zerbina (approved):

tests/threads/tthread_destroy_before_finish.nim
```diff
@@ -0,0 +1,25 @@
...
+var ok = false
+proc whatever() =
+  ok = true
+
+var thread: Thread[void]
+createThread(thread, whatever)
+thread.coreFn.reset()
+
+# this breaks as well. Simulated: variable going out of scope
+# thread.reset()
```
saem: So in order to get the crash we'd have to uncomment the following line?

tests/threads/tthread_destroy_before_finish.nim
```diff
@@ -0,0 +1,27 @@
+discard """
+  output: ""
+  joinable: false
+  description: '''
+  . if `Thread` is freed before the thread finishes, SIGSEGV
+  . reference counting of `Thread.coreFn` is broken
+'''
```
saem, line 8:
```suggestion
  knownIssue: '''
  . if `Thread` is freed before the thread finishes, SIGSEGV
  . reference counting of `Thread.coreFn` is broken
'''
```

That should mark this tests as a known issue and it won't fail in CI.

iacore: Don't you think `knowIssue: true` is better? Maybe put the reason there

saem: `knownIssue` can take the reason directly to afford a literal style. The test description as written is more a failure description. I'm not super fussed either way, using `knownIssue` seems like the right tact, the exact details aren't super important.

PR #721, opened 2023-05-24
cgen: minor code cleanup

iacore:
Removed unnecessary forward declaration and fixed code formatting, no change in behaviour.

### DIFF
--- a/compiler/backend/cbackend.nim
+++ b/compiler/backend/cbackend.nim
@@ -44,8 +44,11 @@ proc generateCode*(graph: ModuleGraph, mlist: sink ModuleList) =
 
   # setup the module for the generated header, if required:
   if optGenIndex in config.globalOptions:
-    let f = if config.headerFile.len > 0: AbsoluteFile config.headerFile
-            else: config.projectFull
+    let f =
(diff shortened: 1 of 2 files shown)


saem:
Updated the PR description, rather than going into all the types of things a PR/commit message should contain, I thought this would be easier to see via example. @iacore if you're fine with the message as it is or after you've made any further edits let me know and I'll merge queue this for merge.

Thanks again for the PR. ❤️

iacore:
It's fine. You can merge it. I'll write better PR message next time.

iacore:
What is this bot shenanigans

Is the PR title used too?

saem:
Yes the title and everything before the horizontal rule become the commit message. The stuff after the horizontal rule is a space to communicate meta-pr stuff that doesn't fit in the commit message like notes to reviewers. The PR template has some inline explanation of all this. Which also means pr messages need hard line wrapping.

We treat the PR body/commit message as change log entries. So no overly terse tithes and bodies, instead explain what, why, how, etc for others on the project or those looking to learn etc. It's more thoroughly covered in the contribution guide.

PR #722, opened 2023-05-25
std/parseopt: only parses command arg vectors

iacore:
## Summary

The `parseopt` module no longer supports parsing arbitrary strings as
command line parameters. Instead a `seq[string]` representing an argv
(argument vector) must be supplied. In addition to removing the API for
this, it no longer automatically fetches command line parameters.

In addition, parsing empty argument lists and short flags with value was
fixed. In particular, input such as `["-s", "foo"]`, where `"-s"`
expects an argument is correctly treated as `-s:foo` or `-s=foo`.

## Details

As part of this change the `string` accepting variant of the
`initOptParser` procedure was removed, along with the similarly misguided
`getOpts` iterator. Lastly, an empty list no longer triggers argument
retrieval via the `os` module, meaning these are handled correctly.

Modified the compiler and various executables to transition to the
revised API, in order to facilitate this introduced the `os.getExecArgs`
procedure.

### DIFF
--- a/compiler/front/commands.nim
+++ b/compiler/front/commands.nim
@@ -9,7 +9,6 @@
 
 ## This module handles the parsing of command line arguments.
 
-
 # Switches specified when the compiler is built (-d:xxx)
 # 
 # Don't use the constant for anything other than printing.
@@ -17,7 +16,6 @@ const bootSwitchEnabled: seq[string] = block:
   var result: seq[string]
   
   template testSwitch(expr, userString) =
-    ## Helper to build boot constants
     if expr:
       result.add(userString)
 
--- a/compiler/front/commands.nim
+++ b/compiler/front/commands.nim
@@ -415,7 +415,7 @@ proc writeVersionInfo(conf: ConfigRef) =
     commitMsg &
     "\nactive boot switches:" & bootSwitchesMsg
 
-proc processCmdLine*(pass: TCmdLinePass, cmd: string; config: ConfigRef) =
+proc processCmdLine*(pass: TCmdLinePass, cmd: openArray[string]; config: ConfigRef) =
   ## Process input command-line parameters into `config` settings. Input is
   ## a joined list of command-line arguments with multiple options and/or
   ## configurations.
--- a/compiler/front/commands.nim
+++ b/compiler/front/commands.nim
@@ -16,6 +16,7 @@ const bootSwitchEnabled: seq[string] = block:
   var result: seq[string]
   
   template testSwitch(expr, userString) =
+    ## Helper to build boot constants
     if expr:
       result.add(userString)
 
--- a/compiler/nim.nim
+++ b/compiler/nim.nim
@@ -74,14 +74,13 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef, argv: openArray[string]):
     processCmdLine: processCmdLine
   )
   self.initDefinesProg(conf, "nim_compiler")
-  if paramCount() == 0:
+  if argv.len == 0:
     return cliErrNoParamsProvided
 
   self.processCmdLineAndProjectPath(conf, argv)
   if conf.errorCounter != 0: return
   
   let graph = newModuleGraph(cache, conf)
-  if conf.errorCounter != 0: return
   
   if not self.loadConfigsAndProcessCmdLine(cache, conf, graph, argv):
     return
@@ -147,9 +146,7 @@ when not defined(selftest):
     proc(conf: ConfigRef, msg: string, flags: MsgFlags) =
       conf.writeHook(conf, msg & "\n", flags)
 
-  var argv = newSeq[string](paramCount())
-  for i in 0..<paramCount():
-    argv[i] = paramStr(i+1)
+  let argv = getAppArguments()
   case handleCmdLine(newIdentCache(), conf, argv)
   of cliErrNoParamsProvided:
     inc conf.errorCounter # causes a non-0 exit, will be replaced soon
--- a/compiler/nim.nim
+++ b/compiler/nim.nim
@@ -146,7 +146,7 @@ when not defined(selftest):
     proc(conf: ConfigRef, msg: string, flags: MsgFlags) =
       conf.writeHook(conf, msg & "\n", flags)
 
-  let argv = getAppArguments()
+  let argv = getExecArgs()
   case handleCmdLine(newIdentCache(), conf, argv)
   of cliErrNoParamsProvided:
     inc conf.errorCounter # causes a non-0 exit, will be replaced soon
--- a/compiler/nim.nim
+++ b/compiler/nim.nim
@@ -65,7 +65,7 @@ type
     cliErrConfigProcessing
     cliErrCommandProcessing
 
-proc handleCmdLine(cache: IdentCache; conf: ConfigRef): CmdLineHandlingResult =
+proc handleCmdLine(cache: IdentCache; conf: ConfigRef, argv: openArray[string]): CmdLineHandlingResult =
   ## Main entry point to the compiler - dispatches command-line commands
   ## into different subsystems, sets up configuration options for the
   ## `conf`:arg: and so on.
@@ -77,19 +77,21 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef): CmdLineHandlingResult =
   if paramCount() == 0:
     return cliErrNoParamsProvided
 
-  self.processCmdLineAndProjectPath(conf)
+  self.processCmdLineAndProjectPath(conf, argv)
   if conf.errorCounter != 0: return
+  
   let graph = newModuleGraph(cache, conf)
-
-  if not self.loadConfigsAndProcessCmdLine(cache, conf, graph) or
-      conf.errorCounter != 0:
+  if conf.errorCounter != 0: return
+  
+  if not self.loadConfigsAndProcessCmdLine(cache, conf, graph, argv):
     return
+  if conf.errorCounter != 0: return
 
   mainCommand(graph)
   if optCmdExitGcStats in conf.globalOptions:
     conf.logGcStats(GC_getStatistics())
-
   if conf.errorCounter != 0: return
+
   when hasTinyCBackend:
     if conf.cmd == cmdTcc:
       tccgen.run(conf, conf.arguments)
@@ -145,7 +147,10 @@ when not defined(selftest):
     proc(conf: ConfigRef, msg: string, flags: MsgFlags) =
       conf.writeHook(conf, msg & "\n", flags)
 
-  case handleCmdLine(newIdentCache(), conf)
+  var argv = newSeq[string](paramCount())
+  for i in 0..<paramCount():
+    argv[i] = paramStr(i+1)
+  case handleCmdLine(newIdentCache(), conf, argv)
   of cliErrNoParamsProvided:
     inc conf.errorCounter # causes a non-0 exit, will be replaced soon
     conf.msgWrite("no command-line parameters provided\n", {msgNoUnitSep})
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -3508,5 +3508,8 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
     if cmpIgnoreCase(f.name, invalid) == 0: return false
 
 proc getAppArguments*(): seq[string] =
+  ## Returns argument to the running executable
+  ##
+  ## Does not contain the executable path (argv[0]). This is equivalent to argv[1..^1] in C.
   for i in 1..paramCount():
     result.add(paramStr(i))
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -3507,7 +3507,7 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
   for invalid in invalidFilenames:
     if cmpIgnoreCase(f.name, invalid) == 0: return false
 
-proc getAppArguments*(): seq[string] =
+proc getExecArgs*(): seq[string] =
   ## Returns arguments given by the OS to the running process
   ##
   ## Does not contain the executable path (argv[0] in C). This is equivalent to argv[1..^1] in C.
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -3506,3 +3506,7 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
     find(f.name, invalidFilenameChars) != -1): return false
   for invalid in invalidFilenames:
     if cmpIgnoreCase(f.name, invalid) == 0: return false
+
+proc getAppArguments*(): seq[string] =
+  for i in 1..paramCount():
+    result &= paramStr(i)
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -3509,4 +3509,4 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
 
 proc getAppArguments*(): seq[string] =
   for i in 1..paramCount():
-    result &= paramStr(i)
+    result.add(paramStr(i))
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -3508,8 +3508,8 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
     if cmpIgnoreCase(f.name, invalid) == 0: return false
 
 proc getAppArguments*(): seq[string] =
-  ## Returns argument to the running executable
+  ## Returns arguments given by the OS to the running process
   ##
-  ## Does not contain the executable path (argv[0]). This is equivalent to argv[1..^1] in C.
+  ## Does not contain the executable path (argv[0] in C). This is equivalent to argv[1..^1] in C.
   for i in 1..paramCount():
     result.add(paramStr(i))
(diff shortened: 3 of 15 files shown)


saem:
I updated the PR body to conform with the PR template this will ensure the merge dependency note won't get included in the final commit.

lib/pure/os.nim
```diff
@@ -3506,3 +3506,7 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
     find(f.name, invalidFilenameChars) != -1): return false
   for invalid in invalidFilenames:
     if cmpIgnoreCase(f.name, invalid) == 0: return false
+
+proc getAppArguments*(): seq[string] =
```
saem:
```suggestion
proc getCmdArguments*(): seq[string] =
  ## retrieves a list of command-line arguments, excluding the command itself.
```

I think `Cmd` instead of `App` makes more sense for the name as they're effectively command-line or launch arguments.

I also suggested some public docs, as that's a general push to make the code base easier to maintain in the long run.

saem (commented):
This is headed in the right direction (too much work was happening in `parseopt`), once the other PR is merged and updates to get CI green I can do another pass.

With a bit more creativity it might be possible to break the dependency between `os` and `parseopt` modules

lib/pure/os.nim
iacore: `getAppXxx` is nim naming convention

iacore:
I fixed the test. However, the original impl of parseopt is bugged.

Basically, it can parse `["-v ABC"]` (pathological) but not `["-v", "ABC"]` (POSIX standard).

It doesn't seem like the nimskull compiler relies on this behavior (only `-v:ABC` is ever used)

iacore:
**I skipped the test** @saem if that's ok for you

lib/pure/os.nim
```diff
@@ -3506,3 +3506,10 @@ func isValidFilename*(filename: string, maxLen = 259.Positive): bool {.since: (1
...
+  ## Returns argument to the running executable
+  ##
+  ## Does not contain the executable path (argv[0]). This is equivalent to argv[1..^1] in C.
```
saem: Please don't mention C, we're not a transpiler and must not require, nor cater to, knowledge of a second language.

lib/pure/os.nim
saem: Nim is irrelevant, the intent is to establish new and more sensible conventions and iterate towards them in Nimskull.

tests/misc/tparseopt.nim
```diff
@@ -1,4 +1,9 @@
 discard """
+  knownIssue: ''' 
+    parseopt can parse ["-v ABC"] (pathological) but not ["-v", "ABC"] (POSIX standard)
+    This issue hasn't been fixed in Nim
```
saem: Nim is irrelevant, no need to mention it. We're not interested in compatibility.

tests/misc/tparseopt.nim
```diff
@@ -1,4 +1,9 @@
 discard """
+  knownIssue: ''' 
+    parseopt can parse ["-v ABC"] (pathological) but not ["-v", "ABC"] (POSIX standard)
```
saem: This might be because of potential reuse in parsing config files, some weird windows-ism, or someone randomly found it "useful". If it's the last category we simply want to nuke that junk from orbit. If it's something like the first two reasons then that requires a little bit more care.

lib/pure/parseopt.nim
```diff
@@ -192,57 +192,12 @@ proc parseWord(s: string, i: int, w: var string,
...
-  result.key = ""
-  result.val = ""
-
-proc initOptParser*(cmdline: seq[string], shortNoVal: set[char] = {},
-                    longNoVal: seq[string] = @[];
-                    allowWhitespaceAfterColon = true): OptParser =
-  ## Initializes the command line parser.
-  ##
-  ## If `cmdline.len == 0`, the real command line as provided by the
+  ## If `args.len == 0`, the real command line as provided by the
```
saem: This diff is a bit awkward to follow here (thanks GitHub), but we don't do retrieval anymore, right? So this part of the comment is no longer valid.

tests/misc/tparseopt.nim
```diff
@@ -1,4 +1,9 @@
 discard """
+  knownIssue: '''
```
saem: At present this will disable the entire test, so marking it as a known issue isn't going to achieve what you want. Instead we should just establish sensible behavior and have the test and module work that way.

saem (changes requested):
Had a read through, your instinct to separate command argument retrieval and have that in the `os` module are good. I'd lean into those further, there is no compatibility goal with Nim whatsoever; if anything we want to actively break things to get people paying attention rather than yoloing things if they're converting code.

I noticed some comments that can now be removed and the known issue isn't going to have the desired effect (too broad).

lib/pure/os.nim
iacore: It is only used to explain it succinctly using ISO C standards.

iacore: `argv` is given by the OS when launching a process with `exec`. Maybe `getExecArguments` is a better name?

tests/misc/tparseopt.nim
iacore: afaik it's not windows-ism

`parseCmdline` already exists for parsing argument list

tests/misc/tparseopt.nim
iacore: It's hard to fix parseopt for me

iacore:
I rebased this branch to devel. Only remaining task: fix parseopt to pass test

compiler/front/commands.nim
```diff
@@ -9,15 +9,13 @@
...
 
-
 # Switches specified when the compiler is built (-d:xxx)
 # 
 # Don't use the constant for anything other than printing.
 const bootSwitchEnabled: seq[string] = block:
   var result: seq[string]
   
   template testSwitch(expr, userString) =
-    ## Helper to build boot constants
```
zerbina: This seems to be a line erroneously dropped during the rebase?

compiler/nim.nim
```diff
@@ -77,19 +77,21 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef): CmdLineHandlingResult =
   if paramCount() == 0:
```
zerbina: Now that it's possible, please replace `paramCount()` with querying the length of the argument array.

compiler/nim.nim
```diff
@@ -77,19 +77,21 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef): CmdLineHandlingResult =
...
+  
   let graph = newModuleGraph(cache, conf)
-
-  if not self.loadConfigsAndProcessCmdLine(cache, conf, graph) or
-      conf.errorCounter != 0:
+  if conf.errorCounter != 0: return
+  
+  if not self.loadConfigsAndProcessCmdLine(cache, conf, graph, argv):
     return
+  if conf.errorCounter != 0: return
```
zerbina: I believe you added the separate `if conf.errorCounter == 0` lines for uniformity, but please remove them again.

Routines should use early returns as little as possible, and the current direction is for `processCmdLineAndProjectPath` and `mainCommand` to return whether they were successful or not, instead of indicating this via the side-channel `errorCounter`.

In addition, `newModuleGraph` never modifies `errorCounter`, so testing the coutner after the call is redundant and (in my opinion) confusing.

compiler/nim.nim
```diff
@@ -145,7 +147,10 @@ when not defined(selftest):
     proc(conf: ConfigRef, msg: string, flags: MsgFlags) =
       conf.writeHook(conf, msg & "\n", flags)
 
-  case handleCmdLine(newIdentCache(), conf)
+  var argv = newSeq[string](paramCount())
+  for i in 0..<paramCount():
+    argv[i] = paramStr(i+1)
+  case handleCmdLine(newIdentCache(), conf, argv)
```
zerbina: Can `getAppArguments` be used here instead?

lib/pure/parseopt.nim
```diff
@@ -192,67 +192,18 @@ proc parseWord(s: string, i: int, w: var string,
...
   ## Initializes the command line parser.
   ##
-  ## If `cmdline == ""`, the real command line as provided by the
-  ## `os` module is retrieved instead if it is available. If the
-  ## command line is not available, a `ValueError` will be raised.
-  ##
-  ## `shortNoVal` and `longNoVal` are used to specify which options
-  ## do not take values. See the `documentation about these
-  ## parameters<#shortnoval-and-longnoval>`_ for more information on
-  ## how this affects parsing.
```
zerbina: This part is not obsolete, please restore it.

lib/pure/parseopt.nim
```diff
@@ -192,67 +192,18 @@ proc parseWord(s: string, i: int, w: var string,
...
-  ## `os` module is retrieved instead if it is available. If the
-  ## command line is not available, a `ValueError` will be raised.
-  ## Behavior of the other parameters remains the same as in
-  ## `initOptParser(string, ...)
-  ## <#initOptParser,string,set[char],seq[string]>`_.
   ##
   ## See also:
+  ## * `os.getAppArguments` for getting command arguments
   ## * `getopt iterator<#getopt.i,seq[string],set[char],seq[string]>`_
+  # xxx: fix the link above after having proper links in doc comments
```
zerbina:
```suggestion
  ## * `getAppArguments <os.html#getAppArguments>`_ for getting command-line
  ##   arguments
  ## * `getopt iterator<#getopt.i,seq[string],set[char],seq[string]>`_
```
This makes a proper link show up in the generated documentation.

lib/pure/parseopt.nim
```diff
@@ -192,67 +192,18 @@ proc parseWord(s: string, i: int, w: var string,
...
   runnableExamples:
-    var p = initOptParser()
+    var p = initOptParser(getAppArguments())
```
zerbina:
```suggestion
    var p = initOptParser()
```
Only a soft suggestion. With the change, something different is tested than what was previously, but I think that's okay in this case. This would also fix the runnable example failing to compile.

zerbina (changes requested):
I also agree that the direction of decoupling the `parseopt` module from the `os` module is good, but there are various things that I thing still need to be adjusted.

Also, please use a more descriptive name instead of `argv` -- for consistency with pre-existing parameters, `cmd` would be okay, I think. It's not immediately clear that `argv` means `argument vector` to people not familiar with C, and even `argument vector` is unclear when used for anything else besides the main procedure's argument.

zerbina:
@iacore Please turn this PR into a draft. This saves CI resources for the `nim-works/nimskull` repository, as the CI is only run for non-draft PRs.

If you have not disabled the CI for your fork (it's enabled by default, I think), then you can still check for CI failures there. Once you're finished with the PR, and the CI in your fork reports no failures, you can mark the PR as "ready for review".

lib/pure/os.nim
saem: I like `getExecArguments` quite a bit, or even `getExecArgs`.

tests/misc/tparseopt.nim
saem: Sounds like more code to nuke from orbit. 😈 😆

compiler/front/commands.nim
iacore: yes

lib/pure/parseopt.nim
iacore: this won't compile

compiler/nim.nim
iacore: yes

lib/pure/os.nim
```diff
@@ -1210,7 +1210,7 @@ proc symlinkExists*(link: string): bool {.rtl, extern: "nos$1",
 
 
 when not defined(windows):
-  const maxSymlinkLen = 1024
+  const maxSymlinkLen = 4096
```
zerbina: Is this change required for decoupling `initOptParser` from `os`? If not, please make a separate PR for it.

nimsuggest/nimsuggest.nim
```diff
@@ -730,16 +730,17 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef) =
...
   graph.onSymImport = graph.onMarkUsed # same callback
-  if self.loadConfigsAndProcessCmdLine(cache, conf, graph):
+  if self.loadConfigsAndProcessCmdLine(cache, conf, graph, argv):
     # defaulting to customizing for the C backend matches what
     # 'nim check' does
     customizeForBackend(graph, conf, backendC)
     mainCommand(graph)
 
 when isMainModule:
+  let argv = getExecArgs()
```
zerbina, line 740: It can stay like this, but in the future, please pass the result to the parameter directly where possible. That is: `handleCmdLine(..., ..., getExecArgs())`.

compiler/nim.nim
```diff
@@ -74,22 +74,23 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef): CmdLineHandlingResult =
...
   if conf.errorCounter != 0: return
+  
   let graph = newModuleGraph(cache, conf)
-
-  if not self.loadConfigsAndProcessCmdLine(cache, conf, graph) or
-      conf.errorCounter != 0:
+  
+  if not self.loadConfigsAndProcessCmdLine(cache, conf, graph, argv):
     return
+  if conf.errorCounter != 0: return
```
zerbina, line 87: Please also merge this line back into the above `if`.

lib/pure/os.nim
```diff
@@ -3124,7 +3124,7 @@ proc getAppFilename*(): string {.rtl, extern: "nos$1", tags: [ReadIOEffect], noW
   ## * `getAppDir proc <#getAppDir>`_
   ## * `getCurrentCompilerExe proc <#getCurrentCompilerExe>`_
 
-  # Linux: /proc/<pid>/exe
+  # Linux: /proc/self/exe
```
zerbina, line 3127: The same with this change.

compiler/nim.nim
iacore: I think this is clearer though

nimsuggest/nimsuggest.nim
iacore: this is clearer

lib/pure/parseopt.nim
```diff
@@ -192,67 +192,22 @@ proc parseWord(s: string, i: int, w: var string,
...
-  ## If `cmdline.len == 0`, the real command line as provided by the
-  ## `os` module is retrieved instead if it is available. If the
-  ## command line is not available, a `ValueError` will be raised.
-  ## Behavior of the other parameters remains the same as in
-  ## `initOptParser(string, ...)
-  ## <#initOptParser,string,set[char],seq[string]>`_.
-  ##
+  ## 
   ## See also:
+  ## * `getAppArguments <os.html#getAppArguments>`_ for getting command-line
```
saem:
```suggestion
  ## * `getExecArgs <os.html#getExecArgs>`_ for getting command-line
```

saem (commented):
To fix the tests I had to append:

```nim

    if p.idx < p.cmds.len and
       card(p.shortNoVal) > 0 and p.key[0] notin p.shortNoVal and
       p.val == "":
      p.val = p.cmds[p.idx]
      inc p.idx
```
to the last if inside `handleShortOption`, so it'll look like:

```nim
proc handleShortOption(p: var OptParser; cmd: string) =
  var i = p.pos
  p.kind = cmdShortOption
  if i < cmd.len:
    add(p.key, cmd[i])
    inc(i)
  p.inShortState = true
  while i < cmd.len and cmd[i] in {'\t', ' '}:
    inc(i)
    p.inShortState = false
  if i < cmd.len and (cmd[i] in {':', '='} or
      card(p.shortNoVal) > 0 and p.key[0] notin p.shortNoVal):
    if i < cmd.len and cmd[i] in {':', '='}:
      inc(i)
    p.inShortState = false
    while i < cmd.len and cmd[i] in {'\t', ' '}: inc(i)
    p.val = substr(cmd, i)
    p.pos = 0
    inc p.idx
  else:
    p.pos = i
  if i >= cmd.len:
    p.inShortState = false
    p.pos = 0
    inc p.idx
    if p.idx < p.cmds.len and
       card(p.shortNoVal) > 0 and p.key[0] notin p.shortNoVal and
       p.val == "":
      p.val = p.cmds[p.idx]
      inc p.idx
```

compiler/nim.nim
saem: @iacore please restore this

saem:
`release_manager` is still expecting the old API: https://github.com/nim-works/nimskull/blob/4df8b0bbefe08fe2b4d5ef08517b93b89d4bb0bf/tools/release_manifest.nim#LL522C1-L522C1

Should just be a `getExecArgs()`.

compiler/nim.nim
iacore: The semantic is the same. Should we make the code clearer?

saem: It is the same semantics and making the code clearer is good; I'm not sure how splitting them apart does that, maybe you meant another way entirely?

The idea behind restoring the line stems from small PR hygiene and keeping file history clean for those spelunking through trying to debug/understand things (false positive).

saem: Following up in chat to discuss this, I don't think we need to block the PR for this.

saem (approved):

saem:
Need to update the PR summary (what will be the merge commit mesage), going to do that in this instance.

PR #724, opened 2023-05-26
lambdalifting: improve generated environment setup code

zerbina:
## Summary

Use an assignment with an object constructor expression instead of a
call to `internalNew`. This makes the resulting code easier to reason
about for the analysis passes, and the `internalNew` definition in the
`system` module becomes obsolete.

## Details

The csources compiler still needs the `internalNew` symbol, so removing
the definition is only possible after the next csources update.

Because of using an assignment, the environment setup code now looks
like
```nim
:env = Env()
```
instead of
```nim
`=destroy`(:env)
internalNew(:env)
```
after the `injectdestructors` pass has run.

### DIFF
--- a/compiler/sem/lambdalifting.nim
+++ b/compiler/sem/lambdalifting.nim
@@ -148,10 +148,10 @@ const
   paramName* = ":envP"
   envName* = ":env"
 
-proc newCall(a: PSym, b: PNode): PNode =
-  result = newNodeI(nkCall, a.info)
-  result.add newSymNode(a)
-  result.add b
+proc newObjConstr(typ: PType, info: TLineInfo): PNode =
+  ## Creates an object constructor node with no arguments for type `typ`, using
+  ## `info` for source line information.
+  newTreeIT(nkObjConstr, info, typ, [newNodeIT(nkType, info, typ)])
 
 proc createClosureIterStateType*(g: ModuleGraph; iter: PSym; idgen: IdGenerator): PType =
   var n = newNodeI(nkRange, iter.info)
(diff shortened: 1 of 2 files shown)


compiler/sem/lambdalifting.nim
```diff
@@ -148,10 +148,10 @@ const
   paramName* = ":envP"
   envName* = ":env"
 
-proc newCall(a: PSym, b: PNode): PNode =
-  result = newNodeI(nkCall, a.info)
-  result.add newSymNode(a)
-  result.add b
+proc newObjConstr(typ: PType, info: TLineInfo): PNode =
+  ## Creates an object constructor node with no arguments for type `typ`, using
+  ## `info` for source line information.
```
saem:
```suggestion
  ## Creates an object construction node with no arguments for type `typ`,
  ## using `info` for source line information.
```

Very soft suggestion to clarify that it's a construction call/expression as opposed to a constructor definition.

lib/system.nim
```diff
@@ -267,6 +267,7 @@ const ThisIsSystem = true
 
 proc internalNew*[T](a: var ref T) {.magic: "New", noSideEffect.}
   ## Leaked implementation detail. Do not use.
+  # TODO: obsolete; remove after the next csource update
```
saem: hmm, did you want to add a check for sym name in magics? I believe that'll allow detection of whether it was an `internalNew` call and that could be an assert (prevent regressions).

saem (approved):
Nice find, and can confirm that the new assignment based code _is_ easier to read.

lib/system.nim
zerbina: I went with wrapping the definition in a check for whether `refc` is used. That way, the symbol is only available during the first bootstrap iteration.

compiler/sem/lambdalifting.nim
zerbina: Yep, that's better, thank you. I think being precise here is important.

PR #725, opened 2023-05-27
lambdalifting: overhaul the implementation

zerbina:
## Summary

Simplify, refactor, and clean up the `lambdalifting` module and
associated passes. This makes the pass independent from `transf`,
which, besides making it more amendable to further changes, removes
the cyclic import dependency between `lambdalifting` and `transf`.

In addition, various aspects of closures are clarified and multiple bugs
are fixed:
- closing over locals inside for-loop bodies where the body is duplicated
  works; the closed over locals are *not* duplicated
- using a closure closure iterator inside another closure iterator now
  works
- using an inner closure iterator inside another inner routine via the
  `for` syntax no longer causes an internal compiler error

Given the complexity of the rewrite, the above list is very likely not
exhaustive.

## Details

The summary of the rewrite is that the `transf`, lambda-lifting, and
closure iterator transformation passes now have a well-defined order in
which they're executed.

### Comparison

In the abstract, the lambda-lifting pass previously worked as follows:
```nim
proc innerOf(r) = # set of all closure routines defined directly in r

proc lambdalifting(root) =
  # ----- the 'detection' part
  var inner

  proc discover(x) =
    inner += x
    transform(x)
    for it in innerOf(x):
      discover(it)

  for x in innerOf(root):
    discover(x)

  var all = inner + root

  for x in all:
    let lifted # all locals of x closed over by some of its inner
               # routines
    env[x] = objectFrom(lifted)

    if x != root:
      setEnvParam x, env[x.owner]

  # ----- the 'lifting' part
  proc lift(x) =
    # for the body of `x`:
    # - (if `x` has lifted locals) inject the environment setup logic
    # - transform access to lifted locals
    # - transform closure routine usage into a closure construction
    #   expression

  for x in all:
    lift(x)

proc transform(x) =
  # note that the result of `transform` is cached
  if x is iterator or callConv(x) isnot closure: 
    lambdalifting(x)
  transf(x)
  closureiters(x)
```

The pass now works, and is integrated, as follows:
```nim
proc lambdalifting(root) =
  # ----- the 'detection' part
  let
    lifted = # all locals of `root` closed over by some of its inner
             # routines
    env = objectFrom(lifted)

  # ----- the 'lifting' part
  for x in innerOf(root):
    setEnvParam x, env

  lift(root)

proc transform(x) =
  # note that the result of `transform` is cached
  lambdalifting(x)
  transf(x)
  closureiters(x)
```

Note that transforming closure iterator symbol usage into a closure
construction expression no longer happens during `lift`, but is now
always the responsibility of `transf`.

The benefits of the new approach are:
- the passes now strictly happen in the order: `lambdalifting`,
  `transf`, `closureiters`
- the `lambdalifting` pass doesn't depend on `transf`, and each
  invocation only modifies symbols and AST directly owned by the
  provided root
- less recursion, and an overall more linear design

One downside is that nested inner closure routines now need to be
traversed multiple times, instead of each just once.

### Other internal changes:

- add the `names` iterators to `ast_query`
- use `typ.base` instead of `skipTypes` for accessing the environment
  parameters' underlying `object` type
- don't traverse the type slots of identdefs in `detectCapturedVars`
  and `liftCapturedVars`; they don't contain imperative code
- don't modify the bodies of inner routines when duplicating for-loop
  bodies

### Changes to the generated code:

- don't add an unnecessary `:state` field to the environment type for
  non-iterator routines
- don't use a temporary local during closure iterator environment
  setup. In addition to making the code easier to reason about for
  analysis passes (less assignments), no extra local is lifted into
  the environment when the setup code is emitted into a `.closure`
  iterator


compiler/ast/ast_query.nim
```diff
@@ -709,4 +709,11 @@ iterator branchLabels*(node: PNode): (int, PNode) =
   assert node.kind in {nkOfBranch, nkElifBranch, nkElse, nkElifExpr,
                        nkElseExpr, nkExceptBranch}
   for i in 0..<node.len-1:
-    yield (i, node[i])
\ No newline at end of file
+    yield (i, node[i])
+
+iterator names*(node: PNode): PNode =
```
saem, line 714: I could have sworn one of these already existed but now I can't find it. It's a gooder.

compiler/sem/lambdalifting.nim
```diff
@@ -304,77 +275,62 @@ proc freshVarForClosureIter*(g: ModuleGraph; s: PSym; idgen: IdGenerator; owner:
...
+      ## gathering the set of locals that need to be lifted
+
+    processed, marker: IntSet
+    capturedVars: seq[PSym]
+      ## all local variables and parameters that need to be lifted into
+      ## the generated environment
+    usedInner: seq[PSym]
+      ## all used inner closure routines defined directly in root. We cannot
+      ## gather these by simply looking for procdef nodes, as we'd miss
+      ## instances of generics routines then
```
saem:
```suggestion
      ## instances of generic routines then
```

saem (commented):
Only did a partial read today, will focus on this tomorrow. The smaller footprint and dropping the dependency cycle is awesome, so I'm focused on finding improvements and onboard with the goal.

zerbina:
I'm still looking into the why `niminst` generates an incorrect makefile (only the Windows target has any output files specified), but so far haven't been able to even build `niminst` (via `koch csource`). Both the compiler from this PR *and* the devel compiler seem to exit prematurely without any stack-trace (even in debug mode).

saem:
> I'm still looking into the why `niminst` generates an incorrect makefile (only the Windows target has any output files specified), but so far haven't been able to even build `niminst` (via `koch csource`). Both the compiler from this PR _and_ the devel compiler seem to exit prematurely without any stack-trace (even in debug mode).

Not sure if it's the same issue, but I ran `./koch.py --nim:bin/nim_dbg csource --lib:lib -d:danger` and got the trace below. I'm going to look at the source code a little bit to see if there are any clues.

<details>
<pre>
nim.nim                  nim
sem.nim                  semStmtAndGenerateGenerics nkStmtList 1663926 nimskull/tools/niminst/niminst.nim(135, 0)
semexprs.nim             semExpr nkStmtList 1663926 nimskull/tools/niminst/niminst.nim(135, 0)
semstmts.nim             semStmtList nkStmtList 1663926 nimskull/tools/niminst/niminst.nim(135, 0)
semexprs.nim             semExpr nkIncludeStmt 1663927 nimskull/tools/niminst/niminst.nim(136, 0)
semstmts.nim             evalInclude
semstmts.nim             incMod
semexprs.nim             semExpr nkStmtList 1671367 nimskull/tools/niminst/nsis.nimf(2, 0)
semstmts.nim             semStmtList nkStmtList 1671367 nimskull/tools/niminst/nsis.nimf(2, 0)
semexprs.nim             semExpr nkProcDef 1671368 nimskull/tools/niminst/nsis.nimf(2, 0)
semstmts.nim             semRoutineDef nkProcDef 1671368 nimskull/tools/niminst/nsis.nimf(2, 0)
semexprs.nim             semProcBody
semexprs.nim             semExpr nkStmtList 1671380 nimskull/tools/niminst/nsis.nimf(3, 2)
semstmts.nim             semStmtList nkStmtList 1671380 nimskull/tools/niminst/nsis.nimf(3, 2)
semexprs.nim             semExpr nkCall 1672477 nimskull/tools/niminst/nsis.nimf(214, 12)
semexprs.nim             semIndirectOp nkCall 1672477 nimskull/tools/niminst/nsis.nimf(214, 12)
(1874 calls omitted) ...
semcall.nim              semOverloadedCall nkInfix 1672853 nimskull/tools/niminst/nsis.nimf(260, 39)
semcall.nim              resolveOverloads nkInfix 1672853 nimskull/tools/niminst/nsis.nimf(260, 39)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1672853 nimskull/tools/niminst/nsis.nimf(260, 39)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand
semexprs.nim             semOperand
semexprs.nim             semExpr nkInfix 1672855 nimskull/tools/niminst/nsis.nimf(259, 27)
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkInfix 1672855 nimskull/tools/niminst/nsis.nimf(259, 27)
semcall.nim              semOverloadedCall nkInfix 1672855 nimskull/tools/niminst/nsis.nimf(259, 27)
semcall.nim              resolveOverloads nkInfix 1672855 nimskull/tools/niminst/nsis.nimf(259, 27)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1672855 nimskull/tools/niminst/nsis.nimf(259, 27)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand
semexprs.nim             semOperand
semexprs.nim             semExpr nkInfix 1672857 nimskull/tools/niminst/nsis.nimf(258, 9)
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkInfix 1672857 nimskull/tools/niminst/nsis.nimf(258, 9)
semcall.nim              semOverloadedCall nkInfix 1672857 nimskull/tools/niminst/nsis.nimf(258, 9)
semcall.nim              resolveOverloads nkInfix 1672857 nimskull/tools/niminst/nsis.nimf(258, 9)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1672857 nimskull/tools/niminst/nsis.nimf(258, 9)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand
semexprs.nim             semOperand
semexprs.nim             semExpr nkInfix 1672859 nimskull/tools/niminst/nsis.nimf(257, 25)
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkInfix 1672859 nimskull/tools/niminst/nsis.nimf(257, 25)
semcall.nim              semOverloadedCall nkInfix 1672859 nimskull/tools/niminst/nsis.nimf(257, 25)
semcall.nim              resolveOverloads nkInfix 1672859 nimskull/tools/niminst/nsis.nimf(257, 25)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1672859 nimskull/tools/niminst/nsis.nimf(257, 25)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand
semexprs.nim             semOperand
semexprs.nim             semExpr nkInfix 1672861 nimskull/tools/niminst/nsis.nimf(256, 42)
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkInfix 1672861 nimskull/tools/niminst/nsis.nimf(256, 42)
semcall.nim              semOverloadedCall nkInfix 1672861 nimskull/tools/niminst/nsis.nimf(256, 42)
semcall.nim              resolveOverloads nkInfix 1672861 nimskull/tools/niminst/nsis.nimf(256, 42)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1672861 nimskull/tools/niminst/nsis.nimf(256, 42)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand
semexprs.nim             semOperand
semexprs.nim             semExpr nkInfix 1672863 nimskull/tools/niminst/nsis.nimf(255, 9)
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkInfix 1672863 nimskull/tools/niminst/nsis.nimf(255, 9)
semcall.nim              semOverloadedCall nkInfix 1672863 nimskull/tools/niminst/nsis.nimf(255, 9)
semcall.nim              resolveOverloads nkInfix 1672863 nimskull/tools/niminst/nsis.nimf(255, 9)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1672863 nimskull/tools/niminst/nsis.nimf(255, 9)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand
semexprs.nim             semOperand
semexprs.nim             semExpr nkInfix 1672865 nimskull/tools/niminst/nsis.nimf(254, 21)
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkInfix 1672865 nimskull/tools/niminst/nsis.nimf(254, 21)
semcall.nim              semOverloadedCall nkInfix 1672865 nimskull/tools/niminst/nsis.nimf(254, 21)
semcall.nim              resolveOverloads nkInfix 1672865 nimskull/tools/niminst/nsis.nimf(254, 21)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1672865 nimskull/tools/niminst/nsis.nimf(254, 21)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand
semexprs.nim             semOperand
semexprs.nim             semExpr nkInfix 1672867 nimskull/tools/niminst/nsis.nimf(253, 35)
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkInfix 1672867 nimskull/tools/niminst/nsis.nimf(253, 35)
semcall.nim              semOverloadedCall nkInfix 1672867 nimskull/tools/niminst/nsis.nimf(253, 35)
semcall.nim              resolveOverloads nkInfix 1672867 nimskull/tools/niminst/nsis.nimf(253, 35)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1672867 nimskull/tools/niminst/nsis.nimf(253, 35)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand
semexprs.nim             semOperand
semexprs.nim             semExpr nkInfix 1672869 nimskull/tools/niminst/nsis.nimf(252, 44)
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkInfix 1672869 nimskull/tools/niminst/nsis.nimf(252, 44)
semcall.nim              semOverloadedCall nkInfix 1672869 nimskull/tools/niminst/nsis.nimf(252, 44)
semcall.nim              resolveOverloads nkInfix 1672869 nimskull/tools/niminst/nsis.nimf(252, 44)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1672869 nimskull/tools/niminst/nsis.nimf(252, 44)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand
semexprs.nim             semOperand
semexprs.nim             semExpr
dollars.nim              $
digitsutils.nim          addInt
digitsutils.nim          addInt
digitsutils.nim          addInt
digitsutils.nim          addIntImpl
digitsutils.nim          addChars
alloc.nim                alloc0
alloc.nim                alloc
</pre>
</details>

compiler/sem/transf.nim
```diff
@@ -1214,12 +1215,15 @@ proc transformBody*(g: ModuleGraph, idgen: IdGenerator, prc: PSym, body: PNode):
   ## This step is not indempotent, and since no caching is performed, it
```
saem, line 1215: Should `transformBody`'s docs say that it applies the `lambdalifting -> transf -> closureIterator` transforms to the body? Or maybe something about the result will have all standard `transf`orms applied, as well as `lambdalifting` and `closureiter`tator transforms.

The above could all be a bad idea, it's partly inspired by your excellent PR message, snippet:
> - the passes now strictly happen in the order: `lambdalifting`, `transf`, `closureiters`.

compiler/sem/lambdalifting.nim
```diff
@@ -924,3 +954,8 @@ proc liftForLoop*(g: ModuleGraph; body: PNode; idgen: IdGenerator; owner: PSym):
 
   loopBody[1] = ibs
   loopBody[2] = body[^1]
+
+proc finishClosureIterator*(g: ModuleGraph, idgen: IdGenerator, iter: PSym) =
```
saem, line 947: little surprised to see this proc here, but I'm not entirely sold on separating general lifting operations (general transforms and queries) into their own module. It would allow `closureiter` and the pass part of `lambdalifting`to use that as common code.

zerbina:
Thank you, @saem; that helped pushing my investigation towards the issue. Too nested call expression cause stack-space exhaustion, and the way source code filters currently transform the source code leads to very deep call expression nesting for input files with many lines.

I'm going to fix that issue first (it's unrelated to this PR), and after that, further investigate why `niminst` produces incorrect makefiles now.

compiler/sem/lambdalifting.nim
```diff
@@ -558,270 +533,330 @@ proc getUpViaParam(g: ModuleGraph; owner: PSym): PNode =
...
+    # a routine prootype or otherwise empty routine -> nothing to do
+    result = (body, nil)
 
 proc liftLambdasForTopLevel*(module: PSym, body: PNode): PNode =
   # XXX implement it properly
   result = body
 
+proc liftIterSym*(g: ModuleGraph; n: PNode; idgen: IdGenerator; owner, currEnv: PSym): PNode =
+  ## Transforms  ``(iter)``  to  ``(let env = newClosure[iter](); (iter, env))``.
+  ## This cannot happen as part of ``liftLambdas``, as the iterators's
```
saem:
```suggestion
  ## This cannot happen as part of ``liftLambdas``, as the iterator's
```

compiler/sem/lambdalifting.nim
```diff
@@ -558,270 +533,330 @@ proc getUpViaParam(g: ModuleGraph; owner: PSym): PNode =
...
+  if upField != nil:
+    # the iterator has an 'up' field, and we have to initialize it here
+    let access = accessEnv(currEnv, upField.typ, n.info, g)
+    result.add(newAsgnStmt(rawIndirectAccess(vnode, upField, n.info), access, n.info))
+
+  result.add makeClosure(g, idgen, iter, vnode, n.info)
+
+proc ensureEnvParam*(graph: ModuleGraph, idgen: IdGenerator, prc: PSym) =
+  ## Problem: top-level anonymous expression can explicitly use the .closure
+  ## calling convention (which doesn't make sense, really). All .closure
```
saem, line 838: Not that I want to change anything this PR, but if/when we can instantiate modules, top level closures would make sense in that they're closing over a specific module instance's `let/var` and/or the module itself?

compiler/sem/lambdalifting.nim
```diff
@@ -558,270 +533,330 @@ proc getUpViaParam(g: ModuleGraph; owner: PSym): PNode =
...
+      result.env = c.envSym
     else:
-      result = body
-    #if fn.name.s == "get2":
-    #  echo "had something to do ", d.somethingToDo
-    #  echo renderTree(result, {renderIds})
+      # nothing to do
+      result = (body, nil)
+  else:
+    # a routine prootype or otherwise empty routine -> nothing to do
```
saem:
```suggestion
    # a routine prototype or otherwise empty routine -> nothing to do
```

saem (commented):
I'm a fan of less code overall.

---

I'm wondering if there is a further (greater?) simplification to be had?

In my rather simplistic mental model in terms of languages levels, first class iterators are a "higher level" feature than first-class routines. So in my mind, I would imagine the transforms _should_ be in the following order and associated responsibilities:

1. `closureiter` (lower away the first-class iterators)
2. `lambdalifting` (lower away first-class routines)
3. `transf` (lower away sugar)

From what I'm piecing together doing `closureiter` first is tricky because it's often a big ol' redo of `lambdalifting` but slightly different, at least that's how it's realized today. Hence the awkward cyclic dependencies.

I guess the question is can we lower first-class iterators to second-class iterators (inline) and first-class routines (less iterators). If the answer is yes, then lambda lifting can have even less awareness of iterators than it does today and know more about the environment's type sooner, perhaps?

It's a bigger discussion and likely out of scope for this PR, so I'll slap together a gist to kick that off.

compiler/sem/transf.nim
zerbina: Yep, I think they should -- the comment has been adjusted.

compiler/sem/lambdalifting.nim
zerbina: I agree that the procedure is misplaced here, but it's a trade-off. Having it in `closureiters.nim` would require `createTypeBoundOpsLL` to either become exported or to be moved somewhere else (such as `liftdestructors.nim`), which both are options I dislike, given `createTypeBoundOpsLL`'s fundamentally hack-y nature.

zerbina:
I figured out the reason `niminst` generated an incorrect makefile: there's a pre-existing bug with inline iterator inlining that now surfaced because of the changes to pass ordering. The fix can be found [here](#731), and needs to be merged first.

There is also a small improvement I integrated: the generated closure iterator instance setup now looks like `(iter, newClosure[iter]())` instead of `(let env = newClosure[iter](); (iter, env)`. In addition, I have fixed the discovered typos (thanks, @saem) and improved the PR message a bit.

> I'm wondering if there is a further (greater?) simplification to be had?

I agree that there's further simplifications to be had, but my main intention with this PR was to fix the bugs blocking #714 -- the improvements, simplifications, and breaking up the cyclic dependency are really only side-effects.

> From what I'm piecing together doing `closureiter` first is tricky because it's often a big ol' redo of `lambdalifting` but slightly different, at least that's how it's realized today. Hence the awkward cyclic dependencies.

Hm, how do you mean? There's no cyclic dependency between `lambdalifting` and `closureiters`.

The reason `closureiters` in its current shape cannot come before `transf` (and thus `lambdalifting`) is that it depends on the lowerings performed by `transf`. For the `closureiters` transformation to work, only the control-flow *primitives* (`while`, `if`, `case`, `block`) must be present, but prior `transf` having run, that is not the case.

saem:
> > From what I'm piecing together doing `closureiter` first is tricky because it's often a big ol' redo of `lambdalifting` but slightly different, at least that's how it's realized today. Hence the awkward cyclic dependencies.
>
> Hm, how do you mean? There's no cyclic dependency between `lambdalifting` and `closureiters`.

Sorry, I should have been more clear, it's not in the module/code sense this change definitely improved that. There remains some at the conceptual/data level as both `lambdalifting` and `closureiters` are closure iterator aware rather than them being lowered once and then `lambdalifting` lowering away HoF (higher order functions/procedures). `closureiters` shouldn't care about the closure/env.

The awesome thing about this PR is that it's starting to show me a path as to how we might build upon it and take it further; whereas the previous code did the opposite. Apologies, if anything is coming off as negative, if it is that's entirely directed at the legacy code.

> The reason `closureiters` in its current shape cannot come before `transf` (and thus `lambdalifting`) is that it depends on the lowerings performed by `transf`. For the `closureiters` transformation to work, only the control-flow primitives (`while`, `if`, `case`, `block`) must be present, but prior `transf` having run, that is not the case.

Yup, I think that's something to address outside this PR, where the result should be: that the level of transform `closureiters` does drops to a language without closure iterators and only HoF and inline iterators, then it moves ahead of `lambdalifting`. Per the comment from `closureiters` (snippet below), I think it's an legacy design issue due to misguided code reuse.

> The transformation should play well with lambdalifting, however depending
> on situation, it can be called either before or after lambdalifting
> transformation. As such we behave slightly differently, when accessing
> iterator state, or using temp variables. *If lambdalifting did not happen,
> we just create local variables, so that they will be lifted further on.*

I bolded the last bit to indicate that's the _only_ manner in which it should operate: converting to inline iterators + closure routines, without managing the env. I believe that would remove the need for an environment. It's possible _I_ am not conceptualizing things correctly, where I'm thinking the lowerings should be:

1. language without closure iterators
2. language without HoF
3. language without inline iterators/for-loops, etc

But there could be key lowerings missing, incorrect responsibility breakdown, and/or the order of application is wrong.

compiler/sem/lambdalifting.nim
zerbina: Hm, I guess that depends on what we want the `.closure` pragma to mean. Right now, at least for procedures, it acts as a calling convention indicating that the procedure has a hidden parameter through which captured state is accessed.

Since all locations explicitly defined outside of routines are currently globals, they don't need to be captured, so top-level `.closure` procedures don't make sense, as there's nothing that they can possibly capture. In fact, semantic analysis currently reports an error for all top-level named procedures that explicitly use the `.closure` calling.

If we want locations defined outside of top-level to be treated as *locals* (which I think would be a good idea, at least for locations defined in nested scopes), then top-level `.closure` would indeed make sense.

zerbina:
> Per the comment from `closureiters` (snippet below), I think it's an legacy design issue due to misguided code reuse.

Ah, thank you for mentioning that comment -- it is outdated now, and needs to be updated to reflect the current state.

zerbina:
The comment in `closureiters` is adjusted, the regression in `liftForLoop` fixed, and the `devel` branch with the shared state fix is merged. Local testing indicates that `niminst` now works correctly; let's see what the CI says.

saem:
Going to merge this, tests pass, there is greater coverage, and we can always iterate/improve later as required.

@Clyybber feel free to add comments, we can still discuss.

PR #727, opened 2023-05-29
fix(os): rename maxSymlinkLen to maxPathLen, and increase it to 4096

iacore:
`PATH_MAX` on Linux is defined as 4096. I don't know if this is large enough for other platforms.

On Linux, the symlink itself doesn't have a length limit, but syscalls may reject longer paths.

### DIFF
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -1210,7 +1210,7 @@ proc symlinkExists*(link: string): bool {.rtl, extern: "nos$1",
 
 
 when not defined(windows):
-  const maxSymlinkLen = 1024
+  const maxPathLen = 4096
 
 const
   ExeExts* = ## Platform specific file extension for executables.
@@ -1254,11 +1254,11 @@ proc findExe*(exe: string, followSymlinks: bool = true;
         when not defined(windows):
           while followSymlinks: # doubles as if here
             if x.symlinkExists:
-              var r = newString(maxSymlinkLen)
-              var len = readlink(x.cstring, r.cstring, maxSymlinkLen)
+              var r = newString(maxPathLen)
+              var len = readlink(x.cstring, r.cstring, maxPathLen)
               if len < 0:
                 raiseOSError(osLastError(), exe)
-              if len > maxSymlinkLen:
+              if len > maxPathLen:
                 r = newString(len+1)
                 len = readlink(x.cstring, r.cstring, len)
               setLen(r, len)
@@ -1769,11 +1769,11 @@ proc expandSymlink*(symlinkPath: string): string {.noWeirdTarget.} =
   when defined(windows):
     result = symlinkPath
   else:
-    result = newString(maxSymlinkLen)
-    var len = readlink(symlinkPath, result.cstring, maxSymlinkLen)
+    result = newString(maxPathLen)
+    var len = readlink(symlinkPath, result.cstring, maxPathLen)
     if len < 0:
       raiseOSError(osLastError(), symlinkPath)
-    if len > maxSymlinkLen:
+    if len > maxPathLen:
       result = newString(len+1)
       len = readlink(symlinkPath, result.cstring, len)
     setLen(result, len)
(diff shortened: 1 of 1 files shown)


lib/pure/os.nim
```diff
@@ -1210,7 +1210,7 @@ proc symlinkExists*(link: string): bool {.rtl, extern: "nos$1",
 
 
 when not defined(windows):
-  const maxSymlinkLen = 1024
```
saem, line 1213: On macos they're 1024, so my guess is that they're hedging their bets. Althought that might be FS limits and not really OS/syscall limit.

https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits

lib/pure/os.nim
```diff
@@ -1254,11 +1254,11 @@ proc findExe*(exe: string, followSymlinks: bool = true;
...
           while followSymlinks: # doubles as if here
             if x.symlinkExists:
-              var r = newString(maxSymlinkLen)
-              var len = readlink(x.cstring, r.cstring, maxSymlinkLen)
+              var r = newString(maxPathLen)
+              var len = readlink(x.cstring, r.cstring, maxPathLen)
               if len < 0:
                 raiseOSError(osLastError(), exe)
-              if len > maxSymlinkLen:
+              if len > maxPathLen:
```
saem, line 1261: IIRC, for max component length (filename), macos's default FS is 255 utf-8 chars, while btrfs 255 bytes, etc... see wikipedia limits link in previous comment.

I think we should keep to conservative maximums as that might be easier or get fancy with conditional limits based on target platform.

saem (commented):
I can see what you're aiming for, but I not sure it's the right move. I think it'd help if I understood your perspective more. Being upfront my own perspective is fairly maleable in this area, so I'm hoping to improve my own through learning yours. Could you please share your perspective?

---

In the meanwhile, my current perspective is that the `os` module should work more often than not, so it's conservative in what it allows/exposes. We can then have more specialized modules that are a bit more platform specific which push boundaries and expose more capabilities if that's what one desires. Again, that's what I currently _believe_ and there are definitely other ways to do it.

There are other options but the one I'm suggesting keeps the public surface area perdictable/uniform and is easier to work with for new comer or someone just looking to bang something out quickly.

Again, this is debatable in my mind, a "strong opinion, weakly held". But maybe you have an approach as to what `os` should/shouldn't do that's much better than what I'm thinking and we should do that instead.

lib/pure/os.nim
iacore: The "conservative choice" should be as large as possible. It's the size used when copying the path from the OS with `readlink(3)` .

lib/pure/os.nim
iacore: That's the whole path, not a component.

On Linux, the symlink content can be any string, not even a valid file name. It's a text file with a special flag. You can use it to store any data.

alaviss (commented):
`maxSymlinkLen` is probably a misnomer. It's more of "safe" length to allocate an initial buffer for. All code using this fallback correctly and allocate a larger buffer if the initial was not large enough, so less of a fix but rather an enhancement.

I'd recommend not renaming it for now, and I'm not sure if the initial buffer should be that large (1-4KiB) in the first place, though.

iacore:
Sorry, I don't have time for adding your suggestion. Maybe I should just close this issue.

iacore:
I'm back! Can anyone merge this?

alaviss (changes requested):
As noted [previously](https://github.com/nim-works/nimskull/pull/727#pullrequestreview-1696071921), this constant does not limit the maximum symlink length but serve as a hint for the initial buffer allocation.

So I'm not sure if the bump would serve any purpose, other than being slightly more efficient if you frequently deal with symlinks over 1024 bytes in size.

Since the buffer doesn't shrink after allocation (`setLen` with len < capacity only limit the length, and does not shrink the actual buffer), this change would introduce a massive penalty per symlink resolved, so I'd have to reject this.

If you want to push this through, please provide details about problems this change would solve.

PR #729, opened 2023-05-30
std/parseopt2: new exec argument parsing library

iacore:
## Summary

Brand new and simpler implementation of 'parseopt', the new module does
not replace `parseopt` to allow the compiler code base time to migrate.

## Details

The parsing approach takes after posix getOpt, taking on conventions
such as space separated flag arguments and treating all args after `--`
as positional.

The new API is main the `opts` iterator which parse input and produces
`Opt` for each option, and input errors are inline via a field.

Along with the new module is a comprehensive set of tests.

### DIFF
--- a/lib/pure/parseopt2.nim
+++ b/lib/pure/parseopt2.nim
@@ -27,6 +27,12 @@ type
     optShort ## -s
     optLong  ## --long
     optPos   ## positional
+
+  OptError* {.pure.} = enum
+    none
+    missing    ## missing val (after parsing all the args, last opt need val)
+    extraneous ## extra val when optValNone
+
   Opt* = object
     case kind*: OptKind
     of optShort:
@@ -36,17 +42,8 @@ type
     of optPos:
       discard
     val*: string
-
-  OptError* = ref object of CatchableError
-    case kind*: OptKind
-    of optShort:
-      keyShort*: char
-    of optLong:
-      keyLong*: string
-    of optPos:
-      discard
-  OptExtraneousVal* = ref object of OptError
-  OptMissingVal* = ref object of OptError
+    error*: OptError
+      # abuse "feature": the default value is OptError.none
 
   OptValExpectation* = enum
     ## Should I expect this option to have a value or not?
@@ -127,8 +124,9 @@ iterator opts*(argv: openArray[string],
           i.inc
           if c in sep:
             if longVal.get(key, longDefault) == optValNone:
-              raise OptExtraneousVal(kind: optLong, keyLong: key)
-            yield Opt(kind: optLong, keyLong: key, val: arg[i..^1])
+              yield Opt(kind: optLong, keyLong: key, val: arg[i..^1], error: OptError.extraneous)
+            else:
+              yield Opt(kind: optLong, keyLong: key, val: arg[i..^1])
             break process_arg
           else:
             key &= c
@@ -148,6 +146,7 @@ iterator opts*(argv: openArray[string],
           yield Opt(kind: optShort, keyShort: c)
         else:
           if i < arg.len: # if not the last char in this arg
+            # implied: expectation == optValOptional or expectation == optValRequired
             if arg[i] in sep:
               # skip separator (e.g. '=') in "-a=c" if it exist
               i.inc
@@ -163,11 +162,6 @@ iterator opts*(argv: openArray[string],
 
   # throw on missing val
   if partial.isSome:
-    let opt = partial.get
-    case opt.kind
-    of optShort:
-      raise OptMissingVal(kind: optShort, keyShort: opt.keyShort)
-    of optLong:
-      raise OptMissingVal(kind: optLong, keyLong: opt.keyLong)
-    else:
-      doAssert false, "unreachable"
+    var opt = partial.get
+    opt.error = OptError.missing
+    yield opt
--- a/lib/pure/parseopt2.nim
+++ b/lib/pure/parseopt2.nim
@@ -11,7 +11,7 @@
 ## --long val
 ## --long:val
 ## --long=val
-## --long=    (empty val)
+## --long=    (empty value)
 ## -- -positional1 --positional2
 ## 
 ## -    is treated as positional
@@ -23,25 +23,25 @@
 import std/[options, strutils]
 
 type
-  OptKind* = enum
-    optShort ## -s
-    optLong  ## --long
-    optPos   ## positional
+  OptKind* {.pure.} = enum
+    short        ## -s
+    long         ## --long
+    positional   ## positional
 
   OptError* {.pure.} = enum
     none
-    missing    ## missing val (after parsing all the args, last opt need val)
-    extraneous ## extra val when optValNone
+    missing    ## missing value (after parsing all the args, last opt need value)
+    extraneous ## extra value when optValNone
 
   Opt* = object
     case kind*: OptKind
-    of optShort:
+    of short:
       keyShort*: char
-    of optLong:
+    of long:
       keyLong*: string
-    of optPos:
+    of positional:
       discard
-    val*: string
+    value*: string
     error*: OptError
       # abuse "feature": the default value is OptError.none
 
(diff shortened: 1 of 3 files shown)


lib/pure/parseopt2.nim
```diff
@@ -0,0 +1,133 @@
...
+    of optShort:
+      keyShort*: char
+    of optLong:
+      keyLong*: string
+    of optPos:
+      discard
+
+# xxx: trivial impl. remove this once == for case objects is implemented
+proc`==`*(a, b: Opt): bool =
+  a.kind == b.kind and a.val == b.val and (
```
saem: A global a.val vs b.val check precludes any optimization where reusing an `Opt` which was previously something else and now an `optPos` won't equal another `optPos` that was cleanly initialized.

Not the end of the world, just noticed it.

lib/pure/parseopt2.nim
```diff
@@ -0,0 +1,133 @@
...
+    case a.kind:
+    of optShort:
+      a.keyShort == b.keyShort
+    of optLong:
+      a.keyLong == b.keyLong
+    of optPos:
+      true
+  )
+
+iterator opts*(argv: openArray[string], shortHasVal: openArray[char] = [], longHasVal: openArray[string] = [], sep: set[char] = {'=', ':'}): Opt =
```
saem:
```suggestion
const defaultSeparators* = {'=', ':'}

iterator opts*(argv: openArray[string], shortHasVal: openArray[char] = [], 
               longHasVal: openArray[string] = [],
               sep: set[char] = defaultSeparators): Opt =
```

The constant opens it up for querying/safely wrapping the routine by the end programmer. I also adjusted for line wrapping, although perhaps premature.

---

RE: `shortHasVal` and `longHasVal`

The reason why negation was chosen is because the space of opt without values is typically closed, while ones with is open. For example: `--hint[Foo]:off`, here we can have potentially 10s/100s of variants of `--hint[xxx]`. By contrast, ones that don't take values tend to be things like `-v|--version` and typically they're not open.

Arguably, we might want to encourage interfaces like `--hint:Foo=off`, where we split on the first `:` and the programmer then handles `Foo=off` maybe via a provided simple `func parseAssignment`, which returns a long/short `Opt` for "simple-compound-strings" like the above.

lib/pure/parseopt2.nim
```diff
@@ -0,0 +1,133 @@
...
+## --long=    (empty val)
+## -- -positional1 --positional2
+## 
+## -    is treated as positional
+## --   skipped once; everything after this is treated as positional
+## ---  is treated as long arg "-"
+## 
+## -:   undefined behavior (separator after dash)
+
+import options, strutils
```
saem:
```suggestion
import std/[options, strutils]
```

We're tending towards qualifying imports with their package, trying to figure out exactly how but want to remove fs path based imports.

lib/pure/parseopt2.nim
```diff
@@ -0,0 +1,133 @@
...
+          break
+    else: # process positional
+      yield Opt(kind: optPos, val: arg)
+
+  # throw on missing val
+  if partial.isSome:
+    let opt = partial.get
+    case opt.kind
+    of optShort:
+      raise OptMissingVal(kind: optShort, keyShort: opt.keyShort)
```
saem: I'd opt (hah!) to have these be inline errors maybe a `Opt(kind: optError, ...)`, or using an intermediate result type; this is user input so errors are expected rather than exceptional. If someone is dying for it we can add a second iterator that wraps the first which raises.

From the programmer's perspective this allows them to more easily implement forgiving parsing and/or recovery in the face of errors; in other words it composes better.

tests/stdlib/tparseopt2.nim
```diff
@@ -0,0 +1,58 @@
+discard """
+"""
+
+import std/[unittest, parseopt2, sequtils]
+
+template checkopts(args; expected) =
+  check(toSeq(args) == expected)
+
```
saem, line 8:
```suggestion

template shortOpt(key: char, value = "") = Opt(kind: optShort, keyShort: key, val: value)
template longOpt(key: string, value = "") = Opt(kind: optLong, keyLong: key, val: value)

```

Using these should make the tests easier to follow.

tests/stdlib/tparseopt2.nim
```diff
@@ -0,0 +1,58 @@
...
+## test edge cases
+
+checkOpts opts(["-c", "c_val"], shortHasVal="c"),
+  @[Opt(kind: optShort, keyShort: 'c', val: "c_val")]
+checkOpts opts(["-c:c_val"], shortHasVal="c"),
+  @[Opt(kind: optShort, keyShort: 'c', val: "c_val")]
+checkOpts opts(["-c=c_val"], shortHasVal="c"),
+  @[Opt(kind: optShort, keyShort: 'c', val: "c_val")]
```
saem:
```suggestion
checkOpts opts(["-c=c_val"], shortHasVal="c"),
  @[Opt(kind: optShort, keyShort: 'c', val: "c_val")]
checkOpts opts(["-cc_val", "-j4"], shortHasVal=['c', 'j']),
  @[Opt(kind: optShort, keyShort: 'c', val: "c_val"), Opt(kind: optShort, keyShort: 'j', val: "4")]
```

Should this last check be supported?

lib/pure/parseopt2.nim
```diff
@@ -0,0 +1,133 @@
...
+## -    is treated as positional
+## --   skipped once; everything after this is treated as positional
+## ---  is treated as long arg "-"
+## 
+## -:   undefined behavior (separator after dash)
+
+import options, strutils
+
+type
+  OptKind* = enum
```
saem: `./foo --whatever ''`  is valid, so it might be worth differentiating long and short with and without values?

Presently the end programmer can't differentiate the result of an `Opt` with an optional value receiving and empty value or none at all.

An alternative approach would be a flag or the ability to query position (how much has been consumed) so the information can be reconstructed.

saem (commented):
This is at least a million times better.

Left a few suggestions for some improvements.

Key items:
- the inclusive vs exclusive approach wrt to long/short opts and values is a pretty important API choice
- potential functionality we may want is support for `-j4` style short args with values
- error values instead of exceptions

lib/pure/parseopt2.nim
iacore: The user should know, because `shortWithVal` and `longWithVal`.

lib/pure/parseopt2.nim
iacore: Sum types in this language is painful. In Rust I would use `Result<T, E>`.

The Nim convention is to use exceptions. (e.g. `parseInt`)

tests/stdlib/tparseopt2.nim
iacore: yes

tests/stdlib/tparseopt2.nim
iacore: they are not necessary though. The output of `check(...)` shows the actual structure, so using shorthands might be more confusing when comparing with error output.

lib/pure/parseopt2.nim
saem: Apologies if I'm being slow, but given an long/short with an optional argument, putting it into `shortWithVal` or `longWithVal` would result in an exception. That would then have to be recovered just to handle an optional value. :/

lib/pure/parseopt2.nim
saem: Counter-point, Nim is wrong. 😆 

We have `Result` (https://github.com/nim-works/nimskull/blob/devel/lib/experimental/results.nim) or this could also be handled as in-band data in the enum itself. Since this is an iterator, basically a stream of results, inband might be a better fit.

lib/pure/parseopt2.nim
iacore: the library doesn't have support for optional argument

saem: We have a bunch in the compiler code base, basically any `on|off` flags treat no value as default "on", it'd be a shame to lose that.

saem:
@iacore next week, ideally this weekend, imma try and use this for the compiler proper.

I intend to create a branch off this one and see what reworking it feels like. That'll be a good practical exercise for the API. Based on that we can adjust, as necessary, the APIs and/or revise the compiler interface (most likely a combination of the two).

If you'd like you can jump ahead and give it a whirl yourself.

lib/pure/parseopt2.nim
iacore: this is added in latest commits

saem:
@iacore please update the PR title and body, as that's what will become the commit message in `devel` following the commit message guidelines -- the PR template itself has some pointers.

Otherwise it should be good to merge.

saem:
Updated the PR message in the following ways:

- what changed (short)
- some reasoning as to why/how it's better
- what changed (details), targeting other devs, change log author, or a curious onlooker

PR #731, opened 2023-05-30
transf: fix inlining iterators into `closure` iterators

zerbina:
## Summary

Don't associate the source symbol with the generated environment field
when creating creating fresh variables for the locals of inlined
iterators.

Previously, using the same `.inline` iterator multiple times inside a
`.closure` iterator led to each inlined iterator body effectively
sharing the state of its locals with that of the others. This is now
fixed.

## Details

Only *top-level* closure iterators were affected. The reason being
that for *top-level* closure iterators, the `transf` pass (which is
responsible for inlining `.inline` iterators) is run *after* the
`lambdalifting` pass (which is responsible for lifting locals).

When inlining the `var`/`let` section, `freshVarForClosureIter` is
called, and while that *does* always create a new field, the subsequent
lookup-by-itemId returns the field first associated with the passed-in
symbol.

`freshVarForClosureIter` now creates a field that has no association
with the symbol of a local, fixing the erroneous sharing.


saem (approved):
Very cool!

I wonder if there are work arounds for this lying around in `sequtils` and other parts of the stdlib? I'm going to do a quick look.

PR #733, opened 2023-06-01
backend: closure iterators as lowering

saem:
## Summary
* what changed and how?
* why are we changing it?

## Details
* anything else

Fixes full_issue_url

---

## Notes for Reviewers
* working based off of these notes: https://gist.github.com/saem/62c16b17749378247d4071af0af4e757


saem:
this should be attempted after a new sem architecture has been put into place.

PR #734, opened 2023-06-01
generate method dispatchers prior to code-gen

zerbina:
## Summary

Move the method dispatcher generation out of the code generators (`cgen`
and `jsgen`) and into the orchestrators. This is another step towards
unifying the backend processing.

Methods are now also subject to dead-code elimination, meaning that if
none of the methods attached to an object hierarchy are called, no code
will be generated for them.

## Details

The dispatchers do not appear in calls prior to `transf`, so the DCE
implementation used by IC backend cannot analyze them. For this reason,
`cgen` continues to special-case methods.

In addition, remove the unused `objHasKidsValid` enum value.

### Future Direction

Given that all backends need to generate method dispatchers, this step
(lowering methods into procedures) should happen through some common
facility in the future.

### DIFF
--- a/compiler/backend/cbackend.nim
+++ b/compiler/backend/cbackend.nim
@@ -12,6 +12,7 @@ import
   compiler/backend/[
     cgen,
     cgendata,
+    cgmeth,
     extccomp
   ],
   compiler/front/[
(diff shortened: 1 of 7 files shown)


saem:
FYI, I made a few minor grammar edits to PR body.

PR #735, opened 2023-06-03
boot: update the csources version

zerbina:
## Summary

Require the latest csources version. In addition, remove support
for using NimSkull compilers without proper `--gc:orc` support
during the first bootstrap iteration.

## Details

The `orc` CI job is now redundant (all bootstrap iterations are
built with ORC enabled), and thus removed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,31 +185,6 @@ jobs:
         if: failure()
         run: bin/nim r tools/ci_testresults
 
-  orc:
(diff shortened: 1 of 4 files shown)


saem:
Awesome, it bootstraps successfully. That means we can merge in sources PR.

saem (changes requested):
New csources are up, ready to update this PR to those ones.

saem (approved):
lgtm

PR #738, opened 2023-06-05
cgen: fix `reset` and `default` for resolved concepts

zerbina:
## Summary

Skip user type-classes when generating code for `reset` and `default`,
fixing semantically invalid C code being generated when the operand is
a resolved user-type-class.

## Details

When not skipped, the branch that assigned `0` was chosen for strings
and seqs. While skipping `tyUserTypeClasses` fixes the bug, it's
fundamentally a workaround; resolved user-type-classes should not reach
the code generators in the first place.


saem (approved):
It's fine for a work around.

Just curious if you got a chance to try having the `mDefault` branch in `magicsAfterOverloadResolution` to skip `tyTypeClasses` prior to fetching the base type?

PR #739, opened 2023-06-05
cgen: don't emit unnecessary case statements

zerbina:
## Summary

Fix a logic bug in `cgen`'s object-init specialization that led to
potentially large but empty C case statements being emitted.

While the bug had no effect on program semantics, it did cause more C
code to be emitted, which translated to more work for both the NimSkull
and C compiler.

## Details

The predicate was the inverse of `isObjLackingTypeField`, but that is
incorrect. `isObjLackingTypeField` also returns false for anything
besides `tyObject`, so in effect, all objects that had fields of
non-object types were erroneously treated as having a type field
somewhere.

Also adjust stale documentation related to object-init specialization.

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -1153,7 +1153,7 @@ proc specializeInitObjectN(p: BProc, accessor: Rope, n: PNode, typ: PType) =
 proc specializeInitObject(p: BProc, accessor: Rope, typ: PType,
                           info: TLineInfo) =
   ## Generates type field (if there are any) initialization code for a
-  ## loaction of type `typ`, where `accessor` is the path of the
+  ## location of type `typ`, where `accessor` is the path of the
   ## location.
   if typ == nil:
     return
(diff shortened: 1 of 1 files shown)


compiler/backend/ccgexprs.nim
```diff
@@ -1152,9 +1152,9 @@ proc specializeInitObjectN(p: BProc, accessor: Rope, n: PNode, typ: PType) =
 
 proc specializeInitObject(p: BProc, accessor: Rope, typ: PType,
                           info: TLineInfo) =
-  ## Generates type field (if there are any) initialization code for an
-  ## variable of `typ` at `accessor`. Specialiaztion for the run-time
-  ## `objectInit` function (that's also only available for the old runtime)
+  ## Generates type field (if there are any) initialization code for a
+  ## loaction of type `typ`, where `accessor` is the path of the
```
saem:
```suggestion
  ## location of type `typ`, where `accessor` is the path of the
```

saem (approved):
Looks good; very minor typo but nbd.

--- 

might be too brittle, but perhaps a test with object initialization test with a `ccodecodecheck` spec, for empty case stmts... that might be a crappy peg to write. Not sure if it's worth it.

zerbina:
> Not sure if it's worth it.

I think I'd skip it for now. In theory, a unit test that directly invokes `specializeInitObject` and then checks its output could work, although setting up all of the contextual state required would make up the major part of test.

PR #740, opened 2023-06-05
codegen: clean up unused code paths

zerbina:
## Summary

Remove unused code paths in `cgen`, `jsgen`, and `vmgen`.

Besides the general benefits that removing unused code paths has, this
also helps with preventing regressions caused by lowerings not being
properly applied, at least until the the code generators are rewritten.

## Details

The AST shapes for which the code paths are removed are:
- `when nimvm` statements
- `if`, `case`, `block`, and `try` *expressions*
- multi-branch `if` statements
- `while x` statements where `x` is something else than a `true` boolean
  literal
- pragma blocks and lambda-expressions
- enum field symbols
- `and`/`or` expressions
- deconstruction in `var`/`let` statements (i.e. `nkVarTuple`) and
  initialization of lifted locals (e.g., `var env.x = 0`)

Except for `skEnumField` symbols (which are lowered away during
`transf`), all of the above are lowered during the MIR phase
already, and never reach code generation.

There are two small changes to the output of the code generators:
`jsgen` no longer emits a redundant `if (!true) break label;` statement
for `while` loops, and `cgen` doesn't lower `if` into gotos anymore. An
example of the latter:
```c
// `if cond: then` now produces: 
if (cond)
{
  then;
}
// instead of:
if (!cond) goto label;
then;
label:
```

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -2346,10 +2346,8 @@ proc expr(p: BProc, n: PNode, d: var TLoc) =
   of nkBlockExpr, nkBlockStmt: genBlock(p, n, d)
   of nkStmtListExpr: genStmtListExpr(p, n, d)
   of nkStmtList: genStmtList(p, n)
-  of nkIfExpr, nkIfStmt: genIf(p, n, d)
+  of nkIfStmt: genIf(p, n)
   of nkWhen:
-    # This should be a "when nimvm" node.
-    expr(p, n[1][0], d)
   of nkObjDownConv: downConv(p, n, d)
   of nkObjUpConv: upConv(p, n, d)
   of nkChckRangeF, nkChckRange64, nkChckRange: genRangeChck(p, n, d)
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -2343,7 +2343,7 @@ proc expr(p: BProc, n: PNode, d: var TLoc) =
   of nkDerefExpr, nkHiddenDeref: genDeref(p, n, d)
   of nkDotExpr: genRecordField(p, n, d)
   of nkCheckedFieldExpr: genCheckedRecordField(p, n, d)
-  of nkBlockExpr, nkBlockStmt: genBlock(p, n, d)
+  of nkBlockStmt: genBlock(p, n)
   of nkStmtListExpr: genStmtListExpr(p, n, d)
   of nkStmtList: genStmtList(p, n)
   of nkIfStmt: genIf(p, n)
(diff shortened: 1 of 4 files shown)


compiler/backend/ccgexprs.nim
```diff
@@ -2461,7 +2391,6 @@ proc expr(p: BProc, n: PNode, d: var TLoc) =
      nkFromStmt, nkTemplateDef, nkMacroDef, nkStaticStmt:
     discard
   of nkPragma: genPragma(p, n)
-  of nkPragmaBlock: expr(p, n.lastSon, d)
```
saem, line 2464: Would it be reasonable to have a `const disallowedExpr = {nkPragmaBLock, ... #[the other items that were remove]#}` in a shared module between the backends and then mark them as unreachable here? Moving towards exhaustiveness?

compiler/backend/ccgstmts.nim
```diff
@@ -763,7 +673,7 @@ proc bodyCanRaise(p: BProc; n: PNode): bool =
       if bodyCanRaise(p, n[i]): return true
     result = false
 
-proc genTryGoto(p: BProc; t: PNode; d: var TLoc) =
```
saem, line 766: Oh this is fantastic, we're not screwing around with `TLoc`s nearly as much! IIRC, there is a bunch of unfortunate AST mutation surrounding those that precluded moving towards a sealable/cacheable AST.

saem (commented):
Very nice ratio of adds/deletes. :D

I found a few more potential cases:
- `nkCall, nkHiddenCallConv, nkInfix, nkPrefix, nkPostfix, nkCommand, nkCallStrLit` in `ccgenexprs` on line 2352-ish
- `nkCommand, nkHiddenCallConv, nkCallStrLit` in `jsgen` for the `nodeKindsNeedNoCopy` const

That might reveal another cascade of removals, but I didn't see those nodes kinds being generated in `mir/astgen`; hopefully I didn't misread that.

compiler/backend/ccgexprs.nim
zerbina: Yep, good call -- I'm going to do that.

zerbina:
@saem Are you okay with me making the case statements in the code generators exhaustive in a follow-up PR? I've adjusted them already, but think that the change would fit better as a separate PR.

saem:
> @saem Are you okay with me making the case statements in the code generators exhaustive in a follow-up PR? I've adjusted them already, but think that the change would fit better as a separate PR.

absolutely.

zerbina:
/merge

Thanks for the review, @saem. I've removed the `nkInfix`, etc. usages you noticed, and will make a follow-up PR for making the case statements exhaustive once this PR is merge.

PR #741, opened 2023-06-06
codegen: make the main case statements exhaustive

zerbina:
## Summary

Introduce a set (`codegenExprNodeKinds`) with all node kinds that can appear
in *general* expression and statement contexts during the code
generation phase. The set is then used to make the `case` statement in
the main AST traversal procedures exhaustive (no more catch-all `else`
branch).

## Details

The set doesn't include node kinds (such as `nkExprColonExpr` or
`nkElse`) that can only appear in nested contexts.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -350,7 +350,7 @@ const
     nkUsingStmt,
   }
 
-  codegenNodeKinds* = {
+  codegenExprNodeKinds* = {
     nkEmpty,
     nkSym,
     nkType,
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -350,6 +350,50 @@ const
     nkUsingStmt,
   }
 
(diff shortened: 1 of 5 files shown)


compiler/ast/ast_types.nim
```diff
@@ -350,6 +350,50 @@ const
     nkUsingStmt,
   }
 
+  codegenNodeKinds* = {
```
```suggestion
  codegenExprNodeKinds* = {
```

Very soft suggestion motivated from the PR description, as it only includes "top level" expression forms, and not some of the ones that can be nested within (eg: `nkExprColonExpr`, etc)?

saem (approved):
Very soft suggestion, ignorable.

Thanks for setting this up, the code gen feels easier to navigate already with the explicit reduction in possible input.

compiler/ast/ast_types.nim
zerbina: Yep, that's better.

PR #743, opened 2023-06-07
cgen: remove more obsolete code

zerbina:
## Summary

Remove leftover obsolete code from `cgen`:
* the `earlyResultAsgn` procedure
* the `nfPreventCg` node flag
* the `lfNoDeepCopy` loc flag
* the `TAssignmentFlag` enum and usages thereof

Thanks to ARC, there's no difference between `nkAsgn` and `nkFastAsgn`
in the C code generator, so the two branches are collapsed into a single
one.


saem (approved):
yay, code is collapsing down further!

PR #744, opened 2023-06-07
transf: adjust for-loop related location lifetimes

zerbina:
## Summary

Change the scoping and lifetime behaviour of locations defined as part
of for-loops to the following:
* in terms of location lifetime, a for-loop body open a new scope.
  Previously, locations defined inside the body were placed inside the
  `yield` statement's surrounding scope
* the lifetime of for-vars is bounded by the scope of the for-loop body,
  instead of starting when control-flow reaches and first `yield`
  statement, and ending when control-flow leaves the *iterator*

This makes the scoping behaviour (in terms of location lifetimes) of
`for` loops consistent with that of other language constructs (such as
`if` and `while`). In addition, `yield` statements in `.inline`
iterators become more similar to invoking a continuation (the for-loop
body) with some arguments (`yield` arguments).

## Details

* don't emit the for-var definitions at the start of the inlined
  iterator
* adjust `transformYieldAsgn` to generate `nkIdentDefs` instead of
  assignments
* also pass the generated identdefs to `introduceNewLocalVars` when
  necessary; no fresh copies of the for-vars would be created otherwise
* wrap the for-loop body in an `nkBlockStmt` before replacing the
  `yield` statement with it; this achieves the wanted scoping behaviour
* add the new `forLoopDefs` iterator to `ast_query`
* consider `nkVarTuple`s with the `isSimpleIteratorVar` optimization
  (they were previously just ignored)

In addition to the consistency improvements, the changed lifetimes also
improves:
* compile times: shorter lifetimes reduce the amount of work for data-
  flow analysis
* efficiency at runtime: assigning to for-vars that are not turned into
  cursors now always uses a bit-wise copy, instead of a call to the
  `=sink` hook

### Tests

* add a specification-style test for the lifetime behaviour
* adjust the test depending on `--expandArc` output

### Before vs. After

Assuming the following iterator and for-loop using it:
```nim
iterator iter(): int =
  var i = 0
  while i < 1:
    yield i
    inc i

for v in iter():
  var x = v
```

inlining the iterator previously produced code that is *semantically*
equivalent to:
```nim
var v
var i = 0
while i < 1:
  v = i
  var x = v
  inc i
```

With this commit, the output becomes:
```nim
var i = 0
while i < 1:
  block:
    var v = i
    var x = v
  inc i
```


tests/arc/topt_no_cursor.nim
```diff
@@ -87,26 +87,26 @@ var lan_ip
...
+            var line = a_cursor[i]
+            splitted = split(line, " ", -1)
+            if `==`(splitted[0], "opt"):
+              var :tmp = splitted[1]
+              `=copy`(lan_ip, :tmp)
+            echo([lan_ip])
+            echo([splitted[1]])
+          finally:
+            `=destroy`(splitted)
+        inc(i, 1)
```
saem, line 109: The placement of the the `inc` outside of the `try` makes more sense, the isolation (clarified lifetimes) that the block provides allows for better intuitional understanding too!

zerbina:
The JavaScript code generator produces incorrect code for taking the address of a `let` location and assigning it to a global (or `var ptr`), which result in the test failure here. I'm going to make a separate PR with the fix.

PR #745, opened 2023-06-07
transf: fix inline iterator invocation semantics

zerbina:
## Summary

Specify in the manual how parameter passing for iterators should work,
and adjust `transf` so that inline iterators work according to the new
specification.

This fixes (in the context of inline iterators):
* arguments to `var` and, depending on the argument expression,
  `openArray` parameters being evaluated on *use* inside the iterator's
  body
* literal `ref` construction expression arguments being evaluated on
  *use* (and thus potentially multiple times)
* `sink` parameters not enforcing ownership (the parameter was not
  always mutable, and it also wasn't destroyed at the end of the
  iterator's scope)
* taking the address of an iterator's parameter resulting in illegal
  code being generated by the code generator (this was dependent on
  the shape of the argument expression) 

Closure iterators used with `for`-loops don't adhere to the new
specification yet, and work the same as they did previously.

## Details

Argument handling is split into two parts:
1. deciding the semantics of what the semantics should be (i.e.
   introduce a full copy, inline at parameter usage, etc.). This only
  takes into consideration the parameter and argument type
2. generating the AST for enforcing the semantics

There are four possible modes of how an argument is passed:
- via forwarding: inline the arguments expression at each parameter
  usage
- by copy: create a full copy of the argument. Used for `sink`, single-
  location view, and mutable `openArray`s types. For views, full copy
  only means copying the view, not the viewed location(s)
- by value: depending on the argument, pass either by name or by shallow
  copy. No full copy must be materialized here
- through temporary: assign the argument to a temporary, and the
  temporary to the local introduced for the parameter. This a workaround
  for `transformConv` eliding to-`openArray` conversions

Except for the forwarding mode, each parameter gets allocated a local
corresponding to it, with all usages of the original parameter symbol in
the inlined body being replaced with the symbol of the local.

Previously, locals (read, temporaries) were only introduced when the
argument is a complex expression (with some additional quirks), such as
calls. The temporaries were initialized via fast assignments, but since
those assignments are turned into normal ones by `mirgen`, this provided
no benefit.

Since the inlining logic now always uses either cursors or views for the
locals corresponding to immutable parameters, efficiency of the produced
code slightly increases in some cases.

One place in the compiler code (the `vmbackend.cpairs` iterator) relied
on the previous forwarding semantics for simple argument expressions -- 
the iterator is changed to use a `var` parameter.

Finally, multiple tests for the changed behaviour, as well as the fixed
issues, are added.

### Misc

* rename the `flatten` procedure from `astgen.nim` to `flattenExpr` and
  move it to `trees.nim`
* move the `canUseView` procedure from `astgen.nim` to the new
  `ast_analysis.nim` module


tests/lang_callable/iter/titer_issues.nim
```diff
@@ -250,3 +250,31 @@ block:
...
+
+  iterator iter(a: RefObj): int =
+    # use `a` multiple times in order to detect the argument expression
+    # being erroneously inlined
+    inc a.i
+    inc a.i
+    yield a.i
+
+  for i in iter(RefObj(i: 1)):
+    doAssert i == 3
```
zerbina, line 280: The problem here was that the argument passing logic treated an `nkObjConstr` node as something with no side-effects, as long as none of the field initializer expression had any. While this is true for the construction of normal objects, it is not true for when the result is of `ref` type. In effect, what happened here was that each occurrence of `a` in the iterator's body was replaced with the `nkObjConstr` AST, and the integer value yielded by the iterator was thus 1.

I've first thought about this while working on the MIR, but I think that given their different behaviour, it would make sense to have a dedicated node kind for `ref` object construction. Alternatively, the `nkObjConstr` nodes for them could be lowered into something like:
```nim
let x = new T
y[] = # object construction expression goes here
x
``` 
early on, but that might complicate some analysis a bit.

saem: Playing off what you're driving towards, I think `ref` construction is perhaps best thought of as:

- `ref`'s location allocation
- `ref`'d memory allocation + the consturctor expression as initialization
- assigning the `ref`'d memory's address to the `ref` location

In which case, the `nkObjectConstr` for a `ref` should be within some node that treats it as an initialization expression and not the usual "value construction". Or in other words, "value construction" fuses (combines & elides) the more granular steps that the `ref` variant takes, just a matter of perspective.

saem (approved):

doc/manual.rst
```diff
@@ -4390,6 +4390,38 @@ parameters of an outer factory proc:
...
+How exactly this is achieved for `inline` iterators is up to the compiler, but
+with the following restrictions / guarantees:
+- no full copy must be introduced when passing to an immutable parameter (a full
+  copy might still be needed in order to ensure the left-to-right evaluation
+  order)
+- except for `seq`s, `string`s, and `ref`s, if the argument is an lvalue
+  expression, the parameter must not be a full location, but only store the
+  handle
+- if the argument is an unamed location (e.g. the result of a call), a full
+  location *may* be used
```
saem: Just noticed this upon re-reading, but we don't have a definition for "full location" prior to this, don't think it needs to be sorted in this PR.

I take it to mean something that stores a direct value, as opposed to a handle (`ref`/`ptr`) or a forwarded expression.

saem:
Noticed one thing about 'full location' in the manual; and that's not a block in my mind. I'm good with hitting `/merge`.

doc/manual.rst
zerbina: Yeah, I was a bit mindless with the term here. I think the more proper description would be "owning location". From a quick skim of the manual, that term also doesn't seem to be defined anywhere yet, but I think it's a bit more self-explanatory (and correct) than "full location", so I'm going to change "full" to "owning" here.

zerbina: I updated it.

saem:
will wait for CI to go green, running too many things at once seems to be an easy way for jobs to start getting cancelled, especially during the day where the infrastructure seems to be busier.

PR #746, opened 2023-06-08
consistent canonical import and package sym naming

saem:
## Summary

Package symbols for modules are now generated in a consistent fashion:

- at the root of a package: `pkgroot/module`
- nested within a package: `pkgroot/pkgpath/module`
- modules within the project package omit `pkgroot/`

This leads to more consistent messages and a variety of improvements for
internals (eg: more stable signature hashes) and a better foundation to
build upon.

## Details

Previously, package symbols were created and modules added to them in a
flat manner, until there was a name collision (two modules in different
directories with the same name). In the case of collisions a mangled
package name was created based on the module that caused the collision
and that was used instead. Although the package symbol information isn't
used in too many parts of the compiler this led to yet more broken data
and subtle issues for anything that relied upon it:

- signature hashing owners and procedures
- c, rod, and various file name generation
- error messages: canonical imports
- and more

Getting package info is now done via `packagehandling.getPkgDesc`, which
retrieves various key details, including dealing with the lack of
package definition. Package detection still remains coupled to nimble
files, for now.

### DIFF
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -1481,12 +1481,17 @@ proc findProjectNimFile*(conf: ConfigRef; pkg: string): string =
 
 proc canonicalImportAux*(conf: ConfigRef, file: AbsoluteFile): string =
   ## canonical module import filename, e.g.: system.nim, std/tables.nim,
-  ## system/assertions.nim, etc
+  ## system/assertions.nim, etc. Canonical module import filenames follow the
+  ## same rules as canonical imports (see `canonicalImport`), except the module
+  ## name is followed by a `.nim` file extension, and the directory separators
+  ## are OS specific.
   let
     desc = getPkgDesc(conf, file.string)
     (_, moduleName, ext) = file.splitFile
   if desc.pkgKnown and
-     AbsoluteFile(getNimbleFile(conf, conf.projectFull.string)) != desc.pkgFile:
+     desc.pkgFile != AbsoluteFile(conf.getNimbleFile(conf.projectFull.string)):
+    # we ignore the pkg root name for intra-package module imports, allows for
+    # easier pkg renames (without changing all files using canonical imports).
     result = desc.pkgRootName
     if desc.pkgSubpath != "":
       result = result / desc.pkgSubpath
@@ -1498,6 +1503,13 @@ proc canonicalImportAux*(conf: ConfigRef, file: AbsoluteFile): string =
 proc canonicalImport*(conf: ConfigRef, file: AbsoluteFile): string =
   ## Shows the canonical module import, e.g.: system, std/tables,
   ## fusion/pointers, system/assertions, std/private/asciitables
+  ## 
+  ## A canonical import path is:
+  ## 
+  ## - typically `pkgroot/pkgsubpath/module`
+  ## - if a module is at the base of a package, then `pkgroot/module`
+  ## - if a module is within the project's package, `pkgroot` is skipped like
+  ##   so `pkgsubpath/module` or `module` (if the module is at the package root).
   let ret = canonicalImportAux(conf, file)
   result = ret.nativeToUnixPath.changeFileExt("")
 
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -254,7 +254,6 @@ type
 
     jsonBuildFile*: AbsoluteFile
     nimStdlibVersion*: NimVer
-    moduleOverrides*: StringTableRef
     cfileSpecificOptions*: StringTableRef ## File specific compilation options for C backend.
     ## Modified by `{.localPassc.}`
     inputMode*: ProjectInputMode    ## how the main module is sourced
@@ -945,7 +944,6 @@ proc newConfigRef*(hook: ReportHook): ConfigRef =
     m: initMsgConfig(),
     headerFile: "",
     packageCache: newPackageCache(),
-    moduleOverrides: newStringTable(modeStyleInsensitive),
     cfileSpecificOptions: newStringTable(modeCaseSensitive),
     inputMode: pimFile,
     projectMainIdx: FileIndex(0'i32), # the canonical path id of the main module
@@ -1356,15 +1354,6 @@ proc rawFindFile2(conf: ConfigRef; f: RelativeFile): AbsoluteFile =
       return canonicalizePath(conf, result)
   result = AbsoluteFile""
 
-proc patchModule(conf: ConfigRef, result: var AbsoluteFile) =
-  ## If there is a known module override for a given
-  ## `package/<name(result)>` replace result with new module override.
-  if not result.isEmpty and conf.moduleOverrides.len > 0:
-    let key = getPackageName(conf, result.string) & "_" & splitFile(result).name
-    if conf.moduleOverrides.hasKey(key):
-      let ov = conf.moduleOverrides[key]
-      if ov.len > 0: result = AbsoluteFile(ov)
-
 when not declared(isRelativeTo):
   proc isRelativeTo(path, base: string): bool =
     # pending #13212 use os.isRelativeTo
@@ -1416,7 +1405,6 @@ proc findFile*(conf: ConfigRef; f: string; suppressStdlib = false): AbsoluteFile
         result = rawFindFile2(conf, RelativeFile f)
         if result.isEmpty:
           result = rawFindFile2(conf, RelativeFile f.toLowerAscii)
-  patchModule(conf, result)
 
 proc findModule*(conf: ConfigRef; modulename, currentModule: string): AbsoluteFile =
   ## Return absolute path to the imported module `modulename`. Imported
@@ -1448,7 +1436,6 @@ proc findModule*(conf: ConfigRef; modulename, currentModule: string): AbsoluteFi
       result = AbsoluteFile currentPath / m
     if not fileExists(result):
       result = findFile(conf, m)
-  patchModule(conf, result)
 
 proc findProjectNimFile*(conf: ConfigRef; pkg: string): string =
   ## Find configuration file for a current project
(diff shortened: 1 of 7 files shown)


saem:
@zerbina if you have any thoughts about canonical import generation that'd be appreciated.

compiler/modules/modules.nim
```diff
@@ -53,39 +53,36 @@ template getModuleIdent(graph: ModuleGraph, filename: AbsoluteFile): PIdent =
...
+        let temp = newSym(skPackage, rootPkg, packageId(), nil, info)
+        graph.packageSyms.strTableAdd(temp)
+        temp
+      else:
+        existingRootPkgSym
+
+  if desc.pkgSubpath == "":
+    result = rootPkgSym
   else:
-    let modules = graph.modulesPerPackage.getOrDefault(result.itemId)
```
zerbina, line 73: I think the `modulesPerPackage` field can be removed now? If I'm reading the code correctly, it's not written to anymore.

compiler/modules/modules.nim
```diff
@@ -53,39 +53,36 @@ template getModuleIdent(graph: ModuleGraph, filename: AbsoluteFile): PIdent =
...
-        # we used to produce an error:
-        localReport(graph.config, info,
-          "module names need to be unique per Nimble package; module clashes with " &
-            toFullPath(graph.config, existing.info.fileIndex))
+    let
+      subPkg = getIdent(graph.cache, desc.pkgName)
+      existingSubPkgSym = graph.packageSyms.strTableGet(subPkg)
+    result =
+      if existingSubPkgSym.isNil:
+        let subPkgSym = newSym(skPackage, subPkg, packageId(), rootPkgSym, info)
```
zerbina, line 81: Apologies if this is obvious, but what are sub-package symbols needed for? Phrased differently, what would be the implications of also only having the `rootPkgSym` in the `subpath != ""` case?

compiler/front/options.nim
```diff
@@ -1493,19 +1480,23 @@ proc findProjectNimFile*(conf: ConfigRef; pkg: string): string =
...
-  let dir = getNimbleFile(conf, $file).parentDir.AbsoluteDir
-  if not dir.isEmpty:
-    let relPath = relativeTo(file, dir)
-    if not relPath.isEmpty and (ret.isEmpty or relPath.string.len < ret.string.len):
-      ret = relPath
-  if ret.isEmpty:
-    ret = relativeTo(file, conf.projectPath)
-  result = ret.string
+  ## canonical module import filename, e.g.: system.nim, std/tables.nim,
+  ## system/assertions.nim, etc
```
zerbina: Could you document what canonical import paths represent?

From reading the code, I believe that the new meaning is "a module's path relative to its package (or the current project directory)". Is that correct? As for its previous meaning, I don't think I understand what the old code tried to do.

compiler/modules/packagehandling.nim
```diff
@@ -34,27 +35,77 @@ proc getNimbleFile*(conf: ConfigRef; path: string): string =
...
+                    "@": "@@",
+                    ":": "@c"})
+  let
+    pkgFile = getNimbleFile(conf, modulePath)
+    (pkgFileRoot, pkgFileName, _) = pkgFile.splitFile
+    pkgKnown = pkgFileName != ""
+    pkgRootName = if pkgKnown: pkgFileName else: "unknown"
+    pkgRoot = if pkgKnown: pkgFileRoot else: conf.projectPath.string
+    relativePath = relativePath(modulePath.parentDir, pkgRoot)
+    pkgSubpath = if relativePath == ".": "" else: relativePath
```
zerbina: Only a very soft suggestions, but I think this code would be easier to parse if there'd only be less `if pkgKnown` predicates, like so:
```nim
...
result =
  if pkgKnown:
   let (pkgFileRoot, pkgFileName, _) = ...
   PkgDesc(pkgKnown: true, ..., pkgRoot: pkgFileRoot, ...)
  else:
    ...

let relativePath = ...
# name and subpath setup goes here

```

zerbina (commented):
I'm not very familiar with this part of the compiler, so my comments are more at the code / surface level.

Regarding usage of canonical paths in test output and how to fix the test failures, I still need to think about it a bit.

compiler/modules/modules.nim
saem: The modules per package, but that's not written to anymore, hmm... that might be a bug that I'm not doing that. The issue is that if we go `foopkg/barmodule`, then it's ambiguous which `barmodule` we actually mean, in case there is more than one:

```
/foopkg
    /a
        barmodule
    /b
        barmodule
```

compiler/front/options.nim
```diff
@@ -1493,19 +1480,23 @@ proc findProjectNimFile*(conf: ConfigRef; pkg: string): string =
...
+  let
+    desc = getPkgDesc(conf, file.string)
+    (_, moduleName, ext) = file.splitFile
+  if desc.pkgKnown:
```
```suggestion
  if desc.pkgKnown and
     AbsoluteFile(getNimbleFile(conf, conf.projectFull.string)) != desc.pkgFile:
```
This fixes the test failures, and is also a sensible rule in general, I think. Essentially, if module A belongs to the same package as the project file, the package qualifier is not included in the canonical path.

Side-note: a call to `toAbsolute` might be needed for the result of `getNimbleFile`, both here and in `getPkgDesc`.

compiler/modules/modules.nim
zerbina: Hm, but providing the full sub-path is always required, no? What I mean is, even if only `foopkg/a/barmodule.nim` would exist, a `foopkg/barmodule` import would still not refer to a file (if I'm understanding the file lookup code correctly, that is).

compiler/front/options.nim
saem: the new code is `pkgroot/<pkgpath>/module`, where `pkgroot/pkgpath` are the "actual package", similar to Java, where `java.utils` and `java.utils.collections` (or whatever they're called) are actually two separate packages, even if they've got common parts.

---

The old code was a bunch of hacks, that sorta kinda worked:
- if we have a pkg, and no configured `path` or it's longer, use pkg relative
- if we did get a relative path based on explicit search `--path`s , use path relative
- if we none of the above, use project file relative

IIUC the second one caused the tests to work before because the current directory is added to the search paths when invoking the compiler.

---

about the only thing I can think of for the interim, is to detect if we're _within_ the package and if so omit the package root, it's not a long term solution, but it shrinks the change and gives time to think. Thoughts?

compiler/modules/packagehandling.nim
saem: Sorry, I slapped this together a bit quick and dirty, I'll clean it up.

compiler/front/options.nim
saem: Hahaha, this is the same conclusion I came to and asked about in the other comment, maybe drop the package root if it's shared. 😆

compiler/front/options.nim
zerbina: I think we arrived at the same solution: https://github.com/nim-works/nimskull/pull/746#discussion_r1223416175

compiler/modules/modules.nim
saem: ### How the Code Used to Work

Yup, but the symbol building worked entirely differently. It would add modules in a flat manner to a package symbol, unless it detected a collision. In the case of collisions, it created a symbol with a mangled package name (based on the module, see `packagehandling.fakePackageName`) and associated the module to that symbol, along with making that package symbol's owner the original untangled packaged symbol... so convoluted. 😮‍💨 

### Impact of this Change

I'm not sure of all the implications myself, I'm sure there are a few considering we do package symbol traversal in a few spots (I can't recall all the details, but here is what I dug up quickly):
- `vmaux.matches` does ignore the fake package and jump straight to the root, but that also weakens lookup expressivity?
- `pluginsupport` don't believe it's impacted, as it just looks for an overall id
- `sighashes.hashOwner` and `hashProc` will have slightly different behaviour now; stable across modules in the same root/subpackage
- `cgen.getSomeNameForModule` will now generate better names?

### Looking Ahead

Regardless, putting modules in flatly seems wrong because that's not what's actually happening, a subpath is required for correct module lookup. Getting this to work correctly might not impact current code but it would open the door on making it possible to evolve things, e.g. module lookup based on packages _not_ file system.

saem: > Apologies if this is obvious, but what are sub-package symbols needed for?

It's not, after having re-reviewed things a bit, the impacts are subtle and mostly correctness oriented. I think the bigger deal is that we have a shot of building on top of the newer approach where we previously didn't.

> Phrased differently, what would be the implications of also only having the `rootPkgSym` in the `subpath != ""` case?

I think the code would truck along and we'd get a few oddities here and there. In particular, we'd have less unique input for things like `sighashes` and some potential vm lookup issues via `vmaux.matches`. These things would likely lie dormant for a long time/not be a problem. But I don't think we could evolve things from that as a base.

In particular the `compilepreter`, which I'd like to ensure isn't too polluted with legacy silliness. I was writing up some integration and I couldn't make sense of package symbols enough such that I felt I'd be able to convert back to `PNode` output for integrating existing backends. I'd have to either implement something from scratch (error prone/second system) or ... not change this up so I can make some sense of it.

compiler/modules/modules.nim
saem: Yup, seems entirely unnecessary now, except for book keeping, but that can be restored at a later time, if required.

compiler/modules/modules.nim
zerbina: Thank you for the explanations! Yeah, having the sub-package symbols sounds like the better choice.

saem:
Cleaned up the code and PR message, going to queue this to merge (feel free to remove it if there is an issue)

PR #747, opened 2023-06-08
threads: free reference cycles when terminating

zerbina:
## Summary

When using `--gc:orc`, run a full cycle collection pass prior to exiting
the internal thread procedure, freeing all heap cells only kept alive
through reference cycles created by the thread.

Previously, they were simply not freed, with manually freeing them being
impossible once the thread terminated.

## Details

Running the cycle collector (via ``GC_fullCollect``) has to happen
*before* deallocating the ``GcThread`` instance, as the latter step
frees - when using emulated thread-local storage - the thread's
thread-local storage (which the cycle collector still depends on).

### DIFF
--- a/lib/system/threads.nim
+++ b/lib/system/threads.nim
@@ -175,6 +175,11 @@ template threadProcWrapperBody(closure: untyped): untyped =
   # However this is doomed to fail, because we already unmapped every heap
   # page!
 
+  when compileOption("gc", "orc"):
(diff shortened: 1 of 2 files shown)


saem (approved):

saem:
Very nice find, this might have been one of the hard to find leaks that was being encountered by the `cps` project.

PR #748, opened 2023-06-08
compiler: partially remove `codegenDecl` pragma

zerbina:
## Summary

Altering the code generators' output through programmable string-
formatting (which is what the pragma provides) does not align with
NimSkull's goals, and the feature is generally becoming more and more
unstable as a side-effect of the backend-processing unification.

It is now mentioned as deprecated in the manual, only allowed on
`proc`s, and treated as having no effect when not using the JavaScript
backend.

## Details

The `asyncjs` module is the only remaining user of the pragma, and since
the VSCode extension currently depends on said module, the pragma is not
fully removed yet.

* remove all `.codegenDecl` support from `cgen`
* disallow the pragma on `var`/`let` statements
* remove all tests related to the pragma
* adjust the manual


compiler/sem/pragmas.nim
```diff
@@ -1324,8 +1318,6 @@ proc applySymbolPragma(c: PContext, sym: PSym, it: PNode): PNode =
       of wNoInit:
         result = noVal(c, it)
         incl(sym.flags, sfNoInit)
-      of wCodegenDecl:
```
saem: I think this needs to stay, but with a `c.config.backend` check against `backendJs` or else error?

doc/manual.rst
```diff
@@ -7088,37 +7088,26 @@ calling syntax: ``obj.method(arg)``.
...
-will generate this C code:
-
-.. code-block:: c
-  int progmem a
+.. warning::
+  This pragma is deprecated and going to be removed.
 
-For procedures, $1 is the return type of the procedure, $2 is the name of
-the procedure, and $3 is the parameter list.
+The `codegenDecl` pragma can be used to directly influence Nim's JavaScript
```
```suggestion
The `codegenDecl` pragma can be used to directly influence |Nimskull|'s JavaScript
```

not strictly necessary, just while it's being changed anyhow, quick win.

saem (commented):
The narrowed usage is much better, it'll make it easier to figure out how to replace it in the JS case down the road (still thinking about it).

compiler/sem/pragmas.nim
zerbina: Yep, it does. The original changeset fully removed the pragma, and this is one place where I missed to revert the change.

zerbina:
/merge

Sorry, I had forgotten to update the PR message.

PR #749, opened 2023-06-09
sequtils: simplify `mapIt` implementation

zerbina:
## Summary

Use an inline iterator instead of a macro for achieving the "evaluate
once" semantics in the `mapIt` template. This makes the code simpler,
and, given that the compiler's code itself uses `mapIt`, also reduces
the amount of macro usage during bootstrapping.

## Details

* introduce the internal `mapIter` iterator
* remove the now obsolete internal `evalOnceAs` macro
* adjust the internal documentation of `mapIt` a bit
* adjust the `t8339.nim` test to not depend on `sequtils`
* don't introduce an unnecessary shallow copy when lifting `op` into a
  procedure

While the `mapIter` doesn't use `lent`, special care is taken for it to
be eligible for the cursor optimization. With the current compiler, for-
vars are always turned into cursors when the invoked inline iterator
only ever yields the *symbol* of a local.

The `t8339.nim` test is intended to make sure that aliases are skipped
when rendering types, but using `sequtils` for testing that is not
required, and only made the test unnecessarily depend on the `mapIt`
implementation details (which now changed, making the test fail).

### DIFF
--- a/lib/pure/collections/sequtils.nim
+++ b/lib/pure/collections/sequtils.nim
@@ -89,27 +89,16 @@ when defined(nimHasEffectsOf):
 else:
   {.pragma: effectsOf.}
 
-macro evalOnceAs(expAlias, exp: untyped,
-                 letAssigneable: static[bool]): untyped =
-  ## Injects `expAlias` in caller scope, to avoid bugs involving multiple
-  ## substitution in macro arguments such as
-  ## https://github.com/nim-lang/Nim/issues/7187.
-  ## `evalOnceAs(myAlias, myExp)` will behave as `let myAlias = myExp`
-  ## except when `letAssigneable` is false (e.g. to handle openArray) where
-  ## it just forwards `exp` unchanged.
-  expectKind(expAlias, nnkIdent)
-  var val = exp
-
-  result = newStmtList()
-  # If `exp` is not a symbol we evaluate it once here and then use the temporary
-  # symbol as alias
-  if exp.kind != nnkSym and letAssigneable:
-    val = genSym()
-    result.add(newLetStmt(val, exp))
-
-  result.add(
-    newProc(name = genSym(nskTemplate, $expAlias), params = [getType(untyped)],
-      body = val, procType = nnkTemplateDef))
+iterator mapIter[T; U; V](src: U, dest: var seq[V]): T =
+  ## Helper iterator for implementing ``mapIt``. The destination sequence is
+  ## pre-allocated here, using the length provided by `src`.
+  let L = src.len
+  dest.newSeq(L)
+  # while we could yield an index + item pair, we do not, as that would currently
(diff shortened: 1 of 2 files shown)


lib/pure/collections/sequtils.nim
```diff
@@ -991,46 +980,34 @@ template mapIt*(s: typed, op: untyped): untyped =
...
-      var it{.inject.}: typeof(items(s), typeOfIter);
-      op), typeOfProc)
+  type
+    InType  = typeof(items(s), typeOfIter)
+    OutType = typeof((
+      block:
+        var it{.inject.}: InType;
+        op), typeOfProc)
+  # we cannot use normal overloading instead of the 'when' here, as the that
+  # force `op` to be ``typed``, which is what we don't want it to be
```
saem:
```suggestion
  # we use `when` instead of overload resolution which would require `op` to be
  # ``typed``, allowing optimization of some `op`s to inline expressions
  # instead of requiring lambdas.
```

Bit more verbose, but I believe it accurately captures the intention (don't assume the `when` is bad) and convey the impact of mucking about with it carelessly.

lib/pure/collections/sequtils.nim
```diff
@@ -991,46 +980,34 @@ template mapIt*(s: typed, op: untyped): untyped =
...
-    # Here, we avoid to create closures in loops.
-    # This avoids https://github.com/nim-lang/Nim/issues/12625
-    when compiles(s.len):
-      block: # using a block avoids https://github.com/nim-lang/Nim/issues/8580
-
-        # BUG: `evalOnceAs(s2, s, false)` would lead to C compile errors
-        # (`error: use of undeclared identifier`) instead of Nim compile errors
-        evalOnceAs(s2, s, compiles((let _ = s)))
-
+    # the normal version
```
saem:
```suggestion
    # the normal version (inline expression)
```

saem (approved):
soft suggestions.

That's a fair bit easier to read! 👓

lib/pure/collections/sequtils.nim
zerbina: Hm, I was trying to communicate that it'd be better for there to be two overloads:
```nim
template mapIt(s: typed, op: proc) = ...
template mapIt(s: typed, op: untyped) = ...
```
instead of implementing manual overloading via `when` (which is only a workaround).

saem: Good thing I wrote out my interpretation.

Could also do:
```nim
  # xxx: drop the `when` by overloading `mapIt`, one where the `op` arg is of
  #      type `proc` and the other `untyped`
```

Again, all soft suggestions, not blockers.

zerbina: Yep, that is better, thanks. I've turned the comment into an `xxx` one, with an additional note that a compiler issue needs to be resolved first.

PR #750, opened 2023-06-09
sequtils: simplify `toSeq` implementation

zerbina:
## Summary

Don't use complex macros, but instead use `compiles` + generic routines
for achieving the same. This should slightly improve compile times for
code often using `toSeq`, and is also another step towards less macro
usage during bootstrapping.

## Details

This change partially reverts https://github.com/nim-works/nimskull/pull/151,
which replaced the usage of `compiles` with macros. However, compared to
macros, `compiles` is the less problematic feature usage here, allowing
the implementation to be significantly less complex. In addition, the
`compiles` here is also easier to replace with type-inspection routines
in the future (such as `x is iterator`, once that works).

### DIFF
--- a/lib/pure/collections/sequtils.nim
+++ b/lib/pure/collections/sequtils.nim
@@ -749,85 +749,25 @@ template anyIt*(s, pred: untyped): bool =
 
 template toSeqBuiltin[T](arg: openArray[T]): seq[T] = @arg
 
-proc toSeqBuiltin[T](arg: set[T]): seq[T] {.inline.} =
-  for it in arg:
-    result.add it
+template toSeqBuiltin[T](arg: seq[T]): seq[T] = arg
 
-proc toSeqBuiltin(a: cstring): seq[char] {.inline.} =
-  let len = a.len
-  result.setLen len
+proc toSeqBuiltin[T](arg: cstring): seq[T] {.inline.} =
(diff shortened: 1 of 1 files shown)


lib/pure/collections/sequtils.nim
```diff
@@ -844,10 +784,16 @@ macro toSeq*(arg: untyped): untyped =
     assert mySeq1 == @[1, 2, 3, 4, 5]
     assert mySeq2 == @[1'i8, 3, 5]
 
-  if arg.kind in {nnkSym, nnkIdent, nnkOpenSymChoice, nnkClosedSymChoice}:
-    result = newCall(bindSym"toSeqImplSingleSym", arg)
+  mixin items
+  when compiles(typeof(items(arg))):
+    # arg is something that supports the ``items`` iterator:
+    toSeqBuiltin[typeof(items(arg))](arg)
+  elif compiles(typeof(arg())):
```
zerbina, line 791: This should be turned into `arg is iterator`, once the compiler support testing for iterator types that.

saem (approved):
lesson learned... hopefully. 🤣 

Type driven is better than AST driven... especially as you said once we have the ability to test for `iterator`.

saem:
/merge

looks like a spurious failure

PR #751, opened 2023-06-09
sequtils: fix double-evaluation bug with `applyIt`

zerbina:
## Summary

Use the `mitems` iterator for implementing the template, fixing the bug
where the argument expression passed to the template was evaluated
multiple times.

The change also makes the template more general: the only requirement
for the input now is that it supports being iterated over via
`mitems`; it no longer needs to be index-able by an integer.


tests/lang_callable/template/twrongmapit.nim
```diff
@@ -30,3 +30,15 @@ import sequtils
 (var i = @[""];i).applyIt(it)
 # now works:
 echo "##", i[0], "##"
+
+block evaluation_order:
```
saem, line 34: is it worth checking this invariant for the other `xxxit` templates?

saem (approved):

tests/lang_callable/template/twrongmapit.nim
zerbina: Yep, it is, but this PR is only meant to fix immediate bug blocking #714. I'd make extending the test case a follow-up PR.

PR #753, opened 2023-06-10
vmgen: fix accessing `var`/`lent` views in objects

zerbina:
## Summary

Make `vmgen` emit the bytecode for loading pointer views into a register
prior to the dereference, making access of `var`/`lent` views stored in
objects work for code being compiled for the VM target (refer to the
added tests for examples; most of them previously crashed the VM).

## Details

The VM can only dereference pointers - which direct single-location
views part of compound types map to for the VM target - if they're
stored in registers. Due to the `vmgen` logic for generating the code
for `nkDotExpr` (field access) and `nkBracketExpr` (array and seq
access) expressions not considering views when testing whether a load-
into-register is required, a handle-to-pointer operand was passed to
`LdDeref`.

`needsRegLoad` now checks whether a load-into-register is required via
the new `fitsRegisterConsiderView`, which treats direct single-location
views as fitting into registers.

### Future Direction

The code generator needs to map types to a register / internal type and
then base the various type queries it performs (`isDirectView`,
`fitsRegister`, etc.) on the mapped type.

### DIFF
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -1377,10 +1377,9 @@ func fitsRegisterConsiderView(t: PType): bool =
   # XXX: introduce a ``mapType`` (similar to the one used by the other code
   #      generators) and base the "fits register" queries on that
   let t = t.skipTypes(IrrelevantTypes)
-  result = fitsRegister(t)
-  if not result:
-    # is it a direct single-location view?
-    result = t.kind in {tyVar, tyLent} and t.base.kind != tyOpenArray
+  fitsRegister(t) or
+    # is it a direct single-location view?:
+    t.kind in {tyVar, tyLent} and t.base.kind != tyOpenArray
 
 template needsRegLoad(): untyped =
   mixin load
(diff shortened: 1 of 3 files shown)


compiler/vm/vmgen.nim
```diff
@@ -1371,9 +1371,20 @@ proc genCard(c: var TCtx; n: PNode; dest: var TDest) =
...
+func fitsRegisterConsiderView(t: PType): bool =
+  ## Returns whether a value of type `t` fits into a register, also
+  ## considering view-types that map to pointers.
+  # XXX: introduce a ``mapType`` (similar to the one used by the other code
+  #      generators) and base the "fits register" queries on that
+  let t = t.skipTypes(IrrelevantTypes)
+  result = fitsRegister(t)
+  if not result:
+    # is it a direct single-location view?
+    result = t.kind in {tyVar, tyLent} and t.base.kind != tyOpenArray
```
```suggestion
fitsRegister(t) or
    t.kind in {tyVar, tyLent} and t.base.kind != tyOpenArray
    # above tests for direct single-location view
```

very soft suggestion, it's slightly less imperative by exploiting mcarthy evaluation; however, fitting in the comment neatly is tricky.

saem (approved):
wow, quite the jackpot you've hit. :D

compiler/vm/vmgen.nim
zerbina: I changed it into an expression, but put the comment (with a colon at the end) above the second condition.

PR #754, opened 2023-06-11
remove the old runtime

zerbina:
## Summary

Remove most of the old runtime, which became obsolete with the
removal of the old GCs. This significantly shrinks downs the `system`
module and its includes, making it easier to maintain and further
refactor them.

Public interface procedures and types related to the old GCs are also
removed. These are:
- the `ForeignCell` API
- `GC_ref` and `GC_unref` for `seq`s and `string`s
- the `GC_setStrategy` procedure

In addition, tracking both implicit and explicit `alloc` and `free`
(when using `--memTracker:on`) now works again.

## Details

Regarding the GC, the removed ones are `boehm`, `go`, `markAndSweep`,
`refc`, `regions`, `gc2`, `hooks`, and `none`. Their direct
implementation is removed, as well as all usages of the conditional
symbols related to them, which are: `boehmgc`, `gogc`, `gcMarkAndSweep`,
`gcRegions`, `gcHooks`, and `nogc`.

Without any stack-scanning GCs, the allocator doesn't need the
`isAllocatedPtr` routine anymore, so it is removed. There were also some
optional checks and `track` calls that were not enabled when using the
new runtime; they're ported over.

A list of removals and changes related to the old
runtime follows:
* remove the version 1 `seq` and `string` implementation
* remove the version 1 `repr` implementation
* keep the old RTTI (`TNimType`; required for `deepcopy` and
  `typeinfo`), but remove the now obsolete fields (`marker` and
  `finalizer`)
* remove `objectInit`
* keep the parts of `genericAssign` that the `typeinfo` module requires
* keep `genericReset`; the deepcopy and `genericAssign` implementation
  need it
* remove various other `.compilerprocs` related to the old runtime

Usage of the `nimV2`, `nimSeqsV2`, and `gcDestructors` conditional
symbol in code that is only available with the C backend are also
removed, as the symbols are always present there.

Finally, the now obsolete internal `cellseqs_v1` module is removed.


saem (approved):
The delete is strong int this one. :D

This gives renewed hope to improved gdb pretty printing support, as well.

saem:
/merge

(sorry, thought I saw something, my mistake)

PR #757, opened 2023-06-11
fix(sem): `typed` `import`s preserve node kind

iacore:
## Summary

`from x import y` and `import x except y` now generate typed AST with
their respective node kind.

## Details

Before this patch, `from x import y` and `import x except y` were
semantically analysed as `nkImportStmt`, resulting in illformed `typed`
AST.

The producers of the relevant AST (`evalFrom` and `evalImportStmt`) are
changed to use the same node kind as that of the input. Additionally,
since these kinds didn't reach `mirgen` before (but now they do),
`mirgen` was updated to ignore them.

Now the AST can roundtrip through semantic analysis, and other consumers
(such as passes, macros, or external tools), and allows proper inspection
of import statements.

The following code will now generate correct `typed` trees:

```nim
import options # nkImportStmt
import os, options # nkImportStmt
from os import joinPath, `/` # nkFromStmt
import os except joinPath, `/` # nkImportExceptStmt
```


tests/lang_callable/macros/tnimskull757.nim
```diff
@@ -0,0 +1,39 @@
+discard """
```
saem: I'd rename this file `ttyped_importstmts_nimskull757.nim`

This would setup a nice convention `ttyped_<nodekind>...`, pretty sure there is a better home to suss out for these tests in the future.

tests/lang_callable/macros/tnimskull757.nim
```diff
@@ -0,0 +1,39 @@
...
+  output: '''
```
saem:
```suggestion
  description: "Test import statements are properly analysed"
  output: '''
```

My description itself isn't the best, here is a recent example of a good one liner from Zerbina's PR: https://github.com/nim-works/nimskull/pull/756/files#diff-e37ff9cec4f49c315bfae4bcdc98161e8ba8f4db3b04346b2c51ae7c38ea7bbbR4

saem (changes requested):
This is a nice fix. I left some recommendations for the test.

Please update the PR description and I can merge it.

tests/lang_callable/macros/tnimskull757.nim
iacore: I've moved it to a better place

zerbina:
Regarding the PR description, for PRs that fix something, I think that the description should answer the following questions:
- what didn't work? what failed? what was wrong?
- (depending on the problem) why didn't it work?
- what is changed to make it work?

---

When writing the description, it can help to ask oneself the above question (and follow up questions) and then try to answer them, like so (using this PR as an example):

*Question:* what was wrong?
*Answer:* `from x` and `import x except y` were always represented as `nkImportStmt` in typed

*Question:* what were the consequences?
*Answer:* there was an irreversible loss of information and the produced typed AST was semantically wrong. The problematic AST was not able to roundtrip through semantic analysis, and other consumers (such as passes, macros, or external tools) were not able to properly inspect import statement

*Question:* what is changed?
*Answer:* the producers of the relevant typed AST (`evalFrom` and `evalImportExcept`) are changed to use the same node kind as that of the input. Since these node kinds previously did not reach `mirgen` (but now they do), `mirgen` now needs to handle (i.e., discard) them 

The answers can then be used as the basis for writing the description.

iacore:
Is the typed AST not used for codegen? I'm confused why the compiler works with this bug.

Also, I don't know *that* much about nimskull

saem:
> Is the typed AST not used for codegen? I'm confused why the compiler works with this bug.

It is used for codegen, but it's not the only input. The compiler pass system has two key data structures built up and available for codegen, besides the AST. They are `modulegraphs.ModuleGraph` and `options.ConfigRef`. In particular the former collects a lot of information about modules, symbol tables, etc that can be used in lieu of information less conveniently encoded within the AST itself. In this case the imports a particular module requires. That's why codegen was able to proceed and the bug was never discovered.

With your fix, we could build upon it (separate PR) that at the very least cross references imports in the AST vs those derived from other sources.

> Also, I don't know _that_ much about nimskull

Not a problem, can help answer questions and lend a hand. This also helps highlight things I, or others, take for granted and could be improved over time.

iacore:
@saem can this be merged?

saem:
@iacore I've updated the PR title/description, please take a look at the content and format.

The title is 50 chars, describing the change. The content contains the `summary` and `details` headings, the former describes the net result today, and the latter describes what changed and how in more detail. The body is formatted to 72 chars lines. The formatting will ensure a legible commit message.

Once you've had a chance to read the updated PR, ping me here or in chat and I can trigger the merge.

saem (approved):

saem:
Just want to do one more quick thing before merge.

PR #758, opened 2023-06-12
Move contributing.rst to doc/

iacore:
Is this good?


saem:
can you try `./koch.py doc`, that should see if the docs still build. note, this is super spammy, sadly. you can check the `doc/html/` directory for a `contributing.html`. 

For a clean build/retry of docs, a `rm -fr doc/html` will do. I'm trying to find the github config for pointing to the contribution doc, we moved it to the `.github/` directory for github integration-y reasons that I can't recall. Maybe @alaviss can?

alaviss:
> I'm trying to find the github config for pointing to the contribution doc, we moved it to the `.github/` directory for github integration-y reasons that I can't recall. Maybe @alaviss can?

That's correct, there are 3 hard-coded locations according to [Github Docs](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors#adding-a-contributing-file):

- `/contributing.rst`
- `/docs/contributing.rst` (notice the `s` in `docs`)
- `/.github/contributing.rst`

IIRC, we selected `.github` to avoid moving the `doc` folder which would "erase" the history of files within (without better tooling). I think Github got better at history for moves though, so we could reconsider the `/doc` -> `/docs` change.

PR #760, opened 2023-06-13
internal: drop `nfBlockArg` usage

saem:
## Summary

`nfBlockArg`, an internal flag used to track whether a block argument,
encoded by the `parser` as a `nkStmtList` is no longer used. This is
largely transparent, except for macros receiving a `typed` argument
block will always be wrapped in a statement list, even if there is only
a single statement.

## Details

Internally the compiler tracked block arguments as statement lists with
an internal flag `nfBlockArg` on the node. This prevented various
normalization within semantic analysis from removing this node. This
flag is no longer used and results in only minor changes to parameter
behaviour within `macros`.

Now, any call to a macro where the block style argument syntax is used,
a statement list node is always passed. Whereas before, singleton
statement list nodes were unwrapped, this is no longer the case. This
is considered acceptable as any non-singleton statement would result in
the macro receiving a statement list, and so this behaviour must be
guarded for regardless.

Within the `parser` `isBlockArg` is no longer tracked and in turn
`nfBlockArg` is not applied to `PNode`s during conversion from the
parsed tree to legacy AST. The flag and field in the `ParsedNodeData`
have been removed.

`semStmtList` now exposes a `collapsed` parameter, which `semExprs` uses
to inform statement list unwrapping, this case only when within a
`concept` matching check. `semOperand` has been updated to check if its
input is a statement list, and if it is, it mimics `semExprs` without
statement unwrapping, such as applying `nfSem` to the semanticized
output.

A described earlier, this change means that `typed` arg blocks will not
be unwrapped, and the `tgettypeinst` was adjusted for this.

### DIFF
--- a/compiler/sem/sigmatch.nim
+++ b/compiler/sem/sigmatch.nim
@@ -2581,11 +2581,10 @@ template isVarargsUntyped(x): untyped =
   x.kind == tyVarargs and x[0].kind == tyUntyped
 
 proc findFirstArgBlock(m: var TCandidate, n: PNode): int =
+  # xxx: this "feature" isn't a great idea and the implementation is awful
   # see https://github.com/nim-lang/RFCs/issues/405
   result = int.high
   for a2 in countdown(n.len - 1, 0):
-    # checking `nfBlockArg in n[a2].flags` wouldn't work inside templates
-    # xxx: ^^ why???
     case n[a2].kind
     of nkStmtList:
       let formalLast = m.callee.n[m.callee.n.len - (n.len - a2)]
--- a/compiler/sem/sigmatch.nim
+++ b/compiler/sem/sigmatch.nim
@@ -2519,24 +2519,6 @@ proc prepareOperand(c: PContext; formal: PType; a: PNode): PNode =
   if formal.kind == tyUntyped:
     assert formal.len != 1
     result = a
-  # elif formal.kind == tyTyped and a.typ.isNil:
-  #   var hasError = false
-  #   case a.kind
-  #   of nkStmtList, nkStmtListExpr:
-  #     result = shallowCopy(a)
-  #     for i, k in a.pairs:
-  #       doAssert k.kind != nkEmpty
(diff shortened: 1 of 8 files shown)


compiler/sem/sigmatch.nim
```diff
@@ -716,6 +716,10 @@ proc matchUserTypeClass*(m: var TCandidate; ff, a: PType): PType =
         diagnostics.add report.semReport
   )
 
+  # xxx: this is where we end up with collapsed bodies and drives the need for
+  #      the check in `semexprs.semExpr`'s `nkStmtList/Expr` branch, a
+  #      `semTryStmt` that doesn't collapse should remove the awkward logic and
```
```suggestion
  #      `semTryExpr` that doesn't collapse should remove the awkward logic and
```

compiler/sem/sigmatch.nim
```diff
@@ -2509,7 +2513,11 @@ proc setSon(father: PNode, at: int, son: PNode) =
 
 # we are allowed to modify the calling node in the 'prepare*' procs:
 proc prepareOperand(c: PContext; formal: PType; a: PNode): PNode =
-  if formal.kind == tyUntyped and formal.len != 1:
+  when defined(nimCompilerStacktraceHints):
+    {.line.}:
+      frameMsg(c.config, a)
```
```suggestion
    frameMsg(c.config, a)
```

zerbina (approved):
I think this is a really good solution to getting rid of `nfBlockArg`! There are two minor things I noticed, but it's nothing blocking.

compiler/sem/sigmatch.nim
saem: ~IIRC the lineinfo gets messed up without `{.line.}:` to unwrap each template application.~

Nevermind, that won't impact anything.

zerbina: Hm, but since not inside a template, it would just override the line information of `frameMsg(c.config, a)` with the line information from some unrelated template expansion, or crash the compiler with an index defect, no?

PR #761, opened 2023-06-13
vmgen: fix double evaluation of `swap` arguments

zerbina:
## Summary

Fix a bug the VM code generator had where the expression used
as the first argument was always evaluated twice -- it is now
only evaluated once.

## Details

The `lowerSwap` procedure that was used for lowering the `swap`
call into assignments duplicated the first argument expression,
resulting in it being evaluated two times.

Usage of `lowerSwap` is replaced with manually emitting the
bytecode for the assignments, which in addition to fixing the
bug and making the generated code slightly more efficient (the
introduced temporary is freed immediately after the swap now),
also has the benefit of removing both the import dependency on
`lowering` and `vmgen` needing access to an `IdGenerator`.

`lowerSwap` has no further usage sites, and is thus removed.

### DIFF
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -51,9 +51,6 @@ import
     msgs,
     options
   ],
-  compiler/sem/[
-    lowerings
-  ],
   compiler/utils/[
     idioms
   ],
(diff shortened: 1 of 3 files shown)


compiler/vm/vmgen.nim
```diff
@@ -1810,7 +1807,37 @@ proc genMagic(c: var TCtx; n: PNode; dest: var TDest; m: TMagic) =
...
     unused(c, n, dest)
-    c.gen(lowerSwap(c.graph, n, c.idgen, if c.prc == nil or c.prc.sym == nil: c.module else: c.prc.sym))
+    let tmp = getTemp(c, n.typ)
+    # XXX: swap doesn't need to be implemented here; lower it into assignment
+    #      with a MIR pass
+    # there's no ``nkHiddenAddr`` on the operands, so we don't need to skip
+    # var types
+    if fitsRegister(n[1].typ):
+      # we need to account for the fact that either operand could be
+      # stored in a register alreay (because it's a local variable)
```
```suggestion
      # stored in a register already (because it's a local variable)
```

saem (approved):
minor typo; still readable.

PR #762, opened 2023-06-13
typed `import` test fix and document known issues

saem:
## Summary

Fixed test for typed `import` statements, it tested `untyped`
imports, and documented known issues discovered during the process.

## Details

The test previously tested `untyped` imports, this meant it was testing
what was parsed, and not what was semantically analysed. The test has
been updated. As part of the update, actual test modules were created in
order for typing to succeed and be introspected.

While updating the test, two known issues were found and documented:
1. `import except` drops the except identifiers
2. `import as` drops the original symbol, retaining only the alias

The behaviour might be acceptable in the long run, but it's flagged for
further consideration.


saem:
FYI: @iacore, that's why my original test had the parameters as `typed`. I didn't read the test carefully enough, the `Ident` node kinds were a give away.

iacore:
> FYI: @iacore, that's why my original test had the parameters as `typed`. I didn't read the test carefully enough, the `Ident` node kinds were a give away.

that didn't compile

saem:
> > FYI: @iacore, that's why my original test had the parameters as `typed`. I didn't read the test carefully enough, the `Ident` node kinds were a give away.
> 
> that didn't compile

I ran it before I sent it over and just confirmed it now and it works, weird. Hopefully the updated PR is useful in providing some insights, it definitely gave me some ideas as to some things to poke at next -- latent bugs.

PR #763, opened 2023-06-14
ccgexprs: fix two copy-on-write string bugs

zerbina:
## Summary

Fix two bugs where incorrect C code was generated in situations where
copy-on-write string are involved. These are:
1. the `s` in `result = s[0]` or `yield s[0]`, where the return type is
   `var char`, was evaluated two times
2. the source string was not prepared for mutation when passing a `s[0]`
   expression to a `var char` parameter or using it as the initializer
   expression for a mutable view

Bug 2 resulted in crashes at run-time when the source string hadn't been
prepared for mutation already.

## Details

Instead of the manual `cow` and `cowBracket` calls (which were the
reason for the double evaluation), emitting the
`nimPrepareStrMutationV2` calls is now left to `genSeqElem`, which
detects when to emit the call by checking for the presence of the
`lfPrepareMutation` loc flag on the destination location.

`genAddr` now accepts an additional `mutate` parameter, which decides
whether `genAddr` includes the `lfPrepareMutation` flag on the
intermediate `TLoc` it creates. For `nkHiddenAddr` nodes that produce
an address meant for mutation, `mutate` is set to 'true'; all other
address-of operation pass 'false'.

This makes `cow` and `cowBracket` obsolete, and preparing a string for
mutation work in all cases where a `nkHiddenAddr` is involved (which is
the case for the two situations where the prepare calls were previously
missing).

Since the arguments to `swap` are not wrapped in `nkHiddenAddr` nodes,
`genSwap` has to handle including the loc flag itself.

### Misc

Remove the redundant elif branch in `genAddr`. It does the same thing as
the `else` branch, and since `mapType` will never map a `tyRef` or
`tyPtr` to `ctArray`, both can be merged.

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -535,29 +535,17 @@ proc genDeref(p: BProc, e: PNode, d: var TLoc) =
       discard getTypeDesc(p.module, e.typ, skVar)
       putIntoDest(p, d, e, "(*$1)" % [rdLoc(a)], a.storage)
 
-proc cowBracket(p: BProc; n: PNode) =
-  if n.kind == nkBracketExpr:
-    let strCandidate = n[0]
-    if strCandidate.typ.skipTypes(abstractInst).kind == tyString:
-      var a: TLoc
-      initLocExpr(p, strCandidate, a)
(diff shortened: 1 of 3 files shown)


saem (approved):
Nice fix.

Maybe it's time to start including a view build like we used to have for arc/orc with the compiler?

PR #764, opened 2023-06-14
frontend: rename default cache folder to nimskull

mrgaturus:
## Summary
* what changed and how?
`~/.cache/nim` is changed to `~/.cache/nimskull`
`C:/Users/myuser/nimcache` is changed to `C:/Users/myuser/nimskullcache`

* why are we changing it?
Changing cache folder name is also a part of finishing project rebranding.

## Details
two hardcoded strings were changed, also there is a comment to easy find `|Nimskull|`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,8 +288,8 @@ jobs:
             --git.commit:'${{ github.sha }}' \
             --git.devel:devel
 
-          # Remove leftover nimcache
-          rm -rf doc/html/nimcache
+          # Remove leftover nimskullcache
+          rm -rf doc/html/nimskullcache
 
       - id: package
         name: Create release package
@@ -346,8 +346,8 @@ jobs:
             --git.commit:'${{ github.sha }}' \
             --git.devel:devel
 
-          # Remove leftover nimcache
-          rm -rf doc/html/nimcache
+          # Remove leftover nimskullcache
+          rm -rf doc/html/nimskullcache
 
       - id: package
         name: Create release package
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,11 @@
 !**/
 !*.*
 
-# Cache
+# |Nimskull| Bootstrap Cache
+nimskullcache/
+rnimskullcache/
+dnimskullcache/
+# Old Cache Path
 nimcache/
 rnimcache/
 dnimcache/
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -49,7 +49,8 @@ const
 
 const
   harmlessOptions* = {optForceFullMake, optNoLinking, optRun, optUseColors, optStdout}
-  genSubDir* = RelativeDir"nimcache"
+  genSubDir* = RelativeDir"nimskullcache"
+  # XXX: |Nimskull| extensions and config files
   NimExt* = "nim"
   RodExt* = "rod"
   HtmlExt* = "html"
@@ -1254,7 +1255,7 @@ include compiler/modules/packagehandling
 
 proc getOsCacheDir(): string =
   when defined(posix):
-    result = getEnv("XDG_CACHE_HOME", getHomeDir() / ".cache") / "nim"
+    result = getEnv("XDG_CACHE_HOME", getHomeDir() / ".cache") / "nimskull"
   else:
     result = getHomeDir() / genSubDir.string
 
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -1311,7 +1311,7 @@ proc toGeneratedFile*(
     path: AbsoluteFile,
     ext: string
   ): AbsoluteFile =
-  ## converts "/home/a/mymodule.nim", "rod" to "/home/a/nimcache/mymodule.rod"
+  ## converts "/home/a/mymodule.nim", "rod" to "/home/a/nimskullcache/mymodule.rod"
   result = getNimcacheDir(conf) / RelativeFile(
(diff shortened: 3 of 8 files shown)


saem:
Ah, there are a few occurences of "nimcache" in various parts of the compiler and a few bits of CI. It's not too bad though, as many of them are variable names and most aren't a string used for the actual directory. Doing a quick sweep of those should reveal a few spots.

Non-exhaustive list I found:
- `.gitignore` should include `nimskullcache`, as well (bootstrapping will likely generate `nimcache` still)
- some of the clean-up in ci.yml likely needs to delete the new relative dir name; maybe both, because the bootstrap compiler still uses `nimcache`
- kochdoc and niminst might need a change, they look at both

zerbina:
Thank you for the PR, @mrgaturus!

Could you remove the mention of "might reuse cache files" from the PR message? I wasn't able to reproduce it so far, and it's also unrelated to the issue you reported. When not using the `-r` or `r` commands, semantic analysis, during which the `requires '.importc' types to be '.completeStruct'` error is reported, is always run, independent from the contents of the cache folder.

mrgaturus:
@saem i didn't change koch nimcache path because as you said that is generated inside compiler folder, do i change that folder name also?

@zerbina i tested again and you're right, it doesn't reuse cache files, but i wonder why it compiled once and then not when i deleted cache, so there might be a possibility. Regarding {.completeStruct.}, that is not related to this small issue but it occurs on one of my [modules](https://github.com/mrgaturus/npainter/blob/master/src/spmc.nim#L222), a quick fix may be remove the `when sizeof(T) > 0` of [create(T) proc](https://github.com/nim-works/nimskull/blob/devel/lib/system/memalloc.nim#LL148C16-L148C16)

saem:
@mrgaturus hoping to clear up some time to pull down the branch today/early tomorrow and see what happens when I:
1. bootstrap
2. build tools
3. run the testament suite
4. build docs

Mostly watching to see what gets created on the file system and see if that identifies anything. That's the plan, if you'd like you can give it whirl ahead of me (LMK if you'd like the commands I was intending to run for the above).

mrgaturus:
@saem tools and tests compiles inside nimskull cache folder, only bootstraping compiles inside nimcache folder relative to koch.py file, i think koch's nimcache folder is fine for now because compiler binary name is still called `nim`

saem:
@alaviss can you review this since you know the build parts better than most? Thank you.

alaviss (changes requested):
There are some parts of the CI/build pipeline that assumes a `nimcache` folder which should be updated:

- First `nimskullcache` should be added to `.gitignore` at the root of this repository.
- Change these to `nimskullcache`: 
  - https://github.com/nim-works/nimskull/blob/57efeec85faf136e5fbd8634c6b0c4da9f965b66/.github/workflows/ci.yml#L291-L292
  - https://github.com/nim-works/nimskull/blob/57efeec85faf136e5fbd8634c6b0c4da9f965b66/.github/workflows/ci.yml#L349-L350
- Adjust this conditional to also ignores `nimskullcache` when packaging: https://github.com/nim-works/nimskull/blob/57efeec85faf136e5fbd8634c6b0c4da9f965b66/tools/niminst/niminst.nim#L274

Most other parts are already using a named `--nimcache` so should not be affected by this change, though we might want to rename them in the future as well.

mrgaturus:
also i will change koch path which is the path CI may use:
https://github.com/nim-works/nimskull/blob/57efeec85faf136e5fbd8634c6b0c4da9f965b66/koch.py#L20-L21

saem:
Wrapping up a proc clean-up in the compiler.

Then I'll use the trick @alaviss metioned, which is to pull down the CI artifacts and have a look at what folders got created/are lingering around.

Thanks for the efforts thus far, @mrgaturus. 🚀

saem (approved):
LGTM

Reviewed the CI artifacts, there is a `nimskullcache` directory in the `compiler Linux` and `compiler macOS` artifacts, but these are used to share state between steps, I believe that's to save time/share across CI steps. Otherwise, the binaries, docs, and all other artifacts are free of any cache directories. Also checked that all the changes @alaviss asked for are made. 🎉.

saem:
/merge

Thanks for doing this, @mrgaturus!

mrgaturus:
i wonder why a macOS CI test fails:
```
NOTINBATCH: tests/ic/timports_temp.nim  --incremental:on -d:nimIcIntegrityChecks 
diff --git a/outputGotten_0_c.txt b/outputExpected_0_c.txt
index 3beb6c5186..9a8bc9fe83 100644
--- a/outputGotten_0_c.txt
+++ b/outputExpected_0_c.txt
@@ -1081,10 +1081,6 @@ megatest:processing: [327] tests/stdlib/metaprogramming/tunittest.nim
 [Suite] test name filtering
 megatest:processing: [328] tests/stdlib/os/osproc/tclose.nim
 megatest:processing: [329] tests/stdlib/os/osproc/texitsignal.nim
-megatest:processing: [0] tests/ambsym/tresolution_identifier_2_ok.nim
-megatest:processing: [1] tests/arithm/tashr.nim
-megatest:processing: [2] tests/arithm/tdiv.nim
-megatest:processing: [3] tests/assign/tassign.nim
 true
 true
 megatest:processing: [330] tests/stdlib/os/tdistros_detect.nim

FAIL: megatest output different, see outputGotten_0_c.txt vs outputExpected_0_c.txt
```

alaviss:
/merge

Should be safe to re-run. It's a known issue that sometimes tests are skipped in megatest and error out.

PR #765, opened 2023-06-14
sigmatch: fix `typeRel` for objects inheriting from phantom types

zerbina:
## Summary

Fix the type-relation (`typeRel`) analysis for object types that have a
phantom type somewhere in the inheritance hierarchy. In the context of
type matching, different instances of a phantom type (e.g. `Type[int]`
and `Type[float]`) were previously treated as being the same.

This also fixes an overloading bug where `Sub = object of Base[T]`
matched equally well against `Base[T]` and `Sub`, resulting in erroneous
ambiguity errors.

## Details

When computing the relation between formal = `tyGenericInst` and
actual = `tyObject`, the closest base type of `actual` that isn't a
`tyObject` object was compared against formal. This skipping was a
previous attempt at fixing the analysis, but it is incorrect, and is
what caused the ambiguity error.

The incorrect type skipping is removed, and replaced with a sub-type
relationship analysis (the new `isSubtypeOfGenericInstance` procedure)
that considers `tyGenericInst`. This fixes phantom types being
ignored when matching a non-generic object type (with a phantom type
somewhere in the hierarchy) against an instance of a generic type.

Consider the generic type `Type[T] = object`, where `T` is phantom
type information (i.e., it is not used in `Type`'s body). While both
`Type[int]` and `Type[float]` produce distinct `tyObject` types
(different IDs), comparing them with `sameObjectType` would yield
`true`, as the procedure only performs a symbol ID + type structure
comparison!

Due to phantom types being ignored, `distinctBase` returning
a skipped type went by unnoticed -- `distinctBase` now returns an
unskipped type.

Finally, the `tinheritance_generic_dispatch.nim` test is extended,
renamed to `tphantom_type_as_base`, and moved to the `typeRel`
category.


compiler/sem/sigmatch.nim
```diff
@@ -471,6 +471,56 @@ proc isGenericSubtype(c: var TCandidate; a, f: PType, d: var int, fGenericOrigin
     d = depth
     result = true
 
+proc isSubtypeOfGenericInstance(c: var TCandidate; a, f: PType, fGenericOrigin: PType): int =
+  ## Computes whether the resolved object-like type `a` is a subtype of `f`,
+  ## where `f` is a ``tyGenericInst``. The inheritance depth is returned,
+  ## or, if the types are not related, -1.
+  ##
+  ## In case of a match, the unresolved generic parameters of `f` are bound to
+  ## the arguments taken from the base type of actual that matched.
```
saem, line 480:
```suggestion
  ## the arguments taken from the base type of `a` that matched.
```

I'm guessing "actual" is more "`a`ctual"

compiler/sem/sigmatch.nim
```diff
@@ -471,6 +471,56 @@ proc isGenericSubtype(c: var TCandidate; a, f: PType, d: var int, fGenericOrigin
...
+  assert f.kind == tyGenericInst
+  var
+    askip = skippedNone
+    fskip = skippedNone
+
+    t = a.skipToObject(askip)
+    last = t ## the unskipped type
+
+  assert t != nil, "'a' is not object-like type"
```
saem, line 489:
```suggestion
  assert t != nil, "'a' is not an object-like type"
```

saem (approved):
I'll give it a shot after this PR is merged, but I'm _guessing_ there are plenty of other places where we ignore phantom type information and shouldn't. 😬

saem:
This feels way more sound than before. 🎉

saem:
/merge

Looks spurious

PR #766, opened 2023-06-17
memalloc: remove compile time sizeof check

mrgaturus:
## Summary
Remove the compile-time `sizeof` check from the `create` and `createU`
procs and instead forbid `UncheckedArray[T]`.

This allows for using both procedures with types that are imported,
contain imported types, or, more generally, have a size that is not
known at the NimSkull side.

## Details
- using `sizeof` with a type of unknown size when in a compile-time
  context results in an `requires .completeStruct` error  
- passing an `UncheckedArray[T]` type to `create` is a common
  mistake that results in undefined behaviour (which is the reason why
  the `sizeof` checks were originally introduced)
- `createShared` and `createSharedU` now also disallow `UncheckedArray`
  types


zerbina (changes requested):
Thank you for contributing!

The checks were introduced by https://github.com/nim-lang/nim/pull/19032, which was implemented as a response to [this](https://github.com/nim-lang/nim/issue/19000) issue (`create(UncheckedArray)` having surprising behaviour). Eventually, Nim faced the same problem as you are doing now (using `create` with a type that has a non-compile-time known size leads to a compilation error) and the checks were thus removed again (https://github.com/nim-lang/Nim/pull/20518).

Disallowing the use of `create` with `UncheckedArray` (or dynamically sized objects in general) does make sense, and I think the `sizeof` checks should be replaced with `T is UncheckedArray`. That won't prevent the problematic use of dynamically sized objects like:
```nim
type Obj = object
  i: int
  data: UncheckedArray[int]
```
with `create`, but it will at least detect and disallow direct usage with `UncheckedArray`.

In addition, please add a test for `create` and `createU` with an imported type, so that it is ensured that the change works as expected.

mrgaturus:
forbid UncheckedArray without ptr and ref makes sense because manipulating raw allocs with UncheckedArray is a common pattern, it think a proc called `toUncheckedArray` would make some elegance:
```nim
let a = create(cint, 2000).toUncheckedArray
# let a  = cast[ptr UncheckedArray[cint]](create(cint, 2000))
a[10] = 1234
a.dealloc()
```

lib/system/memalloc.nim
```diff
@@ -117,8 +117,8 @@ when hasAlloc and not defined(js):
     ## See also:
     ## * `create <#create,typedesc>`_
     static:
-      when sizeof(T) <= 0:
-        {.fatal: "createU does not support types T where sizeof(T) == 0".}
+      when T is UncheckedArray:
+        {.error: "UncheckedArray[T] does not have size, use T instead".}
```
zerbina: Could you remove the redundant wrapping in a `static` block (here and in the other places)?

lib/system/memalloc.nim
```diff
@@ -237,6 +237,9 @@ when hasAlloc and not defined(js):
     ##
     ## See also:
     ## * `createShared <#createShared,typedesc>`_
+    static:
+      when T is UncheckedArray:
+        {.error: "UncheckedArray[T] does not have size, use T instead".}
```
zerbina:
```suggestion
        {.error: "UncheckedArray[T] has no size. Use T instead".}
```
(also applies to the other three places)

tests/system/talloc3.nim
```diff
@@ -0,0 +1,74 @@
+discard """
+  targets: "c"
+"""
```
zerbina:
```suggestion
  description: '''
    Using the higher-level allocation procedures with types that have a size not
    known at the NimSkull side must work
  '''
"""
```
In addition, please rename the test to something more descriptive (maybe `talloc_with_unknown_size.nim`).

tests/system/talloc3.nim
```diff
@@ -0,0 +1,74 @@
...
+
+type
+  # Imported C Types as Opaques
+  M128* {.importc: "__m128", header: "xmmintrin.h".} = object
+  M128i* {.importc: "__m128i", header: "emmintrin.h".} = object
+  M128d* {.importc: "__m128d", header: "emmintrin.h".} = object
```
zerbina:
```suggestion
type
  Obj {.exportc.} = object

# make sure the type is really exported:
var use: Obj

type
  Imported {.importc: "Obj", nodecl.} = object
    # opaque type
```
This makes the test less dependent on C implementation details.

tests/system/talloc3.nim
```diff
@@ -0,0 +1,74 @@
...
+    b {.bitsize: 4.}: cint
+  MixedPixel = object
+    r {.bitsize: 4.}: cint
+    g: int
+    b: M128i
+    a: M128
+  MapNaive = UncheckedArray[NaivePixel]
+  MapSIMD = UncheckedArray[SIMDPixel]
+  MapTiny = UncheckedArray[TinyPixel]
+  MapMixed = UncheckedArray[MixedPixel]
```
zerbina:
```suggestion
  Native = object
    field: int
  WithImported = object
    field: Imported
  WithBitField = object
    field {.bitsize: 4.}: cint
  Mixed = object
    r {.bitsize: 4.}: cint
    g: int
    b: Imported
  ArrayNative       = UncheckedArray[Native]
  ArrayWithImported = UncheckedArray[WithImported]
  ArrayWithBitField = UncheckedArray[WithBitField]
  ArrayMixed        = UncheckedArray[Mixed]
```
I think that this makes the test slightly easier to read and comprehend.

zerbina (commented):
Thanks. I left some small requests regarding the code and tests.

When you're ready, please also update the PR message to reflect what is changed, as it is the message that will be used for the commit merged into the `devel` branch.

For the PR message, I think it makes sense to include a small mention of the reason behind the now removed `sizeof` check.

zerbina (approved):
@mrgaturus Nice work. I have reworded the PR message to be a bit shorter, fit in with the general commit message style, and to have a line length of 72.

If you're okay with the changes, please notify me here or on Matrix and I'll kick off the merge.

mrgaturus:
@zerbina Thank you for your feedback. That is ready for merge 🚀

zerbina:
/merge

Onto the `devel` branch it goes :D

saem:
I just got back from the gym, with merging this on my mind, pleasantly surprised to see it already on the way. :D

Thanks for seeing this through, @mrgaturus and @zerbina.

PR #768, opened 2023-06-19
fix(CI): remove self-executing tests from megatest

saem:
## Summary

Marked `texitsignal` as not joinable to exclude it from the `megatest`
category; this should hopefully limit intermitent CI failures.

## Details

`texitsignal` attempts to execute itself with different parameters and
uses that to branch logic of what to check. This doesn't fit with the
rest of the megatest category.

Also, marked these tests as not `joinable` as they either self-execute
or kickoff compiler processes:

- `texception_message_null_byte`
- `tstaticlib`
- `tstackframes`
- `treadlines`
- `tvmops`

Lastly, disabled `tclose` for macOS testing within the same category, as
it's linux only already to save some CI cycles.

### DIFF
--- a/tests/stdlib/os/osproc/tclose.nim
+++ b/tests/stdlib/os/osproc/tclose.nim
@@ -1,5 +1,6 @@
 discard """
   exitcode: 0
+  disable: osx
 """
 
 when defined(linux):
--- a/tests/stdlib/os/osproc/tclose.nim
+++ b/tests/stdlib/os/osproc/tclose.nim
@@ -1,6 +1,6 @@
 discard """
   exitcode: 0
-  disable: osx
+  disabled: osx
 """
 
 when defined(linux):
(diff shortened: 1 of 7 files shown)


tests/stdlib/os/osproc/tclose.nim
```diff
@@ -1,5 +1,6 @@
 discard """
   exitcode: 0
+  disable: osx
```
zerbina:
```suggestion
  disabled: osx
```
This should fix the test failure.

zerbina (approved):
Another test that self-executes and is joined is the `texception_message_null_byte.nim` test. Marking it as unjoinable would also fix the megatest for the C target failing when no NodeJS installation is present.

tests/stdlib/os/osproc/tclose.nim
saem: I was scouring the logs trying to spot an error and couldn't find the spec parse error. 😅 

Thanks.

saem:
> Another test that self-executes and is joined is the `texception_message_null_byte.nim` test. Marking it as unjoinable would also fix the megatest for the C target failing when no NodeJS installation is present.

Ah, thanks for the heads up. I did a quick search for `getCurrentCompiler` inside the test directory and marked a bunch more as `joinable: false`, I'm going to include them and update the PR description.

PR #769, opened 2023-06-19
jsgen: fix multiple bugs with `addr` and `lent`

zerbina:
## Summary

Fix incorrect JavaScript code being generated for `addr` operations
involving parameters, `let` locations, or imported/exported variables,
and for `lent T` types where `T` is a pointer-like type. In addition,
slightly clean up the address-handling in `jsgen`.

## Details

A "fat pointer" (array storing a base and index) needs to be constructed
when taking the address of locations not using indirection (such as
`let` locations), but this was missing, and the variable that represents
the location instead returned directly.

* consider locations not boxed in an array (e.g. `let` locations,
  parameters) in `genSymAddr`
* remove the scattered special handling for taking the address of
  parameters
* introduce the `isBoxedPointer` procedure and use it to replace the
  manual checks
* simplify the address assignment logic in `genVarInit` (made possible
  by `genSymAddr` now working properly)
* enable and unrelated test case in `taddr.nim` that already succeeded
  previously

The above fixes uncovered two additional JS code generator bugs that
were previously hidden:

* `mapType` didn't consider resolved concept types when testing whether
  a type maps to a representation that has reference semantics
* the hidden address-of operation for `lent T` types where `T` is
  something that maps to a "fat pointer" wasn't skipped

Both resulted in incorrect code being generated in the respective
circumstances, but they are now also fixed.

### DIFF
--- a/compiler/backend/jsgen.nim
+++ b/compiler/backend/jsgen.nim
@@ -2472,12 +2472,11 @@ proc gen(p: PProc, n: PNode, r: var TCompRes) =
   of nkObjConstr: genObjConstr(p, n, r)
   of nkHiddenStdConv, nkConv: genConv(p, n, r)
   of nkAddr, nkHiddenAddr:
-    if mapType(n.typ) != etyBaseIndex:
+    if n.typ.kind == tyLent or mapType(n.typ) != etyBaseIndex:
       # the operation doesn't produce an address-like value (e.g. because the
-      # operand is a JS object and those already have reference semantics)
-      # FIXME: this won't work for ``lent`` views (outside of return values),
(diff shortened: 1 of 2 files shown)


saem (approved):
That's a lot of fixes. :D

Tangent (not in this PR): For parameters, if we really wanted I supposed we could make `arguments` the base of the fat pointer?

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

zerbina:
> Tangent (not in this PR): For parameters, if we really wanted I supposed we could make `arguments` the base of the fat pointer?

Yep, I think that would work, good call. Using `arguments` would also make properly supporting `sink` parameters a lot easier.

PR #770, opened 2023-06-21
testament: small architectural refactoring

zerbina:
## Summary

Make `TSpec` a pure description of the test specification and don't
reuse it for storing test output. This moves decision-making regarding
whether or not to run a test out of specification parsing, and slightly
improves `testament`s general architecture.

## Details

* remove the `err` property from `TSpec`
* make `inCurrentBatch` a property of `TTest` (instead of `TSpec`)
* introduce the `CompileOutput` type, which captures the output of a
  compiler invocation -- `callNimCompiler` now returns an instance of it
* move early decision-making about what to do with a test (is it
  disabled?, the specification invalid?, etc.) into the new
  `computeEarly` procedure
* adjust `testSpec` and `testSpecWithNimcache` to use `computeEarly`

Due to `compilerOutputTests` re-using `CompileOutput` (previously
`TSpec`) for temporary storage, `CompileOutput` still stores a
`TResultEnum` instead of just a simple boolean.

Multiple types and globals in `spec.nim` can now be moved to
`testament.nim`, but for the sake of a smaller changeset, this is left
to a follow-up commit.

### DIFF
--- a/testament/specs.nim
+++ b/testament/specs.nim
@@ -9,7 +9,6 @@
 
 import
   std/[
-    sequtils,
     parseutils,
     strutils,
     os,
@@ -18,7 +17,8 @@ import
     tables,
     hashes,
     sets
-  ]
+  ],
+  system/platforms
 
 type TestamentData* = ref object
   # better to group globals under 1 object; could group the other ones here too
@@ -32,6 +32,7 @@ var
   skips*: seq[string]
 
 let testamentData0* = TestamentData()
+# TODO: ^^ move to ``testament.nim``
 
 type
   TTestAction* = enum
@@ -106,8 +107,13 @@ type
     ccodeCheck*: seq[string] ## List of peg patterns that need to be
     ## searched for in the generated code. Used for backend code testing.
     maxCodeSize*: int ## Maximum allowed code size (in bytes) for the test.
-    err*: TResultEnum
-    inCurrentBatch*: bool
+    disabledOs*: set[OsPlatform]
+      ## the OS'es the test is disabled for
+    disabledCpu*: set[CpuPlatform]
+      ## the cpu architectures the test is disabled for
+    disable32bit*: bool
+      ## whether the test is disabled on 32-bit systems
+
     targets*: set[TTarget]
     specifiedTargets*: set[SpecifiedTarget] ## targets as described by the spec
     matrix*: seq[string]
@@ -125,7 +131,6 @@ type
     timeout*: float ## in seconds, fractions possible, but don't rely on
     ## much precision
     inlineErrors*: seq[InlineError] ## line information to error message
-    debugInfo*: string ## debug info to give more context
     knownIssues*: seq[string] ## known issues to be fixed
     labels*: seq[string] ## user-added metadata
 
@@ -138,6 +143,7 @@ type
 
 # Exported global RetryContainer object.
 var retryContainer* = RetryContainer(retry: false)
+# TODO: ^^ move to ``testament.nim``
 
 proc getCmd*(s: TSpec): string =
   ## Get runner command for a given test specification
@@ -317,6 +323,7 @@ proc initSpec*(filename: string): TSpec =
   result.file = filename
 
 proc isCurrentBatch*(testamentData: TestamentData; filename: string): bool =
+  # TODO: move elsewhere
   if testamentData.testamentNumBatch != 0:
     hash(filename) mod testamentData.testamentNumBatch == testamentData.testamentBatch
   else:
@@ -327,14 +334,6 @@ proc parseSpec*(filename: string,
                 nativeTarget: TTarget): TSpec =
   ## Extract and parse specification for a given file path
   result.file = filename
-
-  when defined(windows):
-    let cmpString = result.file.replace(r"\", r"/")
-  else:
-    let cmpString = result.file
-  if retryContainer.retry and not retryContainer.names.anyIt(cmpString == it[0]):
-    result.err = reDisabled
-    return result
 
   let specStr = extractSpec(filename, result)
   var ss = newStringStream(specStr)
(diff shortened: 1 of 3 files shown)


testament/testament.nim
```diff
@@ -79,6 +83,20 @@ type
...
+    nimout: string ## compiler output
+    output: string ## test output
+
+    # information related to reported compiler errors (unused if
+    # none were reported):
+    file, msg: string
+    line, column: int
+
+    debugInfo: seq[string] ## debug info to give more context
+    err: TResultEnum       ## the reason the test failed (or succeeded)
```
zerbina, line 97: This is going to be turned into a boolean as part of one of the following refactors.

testament/specs.nim
```diff
@@ -431,33 +430,32 @@ proc parseSpec*(filename: string,
...
         of "win", "windows":
-          when defined(windows): result.err = reDisabled
+          result.disabledOs.incl OsPlatform.windows
         of "linux":
-          when defined(linux): result.err = reDisabled
+          result.disabledOs.incl OsPlatform.linux
         of "bsd":
-          when defined(bsd): result.err = reDisabled
+          result.disabledOs.incl {netbsd, freebsd, openbsd}
         of "osx", "macosx": # xxx remove `macosx` alias?
```
haxscramper, line 443: On the `macosx` alias `xxx`: we can print warning/here for the time being

haxscramper (approved):
Aside from one minor comments (technically unrelated), this looks good to me.

testament/testament.nim
```diff
@@ -79,6 +83,20 @@ type
...
+  CompilerOutput = object
+    ## Describes the output of a a compiler invocation.
+    cmd: string    ## the command that was run
+    nimout: string ## compiler output
+    output: string ## test output
+
+    # information related to reported compiler errors (unused if
+    # none were reported):
+    file, msg: string
+    line, column: int
```
haxscramper, line 94: And I think I would've used a dedicated type for the <file:line:col> triple (if we don't have suitable, then create) and put the whole thing as `Optional[]` so it would've be two explicitly optional fields. 


Or use 4-tuple of file:line:col:msg with the same wrapper.

haxscramper: But that's just IMO, so can be ignored

testament/testament.nim
```diff
@@ -1040,16 +1058,44 @@ func defaultTargets(category: Category): set[TTarget] =
...
+  let cmpString =
+    when defined(windows):  spec.file.replace(r"\", r"/")
+    else:                   spec.file
+
+  if retryContainer.retry and not retryContainer.names.anyIt(cmpString == it[0]):
+    # we're retrying failed test and the test is not in the retry-set
+    reDisabled
+  elif spec.parseErrors.len > 0:
+    reInvalidSpec
+  elif targetOS in spec.disabledOs or targetCPU in spec.disabledCpu:
```
saem, line 1072: this is much nicer.

testament/testament.nim
```diff
@@ -79,6 +83,20 @@ type
     startTime: float
     debugInfo: string
 
+  CompilerOutput = object
+    ## Describes the output of a a compiler invocation.
```
```suggestion
    ## Describes the output of a compiler invocation.
```

testament/testament.nim
```diff
@@ -79,6 +83,20 @@ type
...
+    cmd: string    ## the command that was run
```
saem, line 87: Perhaps it's not conveniently accessible enough for the time being, but shouldn't this be in the `TestRun` object?

testament/testament.nim
```diff
@@ -70,6 +71,9 @@ type
     startTime: float
     spec: TSpec
 
+    inCurrentBatch: bool ## whether the test is part of the current
```
saem, line 73: very small thing, entirely ignorable: I _believe_, if the blank line above this one is removed github has an easier time correctly rendering context text, otherwise the context snippet goes all the way upto `  type` on line 57/58.

testament/testament.nim
```diff
@@ -79,6 +83,20 @@ type
     startTime: float
     debugInfo: string
 
+  CompilerOutput = object
```
saem, line 85: Would `TestOutput` be better? As this combines compiler output and possibly also test binary execution output.

I'm not super fussed, it just occurred to me as I was reading and so I'm sharing, up to you.

testament/testament.nim
```diff
@@ -1040,16 +1058,44 @@ func defaultTargets(category: Category): set[TTarget] =
   else:
     standardTargets
 
-proc testSpec(r: var TResults, test: TTest) =
-  var expected = test.spec
-  if expected.parseErrors.len > 0:
-    r.addResult(test)
+proc computeEarly(spec: TSpec, inCurrentBatch: bool): TResultEnum =
+  ## Computes the early result for a test based on its `spec` and
+  ## the current configuration.
```
saem, line 1062: Nice, this is like a "pre-flight check".

saem (approved):
I like it, the small change introduces a good "seam" to break things apart along. :D

testament/specs.nim
zerbina: I've searched the tests, and it seems like there's no usage of `macosx` in any `disabled` specification. I'm going to make a follow-up for removing the alias.

testament/testament.nim
zerbina: Yep, I agree that bundling them makes sense, but I wanted to keep the amount of changes small, so decided against it.

testament/testament.nim
zerbina: `TestOutput` is how I had named it originally, but I later renamed it after noticing that that it's only used for capturing compiler output. Is there a place where it's used for execution output that I missed?

testament/testament.nim
zerbina: Yep, `TestRun` is better, placement wise. I have a next batch of refactors in the work, and, if okay with you, would include moving the field there.

saem: Yup, not at all insistent that this change be included in this PR. Sorry, should have clarified this was more observational wrt this PR.

zerbina:
Thank you for the review, @haxscramper and @saem.

testament/testament.nim
zerbina: It looks like that didn't help, but it was worth a try.

testament/testament.nim
saem: You're right, I assumed test output was being captured via the `output` field below, but that doesn't seem to be happening at this time. Also the check for actual binary execution output remains unchanged (last bit of `testSpecHelper`). Never mind. 😁

testament/specs.nim
```diff
@@ -317,6 +323,7 @@ proc initSpec*(filename: string): TSpec =
   result.file = filename
 
 proc isCurrentBatch*(testamentData: TestamentData; filename: string): bool =
+  # TODO: move elsewhere
```
krux02: It would be nice to elaborate on the reasons why things should be moved elsewhere. As personal reminders (in private projects) TODO's like this are fine, but in community projects they are very confusing when encountered by people who did not write them.

zerbina: Yep, that was an oversight on my side. I have adjusted the comment.

PR #772, opened 2023-06-22
testament: refactor further

zerbina:
## Summary

Further refactor the `testament` implementation, with the goal of
making it a bit easier to maintain and extend.

There are two notable changes in behaviour:
- `macosx` is no longer supported for the `disabled` key in test
  specifications; `osx` should be used instead
- for `run` actions, the compiler output checks are now performed
  *first*, instead of *after* all other test output checks

## Details

The overall theme is moving decision making about what to do with test
results and interacting with the backend into `testSpecHelper`.

This is mainly achieved by performing output-to-backend step
(`addResult`) in at the end of `testSpecHelper`, with the various output
validation routines now returning a `TestResult` object instead of
outputting the result themselves. As a side-effect, duplicated logic
(incrementing the `passed` counter) is removed and the validation
routines become easier to re-use.

In addition:
- move timeout handling out of `addResult` and into `testSpecHelper`
- make the shell command for executing the compiler (`cmd`) a property
  of `TestRun` (it was previously misplaced in `CompilerOutput`)
- run compiler output checks for `run` actions before the other checks,
  which removes a duplicated code path and makes the behaviour more
  intuitive
- remove the `macosx` value for the `disabled` key  
- `CompilerOutput` now represents the "success" state with a bool
  instead of a `TResultEnum`

### DIFF
--- a/testament/testament.nim
+++ b/testament/testament.nim
@@ -577,22 +577,15 @@ proc addResult(
     r: var TResults,
     run: TestRun,
     expected, given: sink string,
-    successOrig: TResultEnum,
+    success: TResultEnum,
     output: ptr CompilerOutput = nil,
     outCompare: TOutCompare = nil
   ) =
   ## Report final test run to backend, end user (write to command-line) and etc
   let
-    duration = epochTime() - run.startTime
-    timeout = run.expected.timeout
-    success =
-      if timeout > 0.0 and duration > timeout:
-        reTimeout
-      else:
-        successOrig
     targetStr = $run.target
     param = ReportParams(
-      duration: duration,
+      duration: epochTime() - run.startTime,
       name: run.getName(),
       origName: run.test.name,
       cat: run.test.cat.string,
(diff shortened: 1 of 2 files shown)


testament/testament.nim
```diff
@@ -847,82 +840,73 @@ proc needsCodegenCheck(run: TestRun): bool =
...
     test: TTest,
     target: TTarget,
-    spec: TSpec,
-    expectedMsg: var string,
-    given: var CompilerOutput
-  ) =
+    spec: TSpec
+  ): TestResult =
   ## Check for any codegen mismatches in file generated from `test` run.
   ## Only file that was immediately generated is tested.
```
saem, line 846: Just remembered how buggy the pre-existing code is wrt reporting all the various issues, code size errors can clobber codegen errors that were detected and only the last check is retained on mismatch.

Soft suggestion: might be worth leaving a `TODO` for someone else or future PR in this series.

```suggestion
  ## Only file that was immediately generated is tested.

  # xxx: only reports last failure, an iterator or early exit might be better?
```

saem (approved):
Very minor suggestion, feel free to ignore it -- I can always come back to it later.

PR #773, opened 2023-06-23
disallow using `.compileTime` locations at run-time

zerbina:
## Summary

Change the `.compileTime` pragma to mean "only usable in a compile-time
context" for locations, with an error being reported if used outside of
one.

This makes the pragma's meaning consistent between locations and
procedures, and removes a special case from the code generators.

## Details

* remove the section documenting the feature from the manual
* add a  report, and integrate the diagnostic into
  `semSym`
* detect and report disallowed usages of `.compileTime` location
  in `sempass2.useVarNoInitCheck`
* remove the implementation from the code generators
* adjust the tests

Two `sempass2` issues had to be fixed in order for the change to work:
* `bind`, `mixin`, and `import`/`export` statements were treated as
  imperative code
* the definitions of `.compileTime` globals were incorrectly treated as
  normal run-time code

In parts, the changes represent a revert of commit https://github.com/nim-works/nimskull/commit/a5e2db2ac53aa61d0e050ee6487e38c647eb442d .

While the original intention seems to have been persisting the
variable's value at the end of compilation over to run-time, the feature
never worked like that, and stopped working at all when the VM moved
away from `PNode` as its data representation.

### `sempass2` vs. sem

The reason the detection is implemented in `sempass2`, as opposed to
during early semantic analysis, is that it's much easier to check for
*location* (not symbol) usage during the second semantic pass. `semSym`
cannot be used here, as that would prevent passing symbols of
`.compileTime` variables to `typed` macro or template parameters.


saem (approved):
👍

It's a rather useful idiom, but I think its design is rather off. It should be redone from first principles, the example in the manual requires a project level global to hold a value that's mutated until the very end of sem and then we should probably emit a `let` variable, or make that controllable. I think that facility having a separate name makes more sense rather than adding more to behaviour to `compiletime`.

zerbina:
> ... and then we should probably emit a `let` variable, or make that controllable. I think that facility having a separate name makes more sense rather than adding more to behaviour to `compiletime`.

That was/is exactly my thinking :D

PR #774, opened 2023-06-27
internal: make `PEvalContext` a standalone type

zerbina:
## Summary

As the first step towards splitting up `vmdef.TCtx` into multiple types,
re-purpose the `PEvalContext` type to represent all state and data
required for running code at compile-time (which is what `TCtx`
previously represented).

## Details

* move `PEvalContext` to the `compilerbridge` module and make it an
  object type
* the `ModuleGraph.vm` now stores an instance of `PEvalContext` instead
  of `PCtx`
* since it's not necessary anymore, don't use inheritance with
  `TCtx`
* remove the `PCtx` type and all usages thereof (this includes the
  deprecated `registerCallback` overload)
* adjust all usage sites of `ModuleGraph.vm`

The `sealRodFile` procedure, which accesses the `PEvalContext`, cannot
be part of the `semdata` module anymore, as that would introduce a
cyclic import dependency. It is moved to `sem`, where its only usage
site is.

`TCtx` no longer representing the "compile-time execution context" makes
it possible to split the type up.

### DIFF
--- a/compiler/front/scripting.nim
+++ b/compiler/front/scripting.nim
@@ -49,14 +49,14 @@ from compiler/vm/vmlegacy import legacyReportsVmTracer
 from std/strutils import cmpIgnoreStyle, contains
 
 proc setupVM*(module: PSym; cache: IdentCache; scriptName: string;
-              graph: ModuleGraph; idgen: IdGenerator): PEvalContext =
+              graph: ModuleGraph; idgen: IdGenerator): TCtx =
   # For Nimble we need to export 'setupVM'.
-  result = newCtx(module, cache, graph, idgen, legacyReportsVmTracer)
+  result = initCtx(module, cache, graph, idgen, legacyReportsVmTracer)
   # for backwards compatibility, allow meta expressions in nimscript (this
   # matches the previous behaviour)
   result.flags = {cgfAllowMeta}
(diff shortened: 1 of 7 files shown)


compiler/front/scripting.nim
```diff
@@ -213,10 +213,10 @@ proc runNimScript*(cache: IdentCache; scriptName: AbsoluteFile;
   # - the ``vmopsDanger`` option has no effect on callbacks registered
   #  during `setupVM`
   # - NimScript has access to the macro/compile-time APIs
-  registerAdditionalOps(vm[], disallowDanger)
```
saem, line 216: Dropping all the derefs is a good sign. 😄

compiler/sem/sem.nim
```diff
@@ -974,6 +974,14 @@ proc appendInstancedGenericRuntimeRoutines(c: PContext, n: PNode) =
         unreachable()
   c.lastGenericIdx = c.generics.len
 
+proc sealRodFile(c: PContext) =
```
saem, line 977: Definitely better, moving away from wide (interface) and/or shallow (minimal implementation/decision encapsulation) is the way to go.

saem (approved):

PR #775, opened 2023-06-27
support `method`s with the VM backend

zerbina:
## Summary

Enable dispatcher generation for the VM backend, which makes `method`
(including multi-methods) available for the standalone VM target. Note
that this doesn't apply to compile-time execution, where using `method`s
is still unsupported.

## Details

* move the `chckNilDisp` compilerproc into the `system` module and
  replace the `== nil` with `isNil` (more on that later)
* only report errors on method usage from `vmgne` when not in the
  `standalone` mode
* rename the `realType` procedure to `routineSignature`
* generate the method dispatcher prior to generating code in
  `vmbackend`
* where possible, enable the `method` tests for the VM target

The VM code generator (`vmgen`) previously checked for whether a
procedure returns something by testing for the presence of a procedure's
`result` AST slot. For methods - where the AST always has a `result`
slot, regardless of whether they have a return value - this test is
no longer enough, and is replaced with inspecting the return type.

Regarding the `== nil` -> `isNil` change: the VM comparison operator for
pointer-like types (which `==` maps to) requires both operands to be of
same dynamic type, but in `chckNilDisp`, this is not necessarily the
case (for example, when the dispatched-through type is a `ref` type).
The proper solution would be for `cgmeth.genDispatcher` to wrap the
argument in a `cast[pointer]`, but the VM doesn't yet support casting
references to pointers.

### DIFF
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -186,6 +186,11 @@ proc genRegLoad(c: var TCtx, n: PNode, dest, src: TRegister)
 template isUnset(x: TDest): bool = x < 0
 
 
+proc realType(s: PSym): PType {.inline.} =
+  ## Returns the signature type of the routine `s`
+  if s.kind == skMacro: s.internal
+  else:                 s.typ
+
 func underlyingLoc(n: PNode): PSym =
   ## Computes and returns the symbol of the complete location (i.e., a location
   ## not part of a compound location) that l-value expression `n` names. If no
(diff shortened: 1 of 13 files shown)


compiler/vm/vmgen.nim
```diff
@@ -186,6 +186,11 @@ proc genRegLoad(c: var TCtx, n: PNode, dest, src: TRegister)
 template isUnset(x: TDest): bool = x < 0
 
 
+proc realType(s: PSym): PType {.inline.} =
```
```suggestion
proc routineSignature(s: PSym): PType {.inline.} =
```

Very soft suggestion, inspired by the doc comment.

saem (approved):
the unification is really highlighted as the tests are applied across all backends.

PR #776, opened 2023-06-27
testament: support executing "known issue" tests

### LINKED ISSUE #61: testament add `knownIssue` and simplify `disabled`

zerbina:
## Summary

Introduce the `--tryFailing` switch for testament, which enables
execution of tests marked as having known issues (and are thus expected
to fail). A "known issue" test successfully finishing in this mode
counts as a test failure.

The plan is to enable this new switch in the CI, in order to
automatically notify about tests that start working after a change. It
is not yet enabled in the CI yet, however.

## Details

* add the `--tryFailing` CLI switch, which sets the `runKnownIssue`
  execution flag
* add the `reUnexpectedSuccess` test result code
* render expectedly failing or untried "known issue" tests as
  `KNOWNISSUE` instead of `SKIP` or `NOTINBATCH` in the output

Refer to https://github.com/nim-works/nimskull/issues/61 for more information.

### DIFF
--- a/doc/testament.rst
+++ b/doc/testament.rst
@@ -43,6 +43,8 @@ Options
                           By default turned on.
 --skipFrom:file           Read tests to skip from ``file`` - one test per
                           line, # comments ignored
+--tryFailing              Run tests marked as "known issue" and verify that
+                          they're still failing
 
 
 Running a single test
--- a/testament/categories.nim
+++ b/testament/categories.nim
@@ -515,7 +515,7 @@ proc runJoinedTest(r: var TResults, targets: set[TTarget], testsDir, options: st
 
 proc processCategory(r: var TResults, cat: Category, targets: set[TTarget],
                      options, testsDir: string,
-                     runJoinableTests: bool) =
+                     runJoinableTests, runKnownIssues: bool) =
   let cat2 = cat.string.split("/")[0].normalize
   case cat2
   of "js":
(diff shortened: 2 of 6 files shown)


saem (approved):
Keeping the code changes small via a boolean flag is a good approach for the interim.

PR #777, opened 2023-06-27
compiler: unify the pre-code-gen processing

zerbina:
## Summary

Introduce a unified backend processing pipeline (the `process` iterator)
based on the implementation of the `vmbackend` module, and integrate it
into all three code generation orchestrators.

Discovery of alive entities (which automatically includes dead-code
elimination), applying transformations, and various other backend-
agnostic processing are now all implemented via a single, shared
implementation, making them much easier to maintain and adjust.

As a side-effect of the reworked lifted global (globals defined in
procedure via the `.global` pragma) handling, multiple issues and bugs
with them are fixed:
- initialization and destruction order for them is the same across all
  backends
- JavaScript backend: lifted global use the pre-init mechanism for
  lifted globals now, meaning that they are initialized during the
  attached-to module's pre-init phase, instead of when control-flow
  first reaches the `var|let` statement
- names of both normal and lifted globals defined inside a for-loop
  always refer to the same location now, regardless of whether the used
  inline iterator contains multiple `yield` statement
- the name of a lifted global defined inside an `.inline` iterator
  always refers to the exact same location across all usages of the
  iterator (each inlined occurrence got its own copy, previously)
- locals defined in the initializer expression of lifted globals are now
  considered by the move analyser and destructor injection. For example,
  the `s` in `var g {.global.} = (block: (var s = @[...]; 1)) ` is
  now destroyed

In addition:
- constants marked with `exportc` are now always included in the
  generated code, even when not explicitly used
- a regression with `--header` not including the prototypes of
  exportc'ed globals is fixed
- the VM backend doesn't generate code for `.compileTime` globals
  anymore
- custom `.dynlib` procedure loading logic is run during the data-init
  phase of a module, instead of as part of the module-init procedure

## Details

The key changes are:
- the code generators don't implement discovery of alive entities
  anymore
- discovery of alive procedures now uses an *iterative* approach instead
  of a *recursive* one for the C and JS backends (the VM backend already
  did)
- the code generators are passed the fully transformed AST as input
- emitting the definition code for globals is now fully the
  responsibility of the orchestrators
- the module destructor produced by `modulelowering` is no longer
  modified in the backend
- destruction of lifted globals happens in the new "post-destructor"
  module-bound operator
- calls to the pre-init and post-destructor operators are now directly
  emitted into the main procedure
- the orchestrators no longer directly call `generateMethodDispatcher`
- locations marked with `.global` are consistently treated as lifted
  globals (even top-level locations) across the backends

In addition, the module-init procedure is now named the same as the
module it belongs to and longer has an "Init" postfix. This gets around
having to introduce a special case for the init procedure in the C code
generator, now that it treats the module-init procedure like a normal
procedure.

### The `process` iterator

At the center of the additions is the `process` inline iterator (which
represents the unified processing pipeline). It implements the discovery
of alive entities, preparing the procedures for code generation (by
invoking `transf` and applying the MIR passes), generating method
dispatchers, and lifting the pre-init and post-destructor operators.

Whenever interaction with the caller is required, the iterator produces
(yields) an event that contains the information necessary for the caller
to act on it. The caller (the orchestrators) then, if necessary, pre-
processes the newly discovered entities and, if the event is about code
becoming available, invokes the code generator. Newly discovered
entities are communicated through the `DiscoverData` structure that both
the iterator and caller have access to.

The iterator's implementation and design is complicated by two things:
late dependencies and (to a lesser degree) first-seen-in-module
tracking.

#### Late Dependencies

A late dependency of a procedure (or code fragment in general) is a
reference to some procedure that is not visible by scanning the MIR
representation. They occur when a code generator emits a call to a
`.compilerproc`. The general goal is to remove raising late dependencies
from `cgen` and `jsgen`, but until then, the `process` iterator has to
support the case where the caller registers new procedures with
`DiscoverData`.

#### First-seen-in-module Tracking

For the inline procedure handling of the C backend to work, it needs to
know about at least one module an inline procedure is used in. The
first-seen-in-module is used for this, but the unified processing
pipeline has to track it.

### `cgen` and `cbackend`

The key changes / important things to note are:
- emitting procedures into the correct C file section is now done by the
  orchestrator
- inline procedures are now only transformed a single time (if used),
  instead of once for each module they're used in
- header file generation (when using the `--header` option) is fully
  managed by the orchestrator
- pre-init procedure generation and everything related to lifted
  globals is removed from the code generator
- `.dynlib` procedure handling is still mostly handled by the code
  generator
- all dedicated module-init procedure is removed; they are, for the
  largest part, treated as normal procedures now
- all dedicated top-level statement processing is removed from the code
  generator

Other changes and internal restructurings:
- introduce the internal `sfTopLevel` flag and use it to: emit a
  `nimTestErrorFlag` call at the exit of a flagged procedure,
  special casing of `emit` and `asm` statements
- split `genProcAux` into `startProc` and `finishProc`
- move code generation logic for the temporary `TNimNode` storage into
  `genDatInitCode` and remove `genInitCode`
- remove support for alive information provided by `dce.nim`, as the
  code generator is not responsible anymore for deciding which
  procedures need to be processed

Making sure that a duplicate of an inline procedure is emitted into each
C file the procedure is used in doesn't happen automatically anymore and
thus requires special support by the orchestrator.

### `jsgen` and `jsbackend`

- split `genProc` into `startProc` and `finishProc`
- integrate and adjust to the `process` iterator
- remove the DCE related bits from `jsgen`

### `vmbackend`

- integrate and adjust to the `process` iterator
- remove the now-obsolete types and procedures

Constants now use the sequence provided by `DiscoverData`, removing a
usage of the `vmdef.LinkState` type from `vmbackend`.

### `transf` and MIR processing

- introduce a dedicated `goIsCompileTime` option for `mirgen`, so that
  definitions of `.compileTime` globals can be properly ignored when
  using the VM backend 
- remove the workarounds related to lifted globals from the `mirbridge`
  and `injectdestructors` modules
- the transformed body of `.inline` procedures is no longer
  unconditionally cached

With the last bits of the new-style DCE (the one implemented by the
`dce` module) support removed from `cgen`, two MIR simplification are
possible:
- `nkConstSection` are discarded by `mirgen`
- the `def` for a procedure doesn't carry the procdef anymore

#### Duplicated Globals

There was a problem regarding globals in `transf`. When introducing
fresh symbols in the context of iterator inlining, globals were also
copied. While that did work previously, it no longer can, as the whole
backend processing now expects all module-level globals to be known
prior to `transf`.

Keeping the original symbol is possible for to-be-lifted globals
(duplicate definitions of those are filtered out by
`produceFragmentsForGlobals`), but not for module-level globals, as the
`injectdestructors` pass requires the latter to have only a single
definition. Therefore, module-level globals are still copied, but
they're now associated with the original symbol, and the
`rewriteGlobalDefs` procedure uses this information to restore the
original symbols after the `injectdestructors` has run.

This also fixes the pre-existing bug where the name of a global defined
inside a for-loop body referred to different locations when the for-
loop body was duplicated (due to inline iterators with multiple
`yield` statements).

### Other

- don't place module-level globals in a module's struct, and reparent
  them to the module's init procedure -- this makes them appear as
  lifted globals to `transf.extractGlobals` (what to do with them was
  previously up to the used code generator) 
- remove the `ModuleGraph.globalDestructors` list and usages thereof;
  the unified backend processing renders it obsolete

### Misc

- implement the `merge` operation for the `Store` container type
- implement the `append` operation for `SourceMap`
- add new iterators for the `SeqMap` and `Store` types

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -506,6 +506,20 @@ func discoverFrom*(data: var DiscoveryData, decl: PNode) =
     else:
       unreachable(n.kind)
 
+func discoverFromValueAst(data: var DiscoveryData, ast: PNode) =
+  ## Discover new routines from `ast`, which is an AST representing a value
+  ## construction expression.
+  case ast.kind
+  of nkSym:
+    let s = ast.sym
+    if s.kind in routineKinds:
+      register(data, procedures, s)
+  of nkWithSons:
+    for n in ast.items:
+      discoverFromValueAst(data, n)
+  of nkWithoutSons - {nkSym}:
+    discard "nothing to do"
+
 func queue(iter: var ProcedureIter, prc: PSym, m: FileIndex) =
   ## If eligible for processing and code generation, adds `prc` to
   ## `iter`'s queue.
@@ -521,20 +535,6 @@ func queueAll(iter: var ProcedureIter, data: var DiscoveryData,
   ## processed/read.
   for _, it in visit(data.procedures):
     queue(iter, it, origin)
-
-func discoveryFromValue(data: var DiscoveryData, ast: PNode) =
-  ## Discover new routines from `ast`, which is an AST representing a value
-  ## construction expression.
-  case ast.kind
-  of nkSym:
-    let s = ast.sym
-    if s.kind in routineKinds:
-      register(data, procedures, s)
-  of nkWithSons:
-    for n in ast.items:
-      discoveryFromValue(data, n)
-  of nkWithoutSons - {nkSym}:
-    discard "nothing to do"
 
 # ----- ``process`` iterator implementation -----
 
@@ -560,7 +560,7 @@ func processConstants(data: var DiscoveryData): seq[(FileIndex, int)] =
   ## and marks the constants as processed.
   assert data.procedures.isProcessed
   for _, s in peek(data.constants):
-    discoveryFromValue(data, astdef(s))
+    discoverFromValueAst(data, astdef(s))
     # the procedure needs to be queued from the context of the module `s` is
     # attached to:
     let m = moduleId(s).FileIndex
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -42,11 +42,13 @@ proc generateMain*(graph: ModuleGraph, modules: ModuleList, result: PNode) =
     # the system module is special cased: its fully initialized during the
     # data-init phase
     if sfSystemModule in it.sym.flags:
+      emitOpCall(graph, it.preInit, result)
       emitOpCall(graph, it.init, result)
 
   # then the modules are initialized and their module-level code executed
   for it in closed(modules):
     if sfSystemModule notin it.sym.flags:
+      emitOpCall(graph, it.preInit, result)
       emitOpCall(graph, it.init, result)
 
 proc generateTeardown*(graph: ModuleGraph, modules: ModuleList, result: PNode) =
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -76,7 +76,7 @@ type
     ## Combines a sequence of items with a "read" cursor.
     data: seq[T]
     progress: int
-      ## remembers the the position until which the items have already
+      ## remembers the position until which the items have already
       ## been processed
 
   DiscoveryData* = object
@@ -428,7 +428,7 @@ proc produceFragmentsForGlobals(data: var DiscoveryData, identdefs: seq[PNode],
         if it[2].kind == nkEmpty:
           # no explicit initializer expression means that the default value
           # should be used
-          # XXX: ^^ it'd make to instead let semantic analysis ensure this
+          # XXX: ^^ it'd make sense to instead let semantic analysis ensure this
           #      (i.e. by placing a ``default(T)`` in the initializer slot)
           r.add(it[2]): MirNode(kind: mnkArgBlock)
           r.add(it[2]): MirNode(kind: mnkEnd, start: mnkArgBlock)
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -59,38 +59,10 @@ proc generateTeardown*(graph: ModuleGraph, modules: ModuleList, result: PNode) =
   for it in rclosed(modules):
     if sfSystemModule notin it.sym.flags:
       emitOpCall(graph, it.destructor, result)
+      emitOpCall(graph, it.postDestructor, result)
 
   emitOpCall(graph, systemModule(modules).destructor, result)
-
-proc finishDeinit*(graph: ModuleGraph, modules: var ModuleList) =
-  ## Appends destructor calls for lifted globals (those marked with
-  ## ``.global``) to the bodies of the modules' de-init procedures. Called at
-  ## the very end of code generation once all alive lifted globals have been
-  ## discovered.
-
-  # XXX: if the called destructors access other globals, or introduce new
-  #      ones, behaviour is undefined. In the best but most unlikely case,
-  #      everything will work as expected, but in the worst case, the code
-  #      is going to misbehave at run-time
-
-  proc prepare(prc: PSym): PNode =
-    # XXX: not pretty, but it's going to removed again anyway
-    case prc.ast[bodyPos].kind
-    of nkEmpty:
-      result = newNodeI(nkStmtList, prc.ast[bodyPos].info)
-      prc.ast[bodyPos] = result
-    of nkStmtList:
-      # already supports appending
-      result = prc.ast[bodyPos]
-    else:
-      result = newTree(nkStmtList, prc.ast[bodyPos])
-      prc.ast[bodyPos] = result
-
-  # XXX: the whole ``globalDestructors`` mechanism is wrong. Moving dead-code
-  #      elimination into the orchestrators will, among other things, allow
-  #      for removing it
-  for (module, call) in ritems(graph.globalDestructors):
-    prepare(modules[module.FileIndex].destructor).add call
+  emitOpCall(graph, systemModule(modules).postDestructor, result)
 
 proc generateMainProcedure*(graph: ModuleGraph, idgen: IdGenerator,
                             modules: ModuleList): PSym =
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -555,6 +555,47 @@ proc next(iter: var ProcedureIter, graph: ModuleGraph,
     # apply all MIR passes:
     process(result.prc, graph, idgen)
 
+proc preprocessDynlib(graph: ModuleGraph, idgen: IdGenerator,
+                      sym: PSym, deps: var seq[PSym]) =
+  # HACK: so that procedures and unexpanded constants used in the
+  #       expressions passed to `.dylib` pragmas are discovered, we
+  #       translate the expression to MIR code, scan it, and then
+  #       translate it back to AST and update the symbol with it. This is
+  #       horrendous, but fortunately, this hack (`preprocessDynlib``) can
+  #       be removed once handling of dynlib procedures and globals is fully
+  #       implemented in the ``process`` iterator
+  if lfDynamicLib in sym.loc.flags:
+    if sym.annex.path.kind in nkStrKinds:
+      # it's a string, no need to transform nor scan it
+      discard
+    else:
+      # XXX: the logic here ignores a large amount of things
+      #      (options, proper owner symbol, etc.)...
+      var
+        t: MirTree
+        m: SourceMap
+
+      generateCode(graph, {}, sym.annex.path, t, m)
+      for dep in deps(t, {}): # just ignore magics here
+        if dep.kind in routineKinds + {skConst}:
+          deps.add dep
+
+      sym.annex.path = generateAST(graph, idgen, sym.owner, t, m)
+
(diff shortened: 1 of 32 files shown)


zerbina:
As mentioned, this PR is a revised version of #550. Too many merge conflicts started to accumulate, and a lot of things changed regarding the architecture of some related compiler bits, so I decided to start over. Between this PR and the original one, there are a few significant differences.

### High-level iterator instead of low-level, manual one

With the predecessor PR, the `ProcedureIter` was the central type for the unified backend processing pipeline. It represented the manual implementation of a closure iterator, with the `hasNext` and `next` procedure being used to query and advance the iterator.

State-wise, `ProcedureIter` managed a set of all seen *procedures*, a queue of what to process and return next, and some configuration state. The intention behind this design was to allow for maximum flexibility in how dependencies are discovered.

However, with all three orchestrators ending up implementing discovery in mostly the same way (with only `cbackend`, due to the inline procedure handling, diverging slightly), this design only lead to a lot of duplication. The `ProcedureIter` type still exists, but it only represents the work queue now, and is not used outside of the `process` iterator.

### No special handling for inner procedures

When the predecessor PR was created, the JS backend didn't use the lambda-lifting pass, so having dedicated processing logic for inner procedures was necessary -- I also still had the plan of implementing lambda-lifting as an MIR pass.

With the JS backend using the lambda-lifting pass now, and me having grown increasingly unsure about the "lambda-lifting as an MIR pass" approach, I ultimately decided to remove the special handling, which significantly simplified the unified backend processing pipeline

### First-seen-in-module tracking

Because it is, strictly speaking, a layering violation to have the unified processing pipeline track the first-seen-in-module for inline procedures, this was left to the orchestrator (`cbackend`) in the predecessor PR. The complexity of this approach, and the divergence from the processing of the other orchestrator it implied, were not worth the cleaner architecture, so I made the tracking a part of the unified processing pipeline.

### Entity discovery

The unified processing pipeline as introduced by the predecessor PR was only concerned with procedures, leaving discovery, etc. of globals and constants to the orchestrators. With discovery of procedures already being handled by `process`, it made sense to also let it take care of the discovery of global and constants. For constants (which can reference procedures), this was also necessary.

zerbina:
It's almost there, only a fix for the regression with `.global` for module-level definitions is missing now.

tests/global/tlifted_global_in_loop_body.nim
```diff
@@ -0,0 +1,49 @@
...
+      var g {.global.} = 1
+      inc g
+      result = g
+
+  doAssert test() == 2
+  # verify that the global's value persists across calls:
+  doAssert test() == 3
+
+block multi_yield:
+  # test with an inline iteratong that has multiple yield statements
```
saem:
```suggestion
  # test with an inline iterator that has multiple yield statements
```

tests/global/tlifted_global_in_inline_iterator.nim
```diff
@@ -0,0 +1,29 @@
...
+
+proc test2(): (int, ptr int) =
+  for r in iter():
+    return r
+
+let (val, address) = test1()
+doAssert val == 2
+
+# invoking the iterator again must not reset the global to its initial
+# value
```
saem:
```suggestion
# value, nor move the global (same address)
```

Figured it'd be worth spelling out both things being checked here.

tests/global/tlifted_global_in_inline_iterator.nim
```diff
@@ -0,0 +1,29 @@
+discard """
+  targets: "c js vm"
+  description: '''
+    Tests to make sure that lifted globals defined inside inline iterators work
```
saem:
```suggestion
    Ensure lifted globals defined inside inline iterators work
```

Shorter, if you prefer.

saem (commented):
Partial review: did a first pass just on the tests themselves.

Now going over the implementation.

compiler/utils/containers.nim
```diff
@@ -60,6 +65,17 @@ iterator pairs*[K, V](m: SeqMap[K, V]): (K, lent V) =
       yield (K(i), m.data[i])
     inc i
 
+iterator mpairs*[K, V](m: var SeqMap[K, V]): (K, var V) =
+  ## Returns, in an unspecified order, the key and value for each entry in the
+  ## map `m`.
```
saem:
```suggestion
  ## Returns, in an unspecified order, the key and mutable value for each entry
  ## in the map `m`.
```

compiler/utils/containers.nim
```diff
@@ -97,6 +107,34 @@ iterator pairs*[I; T](x: Store[I, T]): (I, lent T) =
...
     inc i
 
+func add*[I; T](x: var Store[I, T], it: sink T): I {.inline.} =
+  ## Appends a new item to the Store and returns the ID assigned to
+  ## it
+  rangeCheck x.data.len.BiggestUInt < high(I).BiggestUInt
+  x.data.add it
+  result = I(x.data.high)
+
+iterator mitems*[I; T](x: var Store[I, T]): var T =
```
saem, line 117:
```suggestion
iterator mitems*[I; T](x: var Store[I, T]): var T =
  ## Yields a mutable item for each entry in `x`.
```

compiler/sem/transf.nim
```diff
@@ -1312,14 +1327,25 @@ proc transformBody*(g: ModuleGraph; idgen: IdGenerator; prc: PSym; cache: bool):
...
-      # it is important to transform exactly once to get sym ids and locations right
+    if cache:
       prc.transformedBody = result
     else:
       prc.transformedBody = nil
     # XXX Rodfile support for transformedBody!
 
+proc transformBodyWithCache*(g: ModuleGraph, idgen: IdGenerator, prc: PSym): PNode =
+  ## Transforms the body of `prc` and returns, but doesn't cache, the resulting
+  ## AST. If the transformed body was already cached earlier,
```
saem:
```suggestion
  ## Fetches the cached transformed body of `prc`, transforming it if not
  ## available, new transforms are not cached
```

compiler/mir/mirbridge.nim
```diff
@@ -61,31 +52,43 @@ let reprConfig = block:
...
-proc echoOutput(config: ConfigRef, owner: PSym, body: PNode) =
+proc echoOutput*(config: ConfigRef, owner: PSym, body: PNode) =
   ## If requested via the define, renders the output AST `body` and writes the
   ## result out through ``config.writeLine``.
   if config.getStrDefine("nimShowMirOutput") == owner.name.s:
     writeBody(config, "-- output AST: " & owner.name.s):
       config.writeln(treeRepr(config, body, reprConfig))
 
-proc rewriteGlobalDefs(body: var MirTree, sourceMap: var SourceMap,
+proc restoreGlobal(s: PSym): PSym =
```
saem: Could consider making this an inner proc of `rewriteGlobalDefs` if I'm reading things correctly.

compiler/backend/backends.nim
```diff
@@ -1,28 +1,197 @@
...
+    noImported*: bool
+      ## if ``true``, indicates that a procedure with a body should not be
+      ## treated as imported, even if it's marked as such
+
+  ProcedureIter = object
+    queued: Deque[tuple[prc: PSym, fr: FileIndex]]
+      ## procedures that are queued for code-generation
+    config: BackendConfig
+
+  Queue*[T] = object
```
saem, line 75: it's like a baby kafka. 👶🏽

compiler/backend/backends.nim
```diff
@@ -124,4 +267,523 @@ proc generateMainProcedure*(graph: ModuleGraph, idgen: IdGenerator,
...
+        prepare(r, graph.emptyNode)
+        r.add(it): MirNode(kind: mnkArgBlock)
+        r.add(it[0]): MirNode(kind: mnkGlobal, sym: s, typ: s.typ)
+        r.add(it[0]): MirNode(kind: mnkTag, typ: s.typ)
+        r.add(it[0]): MirNode(kind: mnkName, typ: s.typ)
+        if it[2].kind == nkEmpty:
+          # no explicit initializer expression means that the default value
+          # should be used
+          # XXX: ^^ it'd make to instead let semantic analysis ensure this
+          #      (i.e. by placing a ``default(T)`` in the initializer slot)
```
saem:
```suggestion
          # XXX: ^^ it'd make sense to instead let semantic analysis ensure this
          #      (i.e. by placing a ``default(T)`` in the initializer slot)
```

saem (approved):
Basically done reviewing, I wanted to give `backends` another pass mostly for improvements like docs, simplifications, etc... nothing major/blocking.

I'm good with it merging before hand.

compiler/backend/backends.nim
```diff
@@ -1,28 +1,197 @@
...
+  ProcedureIter = object
+    queued: Deque[tuple[prc: PSym, fr: FileIndex]]
+      ## procedures that are queued for code-generation
+    config: BackendConfig
+
+  Queue*[T] = object
+    ## Combines a sequence of items with a "read" cursor.
+    data: seq[T]
+    progress: int
+      ## remembers the the position until which the items have already
```
saem:
```suggestion
      ## remembers the position until which the items have already
```

saem (approved):

compiler/mir/mirbridge.nim
zerbina: Yep, will do.

compiler/sem/transf.nim
zerbina: That's much easier to parse, thanks.

PR #778, opened 2023-06-28
internal refactor: slim `sem` module

saem:
## Summary

Clean-up the `sem` module, dropping deadcode, moving single use exported
procedures to the use sites, and making one off procedures inner procs.
This is a small step, of many, to reduce dependencies between modules
and public surface area.

## Details

Changes:
- moved `getIdentLineInfo` into `ast_query`
- `isArrayConstr` was only used in `semexprs`, moved there
- `semExprFlagDispatched` move into `semobjconstr`, only user
- move `symFromTyp` to `semtypes`
- dropped `semNodeFromType` as inline in single usage in `semexprs`
- collapsed `typeAllowedCheck` into `paramsTypeCheck

Made inner procs, as they're one offs:
- `resetSemFlag` into `semAfterMacroCall`
- `isImportSystemStmt` & `isEmptyTree` into `semStmtAndGenerateGenerics`

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -489,8 +489,6 @@ proc semIdentVis(c: PContext, kind: TSymKind, n: PNode,
 proc semIdentWithPragma(c: PContext, kind: TSymKind, n: PNode,
                         allowed: TSymFlags): PSym
 
-proc getIdentLineInfo(n: PNode): TLineInfo
-
 proc typeAllowedCheck(c: PContext; info: TLineInfo; typ: PType; kind: TSymKind;
                       flags: TTypeAllowedFlags = {}) =
   let t = typeAllowed(typ, kind, c, flags)
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -204,7 +204,6 @@ proc fitNode(c: PContext, formal: PType, arg: PNode; info: TLineInfo): PNode =
 
   if arg.typ.isNil:
     c.config.localReport(arg.info, reportAst(rsemExpressionHasNoType, arg))
-
     # error correction:
     result = copyTree(arg)
     result.typ = formal
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -518,16 +518,6 @@ proc semTemplateExpr(c: PContext, n: PNode, s: PSym,
 proc semMacroExpr(c: PContext, n: PNode, sym: PSym,
                   flags: TExprFlags = {}): PNode
 
-proc symFromType(c: PContext; t: PType, info: TLineInfo): PSym =
-  if t.sym != nil: return t.sym
-  result = newSym(skType, getIdent(c.cache, "AnonType"), nextSymId c.idgen, t.owner, info)
-  result.flags.incl sfAnon
-  result.typ = t
-
-proc symNodeFromType(c: PContext, t: PType, info: TLineInfo): PNode =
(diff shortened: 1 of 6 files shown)


compiler/sem/sem.nim
```diff
@@ -525,46 +518,15 @@ proc semTemplateExpr(c: PContext, n: PNode, s: PSym,
...
-  result = newSym(skType, getIdent(c.cache, "AnonType"), nextSymId c.idgen, t.owner, info)
-  result.flags.incl sfAnon
-  result.typ = t
-
-proc symNodeFromType(c: PContext, t: PType, info: TLineInfo): PNode =
-  result = newSymNode(symFromType(c, t, info), info)
-  result.typ = makeTypeDesc(c, t)
-
 proc hasCycle(n: PNode): bool =
+  # xxx: this isn't used, we should consider reinstating this?
```
zerbina, line 522: There is the `cyclicTree` procedure in the `trees` module, which compute the same thing as `hasCycle`, but without temporarily mutating `n`. While `hasCycle` is probably a bit faster than `cyclicTree`, I think it's okay to drop `hasCycle`.

compiler/sem/sem.nim
```diff
@@ -696,6 +634,17 @@ proc semAfterMacroCall(c: PContext, call, macroResult: PNode,
   ## reassigned, and binding the unbound identifiers that the macro output
   ## contains.
   c.config.addInNimDebugUtils("semAfterMacroCall", s, macroResult, result)
+
+  proc resetSemFlag(n: PNode) =
```
```suggestion
  proc resetSemFlag(n: PNode) {.nimcall.} =
```

In order to prevent it accidentally closing over some outer local.

compiler/sem/sem.nim
```diff
@@ -839,6 +761,33 @@ proc semStmtAndGenerateGenerics(c: PContext, n: PNode): PNode =
   ## program compilation.
   addInNimDebugUtils(c.config, "semStmtAndGenerateGenerics", n, result)
 
+  proc isImportSystemStmt(g: ModuleGraph; n: PNode): bool =
```
```suggestion
  proc isImportSystemStmt(g: ModuleGraph; n: PNode): bool {.nimcall.} =
```

compiler/sem/sem.nim
```diff
@@ -839,6 +761,33 @@ proc semStmtAndGenerateGenerics(c: PContext, n: PNode): PNode =
...
+          if f == g.systemModule.info.fileIndex:
+            return true
+    of nkImportExceptStmt, nkFromStmt:
+      if n[0].kind == nkIdent:
+        let f = checkModuleName(g.config, n[0], false)
+        if f == g.systemModule.info.fileIndex:
+          return true
+    else: discard
+
+  proc isEmptyTree(n: PNode): bool =
```
```suggestion
  proc isEmptyTree(n: PNode): bool {.nimcall.} =
```

zerbina (approved):
Looks good to me. I left a few soft suggestions about preventing the now-inner procedures from accidentally being turned into closures, but they are non-blocking.

compiler/sem/sem.nim
saem: Good point, I'm going to drop `hasCycle` in a subsequent PR, want to get more clean-up in other places first and possibly help sort out macro sanitization.

zerbina:
Hm, `homebrew` failed to install dependency multiple times now in the past few days. I've restarted the job.

saem:
> Hm, `homebrew` failed to install dependency multiple times now in the past few days. I've restarted the job.

🤞🏽 that it works, thanks

@alaviss can you take a peek today if you have time?

PR #779, opened 2023-06-28
remove the `.dynlib` "extension loader" feature

zerbina:
## Summary

When using a call like `getProcAddr(x)` as the expression
with the `.dynlib` pragma on a procedure, where `x` is a 1-sized
string literal with the single item being a numeric character,
executing the call and assigning its result to the correct global was
emitted into a procedure named `nimLoadProcsX` (where X is the number
from the string literal) that could then be manually called.

The feature was introduced to improve extension loading for the Nim
OpenGL wrapper, but it was never documented, and its existence now
complicates moving more code generator logic into the unified backend
processing pipeline.

## Details

This reverts commit https://github.com/nim-works/nimskull/commit/cf06131decb2d46304874bd243c29267876e0076.

### DIFF
--- a/compiler/backend/cgen.nim
+++ b/compiler/backend/cgen.nim
@@ -706,18 +706,9 @@ proc symInDynamicLib(m: BModule, sym: PSym) =
       initLocExpr(m.initProc, n[i], a)
       params.add(rdLoc(a))
       params.add(", ")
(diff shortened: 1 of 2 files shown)


saem (approved):
The feature seems cute at best, glad to see it go.

PR #780, opened 2023-06-29
testament: fix `--tryfailing` not working with `all`

zerbina:
## Summary

The `--tryfailing` option wasn't passed on to the `testament` processes
spawned when using `all`, meaning that the option had no effect there.

### DIFF
--- a/testament/testament.nim
+++ b/testament/testament.nim
@@ -1391,6 +1391,8 @@ proc main() =
         myself &= " " & quoteShell("--failing")
       if retryContainer.retry:
         myself &= " " & quoteShell("--retry")
+      if runKnownIssues:
+        myself &= " " & quoteShell("--tryfailing")
       if targetsStr.len > 0:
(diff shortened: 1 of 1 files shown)


saem (approved):
ah, yeah it's easy to miss that, no good reminders for it -- I definitely zoned on it. 😅

zerbina:
I think adding a procedure that takes an `Execution` object and returns the command-line argument string necessary to achieve the state (essentially the inverse of `parseOpts`) could help a bit. If using an enum to represent all possible options and switches, one might even be able to use an exhaustive case statement.

saem:
> ... If using an enum to represent all possible options and switches, one might even be able to use an exhaustive case statement.

The enum approach sounds the most promising, nothing that needs to be in this PR, of course. I can make an issue, would be a "good first issue"?

saem:
re-running failed jobs -- macos again.

PR #781, opened 2023-06-30
sem: `typed` AST normalizes out `include` stmts

saem:
## Summary

Previously `include` statements, outside of type sections, were re-added
to `typed` AST. This is no longer the case, they're instead wholy
replaced by the included AST, which prevents repeated processing and
is consistent with `include`'s behaviour in type sections.

## Details

`semstmts` evaluates both declarative and type sections includes. Only
declarative includes, handled in `evalInclude` reintroduced the
`nkIncludeStmt` node within its results. It's been modified to no longer
do that.

In addition, `evalInclude` used a one off procedure `incMod` as part of
its processing, this has been made an inner proc.

Also removed deadcode `semAllTypeSections` and the `rsemRecursiveImport`
legacy report.

Lastly, a test was added to ensure we don't regress.

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -1891,10 +1891,8 @@ proc checkCovariantParamsUsages(c: PContext; genericType: PType) =
       if result:
         c.config.localReport(genericType.sym.info, reportTyp(
           rsemExpectedInvariantParam, t))
-
     of tySequence:
       return traverseSubTypes(c, t[0])
-
     of tyGenericInvocation:
       let targetBody = t[0]
       for i in 1..<t.len:
@@ -1905,7 +1903,6 @@ proc checkCovariantParamsUsages(c: PContext; genericType: PType) =
             if tfCovariant notin formalFlags:
               c.config.localReport(genericType.sym.info, reportSym(
                 rsemCovariantUsedAsNonCovariant, param.sym))
-
             elif tfWeakCovariant in formalFlags:
               param.flags.incl tfWeakCovariant
             result = true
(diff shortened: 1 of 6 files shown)


compiler/sem/semstmts.nim
```diff
@@ -2138,14 +2128,13 @@ proc semAllTypeSections(c: PContext; n: PNode): PNode =
   proc gatherStmts(c: PContext; n: PNode; result: PNode) {.nimcall.} =
     case n.kind
     of nkIncludeStmt:
-      for i in 0 ..< n.len:
-        var f = checkModuleName(c.config, n[i])
+      for p in n.items:
```
zerbina: If you're modifying `semAllTypeSections`, it might make sense to instead remove it, as it's currently unused.

zerbina (approved):
Nice, apart from the consistency improvements, this should also make sanitizing macro output a bit easier.

If we want to go the route of "`typed` AST keeps the shape of the input", then I think the introduction of a `nkExpansion` would make sense. It would be used in places where something is expanded (`include`, macros, templates, etc.), with the first child node storing the typed source-centric AST (`include` statement, macro call, etc.), and the second node the typed expansion.

compiler/sem/semstmts.nim
saem: The haphazard formatting was really bothering me as I was scrolling past it. But I totally missed the fact that it wasn't used, wow.

zerbina: No worries, I just thought it's worth mentioning.

saem: Removed it and the associated legacy report, thanks.

(would have done it yesterday but had a headache 😞)

saem:
/merge

(the macos/brew issue is a github side issue and they're working on it)

PR #782, opened 2023-06-30
internal: add the `asm` and `emit` MIR operators

zerbina:
Summary
=======

Introduce the dedicated `asm` and `emit` operators to the MIR,
replacing the opaque handling via `mnkPNode`. This:
- fixes expressions used in `asm` and `emit` statements not being
  subject to the MIR transformation
- makes MIR-based dependency scanning possible, without
  requiring special traversal logic for `mnkPNode`

Details
=======

Both the `emit` directive and the `asm` statement can include
arbitrary expression, but these (including the entities referenced by
them) were previously hidden at the MIR level. Dependency scanning by
way of iterating over all MIR nodes in a fragment was thus not
guaranteed to work when they were present.

The new MIR nodes act as output operators without any behaviour of
their own, with the string literals and input expression being passed
via an argument block. Phrased differently, `emit` and `asm` act and
look like calls of effect-free procedures at the MIR level.

The arguments are all communicated as being passed-by-value with the
call not modifying them. While this is not correct (the emitted code
could potentially modify the operands), it's easier to implement it
this way. Given the low-level nature of `asm` and `emit`, this
behaviour is deemed okay for now.

To make generation of the `emit` statement work, `nkPragma` processing
in `mirgen` is changed slightly: instead of persisting the whole
`nkPragma` node that contains an interesting directive, only the
interesting directives are persisted (by first moving them into
standalone `nkPragma` nodes).

### DIFF
--- a/compiler/backend/ccgstmts.nim
+++ b/compiler/backend/ccgstmts.nim
@@ -774,7 +774,7 @@ proc genAsmOrEmitStmt(p: BProc, t: PNode, isAsmStmt=false): Rope =
           r = mangleName(p.module, sym)
           sym.loc.r = r       # but be consequent!
         res.add($r)
-    of nkTypeOfExpr:
+    of nkType:
       res.add($getTypeDesc(p.module, it.typ))
     else:
       discard getTypeDesc(p.module, skipTypes(it.typ, abstractPtrs))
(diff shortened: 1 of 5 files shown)


saem (approved):
Nice.

Also normalizing pragma to singletons as part of `sem` or `transf` is a good call. I'll take a peek after wrapping up some existing bits.

PR #784, opened 2023-07-06
cgen: split parameter setup from rendering

zerbina:
## Summary

Separate `TLoc` setup for parameters from the rendering logic. This
reduces the amount of side-effects the rendering has, and is a
preparation for separating `TLoc` from `TSym`.

## Details

* move `TLoc` setup for parameters into `prepareParameters`
* inline the single use of `paramStorage`, and use `mapType` for the
  check
* in `genProcParams`, don't re-compute the information now provided by
  the parameters' `TLoc`
* add calls to `prepareParameters` where needed

This also fixes a regression where a procedure's parameter has no
filled-in `TLoc` if the procedure had no prototype emitted previously.

### DIFF
--- a/compiler/backend/ccgtypes.nim
+++ b/compiler/backend/ccgtypes.nim
@@ -370,6 +370,10 @@ $3endif$N
 
 proc prepareParameters(m: BModule, t: PType) =
   ## Prepares the locs of the parameter symbols for procedure type `t`.
+  ##
+  ## Neither the loc for the 'result' nor for the hidden environment parameter
+  ## are filled-in here. 'result' might not be a parameter, and the hidden
+  ## environment parameter is currently always treated as a local variable.
   assert t.kind == tyProc
   let params = t.n
   for i in 1..<params.len:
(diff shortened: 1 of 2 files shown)


compiler/backend/ccgtypes.nim
```diff
@@ -368,12 +368,36 @@ struct $2_Content { NI cap; $1 data[SEQ_DECL_SIZE];};
...
+            storage)
+
+    if ccgIntroducedPtr(m.config, param, t[0]):
+      # the parameter is passed by address; mark it as indirect
+      incl(param.loc.flags, lfIndirect)
+      param.loc.storage = OnUnknown
+
+  # neither the loc for the result nor for the hidden environment parameter
+  # are filled-in here. The result might not be a parameter, and the hidden
+  # environment parameter is currently always treated as a local variable
```
saem: Might be worth adding this to the proc's doc comment?

saem (approved):
One soft suggestion, but very much looking forward to dropping `TLoc` from `TSym`. :D

PR #785, opened 2023-07-07
cgen: fix regression with `.nodecl` consts

zerbina:
## Summary

Fix a regression where the C code generator ignored the `.nodecl`
pragma for constants.

## Details

Apart from the fix, the single-use `genConstSetup` is inlined into
`useConst`. Since the loc for constants is setup in `genConstDefinition`
(which is always called before `useConst`), the `fillLoc` call is not
added to `useConst` (this is also a step towards `fillLoc` only being
used in procedures responsible for definitions).

In addition, a test for the current behaviour of imported `.nodecl`


saem (approved):
Does fix the issue and it's good to have a regression test for it, as well.

Outside of the PR the behaviour difference between compile and runtime is a bit weird. Although I can't yet articulate what's off/missing.

PR #787, opened 2023-07-07
cgen: stricter symbol handling for `emit`

zerbina:
## Summary

Slightly refactor `ccgstmts.genAsmOrEmitStmt` and make its symbol
handling logic more strict. Apart from fields and types, all symbols
are now required to have their loc name already generated.

This is only an internal change. It prevents regressions, makes the
logic more clear, and is also a preparation for removing `TLoc` from
`TSym`.

## Details

An exhaustive case statement is used instead of the 'if', and the
redundant `$` calls are removed.

Due to the lack of context, field names cannot always be properly
mangled when the field is missing a loc name, but instead of the generic
name mangling (original name + attached-to module name + item id) the
record field mangling scheme is now used (which is incorrect for tuple
fields, but `mangleName` was too).

### DIFF
--- a/compiler/backend/ccgstmts.nim
+++ b/compiler/backend/ccgstmts.nim
@@ -729,28 +729,41 @@ proc genAsmOrEmitStmt(p: BProc, t: PNode, isAsmStmt=false): Rope =
       res.add(it.strVal)
     of nkSym:
       var sym = it.sym
-      if sym.kind in {skProc, skFunc, skIterator, skMethod}:
+      case sym.kind
+      of skProc, skFunc, skIterator, skMethod:
         var a: TLoc
         initLocExpr(p, it, a)
-        res.add($rdLoc(a))
-      elif sym.kind == skType:
-        res.add($getTypeDesc(p.module, sym.typ))
-      else:
+        res.add(rdLoc(a))
+      of skVar, skLet, skForVar, skParam, skResult, skTemp:
+        # make sure the C type description is available:
+        discard getTypeDesc(p.module, skipTypes(sym.typ, abstractPtrs))
+        var a: TLoc
+        initLocExpr(p, it, a)
+        res.add(rdLoc(a))
+      of skType:
+        res.add(getTypeDesc(p.module, sym.typ))
+      of skField:
+        # special support for raw field symbols
         discard getTypeDesc(p.module, skipTypes(sym.typ, abstractPtrs))
         var r = sym.loc.r
         if r == "":
-          # if no name has already been given,
-          # it doesn't matter much:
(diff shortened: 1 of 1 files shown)


compiler/backend/ccgstmts.nim
```diff
@@ -729,28 +729,41 @@ proc genAsmOrEmitStmt(p: BProc, t: PNode, isAsmStmt=false): Rope =
...
         discard getTypeDesc(p.module, skipTypes(sym.typ, abstractPtrs))
         var r = sym.loc.r
         if r == "":
-          # if no name has already been given,
-          # it doesn't matter much:
-          r = mangleName(p.module, sym)
+          # problem: a field doesn't know what type it is part of, so we
+          # can neither use ``fillObjectFields`` nor a ``FieldX`` name
+          # (in the case of tuples). To make sure that the mangled name is
+          # at  least correct for object types, we always use field name
```
saem, line 753:
```suggestion
          # at least correct for object types, we always use field name
```

very soft suggestion, might not be worth retriggering CI and instead rolling this in another PR.

compiler/backend/ccgstmts.nim
```diff
@@ -729,28 +729,41 @@ proc genAsmOrEmitStmt(p: BProc, t: PNode, isAsmStmt=false): Rope =
...
+        res.add(getTypeDesc(p.module, sym.typ))
+      of skField:
+        # special support for raw field symbols
         discard getTypeDesc(p.module, skipTypes(sym.typ, abstractPtrs))
         var r = sym.loc.r
         if r == "":
-          # if no name has already been given,
-          # it doesn't matter much:
-          r = mangleName(p.module, sym)
+          # problem: a field doesn't know what type it is part of, so we
```
saem, line 750: Would `sym.owner.typ` not provide the owning type or perhaps raw field symbols are different?

saem (approved):
One open quesion with respect to fields knowing their container type, but if that doesn't work out this is good to go.

compiler/backend/ccgstmts.nim
zerbina: Unfortunately, no, the `owner` of field symbols cannot be used here. This is because:
- for instances of generic objects, the field owner points to the symbol of the generic object, and from there it's not possible to get the instantiated type
- in some circumstances, the owner for object fields doesn't point to an `skType` symbol
- fields of named tuples cannot point to the tuple type's symbol (because there is none)

saem: sem lets us down yet again. :D

Thanks for providing scenarios where this doesn't work. OK, let's merge this and I'll see if I can maybe plug a few of these holes.

PR #789, opened 2023-07-07
internal: clean-up `modulegraphs` module

saem:
## Summary

Remove unused `ModuleGraph` fields `strongSemCheck` and
`compatibleProps`. Make various one off top-level procs as inner procs.

## Details

Proc changes:
- `resolveBackendSym` was only used in `getAttachedOp`, made inner
- `resolveInst` was only used in `procInstCacheItems`, made inner
- `resolveType` was only used in `typeInstCacheItems`, made inner
- `transitiveClosure` was only used in `markClientsDirty` made inner

### DIFF
--- a/compiler/modules/modulegraphs.nim
+++ b/compiler/modules/modulegraphs.nim
@@ -132,8 +132,6 @@ type
     cacheCounters*: Table[string, BiggestInt] # IC: implemented
     cacheTables*: Table[string, BTree[string, PNode]] # IC: implemented
     passes*: seq[TPass]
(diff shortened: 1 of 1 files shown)


zerbina (approved):
Looks good to me.

Maybe consider using the explicit `.nimcall` calling convention for the inner procedures, but I don't think that's anything blocking.

saem:
> Looks good to me.

thanks for the review.

> Maybe consider using the explicit `.nimcall` calling convention for the inner procedures, but I don't think that's anything blocking.

ugh, I forgot! I've gotten a bit gun shy wrt to CI passing on OSX... so I'll poke at it in a follow-up PR.

PR #790, opened 2023-07-08
internal: separate `TLoc` from `TSym` and `TType`

zerbina:
## Summary

Remove the `loc` fields from both `TSym` and `TType` and leave it to
the code generators how and if they use `TLoc` -- `TSym` only stores the
external name plus interface flags now. The benefits:
- another point of coupling (through data-types) between semantic
  analysis and code generation is removed
- the size of an `TType` instance shrinks from 120 to 96 bytes (with a
  64-bit target)
- the size of an `TSym` instance shrinks from 168 to 160 bytes

Because of its `lode` field, `TLoc` not being defined in the `ast_types`
module is also an important prerequisite for introducing a code IR for
the code generators.

## Details

The key changes:
- move `TLoc` and the associated types only relevant to the C code
  generator to `cgendata`
- remove the `loc` field from `TSym` and `TType` -- `TSym` stores the
  external name and interface flags via the new `extname` and `locFlags`
  fields
- as an interim solution, add the `locId` field to `TSym`
- fix the name slot in the `ast` of dispatcher storing the wrong symbol
- NDI file generation is now the responsibility of the orchestrator

### `cgen`

For associating a `PSym` with a `TLoc`, multiple `Store`s shared across
all C modules are used. The `TSym.locId` is an index into the
`SymbolMap` (which is a `Store` underneath) for the respective symbol
kind (globals, constants, procedures, fields). While a single
`SymbolMap` could be used (for the non-procedure entities), the plan is
to use dedicated types for the entities in the future, and using
separate stores is a preparation for that.

In order to not re-mangle parameters every time when emitting a
prototype, the locs for parameters are stored with their procedure.

Since locals are known to not outlive their surrounding procedure, the
locs for them are stored in a dedicated `SymbolMap` with the procedure
context -- once a procedure context goes away, the locals' `TLoc`s can
be, and are, freed already.

In general, the C code generator is now more strict with who is
responsible for filling-in loc information: for constants and globals,
it must only happen once during definition. Procedures and fields still
need to support ad-hoc loc setup, and fields need special-casing because
of `.inline` procedures and how `finally` clauses work (i.e., they're
duplicated at a too late time).

A table-based approach where the symbol IDs are mapped to `TLoc`s would
have also worked, but measurements showed that it was ~1.4% slower.

#### NDI files

Instead of directly updating the NDI file with a new entry when
generating a mangled name, as was done previously, adding the entries is
now the responsibility of the orchestrator: entries for whole-program
entities are added after code generation has finished -- entries for
locals once code generation for a procedure has finished.

For NDI file generation to work, the symbol. This was already the case
for all entities except constants, so constants now also store a `nkSym`
node in their `lode` field.

### `jsgen`

Due to it being less complex than the C code generator, a table-based
approach is okay for the JavaScript code generator. Only the JavaScript
name is relevant to the code generator, so it maps the symbol IDs
directly to strings and doesn't use `TLoc`. Separate tables are used
for global entities (globals, constants, procedures, and fields),
locals, and parameters.

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -309,14 +309,6 @@ proc newType*(kind: TTypeKind, id: ItemId; owner: PSym): PType =
       echo "KNID ", kind
       writeStackTrace()
 
-
-proc mergeLoc(a: var TLoc, b: TLoc) =
-  if a.k == low(typeof(a.k)): a.k = b.k
-  if a.storage == low(typeof(a.storage)): a.storage = b.storage
-  a.flags.incl b.flags
-  if a.lode == nil: a.lode = b.lode
-  if a.r.len == 0: a.r = b.r
-
 proc newSons*(father: Indexable, length: int) =
   setLen(father.sons, length)
 
@@ -333,7 +325,9 @@ proc assignType*(dest, src: PType) =
     if dest.sym != nil:
       dest.sym.flags.incl src.sym.flags-{sfExported}
       if dest.sym.annex == nil: dest.sym.annex = src.sym.annex
-      mergeLoc(dest.sym.loc, src.sym.loc)
+      dest.sym.locFlags.incl src.sym.locFlags
+      if dest.sym.extname.len == 0:
+        dest.sym.extname = src.sym.extname
     else:
       dest.sym = src.sym
   newSons(dest, src.len)
@@ -355,7 +349,8 @@ proc copySym*(s: PSym; id: ItemId): PSym =
   result.magic = s.magic
   result.options = s.options
   result.position = s.position
-  result.loc = s.loc
+  result.extname = s.extname
+  result.locFlags = s.locFlags
   result.annex = s.annex      # BUGFIX
   result.constraint = s.constraint
   if result.kind in {skVar, skLet, skField}:
@@ -372,7 +367,8 @@ proc createModuleAlias*(s: PSym, id: ItemId, newIdent: PIdent, info: TLineInfo;
   result.flags = s.flags
   result.options = s.options
   result.position = s.position
-  result.loc = s.loc
+  result.extname = s.extname
+  result.locFlags = s.locFlags
   result.annex = s.annex
 
 proc initStrTable*(x: var TStrTable) =
@@ -510,7 +506,7 @@ template transitionSymKindCommon*(k: TSymKind) =
   s[] = TSym(kind: k, itemId: obj.itemId, magic: obj.magic, typ: obj.typ, name: obj.name,
              info: obj.info, owner: obj.owner, flags: obj.flags, ast: obj.ast,
              options: obj.options, position: obj.position, offset: obj.offset,
-             loc: obj.loc, locId: obj.locId,
+             extname: obj.extname, locFlags: obj.locFlags, locId: obj.locId,
              annex: obj.annex, constraint: obj.constraint)
   when defined(nimsuggest):
     s.allUsages = obj.allUsages
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -510,7 +510,8 @@ template transitionSymKindCommon*(k: TSymKind) =
   s[] = TSym(kind: k, itemId: obj.itemId, magic: obj.magic, typ: obj.typ, name: obj.name,
              info: obj.info, owner: obj.owner, flags: obj.flags, ast: obj.ast,
              options: obj.options, position: obj.position, offset: obj.offset,
-             loc: obj.loc, annex: obj.annex, constraint: obj.constraint)
+             loc: obj.loc, locId: obj.locId,
+             annex: obj.annex, constraint: obj.constraint)
   when defined(nimsuggest):
     s.allUsages = obj.allUsages
 
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1590,19 +1590,13 @@ type
     data*: seq[PSym]
 
   # -------------- backend information -------------------------------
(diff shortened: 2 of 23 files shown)


compiler/ast/ast.nim
```diff
@@ -355,7 +349,8 @@ proc copySym*(s: PSym; id: ItemId): PSym =
   result.magic = s.magic
   result.options = s.options
   result.position = s.position
-  result.loc = s.loc
+  result.extname = s.extname
+  result.locFlags = s.locFlags
```
saem, line 353: too big a change to name it `extFlags`?

a soft suggestion as a fair bit of this is under heavy transition.

compiler/ast/ast_types.nim
```diff
@@ -1590,19 +1590,13 @@ type
...
-    locGlobalVar,             ## location is a global variable
-    locParam,                 ## location is a parameter
-    locField,                 ## location is a record field
-    locExpr,                  ## "location" is really an expression
-    locProc,                  ## location is a proc (an address of a procedure)
-    locData,                  ## location is a constant
-    locCall,                  ## location is a call expression
-    locOther                  ## location is something other
   TLocFlag* = enum
+    # XXX: `TLocFlag` conflates two things:
```
saem, line 1594: ah, even more reason to hold off on `extFlags` until these have been separated.

compiler/ast/ast_types.nim
```diff
@@ -1713,7 +1696,12 @@ type
                               ## to the module's fileIdx
                               ## for variables a slot index for the evaluator
     offset*: int              ## offset of record field
-    loc*: TLoc
+    extname*: string          ## the external name of the type, or empty if a
+                              ## generated name is to be used
+    locFlags*: TLocFlags      ## additional flags that are relevant to code
+                              ## generation
+    locId*: uint32            ## associates the symbol with a loc in the C code
```
saem, line 1703: If two symbols can't be associated with the same location, could we piggy back off of `itemId`?

(I'm guessing this will result in a 'no' once I've read further in the PR)

It would mean we'd have to store _something_ for each symbol (more memory), but that might balance out as we're storing an extra 4 bytes with the current change. The lack of `loc` information would have to be tracked on the loc data end, which does mean a deref/index-lookup prior to knowing it doesn't exist.

compiler/backend/cbackend.nim
```diff
@@ -93,6 +94,14 @@ type
 
 func hash(x: PSym): int = hash(x.id)
 
+proc writeMangledLocals(p: BProc) =
+  ## Writes the mangled names of `p`'s locals to the module's NDI file.
```
saem, line 98: this is a nice secondary benefit developing over time, as the mangling becomes more and more consistent it's going to make `gdb` pretty printers' life all that much easier.

compiler/backend/ccgstmts.nim
```diff
@@ -169,7 +169,13 @@ proc genSingleVar(p: BProc, v: PSym; vn, value: PNode) =
 
   if value.kind != nkEmpty:
     genLineDir(p, vn)
-    loadInto(p, vn, value, v.loc)
+    # we have an parameter aliasing issue here: passing `p.locals[v]`
```
saem:
```suggestion
    # we have a parameter aliasing issue here: passing `p.locals[v]`
```

compiler/backend/ccgtypes.nim
```diff
@@ -501,35 +490,46 @@ proc genRecordFieldsAux(m: BModule, n: PNode,
     if field.typ.kind == tyVoid: return
     #assert(field.ast == nil)
     let sname = mangleRecFieldName(m, field)
-    fillLoc(field.loc, locField, n, unionPrefix & sname, OnUnknown)
+    if field.locId == 0:
+      # XXX: the C struct definition for the type is re-generated in every C
+      #      file the type is used in, so the field might have an associated
+      #      loc. Eventually, each C type is only generated once, and then the
+      #      guard can be removed
```
saem:
```suggestion
      #      loc. Eventually, each C type will only be generated once, and then
      #      the guard can be removed
```

compiler/backend/jsgen.nim
```diff
@@ -267,7 +275,40 @@ proc mangleName(m: BModule, s: PSym): Rope =
       result.add("_")
       result.add(rope(s.id))
 
-    s.loc.r = result
+proc mangledName(p: PProc, s: PSym, info: TLineInfo): string =
+  ## Returns the cached JavaScript name for `s`.
+  ## At the time of writing, a borrowed string (``lent string``) from
+  ## can't be returned because of compiler shortcomings.
```
```suggestion
  ## At the time of writing, a borrowed string (``lent string``)
  ## can't be returned because of compiler shortcomings.
```

compiler/ast/ast_types.nim
saem: revisiting this comment, I'm pretty sure this wouldn't be a good approach at present, better to defer it as the necessary data and consequent layout are likely to change dramatically.

saem (approved):
Looking forward to having this merged in, further breaking dependencies like this is great.

Few minor items, they're mostly typographical and non-blocking.

compiler/ast/ast_types.nim
zerbina: Yep, I'm going to split the enum up immediately after this PR.

compiler/backend/ccgstmts.nim
zerbina: Ah, thanks -- I thought I had caught all of those. The sentence originally was "... have an issue with ...", but I reworded it and missed adjusting the 'an' :P

PR #791, opened 2023-07-09
mirgen: improve the MIR generation DSL

zerbina:
## Summary

Make the atom routines take an `MirNodeSeq` as input (instead of the
much more broad `TCtx`) and change the context-provider templates to
accept a `TCtx` instance as the first argument.

```nim
chain: input(c) => op(c) => output(c)
```

is now written as:

```nim
chain(c): input() => op() => output()
```

This makes the chains easier to write, and is a preparation for moving
the definitions that make up the DSL out of `mirgen` and into the
`mirconstr` module.

## Details

* make the destination buffer available to the adapter templates by
  injecting a nullary template (`buffer`) into the scope, removing the
  need to manually pass the buffer to each atom routine
* generate adapters for the procedures that represent input operands
* for compatibility, also generate input adapters that accept a `TCtx`.
  This reduces the amount of changes necessary
* adjust the usages of `chain`, `eval`, and `forward`
* don't use dirty templates for the adapter-generators -- normal
  templates suffice


saem (approved):
Will have to see how it plays out with further usage, but seems like a net win as it avoids getting the context mixed up along the transformation chain.

PR #792, opened 2023-07-10
internal: split up `TLocFlag`

zerbina:
## Summary

Separate the enum into two enums: one for the flags describing a
symbol's external interface and one for the actual location flags. This
fixes an implementation detail of the C code generator (location flags)
leaking into `TSym`.

## Details

The enum for the external interface flags is named `ExternalFlag` and
the location flag enum `LocFlag`. Since the latter is only relevant to
the C code generator, it is moved to `cgendata`.

The field storing the external interface flags in `TSym` and `PackedSym`
is renamed to `extFlags`. Due to the `ef` prefix being already taken by
the `TExprFlag` enum, the `ExternalFlag` enum uses the `exf` prefix.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1589,30 +1589,18 @@ type
     counter*: int
     data*: seq[PSym]
 
-  # -------------- backend information -------------------------------
-  TLocFlag* = enum
-    # XXX: `TLocFlag` conflates two things:
-    #      - flags regarding the external interface (e.g., `lfHeader`; set by
-    #        sem, used by sem and the code generators)
-    #      - location-related flags (e.g., ``lfIndirect``, ``lfSingleUse``,
-    #        etc.; only relevant to the C code generator)
-    #      Split the enum up.
-    lfIndirect,               ## backend introduced a pointer
-    lfFullExternalName, ## only used when 'conf.cmd == cmdNimfix': Indicates
+  ExternalFlag* = enum
+    ## Flags that describe a symbol's external interface.
+    exfFullExternalName  ## only used when 'conf.cmd == cmdNimfix': Indicates
       ## that the symbol has been imported via 'importc: "fullname"' and
       ## no format string.
-    lfNoDecl,                 ## do not declare it in C
-    lfDynamicLib,             ## link symbol to dynamic library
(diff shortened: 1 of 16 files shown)


compiler/ast/ast_types.nim
```diff
@@ -1589,30 +1589,18 @@ type
...
-  TLocFlag* = enum
-    # XXX: `TLocFlag` conflates two things:
-    #      - flags regarding the external interface (e.g., `lfHeader`; set by
-    #        sem, used by sem and the code generators)
-    #      - location-related flags (e.g., ``lfIndirect``, ``lfSingleUse``,
-    #        etc.; only relevant to the C code generator)
-    #      Split the enum up.
-    lfIndirect,               ## backend introduced a pointer
-    lfFullExternalName, ## only used when 'conf.cmd == cmdNimfix': Indicates
+  ExternalFlag* = enum
```
saem, line 1592: Could also call it `ExternFlag`, leaning a bit on `extern` being evocative of effecting the ABI or external binary interface.

---

Thought of this after, I think this is to do with how the symbol manifests in the intended compiled artifact(s)'. Sticking with that some alternatives are:

- `ArtifactFlag`
- `BinaryInterfaceFlag`

Personally, I like the existing `ExternalFlag` more than the above suggestions, merely noting them in case they inspire a better name and to track what we considered.

Finally, perhaps this flag is a misplaced symbol flag?

saem (approved):
Tried to come up with a name suggestion but I think the current one is still the best, except perhaps it should simply be a symbol flag.

Otherwise the code seems more sensible than before.

compiler/ast/ast_types.nim
zerbina: Hmm, about it being a misplaced symbol flag, I'm not sure.

While a normal symbol flag is relevant to both semantic analysis and the backend (the flags modify what the symbol represents), `ExternalFlag`s are, for the largest part, only relevant to the code generators (they only affect code generation [^1]), so I *think* that having them separate makes sense, at least for now.

[^1]: not entirely; `exfNoDecl` and `exfDynamicLib` do have an effect on observable behaviour

saem:
/merge

feel free to stop it in case you wanted to change things further.

PR #793, opened 2023-07-10
internal: move MIR generation DSL to `mirconstr`

zerbina:
## Summary

Move the routines and types that make up the DSL to the `mirconstr`
module and use the DSL in both the `injectdestructors` and `mirbridge`
module.

Apart from making the MIR code generation in both modules less complex,
manual, and error-prone, this is also significant progress towards the
MIR code coming out of the `injectdestructors` pass having proper type
information, which is a requirement for having MIR passes that run after
`injectdestructors`.

## Details

* move the `EValue` and `ChainEnd` types to `mirconstr`
* move the atomic DSL operands to `mirconstr`. The `notOp`, `modify`,
  `outOp`, and `tupleAccess` operands stay in `mirgen`
* add wrapper templates for DSL operands that `mirgen` still calls with
  a `TCtx` instance instead of a `MirNodeSeq`

As part of moving them, some operand routines that had a 'gen' prefix
are changed to not have it, making the routine names more consistent.

In addition, some cleanup in `mirconstr` is performed, and the DSL
extended a bit:
* `|=>`, `=>|`, and `previous` are removed. They were unused and didn't
  fit in with the overall design
* the `emit`, `symbol`, and `opParam` operands are added, all of which
  are only useful outside of `mirgen`

Another addition is the `predicate` meta operand, which makes it
possible to write chains where an operand is conditionally excluded:

```nim
chain(...): a() => predicate(cond) => b() => c()
```

Here, `b()` is only evaluated if `cond` evaluates to 'true'. Using two
chains plus an in-between 'if' statement was previously necessary to
achieve the same.

Finally, documentation about the DSL is added, and the
`injectdestructors` and `mirbridge` modules are changed to use the DSL
for generating MIR code. The logic from the `genInjectedSink` procedure
was previously duplicated into `genSinkFromTemporary`, but this is now
fixed.

### DIFF
--- a/compiler/mir/mirconstr.nim
+++ b/compiler/mir/mirconstr.nim
@@ -119,6 +119,12 @@ template follows*(): Sink =
 
 {.push inline.}
 
+func emit*(s: var MirNodeSeq; n: sink MirNode): EValue =
+  ## Adds `n` to the node sequences, with its type providing
+  ## the initial type information for the result.
+  result = EValue(typ: n.typ)
+  s.add n
+
 func procLit*(s: var MirNodeSeq, sym: PSym): EValue =
   s.add MirNode(kind: mnkProc, typ: sym.typ, sym: sym)
   result = EValue(typ: sym.typ)
@@ -137,6 +143,16 @@ func constr*(s: var MirNodeSeq, typ: PType): EValue =
 
 func tempNode*(s: var MirNodeSeq, typ: PType, id: TempId): EValue =
   s.add MirNode(kind: mnkTemp, typ: typ, temp: id)
+  result = EValue(typ: typ)
+
+func symbol*(s: var MirNodeSeq, kind: range[mnkConst..mnkLocal],
+             sym: PSym): EValue =
+  s.add MirNode(kind: kind, typ: sym.typ, sym: sym)
+  result = EValue(typ: sym.typ)
+
+func opParam*(s: var MirNodeSeq, i: uint32, typ: PType): EValue =
+  assert typ != nil
+  s.add MirNode(kind: mnkOpParam, param: i, typ: typ)
   result = EValue(typ: typ)
 
 func magicCall*(s: var MirNodeSeq, m: TMagic, typ: PType): EValue =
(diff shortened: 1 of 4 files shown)


saem:
Updated the PR description, under the details section, the second bullet point said that `notOp` etc stayed in `mirconstr`, which isn't the case, updated to `mirgen`.

zerbina:
Yep, the bullet point was wrong, thank you.

compiler/mir/mirconstr.nim
```diff
@@ -45,6 +143,188 @@ template follows*(): Sink =
...
+func arg*(s: var MirNodeSeq, val: EValue) =
+  s.add MirNode(kind: mnkArg, typ: val.typ)
+
+func consume*(s: var MirNodeSeq, val: EValue) =
+  s.add MirNode(kind: mnkConsume, typ: val.typ)
+
+func name*(s: var MirNodeSeq, val: EValue) =
+  s.add MirNode(kind: mnkName, typ: val.typ)
+
+func voidOut*(s: var MirNodeSeq, val: EValue) =
```
saem, line 257:
```suggestion
func voidOut*(s: var MirNodeSeq, val: EValue) =
  ## end the mir input expression as a void-expression
```

not sure I have that quite right, this is from what I could gather based on usage. initially I was seeing if I could come up with a better name, settled on a doc string might help. (soft suggestion)

saem (approved):
DSL definitely makes it more compact, while keeping the tree structure relatively straightforward to follow.

compiler/mir/mirconstr.nim
zerbina: Yup, that sounds correct. I added it plus some additional elaboration.

In general, the 'void' operator/node acts as both a `discard` and as a delimiter for expressions ('call' operators that yield nothing also need to be followed by 'void').

zerbina:
Apologies for the long time it took me to respond, I didn't manage to get any work done yesterday.

saem:
> Apologies for the long time it took me to respond, I didn't manage to get any work done yesterday.

no worries. this is a fun project and I don't have any expectations; some days it's not meant to be.

saem:
hand slipped on the trackpad and I clicked the wrong button (close with comment). 🤦🏽

PR #797, opened 2023-07-15
internal: data-oriented storage for `TLib`

zerbina:
## Summary

Replace storing `TLib` objects as `ref`s in `TSym` with storing all
`TLib` instances belonging to a module in a sequence that a `TSym` then
stores an index-like reference into.

This makes serializing the symbols and associated `TLib` instances
easier: instead of storing a `PackedLib` for *each* packed symbol, the
packed symbols now only store a module+index tuple.

## Details

For simplicity, the list with the libs for a module is stored in a
separate `seq` in `ModuleGraph`. The list cannot be stored in
`PContext`, as it needs to be accessible after semantic analysis, at
which point `PContext` is already gone.

While the handle to the `TLib` instance stored with `TSym` would
ideally be a single index into the `TLib` list of the module a
symbol is attached to, the way symbols of aliased structural types
currently work (the aliased type's `TSym` content is partially
copied) makes this not possible, meaning that a symbol also has to
store the ID of the module the referenced `TLib` instance is part
of.

Storing the `TLib` objects for a module happens when the module is
closed. While not strictly necessary at the moment, this is a
preparation for storing a `PSym` in `TLib.name` that is only setup
when the module is closed.

Aside from reducing the size of rodfiles, not storing a `PackedLib`
for each `PackedSym` also fixes another inefficiency, namely that
each deserialized `TSym` got a unique, possibly duplicated, `TLib`
instance.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1,7 +1,6 @@
 import compiler/ast/lineinfos
 import compiler/utils/ropes
 import std/[hashes]
-import experimental/dod_helpers
 
 from compiler/ast/idents import PIdent, TIdent
 
@@ -45,15 +44,10 @@ type
 
 type
   NodeId* = distinct int32
-  LibId* = distinct uint32
 
 proc `==`*(a, b: NodeId): bool {.borrow.}
 proc hash*(a: NodeId): Hash {.borrow.}
 proc `$`*(a: NodeId): string {.borrow.}
-
-template indexLike*(_: typedesc[LibId]) =
-  # makes ``LibId`` available to use with ``opt``
-  discard
 
 type
   TNodeKind* = enum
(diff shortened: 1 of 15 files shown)


compiler/ast/ast_types.nim
```diff
@@ -1615,6 +1615,20 @@ type
     name*: Rope
     path*: PNode              ## can be a string literal!
 
+  LibId* = object
+    ## Identifies a ``TLib`` instance. The default value means 'none'.
```
saem, line 1619: Might be worth adding a `unknownLibId` or the like as a constant below to further 'reserve' the concept of the default value `LibId(module: 0, index: 0)` as none.

---

Update: reading more of the PR, there is a similar tuple type for IC, so we wouldn't end up using that const/this object type all that much. Seems less worth it practically speaking, mostly there to signal intent.

compiler/modules/modulegraphs.nim
```diff
@@ -377,6 +380,24 @@ proc getToStringProc*(g: ModuleGraph; t: PType): PSym =
...
+  result = g.libs[id.module][id.index - 1]
+
+proc addLib*(g: ModuleGraph, module: int, lib: sink TLib): LibId =
+  ## Registers (adds) `lib` with the given module and returns the ID through
+  ## which the instance can be accessed from now on.
+  g.libs[module].add lib
+  # the index is 1-based, so we can directly use the length
+  result = LibId(module: module.int32, index: g.libs[module].len.uint32)
+
+proc storeLibs*(g: ModuleGraph, module: int) =
```
saem, line 394:
```suggestion
proc ownTheLibs*(g: ModuleGraph, module: int) =
```

jk jk 😆

saem (approved):
I like the overall approach, I can see it nudging us towards further convergence between existing compiler and IC.

PR #798, opened 2023-07-15
internal: `astrepr.debug` now useful for `PType`

saem:
## Summary

`astrepr.debug` overload for `PType` now provides more tailored output,
reducing spam and surfacing more useful information.

## Details

Previously, `astrepr.debug` used a generic `implicitTReprConf` when
generating output. This meant continually fiddling with that set and/or
having largely useless/spammy output for `PType` vs `PSym` vs `PNode`,
etc. Now the `PType` overload has its own config that can be tailored
for types.

### DIFF
--- a/compiler/utils/astrepr.nim
+++ b/compiler/utils/astrepr.nim
@@ -1248,7 +1248,7 @@ template debug*(it: PType) =
 
                   # trfShowNodeTypes,
                   trfShowNodeLineInfo, trfShowNodeFlags}
-  debugAux(implicitDebugConfRef, it, implicitTReprConf, instLoc())
+  debugAux(implicitDebugConfRef, it, conf, instLoc())
 
 template debug*(it: PIdent) =
   ## Print tree representation of a PIdent for compiler debugging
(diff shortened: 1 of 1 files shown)


compiler/utils/astrepr.nim
```diff
@@ -1234,7 +1234,20 @@ template debug*(it: PSym) =
...
+
+                  trfShowTypeFlags
+                } -
+                {
+                  trfShowTypeSym, trfShowSymLineInfo,
+                  trfShowSymFlags, trfShowSymMagic,
+
+                  # trfShowNodeTypes,
+                  trfShowNodeLineInfo, trfShowNodeFlags}
   debugAux(implicitDebugConfRef, it, implicitTReprConf, instLoc())
```
zerbina: Did you mean to pass the new `conf` to `debugAux` here?

saem: D'oh, that got messed up in the copy paste, thanks for catching that.

zerbina (approved):

PR #802, opened 2023-07-17
backend: fix inconsistencies in `--genScript:on` mode

CyberTailor:
## Summary
* Bugfix: Make compiler executable configurable in `--genScript:on`
  mode.

  That means you can now use command-line options like `--gcc.exe` or
  nim.cfg options like `clang.exe` with `--genScript:on` flag, and they
  will be respected.

  This is needed if compiler on the target machine (when
  cross-compiling) is named differently than the default.

* Change: Use `--cincludes` option in `--genScript:on` mode.

  Since `cIncludes` paths can be specified only via cli, there should be
  no unintended side effects. This change is made for consistency.

## Details
This PR changes behavior of the `getCompileCFileCmd` function in the
`extccomp` module, in particular:

* Make sure `getConfigVar(conf, c, ".exe")` always takes precedence over
  `getCompilerExe`, even when relative paths are requested.
* Remove outdated code related to compiler executable formatting. None of
  defined compilers used format strings in the executable.

### DIFF
--- a/compiler/backend/extccomp.nim
+++ b/compiler/backend/extccomp.nim
@@ -626,7 +626,7 @@ proc getCompileCFileCmd*(conf: ConfigRef; cfile: Cfile,
           kind: rbackCannotProduceAssembly,
           usedCompiler: CC[conf.cCompiler].name)
 
-  result = exe & ' ' & CC[c].compileTmpl % [
+  result = exe.quoteShell & ' ' & CC[c].compileTmpl % [
     "dfile", dfile,
     "file", cfsh, "objfile", quoteShell(objfile),
     "options", options, "include", includeCmd,
--- a/compiler/backend/extccomp.nim
+++ b/compiler/backend/extccomp.nim
@@ -569,9 +569,16 @@ proc getCompileCFileCmd*(conf: ConfigRef; cfile: Cfile,
   var options = cFileSpecificOptions(conf, cfile.nimname, cfile.cname.changeFileExt("").string)
 
   var exe = getConfigVar(conf, c, ".exe")
-  if exe.len == 0: exe = getCompilerExe(conf, c, cfile.cname)
+  if exe.len == 0:
+    # fallback to the default
+    exe = getCompilerExe(conf, c, cfile.cname)
 
-  if needsExeExt(conf): exe = addFileExt(exe, "exe")
+  if needsExeExt(conf):
+    exe = exe.addFileExt("exe")
+
+  if not noAbsolutePaths(conf):
+    exe = joinPath(conf.cCompilerPath, exe)
+
   if optGenDynLib in conf.globalOptions and
       ospNeedsPIC in platform.OS[conf.target.targetOS].props:
     options.add(' ' & CC[c].pic)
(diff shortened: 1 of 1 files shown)


saem:
Thanks for the PR.

saem (changes requested):
Please update the PR message, it's used as the final commit message the pre-existing PR template walks you through it.

I took a quick look at the code and it seems it `compilePattern`, and possibly `includeCmd`, could be made `let`. The mutable variable and awkward/error prone logic is facilitated by the mutable variable.

CyberTailor:
> Please update the PR message

Done, although the description is kinda short.

> it seems it compilePattern, and possibly includeCmd, could be made let

I don't see how

zerbina:
> Done, although the description is kinda short.

A small request regarding the summary: could you reword it so that it reflects the bugfix nature of the PR? The current wording makes it sound more like a feature addition.

It's not directly mentioned in the template, but since the PR message is used as the final commit message, please also make sure that the lines have a maximum length of 72 characters.

> I don't see how

At least for `compilePattern`, the following would work:
```nim
let compilePattern =
  if noAbsolutePaths(conf): joinPath(conf.cCompilerPath, exe)
  else:                     exe
```

compiler/backend/extccomp.nim
```diff
@@ -580,17 +580,16 @@ proc getCompileCFileCmd*(conf: ConfigRef; cfile: Cfile,
     options.add ' '
     options.add cfile.customArgs
 
-  var compilePattern: string
+  let compilePattern =
+    if noAbsolutePaths(conf): exe
```
saem: Just confirming my understand: this is in fact the crux of the change because regardless of flags the pattern is configured in the same fashion?

CyberTailor: `getCompilerExe(conf, c, cfile.cname)` is a fallback, but it's currently forced in `--genScript:on` mode

saem:
thanks again for the PR, @CyberTailor.

The PR is used as a fairly detailed change log in order to:
- derive a change log
- help debug via git bisect
- learning the code base/code archeology

All of these require clear PR messages that describe:
- the behavioural change (before and after)
- implementation changes
- reasons for change
- any additional context/tidbits

I've pulled together a todo for things I've noticed:
- [ ] the PR title ("refactor") doesn't reflect this is a bug fix; your first line in the summary is closer to summarizing the effect of the PR
- [ ] update the `summary`:
  - [ ] describe the new behaviour, what can people expect
  - [ ] add in a reason for the change (perhaps something like: it's inconsistent between modes for no good reason/without cause)
- [ ] update the `details`:
  - [ ] with the title update, please ensure that the module `extccomp` and changed routine `getCompileCFileCmd` are mentioned
  - [ ] even though they're `details`, they're a summary relative to the code, most people won't know local variables, so instead of `exe` and `compilerPattern`, instead take the perspective of behaviour of the compiler before vs after and the changes required to make that happen outward behaviour was before (the executable wasn't configurable) and 

N.B. `refactor` means changing code without changing behaviour

Just in case it helps: my approach to writing PR message is I brain dump everything under `details` first, clean it up, then extract the `summary` from it, and a `title` from the `summary`. If you're stuck collect the bullet points and I can help. Finally, recent commits are PR summaries and you can use those as examples. Zerbina's are especially good.

saem (approved):
That's a much improved description, thanks

PR #803, opened 2023-07-17
front: skip non-existent paths in gcc depfiles

CyberTailor:
## Summary
There could be non-existent files in `conf.m.fileInfos` that prevented GCC-style depfiles from behaving correctly. This commit filters them out.

## Details
Also create a helper function to quote paths for Unix Makefiles.

### DIFF
--- a/compiler/front/depfiles.nim
+++ b/compiler/front/depfiles.nim
@@ -6,7 +6,7 @@
 ## Functions for writing target's dependencies in various formats.
 
 import
-  std/[strutils, tables]
+  std/[os, strutils, tables]
 
 import
   compiler/utils/[pathutils],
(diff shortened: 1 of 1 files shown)


compiler/front/depfiles.nim
```diff
@@ -25,13 +25,14 @@ proc writeDepsFile*(g: ModuleGraph) =
       f.writeLine(toFullPath(g.config, k))
   f.close()
 
+func makefileQuoted(path: string): string =
```
saem: Isn't there already a pre-existing procedure for this like quote shell?

```suggestion
func quoteFilepath(path: string): string =
```

Also, this seems like a better name

compiler/front/depfiles.nim
```diff
@@ -25,13 +25,14 @@ proc writeDepsFile*(g: ModuleGraph) =
...
   ## Outputs one `make` rule containing target's file name, a colon, and the
   ## names of all specified paths. Spaces and hashes are escaped.
-  let target = target.multiReplace((" ", "\\ "), ("#", "\\#"))
-  depfile.write(target & ": \\" & '\n')
-  for it in paths:
-    let path = it.multiReplace((" ", "\\ "), ("#", "\\#"))
-    if path.len == 0:
+  depfile.write(target.makefileQuoted & ": \\" & '\n')
+  for path in paths:
+    if path.len == 0 or not fileExists(path):
```
saem, line 37: I don't recall when this is called, but if this is meant to produce a reusable description then this shouldn't be filtered here.

saem (changes requested):
Thanks for opening a PR.

As per the other PR the template needs to be updated. Some of the key questions as part of the template would surface some deeper insights and help produce a deeper fix.

compiler/front/depfiles.nim
CyberTailor: > Isn't there already a pre-existing procedure for this like quote shell?

Yes, but only subset of that is needed

CyberTailor: > Also, this seems like a better name

Committed.

compiler/front/depfiles.nim
CyberTailor: Updated PR message

CyberTailor:
Closing until I remember what package caused this behavior

PR #805, opened 2023-07-18
cgen: replace `TNodeTable` usage with `Table`

zerbina:
## Summary

Replace the usage of `TNodeTable` with `Table` in `cgen`. This is a
preparation for moving away from `TNode` in the code generators.

In addition, the change allowed for fixing an issue with the
structural comparison logic that led to `-0.0` and `0.0` being treated
as equal and thus not-equal array construction expressions (e.g.
`[-0.0, 0.0]` and `[0.0, 0.0]`) collapsed into the the same constant
when using the C backend.

## Details

The implementation of `TNodeTable` is almost exactly the same as that of
`Table`, with the largest difference being that `TNodeTable` uses the
`key` and not its stored hash for testing whether a slot is empty. To
be able to use a `PNode` as the key, a distinct type (`ConstrTree`; for
"construction tree") is created that has a structural hash and
structural equality operator.

Both operators are based on the previously used `hashTree` and
`treesEquivalent` procedures from the `treetab` module, with adjustments
made for the specific use case:
- `nkIdent` nodes don't reach the code generator (and thus don't need to
  be considered)
- nodes of integer kind always have their value hashed, as opposed to
  the value being used as the hash (which means a hash is also computed
  for >= 2^32 values when using a 32-bit compiler)
- the bit pattern of floats are compared, not the values themselves

The last change fixes the aforementioned issue with two non-bit-pattern-
equal values resulting from constant array construction expressions
using into the same constant.

### DIFF
--- a/compiler/backend/cgendata.nim
+++ b/compiler/backend/cgendata.nim
@@ -11,6 +11,7 @@
 
 import
   std/[
+    hashes,
     intsets,
     tables,
     sets
@@ -18,7 +19,8 @@ import
   compiler/ast/[
     ast,
     lineinfos,
-    ndi
+    ndi,
+    types
   ],
   compiler/modules/[
     modulegraphs
@@ -28,6 +30,7 @@ import
   ],
   compiler/utils/[
     containers,
+    idioms,
     ropes,
     pathutils
   ]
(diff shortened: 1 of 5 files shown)


compiler/backend/cgendata.nim
```diff
@@ -332,4 +342,70 @@ func contains*[T](m: SymbolMap[T], sym: PSym): bool {.inline.} =
 
 iterator items*[T](m: SymbolMap[T]): lent T =
   for it in m.store.items:
-    yield it
\ No newline at end of file
+    yield it
+
+proc hash(n: ConstrTree): Hash =
+  ## Computes a hash over the structure of `tree`. The hash function is
```
saem:
```suggestion
  ## Computes a hash over the structure of a tree (`n`). The hash function is
```

or rename the paramter to `tree`?

saem (approved):
One soft suggestion to do with docs/naming. Otherwise this is good to go.

PR #806, opened 2023-07-18
contributing: update the commit guidelines

zerbina:
## Summary

Remove mentions of the outdated single commit requirement from both the
PR message template and `contributing.rst`. In addition, also add
details regarding what happens when merging the PR to both.

The intention with the latter is to make it easier for first-time
contributors to get an overview of the merge process, requirements, and
the role of the PR message.

## Details

With bors replaced with merge queues, the PR having a single commit is
not required anymore, and manually squashing the commits is in fact
oftentimes detrimental to understanding a merged PR. The section is
replaced with an encouragement to use multiple commits and a
description of what happens when the PR merged.

Rewriting history that has already been reviewed can make incremental
reviews harder (GitHub only shows the difference between the most
recent force push and the prior state), so an encouragement against
doing so is added to 'Anatomy of a Pull Request' section.

Since the line length requirement was only mentioned a single time in
the rather large `contributing.rst` document, it is not also mentioned
at the top of the PR message template.

### DIFF
--- a/.github/contributing.rst
+++ b/.github/contributing.rst
@@ -87,50 +87,23 @@ Anatomy of a Pull Request
 =========================
 
 There is a GitHub Pull Request template to help guide crafting a description,
-and you can liberally copy content from the commit message as needed.
+and you can liberally copy content from the commit messages as needed.
 
 If your PR isn't quite ready feel free to create it as a draft, then once
 you're all set feel free to flip it to "Ready for review".
 
+The PR can have as many commits as you want, with multiple commits being
+preferred, as it can make review easier, especially when there are many
+changes.
 
-One Commit Per Pull Request
----------------------------
+When the PR is merged, a squashed version of the commits is automatically
+created (leaving both the original branch and the commits as they show up in
+the history on GitHub intact), with everything coming before the first section
+break (`---`) in the PR message being used as the commit message.
 
-Pull request should contain a single **passing** commit. This is necessary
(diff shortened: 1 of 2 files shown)


.github/contributing.rst
```diff
@@ -87,50 +87,23 @@ Anatomy of a Pull Request
...
 
 There is a GitHub Pull Request template to help guide crafting a description,
-and you can liberally copy content from the commit message as needed.
+and you can liberally copy content from the commit messages as needed.
 
 If your PR isn't quite ready feel free to create it as a draft, then once
 you're all set feel free to flip it to "Ready for review".
 
+The PR can have as many commits as you want, with multiple commits being
+preferred, as it can make review easier, especially when there are many
```
haxscramper, line 96: I would consider some more of the explanation from the PR description -- why not squash manually and other things mentioned there.

haxscramper (approved):
Aside from the minor comments it looks good to me

.github/contributing.rst
saem: No, this is how we end up with hard to follow docs. The writing style should be, keep everything very brief and to the point upfront, then elaborate later on.

Most people just want the instructions to follow and get going. We don't have to defend each point, nor do we have to convince people, right at that moment. If a would be contributor demands/feels they need that for them to follow along then they can get fucked.

saem (approved):

saem:
/merge

thanks for thinking of this, I forgot that the guidance was stale.

PR #807, opened 2023-07-19
cgen: fix performance regression with `x in {...}`

zerbina:
## Summary

Fix a regression in the C code generator that led to slightly less
efficient code being generated for `x in set` where set is a constant
set construction expression.

## Details

The decision of whether to emit set construction code or to use an
`||` (or) chain of comparisons was still based on the presence of the
`nfAllConst` flag, but node flags don't reach the code generators ever
since the introduction of the MIR and `astgen`, meaning that the `||`
chain was always chosen for set literals with less than 8 elements.

Testing for the flag is replaced with a call to `isDeepConstExpr`,
restoring the originally intended behaviour (i.e., emit set construction
code for sets with a size-in-bytes <= `sizeof(int)`).

In addition, the other remaining usage of `nfAllConst` in
`ccgexprs.genSetConstr` is removed completely: fully constant
construction expressions are already handled by its single callsite
(`ccgexprs.expr`).

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -1451,7 +1451,7 @@ proc fewCmps(conf: ConfigRef; s: PNode): bool =
   # this function estimates whether it is better to emit code
   # for constructing the set or generating a bunch of comparisons directly
   if s.kind != nkCurly: return false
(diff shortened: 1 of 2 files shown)


zerbina:
Using `ccodecheck` for detecting these kinds of issue works for now, but we're sooner or later going to need a different approach for reliably testing the output of the code generators.

saem (approved):

PR #808, opened 2023-07-20
internal: remove obsolete `nkExprColonExpr` detection

zerbina:
## Summary

Remove detection of `nkExprColonExpr` nodes in the context of tuple,
array, and object construction expressions in the code generators. The
ASTs reaching the code generators are expected to have a specific shape;
violations thereof need to be treated as internal errors and should
not be silently supported.

## Details

The AST coming out of `astgen` does not contain `nkExprColonExpr` nodes
in `nkTupleConstr` trees, so the detection for them in
`ccgexprs.genTupleConstr` and `vmgen.genTupleConstr` is obsolete and
thus removed. Since `jsgen.genTupleConstr` is used for both normal
procedure-level code and constant initializers, it cannot have the
detection removed, as constant initializers do use `nkExprColonExpr`
nodes for named tuple constructions (`nkTupleConstr`).

In addition, the constant initializer AST that comes out of both
`semfold` or `vmcompilerserdes`:
- doesn't have `nkExprColonExpr` nodes in `nkBracket` (array
  construction) trees
- always has `nkExprColonExpr` nodes in `nkObjConstr` trees

Therefore, `ccgexprs.getNullValueAux` (code generation for constant
object constructions) can depend on the presence of `nkExprColonExpr`
nodes and `ccgexprs.genConstSimpleList` (code generation for constant
array constructions) on their absence.

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -1995,9 +1995,7 @@ proc genTupleConstr(p: BProc, n: PNode, d: var TLoc) =
     let t = n.typ
     discard getTypeDesc(p.module, t) # so that any fields are initialized
     if d.k == locNone: getTemp(p, t, d)
-    for i in 0..<n.len:
-      var it = n[i]
-      if it.kind == nkExprColonExpr: it = it[1]
+    for i, it in n.pairs:
       initLoc(rec, locExpr, it, d.storage)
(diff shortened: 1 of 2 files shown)


saem (approved):
oh nice, I imagine there is a fair number of such simplifications that can be taken advantage of in the backends.

PR #809, opened 2023-07-20
vmgen: remove diagnostics representing internal errors

zerbina:
## Summary

Remove `vmgen`-related diagnostics that were about violations of
internal expectations/requirements. Detecting and reporting these kinds
of issues should not use diagnostics and reports, but rather `assert`
and `unreachable`.

## Details

The removed internal diagnostics are `adVmGenNotAFieldSymbol`,
`adGenDiagInvalidObjectConstructor`, and
`adVmGenCodeGenGenericInNonMacro` (which was unused).

All `nkDotExpr` (not-a-field-symbol diagnostic) and `nkObjConstr`
(invalid-object-constructor diagnostic) trees reaching `vmgen` are
produced by `astgen`, meaning that if they have an unexpected structure,
`astgen` doesn't work as it should, which is a severe internal error.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1045,13 +1045,13 @@ type
     adVmGenTooManyRegistersRequired       # |       defined in the `vmdef`
     adVmGenCannotFindBreakTarget          # |       module. There should be a
     adVmGenNotUnused                      # |       way to cross-reference data
-    adVmGenNotAFieldSymbol                # |       without introducing direct
+                                          # |       without introducing direct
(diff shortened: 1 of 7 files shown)


saem (approved):
the change looks good

although i don't recommend the `internalError` etc guidance in the PR message itself as their current implementation is corrupted by reports.

zerbina:
> although i don't recommend the `internalError` etc guidance in the PR message itself as their current implementation is corrupted by reports.

Yeah, it's probably better to leave it out -- I've removed it.

PR #811, opened 2023-07-22
sem: improve and rework `semobjconstr`

zerbina:
## Summary

Rework the semantic analysis of `nkObjConstr` nodes. The logic is now
simpler, compile-time evaluation happens from left to right, non-follow-
up errors are not dropped, and less follow-up errors are reported.

## Details

The idea is to split the processing into two phases:
1. AST validation and semantic analysis (i.e. figuring out the meaning)
2. filling in the gaps (e.g., fitting types) and making sure language
   rules are not violated

### Phase 1

During the first phase, for each slot, it is verified that the AST has
the expected shape and that a field with the given name exists in the
used object type; the initializer expression is also semantically
analyzed.

If the AST for a field assignment is not as expected, it is not analyzed
further. Compared to the previous logic, the "invalid field assignment"
error is not specialized further -- this is left to error message
rendering (currently `astDiagToLegacyReport`), thus making
`adSemFieldAssignmentInvalidNeedSpace` obsolete.

Errors in phase 1 do *not* short-circuit the second phase.

### Phase 2

The second phase is split up into two parts:
1. post-processing of the initializer expression and basic validity
  checks that only need the object construction expression
2. validity checks that also need to consider the object type

The post-processing includes (if possible) folding the initializer for
discriminators and fitting the expression's type to the formal type; the
basic validity checks are the field visibility and duplicated field
checks. Errors nodes produces here are placed directly into the
temporary AST.

This allows for dropping the `semExprFlagDispatched` procedure and the
`efPreferStatic`, `efPreferNilResult`, and `efPreferStatic` expression
flags.

Running the checks that require traversal/inspection of the object type
re-uses the pre-existing `semConstructTypeAux` and `semConstructFields`
(both which have their `sem` prefix replaced by `check`). Both only
inspect the construction expression now (no sem-checking or modification
is performed) and rely on the field assignments that reach them to either
be an `nkError` node or valid, which significantly simplifies error
handling. 

Since it is obsolete now, the `rsemFieldOkButAssignedValueInvalid` is
removed. A specialized wrapper for incorrect object constructions is not
needed (the generic wrapper suffices), so
`adSemObjectConstructorInvalid` is also removed.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1239,11 +1239,8 @@ type
     adSemConstExprExpected
     adSemExpectedRangeType
     # semobjconstr
-    adSemFieldAssignmentInvalidNeedSpace
     adSemFieldAssignmentInvalid
     adSemFieldNotAccessible
-    adSemFieldOkButAssignedValueInvalid
-    adSemObjectConstructorIncorrect
     adSemObjectRequiresFieldInitNoDefault
     adSemExpectedObjectType
     adSemExpectedObjectOfType
@@ -1364,9 +1361,7 @@ type
         adSemNamedExprNotAllowed,
         adSemNoReturnTypeDeclared,
         adSemReturnNotAllowed,
-        adSemFieldAssignmentInvalidNeedSpace,
         adSemFieldAssignmentInvalid,
-        adSemObjectConstructorIncorrect,
         adSemExpectedObjectType,
         adSemFoldOverflow,
         adSemFoldDivByZero,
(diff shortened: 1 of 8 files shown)


saem:
Just an initial observation, this separates expansion/synthesis (phase 1) and reduction/inheritance (phase 2). Definitely reads a lot cleaner and the amount of lines removed is great.

I'm doing a trace of the logic just to make sure I follow along, but a big fan thus far. 🥳

saem (approved):
Much improved. I followed along the logic and it seems to cover all the bases, nicely done.

PR #813, opened 2023-07-25
internal: use custom rendering for `--expandArc`

zerbina:
## Summary

Report the output for `--expandArc` via manual `msgWrite` calls instead
of using `localReport` and use custom render-to-text logic for the
`PNode` tree. The rendering logic is kept simple, and while the
output of `--expandArc` stays largely the same, it doesn't represent
valid NimSkull code anymore.

This is a preparation for the introduction of a code-generator IR, as
with it, `astgen` is not going to output `PNode` AST anymore, meaning
that `renderTree` cannot be used there.

## Details

The `rsemExpandArc` report is removed and everything associated with it.
While it could be kept, the general direction it to move aways from
reports for compiler tracing, and so `msgWrite` is instead used. This
also helps with getting around cyclic imports once the new IR is
introduced.

The rendering logic is a simplified version of `renderTree`, with
more complex text-layouting and conditional logic removed. The supported
AST shapes are similar to that of the planned initial version of the
code-generator IR, and the routines are added to the new `cgirutils`
module.

In order for changes to the `expandArc`-using tests to stay small, the
rendered output is kept compatible with that of
`renderTree(n, {renderIr, renderNoComment})`, where reasonable.

### DIFF
--- a/compiler/ast/report_enums.nim
+++ b/compiler/ast/report_enums.nim
@@ -818,7 +818,6 @@ type
 
     rsemEffectsListingHint
     rsemExpandMacro = "ExpandMacro" ## Trace macro expansion progress
-    rsemExpandArc = "ExpandArc"
 
     rsemCompilesReport
     rsemNonMatchingCandidates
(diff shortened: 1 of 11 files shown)


saem (approved):
overall change looks good.

taking a peek at CI seems `tests/lang_objects/destructor/tv2_cast.nim` needs an update.

PR #815, opened 2023-07-26
prevent RVO and in-place construction via an MIR pass

zerbina:
## Summary

Add a MIR pass that injects write-to-temporaries where necessary, which,
if enabled, guarantees to the C code generator that it can always use
RVO and in-place aggregate construction. This removes the dependency on
`isPartOf` from `cgen`, makes the return-value and in-place-construction
optimization related logic backend-agnostic, and also adds a basic
framework for MIR passes.

In theory, the VM code-generator could also use the in-place aggregate
construction optimization now.

In addition, multiple bugs where the in-place construction optimization
was not correctly prevented when using the C backend are fixed:
```nim
  # in the following statements, the left-hand side is (in terms of
  # behaviour) now fully evaluated first (previously it wasn't)
  arr = [1, arr[0]]
  arr = [1, (let v = arr[0]; v)]
  arr = [1, call(arr[0])]
  obj = Obj(a: 1, b: (var v = obj.a; v))

  # the following doesn't crash with an NPE anymore:
  obj = Obj(prc: ...)
  obj = Obj(prc: nil, other: obj.prc())
```

## Details

In preparation for the introduction of more MIR passes, a dedicated
module with the very basic framework is added. The only public routine
is the `applyPasses` procedure, which runs all passes enabled for the
specified backend. Since the targeted backend can be different from the
one selected by the user (because of compile-time execution), a
dedicated enum is used.

Due to their similarity in processing, the temporary injection for calls
and aggregate constructions is combined into a single pass. The pass
works by searching for assignments where the source operand is the
result of a call or aggregate construction and then running an analysis
for whether the assignment destination is used in a way that prevents
RVO or in-place construction. If the destination does, the source r-
value is assigned to an intermediate temporary first.

In the abstract, the used analysis works similar to `isPartOf`, but it
is, apart from the type-analysis, implemented from scratch for the MIR.

### `transf`

For the purpose of making assignments like `x = (x.a, x.b)` work when
using the C backend, `transf` previously unpacked the literal tuple
construction expression and then repacked it, like so:
```nim
let
  a = x.a
  b = x.b
x = (a, b)
```
While this does what is intended, it introduced an left-to-right
evaluation-order violation when the `x` has side effects. The new MIR
pass taking care of introducing a temporary allows for the removal of
``transformAsgn``, fixing the aforementioned issue and slightly reducing
the amount of work for the `injectdestructors` pass (because there are
less locals to analyze).

The removal does have an unintended side-effect: r-value inputs to
literal tuple construction expressions are not properly destroyed when
an exception is raised (see the changed `ttuple_unpacking.nim` test).
This is a known and documented issue for array and object constructions,
which now also affects tuple constructions.

### `cgen`

Calls that are the source operand to an assignment and return the result
via an out parameter always use the assignment destination as their
`Result` argument now; the `preventNrvo` procedure is renamed to
`reportObservableStore` and the related injection of temporaries
removed.

In `genObjConstr`, a temporary now only has to be created for `ref`s and
when the destination is an unset location.

For literal `seq` construction expressions (e.g., `@[1, 2]`), in-place
construction cannot be used anymore, since `isPartOf` usage is phased
out and the new MIR pass doesn't special-case `mArrToSeq`. However, with
the removal of the legacy GCs, the in-place `seq` construction had only
very little benefit, as using a temporary there only implies an
additional `(int, pointer)` local plus assignment thereof.

### Misc

Two internal bugs in the `mirtree` module are fixed:
- `operand` incorrectly ignored `mnkConsume` nodes
- the `SingleInputNode` set was missing `mnkStdConv`, `mnkPathNamed`,
  `mnkPathPos`, and `mnkPathVariant`. The aforementioned nodes now work
  with `unaryOperand`, as they should

### DIFF
--- a/compiler/mir/mirpasses.nim
+++ b/compiler/mir/mirpasses.nim
@@ -1,0 +1,286 @@
+## Implements the various MIR-based passes. The entry point is the
+## `applyPasses <#applyPasses>`_ procedure.
+
+import
+  compiler/ast/[
+    ast_query,
+    ast_types,
+    types
+  ],
+  compiler/mir/[
+    analysis,
+    mirchangesets,
+    mirconstr,
+    mirtrees,
+    sourcemaps
+  ],
+  compiler/sem/[
+    aliasanalysis
+  ],
+  compiler/utils/[
+    idioms
+  ]
+
+# for type-based alias analysis
+from compiler/sem/aliases import isPartOf, TAnalysisResult
+
+type
+  TargetBackend* = enum
+    ## The backend that is going to consume the MIR code. Used to select what
+    ## passes to run.
+    # XXX: this is the wrong abstraction. Application of the MIR passes doesn't
+    #      care about what backend is the target -- it cares about what the
+    #      targeted *language level* is (which is what the backend implies)
+    targetC
+    targetJs
+    targetVm
+
+  ArgIter = object
+    ## State for an iterator over an argument block. For efficiency, the inputs
+    ## to the arguments are returned in *reverse*
+    pos: NodePosition ## node position
+
+const
+  LocSkip = abstractRange + tyUserTypeClasses
+    ## types to skip to arrive at the underlying concrete value type
+
+func isRvalue(tree: MirTree, a: OpValue): bool {.inline.} =
+  ## Returns whether `a` represents an r-value, that is, something that
+  ## doesn't have a name and cannot be assigned to.
+  tree[a].kind in {mnkLiteral, mnkConstr, mnkObjConstr, mnkCall}
+
+func getRoot(tree: MirTree, n: OpValue): NodePosition =
+  ## Returns the root of a value. The root is either:
+  ## - the first operation yielding an lvalue (e.g., a pointer dereference)
+  ## - the name of a location (e.g., ``a`` in ``a.b.c``)
+  ## - an r-value (e.g., ``call()`` in ``call().b.c``)
+  const PathNodes = { mnkPathArray, mnkPathNamed, mnkPathPos, mnkPathVariant,
+                      mnkConv }
+    ## all operations that (can) take an lvalue as input and produce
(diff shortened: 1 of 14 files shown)


zerbina:
This is not fit for merging yet -- I've found some implementation bugs that need to be fixed first. The PR is still ready for review, however.

compiler/mir/mirpasses.nim
```diff
@@ -0,0 +1,286 @@
...
+    #       (pointers, refs, openArrays).
+    if tree[x].kind == mnkStdConv and
+       tree[x].typ.skipTypes(LocSkip + {tyVar}).kind == tyOpenArray:
+      OpValue(NodePosition(x) - 1)
+    else:
+      x
+
+  # we don't need to consider defs or initializing assignments (``mnkInit``)
+  # here, because there it is guaranteed that the destination does not appear
+  # anwhere in the source expression
```
```suggestion
  # anywhere in the source expression
```

tests/assign/tin_place_construction.nim
```diff
@@ -0,0 +1,93 @@
+discard """
+  targets: "c js vm"
+  description: '''
+    Ensure that the in-place aggregate construction optimization doesn't affect
+    observable semantics
+  '''
+"""
+
+# assignments are always evluated left-to-right. Optimizing an assignment with
```
saem:
```suggestion
# assignments are always evaluated left-to-right. Optimizing an assignment with
```

saem (approved):
Only found some doc typos, they aren't blockers.

The tests are nice and thorough, the new pass seems tighter and easier to follow.

zerbina:
Thank you for the review, @saem! The logic unfortunately ended up more complex than I had hoped, but I think that this can be alleviated/remedied through some adjustments of the MIR's design.

I've fixed the issues I mentioned I had noticed. They were mostly with sets missing some node kinds.

PR #816, opened 2023-07-26
sem: all tuple fields now have an owning sym

saem:
## Summary

All tuple field symbols now have an owning tuple type symbol.

## Details

For tuples without names create a anonymous sym as
an owner. This means all fields in a tuple will have a shared owner,
allowing for querying for reflection or code-gen.

This required modifying `semAnonTuple` and `semTuple` procedures in the
`semtypes` module. Now these procedures check to see if an owning symbol
already exists and if not an anonymous one is created.

---

## Notes for Reviewers
* first sketch of the solution, not all that pretty
* either in this PR, or in subsequent PRs, clean things up; lots of unprincipled code here

### DIFF
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -497,6 +497,17 @@ proc semAnonTuple(c: PContext, n: PNode, prev: PType): PType =
   if n.len == 0:
     localReport(c.config, n, reportSem rsemTypeExpected)
   result = newOrPrevType(tyTuple, prev, c)
+  let setupOwner = c.inGenericContext == 0 and result.sym.isNil
+  if setupOwner:
+    # xxx: instead of "AnonType" maybe generate a name from the field
+    #      names+types with some caching so the same structures have the same
+    #      names/symbols.
+    let sym = newSym(skType, getIdent(c.cache, "AnonType"),
+                      nextSymId c.idgen, c.getCurrOwner(), n.info)
+    sym.flags.incl sfAnon
+    sym.typ = result
+    result.owner = sym
+    pushOwner(c, sym)
   for it in n:
     let t = semTypeNode(c, it, nil).skipIntLit(c.idgen)
     if prev != nil and prev.aliasesType(t):
@@ -505,6 +516,8 @@ proc semAnonTuple(c: PContext, n: PNode, prev: PType): PType =
       break
     else:
       rawAddSon(result, t)
+  if setupOwner:
+    popOwner(c)
 
 proc semTuple(c: PContext, n: PNode, prev: PType): PType =
   # TODO: replace with a node returning variant that can in band errors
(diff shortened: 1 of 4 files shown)


saem:
tests seem to pass locally, just doing a quick check here.

compiler/sem/semtypes.nim
```diff
@@ -500,6 +497,17 @@ proc semAnonTuple(c: PContext, n: PNode, prev: PType): PType =
   if n.len == 0:
     localReport(c.config, n, reportSem rsemTypeExpected)
   result = newOrPrevType(tyTuple, prev, c)
+  let setupOwner = c.inGenericContext == 0 and result.sym.isNil
```
zerbina, line 500: I haven't verified it, but I think the check is an issue in the following case:
```nim
  type
    Generic[T] = object
      f: tuple[a, b: int]
```
When the tuple type is sem-checked, `inGenericContext` is greater than 0 and `result.sym` is nil.

compiler/sem/semtypes.nim
```diff
@@ -500,6 +497,17 @@ proc semAnonTuple(c: PContext, n: PNode, prev: PType): PType =
...
+  if setupOwner:
+    # xxx: instead of "AnonType" maybe generate a name from the field
+    #      names+types with some caching so the same structures have the same
+    #      names/symbols.
+    let sym = newSym(skType, getIdent(c.cache, "AnonType"),
+                      nextSymId c.idgen, c.getCurrOwner(), n.info)
```
```suggestion
    let sym = newAnonSym(c, skType, n.info)
```
While it's slightly different than what you have (`c.cache.idAnon` is used as the identifier), it would mean that all anonymous symbols are created the same way. 

`newAnonSym` in `semexprs` would have to be moved into `sem` (or maybe `semdata`), however.

compiler/sem/semtypes.nim
```diff
@@ -508,13 +516,26 @@ proc semAnonTuple(c: PContext, n: PNode, prev: PType): PType =
...
   var typ: PType
   result = newOrPrevType(tyTuple, prev, c)
   result.n = newNodeI(nkRecList, n.info)
+  let setupOwner = c.inGenericContext == 0 and result.sym.isNil
+  if setupOwner:
+    # xxx: instead of "AnonType" maybe generate a name from the field
+    #      names+types with some caching so the same structures have the same
+    #      names/symbols.
+    let sym = newSym(skType, getIdent(c.cache, "AnonType"),
+                      nextSymId c.idgen, c.getCurrOwner(), n.info)
```
```suggestion
    let sym = newAnonSym(c, skType, n.info)
```

compiler/sem/semtypes.nim
```diff
@@ -500,6 +497,17 @@ proc semAnonTuple(c: PContext, n: PNode, prev: PType): PType =
...
+    sym.flags.incl sfAnon
+    sym.typ = result
+    result.owner = sym
```
zerbina, line 508: If a symbol directly represents the type, then it is generally assigned to the types's `sym` field instead of `owner`, which for consistency would also make sense here, I think. (`typesrenderer` already checks for `sfAnon` symbols, so this *should* not cause problems)

zerbina (commented):
I've noticed a potential issue with the `setupOwner` condition, and there might be an expectation violation with assigning the anonymous symbol to the type's `owner`, but I think that the direction of assigning symbols to anonymous types is good.

It's nothing that has to happen as part of this PR, but using an anonymous symbol would also make sense for anonymous procedural types.

compiler/sem/semtypes.nim
saem: Doh, good call.

I was suspicious of the owner pushing in the right hand side type section analysis for generics, and I think that's coming home to roost here.

compiler/sem/semtypes.nim
saem: It did, but more so for the silly `getTypeDecl` and that whole silly API in `std/macros`. I'll have to dig in further to see what's going wrong there.

saem:

compiler/sem/semtypes.nim
saem: combining them is a good idea, `newAnonSym` can also use the passed in symbol kind param to specialize things like the name (anon identifier vs type vs whatever).

compiler/sem/semtypes.nim
saem: Done

compiler/sem/semtypes.nim
saem: done

saem:
the approach I'm taking here isn't quite right. I think more needs to be cleaned up and then a less hacky approach to symbol creation can be taken.

PR #818, opened 2023-07-29
mirpasses: make call-argument fixup a MIR pass

zerbina:
## Summary

Replace the `PNode`-based fixup for call arguments in `ccgcalls` with a
MIR pass. This removes a dependency on `PNode`-based analysis from
`cgen`, and also makes it possible to, in the future, enable the fixup
for the JS or VM backends.

While the used analysis stays mostly the same, an observable evaluation-
order violation is fixed. Injecting (shallow) copies for values passed
to both immutable and mutable parameters now considers *all* parameters,
instead of only parameters to the right of immutable ones. For example,
given:
```nim
f(a, a) # proc f(x: var T, y: T)
```
this fixes mutations through `x` inside `f` being visible on `y`.

## Details

The analysis used by the MIR pass works mostly the same as the one
previously used in `ccgcalls`: for each argument value that is not
explicitly passed by-reference, it is analyzed whether:
- the value is potentially mutated *after* it is bound to the parameter
  but *before* the procedure is called
- the value is potentially also passed to a `var` parameter
If either is the case, the argument is shallow-copied to a temporary
that is then passed to the parameter instead.

The differences compared to the previous analysis are that:
- checking whether the argument value (or something that potentially
  overlaps with it in memory) is also passed to a mutable parameter
  now also considers preceding parameters. Previously, only the
  following parameters were checked
- testing for overlapping values considers the whole path now, instead
  of only the root. In effect, this means that for `f(a.x, a.y)`, where
  the second parameter is mutable, no temporary is (unnecessarily)
  injected for the first parameter

For overlap testing, the `maybeSameMutableLocation` procedure is
introduced, which mirrors the behaviour of `dfa.aliases` (the routine
used by the previous `PNode`-based analysis). Since dereferences of
pointer-like values are treated like a normal field access, calls like
`f(a[].x, b[].y)`, where one of the parameters is mutable and `a` and
`b` point to the same location, still cause observable evaluation-order
violations.

Finally, the analysis and temporary injection in `ccgcalls` is removed,
and a test is added for the fixed issue.

### DIFF
--- a/compiler/mir/mirpasses.nim
+++ b/compiler/mir/mirpasses.nim
@@ -335,8 +335,8 @@ proc fixupCallArguments(tree: MirTree, config: ConfigRef,
   ## Do note that due to the placement of this pass (it happens after the
   ## ``injectdestructors`` pass), only *shallow*, non-owning copies of the
   ## affected arguments are made, meaning that there's the issue of resource-
-  ## like values (refs, seqs, strings, everything else that has a destructor)
-  ## being freed while they're still reachable through an immutable parameter.
+  ## like values (refs, seqs, strings, everything else that has non-trivial
+  ## copy behaviour) not being duplicated properly.
 
   proc maybeSameMutableLocation(tree: MirTree, a, b: OpValue): bool {.nimcall.} =
     ## Compares the path of both `a` and `b` to check whether they could
--- a/compiler/mir/mirpasses.nim
+++ b/compiler/mir/mirpasses.nim
@@ -442,7 +442,7 @@ proc fixupCallArguments(tree: MirTree, config: ConfigRef,
           continue
 
         var needsTemp = false
-        block:
+        block checkIfArgNeedsTemp:
           # first, check whether the argument value is passed to an earlier
           # ``var`` parameter of the procedure (e.g., given the call
           # ``f(a, b, c, d)`` and analysing 'c', check 'a' and 'b'. 'd' is
@@ -453,15 +453,14 @@ proc fixupCallArguments(tree: MirTree, config: ConfigRef,
             if tree[arg2].kind == mnkName and tree[arg2 - 1].kind == mnkTag:
               if maybeSameMutableLocation(tree, val, OpValue(arg2 - 2)):
                 needsTemp = true
-                break
+                break checkIfArgNeedsTemp
 
-        if not needsTemp:
           # look for potential mutations of the argument that happen after
           # the argument is bound
           for mut in potentialMutations(tree, arg + 1, i - 2):
             if maybeSameMutableLocation(tree, val, mut):
               needsTemp = true
-              break
+              break checkIfArgNeedsTemp
 
         if needsTemp:
           let temp = MirNode(kind: mnkTemp, typ: tree[arg].typ,
--- a/compiler/mir/mirpasses.nim
+++ b/compiler/mir/mirpasses.nim
@@ -387,7 +387,7 @@ proc fixupCallArguments(tree: MirTree, config: ConfigRef,
     result = true # until proven otherwise
     for i in 1..min(an.len, bn.len):
       let a {.cursor.} = tree[an[^i]]
-      let b {.cursor.} = tree[bn[^1]]
+      let b {.cursor.} = tree[bn[^i]]
 
       template check(cond: bool) =
         if cond:
@@ -407,7 +407,7 @@ proc fixupCallArguments(tree: MirTree, config: ConfigRef,
       of mnkPathPos:
         check(a.position != b.position)
       of mnkPathArray:
-        check(sameIndex(tree[an[i] - 3], tree[bn[i] - 3]) == no)
+        check(sameIndex(tree[an[^i] - 3], tree[bn[^i] - 3]) == no)
       else:
         unreachable()
 
--- a/compiler/mir/mirpasses.nim
+++ b/compiler/mir/mirpasses.nim
@@ -24,6 +24,10 @@ import
 # for type-based alias analysis
 from compiler/sem/aliases import isPartOf, TAnalysisResult
 
+# XXX: required for computing the size of types. Ideally, the size would
+#      be fully computed already at this stage of compilation...
+from compiler/front/options import ConfigRef
+
 type
   TargetBackend* = enum
     ## The backend that is going to consume the MIR code. Used to select what
@@ -88,6 +92,17 @@ func skipTag(tree: MirTree, a: OpValue): OpValue {.inline.} =
   if tree[a].kind == mnkTag: OpValue(NodePosition(a) - 1)
   else:                      a
 
+func skipOpParam(tree: MirTree, n: OpValue): OpValue =
+  ## Returns the value that was passed to the surrounding region's argument
+  ## corresponding to the given ``mnkOpParam``.
+  assert tree[n].kind == mnkOpParam
+  let p = findParent(tree, NodePosition n, mnkRegion)
+  if isArgBlock(tree, p - 1):
+    result = operand(tree, Operation(p), tree[n].param)
+  else:
+    # it's a single argument region
+    result = getStart(tree, p - 1).OpValue
+
 func initArgIter*(tree: MirTree, call: NodePosition): ArgIter =
   assert tree[call].kind in { mnkCall, mnkMagic, mnkRegion, mnkAsgn,
                               mnkFastAsgn, mnkInit, mnkPathArray, mnkAsm,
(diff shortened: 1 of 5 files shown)


compiler/mir/mirpasses.nim
```diff
@@ -140,6 +160,22 @@ iterator uses(tree: MirTree, start, last: NodePosition): OpValue =
...
+iterator potentialMutations(tree: MirTree, start, last: NodePosition): OpValue =
+  ## Returns in an unspecified order all values that are:
+  ## - passed to parameters that allow for mutations
+  ## - turned into pointer rvalues (i.e., have their address taken)
+  ## - have mutable views created of them
+  ## None of the above means that the used value is really mutated in the
+  ## specified range, rather it means that the value *could* be mutated.
+  var i = start
+  while i <= last:
+    # all ``mnkTag`` nodes currently imply some sort of mutation/change
```
zerbina, line 172: I'm increasingly starting to think that generalizing the original `mnkModify` into `mnkTag` was a mistake. As the comment mentions, all current value tags represent some form of mutation, but the name `mnkTag` doesn't really convey that.

I'll think about it some more, but my current opinion is that `mnkTag` should become `mnkMutation` (or similar).

saem: unless the intention is to generalize across other effects, I agree I think a name indicating mutation/memory makes sense.

compiler/mir/mirpasses.nim
```diff
@@ -277,8 +313,174 @@ proc preventRvo(tree: MirTree, changes: var Changeset) =
...
+  ## There is another, somewhat related, problem: the underlying location of an
+  ## immutable pass-by-reference parameter can overlap with that of a ``var``
+  ## parameter, in which case changes through the ``var`` parameter are
+  ## observable on the immutable parameter. This problem is also addressed by
+  ## the fixup.
+  ##
+  ## Do note that due to the placement of this pass (it happens after the
+  ## ``injectdestructors`` pass), only *shallow*, non-owning copies of the
+  ## affected arguments are made, meaning that there's the issue of resource-
+  ## like values (refs, seqs, strings, everything else that has a destructor)
```
saem: I'm increasingly convinced that destructors, as realized are a misfeature, or one that should be deemphasized. instead we should trigger ops on the containing type, which would be better for data oriented programming and allow for better composition. the op interface might be: `=receiveCopy`, `=takeOwnership`, `=endOfLife`, etc.

compiler/mir/mirpasses.nim
```diff
@@ -277,8 +313,174 @@ proc preventRvo(tree: MirTree, changes: var Changeset) =
...
+    # we now compare the two paths, and if one ends while they haven't
+    # diverged, we treat the values as overlapping in memory. Since derefs are
+    # skipped, there can be both false positives and false negatives, as the
+    # underlying location cannot be statically known when dereferences are
+    # involved. For the immediate use case here, the heuristic works okay
+
+    result = true # until proven otherwise
+    for i in 1..min(an.len, bn.len):
+      let a {.cursor.} = tree[an[^i]]
+      let b {.cursor.} = tree[bn[^1]]
```
saem: if this is consistent across iterations should it be pulled out of the loop.

tests/ccgbugs/targ_lefttoright.nim
```diff
@@ -65,6 +65,16 @@ template test =
   var h = 1
   say h, h +=+ 1 # 1,2
 
+  proc mutateFirst(a: var C, b: C) =
+    a.i = 2
+    doAssert b.i == 1
+
+  var i = C(i: 1)
+  when nimvm: # XXX: doesn't work yet
+    discard
```
saem, line 74: this gave me an idea, maybe we should add a `doAssertKnownIssue` that'll work with inverted behaviour.

saem (approved):
very cool

compiler/mir/mirpasses.nim
zerbina: Oh, thank you for noticing this. That's a severe bug, it should have been `^i`.

tests/ccgbugs/targ_lefttoright.nim
zerbina: I like it, I've come across multiple cases were that would have been useful.

compiler/mir/mirpasses.nim
```diff
@@ -277,8 +313,174 @@ proc preventRvo(tree: MirTree, changes: var Changeset) =
...
+                needsTemp = true
+                break
+
+        if not needsTemp:
+          # look for potential mutations of the argument that happen after
+          # the argument is bound
+          for mut in potentialMutations(tree, arg + 1, i - 2):
+            if maybeSameMutableLocation(tree, val, mut):
+              needsTemp = true
+              break
```
Clyybber: We could move both loops into a block and break out of that common block. I think the current way is quite readable too though.
```suggestion
        block checkIfArgNeedsTemp:
          # first, check whether the argument value is passed to an earlier
          # ``var`` parameter of the procedure (e.g., given the call
          # ``f(a, b, c, d)`` and analysing 'c', check 'a' and 'b'. 'd' is
          # analysed as part of the ``potentialMutations`` loop below)
          var iter2 = iter
          while iter2.hasNext:
            let arg2 = rawNext(iter2, tree)
            if tree[arg2].kind == mnkName and tree[arg2 - 1].kind == mnkTag:
              if maybeSameMutableLocation(tree, val, OpValue(arg2 - 2)):
                needsTemp = true
                break checkIfArgNeedsTemp

          # look for potential mutations of the argument that happen after
          # the argument is bound
          for mut in potentialMutations(tree, arg + 1, i - 2):
            if maybeSameMutableLocation(tree, val, mut):
              needsTemp = true
              break checkIfArgNeedsTemp
```

Clyybber (approved):

compiler/mir/mirpasses.nim
Clyybber: What does `instead we should trigger ops on the containing type` mean concretely, since the current hooks are attached to the respective type?

compiler/mir/mirpasses.nim
zerbina: Hm, yeah, that's probably cleaner. With the separate `if` statement, I think my intention was to make it clear that `iter2` is only meant to be used with the first loop, but the code is short enough for this to be visible without the separate scope.

compiler/mir/mirpasses.nim
saem:
```nim
type
  seq[T] = object # for discussion assume this definition
    data: uncheckedArray[T]
    len: int

proc `=takeOwnership`[T](s: seq[T], t: T) = # ... whatever default impl we want

# `seq` is a container and its 'containers ops' are triggered based on a `T`'s lifetime (like current type ops), but call `seq[T]`'s 'container ops'
# this also opens the door on specialization based on the contained type and overriding might be feasible

proc `=takeOwnership`(s: seq[Uri], u: Uri) = # ... specialized impl
```

I'm not sure about the actual events/hooks we want, but what I'm somewhat certain about is type ops triggered and called on the same type is 'off'. To clarify, `Uri` could have container ops of its own under this scheme, but they'd apply to the pile of `string` fields that are in its type definition and not the `Uri` type itself.

---

I had separately started thinking about triggered ops on containers before and I was coming at it from a DOD perspective. That further corroborated that type ops on the actual type aren't as useful wrt correctness or saving work, below is a snippet I was using to think about this.

Using the below example, if we specialized the `Component` types with type ops we'd end up with something "wrong". Those types would start having to know where and how they're stored, plus others couldn't reuse the components and associated systems easily (poor composition).

```nim
type
  Position* = object
    x*, y*: int64
  Movement* = object
    dx*, dy*, ddx*, ddy*: float32
  Component* = Position | Movement
  EntityData = object
  Store* = object
    allocator: Allocator
    entities*: seq[EntityData]
    positions*: seq[Position]
    movements*: seq[Movement]
proc `=alloc`[T=Component or EntitData](s: Store, t: typedesc[T])
proc `=dealloc`[T=Component or EntitData](s: Store, a: T)
proc `=transferInto`[T=Component or EntitData](s: Store, a: T)
proc `=copyInto`[T=Component or EntitData](s: Store, a: T)
```

The last bit I should add is that we effectively do this sort of logic for the stack, the push/pop of activation records, lifting of envs and associated allocs/deallocs, etc. All that is more call stack/container focused, that's what's being bumped, that's the ever present context. If we were actually modelling a call stack, we wouldn't want to put the type ops on each call frame type (that's a _lot_ of object types 😆), we'd want activation record lifecycle events to trigger container ops on the call stack (container).

Anyhow, I hope that makes some sense.

PR #821, opened 2023-07-31
internal: less generic-expression workarounds

zerbina:
## Summary

Prevent expressions-depending-on-unresolved-type-variables passed to
`tryConstExpr` from reaching into the compile-time evaluation machinery.
This allows for removing multiple workarounds from `mirgen` and `vmgen`.

## Details

Before `tryConstExpr` passes the expression to `evalConstExpr`, it now
first checks whether the expression either contains unresolved generic
parameters or is an unresolved type-parameter-lookup (e.g., `T.param`).
If either is the case, `nil` (which signals that the expression cannot
be evaluated at compile-time) is returned without passing the
expression to `evalConstExpr`.

While this approach is fundamentally also a workaround (expressions
containing unresolved generic parameters shouldn't be passed to
`tryConstExpr` in the first place), it at least prevents unresolved
expressions from reaching into `transf`, `mirgen`, and `vmgen`.

`mirgen` now doesn't have to push unresolved generic parameters or
type-parameter-lookups as `mnkLiteral`s through the MIR phase, and
`vmgen` doesn't have to special-case them. More generally, `vmgen` can
now treat unexpected nodes reaching it as a defect (via `unreachable`).

Since `vmGenDiagCodeGenUnexpectedSym` and `vmGenDiagCannotGenerateCode`
are now unused, they are, together with everything related to them,
removed.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1046,13 +1046,13 @@ type
     adVmGenCannotFindBreakTarget          # |       module. There should be a
     adVmGenNotUnused                      # |       way to cross-reference data
                                           # |       without introducing direct
-    adVmGenCannotGenerateCode             # |       import or type
+                                          # |       import or type
     adVmGenCannotEvaluateAtComptime       # |       dependencies. Likely this
                                           # |       involves data oriented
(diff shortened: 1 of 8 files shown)


saem (approved):
Wow, it's surprising how much that code proliferated.

PR #822, opened 2023-07-31
semfold: fix range-check folding

zerbina:
## Summary

Fix literals resulting from folding range-check conversions having a
type that doesn't match the node kind. This, in turn, fixes a VM code
generator issue where the code generated for an access of the result of
an integer-literal-converted-to-float-range would cause the VM to crash.

## Details

The folding logic directly assigned the destination type to the literal,
which led to, for example, integer literals having a `tyFloat` type.
This breaks the assumption of only float-literal nodes having a float
type, and is what caused the `vmgen` bug (`vmgen` only looks at the node
kind).

Same as for normal conversion, range-check conversion are now also
folded via `foldConv` (but only in case the range check succeeds),
fixing the issue.


saem (approved):
The explicit conversion and applying the type there makes more sense.

PR #823, opened 2023-08-01
semfold: fix conversion folding

zerbina:
## Summary

Make the logic handling folding of integer- and float-like conversion
more exhaustive and slightly simplify it. This fixes the internal issue
of literal-nodes having unexpected types, and it also fixes no range
error being reported at compile-time for some to-`char` conversion.

## Details

Folding conversions to integer, float, and bool types now considers all
valid source types, with unexpected types now being treated as a defect
instead of silently ignoring them. Named set constants are used for
this, with `tyChar` being included in the integer-like set, meaning that
to-`char` conversions are now properly subjected to range checks during
constant folding.

For the integer-like branch handling, the logic is simplified to only
perform the range check in a single place, and the unnecessary
transition-to-`nkUIntLit` removed (`newIntNodeType` already makes sure
that the node has the correct kind based on the type).

In general, this is a step towards the type of a literal-node matching
its kind. Both the C and VM code generator use the node kind for
deciding whether a float literal is a 32-bit or 64-bit one, so the
float-literal-nodes produced by constant folding now having the correct
kind fixes a bug where the emitted float literals had the wrong target-
language type (see the fixed test `tfloats.nim`).

### DIFF
--- a/compiler/sem/semfold.nim
+++ b/compiler/sem/semfold.nim
@@ -439,28 +439,26 @@ proc foldConv(n, a: PNode; idgen: IdGenerator; g: ModuleGraph; check = false): P
              PAstDiag(kind: adSemFoldRangeCheckForLiteralConversionFailed,
                       inputLit: a))
 
-  # if srcTyp.kind == tyUInt64 and "FFFFFF" in $n:
-  #   echo "n: ", n, " a: ", a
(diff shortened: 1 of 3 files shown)


saem (approved):
I'll try a regular build that tomorrow, IIRC these were one of the issues. 👍🏽

zerbina:
I've reverted the unrelated change to the test such that there are no merge conflicts with #824.

PR #824, opened 2023-08-02
jsgen: fix incorrect `float32`-literal behaviour

zerbina:
## Summary

Narrow the value of 32-bit float-literal nodes prior to rendering them,
which fixes `float32` literals having an unexpected run-time value
when their value as-written is not fully representable with the
precision of a 32-bit float.

## Details

Instead of using `toStrMaxPrecision`, which would add an unwanted `f`
suffix, rendering the float value to text is done by directly calling
`addFloatRoundtrip`, which `toStrMaxPrecision` uses internally.

### DIFF
--- a/compiler/backend/jsgen.nim
+++ b/compiler/backend/jsgen.nim
@@ -21,6 +21,9 @@ Check `mapType` for the details.
 """
 
 import
+  system/[
+    formatfloat
+  ],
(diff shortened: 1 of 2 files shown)


saem (approved):
nice and things are more uniform to boot.

PR #825, opened 2023-08-05
`std/json`: remove macro type API usage

saem:
## Summary

Remove usage of `macros` type API usage in `std/json`, specifically
`getTypeImpl` and `getTypeInst`. This is part of a broader effort to
drop this API from the standard library. The current API enforces a
number of poor design decisions in compiler internals. The API itself
is questionable and much of this work should happen via generics.

## Details

As part of removing the usage of these APIs, there was a slight
regression in the precision of errors when a tuple without named fields
is used. The error message is equivalent, but the error location will
be within the `json` module itself.

The change itself uses the `fieldPairs` iterator which relies upon
compiler support in `semfields`. Some minor adjustments were made here
in particular the ability to handle `nkError` nodes.

The `json` module relies upon the ability to forward declare generic
procedures, this was not possible with the earlier bootstrap compilers,
and existed behind a conditional compilation flag
(`nimFixedForwardGeneric`). Seeing as this is no longer the case the
flag has been removed as part of the module rework.

Finally, the test `tjsonmacro_reject` existed for the previous approach,
but the test are combined into `tjsonmacro` and the previous test is
removed.


saem:
~Next thing to fix is handling the error pragma handling, presently only in the VM, it probably needs to be in `sempass2`, as well.~

`error` in macros is implemented a bit oddly, it can only execute in the VM and it's far too general a name. the whole api is infuriating, what should be a simple static call that can be folded is instead tangled up in macro land.

lib/pure/json.nim
```diff
@@ -1046,310 +1046,241 @@ template verifyJsonKind(node: JsonNode, kinds: set[JsonNodeKind],
...
-    when T is uint|uint64 or (not defined(js) and int.sizeof == 4):
-      verifyJsonKind(jsonNode, {JInt, JString}, jsonPath)
-      case jsonNode.kind
-      of JString:
-        let x = parseBiggestUInt(jsonNode.str)
-        dst = cast[T](x)
-      else:
-        dst = T(jsonNode.num)
+macro isRefSkipDistinct*(arg: typed): untyped =
+  ## internal only, do not use
```
zerbina: The macro can now be moved into the `tjson.nim` test, as that's the only remaining place where it's used.

saem: I was thinking of deleting it and the tests.

zerbina: Oh, that's indeed the better choice. I've disregarded that without any usage, there's no point in testing the macro :D

compiler/sem/semtypes.nim
```diff
@@ -986,7 +984,7 @@ proc skipGenericInvocation(t: PType): PType {.inline.} =
 
 proc addInheritedFields(c: PContext, check: var IntSet, pos: var int,
                         obj: PType) =
-  assert obj.kind == tyObject
+  assert obj.kind == tyObject, $obj.kind
```
zerbina, line 987:
```suggestion
  assert obj.kind == tyObject
```

lib/pure/json.nim
```diff
@@ -1046,310 +1046,211 @@ template verifyJsonKind(node: JsonNode, kinds: set[JsonNodeKind],
...
-  proc initFromJson[S,T](dst: var array[S,T]; jsonNode: JsonNode; jsonPath: var string) =
-    verifyJsonKind(jsonNode, {JArray}, jsonPath)
-    let originalJsonPathLen = jsonPath.len
-    for i in 0 ..< jsonNode.len:
-      jsonPath.add '['
-      jsonPath.addInt i
-      jsonPath.add ']'
-      initFromJson(dst[i.S], jsonNode[i], jsonPath) # `.S` for enum indexed arrays
-      jsonPath.setLen originalJsonPathLen
+# The following forward declarations don't work in older versions of Nim
```
zerbina:
```suggestion
```
Given that the conditional symbol is removed, this comment is probably not needed anymore.

lib/pure/json.nim
```diff
@@ -1046,310 +1046,211 @@ template verifyJsonKind(node: JsonNode, kinds: set[JsonNodeKind],
...
+      dst = some(default(T))
+    initFromJson(dst.get, jsonNode, jsonPath)
+
+from std/typetraits import isNamedTuple, distinctBase
+
+proc initFromJson[T: object|tuple](dst: var T, jsonNode: JsonNode, jsonPath: var string) =
+  let originalJsonPathLen = jsonPath.len ## so we can trucate/reset after
+  when T is tuple:
+    when T.isNamedTuple:
+      # TODO: nimvm workaround for distinct?
```
zerbina:
```suggestion
```
The comment seems to be stale.

lib/pure/json.nim
```diff
@@ -1046,310 +1046,211 @@ template verifyJsonKind(node: JsonNode, kinds: set[JsonNodeKind],
...
-      doAssert data.person.age == 21
-      doAssert data.list == @[1, 2, 3, 4]
-
-    var jsonPath = ""
-    initFromJson(result, node, jsonPath)
+      {.error("Use a named tuple instead of: " & T.repr).}
+  elif T is object:
+    jsonPath.add "."
+    for name, value in dst.fieldPairs:
+      # xxx: might need unsafe assign due to kind fields
```
zerbina:
```suggestion
```
This one too.

lib/pure/json.nim
```diff
@@ -1046,310 +1046,211 @@ template verifyJsonKind(node: JsonNode, kinds: set[JsonNodeKind],
...
-      # both, `dst` and `typeSym` don't have good lineinfo. But nothing
-      # else is available here.
-      detectIncompatibleType(typeSym, dst)
-      foldObjectBody(result, typeSym, dst, jsonNode, jsonPath, originalJsonPathLen)
+      dst = some(default(T))
+    initFromJson(dst.get, jsonNode, jsonPath)
+
+from std/typetraits import isNamedTuple, distinctBase
+
+proc initFromJson[T: object|tuple](dst: var T, jsonNode: JsonNode, jsonPath: var string) =
```
zerbina: I think splitting this into two overloads (one for tuples, one for objects) would be better, as there's no need for the outer `when` statement then.

zerbina (approved):
Looks good to me! It's a lot less code now, and the version using generics is much easier to understand too. I've left a few clean-up suggestions.

As an aside, the removal of `nimFixedForwardGeneric` makes the final change-set a bit hard to read. Turning the removal into a separate PR would be nice, but it's not something blocking -- no change expected from my side here.

zerbina:
I've applied some small typo and grammar fixes to the PR message.

compiler/sem/semtypes.nim
saem: wouldn't it be easier to debug by leaving the kind in there?

lib/pure/json.nim
saem: Done.

compiler/sem/semtypes.nim
zerbina: It would, although I personally think that assertions with `$x` are visually a bit noisy, hence my suggestion. It's nothing critical, however, and I'm fine with it staying as it is.

It's nothing for this PR, but since asserting that the discriminator of some object has a certain value is a common pattern in the compiler, I think that adding a new idiom `assertKind` (which automatically renders the value on failure) would make sense.

compiler/ast/ast_query.nim
```diff
@@ -331,7 +331,7 @@ proc hasNilSon*(n: PNode): bool =
 proc containsNode*(n: PNode, kinds: TNodeKinds): bool =
   if n == nil: return
   case n.kind
-  of nkEmpty..nkNilLit: result = n.kind in kinds
+  of nkEmpty..nkNilLit, nkError: result = n.kind in kinds
```
```suggestion
  of nkWithoutSons: result = n.kind in kinds
```
This fixes `containsNode` for all other nodes that don't have sub-nodes.

PR #826, opened 2023-08-06
codegen: introduce dedicated node kind for magics

zerbina:
## Summary

Add the `cnkMagic` node kind, which is meant for referencing magic
procedures,  and integrate it into the code generators. References
to magic procedures that don't use symbols during the MIR phase (i.e.,
`mnkMagic` nodes) don't require the creation of a `PSym` in `cgirgen`
now.

For compatibility with the existing implementation, the `magic` field
of symbols is still inspected during code generation, but the direction
is to have only `cnkMagic` mean "magic procedure" in the code
generators.

## Details

In the code generators, all manual `TSym.magic` lookups, outside of type
contexts, are replaced with calls to either `compat.getMagic` or the new
`compat.getCalleeMagic`. Both procedures consider both `cnkSym` and
`cnkMagic` nodes.

An internal problem that the new node fixes is "improper symbols"
reaching the code generators: a symbol created via `createMagic` is
missing much of what is need for a proper procedure symbol, such as a
type, associated AST, owner, etc. -- with `mnkMagic`, these symbols are
no longer created.

As a side-effect, `render` now uses the enum name for magic-procedure
references using `cnkMagic`; the tests using `--expandArc` are adjusted
to account for this.

### DIFF
--- a/compiler/backend/cgirutils.nim
+++ b/compiler/backend/cgirutils.nim
@@ -133,7 +133,7 @@ proc render(c: var RenderCtx, ind: int, n: CgNode, res: var string) =
     if sfCursor in n.sym.flags:
       res.add "_cursor"
   of cnkMagic:
-    # cut of the 'm' prefix and use lower-case for the first character
+    # cut off the 'm' prefix and use lower-case for the first character
     var name = substr($n.magic, 1)
     name[0] = toLowerAscii(name[0])
     res.add name
--- a/compiler/backend/cgirutils.nim
+++ b/compiler/backend/cgirutils.nim
@@ -47,6 +47,9 @@ proc treeRepr*(n: CgNode): string =
       result.add n.sym.name.s
       result.add " id: "
       result.add $n.sym.itemId
+    of cnkMagic:
+      result.add "magic: "
+      result.add $n.magic
     of cnkEmpty, cnkInvalid, cnkType, cnkAstLit, cnkNilLit, cnkReturnStmt:
       discard
     of cnkWithItems:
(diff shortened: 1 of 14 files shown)


compiler/backend/cgirutils.nim
```diff
@@ -129,6 +132,11 @@ proc render(c: var RenderCtx, ind: int, n: CgNode, res: var string) =
     # highlight cursor locals
     if sfCursor in n.sym.flags:
       res.add "_cursor"
+  of cnkMagic:
+    # cut of the 'm' prefix and use lower-case for the first character
```
saem:
```suggestion
    # cut off the 'm' prefix and use lower-case for the first character
```

compiler/backend/compat.nim
```diff
@@ -156,10 +159,9 @@ proc isLValue*(n: CgNode): bool =
 
 proc canRaiseConservative*(fn: CgNode): bool =
```
saem, line 160: no change in this PR, but I think a better name might be `allowedToRaise`, whereas `canRaise` conjures up a meaning of "could this callee raise", the former is more like "regardless of whether it does, do we allow the possibility of it raising".

saem (approved):
🪄 

found one non-blocking typo in a doc comment, otherwise it's good to merge.

compiler/backend/cgirutils.nim
zerbina: Thanks.

compiler/backend/compat.nim
zerbina: At least in the way it's currently used, `canRaiseConservative` does answer the question "could this callee raise". My current plan is to collapse both `canRaiseConservative` and `canRaise` into just `canRaise`.

PR #828, opened 2023-08-06
codegen: lower `swap` via a MIR pass

zerbina:
## Summary

Implement the lowering of `swap` as a MIR pass instead of directly in
each code generator. This removes the amount of code duplication in the
compiler and fixes a double-evaluation bug for call-expression `swap`
operands when using the JS backend.

MIR passes are now also applied for code run at compile time.

## Details

The MIR passes were previously not applied to code processed by `vmjit`
(compile-time code NimScript execution), which is fixed now.

To be able to efficiently detect and lower the call with a MIR pass,
`mSwap` is transformed into the MIR magic call representation (which
uses `mnkMagic` instead of `mnkCall`). The lowering pass then only has
to search for the `mnkMagic` node and replace it.

Implementation-wise, a `swap` call is lowered into shallow assignments
(`mnkFastAsgn`). For the C target, the code that is later generated
stays the same; for the VM target, the efficiency slightly increases for
operands that aren't locals; and for the JS target, efficiency regresses
for operands of aggregate types. The latter is the case because the
JavaScript code generator currently ignores shallow assignments
(`cnkFastAsgn`) and performs full copies.

### DIFF
--- a/compiler/mir/mirpasses.nim
+++ b/compiler/mir/mirpasses.nim
@@ -148,7 +148,7 @@ iterator search(tree: MirTree, kinds: static set[MirNodeKind]): NodePosition =
     inc i
 
 iterator search(tree: MirTree, magic: static TMagic): NodePosition =
-  ## Returns in order appearance the positions of all ``mnkMagic`` nodes
+  ## Returns in appearance order the positions of all ``mnkMagic`` nodes
   ## that match `magic`.
   var i = 0
   while i < tree.len:
--- a/compiler/mir/mirpasses.nim
+++ b/compiler/mir/mirpasses.nim
@@ -147,6 +147,15 @@ iterator search(tree: MirTree, kinds: static set[MirNodeKind]): NodePosition =
       yield NodePosition(i)
     inc i
 
+iterator search(tree: MirTree, magic: static TMagic): NodePosition =
+  ## Returns in order appearance the positions of all ``mnkMagic`` nodes
+  ## that match `magic`.
+  var i = 0
+  while i < tree.len:
+    if tree[i].kind == mnkMagic and tree[i].magic == magic:
+      yield NodePosition(i)
+    inc i
+
 iterator uses(tree: MirTree, start, last: NodePosition): OpValue =
   ## Returns in an unspecified order all values used for reads/writes
   ## in the code range ``start..last``. Tags are already skipped.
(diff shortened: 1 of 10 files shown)


zerbina:
Run-time efficiency regressions are very likely going to be common as part of the "lowerings as MIR passes" project. Previously, the code generators could generate specialized target code for magics, which MIR passes cannot do (directly).

The plan is to perform optimizations on the core-language code (`CgNode`) as a whole, instead of achieving efficiency by producing specialized target code for magics. For example, one such optimization could be the removal of unnecessary intermediates. This has the benefits that:
- the lowering passes can focus on correctness
- the optimizations also apply to non-magic code
- optimization passes can be easily shared between code generators

Of course, this approach doesn't mean that the lowering passes should completely disregard the efficiency of the code they produce.

compiler/mir/mirpasses.nim
```diff
@@ -478,6 +487,38 @@ proc fixupCallArguments(tree: MirTree, config: ConfigRef,
...
+        # we're just swapping the values, no full copy is needed
+        argBlock(buf):
+          chain(buf): opParam(0, typ) => tag(ekReassign) => name()
+          chain(buf): opParam(1, typ) => arg()
+        buf.add MirNode(kind: mnkFastAsgn)
+        argBlock(buf):
+          chain(buf): opParam(1, typ) => tag(ekReassign) => name()
+          chain(buf): emit(temp) => arg()
+        buf.add MirNode(kind: mnkFastAsgn)
+    changes.remove() # remove the 'void' sink
```
zerbina, line 520: Having to remove the `mnkVoid` node isn't pretty, and the same thing is going to come up for the lowering passes of other magics. I think it'd make sense to introduce the new `mnkInlined` (or similar) node which acts much like an `mnkRegion`, with the difference that an `mnkInlined` returns a value ('unit' when the replaced operator has a `void` return).

compiler/mir/mirpasses.nim
```diff
@@ -147,6 +147,15 @@ iterator search(tree: MirTree, kinds: static set[MirNodeKind]): NodePosition =
       yield NodePosition(i)
     inc i
 
+iterator search(tree: MirTree, magic: static TMagic): NodePosition =
+  ## Returns in order appearance the positions of all ``mnkMagic`` nodes
```
```suggestion
  ## Returns in appearance order the positions of all ``mnkMagic`` nodes
```

or `order of appearance`, but this way is shorter.

saem (approved):
LGTM.

only one non-blocking suggestion

compiler/mir/mirpasses.nim
zerbina: Oops, yea, `order of appearance` was what I meant to write.

PR #834, opened 2023-08-08
testament: support per-target `knownIssue`s

zerbina:
## Summary

Add support for per-target `knownIssue` specification to testament.

Previously, only the full test (including all targets) could be marked
as a "known issue". To emulate per-target `knownIssue`s, tests had to
explicitly disable the target for which the test was failing and add a
separate comment, but this has the issue of `--tryfailing` not applying.

The `knownIssue` key in the test specifications now supports a
*sub-key*, like so: `knownIssue.target`, where `target` is one of `c`,
`js`, or `vm`. A `knownIssue` key without a sub-key means "applies to
all targets", and specifying multiple targets works by separating them
via spaces.

## Details

At the core of the changes is that `knownIssues` are now per-target
instead of per-spec. The spec parser already supports dots in keys, so
the only addition there is splitting the parsed key string into the main
and sub part prior to processing the key.

Since "has known issue" is now a per-run attribute instead of a per-spec
one, disabling a test early by returning `reKnownIssue` from
`computeEarly` cannot work anymore -- everything related to whole-spec
known issues is removed. Instead, test-run descriptions are now always
generated, but then later skipped by `run` (which now accepts a
`runKnownIssues` parameter) if they have known issues and trying those
is disabled.

Since skipped "known issue" test runs would have no associated "known
issue" information in the backend otherwise, `addResult` for `TestRun`
now always includes the runs "known issues" information.

Tests using the disable + comment approach are adjusted to use per-
target `knownIssue`s, but without checking whether the issue is still
current.

### DIFF
--- a/doc/testament.rst
+++ b/doc/testament.rst
@@ -127,7 +127,16 @@ Test execution options
 
 - ``knownIssue`` description of the test that currently fails but should
   execute successfully; it is a known bug that must be fixed in the future.
-  Can be used several times in specification.
+  Can be used several times in specification. One can also specify the
+  affected targets.
+
+  .. code-block:: nim
+
+    knownIssue.vm: "..."   # the test only fails with the vm target
+    knownIssue.c js: "..." # the fails with both the c and js target
+
+  A standalone ``knownIssue`` key means that all selected targets are
+  affected by the issue(s).
 
 Compiler output assertions
 --------------------------
--- a/doc/testament.rst
+++ b/doc/testament.rst
@@ -133,7 +133,7 @@ Test execution options
   .. code-block:: nim
 
     knownIssue.vm: "..."   # the test only fails with the vm target
-    knownIssue.c js: "..." # the fails with both the c and js target
+    knownIssue.c js: "..." # fails with both the c and js target
 
   A standalone ``knownIssue`` key means that all selected targets are
   affected by the issue(s).
--- a/testament/specs.nim
+++ b/testament/specs.nim
@@ -121,7 +121,7 @@ type
     timeout*: float ## in seconds, fractions possible, but don't rely on
     ## much precision
     inlineErrors*: seq[InlineError] ## line information to error message
-    knownIssues*: seq[string] ## known issues to be fixed
+    knownIssues*: array[TTarget, seq[string]] ## known issues to be fixed
     labels*: seq[string] ## user-added metadata
 
 proc getCmd*(s: TSpec): string =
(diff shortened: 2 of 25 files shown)


zerbina:
The PR is ready to reviewed, I've only created it as draft in order to not put more pressure on the CI.

testament/specs.nim
```diff
@@ -289,6 +289,16 @@ proc parseSpecifiedTargets*(value: string): set[SpecifiedTarget] =
       else: raise newException(ValueError,
                                "invalid target specificatoin: '$#'" % v)
 
+proc splitKey*(key: string): tuple[key, sub: string] =
```
saem, line 292: did you already consider using section start/end, IIRC this is all based on `parsecfg` so that should be available to us?

```
[knownIssues]
target.js: "lorem ipsum ..."
```

Having said that, this might be a better evolution once we support more kinds of known issues.

saem (approved):
the changed spec definition does a good job of cleaning up those tests.

---

thinking about things broadly and how things might need to work in the future:

target based disabling is the biggie, and will likely remain so, the other one is platform based (i.e. got something working in posix environments, but not windows, etc). the latter is being handled by `disabled` + comments, so one day we'll need to support platform based known issues, but for now it's fine and I suspect that we will want the run based approach in that scenario as well.

testament/specs.nim
zerbina: I did consider sections, both for `[target]` and, like you mentioned, `[knownIssues]`, but ultimately decided against using them here (for now), as that would require larger changes to the spec parsing (e.g., the enclosing section needs to be tracked, keys not allowed for the active sections detected/rejected, etc.).

Once specification of known issues becomes more elaborate (platform, matrix entry, etc.), I agree that sections are likely the better approach.

zerbina:
I've updated the testament documentation with a mention `knownIssue`'s new capability.

doc/testament.rst
```diff
@@ -127,7 +127,16 @@ Test execution options
...
 - ``knownIssue`` description of the test that currently fails but should
   execute successfully; it is a known bug that must be fixed in the future.
-  Can be used several times in specification.
+  Can be used several times in specification. One can also specify the
+  affected targets.
+
+  .. code-block:: nim
+
+    knownIssue.vm: "..."   # the test only fails with the vm target
+    knownIssue.c js: "..." # the fails with both the c and js target
```
saem:
```suggestion
    knownIssue.c js: "..." # fails with both the c and js target
```

saem (approved):
soft non-blocking typo

zerbina:
/merge

The CI failure was due to what looks like a network error while downloading one of the action dependencies.

PR #835, opened 2023-08-09
internal: use dedicated context type for `vmbackend`

zerbina:
## Summary

Add and use a dedicated type for the contextual state of the VM code-
generation orchestrator. This removes the hard coupling between
`vmbackend` and `vmdef.TCtx` and is another preparation for splitting up
the latter.

## Details

The new `GenCtx` type replaces `TCtx` as the contextual-state type
for the orchestrator. For now, `GenCtx` still stores an instance of
`TCtx`, as the latter currently contains the code generator state.

In preparation for splitting up `TCtx`, usage of its `idgen` and
`functions` fields are removed from `vmbackend.nim`. The ID generator is
directly passed to procedures that need one, and the function table
(`functions`) is now stored in `GenCtx` -- prior to serializing the VM
environment, the function table stored with `GenCtx`'s is moved into the
`TCtx` instance.

With the new dedicated type for the orchestrator, the collected globals
don't have to be stored in `TCtx` anymore. The `collectedGlobals` field
is made part of `GenCtx` and is renamed to just `globals`.

### Misc

- add the `base` template for `OrdinalSeq` and use it to reduce the
  noise in the `OrdinalSeq` routines
- fix the `add` template for `OrdinalSeq`

### DIFF
--- a/compiler/utils/containers.nim
+++ b/compiler/utils/containers.nim
@@ -152,9 +152,9 @@ template `[]`*[I; T](x: OrdinalSeq[I, T], i: I): untyped =
 template `[]=`*[I; T](x: OrdinalSeq[I, T], i: I, item: T): untyped =
   (seq[T])(x)[ord i] = item
 
-func add*[I; T](x: OrdinalSeq[I, T], item: sink T): I {.inline.} =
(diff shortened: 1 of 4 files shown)


saem (approved):
good start to the split. :D

PR #836, opened 2023-08-09
codegen: optimize single-operand `CgNode`s

zerbina:
## Summary

Add a dedicated `CgNode` branch for expression-like nodes that always
have a single operand. This reduces the amount of allocations and access
indirections, and it also makes the code more self-describing
(`.operand` now vs. the previous `[0]`).

## Details

For now, only expression-like nodes use the `operand` branch, but some
statement nodes (for example, `mnkRepeat`) could also be considered for
using the `operand` branch in the future. The set previously denoting
all atoms (`cnkWithoutItem`) is renamed to `cnkAtoms` and the usage
sites are adjusted to take the new meaning into account: only traversing
`mnkWithItem` nodes is not enough anymore.

The `cnkCast`, `cnkConv`, and `cnkHiddenConv` nodes also use the
`operand` branch. Them using two subnodes was inherited from `PNode`,
but it's not something that's needed for `CgNode` IR.

`cgirgen` is changed to create `operand`-nodes where required, and all
`[0]` and `[1]` (for `cnkCast|cnkConv|cnkHiddenConv`) access for the
changed nodes is replaced with `.operand`.

### Allocation Stats

The accumulated allocation stats (using `-d:nimAllocStats`) of
`generateIR` when compiling the compiler at https://github.com/nim-works/nimskull/commit/48a07fbfc214491754761d09b178b71137cc1dd5 are:

|version|allocs|deallocs|
|:------|:-----|:-------|
|this commit|3162007|752474|
|48a07fbfc2|3320434|758663|

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -485,20 +485,22 @@ proc unaryArith(p: BProc, e: CgNode, d: var TLoc, op: TMagic) =
     assert false, $op
 
 proc genDeref(p: BProc, e: CgNode, d: var TLoc) =
-  let mt = mapType(p.config, e[0].typ, mapTypeChooser(e[0]))
+  let
+    src = e.operand
+    mt = mapType(p.config, src.typ, mapTypeChooser(src))
   if mt in {ctArray, ctPtrToArray} and lfEnforceDeref notin d.flags:
     # XXX the amount of hacks for C's arrays is incredible, maybe we should
     # simply wrap them in a struct? --> Losing auto vectorization then?
-    expr(p, e[0], d)
-    if e[0].typ.skipTypes(abstractInst).kind == tyRef:
+    expr(p, src, d)
+    if src.typ.skipTypes(abstractInst).kind == tyRef:
       d.storage = OnHeap
   else:
     var a: TLoc
-    var typ = e[0].typ
+    var typ = src.typ
     if typ.kind in {tyUserTypeClass, tyUserTypeClassInst} and typ.isResolvedUserTypeClass:
       typ = typ.lastSon
     typ = typ.skipTypes(abstractInst)
-    initLocExprSingleUse(p, e[0], a)
+    initLocExprSingleUse(p, src, a)
     if d.k == locNone:
       # dest = *a;  <-- We do not know that 'dest' is on the heap!
       # It is completely wrong to set 'd.storage' here, unless it's not yet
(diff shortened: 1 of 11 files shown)


compiler/backend/ccgexprs.nim
```diff
@@ -1534,12 +1536,16 @@ proc genOrd(p: BProc, e: CgNode, d: var TLoc) =
 proc genSomeCast(p: BProc, e: CgNode, d: var TLoc) =
   const
     ValueTypes = {tyTuple, tyObject, tyArray, tyOpenArray, tyVarargs, tyUncheckedArray}
+
+  let src =
+    if e.kind == cnkCall: e[1]
+    else:                 e.operand
```
zerbina: `genSomeCast` is also used to generate the code for `mChr`, which is why this dispatching is needed here.

saem: perhaps overly defensive, but perhaps this should be exhaustive over kind?

compiler/vm/vmgen.nim
```diff
@@ -205,8 +205,11 @@ func underlyingLoc(n: CgNode): PSym =
   var root {.cursor.} = n
   # skip nodes that don't change the location until we arrive at either one
   # that does, or a symbol
-  while root.kind in {cnkConv, cnkStmtListExpr}:
-    root = root.lastSon
+  while true:
```
saem, line 208:
```suggestion
  while root.kind in {cnkConv, cnkStmtListExpr}:
```
at first i thought this might be a good idea, but it ends up meaning two places have to be updated correctly in order for things to work. thinking about it a bit further and i realize that i'm suggesting this to avoid leaving infinite loops around.

separate from this PR, do you think we should have an idiom, like so?

```nim
template loopForever(body: untyped) =
  var i = 0
  const maxIterations = 1000
  while i < maxIterations:
    body
    inc i
  doAssert i < maxIterations, "fixed point calculation exceeded maximum iteration attempts"
```

saem (approved):
the clear name over 'idiomatic' traversal is much more pleasant to read, with storage benefits to boot. :D

compiler/vm/vmgen.nim
zerbina: Yea, this did bother me a bit too. A `while true` is always a bit meh, though I think that, just like you said, not duplicating the stats is the better choice here.

Regarding the idiom, I'm a bit unsure. To me, it feels like a work-around to missing static analysis -- essentially, the compiler could (and I think should) detect `while` loops where a control-flow path exists on which no progress is made (i.e., no state is modified that is used in conditions guarding loop exits).

Adding the idiom likely still makes sense, but figuring out a `maxIterations` value that works well across different use cases could be hard.

saem: Thanks for the input. Static analysis + some wise API design choices is definitely the way to go. In the meantime, we can make the `maxIteration` a parameter (tacked on the end so it doesn't always have to be specified).

PR #837, opened 2023-08-10
remove support for enum inheritance

zerbina:
## Summary

Enum types supported an internal base type that could only be specified
in macro-generated enum types (there's no syntax for representing it),
but there aren't any tests for it, the features seems to have never
been documented anywhere, and enums that inherit from other enums are
not consistently considered in the compiler. The feature is thus
removed.

## Details

- remove support for `nkEnumTy` AST having a non-empty first slot. This
  now results in an `illformed AST` error
- remove support for enum inheritance from `tryReadingTypeField`
- remove the `rsemInheritanceOnlyWorksWithAnEnum` report

### DIFF
--- a/compiler/ast/report_enums.nim
+++ b/compiler/ast/report_enums.nim
@@ -371,9 +371,6 @@ type
     rsemCannotInstantiateWithParameter
     rsemCannotGenerateGenericDestructor
     rsemUndeclaredField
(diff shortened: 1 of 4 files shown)


saem (approved):
Very cool. Eventually with a bit of work we'll be able to remove the unnecessary hole in the ast. 🎉

PR #838, opened 2023-08-11
vm: pass data for compiler interfacing in `VmArgs`

zerbina:
## Summary

Provide the data for interfacing with the compiler in the `VmArgs`
object, which allows for making the callbacks registered by `vmops`
independent of the `TCtx` instance they're first registered with.

## Details

Closing over fields of the registered-with-`TCtx` is replaced with
accessing the corresponding `VmArgs` fields.

While the `ConfigRef` instance stored with `TCtx` is always the same
as the one stored in `TCtx`'s `graph`, a `ModuleGraph` is not always
present (for example, when the VM runs in standalone mode), meaning
that the `ConfigRef` instance has to also be passed in `VmArgs`.

In addition to making the callbacks context-independent, this change 
also means that the callbacks in `vmops` aren't (first-class) closures
anymore. To prevent them from accidentally capturing something again,
all manual callback procedures in `vmops` are assigned the `.nimcall`
calling convention.

### DIFF
--- a/compiler/vm/vmops.nim
+++ b/compiler/vm/vmops.nim
@@ -69,7 +69,6 @@ from system/formatfloat import writeFloatToBufferSprintf
 
 from compiler/modules/modulegraphs import `$`
 
-
 func emptyCallback(a: VmArgs) =
   discard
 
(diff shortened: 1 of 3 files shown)


compiler/vm/vmops.nim
```diff
@@ -223,19 +222,11 @@ proc getEffectList(cache: IdentCache, idgen: IdGenerator; a: VmArgs;
...
-#      environment related ones could be passed to the callbacks instead,
-#      making most of the capturing unnecessary
-
 proc registerBasicOps*(c: var TCtx) =
   ## Basic system operations as well as callbacks for some stdlib functions
   ## that don't interact with the host environement, but use language features
   ## that the VM doesn't directly support (such as 'importc'-ed functions)
 
-  # captured vars:
-  let isJs = c.config.backend == backendJs
```
saem, line 237: With the removal of captures should the calling convention be explicit in order to avoid future captures?

saem (approved):
One nonblocking tidbit for consideration, but otherwise good to go. I like how these modules are genuinely become more independent of each other.

compiler/vm/vmops.nim
zerbina: Ah, good call, I'll adjust the definitions.

zerbina: I've changed all callback procedures to use the `.nimcall` calling convention.

PR #839, opened 2023-08-11
sem: decide storage type for enums on definition

zerbina:
## Summary

Centralize the decision about the underlying storage types for enum
values in semantic analysis. Since the decision now happens in a single
place instead of being scattered across the compiler, this makes it
significantly easier to change the enum storage behaviour later on.

In addition, this fixes two bugs with 1- and 2-byte enum values (which
use unsigned integer storage) where the most significant bit was set
that affected code running in the VM:
- loading the values from memory resulted in non-valid values
- (only affected the VM and JS backend) storing them inside aggregate
  `const`s resulted in them having invalid values at run-time

## Details

The key changes are:
- `tyEnum` now stores the storage type in the second type slot, meaning
  that `skipTypes(tyEnum)` yields the storage type
- size and alignment computation redirects to the storage type
- `isUnsigned` considers a `tyEnum`'s storage type
- the code generators use the `tyEnum`'s provided storage type and don't
  decide it themselves

Using the second type slot of `tyEnum` is done for both forward- and
backward-compatibility: it's possible that some of the current code
still depends on an enum type's first slot always being `nil`, and in
the future the slot could possibly be used to store the enum's base
type.

The storage type is directly set on enum definition (`semEnum`). Before
analyzing the enum fields, a preliminary storage type (`tyInt`) is
already added for the enum: this allows for the enum's values to be used
*before* the enum type is fully produced. Once the full value range is
known, the correct storage type is selected.

Choosing the storage type happens in the same way as it previously did
in the C code generator (`ccgtypes`). Apart from the information now
being available during semantic analysis already, this only affects the
VM target, where non-negative enum values requiring 32 or 64 bit of
storage were previously stored as unsigned integer (they're now stored
as *signed* integers, in line with the C target).

For the VM code generator, `tyEnum` is added to the `IrrelevantTypes`
set, meaning that the code generator now transparently treats all enum
types as their storage type, with an exception being made for the to-
string operation.

### DIFF
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -41,6 +41,10 @@ proc semEnum(c: PContext, n: PNode, prev: PType): PType =
     localReport(c.config, n[0], reportAst(rsemIllformedAst, n[0]))
 
   rawAddSon(result, nil) # base type; always nil
+  # add a preliminary storage type such that the enum can already be
+  # used in its own definition
+  rawAddSon(result, getSysType(c.graph, n.info, tyInt))
+
   let isPure = result.sym != nil and sfPure in result.sym.flags
   var symbols: TStrTable
   if isPure: initStrTable(symbols)
@@ -143,6 +147,33 @@ proc semEnum(c: PContext, n: PNode, prev: PType): PType =
       wrongRedefinition(c, e.info, e, conflict)
 
     inc(counter)
+
+  # now that we know the full value range, select the correct storage type
+  # of the enum:
+  let tk =
+    if firstOrd(c.config, result) < Zero:
+      tyInt32 # use signed int32
+    elif result.size != szUncomputedSize:
+      # use the manually specified size for selecting the storage type
+      let s = result.size
+      if   s <= 1: tyUInt8
+      elif s <= 2: tyUInt16
+      elif s <= 4: tyInt32
+      elif s <= 8: tyInt64
+      else:        unreachable()
+    else:
+      let lastOrd = lastOrd(c.config, result)
+      if lastOrd < (1 shl 8):
+        tyUInt8
+      elif lastOrd < (1 shl 16):
+        tyUInt16
+      elif lastOrd < (BiggestInt(1) shl 32):
+        tyInt32
+      else:
+        tyInt64
+
+  result[1] = getSysType(c.graph, n.info, tk)
+
   if isPure and sfExported in result.sym.flags:
     addPureEnum(c, LazySym(sym: result.sym))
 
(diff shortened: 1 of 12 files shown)


saem:
I started asking myself why the "inconsistency" between unsigned for 8 and 16 bits, but then this paragraph alluded to it, leaving this message here as a remark for those that might need additional clarification.

> This also fixes the internal
> inconsistency of vmtypegen using uint32 and uint64 for 32- and 64-
> bit enum values (as opposed to int32 and int64 used by ccgtypes).
> In effect, this means that a pointer-to-32-or-64-bit-enum-value cannot
> be reinterpreted as a pointer-to-unsigned-int when using the VM
> backend anymore.

---

side note: I really like the fact that this brings us that much closer to having first class enum values. 🚀

compiler/sem/semtypes.nim
```diff
@@ -143,6 +147,33 @@ proc semEnum(c: PContext, n: PNode, prev: PType): PType =
       wrongRedefinition(c, e.info, e, conflict)
 
     inc(counter)
+
+  # now that we know the full value range, select the correct storage type
+  # of the enum:
+  let tk =
+    if firstOrd(c.config, result) < Zero:
+      tyInt32 # use signed int32
```
saem, line 155: shouldn't this do an internal check like the one below for `lastOrd`?

```nim
      if firstOrd(c.config, result) < (BiggestInt(1) shr 32):
        tyInt64
      else:
        tyInt32
```

saem: an alternative strategy with perhaps simpler logic might be to check computed vs non-computed size, then for computed ones, determine the base kind independently based on first and last ordinal value, then take the biggest one of the two determined sizes?

saem (approved):
Looks good and consolidating this has a number of nice benefits as you found.

I left one comment for consideration wrt determining the storage size, but I might be mistaken in my reading of things.

---

Remarks about tests (no change is necessarily required for this PR):

I did a quick scan of the enum tests and the `sizeof` related tests seem to have a number of gaps. I couldn't do a full inventory, but considering how spotty/messy the language spec are alone, those should get covered in a separate endeavour. From what I can tell the non-zero first ordinal and various boundary conditions have significant gaps.

compiler/sem/semtypes.nim
zerbina: > shouldn't this do an internal check like the one below for lastOrd?

Yep, to be a able to properly store enums that require signed values and exceed the 32-bit int range, that'd be necessary. At the moment, these simply misbehave at run-time.

zerbina:
> I started asking myself why the "inconsistency" between unsigned for 8 and 16 bits, but then this paragraph alluded to it, leaving this message here as a remark for those that might need additional clarification.

Apologies, the previous PR message (I've adjusted it) didn't make it clear that this PR only promotes the behaviour with the C backend to the official/correct one, but doesn't change it.

As to why 8 and 16 bit values use unsigned integers while the larger ones do not, I don't know. I'm not aware of any discussions related to this, and the behaviour seems to have been this way from the very first (public) commit.

compiler/sem/semtypes.nim
saem: Ah, understood.

PR #840, opened 2023-08-13
treat `gorge`/`staticExec` as a dangerous operation

zerbina:
## Summary

Treat both `gorge` procedure and its extended version as dangerous
operations, meaning that `--experimental:vmopsDanger` has to be
explicitly enabled for them to not be treated as no-ops.

This fixes an inconsistency: all other compile-time-available
operations that are able to modify the host's environment were
disabled without said option, but `gorge` (which can indirectly
modify the host's environment) was not.

## Details

To be able to treat `gorge` as a dangerous operation, usage of it in
the context of NimScript had to be removed, as otherwise `exec` would
be a no-op when dangerous VM operations are disabled. The extended
`exec` procedure now calls the `rawExecEx` callback (which uses the
gorge implementation internally), and a call to `gorge` in
`distros.nim` is replaced with a call to `exec`.

The `opcGorge` implementation now considers the `vmopsDanger` feature,
and the `gorgeEx` callback is included in the callbacks that are turned
into no-ops when `vmopsDanger` is not enabled. For tests that use
`gorge` or `gorgeEx`, dangerous compile-time operations are enabled.

### DIFF
--- a/compiler/front/scripting.nim
+++ b/compiler/front/scripting.nim
@@ -30,6 +30,7 @@ import
   ],
   compiler/vm/[
     compilerbridge,
+    gorgeimpl,
     vmconv,
     vmdef,
     vmhooks,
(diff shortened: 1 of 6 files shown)


saem (approved):
Yay, more consistent application of principles

PR #841, opened 2023-08-14
vmops: decouple registration from discovery

zerbina:
## Summary

Separate *discovering* overrides from *registering* them by turning the
`register` procedures into iterators that don't require a `TCtx`. This
makes the interface more flexibile, and is a preparation for moving the
the `callbackKeys` list out of `TCtx`.

## Details

As part of the changes, the used terminology is improved: a *callback*
is what the VM invokes in the end, while an *override* is a pattern +
callback to redirect to. The places where the `register` procedures
were previously used are updated to iterate the contents of the new
iterators and register the callbacks/overrides as needed.

The `registerX` procedures are turned into iterators, and the calls to
`registerCallback` in `vmops` are replaced with calls to the `override`
template, which expands to a `yield` statement. Iterators are used
instead because they're slightly more flexible, but, at least with the
current usage, constant `seq`s would have worked as well.

Although not yet clear from data placement, "overrides" are not
something that the VM implements -- rather, it's a feature of the
JIT and linker. Therefore, the `registerCallback` is moved from the
`vmdef` into the `vmjit` module, and updated to also verify that the
callbacks and patterns list stay synchronized. The
`registerAdditionalOps`, which has the intended use of registering
the additional ops for compile-time code execution is moved to
`compilerbridge`.

As the result, `vmops` is now only concerned with providing overrides,
but not with how/if they're registered.

### DIFF
--- a/compiler/vm/vmjit.nim
+++ b/compiler/vm/vmjit.nim
@@ -346,6 +346,6 @@ proc registerCallback*(c: var TCtx; pattern: string; callback: VmCallback) =
   ## `pattern` is added to the VM's function table, all invocations of the
   ## procedure at run-time will invoke the callback instead.
   # XXX: consider renaming this procedure to ``registerOverride``
-  c.callbacks.add(callback)
+  c.callbacks.add(callback) # some consumers rely on preserving registration order
   c.callbackKeys.add(IdentPattern(pattern))
   assert c.callbacks.len == c.callbackKeys.len
--- a/compiler/vm/vmjit.nim
+++ b/compiler/vm/vmjit.nim
@@ -343,8 +343,8 @@ proc loadProc*(jit: var JitState, c: var TCtx, sym: PSym): VmGenResult =
 proc registerCallback*(c: var TCtx; pattern: string; callback: VmCallback) =
   ## Registers the `callback` with `c`. After the ``registerCallback`` call,
   ## when a procedures of which the fully qualified identifier matches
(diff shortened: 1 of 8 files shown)


saem:
The last sentence in the first paragraph of the 'details' section is cut off part way through.

zerbina:
Oh, thanks for the hint. I had forgot finishing the sentence.

saem:
Made some minor PR message body updates for grammar.

compiler/vm/vmjit.nim
```diff
@@ -338,4 +338,14 @@ proc loadProc*(jit: var JitState, c: var TCtx, sym: PSym): VmGenResult =
...
\ No newline at end of file
+    compile(jit, c, idx)
+
+proc registerCallback*(c: var TCtx; pattern: string; callback: VmCallback) =
+  ## Registers the `callback` with `c`. After the ``registerCallback`` call,
+  ## when a procedures of which the fully qualified identifier matches
+  ## `pattern` is added to the VM's function table, all invocations of the
+  ## procedure at run-time will invoke the callback instead.
+  # XXX: consider renaming this procedure to ``registerOverride``
+  c.callbacks.add(callback)
```
```suggestion
  c.callbacks.add(callback) # some consumers rely on preserving registration order
```

saem (approved):
one minor non-blocking suggestion, but the code is cleaner wrt responsibility separation and style (i.e. `register: foo`)

PR #842, opened 2023-08-14
codegen: remove the `cnkChckRange` node kinds

### LINKED ISSUE #830: `sem`-driven run-time checks
Make emission of run-time checks driven by `sem` instead of by the code generators and treat them as an extensions to the compile-time checks.

## The Problem

How run-time checks (overflow checks, bound checks, field checks, range checks, object-conversion checks, float checks, and, in the future, possibly nil checks) work and when they're actually generated is currently fully the responsibility of the code generators. There are multiple issues with this:
1. the checks decide program behaviour -- *what* and *where* thus need to be decisions of `sem` 
2. the relevant logic is duplicated for each code generator
3. their presence is unknown during the early MIR phase, which means that the exceptional control-flow arising from failed run-time checks (when `--panics:off`) cannot be considered during by the data-flow analysis
4. information about where defects are raised is not present during semantic analysis (`sempass2`)
5. options enabled/disabled via `push` and `pop` don't affect top-level or procedure-level code, since both pragmas don't reach the code generators 

With the current architecture/design, compile- and run-time checks are completely separate. A much better approach would be to treat run-time checks as a fallback for where the compiler cannot statically prove that something is safe. For example, field checks (i.e. ensuring that a field is part of an active variant branch) are not necessary in the following snippet:
```nim
case x.kind
of enumA:
  discard x.a + x.b
  # assuming that both `a` and `b` are part of a branch that is
  # discriminated by `enumA`, no checks are necessary here
```

In the following, only the first access of `y.a` needs a check:
```nim
echo y.a + y.a
```

These optimizations could greatly benefit code with enabled run-time checks.

## How to get there

- [ ] ensure that code for all checks is present for AST reaching `mirgen`. Range (`nkChckRange`) and field (`nkCheckedFieldExpr`) already have AST representation, overflow, bound, object conversion, and float checks still need some way to represent them, either via new magics or new node kinds
- [x] remove all decision making regarding run-time checks from the code generators
- [ ] move the insertion of run-time checks into either `sempass2` or a separate pass running earlier. This is so that they can be taken into account by defect tracking
- [ ] only insert checks where the success of an operation cannot be proven statically

Once the checks are all present in AST reaching `sempass2`, (internal) defect tracking can be implemented, which would make it possible to remove the heuristics regarding exceptions/defects (`canRaiseConservative`, `canRaise`, etc.) from the code generators. This is a step towards removing the distinction between exception and defects from the code generators.

zerbina:
## Summary

Replace the `cnkChckRange` node kinds with a magic call that is already
inserted during MIR synthesis (`mirgen`) and remove all decision making
regarding range-checks from the code generators.

This is a step towards moving decision making regarding run-time checks
fully into late semantic analysis. In addition, this fixes destructors
not being run for scopes where the only escaping unstructured control-
flow is due to range checks.

## Details

A new internal magic is added for representing conversions-with-range-
checks during the MIR and code generation phase: `mChckRange`. `mirgen`
translates `nkChckRange`, `nkChckRange64`, and `nkChckRangeF` to the
new magic, but only if its not a to-unsigned conversion and range checks
are enabled for the current owner.

For now, these guards mirror those previously used in the code
generators, but the direction is to move to a correct-by-construction
approach and only inject `nkChckRange` nodes where range checks should
be performed.

The `mChckRange` magic is included in the `magicsThatCanRaise` set, and
the duplicate set of the same name is removed from `mirexec`. As a
consequence, the exceptional control-flow arising from failed range
checks is now picked up on by MIR-based data- and control-flow analysis.
However, since the set is constant, this also means that range-check
calls will be treated as raising even when panics are enabled. This is a
known problem with magics that needs to be solved at one point.

Multiple things are now obsolete and thus removed:
- the `cnkChckRange` node kinds
- the `cgirgen` logic for producing `cnkChckRange` nodes (which was
  originally duplicated from the logic in `transf.transformConv`)
- the code generator handling for `cnkChckRange`. Parts of it are
  re-used for implementing `mChckRange` support

Two additional problems are fixed as a side-effect of the changes:
- the VM code generator now respects the `rangeChecks` option in
  the same way as the other code generator
- disabled range checks (either via the `rangeChecks` option or
  because the target type is unsigned) are not treated as no-op
  by the JS code generator, but instead use unchecked conversions
  (the `tdollars.nim` test is adjusted accordingly), in line with
  the other code generators

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -609,7 +609,8 @@ proc isSinkType*(t: PType): bool {.inline.} =
   t.kind == tySink
 
 const magicsThatCanRaise* = {
-  mNone, mSlurp, mStaticExec, mParseExprToAst, mParseStmtToAst, mEcho}
+  mNone, mSlurp, mStaticExec, mParseExprToAst, mParseStmtToAst, mEcho,
+  mChckRange }
 
 proc canRaiseConservative*(fn: PNode): bool =
   if fn.kind == nkSym and fn.sym.magic notin magicsThatCanRaise:
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -814,6 +814,9 @@ type
 
     # magics only used internally:
     mAsgnDynlibVar
+    mChckRange
+      ## chckRange(v, lower, upper); conversion + range check -- returns
+      ## either the type-converted value or raises a defect
 
 # things that we can evaluate safely at compile time, even if not asked for it:
 const
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -1414,8 +1414,8 @@ proc genInOp(p: BProc, e: CgNode, d: var TLoc) =
     # do not emit the set, but generate a bunch of comparisons; and if we do
     # so, we skip the unnecessary range check: This is a semantical extension
     # that code now relies on. :-/ XXX
-    let ea = if e[2].kind in {cnkChckRange, cnkChckRange64}:
-               e[2][0]
+    let ea = if getMagic(e[2]) == mChckRange:
+               e[2][1]
              else:
                e[2]
     initLocExpr(p, ea, a)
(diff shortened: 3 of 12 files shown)


saem (approved):
the greater regularity in handling of exceptional control is particularly cool.

PR #843, opened 2023-08-16
internal: split `vmgen`/linker state from `TCtx`

zerbina:
## Summary

Move much of the data and associated types for the VM code generator and
linker into `vmgen` and the new `vmlinker` module. Having the code-
generation-related state in the `vmgen` module, means that the VM code-
generator state can now depend on modules that should not be a
transitive dependency (through `vmdef`) of all other VM-related modules.

## Details

The linker-related types `IdentPattern` and `LinkIndex` plus the
routines for working with them are moved to the `vmlinker` module -- the
linker-related state is moved into the new `LinkerData` type. Due
`vmcompilerserdes` currently requiring access to the link table, the
`LinkerData` instance has to be still stored with `TCtx` instead of
the more correct `JitState`.

For the VM code generator, all relevant types, except for the
diagnostics, are moved from `vmdef` into `vmgen`. The `TCtx` used by
`vmgen` is now an alias to the new `CodeGenCtx` type, which stores all
the input, output, and other contextual state required for code
generation.

When using `vmjit`, much of the data read or written to by the code
generator is part of the VM execution environment (`TCtx`), and in order
to not having to either use pointers or pass all of it along as
procedure parameters, the relevant data is swapped between `TCtx` and
`CodeGenCtx` both *before* and *after* invoking the code generator.
`vmjit.genExpr` is slightly adjusted for this to work uniformly across
the `genX` procedures: it now also emits an `Eof` instruction.

`vmbackend` now stores an instance of `CodeGenCtx` instead of `TCtx`.
Once code generation is finished, a temporary VM execution environment
is set-up, populated, and then serialized.

Finally, `vmrunner` is updated to not rely on (now non-existent)
`TCtx.callbackKeys` field. The list of expected override patterns is
instead taken from the packed context and passed to `registerCallbacks`
directly.

### DIFF
--- a/compiler/vm/vmjit.nim
+++ b/compiler/vm/vmjit.nim
@@ -81,7 +81,7 @@ func swapState(c: var TCtx, gen: var CodeGenCtx) =
   template swap(field: untyped) =
     swap(c.field, gen.field)
 
-  # inpute parameters:
+  # input parameters:
   swap(graph)
   swap(config)
   swap(mode)
--- a/compiler/vm/vmjit.nim
+++ b/compiler/vm/vmjit.nim
@@ -41,6 +41,7 @@ import
     vmcompilerserdes,
     vmdef,
     vmgen,
+    vmlinker,
     vmmemory,
     vmtypegen
   ],
@@ -152,22 +153,22 @@ func discoverGlobalsAndRewrite(data: var DiscoveryData, tree: var MirTree,
   # can rewrite all defs in one go:
   rewriteGlobalDefs(tree, source, outermost = false)
 
-func register(c: var TCtx, data: DiscoveryData) =
+func register(linker: var LinkerData, data: DiscoveryData) =
   ## Registers the newly discovered entities in the link table, but doesn't
   ## commit to them yet.
   for i, it in peek(data.procedures):
-    c.symToIndexTbl[it.id] = LinkIndex(i)
+    linker.symToIndexTbl[it.id] = LinkIndex(i)
 
   for i, it in peek(data.constants):
-    c.symToIndexTbl[it.id] = LinkIndex(i)
+    linker.symToIndexTbl[it.id] = LinkIndex(i)
 
   # first register globals, then threadvars. This order must be the same as
   # the one they're later committed to in
   for i, it in peek(data.globals):
-    c.symToIndexTbl[it.id] = LinkIndex(i)
+    linker.symToIndexTbl[it.id] = LinkIndex(i)
 
   for i, it in peek(data.threadvars):
-    c.symToIndexTbl[it.id] = LinkIndex(i)
+    linker.symToIndexTbl[it.id] = LinkIndex(i)
 
 proc generateMirCode(c: var TCtx, n: PNode;
                      isStmt = false): (MirTree, SourceMap) =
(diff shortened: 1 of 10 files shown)


compiler/vm/vmjit.nim
```diff
@@ -71,6 +74,28 @@ func selectOptions(c: TCtx): set[GenOption] =
...
     result.incl goIsCompileTime
 
+func swapState(c: var TCtx, gen: var CodeGenCtx) =
+  ## Swaps the values of the fields shared between ``TCtx`` and ``CodeGenCtx``.
+  ## This achieves reasonably fast mutable-borrow-like semantics without
+  ## resorting to pointers.
+  template swap(field: untyped) =
+    swap(c.field, gen.field)
+
+  # inpute parameters:
```
```suggestion
  # input parameters:
```

compiler/vm/vmlinker.nim
```diff
@@ -0,0 +1,57 @@
...
+  IdentPattern* = distinct string
+    ## A matcher pattern for a fully qualified symbol identifier
+
+  LinkerData* = object
+    ## All data required for mapping symbols to the entities they represent
+    ## inside the VM (procedures, globals, constants).
+    symToIndexTbl*: Table[int, LinkIndex]
+      ## the link table. Keeps track of all symbols known to the linker
+    callbackKeys*: seq[IdentPattern]
+      ## the matcher patterns used for implementing overrides
```
saem, line 31: an invariant we need is that all `callbackKeys` entries, if linked, must have an entry at the corresponding `int` in `symToIndexTbl`? or perhaps, if a key exists in `symToIndexTbl` that overlaps with the index of `callbackKeys`, that must be the `LinkIndex` for that `IdentPattern`?

If that's the case, I think it might be worth noting this invariant, as it indicates not only the cardinality/relationship, but clearly hints as to how data should be organized with the overall type.

saem (approved):
Looks good, I like how `vmdef` is becoming more focused and the modules become more independent of each other.

I did find a comment typo, but it's non-blocking. The suggestion for documenting the invariant is also non-blocking, just occurred to me that the trick with more DOD style approaches is signalling cardinality/relationships of the data contained therein -- I might be mistaken.

compiler/vm/vmlinker.nim
zerbina: > an invariant we need is that all `callbackKeys` entries, if linked, must have an entry at the corresponding `int` in `symToIndexTbl`?

Not directly, no. There exists no relation between `callbackKeys` and `LinkIndex` (and thus `symToIndexTbl`).

Once the necessary progress has been made in the code generators, `symToIndexTbl` is going be replaced with three `OrdinalSeq`s (one for each relevant symbol kind), where each uses a key type that corresponds to the ID of a `Store` stored elsewhere. That way, the relationship is going to be more clear.

compiler/vm/vmjit.nim
zerbina: Heh, I missed that one, thanks.

compiler/vm/vmlinker.nim
saem: apologies, had a dyslexic moment, I thought I read the the second `lookup` procedure delegating to the first but that's clearly not the case.

PR #844, opened 2023-08-17
make `std/marshal` compile with `--styleCheck:error`

iacore:
Change names of types in `std/marshal` that start with a lower-case
letter to start with an upper-case letter, making the module compatible
with the `--styleCheck:error` mode.

### DIFF
--- a/lib/core/typeinfo.nim
+++ b/lib/core/typeinfo.nim
@@ -86,8 +86,8 @@ type
     else:
       rawTypePtr: pointer
 
-  ppointer = ptr pointer
-  pbyteArray = ptr array[0xffff, int8]
+  PPointer = ptr pointer
+  PbyteArray = ptr array[0xffff, int8]
 
 include system/seqs_v2_reimpl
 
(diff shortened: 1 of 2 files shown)


zerbina (changes requested):
Thank you for the PR. I've updated the title to better reflect what is changed.

Please also update the PR message regarding what works now (the `marshal` module could already be compiled previously, it only failed when `--styleCheck:error` is enabled) and ensure that it adheres to the rules mentioned in the template.

For making sure that it really does work now, please also change `tests/stylecheck/tstyle_imports.nim` such that it imports the `marshal` module -- you only need to uncomment line 56.

zerbina:
> @zerbina says that `std/marshal` module could already be compiled previously, it only failed when `--styleCheck:error` is enabled. I think this is the default behavior now, since it doesn't compile by default.

`--styleCheck:error` is not the default. See, for example, the `stdlib/formats/tmarshal.nim` test, which imports the `marshal` module without configuring style checks.

The reason importing the module caused problems for you is that you tried to import the module in the compiler, where standard style *is* enforced (see `compiler/nim.cfg`).

zerbina:
Apart from the PR message structure, content, and formatting, this is good to go.

iacore:
> Apart from the PR message structure, content, and formatting, this is good to go.

@zerbina Can you fix the PR message for me? I don't know what it is supposed to look like.

zerbina:
Okay, I will do so tomorrow.

zerbina:
I've updated the message. The problems with the previous one were that:
- the lines had more than 72 characters
- the description implied that `std/marshal` couldn't be compiled previously, which is incorrect

@iacore If you're okay with the changes, please ping me, and I'll start the merge.

iacore:
@zerbina this is fine.

zerbina (approved):

PR #847, opened 2023-08-18
fix object subtype-relation between generic instances

zerbina:
## Summary

Fix a bug with the subtype-of-generic-object logic that caused generic
instances of object-like types to be treated as unrelated when one of
their in-between types is a `ref`/`ptr` type while the formal type is
not (or vice versa). Refer to the added test for an example of where
this issue surfaced.

## Details

The mistake was considering the skipped ptr/ref at each inheritance
step, while it must only be considered once for the formal and original
actual type. This properly considers `ref` and `ptr` types being used
as object base types.

In addition to the fix, the documentation of and inside
`isSubtypeOfGenericInstance` is slightly improved.

### DIFF
--- a/tests/typerel/tphantom_type_as_base.nim
+++ b/tests/typerel/tphantom_type_as_base.nim
@@ -73,3 +73,24 @@ block with_ptr_object:
   proc p(x: ptr Sub): int = 3
   # the more precise overload matches:
   doAssert p(addr obj) == 3
+
+block with_intermediate_non_object_base:
+  # test the case where an intermediate base type is a ``ref`` or
+  # ``ptr`` while formal type is not
+  type
+    Root[T] = object of RootObj
+
+    First   = ref object of Root[int]
+    Second  = object of First
+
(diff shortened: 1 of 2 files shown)


tests/typerel/tphantom_type_as_base.nim
```diff
@@ -72,4 +72,25 @@ block with_ptr_object:
...
+  doAssert p(addr obj) == 3
+
+block with_intermediate_non_object_base:
+  # test the case where an intermediate base type is a ``ref`` or
+  # ``ptr`` while formal type is not
+  type
+    Root[T] = object of RootObj
+
+    First   = ref object of Root[int]
+    Second  = object of First
```
saem, line 84: no change for this PR. musing outloud really, but I do wonder about this feature... maybe it's ok under the "what else could it be?" although those sorts of things preclude future extension of the language.

saem (approved):
this is a really good find, and fix. 😃

PR #848, opened 2023-08-19
generics: fix coercions to distinct phantom types

zerbina:
## Summary

Fix phantom type information being lost in conversions-with-inference
(i.e, `Typ(x)` where `Typ` is a generic type) to distinct types.

## Details

Matching a `B[U]` against a `A[T] = distinct T`, which is what
eventually happens for conversions like `A(B)`, results in the actual
type `B[U]` being matched against the formal type `distinct T`. Since
the invocation information (i.e., `tyGenericInst`) of the actual
type is skipped when the formal type is a `tyDistinct`, phantom type
information present on the instance was lost.

Skipping the invocation information is now removed, and phantom types
are thus passed on properly.

### DIFF
--- a/compiler/sem/sigmatch.nim
+++ b/compiler/sem/sigmatch.nim
@@ -1082,7 +1082,7 @@ typeRel can be used to establish various relationships between types:
 
   if a.kind == tyGenericInst and
       skipTypes(f, {tyStatic, tyVar, tyLent, tySink}).kind notin {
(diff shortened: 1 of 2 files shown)


saem (approved):
Sorry it took a while to review, I was thinking a bunch about whether the semantics made sense given current state and future direction. They definitely do. 🥳

It's a really good change, making things far more consistent.

PR #849, opened 2023-08-19
internal: always-on logging of generated VM bytecode

zerbina:
## Summary

Always include support for logging the VM bytecode generated during
compile-time execution in the compiler. `-d:expandVmListing` is re-used
for selecting what to output. In addition, the feature is decoupled from
the legacy reporting machinery.

The goal is improve the debugging experience when working on the VM, by
not having to temporarily re-build the compiler with
`nimVMDebugGenerate` when wanting to log the the generated bytecode.

## Details

* move rendering of VM code listing from `cli_reporter` to `vmutils`,
  with some minor code-style adjustments applied
* add the `logBytecode` procedure to the `compilerbridge` module. It is
  responsible for deciding what to render, invoking the renderer, and
  writing the result to the output (via `msgWrite`)
* call `logBytecode` whenever new bytecode becomes available
* remove the `rdbgVmCodeListing` report, the data types related to it,
  and the special handling for it in `options`
* change `codeListing` to be a proc, as it doesn't adhere to the
  "strict funcs" requirements
* remove all usages of `nimVMDebugGenerate`
* update the feature's documentation in `debug.rst`

The key difference that allows for always including the logging support
is that a code listing is only created when it's meant to be rendered.
Previously, a listing was created for *all* code (even if it was only
discarded afterwards), which introduced non-negligible overhead.

### DIFF
--- a/compiler/ast/report_enums.nim
+++ b/compiler/ast/report_enums.nim
@@ -838,7 +838,6 @@ type
     #----------------------------  Debug reports  ----------------------------#
     rdbgVmExecTraceFull
     rdbgVmExecTraceMinimal
-    rdbgVmCodeListing
 
     rdbgOptionsPush
     rdbgOptionsPop
--- a/compiler/ast/reports_debug.nim
+++ b/compiler/ast/reports_debug.nim
@@ -6,7 +6,6 @@
 
 import
   compiler/ast/[
-    ast_types,
     lineinfos,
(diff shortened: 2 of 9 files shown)


saem (approved):
The delete to add ratio, the significant simplification in logic within the default reporter, shrinking of legacy reports, more convenient debugging, what's not to love?

PR #850, opened 2023-08-20
better dependent-expression support for range types

### LINKED ISSUE #846: simplify instantiation of generic types (`semtypinst`)
## Goal

## Why?

## Concrete benefits

## How?

zerbina:
## Summary

Fix complex type-variable-dependent expressions not working reliably
for range and array-index types used in routine signatures.
Internally, expressions in both contexts are now first treated as
generic, and using `replaceTypesInBody` for expressions with
potentially unbound type variables is removed.

**Language changes:**
- syntax macros and templates (i.e., those where all parameters are 
  `untyped`) used in range- and index-type expression are expanded
  before all further analysis (and thus typed macros/template, static
  expressions, etc.)
- typed macros and templates used in range- and index-type expressions
  depending on type variables are potentially expanded multiple times

**Fixes:**
- type-variable-dependent range and array-index types in routine
  signatures can now be arbitrarily complex (refer to the added tests
  for an example of what is now possible)

## Details

**Overview:**
- all expressions used with range types and in array-index positions
  now use the generic pre-pass
- type-variable-dependent expressions in range type constructors are
  now always generic expressions (i.e., symbols are looked up, but
  not AST is not typed)
- `replaceTypesInBody` is renamed to `instantiateTypesInBody` and
  support for unresolved type variables is removed -- the new
  `replaceTypeVarsInBody` now has to be used for that case

Semantic analysis has very basic, scattered support for expressions
depending on unresolved type variables, allowing `semExprWithType` to
be used for typing expressions in range and array type constructors.
However, in more complex cases, this usually results in errors with
unclear descriptions. Until more complete support for typing expressions
depending on unresolved type variables is implemented, it is more
robust to treat them as fully generic.

### Analyzing the expressions

Both `semRangeAux` and `semArrayIndex` now always run the generic pre-
pass first (`semGenericStmt`, which also expands syntax macros), which
is required for detecting whether unresolved type variables are used.
When no unresolved type variables exist in the analyzed expression, it
is directly passed on to `semExprWithType`.

The procedure for querying whether a symbol is that of an unresolved
type variable (`isUnresolvedSym`), is moved to the `ast_query`
module.

Looking for unresolved type variables is combined with making sure that
all type variables are referenced in the AST via their symbols
(`fixupTypeVars`) -- something that is not always already the case. Some
type variables not having their symbol bound would mean that the later
pass for replacing type variables cannot detect them.

As an unintended side-effect of using `semGenericStmt`, unrelated
symbols are marked as used, which is what causes the different error
message in `tgeneric_backtrack_type_inference.nim` (the presence of the
`sfUsed` flag for generic parameters of routines decides whether a
routine is marked with `tfUnresolved`).

### Processing the generic expressions

For processing generic expressions used in the context of `static`
parameter inference, where unbound type variables are possible, the
`replaceTypeVarsInBody` procedure is introduced. While using 
`replaceTypesInBody` with `allowMetaTypes = true` would have sufficed,
this is intended as a step towards removing the "meta-types allowed"
mode from `semtypinst`.

Implementation-wise, `replaceTypeVarsInBody` is similar to `prepareNode`,
with the differences being that `replaceTypeVarsInBody`:
- does not instantiates or traverses into any types
- only considers symbol nodes referring to type variables

Expression already having a type are not analyzed again when used as call
operands, and thus two compromises are required for the generic-expression
approach to work:
- both `prepareNode` and `replaceTypeVarsInBody` have to mirror
  how `semSym` assigns types to symbol nodes of type variables (`fixType`
  implements this)
- `replaceTypeVarsInBody` has to remove from the expression the types
  previously assigned by both `semRangeAux` and `semArrayIndex`

`replaceTypesInBody` is renamed to `instantiateTypesInBody`. In addition,
the unnecessarily exported `replaceTypeVarsN` is un-exported.

### Misc

- restore the name of the `tgenericshardcases` test, which was
  erroneously changed to `tsharedcases` during a test suite cleanup

### DIFF
--- a/compiler/sem/semtypinst.nim
+++ b/compiler/sem/semtypinst.nim
@@ -802,7 +802,7 @@ proc replaceTypeVarsInBody*(c: PContext, pt: TIdTable, n: PNode): PNode =
   ## the generic AST `n`, with unbound type variables being ignored. Generic
   ## types are not instantiated and static expression not evaluated.
   ##
-  ## Use this procedure instead of ``instantiateTypesInBody`` when its not
+  ## Use this procedure instead of ``instantiateTypesInBody`` when it's not
   ## guaranteed whether all type variables have been resolved.
   template lookup(t: PType): PType = PType(idTableGet(pt, t))
 
--- a/compiler/sem/semtypinst.nim
+++ b/compiler/sem/semtypinst.nim
@@ -32,7 +32,9 @@ import
     semdata,
   ],
   compiler/utils/[
-    astrepr
+    astrepr,
+    debugutils,
+    idioms
   ]
 
 # xxx: reports are a code smell meaning data types are misplaced
@@ -115,7 +117,7 @@ type
 
 proc replaceTypeVarsTAux(cl: var TReplTypeVars, t: PType): PType
 proc replaceTypeVarsS(cl: var TReplTypeVars, s: PSym): PSym
-proc replaceTypeVarsN*(cl: var TReplTypeVars, n: PNode; start=0): PNode
+proc replaceTypeVarsN(cl: var TReplTypeVars, n: PNode; start=0): PNode
 
 proc initLayeredTypeMap*(pt: TIdTable): LayeredIdTable =
   result = LayeredIdTable()
@@ -154,6 +156,26 @@ proc isTypeParameterAccess(cl: TReplTypeVars, s: PSym): bool =
   s.kind == skType and s.typ.kind == tyTypeDesc and
     cl.typeMap.lookup(s.typ[0]) != nil
 
+proc fixType(c: PContext, s: PSym, t: PType): PType =
+  ## Computes the correct type to assign to the symbol node of a replaced
+  ## type variable.
+  # XXX: ideally, this would be left to ``semSym``, but at the time of writing,
+  #      an expression having a type prevents it from being analyzed again in
+  #      some cases (e.g., call operands), meaning that we have to already
+  #      assign the expected type here
+  case s.kind
+  of skGenericParam:
+    if s.typ.kind == tyStatic:
+      t
+    else:
+      makeTypeDesc(c, t)
+  of skType:
+    makeTypeDesc(c, t)
+  of skParam:
+    t
+  else:
+    unreachable(s.kind)
+
 proc prepareNode*(cl: var TReplTypeVars, n: PNode): PNode =
   ## Replaces unresolved types referenced by the nodes of expression/statement
   ## `n` with the corresponding bound ones.
@@ -161,6 +183,8 @@ proc prepareNode*(cl: var TReplTypeVars, n: PNode): PNode =
   ## Because of how late-bound static parameters are currently implemented,
   ## ``prepareNode`` is also responsible for replacing sub-expression of type
   ## ``tyStatic`` with the computed value.
+  addInNimDebugUtils(cl.c.config, "prepareNode", n, result)
+
   template resolveStatic() =
     ## Redirects to the evaluated (or unevaluated) value of the static type, if
     ## `n` resolves to a ``tyStatic``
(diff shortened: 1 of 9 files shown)


zerbina:
The PR message was quite the mess, sorry for that. I've cleaned it up a bit, and hopefully it's easier to read and parse now.

Looking ahead, the new `fixupTypeVars`, if merged, would be a great help in fixing the issues with `when` statements inside generic objects (they too use `semExprWithType` at the moment).

compiler/sem/semtypinst.nim
```diff
@@ -154,13 +156,35 @@ proc isTypeParameterAccess(cl: TReplTypeVars, s: PSym): bool =
   s.kind == skType and s.typ.kind == tyTypeDesc and
     cl.typeMap.lookup(s.typ[0]) != nil
 
+proc fixType(c: PContext, s: PSym, t: PType): PType =
+  ## Computes the correct type to assign to the symbol node of a replaced
+  ## type variable.
```
saem, line 161:
```suggestion
  ## Computes the correct type to assign to the symbol of a replaced type
  ## variable.
```

not sure if that's too pendantic, but we're technically working with the symbol and not an `nkSym`.

compiler/sem/semtypinst.nim
```diff
@@ -755,15 +786,74 @@ proc initTypeVars*(p: PContext, typeMap: LayeredIdTable, info: TLineInfo;
...
   result = replaceTypeVarsN(cl, n)
   popInfoContext(p.config)
 
+proc replaceTypeVarsInBody*(c: PContext, pt: TIdTable, n: PNode): PNode =
+  ## Replaces with their bound types (provided by `pt`) all type variables in
+  ## the generic AST `n`, with unbound type variables being ignored. Generic
+  ## types are not instantiated and static expression not evaluated.
+  ##
+  ## Use this procedure instead of ``instantiateTypesInBody`` when its not
+  ## guaranteed whether all type variables have been resolved.
```
```suggestion
  ## Use this procedure instead of ``instantiateTypesInBody`` when it's not
  ## guaranteed whether all type variables have been resolved.
```

very minor.

tests/lang_callable/generics/tgeneric_backtrack_type_inference.nim
```diff
@@ -1,5 +1,5 @@
 discard """
-  errormsg: "\'vectFunc\' doesn't have a concrete type, due to unspecified generic parameters."
+  errormsg: "type mismatch: got <proc (x: vector[vectFunc.T]): vector[vectFunc.T], vector[3]>"
```
saem, line 2: this is _really_ nice.

saem (approved):
The way it works now is quite a bit better. Only some non-blocking suggestions.

Test coverage seems good. I tried to come up with some meaningful `range` tests to suggest, but nothing particularly novel occurred to me. The best I could come up with is:

```nim
proc r(t: typedesc, rng: range[-typeNameLen(t), typeNameLen(t]): int =
  result = rng.type.high - rng.type.low

doAssert r(MyType, 0) == 12
```

saem:
> The PR message was quite the mess, sorry for that. I've cleaned it up a bit, and hopefully it's easier to read and parse now.

thanks for the additional edits, in fairness i found the earlier version was already pretty good, but the improvement is appreciated all the same.

> Looking ahead, the new `fixupTypeVars`, if merged, would be a great help in fixing the issues with `when` statements inside generic objects (they too use `semExprWithType` at the moment).

considering i'd like to lean on `when` and generics a whole lot more (reducing macro usage) that's a win all around. :D

---

Awesome PR.

compiler/sem/semtypinst.nim
zerbina: Yep, we're working with the symbol, but the returned type is meant to be assigned to the node, hence the explicit mention.

zerbina:
Thanks a lot for the review, @saem.

Adding a test for `range` was a good call, as it uncovered unrelated issues with `typeRel` for range types. They are not hard to fix, but I think it'd be better to do so in a follow-up PR (I've marked the added test as a known issue).

PR #851, opened 2023-08-22
sem: use `nkError` in `break`/`continue` analysis

zerbina:
## Summary

Split semantic analysis of 'break' and 'continue' statements into two
procedures and use `nkError` propagation instead of `localReport`. The
quality of errors reported for invalid 'break' statements is also
improved.

## Details

* split `semBreakOrContinue` into two procedures, as the handling of
  both had only very little in common
* refactor the analysis of 'break' statements. `ElaborateAst` is used
  for easier error propagation. In addition, the input AST is no longer
  modified
* report errors through `nkError` nodes -- `AstDiagKind` counterparts
  for the previously used reports are added

**Changed errors:**

* don't report an "invalid control flow" error when the provided
  identifier is erroneous. The identifier error is forwarded instead
* introduce the new "label expected" error and use it to replace
  "ill-formed AST" and "invalid control flow" errors when the AST in
  the label slot is not the symbol of a label

Two new error message tests are added for the changed errors.


saem (approved):
nice PR, the analysis code is easier to follow and the error messages are more actionable.

PR #853, opened 2023-08-23
fix type matching failing for generic range types

zerbina:
## Summary

Fix two bugs with `typeRel` that caused no relationship to be detected
when a formal range type's underlying type is dependent on an unresolved
expression.

#### Problem One

In case they hadn't been resolved yet (because they depend on late-bound
type variables), the bounding values of the formal range type were
*modified* when computing the relationship between two range types,
which in effect meant that all following queries of the routine
signature saw an instantiated range type, instead of the original
generic one.

The resolved expressions are now stored locally and then directly passed
on to `typeRangeRel`.

#### Problem Two

Neither range-against-range nor non-range-against-range considered the
case where the range's underlying type was dependent on an unresolved
expression -- both returning `isNone` when the formal range type's base
is a `tyFromExpr`.

If the formal range type's underlying type is a `tyFromExpr`, the
dependent type is now resolved first. Were `f` was used previously
directly, the resolved underlying type is used instead.

#### Tests

An incorrect test case from the now-working `tdependent_range_type` test
is removed.

### DIFF
--- a/compiler/sem/sigmatch.nim
+++ b/compiler/sem/sigmatch.nim
@@ -659,7 +659,7 @@ proc procTypeRel(c: var TCandidate, f, a: PType): TTypeRelation =
     result = f.allowsNil
   else: discard
 
-proc typeRangeRel(f, a: PType): TTypeRelation {.noinline.} =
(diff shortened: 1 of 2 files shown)


saem (approved):
this makes ranges quite a bit more useful, yay!

PR #857, opened 2023-08-25
internal: detect illegal access of locals earlier

zerbina:
## Summary

Move detection of inaccessible locals from `vmgen` into a pre-pass over
standalone expressions/statements passed to compile-time execution. This
makes sure that all locals reaching `transf` and the MIR phase are
valid, which is a requirement for moving away from using `PSym` for
locals in the code generators.

## Details

- implement the check for whether unavailable locals are accessed in the
  new `semcomptime` module
- integrate the check into `evalConstExprAux`, where it short-circuits
  code generation in case of an error
- add a dedicated diagnostic for using unavailable locals in compile-
  time contexts (`adSemUnavailableLocation`/`rsemUnavailableLocation`)
- adjust the tests using the old error message
- remove from `vmgen` the checks regarding whether access of a local is
  valid

Implementation-wise, the check works by traversing the AST, registering
all symbols in a set as they're declared, and then, on use, checking
whether non-global symbol are present in the set. If not, the node
with the problematic symbol is propagated back to the callsite, where it
is wrapped in an error node.

While it would be possible to integrate the check into `sempass2`, the
future of the latter isn't entirely clear and the pass is at present
also not run for standalone compile-time expressions/statements.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -490,6 +490,10 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
 
       result.add("; routine: ", r.symstr)
 
+    of rsemUnavailableLocation:
+      result = ("accessed location '$1' doesn't exist in the current " &
+               "compile-time context") % [r.ast.render]
+
     of rsemIllegalCallconvCapture:
       let s = r.symbols[0]
       let owner = r.symbols[1]
@@ -3284,7 +3288,8 @@ func astDiagToLegacyReport(conf: ConfigRef, diag: PAstDiag): Report {.inline.} =
       adSemExpectedIdentifierQuoteLimit,
       adSemExpectedRangeType,
       adSemExpectedLabel,
-      adSemContinueCannotHaveLabel:
+      adSemContinueCannotHaveLabel,
+      adSemUnavailableLocation:
     semRep = SemReport(
         location: some diag.location,
         reportInst: diag.instLoc.toReportLineInfo,
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1113,6 +1113,8 @@ type
     # vm
     adVmError
     adVmQuit
+    # semcomptime
+    adSemUnavailableLocation
     # semcall
     adSemRawTypeMismatch
     adSemExpressionCannotBeCalled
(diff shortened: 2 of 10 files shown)


zerbina:
Since `check` happens prior to `transf`, where `len(array)` calls are folded away, it picks up references to locals that previously didn't (and still don't) reach `vmgen`, treats that as an error, and thus causes the test failures.

For now, I'm going to work around this by having `check` ignore calls to `mLengthArray`, but ideally, the `len` calls should have already been folded away at this point.

compiler/front/cli_reporter.nim
```diff
@@ -490,6 +490,10 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
 
       result.add("; routine: ", r.symstr)
 
+    of rsemUnavailableLocation:
+      result = ("accessed location '$1' doesn't exist in the current " &
+               "compile-time context") % [r.ast.render]
```
saem, line 495:
```suggestion
      result = ("accessed non-compile-time location '$1' within a " &
               "compile-time context") % [r.ast.render]
```

I believe this makes it slightly more actionable, and reads a bit better from the perspective of "what's wrong with this line?".

saem (approved):
wow, it really should have worked this way all along. :D

---

I suggested what I believe to be an improvement to the error message. My thinking is the end programmer's internal monologue is "what's wrong with this line?". Then the compiler 'replies', this line "accessed a non-compile-time location 'foo' within a compile-time context".

I think it's _slightly_ better than:
1. "what's wrong with this line?"
2. "accessed location 'foo' doesn't exist in the current compile-time context"

my reasoning being that the former gives a stronger hint to the end programmer that they're either accessing something they shouldn't (with the understanding they're in a compile-time context) or that they aren't in the context they think they are (run vs compile-time).

compiler/front/cli_reporter.nim
zerbina: Yep, I think that's better, thank you.

zerbina: Hm, on further thought, the message would be incorrect in the following case:
```nim
static:
  var x = 0
  static:
    echo x
```
Here `x` is a compile-time location, but its definition happens at a "later" time (i.e., the inner `static` statement is evaluated first) than its usage, meaning that the usage of `x` is invalid/disallowed.

saem: Ah, multi-staged compilation for the win. 😆 

~It's more like the location doesn't _yet_ exist in the current compile-time context. Maybe the word 'yet' is the hint for the end programmer to think about timing/stages?~

I would stick with the original message.

zerbina: Do you mean the "cannot evaluate at compile-time" one or the "accessed location doesn't exist"?

saem: The "accessed location doesn't exist" one

zerbina:
/merge

PR #858, opened 2023-08-25
internal: improve `tryGenerateInstance`

### LINKED ISSUE #846: simplify instantiation of generic types (`semtypinst`)

zerbina:
## Summary

Replace the meta-type instantiation plus subsequent test for whether
the resulting type is still generic with directly testing for whether
the to-be-instantiated type contains unbound type variables.

Apart from being more efficient, this also allows for removing the now-
unused `prepareMetatypeForSigmatch` procedure, which is a step
towards removing the "meta-types allowed" mode from `semtypinst`.

## Details

Compared to `containsGenericType`, the new `containsUnresolvedTypeVar`
looks for unbound type variables instead of for generic types. This
slightly changes the meaning of `tryGenerateInstance` from "produces
non-meta, concrete type on success" to "produces concrete type on
success". However, since the single usage-site of `tryGenerateInstance`
(`procParamTypeRel`) already tests for meta types being returned, no
additional adjustment is required.

### DIFF
--- a/compiler/sem/semtypinst.nim
+++ b/compiler/sem/semtypinst.nim
@@ -984,46 +984,72 @@ proc generateTypeInstance*(p: PContext, pt: TIdTable, info: TLineInfo,
     var position = 0
     recomputeFieldPositions(objType, objType.n, position)
 
-proc prepareMetatypeForSigmatch*(p: PContext, pt: TIdTable, info: TLineInfo,
-                                 t: PType): PType =
-  var typeMap = initLayeredTypeMap(pt)
-  var cl = initTypeVars(p, typeMap, info, nil)
-  cl.allowMetaTypes = true
-  pushInfoContext(p.config, info)
-  result = replaceTypeVarsT(cl, t)
-  popInfoContext(p.config)
-
 template generateTypeInstance*(p: PContext, pt: TIdTable, arg: PNode,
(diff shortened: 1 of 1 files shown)


compiler/sem/semtypinst.nim
```diff
@@ -984,46 +984,72 @@ proc generateTypeInstance*(p: PContext, pt: TIdTable, info: TLineInfo,
...
-  result = replaceTypeVarsT(cl, t)
-  popInfoContext(p.config)
-
 template generateTypeInstance*(p: PContext, pt: TIdTable, arg: PNode,
                                t: PType): untyped =
   generateTypeInstance(p, pt, arg.info, t)
 
+proc containsUnboundTypeVar(pt: TIdTable, t: PType): bool =
+  ## Given the type `t`, answers the question of whether `t` either is or uses
+  ## an unbound type variable somewhere. `pt` provides the bindings.
```
saem, line 993: no change for this PR, but just in case you hadn't encountered it, but `pt` is typically called a `type environment` in literature. hoping that's useful in future renaming considerations if and when they come up.

saem (approved):

saem:
/merge

sorry about the delay in reviewing

PR #859, opened 2023-08-26
internal: speed up `FileInfoIdx`-from-path lookup

### LINKED ISSUE #546: Canonicalized filepaths has a large performance impact on compilation speed
Currently the compiler supports emiting "canonical" paths for modules, instead of say absolute, project file relative, etc, a list of possibilities can be seen in `compiler/front/in_options.FilenameOption`. Most pressingly, `foCanonical` results in massive performance problems as each attempt at canonicalzing a path results in a lot of filesystem IO.

The goal is fairly simple:

1. get the baseline performance of compiling with `--filenames=canonical` and without
2. track down the various use cases and classify them by purpose/intention
3. with the benchmark data and use cases in hand, eliminate/reduce the filesystem access (AKA cache the data)

To measure the impact search the `/tests` directory for tests that specify the `--filenames=canonical` option, then run those with and without to observe the difference. The largest impact was observed with `./koch.py temp --stacktrace --stacktracemsgs -d:nimCompilerStacktraceHints --lib:lib -d:debug -d:usenodeids c --filenames=canonical --msgFormat=sexp tests/compilerfeatures/tstructured_parse_fail.nim`, as this resulted in `setMsgFrame` calls which converted `TLineInfo` to strings triggering the canonicalization for many hot compiler procedures.

If stuck at any point, ask @saem.

bung87:
## Summary

Cache the result of file-info-from-absolute-path lookups, avoiding
unnecessary filesystem IO incurred by expanding file paths. This
significantly increases the speed of the incremental compilation mode
(which performs a lot of those lookups), with compile time reductions
of up to 90% (measured with the `tstdlib_import_changed.nim` test).

## Details

Add the `rawPathToIndexTbl` table to `MsgFormat`. It associates the
unprocessed absolute file paths passed to `fileInfoIdx` with their
resolved-to `FileInfoIdx`, preventing the repeated expansion of the
same file paths.

In addition, the line overrides in the `assert` and `jsAssert`
templates are changed to use the no-argument form (`{.line.}`). The
latter is semantically equivalent to the previous `{.line: loc.}`,
while not performing an unnecessary `FileInfoIdx` -> filename ->
`FileInfoIdx` round-trip.

### DIFF
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -1167,9 +1167,9 @@ proc fileInfoKnown*(conf: ConfigRef; filename: AbsoluteFile): bool =
   result = conf.fileIdxTbl.hasKey(filename.string)
 
 proc fileInfoIdx*(conf: ConfigRef; filename: AbsoluteFile; isKnownFile: var bool): FileIndex =
-  let hasKey = conf.fileIdxTbl.hasKey(filename.string)
-  if hasKey:
-    return conf.fileIdxTbl[filename.string]
+  result = conf.fileIdxTbl.getOrDefault(filename.string, InvalidFileIdx)
+  if result != InvalidFileIdx:
+    return
   var
     canon: AbsoluteFile
     pseudoPath = false
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -95,8 +95,8 @@ type
     writtenSemReports*: ReportSet
     lastError*: TLineInfo
     filenameToIndexTbl*: Table[string, FileIndex]
-    rawPathToIndexTbl*: TableRef[string, FileIndex] ## maps non-canonicalized paths 
-    ## of known-files to the corresponding file index
+    rawPathToIndexTbl*: Table[string, FileIndex] ## maps non-canonicalized
+    ## paths of known-files to the corresponding file index
     fileInfos*: seq[TFileInfo] ## Information about all known source files
     ## is stored in this field - full/relative paths, list of line etc.
     ## (For full list see `TFileInfo`)
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -21,7 +21,7 @@ when not FileSystemCaseSensitive:
 from terminal import isatty
 from times import utc, fromUnix, local, getTime, format, DateTime
 from std/private/globs import nativeToUnixPath
-import tables
+
 from compiler/ast/ast_types import
   TNodeKind,  # used in conversion from `ParsedNode` to `PNode`
   TNodeFlag,  # used in conversion from `ParsedNode` to `PNode`
@@ -95,6 +95,8 @@ type
     writtenSemReports*: ReportSet
     lastError*: TLineInfo
     filenameToIndexTbl*: Table[string, FileIndex]
+    rawPathToIndexTbl*: TableRef[string, FileIndex] ## maps non-canonicalized paths 
+    ## of known-files to the corresponding file index
     fileInfos*: seq[TFileInfo] ## Information about all known source files
     ## is stored in this field - full/relative paths, list of line etc.
     ## (For full list see `TFileInfo`)
@@ -112,6 +114,7 @@ proc initMsgConfig*(): MsgConfig =
   result.fileInfos = @[]
   result.errorOutputs = {eStdOut, eStdErr}
   result.filenameToIndexTbl["???"] = FileIndex(-1)
+  result.rawPathToIndexTbl = newTable[string, FileIndex]()
 
 func incl*(s: var ReportSet, id: NodeId) = s.ids.incl uint32(id)
 func contains*(s: var ReportSet, id: NodeId): bool = s.ids.contains uint32(id)
@@ -312,7 +315,6 @@ type
 
     when defined(nimDebugUnreportedErrors):
       unreportedErrors*: OrderedTable[NodeId, PNode]
-    fileIdxTbl*: TableRef[string, FileIndex]
 
 template `[]`*(conf: ConfigRef, idx: FileIndex): TFileInfo =
   conf.m.fileInfos[idx.uint32]
@@ -934,7 +936,6 @@ proc newConfigRef*(hook: ReportHook): ConfigRef =
     maxLoopIterationsVM: 10_000_000,
     vmProfileData: newProfileData(),
     spellSuggestMax: spellSuggestSecretSauce,
-    fileIdxTbl: newTable[string, FileIndex]()
   )
   initConfigRefCommon(result)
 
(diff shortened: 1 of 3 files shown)


compiler/front/options.nim
```diff
@@ -21,7 +21,7 @@ when not FileSystemCaseSensitive:
 from terminal import isatty
 from times import utc, fromUnix, local, getTime, format, DateTime
 from std/private/globs import nativeToUnixPath
-
+import tables
```
zerbina: Please move this together with other `std` imports.

The convention is to have all full imports grouped together at the top of the file, followed by `export`s, `from` imports, and other imports that have some commentary attached.

compiler/front/options.nim
```diff
@@ -312,6 +312,7 @@ type
 
     when defined(nimDebugUnreportedErrors):
       unreportedErrors*: OrderedTable[NodeId, PNode]
+    fileIdxTbl*: TableRef[string, FileIndex]
```
zerbina: While not strictly related to the "message configuration", please move this field into `MsgConfig`, placed together with the `filenameToIndexTbl`. This will make it easier to split out the file-related state in the future.

In addition:
- please document what the field does (e.g., "maps non-canonicalized paths of known-files to the corresponding file index")
- give it a more descriptive name (for example, `rawPathToIndexTbl`)

compiler/front/options.nim
```diff
@@ -1162,16 +1164,12 @@ proc canonicalCase(path: var string) =
...
-  except OSError:
-    canon = filename
-  canon.string.canonicalCase
-  result = conf.m.filenameToIndexTbl.hasKey(canon.string)
+  result = conf.fileIdxTbl.hasKey(filename.string)
 
 proc fileInfoIdx*(conf: ConfigRef; filename: AbsoluteFile; isKnownFile: var bool): FileIndex =
+  let hasKey = conf.fileIdxTbl.hasKey(filename.string)
+  if hasKey:
+    return conf.fileIdxTbl[filename.string]
```
```suggestion
  result = conf.fileIdxTbl.getOrDefault(filename.string, InvalidFileIdx)
  if result != InvalidFileIdx:
    return
```
This way it's slightly faster (one table lookup instead of two)

compiler/front/options.nim
```diff
@@ -1197,6 +1195,8 @@ proc fileInfoIdx*(conf: ConfigRef; filename: AbsoluteFile; isKnownFile: var bool
     conf.m.fileInfos.add(newFileInfo(canon, if pseudoPath: RelativeFile filename
                                             else: relativeTo(canon, conf.projectPath)))
     conf.m.filenameToIndexTbl[canon2] = result
+  if not hasKey:
+    conf.fileIdxTbl[filename.string] = result
```
```suggestion

  conf.fileIdxTbl[filename.string] = result
```
The `if` is redundant, as the condition will always be 'true' here.

lib/js/jsconsole.nim
```diff
@@ -107,7 +107,7 @@ since (1, 5):
     const
       loc = instantiationInfo(fullPaths = compileOption("excessiveStackTrace"))
       msg = getMsg(loc, astToStr(assertion)).cstring
-    {.line: loc.}:
+    {.line: instantiationInfo(fullPaths = true).}:
```
zerbina: What is the purpose of this change?

compiler/front/options.nim
```diff
@@ -1162,16 +1164,12 @@ proc canonicalCase(path: var string) =
...
 proc fileInfoKnown*(conf: ConfigRef; filename: AbsoluteFile): bool =
-  var
-    canon: AbsoluteFile
-  try:
-    canon = canonicalizePath(conf, filename)
-  except OSError:
-    canon = filename
-  canon.string.canonicalCase
-  result = conf.m.filenameToIndexTbl.hasKey(canon.string)
+  result = conf.fileIdxTbl.hasKey(filename.string)
```
zerbina: This is not correct, as it means that `fileInfoKnown(x)` would return 'false' if there's no preceding call to `fileInfoIdx(x)`.

zerbina (changes requested):
Welcome, and thank you for working on this.

I've left a few requests regarding code style, data placement, and documentation. In addition, please update the PR message and title according to the contribution guidelines. The guidelines contain examples of how the title and message should look like, and for inspiration, you can also look at PRs merged previously.

lib/js/jsconsole.nim
bung87: the line pragma always expected absolute path. even it can handle relative path in down stream, it will occurs several file io error on compile time.

compiler/front/options.nim
bung87: this proc is only used in nimsuggest api, I assuming the project already compiled and  `fileInfoIdx` called on every project files, no?

compiler/front/options.nim
bung87: how to move this together with other `std` imports? there's only  one `from std/... import ...` while I want import the whole module.

bung87: oh, I see , it's already imported

compiler/front/options.nim
bung87: am not sure now, just revert it.

compiler/front/options.nim
```diff
@@ -95,6 +95,8 @@ type
     writtenSemReports*: ReportSet
     lastError*: TLineInfo
     filenameToIndexTbl*: Table[string, FileIndex]
+    rawPathToIndexTbl*: TableRef[string, FileIndex] ## maps non-canonicalized paths 
+    ## of known-files to the corresponding file index
```
```suggestion
    rawPathToIndexTbl*: Table[string, FileIndex] ## maps non-canonicalized
    ## paths of known-files to the corresponding file index
```
I've missed a `TableRef` being used in my previous review, apologies. There's no need for using one here.

compiler/front/options.nim
```diff
@@ -112,6 +114,7 @@ proc initMsgConfig*(): MsgConfig =
   result.fileInfos = @[]
   result.errorOutputs = {eStdOut, eStdErr}
   result.filenameToIndexTbl["???"] = FileIndex(-1)
+  result.rawPathToIndexTbl = newTable[string, FileIndex]()
```
```suggestion
  result.rawPathToIndexTbl = initTable[string, FileIndex]()
```

lib/js/jsconsole.nim
zerbina: Got it, thanks. Since you're changing this line anyway, please use `{.line.}` (without any parameter) instead -- it does the same thing, while being much more efficient.

lib/system/assertions.nim
```diff
@@ -44,7 +44,7 @@ template assertImpl(cond: bool, msg: string, expr: string, enabled: static[bool]
       ploc = $loc
     bind instantiationInfo
     mixin failedAssertImpl
-    {.line: loc.}:
+    {.line: instantiationInfo(fullPaths = true).}:
```
zerbina:
```suggestion
    {.line.}:
```

zerbina (changes requested):
Thanks. There are few things I missed during my earlier review that still need to be changed, but apart from those and the PR message/title, this is good to.

---

For the PR title, make sure it reads like a changelog entry (it should also not include the fixed issue) and is 50 characters in length (a few more are generally okay).

The "summary" section should provide a short overview of what changed and why it changed. The "details" section should provide details about the implementation and steps taken. In addition, each line in the message body must be no longer than 72 characters.

As it stands, most of what you have placed in the "summary" section should be part of the "details" section.

zerbina:
The `pragmaLine` procedure ignores the provided column information, which is why the `t9768.nim` test failed when you used the `.line` pragma without any argument.

I'll quickly make a separate PR with a fix, after which you can merge `devel` into this PR and use `.line` again.

zerbina:
Okay, the fix is merged. After merging the `devel` branch into this PR, you should be able to use only `{.line.}`.

bung87:
now all passed, am not native English speaker, feel free to modify the title and description.

zerbina (approved):
I've updated both the PR message and title. Thank you seeing this through. Nice work! IC is *a lot* faster now.

I did remove the mention of #546, however, as the caching doesn't affect the `--filenames=canonical` mode more than it affects the other modes (`canonicalImportAux` doesn't use `fileInfoIdx`).

bung87:
Thank you for review my PR and make a better description. glad to hear it getting faster, nice!

saem:
That's a very nice performance improvement, thanks @bung87

PR #860, opened 2023-08-26
fix `.line` pragma ignoring column information

zerbina:
## Summary

Use the column information provided by the tuple value, which fixes,
for example, stack traces produced by assertion failures in compile-time
contexts pointing to the wrong column.

In addition, the tuple passed to the `.line` pragma is now verified to
have the correct number of elements, which fixes the compiler crashing
when it's a tuple with less than the expected number of elements.

## Details

In `pragmaLine`:
- don't allow `nkPar` -- all literal tuple construction expressions have
  to use `nkTupleConstr` nodes
- require the provided tuple node to have exactly three elements
- use more descriptive names instead of `x` and `y`
- use the third tuple element's value (which is required to be an
  integer value) as the column information
- build the `TLineInfo` with `newLineInfo`, which also makes sure
  that out-of-range line and column numbers are properly handled (by
  clamping them)
- use the correct node when creating errors and use `wrapError` for
  proper propagation

In addition, the error message for when the provided tuple value doesn't
have the correct shape is slightly improved, now detailing how the
tuple is required to look like.

A test (`t9768.nim`) that relied on incorrect column being set is
adjusted.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -888,7 +888,8 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       result = "empty 'asm' statement"
 
     of rsemLinePragmaExpectsTuple:
-      result = "tuple expected"
+      result = "a tuple value of the form '(string, int, int)' is expected," &
+               " but got: '" & r.ast.render & "'"
 
     of rsemRaisesPragmaExpectsObject:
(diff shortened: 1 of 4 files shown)


zerbina:
/merge

@saem: The change is relatively small, and in order to unblock #859, I'm merging this already.

PR #861, opened 2023-08-26
internal: don't instantiate concepts when lifting types

### LINKED ISSUE #846: simplify instantiation of generic types (`semtypinst`)

zerbina:
## Summary

Don't instantiate invocations of generic concepts when lifting types
(`liftParamType`). Instead, only turn the invocation into a
`tyUserTypeClassInst` type.

This removes a usage of `semtypinst`'s "meta-types allowed" mode, and is
thus a step towards removing the latter.

## Details

Instead of using `instGenericContainer(allowedMetaTypes = true)` for
lifting the invocation of generic concepts into a `tyUserTypeClassInst`,
the invocation is now copied, transitioned to a `tyUserTypeClassInst`,
and then has the underlying `tyUserTypeClass` the invocation applies to
added as the last element.

The above is **not** the same as what `instGenericContainer` did. Most
importantly, generic invocations used as arguments to the invocation
are not instantiated. In practice, this means that a parameter type like
`Concept[Generic[T]]` will reach `typeRel` and subsequently
`matchUserTypeClass` as

 `(UserTypeClassInst (...) (GenericInvocation (...) (...)) (...))`

instead of as

`(UserTypeClassInst (...) (GenericInst (...) (...) (...)) (...))`

The `tyGenericInvocation` has the `tfHasMeta` flag included and a
`tyInferred` is thus later created when matching against the concept,
which is correct. Since `typeRel` does support resolving `tyInferred`
types that have a `tyGenericInvocation` as the base, this doesn't cause
problems.

<!-- Note: section break (`---`) onwards is not in CI merge commit

## Notes for Reviewers

Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -1365,9 +1365,15 @@ proc liftParamType(c: PContext, procKind: TSymKind, genericParams: PNode,
       return
 
     if body.lastSon.kind == tyUserTypeClass:
-      let expanded = instGenericContainer(c, info, paramType,
-                                          allowMetaTypes = true)
-      result = recurse(expanded, true)
+      # turn the user-type-class invocation into a ``tyUserTypeClassInst``,
+      # making handling of the type easier for ``typeRel``
+      result = copyType(paramType, nextTypeId(c.idgen), body.owner)
+      result.kind = tyUserTypeClassInst
+      # an unresolved ``tyUserTypeClassInst`` currently expects the
+      # ``tyUserTypeClass`` as the last element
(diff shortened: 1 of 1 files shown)


saem (approved):
Hopefully I'm following along correctly, but given the change, if we're doing a sigmatch, then we're deferring instantiation until we're actually comparing an argument and parameter. This seems entirely correct, as we're not losing information/making the decision too quickly, as we're aiming for "late static binding".

PR #862, opened 2023-08-29
warn if `untyped` params followed by non-`untyped`

saem:
## Summary

Emit a warning if an `untyped`, or `varargs[untyped]`, parameter is
followed by a non-`untyped` parameter. This will eventually become an
error, as it violates the principle where previously analysed AST cannot
be "unanalysed" during overload dispatch resolution.

## Details

`semtypes.semProcTypeNode` to emit a warning if an `untyped`, or
`varargs[untyped]`, parameter is followed by a non-`untyped` parameter.
This is a new warning message so new entries were required in legacy
reports, unfortunately.

The manual has been updated to indicate the future direction where this
will become an error.
Finally added a test to ensure that warnings are generated as required.

As part of converting it to an error the warning will be used to track
down all APIs/signatures that require updating.

### DIFF
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -1423,8 +1423,12 @@ proc semProcTypeNode(c: PContext, n, genericParams: PNode,
   # and then 'isType' is false; this is of course all terrible design
   checkMinSonsLen(n, 1, c.config)
   result = newProcType(c, n.info, prev)
-  var check = initIntSet()
-  var counter = 0
+  var
+    check = initIntSet()
+    counter = 0
+    untypedParamPos = 0 ## the first untyped param, if one exists at all, used
+                        ## to enforce: all params following an untyped param
+                        ## must be untyped
 
   for i in 1..<n.len:
     var a = n[i]
@@ -1454,6 +1458,17 @@ proc semProcTypeNode(c: PContext, n, genericParams: PNode,
             c.config,
             a[^2].info,
             reportTyp(rsemMisplacedMagicType, typ))
+
+      template isUntyped(t: PType): bool =
+        t.kind == tyUntyped or
+          t.kind == tyVarargs and t.len > 0 and t[0].kind == tyUntyped
+
+      if untypedParamPos == 0:
+        if typ.isUntyped:
+          untypedParamPos = i
+      elif not isUntyped(typ):
+        localReport(c.config, a[^2],
+                    reportTyp(rsemUntypedParamsFollwedByMoreSpecificType, typ))
 
     if hasDefault:
       def = a[^1]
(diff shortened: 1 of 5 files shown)


zerbina:
Hmm, that's a weird test failure. I've pulled down the branch and ran the test locally, both via `cat lang_callable` and `r tests/lang_callable/overload/tuntyped_param_not_followed_by_typed.nim`, but it didn't fail there.

saem:
> Hmm, that's a weird test failure. I've pulled down the branch and ran the test locally, both via `cat lang_callable` and `r tests/lang_callable/overload/tuntyped_param_not_followed_by_typed.nim`, but it didn't fail there.

hmm, I'm doing a `check` so it won't produce an exe, and if it was compiled in the meantime it'll silently pass because there will be an old exe lying around. I think I have to change it to `action: compile`.

compiler/sem/semtypes.nim
```diff
@@ -1455,6 +1459,17 @@ proc semProcTypeNode(c: PContext, n, genericParams: PNode,
             a[^2].info,
             reportTyp(rsemMisplacedMagicType, typ))
 
+      template isUntyped(t: PType): bool =
+        t.kind == tyUntyped or
+          t.kind == tyVarargs and t.len > 0 and t[0].kind == tyUntyped
+
+      if untypedParamPos == 0:
+        if typ.isUntyped:
```
```suggestion
        if isUntyped(typ):
```
This makes skimming this part a little bit easier, as both `isUntyped` calls use the the same syntax.

zerbina (approved):
Looks good to me, and it's a good first step towards resolving the issues around too eager compile-time evaluation. I've left a soft suggestion regarding a very minor style change -- feel free to ignore it.

---

I've looked a bit into where the warnings are reported in the compiler's code base. Some can be fixed by replacing `untyped` with generic parameters, but in many other cases, `untyped` is used as a way to pass raw identifiers.

Some rule like allowing AST-constrained `untyped` parameters (e.g., `untyped{nnkIdent}`) to be followed by non-`untyped` *could* make sense here (nothing for this PR), although I'm not yet sure how/if that would work with overload sorting.

compiler/sem/semtypes.nim
saem: I switched it for the negation case because that read a fair bit better, but thought this might be ok. I'll change it, thanks.

saem:
> Looks good to me, and it's a good first step towards resolving the issues around too eager compile-time evaluation. I've left a soft suggestion regarding a very minor style change -- feel free to ignore it.

thanks for the review

---

> I've looked a bit into where the warnings are reported in the compiler's code base. Some can be fixed by replacing `untyped` with generic parameters, but in many other cases, `untyped` is used as a way to pass raw identifiers.
> 
> Some rule like allowing AST-constrained `untyped` parameters (e.g., `untyped{nnkIdent}`) to be followed by non-`untyped` _could_ make sense here (nothing for this PR), although I'm not yet sure how/if that would work with overload sorting.

Yup, the raw identifiers case is frequent. I was half considering we introduce `identifier` as a meta-type where an identifier, accent quoted, etc can be passed in, something like: `template myTemplate(name: identifier)`. `untyped{nnkIdent}` does tie us down less in the meantime. 🤔

PR #863, opened 2023-08-30
make `.cursor` fields work according to spec

zerbina:
## Summary

Don't inject lifetime hook calls for `.cursor` fields when lifting
hooks, making the `.cursor` pragma on object fields behave as described
in the specification.

This also fixes double frees and other ownership issues for closure
iterators that contained `.cursor` locals.

## Details

The specification says that the `.cursor` annotation prevents
construction/destruction pairs, and while this was already true for all
cursor locals and assignments to cursor fields, inside lifted hooks it
was only true for `ref` and `proc` fields when `--gc:arc|orc` was used.

The special-casing of `ref` and `proc` and in `fillBodyObj` is removed,
meaning that hook calls are now always prevented for all cursor fields.

A direct test for cursor fields is added plus a test for cursor locals
lifted into closure iterator environments (where the previous behaviour
caused double frees for non-`ref`/non-`proc` fields).

### DIFF
--- a/tests/arc/tnon_ref_cursor_in_object.nim
+++ b/tests/arc/tnon_ref_cursor_in_object.nim
@@ -1,0 +1,49 @@
+discard """
+  targets: "c js vm"
+  description: "Ensure that `.cursor` works for non-ref types when used on object fields"
+"""
+
+type
+  Resource = object
+    value: int
+
+  Object = object
+    r {.cursor.}: Resource
+    s {.cursor.}: seq[Resource]
+
+var numDestroy = 0
+
+proc `=copy`(x: var Resource, y: Resource) {.error.} # disallow full copies
+proc `=destroy`(x: var Resource) =
+  inc numDestroy
+
+proc test() =
+  # perform the test in procedure so that globals aren't used (their different
+  # semantics with regards to destruction would interfere)
+  var
+    r = Resource(value: 1) # initialize a resource
+    s = @[Resource(value: 2)]
+
+  # make sure no copy is required in the initializer expression:
+  var o = Object(r: r, s: s)
+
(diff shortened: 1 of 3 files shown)


zerbina:
With this PR, the [test](https://github.com/zero-functional/zero-functional/blob/0d2864eb611ea9aaf316e37af07e9c90d2887f4b/test.nim) from `zero-functional` mostly succeeds. The remaining failure is due to an unrelated problem with cursor inference.

tests/arc/tnon_ref_cursor_in_object.nim
```diff
@@ -0,0 +1,49 @@
...
+  var
+    r = Resource(value: 1) # initialize a resource
+    s = @[Resource(value: 2)]
+
+  # make sure no copy is required in the initializer expression:
+  var o = Object(r: r, s: s)
+
+  # copying the object doesn't perform a full copy of the cursor fields:
+  var o2 = o
+  discard addr(o2) # prevent `o2` from being turned into a cursor
```
saem, line 32: apologies for the possibly dumb question, but I'm pretty far behind on cursor inferencing:

does taking the address block it from being a cursor because the analysis assumes taking an `addr` means all bets are off (conservatism)?

zerbina: I don't think it's a dumb question, no worries.

Yep, that's the behaviour of the current implementation. However, relying on `addr` blocking `o2` from being turned into a cursor is a bit brittle -- the better approach is likely to run the test with `--cursorInference:off`.

saem (approved):
good thing you tried out zero-functional's tests, and nice fix -- libraries that push composition like that are good testing grounds.

PR #865, opened 2023-08-31
sem: perform constant folding in sem

zerbina:
## Summary

Perform constant folding at the end of the first semantic pass, prior to
`sempass2`. Inlining constants and folding constant expression can
produce errors and, more generally, decides program behaviour, meaning
that performing these steps in `transf` is too late. In addition to
reducing the amount of repeated work for the compiler, this change
should also make further improvements to `semfold` easier.

**Fixes/Improvements**:
- the compiler doesn't crash when errors are detected during constant
  folding
- static range, array index, conversion, and other constant folding
  errors can now be detected via `compiles`, and they're now also
  reported with `nim check`
- legacy not-nil annotations and the `ProveInit` warnings now work
  better and more reliably in code using enum values

**Breaking changes**:
- the AST returned by `getImpl` now contains the result of constant
  folding

## Details

The core of the change is the introduction of `foldInAst` and
`foldConstExprAux` (the 'Aux' postfix is used to make the future
introduction of a public `foldConstExpr` easier). They implement a
depth-first traversal of the AST and inline/fold constants in almost the
same way as `transf` previously did. Compared to `transf`, they do
support error propagation.

### Implementation of the pass

A standard AST traversal is used, but expressions and statements are
separated as a preparation for further changes to the pass. The
introduced `Folded` type both tracks whether there's an error in a tree
and also implements a simple copy-on-write mechanism that prevents
unnecessary copies when nothing changes in a sub-tree. Same as `transf`,
`getConstExpr` is called on each sub-expression, with the result being
inlined, if allowed.

### Differences to `transf`

* statement-list expressions where the first sub-node is a doc comment
  are not folded away -- the documentation generator still needs them
* `getConstExpr` is not called on nodes of literal values (`nkIntLit`,
  `nkStrLit`, etc.) as that would only introduce unnecessary copies
  (potentially forcing a whole tree to be copied)

### Updating the compiler

* all places where AST is passed on to either `sempass2` or `transf`
  now run the constant folding pass (`foldInAst`)
* since the AST is passed on to the VM (and thus `transf`) in case the
  expression is not fully constant, `evalConstExpr` and `tryConstExpr`
  first run the folding pass before calling `getConstExpr`. This is
  slightly more efficient than the other way around
* `getConstExpr2` is a convenience wrapper around `getConstExpr` that
  allows passing in `nkError` nodes

With constant expressions being already folded in AST reaching
`sempass2`, two simplifications are possible:
* `sempass2` doesn't have to tentatively fold a call in order to know
  whether a call expression should be analyzed or not
* `semcomptime` doesn't have to special-case `mLengthArray` magics

Since the constant folding pass is now *not* run for the bodies of
lifted operations (e.g., lifetime hooks), the lifting has to make sure
that the output can be consumed by logic that expects constant-folded
AST:
* `liftdestructors` attempts to directly fold `offsetOf` and `alignOf`
* `genEnumToStrProc` inserts enum literals directly as `nkIntLit`
  instead of as `skEnumField` symbols

In `semGenericArgs`, a defensive copy is of the evaluated constant
expression is created prior to assigning the created `static` type. This
prevents recursive types when `opr` is the exact same node as
`evaluated`.

Finally, the `semfold` integration is removed from `transf`. The logic
for collapsing `nkCheckedFieldExpr`s is obsolete (done by the folding
pass) and the labels in `nkOfBranch` trees can be ignored.

### Other changes

* remove support for folding `nkChckRange`, `nkStringToCString`, and
  `nkCStringToString` nodes -- these nodes are first introduced in
  `transf`, which folding now happens prior to
* remove `adSemfoldInvalidConversion` diagnostic and report associated
  with range-conversion folding errors
* remove a leftover usage of `nkPar` for meaning "tuple constructor"

The `guards` module (which implements facts and implications used in
guarded data-flow analysis) already supported enum operation (e.g.,
`==`, `in`, etc.) but since enum values reached `sempass2` as
`skEnumField` symbols (which `guards` doesn't understand), facts
involving enum values generally didn't apply/work. With enum values now
lowered into integer literals, `guards` can automatically work with
them. An existing test for the `guards` module that now works properly
is update and changed in way so that silent regressions are prevented.

Since `nkHiddenConv` nodes in the index slot of `nkBracketExpr` nodes
now still exist when folding (the nodes are removed in `transf`),
different errors are produced for static out-of-bounds array access. The
associated tests are marked as known issues until it's clearer how this
should ideally work.

### DIFF
--- a/compiler/sem/pragmas.nim
+++ b/compiler/sem/pragmas.nim
@@ -41,6 +41,7 @@ import
   ],
   compiler/sem/[
     semdata,
+    semfold,
     lookups
   ],
   compiler/backend/[
@@ -375,8 +376,10 @@ proc expectDynlibNode(c: PContext, n: PNode): PNode =
     # For the OpenGL wrapper we support:
     # {.dynlib: myGetProcAddr(...).}
     result = c.semExpr(c, n[1])
-    if result.kind == nkSym and result.sym.kind == skConst:
-      result = result.sym.ast # look it up
+    # this is AST that later gets passed to code generation, so we have
+    # perform constant folding
+    result = foldInAst(c.module, result, c.idgen, c.graph)
+    # XXX: sempass2 is missing here...
     if result.typ == nil or result.typ.kind notin {tyPointer, tyString, tyProc}:
       result = c.config.newError(n, PAstDiag(kind: adSemStringLiteralExpected))
 
--- a/compiler/sem/pragmas.nim
+++ b/compiler/sem/pragmas.nim
@@ -376,7 +376,7 @@ proc expectDynlibNode(c: PContext, n: PNode): PNode =
     # For the OpenGL wrapper we support:
     # {.dynlib: myGetProcAddr(...).}
     result = c.semExpr(c, n[1])
-    # this is AST that later gets passed to code generation, so we have
+    # this is AST that later gets passed to code generation, so we have to
     # perform constant folding
     result = foldInAst(c.module, result, c.idgen, c.graph)
     # XXX: sempass2 is missing here...
--- a/compiler/sem/semfold.nim
+++ b/compiler/sem/semfold.nim
@@ -982,5 +982,6 @@ proc getConstExprError*(m: PSym, n: PNode, idgen: IdGenerator,
                         g: ModuleGraph): PNode =
   ## If `n` is an error, returns the error. Otherwise returns the folded
   ## value, or nil, if `n` isn't constant.
-  if n.kind == nkError: n
-  else:                 getConstExpr(m, n, idgen, g)
+  case n.kind
+  of nkError: n
+  else:       getConstExpr(m, n, idgen, g)
--- a/compiler/sem/semfold.nim
+++ b/compiler/sem/semfold.nim
@@ -77,7 +77,7 @@ proc prepareWrite(f: var Folded) =
     f.node = shallowCopy(orig)
     f.wasCopied = true
     # only copy the nodes up-to `num`, the rest
-    # is copied manually
+    # are copied manually
     for i in 0..<f.i:
       f.node[i] = orig[i]
 
--- a/compiler/sem/semfold.nim
+++ b/compiler/sem/semfold.nim
@@ -791,7 +791,7 @@ proc getConstExpr(m: PSym, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNode 
   #    if a == nil: return nil
   #    result[i][1] = a
   #  incl(result.flags, nfAllConst)
-  of nkPar, nkTupleConstr:
+  of nkTupleConstr:
     # tuple constructor
     result = copyNode(n)
     if (n.len > 0) and (n[0].kind == nkExprColonExpr):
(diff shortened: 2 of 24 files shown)


compiler/sem/pragmas.nim
```diff
@@ -375,8 +376,10 @@ proc expectDynlibNode(c: PContext, n: PNode): PNode =
     # For the OpenGL wrapper we support:
     # {.dynlib: myGetProcAddr(...).}
     result = c.semExpr(c, n[1])
-    if result.kind == nkSym and result.sym.kind == skConst:
-      result = result.sym.ast # look it up
+    # this is AST that later gets passed to code generation, so we have
```
saem:
```suggestion
    # this is AST that later gets passed to code generation, so we have to
```

compiler/sem/semfold.nim
```diff
@@ -42,6 +42,65 @@ from compiler/front/msgs import internalError
...
+func fromAst(n: PNode, num: int): Folded =
+  result = Folded(node: n, i: num, hasError: false)
+
+proc prepareWrite(f: var Folded) =
+  if not f.wasCopied:
+    let orig = f.node
+    f.node = shallowCopy(orig)
+    f.wasCopied = true
+    # only copy the nodes up-to `num`, the rest
+    # is copied manually
```
saem:
```suggestion
    # are copied manually
```

compiler/sem/semfold.nim
```diff
@@ -794,3 +838,148 @@ proc getConstExpr(m: PSym, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNode
...
+  if f.hasError:
+    result = g.config.wrapError(f.node)
+  else:
+    result = f.node
+
+proc getConstExpr2*(m: PSym, n: PNode, idgen: IdGenerator, g: ModuleGraph): PNode =
+  ## If `n` is an error, returns the error. Otherwise returns the folded
+  ## value, or nil, if `n` isn't constant.
+  if n.kind == nkError: n
+  else:                 getConstExpr(m, n, idgen, g)
```
saem:
```suggestion
  case n.kind
  of nkError: n
  else:       getConstExpr(m, n, idgen, g)
```

this is incredible soft and I'm not entirely convinced, half putting it up for myself. This gives up one vertical line, but horizontally it's more compact.

compiler/sem/semstmts.nim
```diff
@@ -3276,9 +3281,10 @@ proc semStaticStmt(c: PContext, n: PNode): PNode =
   #writeStackTrace()
   inc c.p.inStaticContext
   openScope(c)
-  let a = semStmt(c, n[0], {})
+  var a = semStmt(c, n[0], {})
   closeScope(c)
   dec c.p.inStaticContext
+  a = foldInAst(c.module, a, c.idgen, c.graph)
```
saem, line 3287: it was in this file's changes that I realized that it's a bit awkward to remember to `foldInAst`

saem (approved):
LGTM

Longer term, we'll likely have to push enum lowering further out of sem into later layers, but this is still better.

---

@zerbina and I discussed using `distinct PNode` types or a runtime field `level` to track whether an AST has been appropriately lowered (e.g. folded + sempass2 for comptime).

tests/misc/tinvalidarrayaccess.nim
```diff
@@ -1,6 +1,10 @@
 discard """
   errormsg: "index 2 not in 0 .. 1"
-  line: 18
+  knownIssue: '''
+    the conversion in the array-index position is folded first, resulting in a
+    range error instead of an index error
+  '''
+  line: 23
```
Clyybber: Doesn't matter but should be 22 I think.

compiler/sem/semfold.nim
```diff
@@ -794,3 +838,148 @@ proc getConstExpr(m: PSym, n: PNode; idgen: IdGenerator; g: ModuleGraph): PNode
...
+  if n.kind == nkError:
+    return n # already an error -> nothing to do
+
+  let f = foldInAstAux(m, n, idgen, g)
+  if f.hasError:
+    result = g.config.wrapError(f.node)
+  else:
+    result = f.node
+
+proc getConstExpr2*(m: PSym, n: PNode, idgen: IdGenerator, g: ModuleGraph): PNode =
```
Clyybber: A name like `getConstExprForwardError` or `getConstExprError` would be more descriptive IMO.

Clyybber (approved):
Very nice improvement, not sure what's the best way to restore these array index errors, but maybe it'll become clearer in the future.

zerbina:
/merge

Thank you for the review, @saem and @Clyybber.

PR #866, opened 2023-09-01
semtypinst: remove the "meta-types allowed" mode

### LINKED ISSUE #846: simplify instantiation of generic types (`semtypinst`)

zerbina:
## Summary

With all users of the mode gone, the mode itself is removed, preventing
it from becoming accidentally relied on again.

## Details

* remove all usages of the `TReplTypeVars.allowMetaTypes` field
* perform some small clean-up of unused code
* remove the `allowedMetaTypes` field itself
* remove the `allowedMetaTypes` parameter from `instGenericContainer`

Closes https://github.com/nim-works/nimskull/issues/846

### DIFF
--- a/compiler/sem/semtypinst.nim
+++ b/compiler/sem/semtypinst.nim
@@ -105,11 +105,9 @@ type
     typeMap*: LayeredIdTable  # map PType to PType
     symMap*: TIdTable         # map PSym to PSym
     localCache*: TIdTable     # local cache for remembering already replaced
-                              # types during instantiation of meta types
-                              # (they are not stored in the global cache)
+                              # types. Used as a work-around for instantiation
+                              # cache issues
     info*: TLineInfo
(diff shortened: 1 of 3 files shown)


compiler/sem/semtypinst.nim
```diff
@@ -105,11 +105,9 @@ type
     typeMap*: LayeredIdTable  # map PType to PType
     symMap*: TIdTable         # map PSym to PSym
     localCache*: TIdTable     # local cache for remembering already replaced
-                              # types during instantiation of meta types
-                              # (they are not stored in the global cache)
+                              # types. Used as a work-around for instantiation
+                              # cache issues
```
saem, line 109: excellent, emphasizes that it needs to be fixed rather than actually required.

saem (approved):
awesome, very glad it's gone.

PR #867, opened 2023-09-01
treat `when` within generic objects as generic

zerbina:
## Summary

Clarify the specification regarding how `when` statements within generic
object definitions work and adjust the implementation accordingly.

`when` statements inside generic object definitions are now treated as
generic. This is a breaking change, but it makes using generic
parameters in more complex expressions possible. Most notably,
`isnot` now works properly inside generic objects' `when` conditions.

## Details

* process `nkElifBranch` conditions of `nkWhenStmt` nodes within generic
  objects via `semGenericStmt` when analyzing the definition
* remove the redundant AST checks in `replaceObjBranches` and
  `replaceTypeVarsN`; the  already take place in `semRecordNode`
* use `semConstBoolExpr` instead of `semConstExpr` when evaluating
  generic `when` conditions
* `semConstBoolExpr` not returning an `nkIntLit` happens in case of
  cascading errors, so don't treat this as an internal error

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -3084,10 +3084,24 @@ exceptions:
   translated by the compiler, the other statements are not checked for
   semantics! However, each condition is checked for semantics.
 
-The `when` statement enables conditional compilation techniques. As
-a special syntactic extension, the `when` construct is also available
+The `when` statement enables conditional compilation techniques.
+
+When statement inside `object`
+------------------------------
+
+As a special syntactic extension, the `when` construct is also available
 within `object` definitions.
 
+At which point the `when` construct is evaluated and collapsed depends on
+whether the object has generic parameters:
+* if it does not, the conditions are evaluated and a branch is chosen
+  when inside the definition
+* it it does, all condition expressions are treated as generic expressions
+  that are only instantiated, and a branch chosen, when the object type is
+  instantiated
+
+Same as for `when` statements outside of `object` definitions, each condition
+is checked for semantics and evaluated.
 
 When nimvm statement
 --------------------
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -3093,10 +3093,10 @@ As a special syntactic extension, the `when` construct is also available
 within `object` definitions.
 
 At which point the `when` construct is evaluated and collapsed depends on
-whether the object has generic parameters:
-* if it does not, the conditions are evaluated and a branch is chosen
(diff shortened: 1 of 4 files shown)


doc/manual.rst
```diff
@@ -3084,10 +3084,24 @@ exceptions:
...
+------------------------------
+
+As a special syntactic extension, the `when` construct is also available
 within `object` definitions.
 
+At which point the `when` construct is evaluated and collapsed depends on
+whether the object has generic parameters:
+* if it does not, the conditions are evaluated and a branch is chosen
+  when inside the definition
+* it it does, all condition expressions are treated as generic expressions
```
```suggestion
* if it does, all condition expressions are treated as generic expressions
```

doc/manual.rst
```diff
@@ -3084,10 +3084,24 @@ exceptions:
...
+
+When statement inside `object`
+------------------------------
+
+As a special syntactic extension, the `when` construct is also available
 within `object` definitions.
 
+At which point the `when` construct is evaluated and collapsed depends on
+whether the object has generic parameters:
+* if it does not, the conditions are evaluated and a branch is chosen
```
saem: Hmm, I wonder if this is a case of a null/zero-ary generic?

To clarify what a null/zero-ary generic is:
```nim
type
  Foo = object
  Foo[] = object
```

Although very similar, the second `Foo[]` is not really equivalent and a generic with zero your parameters. The latter triggers instantiation.

saem (approved):
I found a minor typo, but since it's in the manual I recommend changing it.

There is one non-blocking discussion point when it comes to zero explicit parameter objects with a `when` in the body.

Overall, this change in behaviour is excellent, it's one of those things where one wonders why it would work any other way.

doc/manual.rst
zerbina: Ah, darn, I forgot about nullary generics again :D. For consistency with generic procedures, where a nullary generic procedure requires instantiation, I think the same should be true for types.

I'll adjust the specification to be explicit about nullary generic types (the implementation already handles nullary generics like you described).

zerbina: I've updated the specification to account for nullary generic types.

saem: nullary generics really only exist to:
1. support generic paratial application and "type parameter subtraction" like scenarios
2. trick us

most of the time it's the latter. 😆 

---

imma read over the commit now.

PR #870, opened 2023-09-03
semtypinst: fix the type instantiation cache

zerbina:
## Summary

Fix a severe issue with the type instantiation cache that led to the
wrong types being sometimes retrieved from the cache and multiple,
redundant instances of the same type being created. With the fix, an
important compiler-internal invariant is now true: if two `PType`s
representing nominal types have a different `itemId`, they are not the
same (in terms of type equality).

## Details

The generic type instantiation cache is used when a type application
(`tyGenericInvocation`) is evaluated (`handleGenericInvocation`). The
latter previously operated as follows:
1. the invocation as-is is looked-up in the cache. If the lookup
   succeeds (which is only possible for invocations where all arguments
   are concrete types), the cached resolved instance is returned,
   otherwise, the procedure proceeds
2. immediate type parameters used as arguments are replaced (e.g., the
   `T` for `Invoc[T]`, but not for `Invoc[Invoc[T]]`) and the
3. the modified invocation is looked-up in the cache again. On success,
   the cached resolved instance is returned. In case of no success, the
   invocation is treated as producing a not-yet-seen-before resolved
   instance

In practice, this meant that all invocations where the arguments were
not immediate `tyGenericParam`eters and concrete types always resulted
in cache misses and thus new, and possibly duplicate, instances.

Since the presence of an unfinished instance of a generic type being
present in the cache always resulted in a cache miss (the
`inst.len < key.len` check), recursive generic instantiation would lead
to infinite recursion that was only bailed out of through (mis)-use of
the `localCache` table. However, since `localCache` only remembered the
most recent `PType` associated with a generic type, and the recursion
limit counter isn't reset, this, in turn, would lead to the wrong type
being committed to the global instance cache in contrived scenarios.

### Reworked Implementation

The arguments are now fully resolved prior to cache lookup, and there's
only a single lookup instead of two. A different way to look at it is
that instantiation now happens inside-to-outside instead of outside-to-
inside.

* only inherit the flags relevant for type equality from the
  invocation -- the other flags are inherited from the resolved type-
  arguments. Inheriting all flags from the invocation like was done
  previously would cause the `tfHasMeta` being erroneously included on
  the type in case of recursive generic types
* remove the unnecessary `body.kind == tyError` check. There's an
  internal assertion at the start of `handleGenericInvocation` that
  makes the check redundant
* remove the `localCache` table; it's not needed anymore
* un-export the `searchInstTypes` procedure
* for robustness, and to prevent too deep generic type recursion, keep
  the recursion limit for type instantiations

### Removed Workarounds

Tow workarounds where necessitated by multiple `PType` instances of the
same nominal type existing:
- `sameType` had to fall back to structural type equivalence for
  `tyObject` and `tyDistinct` types that resulted from generic
  invocations (indicated by the `tfFromGeneric` flag)
- the VM's type creation logic had to canonicalize `tyObject` types so
  as to not create duplicate `VmType` instances for the same object

Both workarounds are obsolete now and thus removed.

### DIFF
--- a/compiler/sem/semtypinst.nim
+++ b/compiler/sem/semtypinst.nim
@@ -98,9 +98,6 @@ type
     c*: PContext
     typeMap*: LayeredIdTable  # map PType to PType
     symMap*: TIdTable         # map PSym to PSym
-    localCache*: TIdTable     # local cache for remembering already replaced
-                              # types. Used as a work-around for instantiation
-                              # cache issues
     info*: TLineInfo
     skipTypedesc*: bool       # whether we should skip typeDescs
     isReturnType*: bool
@@ -653,7 +650,6 @@ proc replaceTypeVarsTAux(cl: var TReplTypeVars, t: PType): PType =
   of tyUserTypeClassInst:
     bailout()
     result = instCopyType(cl, t)
-    idTablePut(cl.localCache, t, result)
     for i in 1..<result.len:
       result[i] = replaceTypeVarsT(cl, result[i])
     propagateToOwner(result, result.lastSon)
@@ -668,7 +664,6 @@ proc replaceTypeVarsTAux(cl: var TReplTypeVars, t: PType): PType =
       bailout()
       result = instCopyType(cl, t)
       result.size = -1 # needs to be recomputed
-      idTablePut(cl.localCache, t, result)
 
       for i in 0..<result.len:
         if result[i] != nil:
@@ -807,7 +802,6 @@ proc replaceTypeParamsInType*(c: PContext, pt: TIdTable, t: PType): PType =
 proc initTypeVars*(p: PContext, typeMap: LayeredIdTable, info: TLineInfo;
                    owner: PSym): TReplTypeVars =
   initIdTable(result.symMap)
-  initIdTable(result.localCache)
   result.typeMap = typeMap
   result.info = info
   result.c = p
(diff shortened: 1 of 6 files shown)


compiler/sem/semtypinst.nim
```diff
@@ -423,72 +414,73 @@ proc instCopyType*(cl: var TReplTypeVars, t: PType): PType =
...
+      echo "Generic instantiation cached ", typeToString(result), " for ",
+        typeToString(t), " header ", typeToString(header)
+    return
+
+  # not cached yet
   result = newType(tyGenericInst, nextTypeId(cl.c.idgen), t[0].owner)
-  result.flags = header.flags
+  # make sure the 'not-nil' flag (which is essentialy a type modifier) is
+  # already present on the instance prior to caching
+  result.flags = header.flags * {tfNotNil}
```
saem: sorry, if this is clarified elsehwere but are we assuming that the instantiated type is not nil because we are instantiating it therefore any function pointer to said type would resolve?

saem (approved):
LGTM

compiler/sem/semtypinst.nim
zerbina: The `tfNotNil` flag is the only flag contributing to type equality that can currently be set on the invocation, and these flags have to be present on the instance prior to recursing, otherwise there would be issues with recursive generics.

An example of where problems would occur (`a` would get the same instance as `y`, which would be wrong):  
```nim
type Generic[T] = ref object
  x: Generic[T]

var y: Generic[int] not nil
```

zerbina: In order for the code to be a bit more future-proof, I'm going to change `{tfNotNil}` to `eqTypeFlags`.

PR #871, opened 2023-09-04
internal: cache transformations in `ModuleGraph`

zerbina:
## Summary

Store transformed routine bodies in the `ModuleGraph` instead of in the
associated routines' `TSym`. This makes it possible to discard the
cached transformations for selected modules, and macros can now also
have their transformed body cached. More generally, the change is a step
towards a more data-oriented symbol design.

## Details

* remove the `transformedBody` field from `TSym`
* store the transformed bodies via a `Table[int32, PNode]`, with a
  separate table instance for each module.
* read and write access is done through the new `setTransformed` and
  `getTransformed` procedures; the `transformBody` procedures are
  adjusted
* the cached transformations are discarded when a module is re-opened
  for analysis


saem (approved):
dod baby steps!

PR #872, opened 2023-09-04
suggest: discard the compile-time evaluation state

zerbina:
## Summary

Discard the state of the VM and JIT prior to executing a `nimsuggest`
command, which fixes VM-related crashes and misbehaviour caused by the
VM's internal environment state becoming stale when partially
recompiling the project.

## Details

If a module is recompiled, which is what happens when executing
`nimsuggest` commands, the module's ID generator is effectively reset,
meaning that the IDs of symbols and types are in the same ID space as
those used during the previous compilation. This is a problem for the
JIT and the VM's caches, both which use the IDs to identify symbols and
types.

Prior to executing a `nimsuggest` command that later causes the
recompilation of a module, all compile-time execution state is now
discarded first (by assigning 'nil' to `graph.vm`).

As a consequence, it becomes possible for compile-time-only routines
to be passed to `transformBody` multiple times, which is problematic if
they contain inner closure procedures (who must only be processed by
the destructive lambda-lifting pass once). When in the context of
`nimsuggest`, compile-time-only routines are now also cached, fixing this
follow-up problem.

While using a new module ID would also fix the problem, it would not
fix the problem of recompilation causing the accumulation of obsolete
VM environment and JIT state (e.g., bytecode, RTTI, etc.).

### DIFF
--- a/compiler/vm/vmjit.nim
+++ b/compiler/vm/vmjit.nim
@@ -282,13 +282,15 @@ proc genProc(jit: var JitState, c: var TCtx, s: PSym): VmGenResult =
   c.removeLastEof()
 
   let body =
-    if s.kind == skMacro:
+    if isCompileTimeProc(s) and not defined(nimsuggest):
+      # no need to go through the transformation cache
       transformBody(c.graph, c.idgen, s, s.ast[bodyPos])
     else:
-      # watch out! While compile-time only procedures don't need to be cached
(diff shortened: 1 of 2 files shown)


saem:
minor grammar related edits to the PR description.

saem (approved):
good call putting a clear warning about repeated body transformations, that should be some good inspiration for someone motivated and wanting to improve internals.

PR #874, opened 2023-09-05
sem: use `nkError` for `semOf`

zerbina:
## Summary

Replace `localReport` usage in `semOf` with `nkError` and also clean up
the procedure in general. In addition, fix the internal issue of
folded-to-false 'of' operations being replaced with their original
expression.

## Details

* remove the argument count guard. `semOf` is only called after overload
  resolution succeeded, meaning that the number of arguments must be
  correct
* don't analyze the operands again; they are already fully analyzed when
  `semOf` is called
* use `newIntTypeNode` instead of `newIntNode` + manual type assignment
* fix the folded-to-false AST being replaced with the original
  expression
* replaced the `localReport` usages for errors with error nodes; new
  diagnostic kinds and their associated logic are required
* add a documentation comment for `semOf`

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1206,6 +1206,8 @@ type
     adSemExpectedOrdinal
     adSemConstExprExpected
     adSemExpectedRangeType
+    adSemExpectedObjectForOf
+    adSemCannotBeOfSubtype
     # semobjconstr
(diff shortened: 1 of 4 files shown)


saem (approved):
Nice fix. Also the cleanups like `newIntTypeNode` make things much easier to follow.

PR #875, opened 2023-09-06
sem: correctly compute size for partial objects

zerbina:
## Summary

Treat partial objects as having an unknown size until they're completed.
This prevents their usage in compile-time contexts prior to being
completed and fixes memory corruption issues with the C target, which
were caused by the size computed for the partial object types being
wrong.

Since the JS and VM targets don't support `sizeof`/`alignof`/`offsetof`
for incomplete types, attempting to call either magic for an incomplete
partial object will result in compiler errors.

## Details

The size and alignment for all types is computed and set for all types
at the end of a type section, and this led to the size of the incomplete
package-level (i.e., partial) objects being set already.

When a package-level object is started (that is, first added to the
`packageTypes` list), its size and alignment is now set to 'unknown',
preventing the value from being incorrectly computed and disabling
folding of `sizeof`/`alignof`/`offsetof` magics. Once finished (i.e.,
the final record AST is analyzed), the size and alignments values are
set to 'uncomputed'.

Unfortunately, this means that other aggregate types embedding partial
objects prior to them being completed are also treated as incomplete,
and, since there's currently no way to propagate the completion of a
partial object to all its users, stay as incomplete, preventing their
use with `sizeof`/`alignof`/`offset` at compile-time and with the JS and
VM targets.

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -1807,6 +1807,10 @@ proc typeDefLeftSidePass(c: PContext, typeSection: PNode, i: int) =
         s = semIdentDef(c, name[1], skType)
         s.typ = newTypeS(tyObject, c)
         s.typ.sym = s
+        # until the type is completed, the size and alignment are treated as
+        # unknown
+        s.typ.size = szUnknownSize
+        s.typ.align = szUnknownSize
         s.flags.incl sfForward
         c.graph.packageTypes.strTableAdd s
         addInterfaceDecl(c, s)
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -1142,6 +1142,12 @@ proc semObjectNode(c: PContext, n: PNode, prev: PType; flags: TTypeFlags): PType
   else:
     # partial object so add things to the check
     addInheritedFields(c, check, pos, result)
+    if not incompleteType(result):
+      # this record AST represents the final addition to the object; the
+      # size can be computed now
+      result.size = szUncomputedSize
+      result.align = szUncomputedSize
+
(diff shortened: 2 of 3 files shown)


saem (approved):
One thing for consideration:

I wonder if we should use a negative set of values to store partly calculated values?

Might not be worth it.

zerbina:
> I wonder if we should use a negative set of values to store partly calculated values?

Hm, could you elaborate on how you imagine this to work? I'm not sure I understand how it would help.

saem:
> > I wonder if we should use a negative set of values to store partly calculated values?
> 
> Hm, could you elaborate on how you imagine this to work? I'm not sure I understand how it would help.

It'd be a bit weird, but we'd store the size of partial objects as negative numbers (with a possible offset or using a mask). Then when calculating the size of a type we'd first check if a field is partial (negative) and if so we'd add everything (undoing the negative/mask for the summation) and store the computed size as a negative value. So a partial field with a minimum size of 32 bytes, regardless of encoding, would contribute 32 bytes of size to the containing type.

This would give us an "at least this big" size for all partial types or types containing partial types.

zerbina:
Ah, I get it, thank you for explaining.

However, I'm still not sure about what knowing the lower bound of the a type's possible size would get us. More generally, do you think it is something that should be part of this PR?

saem:
> Ah, I get it, thank you for explaining.
> 
> However, I'm still not sure about what knowing the lower bound of the a type's possible size would get us. More generally, do you think it is something that should be part of this PR?

Especially after discussing it here, I don't think it should be part of this PR, I mostly wanted to consider the idea. I think the use case will be pretty niche, something we might discover to be useful at a later date.

PR #878, opened 2023-09-07
vm: convert `gorge` and `staticExec` to vmops

saem:
## Summary

Convert `gorge` and `staticExec` in the `system` module from magics to
VM callback operations.

## Details

Removal of the magic requires creating a conditional compilation step
based on the `nimskullReworkStaticExec` conditional symbol (see:
`condsym`). Two definitions of `gorge` and `staticExec` now exist, one
with the magic and one without. Without this, bootstrapping would no
longer work.

Also marked `gorge` as deprecated, `staticExec` is a far better name,
and noted that `staticExec` will have its return type revised to return
the exit code in addition to the output. The reason for the deprecation
is that `gorge` is just not as clear as `staticExec`, same for `slurp`.

This change enables future removal of the `mStaticExec` magic enum, and
more importantly the `opcGorge` VM opcode in the near future.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -742,7 +742,9 @@ type
     mDefined, mDeclared, mDeclaredInScope, mCompiles, mArrGet, mArrPut, mAsgn,
     mLow, mHigh, mSizeOf, mAlignOf, mOffsetOf, mTypeTrait,
     mIs, mOf, mAddr, mType, mTypeOf,
-    mPlugin, mEcho, mShallowCopy, mSlurp, mStaticExec, mStatic,
+    mPlugin, mEcho, mShallowCopy, mSlurp,
+    mStaticExec ## deprecated, remove me
+    mStatic,
     mParseExprToAst, mParseStmtToAst, mExpandToAst, mQuoteAst,
     mInc, mDec, mOrd,
     mNew, mNewSeq, mNewSeqOfCap,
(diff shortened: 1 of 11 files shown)


saem:
going to revise the approach to use a registered procedure approach for gorge

saem:
Marked ready for review to run it through CI, still need to clean-up the PR message.

zerbina:
If you turn `slurp` into a template that just calls `staticExec`, then you don't need the special handling in `gorgeOps`.

Also, what's the reason behind deprecating the `slurp` and `gorge` aliases?

saem:
> Also, what's the reason behind deprecating the `slurp` and `gorge` aliases.

Added to the PR body, they're less clear as to what they do and what they don't (only work in static contexts), these very general names are better reserved for other parts of the stdlib.

saem:
> If you turn `slurp` into a template that just calls `staticExec`, then you don't need the special handling in `gorgeOps`.

Did you mean `gorge` instead of `slurp`? I'm guessing so, but perhaps I'm missing something.

zerbina:
> Did you mean `gorge` instead of `slurp`? I'm guessing so, but perhaps I'm missing something.

Yep, I did mean to say `gorge`.

saem:
> > Did you mean `gorge` instead of `slurp`? I'm guessing so, but perhaps I'm missing something.
> 
> Yep, I did mean to say `gorge`.

I thought about slimming it down but decided to keep the change "smaller" in this regard because of the deprecation and eventual removal.

zerbina:
I've updated the PR message with two small fixes.

saem:
@zerbina this change good to go in your mind?

zerbina (approved):
Yep, I think it's good to go. A very minor request of mine would be to revert the formatting changes in `vmdef`, but feel free to not.

Besides `gorge` moving away from first-class VM support, which is very nice in itself, this also addresses the confusion that could come from `gorge` using a dedicated opcode while `gorgeEx` uses a VM op, despite both having more or less the same implementation.

saem:
Undid the `vmdef` formatting change.

PR #881, opened 2023-09-08
nimsuggest: remove `v1` implementation

bung87:
## Summary

Remove the implementation of the legacy v1 `nimsuggest` version. This
simplifies the code and thus helps with restructuring/refactoring
`nimsuggest`, which is in the process of moving to the language-server
protocol (LSP).

The `--v1` CLI option is removed from `nimsuggest`, making the `--v2`
option, which is kept for backwards compatibility, the only remaining
version option.

## Details

Since there's only a single version left, the fields for tracking the
selected version are obsolete. `suggestVersion` having the value 0 meant
that the `v2` version is selected, while a value of 1 meant that
the now-removed `v1` version is selected.

* remove the `suggestVersion` field and all its usages
* remove the `version` field from the `Suggest` type
* remove the `findUsages` procedure, which was only used with version
  `v1`

### DIFF
--- a/compiler/tools/suggest.nim
+++ b/compiler/tools/suggest.nim
@@ -389,7 +389,7 @@ proc suggestFieldAccess(c: PContext, n, field: PNode, outputs: var Suggestions) 
   var typ = n.typ
   var pm: PrefixMatch
   when defined(nimsuggest):
-    if n.kind == nkSym and n.sym.kind == skError and c.config.suggestVersion == 0:
+    if n.kind == nkSym and n.sym.kind == skError:
       # consider 'foo.|' where 'foo' is some not imported module.
       let fullPath = findModule(c.config, n.sym.name.s, toFullPath(c.config, n.info))
       if fullPath.isEmpty:
@@ -482,16 +482,6 @@ when defined(nimsuggest):
       if infoB.infoToInt == infoAsInt: return
     s.allUsages.add(info)
 
-proc findUsages(g: ModuleGraph; info: TLineInfo; s: PSym; usageSym: var PSym) =
-  if g.config.suggestVersion == 1:
-    if usageSym == nil and isTracked(info, g.config.m.trackPos, s.name.s.len):
-      usageSym = s
-      suggestResult(g.config, symToSuggest(g, s, isLocal=false, ideUse, info, 100, PrefixMatch.None, false, 0))
-    elif s == usageSym:
-      if g.config.lastLineInfo != info:
-        suggestResult(g.config, symToSuggest(g, s, isLocal=false, ideUse, info, 100, PrefixMatch.None, false, 0))
-      g.config.lastLineInfo = info
-
 when defined(nimsuggest):
   proc listUsages*(g: ModuleGraph; s: PSym) =
     #echo "usages ", s.allUsages.len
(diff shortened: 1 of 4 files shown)


compiler/tools/suggest.nim
```diff
@@ -518,20 +508,13 @@ proc suggestSym*(g: ModuleGraph; info: TLineInfo; s: PSym; usageSym: var PSym; i
...
 
-    if conf.ideCmd == ideUse:
-      findUsages(g, info, s, usageSym)
-    elif conf.ideCmd == ideDef:
+    if conf.ideCmd == ideDef:
       findDefinition(g, info, s, usageSym)
-    elif conf.ideCmd == ideDus and s != nil:
-      if isTracked(info, conf.m.trackPos, s.name.s.len):
-        suggestResult(conf, symToSuggest(g, s, isLocal=false, ideDef, info, 100, PrefixMatch.None, false, 0))
-      findUsages(g, info, s, usageSym)
```
zerbina: Why did you remove this as a whole? Only the `findUsage` call is obsolete, as far as I can tell.

zerbina (changes requested):
Less baggage is a good thing, thank you!

Since the general direction is to move away from the custom suggestion server, I think `suggestVersion` should be removed too, as part of this PR.

---

As for the PR message, please detail what "removing backwards compatibility" means in practice, both for the user and internally in the compiler and `nimsuggest`.

compiler/tools/suggest.nim
bung87: `dus` is alias for `use`, in `nimlsp` it never been called
```nim
if conf.ideCmd in {ideUse, ideDus}:
    let u = if conf.suggestVersion != 1: graph.symFromInfo(conf.m.trackPos) else: graph.usageSym
    if u != nil:
      listUsages(graph, u)
```

zerbina: While `dus` is not an alias for `use`, they indeed perform effectively the same thing in the end. Since it is related, I would suggest removing `ideDus` as part of this PR.

It's not immediately obvious from the code, so please also document in the PR message why `dus` is obsolete and subsumed by `use`.

bung87: I suggest preserve `ideDus`, it covered by `ideUse` just because `ideUse` also list definition

zerbina: I'm aware, but with the removal of the lines I asked about, you removed the only thing that made `ideDus` slightly different compared to `ideUse`. Removing `ideDus` (but not the `dus` command) would make no difference now.

bung87: "Removing ideDus (but not the dus command) " I don't understand, there's no `dus` command

zerbina: There is, see the parsing logic within `nimsuggest.execCmd`.

bung87: I prefer remain these, `ideUse` will be conditional list definition in future.

zerbina: Well, okay, but then what is the point of removing `ideDus` handling?

Also, do you have your plans regarding `nimsuggest` written down somewhere public? If not, please do so.

bung87: remove `ideDus` handling because covered by `listUsages`, where the place to write down the plan?

zerbina: Not entirely, the extra `suggestResult` call is not covered by `listUsages`. Sure, it results in a duplicate 'def' being reported, as `listUsages` already reports definitions, but still.

For writing down the plan, I think a GitHub discussion makes sense.

bung87: okay, I revert `ideDus` handle, it's fine without this change in this PR.

nimsuggest/nimsuggest.nim
```diff
@@ -659,8 +654,6 @@ proc processCmdLine*(pass: TCmdLinePass, argv: openArray[string]; conf: ConfigRe
         gMode = mepc
         conf.verbosity = compVerbosityMin  # Port number gotta be first.
       of "debug": conf.incl optIdeDebug
-      of "v2": conf.suggestVersion = 0
-      of "v1": conf.suggestVersion = 1
```
zerbina, line 663: Please keep the `v2` option for now (by treating it as a no-op). Other external tools, such as Saem's extension, still use it.

zerbina (changes requested):
Apart from one issue with backwards compatibility and two leftovers that should also be removed, I think this PR is good to be merged.

---

The `version` field in the `Suggest` type also needs to be removed, and please also remove the `--v1` mention in the `nimsuggest.Usage` string constant.

zerbina (approved):
Thanks.

I've updated both the PR title and description:
- only the second version remaining is an important piece of information, but it's not relevant in the title -- I've moved it to the summary
- the "suggest remove v1" wording made it sound like the removal is *suggested*, but not that the removal was related to `nimsuggest`
- the previous summary didn't mention *why* the old version is removed
- the details section was missing the removal of the `version` field

PR #882, opened 2023-09-08
stdlib: temporarily add AST pattern matching

bung87:
## Summary

Add the `ast_pattern_matching` module to the experimental standard
library. It implements pattern matching for `NimNode` AST and is copied
from https://github.com/krux02/ast-pattern-matching.

The first iteration of the language-server protocol (LSP) implementation
is going to use the pattern matching, but it's likely that the module
will be removed from the standard library again.

## Details

A copy of the `ast-pattern-matching` implementation already existed
within the `tests` directory, where it is used by the `tastpec.nim`
test. The module is moved to the `lib/experimental` directory and the
import is adjusted.


lib/experimental/ast_pattern_matching.nim
zerbina, line 1: Suggestion for a new comment:
```nim 
## Implements pattern matching for ``NimNode`` AST. This module is copied
## from https://github.com/krux02/ast-pattern-matching.
```

zerbina:
I've update the PR message and title. The issues with the previous one were as follows:
- the summary was missing a short description of what the module does and why it's added to the experimental standard library
- the reason provided in the details section was not something concrete ("it may be useful")

Since it's intended for use by the LSP during the latter's development, I've explicitly mentioned the addition's uncertain future.

zerbina (approved):

PR #883, opened 2023-09-09
internal: remove `msgs.addSourceLine`

bung87:
Inline the private `msgs.addSourceLine` procedure at the single
place where it is used. If a procedure is very short, not exported,
undocumented, and only used a single time, it is better to not have
it in the first place.


zerbina (approved):
Looks good to me! This also removes a unnecessary copy of the line (the `line` parameter was missing a `sink` annotation).

I've shortened the title a bit (try to keep it below 50 characters) and added some elaboration on the "why". If a change is very small in impact (for example, because it's a tiny clean up), you don't have to include the `Summary` and `Detail` headers.

PR #886, opened 2023-09-09
FFI: disallow interface specification on locals

zerbina:
## Summary

Change the language specification and implementation to disallow using
the `.nodecl`, `.header`, `.importc`, `.importjs`, `.exportc`,
`.extern`, and `.dynlib` pragmas on local 'var'/'let' definitions --
locals exist only inside routines and thus cannot be foreign.

## Details

Except for parameters, which can be detected through the `skParam` kind,
locals cannot be distinguished from globals through just the symbol
kind. Therefore, they need to be detected through dedicated logic that
inspects the full symbol.

This check is implemented by the new `isLocal` procedure added to the
`pragmas` module. The `sfGlobal` flag cannot be used for detecting
globals at this point, as the flag is only included on the symbols
for declarations outside of routines *after* pragma processing is
finished. Instead, `isLocal` checks for whether the owner of the symbol
is a module -- nonetheless testing for the `sfGlobal` flag ensures that
declarations with `.global` can also use the FFI.

During application of the FFI pragmas, it is now first checked whether
the applied-to symbol is that of a local, and if it is, an error is
returned. The error uses the new `adSemExternalLocalNotAllowed`
diagnostic plus associated report.

Finally, the documentation for the pragmas is updated.

### Code-generator cleanup

- the name mangling in the C code generator is updated to not inspect
  the `extname` for parameters and locals -- the string is now always
  empty for those
- testing for the `sfImportc` and `extNoDecl` flag for locals is removed
  from the JS code generator -- the flags are never present on the
  symbol now

### Other adjustments

The only place in the compiler where using the FFI with locals was
relied on is with the `asyncjs` module. There, `.importjs` is used for
getting access to JavaScript's `undefined` value. However, using a
local is not necessary for achieving this -- a `proc` using `.importjs`
works as well.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1822,6 +1822,10 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemIllegalCustomPragma:
       result = "cannot attach a custom pragma to '$1'" % r.symstr
 
+    of rsemExternalIfaceNotAllowed:
+      result = "an external interface cannot be specified on a parameter or" &
+               " local 'var'/'let'"
+
     of rsemCallingConventionMismatch:
       assert false, "REMOVE"
 
@@ -3290,7 +3294,8 @@ func astDiagToLegacyReport(conf: ConfigRef, diag: PAstDiag): Report {.inline.} =
       adSemExpectedRangeType,
       adSemExpectedLabel,
       adSemContinueCannotHaveLabel,
-      adSemUnavailableLocation:
+      adSemUnavailableLocation,
+      adSemExternalIfaceNotAllowed:
     semRep = SemReport(
         location: some diag.location,
         reportInst: diag.instLoc.toReportLineInfo,
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1822,9 +1822,9 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemIllegalCustomPragma:
       result = "cannot attach a custom pragma to '$1'" % r.symstr
 
-    of rsemExternalIfaceNotAllowed:
-      result = "an external interface cannot be specified on a parameter or" &
-               " local 'var'/'let'"
+    of rsemExternalLocalNotAllowed:
+      result = "parameters and local 'let'/'var' cannot be part of an" &
+               " external interface"
 
     of rsemCallingConventionMismatch:
       assert false, "REMOVE"
@@ -3295,7 +3295,7 @@ func astDiagToLegacyReport(conf: ConfigRef, diag: PAstDiag): Report {.inline.} =
       adSemExpectedLabel,
       adSemContinueCannotHaveLabel,
       adSemUnavailableLocation,
-      adSemExternalIfaceNotAllowed:
+      adSemExternalLocalNotAllowed:
     semRep = SemReport(
         location: some diag.location,
         reportInst: diag.instLoc.toReportLineInfo,
--- a/compiler/sem/pragmas.nim
+++ b/compiler/sem/pragmas.nim
@@ -160,6 +160,16 @@ proc illegalCustomPragma*(c: PContext; n: PNode, s: PSym): PNode =
     n,
     PAstDiag(kind: adSemIllegalCustomPragma, customPragma: s))
 
+func isLocal(s: PSym): bool =
+  ## Returns whether `s` is a parameter or local `var`/`let`.
+  # XXX: ideally, we could rely on ``sfGlobal``, but for declarations
+  #      outside of routines, the flag is currently only included after pragma
+  #      processing
+  s.kind in skLocalVars and sfGlobal notin s.flags and s.owner.kind != skModule
+
+proc disallowedExternalIfaceSpec(c: PContext, n: PNode): PNode =
+  c.config.newError(n, PAstDiag(kind: adSemExternalIfaceNotAllowed))
+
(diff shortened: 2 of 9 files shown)


mrgaturus:
What about naming error with `ExternalSym` instead `ExternalIface`?, it looks more specific because external interfaces are applied to symbols

zerbina:
> What about naming error with `ExternalSym` instead `ExternalIface`?, it looks more specific because external interfaces are applied to symbols

I'm... not sure I understand. My thinking was that the error is about an interface specification being applied where it must not, so `ExternalI(nter)face` made sense to me.

Do you mean that the report name should include a mention of locals?

mrgaturus:
> Do you mean that the report name should include a mention of locals?

yes, now that you mention it, i think naming as `adSemExternalLocalNotAllowed` and `disallowedExternalLocalSpec` is also more specific with report message

saem:
In the PR message body, should the following:

> the flags are never now never present on the symbol

read as:

> the flags are always present on the symbol

compiler/front/cli_reporter.nim
```diff
@@ -1822,6 +1822,10 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemIllegalCustomPragma:
       result = "cannot attach a custom pragma to '$1'" % r.symstr
 
+    of rsemExternalIfaceNotAllowed:
+      result = "an external interface cannot be specified on a parameter or" &
+               " local 'var'/'let'"
```
saem: Would a message closer to this be a bit more actionable?

> parameters or local 'let'/'var' cannot be a part of an external interface

compiler/sem/pragmas.nim
```diff
@@ -160,6 +160,16 @@ proc illegalCustomPragma*(c: PContext; n: PNode, s: PSym): PNode =
     n,
     PAstDiag(kind: adSemIllegalCustomPragma, customPragma: s))
 
+func isLocal(s: PSym): bool =
+  ## Returns whether `s` is a parameter or local `var`/`let`.
+  # XXX: ideally, we could rely on ``sfGlobal``, but for declarations
```
saem, line 165: instead of "global" are we really after top level module scope, barring forcing global via the pragma?

asking because I can think of a pretty quick PR I can bang out to improve things.

zerbina:
> In the PR message body, should the following:
> 
> > the flags are never now never present on the symbol
> 
> read as:
> 
> > the flags are always present on the symbol

Ah, no, it should read as "they're never present on the symbol now" -- I've updated it.

saem (approved):
this is good stuff, one more thing that's correct by design, rather than having weird gotchas/sharp edges in the language. 🚀

compiler/front/cli_reporter.nim
zerbina: The whole terminology around the foreign-function-interface seems quite inconsistent too me, so I'm not sure.

For example, I'm not sure what is (or should be) considered an "external interface". Is it meant in the global sense, applying to the whole program? Or is it more meant in the local sense, only applying to symbols (the definition I went with when writing the error message)?

compiler/sem/pragmas.nim
zerbina: I'm not entirely sure. "globals" are a bit fuzzy at the moment, as `var`/`let` definitions placed outside of procedures but *not* at the top-level module scope are also treated as globals (see `semIdentDef`). That is:

```nim
# module scope
block:
  var a = 0 # <- currently treated as a global
```

Aside: I think only definitions in the top-level module scope should be considered "globals". This would significantly simplify the backends, but it would also be a significant language change requiring large adjustments to the test suite.

compiler/front/cli_reporter.nim
saem: I've been personally avoiding the module/compilation unit question(s) because I didn't feel like I've built up enough experience to formulate a passable answer. So yeah, it's also ambiguous for me whether we're talking about things on a per module basis or whole program (we do unity builds after all).

I think there is a bias towards it being on a module by module basis (compilation unit), so we can go with that for now until we have a more reasoned answer.

compiler/sem/pragmas.nim
saem: I'm with you on this one, if it's not top level scope it's not global to the module, at least that's where my instincts are at.

Looking at it from the other point of view:

```nim
# module scope
block:
  var a = 0 # should this be global or not
```

Here the location backing `a` the handle is global, while the handle is short lived (scope of the `block`). I guess the question is should we allow for this sort of thing? Peeling back some more layers if we think about a module as a struct + init function, then a global is any field in the struct, while a non-global is a local to the init function body. Going by that, the code in the `block` in the above example seems like something that would occur in the body and we shouldn't lift `a` into the module struct.

I can probably go around in circles for a bit, but I think what's really happening with global is we're decided whether we want to lift it into the struct (location) or not and what level of control do we want to give the user over all that. Layering FFI atop that is where things get a bit messy as the location is valid but depending upon timing the handle isn't and that seems like a bunch of UB lying in wait. 🤔

Hopefully the above isn't all just blindingly obvious. One thing that seems clear is that this might require a fair bit more discussion and any changes are all outside of this PR. 😆

compiler/front/cli_reporter.nim
zerbina: Using "external interface" to refer to the external interface of the module, I think your suggested message fits better, so I replaced the previous one with it.

zerbina:
/merge

Thank you for the review, @saem, and thank you @mrgaturus for suggesting a name change -- the previous one was indeed a bit too broad.

PR #888, opened 2023-09-10
pragmas: remove obsolete built-in pragmas

bung87:
## Summary

Remove the built-in no-op pragmas `delegator`, `deadCodeElim`,
`unchecked`, and `merge`. They have been no-ops for a long time, and
removing them means that their names are now available to be used by
user-defined pragmas.

## Details

* remove usage of the no-op `merge` pragma from `system.&`
* remove the pragmas' corresponding `TSpecialWord` enum value:
  `wDelegator`, `wDeadCodeElimUnused`, `wUnchecked`, `wMerge`

### DIFF
--- a/compiler/ast/wordrecg.nim
+++ b/compiler/ast/wordrecg.nim
@@ -38,14 +38,14 @@ type
     wUncheckedAssign = "uncheckedAssign",
 
     wImmediate = "immediate", wConstructor = "constructor", wDestructor = "destructor",
-    wDelegator = "delegator", wOverride = "override", wImportCompilerProc = "importCompilerProc",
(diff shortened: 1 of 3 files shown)


zerbina:
The clean up looks good, thank you.

---

I've updated the commit message and title.

Please make sure that the message describes the full impact. Removing the `TSpecialWord` enum values, while maybe the main intention, is not only an internal change, it also means that the corresponding built-in pragmas (`.merge`, `.deadCodeElim`, and `.delegator`) are no longer available, which in turn means that their names can now be used by user-defined pragmas.

zerbina (approved):

PR #879, opened 2023-09-07
tool: LSP support

bung87:
## Summary
* LSP support

## Details

---

## Notes for Reviewers
* get some improvements idea in my mind but first release initial version

<!--
Pull Request(PR) Help

Before Merge Ensure:
* title reads like a short changelog line entry
* code includes tests and is documented
* leave the source better than before, but split out big reformats

See contributor (guide)[https://nim-works.github.io/nimskull/contributing.html]
for details, especially if you're new to this project.

Tips that make PRs easier:
* for big/impactful changes, start with chat/discussions to refine ideas
* refine the pull request message over time; don't have to nail it in one go

PR #892, opened 2023-09-12
nimsuggest: move library code into `suggest`

bung87:
## Summary

Extract procedures that do not rely on the command-line from the
`nimsuggest` module to the `suggest` module.

## Details

Changes:

- procedure `findNode` was moved and renamed to `findTrackedNode`
- procedure `symFromInfo` was moved and renamed to `findTrackedSym`
- the ide command execution part of `executeNoHooks` was extracted and
  moved into `executeCmd`

This makes the `suggest` module more of a reusable component for other
non-`nimsuggest` tools, e.g. the LSP server.

### DIFF
--- a/compiler/tools/suggest.nim
+++ b/compiler/tools/suggest.nim
@@ -489,8 +489,8 @@ proc executeCmd*(cmd: IdeCmd, file, dirtyfile: AbsoluteFile, line, col: int;
   var isKnownFile = true
   let dirtyIdx = fileInfoIdx(conf, file, isKnownFile)
 
-  if not dirtyfile.isEmpty: msgs.setDirtyFile(conf, dirtyIdx, dirtyfile)
-  else: msgs.setDirtyFile(conf, dirtyIdx, AbsoluteFile"")
+  if dirtyfile.isEmpty: msgs.setDirtyFile(conf, dirtyIdx, AbsoluteFile"")
+  else: msgs.setDirtyFile(conf, dirtyIdx, dirtyfile)
 
   conf.m.trackPos = newLineInfo(dirtyIdx, line, col)
   conf.m.trackPosAttached = false
--- a/compiler/tools/suggest.nim
+++ b/compiler/tools/suggest.nim
@@ -479,6 +479,7 @@ proc findTrackedNode(n: PNode; trackPos: TLineInfo): PSym =
 proc findTrackedSym*(g: ModuleGraph;): PSym =
   let m = g.getModule(g.config.m.trackPos.fileIndex)
   if m != nil and m.ast != nil:
+    # xxx: the node finding should be specialized per symbol kind
     result = findTrackedNode(m.ast, g.config.m.trackPos)
 
 proc executeCmd*(cmd: IdeCmd, file, dirtyfile: AbsoluteFile, line, col: int;
--- a/compiler/tools/suggest.nim
+++ b/compiler/tools/suggest.nim
@@ -56,6 +56,7 @@ import
     options,
   ],
   compiler/modules/[
+    modules,
     modulegraphs,
   ],
   compiler/sem/[
@@ -66,7 +67,8 @@ import
   compiler/utils/[
     prefixmatches,
     astrepr,
-    debugutils
+    debugutils,
+    pathutils
   ]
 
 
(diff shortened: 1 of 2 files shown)


saem:
@bung87 does this block #879? If so, can you note blockers/dependencies like that in the `Notes for Reviewers` section, thanks.

bung87:
updated.

compiler/tools/suggest.nim
```diff
@@ -466,6 +468,48 @@ proc isTracked*(current, trackPos: TLineInfo, tokenLen: int): bool =
...
+    for i in 0 ..< safeLen(n):
+      let res = findTrackedNode(n[i], trackPos)
+      if res != nil: return res
+
+proc findTrackedSym*(g: ModuleGraph;): PSym =
+  let m = g.getModule(g.config.m.trackPos.fileIndex)
+  if m != nil and m.ast != nil:
+    result = findTrackedNode(m.ast, g.config.m.trackPos)
+
+proc executeCmd*(cmd: IdeCmd, file, dirtyfile: AbsoluteFile, line, col: int;
```
saem, line 485: public procedures of modules, especially important ones like these should have a doc comment describing them.

saem:
I've done a quick read, but it looks like there was a move plus renames of 2 procedures and the extraction of another procedure.

Were there any other changes that I didn't spot? The extraction of the `execueCmd` procedure is a highlight that should be covered in the PR description.

compiler/tools/suggest.nim
```diff
@@ -466,6 +468,48 @@ proc isTracked*(current, trackPos: TLineInfo, tokenLen: int): bool =
...
+    if isTracked(n.info, trackPos, n.sym.name.s.len): return n.sym
+  else:
+    for i in 0 ..< safeLen(n):
+      let res = findTrackedNode(n[i], trackPos)
+      if res != nil: return res
+
+proc findTrackedSym*(g: ModuleGraph;): PSym =
+  let m = g.getModule(g.config.m.trackPos.fileIndex)
+  if m != nil and m.ast != nil:
+    result = findTrackedNode(m.ast, g.config.m.trackPos)
```
saem, line 483:
```suggestion
    # xxx: the node finding should be specialized per symbol kind
    result = findTrackedNode(m.ast, g.config.m.trackPos)
```

saem: noting an issue with the current code for future improvement.

compiler/tools/suggest.nim
```diff
@@ -466,6 +468,48 @@ proc isTracked*(current, trackPos: TLineInfo, tokenLen: int): bool =
...
+
+proc executeCmd*(cmd: IdeCmd, file, dirtyfile: AbsoluteFile, line, col: int;
+             graph: ModuleGraph) =
+  let conf = graph.config
+  conf.ideCmd = cmd
+  var isKnownFile = true
+  let dirtyIdx = fileInfoIdx(conf, file, isKnownFile)
+
+  if not dirtyfile.isEmpty: msgs.setDirtyFile(conf, dirtyIdx, dirtyfile)
+  else: msgs.setDirtyFile(conf, dirtyIdx, AbsoluteFile"")
```
saem:
```suggestion
  if dirtyfile.isEmpty: msgs.setDirtyFile(conf, dirtyIdx, AbsoluteFile"")
  else: msgs.setDirtyFile(conf, dirtyIdx, dirtyfile)
```

negated if/else are just awkward to read.

compiler/tools/suggest.nim
```diff
@@ -466,6 +468,48 @@ proc isTracked*(current, trackPos: TLineInfo, tokenLen: int): bool =
...
+
+  conf.m.trackPos = newLineInfo(dirtyIdx, line, col)
+  conf.m.trackPosAttached = false
+  conf.errorCounter = 0
+  if not isKnownFile:
+    graph.compileProject(dirtyIdx)
+  if conf.ideCmd in {ideUse, ideDus} and
```
saem, line 503: hmm, that's a bit weird because `ideDus` is definition + usage, IIRC, in which case definition (`ideDef`?) should be included... or `ideDus` shouldn't be included here. Probably not worth changing right now, mostly and observation

saem (changes requested):
I spotted a few things, but I really like the idea of separating these things out, so we have a general core `suggest` module and then wrap it with `nimsuggest` or lsp specific modules to specialize it.

I can help with the PR phrasing, I just need you to provide the core content (what and why).

compiler/tools/suggest.nim
bung87: yep, this piece of code is questionable, maybe other ideCmds should include too.

compiler/tools/suggest.nim
bung87: could you help me with this?

saem: Something like:

> executes the given suggest command, `cmd`, for a given `file`, at the position described by `line` and `col`umn. If `dirtyFile` is non-empty, then its contents are used as part of the analysis.

Something like that, not sure if I'm missing key bits of functionality?

saem: A doc comment for a routine should cover what it does and any pre/post conditions about the routine/its inputs.

compiler/tools/suggest.nim
```diff
@@ -484,6 +484,9 @@ proc findTrackedSym*(g: ModuleGraph;): PSym =
 
 proc executeCmd*(cmd: IdeCmd, file, dirtyfile: AbsoluteFile, line, col: int;
              graph: ModuleGraph) =
+  ## executes the given suggest command for a given file,
+  ## at the position described by line and column.
+  ## If dirtyFile is non-empty, then its contents are used as part of the analysis.
```
saem:
```suggestion
  ## executes the given suggest command, `cmd`, for a given `file`, at the
  ## position described by `line` and `col`umn. If `dirtyFile` is non-empty,
  ## then its contents are used as part of the analysis.
```

comments and lines in general should be formatted to 80 chars, and backticks should be used to identify/highlight parameter names.

bung87:
updated

saem:
> updated

Yup, I just need a some time to clean-up the PR message. In the meantime, if you can take another attempt at it based on the recent PR messages I've cleaned up, that'll save me some time, thanks.

saem:
@bung87 have a read through the PR title + body

saem (approved):

PR #894, opened 2023-09-12
remove obsolete comments related to nimsuggest

bung87:
Removed obsolete comments related to nimsuggest from the `nimsuggest`
and `suggest` modules.


saem (approved):
Yup, these are old and would misguide people.

PR #895, opened 2023-09-12
fix instantiation errors with generic `=deepCopy`

zerbina:
## Summary

Fix "unresolved generic parameter" error occurring when instantiating
generic `=deepCopy` implementations that create locals of the
instantiated type the generic operator is instantiated for.

## Details

A generic `=deepCopy` operator attached to a generic type is
instantiated when an instance of the generic type is created. At the
end of instantiating body of the operator, the second semantic pass
(`sempass2`) lifts the type-bound operators for types used in the body.

If the type the generic `=deepCopy` operator is instantiated for is used
in the body, this means that the other type-bound operators are also
lifted already. However, if one of the other type-bound operators is
generic, this leads to `liftdestructors.inst` being called, which then
reports an `rsemUnresolvedGenericParameter` error, as `typeInst` is nil.

The reason for `typeInst` being 'nil' at this point is that
`handleGenericInvocation`, from where the generic `=deepCopy` is
instantiated, only assigns the `typeInst` field *after* the hook
is instantiated.

The `typeInst` field for an instantiation is now set *before*
instantiating the generic `=deepCopy` operator, fixing the issue. In
addition, `typeInst` not being set when trying to instantiate a generic
type-bound operator is changed to an internal error (`unreachable`)
and the `rsemUnresolvedGenericParameter` report is removed.

### DIFF
--- a/tests/system/tdeepcopy.nim
+++ b/tests/system/tdeepcopy.nim
@@ -93,3 +93,27 @@ doAssert(sizeof(PtrTable) == 2*sizeof(int)+sizeof(pointer)*2*100)
 
 main()
 echo "ok"
+
+block generic_deep_copy_using_instantiated_for_type:
+  # a generic ``=deepCopy`` implementation using the type instance it is
+  # instantiated for would lead to "unresolved generic parameter" errors,
+  # when the generic type also had other generic type-bound operators
+  # attached
+  type Generic[T] = object
+    x: T
+
+  var numDestroy {.global.} = 0
+
+  proc `=destroy`[T](x: var Generic[T]) =
+    inc numDestroy
+
+  proc `=deepCopy`[T](x: ref Generic[T]): ref Generic[T] =
+    var v = Generic[T]() # create something that requires the destroy
+                         # hook to be instantiated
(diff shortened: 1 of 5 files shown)


tests/system/tdeepcopy.nim
```diff
@@ -93,3 +93,29 @@ doAssert(sizeof(PtrTable) == 2*sizeof(int)+sizeof(pointer)*2*100)
...
+
+  proc `=deepCopy`[T](x: ref Generic[T]): ref Generic[T] =
+    var v = Generic[T]() # <- forces the generic `=destroy` operator to be
+                         #    instantiated
+    result = x
+    # `v` is destroyed here, incrementing `numDestroy` by one
+
+  let v = new(Generic[int]) # <- the deep-copy operator is instantiated here
+
+  # make sure the operatore really works:
```
saem:
```suggestion
  # make sure the operator really works:
```

this is a soft suggestion, not really a big deal.

saem (approved):
Nice, one less report and the construction order makes a lot more sense.

PR #897, opened 2023-09-13
nimsuggest: make `def` work for all `import` syntaxes

bung87:
## Summary

Fix the `def` command producing no results when used with certain import
syntaxes and styles.

* multiple-module-imports: definitions are now reported for all module
  names (e.g., `z` in `import x/[y, z]`), not only the first one
* import-as: definitions are now reported for the alias name (e.g., `z`
  in `import x/[y as z]`)
* definitions are now reported for single-module imports that use the
  bracket syntax (`x/[y]`) where the module name is on a different line
  than the opening bracket

The `def` command is usually used for goto-definition support in IDEs,
meaning that goto-definition on module names in import statements now
work for the above cases.

## Details

Instead of passing the source position of the transformed node to
`onSymImport`, the source position of the original node is passed to it.
So that the original source position can be passed to `onSymImport`,
`myImportModule` and `impMod` now take an additional `info: TLineInfo`
parameter.

For `import x as y`, a `def` of `x` is only reported on `y` right now
(and nothing is reported on `x`). Eventually, a proper `def` should
be reported on either both `x` and `y` or only on `x`.


zerbina:
Could you rephrase the PR message? As it stands, it's not clear, at least to me, what "resolved" and "now works" mean in this context.

If I'm understanding correctly, this intends to fix a `nimsuggest` related issue. If so, please also add a test for the fixed issue, thanks.

bung87:
as the plan is drop the old protocol, still need write text based test case ?

zerbina:
> as the plan is drop the old protocol, still need write text based test case ?

Yep. If an issue is fixed, a test case for ensuring that this really is the case should exist. 

In addition, all `nimsuggest` tests can and should be ported once the custom protocol is removed, so this is also going to help with development of the LSP server.

bung87:
I'll write test case after some refactoring PRs get merged.

compiler/modules/importer.nim
```diff
@@ -294,7 +294,7 @@ proc transformImportAs(c: PContext; n: PNode): tuple[node: PNode, importHidden:
     ret.node = n.processPragma
   return ret
 
-proc myImportModule(c: PContext, n: var PNode, importStmtResult: PNode): PSym =
+proc myImportModule(c: PContext, n: var PNode, info: TLineInfo, importStmtResult: PNode): PSym =
```
zerbina:
```suggestion
proc myImportModule(c: PContext, n: var PNode, info: TLineInfo,
                    importStmtResult: PNode): PSym =
  ## `info` provides the source position (which may be different from the one
  ## of `n`) to use for symbol suggestions.
```
It's not immediately obvious why this procedure needs an extra `TLineInfo`, so documenting it makes sense.

nimsuggest/tests/timport_multi_modules.nim
```diff
@@ -0,0 +1,15 @@
...
+$nimsuggest --tester $file
+>def $1
+def;;skModule;;minclude_types;;*;;$file;;10;;15;;"";;100
+>def $2
+def;;skModule;;minclude_types;;*;;$file;;11;;15;;"";;100
+>def $3
+def;;skModule;;minclude_import;;*;;$file;;13;;10;;"";;100
+"""
+import fixtures/[minclude_import, minclud#[!]#e_types]
+import fixtures/[minclude_types as typ#[!]#es]
```
zerbina: It looks like suggestions for the `import x as y` syntax (without the brackets) also didn't work previously. Please add an additional test for this case here.

In addition, I don't think the 'def' should be on the alias. Rather it should be produced on the original module name (`minclude_types` in this case).

nimsuggest/tests/timport_multi_modules.nim
```diff
@@ -0,0 +1,15 @@
...
+import
```
zerbina:
```suggestion
# test with multi-line import statements:
import
```
Important requirements that are not immediately obvious need to be documented.

nimsuggest/tests/timport_multi_modules.nim
```diff
@@ -0,0 +1,15 @@
...
+  fixtures/[
+    minclu#[!]#de_import
+  ]
```
zerbina: While suggestions already worked previously for them, please also add test cases for the "from-import" and "import-except" syntax. This ensure that they still work and keep working.

nimsuggest/nimsuggest.nim
```diff
@@ -775,6 +775,9 @@ else:
     myLog(conf, "START " & conf.projectFull.string)
 
     var graph = newModuleGraph(cache, conf)
+    graph.onMarkUsed = proc (g: ModuleGraph; info: TLineInfo; s: PSym; usageSym: var PSym; isDecl: bool) =
+      suggestSym(g, info, s, usageSym, isDecl)
+    graph.onSymImport = graph.onMarkUsed # same callback
```
zerbina: What's the purpose of this change?

zerbina (changes requested):
Thank you for adding the test. I've left some request regarding improvements to it.

---

I'm asking you again to stop force pushing. If the history is very large and messy, then force-pushing *may* be considered (but it's better when the history stays clean from the get go); otherwise review just gets complicated, as all changes need to be inspected again.

zerbina:
In addition, please also update the PR message and title. For this PR, the title should focus on the feature rather than on the implementation.

nimsuggest/nimsuggest.nim
bung87: the library usage lose these callbacks since https://github.com/nim-works/nimskull/pull/363

nimsuggest/tests/timport_multi_modules.nim
bung87: well, from my point of view, it should be on both side, if only choose one side, I'll choose the alias side, let's say `import x as y`, then the code `y.Xfield` when hover on `y` should show the module document.

nimsuggest/nimsuggest.nim
zerbina: Ah, I see, thanks. However, it's not something related to the import suggestion changes, no?

bung87: yes.

nimsuggest/tests/timport_multi_modules.nim
zerbina: > let's say `import x as y`, then the code `y.Xfield` when hover on `y` should show the module document.

Yes, but I'm not talking about usage. On usage, only the alias is available and visible, but in the import statement, the original name is immediately visible.

In addition, `onSymImport`, as the name implies, is called for the symbol of the module (and not the alias) -- the source position (and thus the `def`) should reflect that.

bung87: you mean the user have to hover on the original module name not the alias? that's weird.

zerbina: I think having to hover over the original module name within the `import` statement to show the module's document comment makes sense, yes.

In any way, given that the whole importing logic needs to be reworked (probably from the ground up) anyway, I think it can stay like it is for now. However, please document the behaviour and surrounding uncertainty in the PR body.

zerbina:
I've updated the title to better showcase what I mean. Instead, if the PR is about a feature of a tool or the language, the title should focus on the user-facing impact: `onSymImport` now using the original line info only means something to someone very familiar with `nimsuggest` and the compiler.

---

Please update the commit message (PR body) with the exact list of syntaxes where no `def`s were previously reported but now are, and please also highlight that this means that "goto definition" and hover now work for imports as one would expect.

Also, the note regarding the import-as syntax should go in the commit message, not in the reviewer notes.

bung87:
updated

zerbina:
Okay, thanks.

I'm in the process of rewording the PR message. In the meantime, could you rename the test file? It's not only related to multiple-module import statements anymore (`tdef_import.nim` is a good name, I think).

zerbina (approved):
Notes on the commit message:
- user-facing impact should be documented in the summary
- implementation details should go in the details section
- the implementation details need to be described in full

Please make sure to follow these guidelines in the futures.

Once you've read the new PR message and are okay with the changes, please notify me and I'll start the merge.

bung87:
thanks! am ok with that, that's much detail without needs looking at the code.

PR #898, opened 2023-09-13
tool: nimsuggest remove `terse` flag

bung87:
## Summary

Remove `terse` flag from `nimsuggest`. It's not used by any of the
existing extensions, nor is it tested.

## Details

The option itself is rather arbitrary in what it filters and the design
is confusing in terms of determining what should and shouldn't qualify.

### DIFF
--- a/compiler/front/in_options.nim
+++ b/compiler/front/in_options.nim
@@ -39,7 +39,6 @@ type
     optEmbedOrigSrc           ## embed the original source in the generated
                               ## code also: generate header file
     optIdeDebug               ## idetools: debug mode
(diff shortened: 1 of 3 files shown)


saem:
@bung87 please see the PR, there is a part of the original body that I can't work out what you mean, can you add more info, I'll help with the grammar?

Also, note how I changed the body, including reviewer notes. These messages are not simply for the PR review now, but also for future retrospectives, someone putting together change logs, learning from the compiler etc... so we ensure they do a good job describing the goal, change, and impact.

saem:
After the PR body is updated this is good to merge, `terse` is pointless, so happy to see it go.

PR #899, opened 2023-09-13
internal: separate VM profiler data from `ConfigRef`

zerbina:
## Summary

Store the data collected by the VM profiler with the profiler itself,
instead of storing it with the global configuration state (`ConfigRef`).

This allows for moving the profiler-related data types closer to the
`vmprofiler` module, and it also shrinks the surface of `ConfigRef` and
the `options` module.

## Details

* move all profiler-related data types to `vmdef`; they should ideally
  be placed in the `vmprofiler` module, but that would currently
  introduce cyclic dependencies
* `vmprofiler.dump` now accepts a `Profiler` instance; for rendering the
  data associated with a VM context, the new
  `compilerbridge.dumpVmProfilerData` procedure is now used

### DIFF
--- a/compiler/ast/lineinfos.nim
+++ b/compiler/ast/lineinfos.nim
@@ -26,12 +26,6 @@ type
     col*: int16
     fileIndex*: FileIndex
 
-  SourceLinePosition* = tuple
-    ## Identifies a line in a source file. Only intended for use by
-    ## the profiler.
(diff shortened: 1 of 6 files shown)


saem (approved):
Heck yeah, more modular compiler, all that much easier to understand. 🎉

PR #902, opened 2023-09-15
refactoring suggest hooks, Config.writeHook default to msgs.msgWrite

bung87:
## Summary

Remove `msgs.suggestWriteln` and unnecessary callback wrapper for
`ConfigRef.writeHook` registration.

## Details 
* Remove procedure `suggestWriteln` and its call in `suggest`,
`suggestWriteln` wraps around `writelnHook`, `writelnHook` original 
purpose is write message to stdio or some like that, not for suggest result.
* `config.writeHook` assign to `msgs.msgWrite` directly in `nim.nim` and 
  `vmrunner.nim`, now not a wrapper procedure  around it.

### DIFF
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -55,10 +55,6 @@ proc flushDot*(conf: ConfigRef) =
 
 const gCmdLineInfo* = newLineInfo(commandLineIdx, 1, 1)
 
-proc suggestWriteln*(conf: ConfigRef; s: string) =
-  if eStdOut in conf.m.errorOutputs:
-    writelnHook(conf, s)
-
 proc msgQuit*(x: int8) = quit x
 proc msgQuit*(x: string) = quit x
 
--- a/compiler/nim.nim
+++ b/compiler/nim.nim
@@ -145,10 +145,7 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef, argv: openArray[string]):
 when not defined(selftest):
   var conf = newConfigRef(cli_reporter.reportHook)
   conf.astDiagToLegacyReport = cli_reporter.legacyReportBridge
-  conf.writeHook =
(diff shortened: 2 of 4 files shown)


compiler/front/msgs.nim
```diff
@@ -55,10 +55,6 @@ proc flushDot*(conf: ConfigRef) =
 
 const gCmdLineInfo* = newLineInfo(commandLineIdx, 1, 1)
 
-proc suggestWriteln*(conf: ConfigRef; s: string) =
```
saem, line 58: agreed, this proc being here seems to be odd.

compiler/nim.nim
```diff
@@ -145,10 +145,7 @@ proc handleCmdLine(cache: IdentCache; conf: ConfigRef, argv: openArray[string]):
 when not defined(selftest):
   var conf = newConfigRef(cli_reporter.reportHook)
   conf.astDiagToLegacyReport = cli_reporter.legacyReportBridge
-  conf.writeHook =
-    proc(conf: ConfigRef, msg: string, flags: MsgFlags) =
-      msgs.msgWrite(conf, msg, flags)
-
+  conf.writeHook = msgs.msgWrite
```
saem, line 148: yup, this is simpler.

compiler/tools/suggest.nim
```diff
@@ -245,8 +245,6 @@ proc `$`*(suggest: Suggest): string =
 proc suggestResult(conf: ConfigRef; s: Suggest) =
   if not isNil(conf.suggestionResultHook):
     conf.suggestionResultHook(s)
-  else:
-    conf.suggestWriteln($s)
```
saem, line 249: these likely were put here for debugging in some cases, not the end of the world if we drop them.

saem (approved):

PR #903, opened 2023-09-15
move `msgs.log` to `nimsuggest`

bung87:
Move the `log` procedure, which is only intended for use by the
`nimsuggest` tool, from the `msgs` module to `nimsuggest`.

### DIFF
--- a/nimsuggest/nimsuggest.nim
+++ b/nimsuggest/nimsuggest.nim
@@ -114,7 +114,9 @@ proc writelnToChannel(line: string) =
 proc sugResultHook(s: Suggest) =
   results.send(s)
 
-proc log*(s: string) =
+proc log(s: string) =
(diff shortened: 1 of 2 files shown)


nimsuggest/nimsuggest.nim
```diff
@@ -114,6 +114,12 @@ proc writelnToChannel(line: string) =
 proc sugResultHook(s: Suggest) =
   results.send(s)
 
+proc log*(s: string) =
```
zerbina:
```suggestion
proc log(s: string) =
  ## Appends `s` as new line to the `nimsuggest.log` file in the user's home
  ## directory. 
```
Since it's only used in this module, the procedure doesn't need to be exported.

zerbina:
@bung87: I've updated both the PR title and body.

If you're referencing module, procedure, type, etc. names (or files) in the text (including the title), please wrap them in backticks, so that the names are highlighted.

PR #905, opened 2023-09-15
sem: propagate errors through `semWhile`

zerbina:
## Summary

Enable `nkError` propagation for `semWhile` and also change the
procedure to not modify the input AST.

## Details

For automatic error propagation, `semWhile` now returns an
`ElaborateAst` instead of a raw `PNode`.


saem (approved):
`ElaborateAst` makes things write concise.

PR #906, opened 2023-09-15
internal: make NDI mappings independent from `PSym`

zerbina:
## Summary

Change how symbol-to-name mappings are stored in an `NdiFile` such
that it's not dependent on getting a `PSym` as the input. This is a
preparation for moving away from using `PSym` in the code generators.

## Details

Only the source position (`TLineInfo`) and the user-provided name
(`PIdent`) were previously taken from the `PSym`. Both are now stored in
the mapping entry directly, making it possible to register a mapping
when not having access to `PSym`.

A new and exported, but currently unused, `writeMangledName` overload is
added, which allows for passing the `TLineInfo` and `PIdent` directly.

To aid with debugging, an additional ID value is stored in the mapping
entry (but it's not written to the `.ndi` file). This value is meant for
tracing the entry back to originating symbol.

### DIFF
--- a/compiler/ast/ndi.nim
+++ b/compiler/ast/ndi.nim
@@ -20,7 +20,6 @@ import
     options
   ],
   compiler/utils/[
-    ropes,
     pathutils
   ]
 
--- a/compiler/ast/ndi.nim
+++ b/compiler/ast/ndi.nim
@@ -12,7 +12,8 @@
 
 import
   compiler/ast/[
-    ast
+    ast,
+    lineinfos
   ],
   compiler/front/[
     msgs,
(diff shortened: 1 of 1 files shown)


compiler/ast/ndi.nim
```diff
@@ -24,24 +25,38 @@ import
   ]
 
 type
+  Mapping = tuple
+    info: TLineInfo ## the source code position where the symbol is defined
+    orig: PIdent ## the user-defined name
+    name: string ## the mangled name
```
saem, line 30: Would it be worth storing the symbol I'd as a last ditch but of information for tracing issues?

saem (approved):
Very mild suggestion about storing the symbol id, if available, in the mapping. This is a good change!

compiler/ast/ndi.nim
zerbina: Hm, I'm not sure. I think issues in this area are unlikely, but storing an additional ID here is rather cheap, so I'm going to add it.

PR #907, opened 2023-09-16
internal: reorder `ParsedNodeKind` for range checks

bung87:
## Summary

Reorder `ParsedNodeKind` for range checks, and added named sets to
support querying parsed node kinds.

## Details

Reorder `ParsedNodeKind` in order to support ranges over similar kinds.
This is considered experimental as the motivation for the reodering came
from usage in `suggest` (a consumer). The ordering should be determined
by the `parser` (the producer). Creating these seemingly convenient
consumer/read side ranges end up quietly coupling the producer and
consumer(s). Which results in refactorings to a small collection of
parser modules spilling out to changes in various consumer locations. In
this case these might coincide with the parser's needs, as well.

Additionally, added `pnkDeclarativeDefs` and `pnkRoutineDefs` similar to
`PNode`'s kind, `TNodeKind`.

### DIFF
--- a/compiler/ast/ast_parsed_types.nim
+++ b/compiler/ast/ast_parsed_types.nim
@@ -173,8 +173,8 @@ type
     pnkIncludeStmt
     pnkExportStmt
     pnkExportExceptStmt
-    pnkConstSection
     pnkTypeSection
+    pnkConstSection
     pnkLetSection
     pnkVarSection
     pnkProcDef
--- a/compiler/ast/ast_parsed_types.nim
+++ b/compiler/ast/ast_parsed_types.nim
@@ -117,7 +117,6 @@ type
     pnkPostfix
     pnkExprEqExpr
     pnkExprColonExpr
-    pnkIdentDefs
     pnkVarTuple
     pnkPar
     pnkSqrBracket
@@ -186,6 +185,7 @@ type
     pnkTemplateDef
     pnkConstDef
     pnkTypeDef
+    pnkIdentDefs
     pnkEnumTy
     pnkEnumFieldDef
     pnkObjectTy
--- a/compiler/ast/ast_parsed_types.nim
+++ b/compiler/ast/ast_parsed_types.nim
@@ -117,7 +117,6 @@ type
     pnkPostfix
     pnkExprEqExpr
     pnkExprColonExpr
-    pnkVarTuple
     pnkPar
     pnkSqrBracket
     pnkCurly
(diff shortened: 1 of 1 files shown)


compiler/ast/ast_parsed_types.nim
```diff
@@ -175,6 +174,7 @@ type
     pnkExportStmt
     pnkExportExceptStmt
     pnkConstSection
+    pnkTypeSection
```
saem:
```suggestion
    pnkTypeSection
    pnkConstSection
```

const, let, and var sections are very similar, unlike type sections, type section shouldn't break them up.

compiler/ast/ast_parsed_types.nim
```diff
@@ -184,7 +184,7 @@ type
     pnkIteratorDef
     pnkMacroDef
     pnkTemplateDef
-    pnkTypeSection
+    pnkConstDef
```
saem, line 186: ident def and const def are _very_ similar, why move const def here and why doesn't ident def belong here?

saem (changes requested):
I'm trying to understand the reasoning for the movement, or put another way, what are the group principles being applied to come up with the new ordering.

the type section I can sort of understand (guessing), but I'm struggling to come up with the reasons for the const def move, especially when taking ident def into account. The important part isn't whether identdef should move or not, but what are the rules/principles we should be following so we can make sure things are consistent.

compiler/ast/ast_parsed_types.nim
bung87: I just switch these two, it's `pnkIdentDefs` part of section, structure similar to `xxDef`

compiler/ast/ast_parsed_types.nim
```diff
@@ -184,8 +183,9 @@ type
...
     pnkTypeDef
+    pnkIdentDefs
```
saem: This should appear before `pnkConstDef` and `pnkTypeDef`, a range for these would read awkwardly otherwise `{pnkConstDef .. pnkIdentDefs}` -- the important part is that they're all ident definitions.

saem (commented):
Generally speaking, I'm not a huge fan of reordering the enum based on the needs of a consumer, the order needs to be producer centric as the primary need is parsing.

Creating these seemingly convenient consumer/read side ranges end up quietly coupling the producer and consumer(s). Which results in refactorings to a a small collection of parser modules spilling out to changes in various consumer locations.

Consumers are better off using provided or created named sets for their querying need. I'm ok with this change going through because it doesn't trouble the parser yet, but this is something to be mindful of going forward.

compiler/ast/ast_parsed_types.nim
zerbina: `pnkVarTuple` also needs to be considered here, as it's also an identifier definition.

saem:
Updating the PR title and body.

If I have further questions, i'll ping @bung87 in chat.

PR #908, opened 2023-09-16
suggest: `symToSuggest` limit data per `IdeCmd`

bung87:
## Summary

`suggest.symToSuggest` now restrict data fetched based on the `IdeCmd`
given. Meaning some `Suggest` fields are only filled in/used given
specific `IdeCmd`(s), avoiding unnecessary processing.

## Details

* Fill `contextFits`,`globalUsages`,`localUsages` fields when `IdeCmd` in 
`{ideSug, ideCon}`. they're used in `cmpSuggestions` -> `produceOutput`.
* Fill `doc` field when `IdeCmd` is not `ideHighlight`. `ideHighlight`
  results don't render the documentation comment, see `$(Suggest)`

### DIFF
--- a/compiler/tools/suggest.nim
+++ b/compiler/tools/suggest.nim
@@ -162,15 +162,17 @@ proc symToSuggest(g: ModuleGraph; s: PSym, isLocal: bool, section: IdeCmd, info:
   result.quality = quality
   result.isGlobal = sfGlobal in s.flags
   result.prefix = prefix
-  result.contextFits = inTypeContext == (s.kind in {skType, skGenericParam})
+  if section in {ideSug, ideCon}:
+    result.contextFits = inTypeContext == (s.kind in {skType, skGenericParam})
   result.scope = scope
   result.name = addr s.name.s
   when defined(nimsuggest):
-    result.globalUsages = s.allUsages.len
-    var c = 0
-    for u in s.allUsages:
-      if u.fileIndex == info.fileIndex: inc c
-    result.localUsages = c
+    if section in {ideSug, ideCon}:
+      result.globalUsages = s.allUsages.len
+      var c = 0
+      for u in s.allUsages:
+        if u.fileIndex == info.fileIndex: inc c
+      result.localUsages = c
   result.symkind = byte s.kind
   if optIdeTerse notin g.config.globalOptions:
     result.qualifiedPath = @[]
@@ -192,7 +194,8 @@ proc symToSuggest(g: ModuleGraph; s: PSym, isLocal: bool, section: IdeCmd, info:
     else:
       result.forth = ""
     when defined(nimsuggest) and not defined(noDocgen) and not defined(leanCompiler):
-      result.doc = extractDocComment(g, s)
+      if section in {ideSug, ideCon, ideDef, ideChk}:
+        result.doc = extractDocComment(g, s)
   let infox =
     if useSuppliedInfo or section in {ideUse, ideHighlight, ideOutline}:
       info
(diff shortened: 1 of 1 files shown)


compiler/tools/suggest.nim
```diff
@@ -200,10 +203,10 @@ proc symToSuggest(g: ModuleGraph; s: PSym, isLocal: bool, section: IdeCmd, info:
...
   result.line = toLinenumber(infox)
   result.column = toColumn(infox)
-  result.tokenLen = if section != ideHighlight:
-                      s.name.s.len
-                    else:
-                      getTokenLenFromSource(g.config, s.name.s, infox)
+  if section in {ideHighlight}:
+    result.tokenLen = getTokenLenFromSource(g.config, s.name.s, infox)
+  else:
+    result.tokenLen = s.name.s.len
```
zerbina: Expression-style is generally preferred within the compiler's code-base, meaning that the way it was previously written (apart from the inverted condition) was fine.

zerbina (changes requested):
I think this change is okay, less processing is certainly nice, but I dislike that the data type (`Suggest`) doesn't reflect that not all fields are valid for each `section` value.

Eventually, `Suggest` should either be turned into a variant object or replaced with something completely different, but for now, could you document on the fields for which section the fields apply?

---

Regarding the PR message and title:
- please try to keep the PR titles short. They're used as the commit title, and should not be much longer than 50 characters
- the lines in the body must be no longer than 72 characters
- please document where you got the information of what the sets should be from. Skimming over the `symToSuggest` usage sites, `ideChk` seems to never be passed to it

compiler/tools/suggest.nim
bung87: it makes line longer and this processing may apply to other command in future.

bung87:
updated.

compiler/tools/suggest.nim
zerbina: I understand the line length concern (the `if` expression usually goes on the next line for this reason), but tentatively changing an otherwise unrelated line only adds noise and complicates traversing the git history.

zerbina:
Thank you for updating the PR body with the context regarding the sets.

The set for the `doc` field including `ideChk` is still a bit confusing, given that `symToSuggest` is never invoked with `ideChk`. However, since the whole suggestion reporting is going to be reworked anyway, I think that's fine for now.

bung87:
confirmed, `ideChk` doesn't goes to `symToSuggest`, `structuredReportHook` take the responsibility that accept `Report`, now leave it as is, as `symToSuggest` accept all commands and `Suggest` is not object variant, so take care all kinds of command is fine. I haven't trigger diagnostics so far, I'll revisit the code handle `ideChk`.

zerbina:
Apart from the style change and the PR title, this PR is good to go.

compiler/tools/suggest.nim
zerbina:
```suggestion
  result.tokenLen = if section != ideHighlight:
                      s.name.s.len
                    else:
                      getTokenLenFromSource(g.config, s.name.s, infox)
```
This should revert the style change.

compiler/tools/suggest.nim
```diff
@@ -191,7 +193,8 @@ proc symToSuggest(g: ModuleGraph; s: PSym, isLocal: bool, section: IdeCmd, info:
   else:
     result.forth = ""
   when defined(nimsuggest) and not defined(noDocgen) and not defined(leanCompiler):
-    result.doc = extractDocComment(g, s)
+    if section in {ideSug, ideCon, ideDef, ideChk}:
```
```suggestion
    if section != ideHighlight:
```

zerbina (changes requested):
Sorry for not having noticed this earlier.

While rewording the PR message and looking for a more general reason for the `doc` field change, I noticed that the `ideMod` section is not included in the set, which is a breaking change (see the `$` operator for `Suggest`).

I know that the custom protocol is going to be removed, but until it is, `nimsuggest` has to continue working.

saem:
@zerbina FYI, I just reworded the PR message, nothing major but hopefully I didn't overwrite your changes. 🤞🏽

zerbina:
/merge

I've updated the PR message according to the recent changes.

PR #910, opened 2023-09-16
ci: Remove comments from PR description before merging

Clyybber:
Remove HTML comments from the PR description before merging, as otherwise
they can end up in the final commit message.

### DIFF
--- a/.github/workflows/slash-command-merge.yml
+++ b/.github/workflows/slash-command-merge.yml
@@ -52,7 +52,7 @@ jobs:
             }
 
             let [ body, body_tail ] = pr_info.body.split(/^---\s*$/m, 2);
-            body = body.replace(/<\!--.*?-->/sg, "");
+            body = body.replace(/<\!--.*?-->/s, "");
             body = body.trimEnd();
             body_tail = body_tail?.trimEnd() || '';
             // Only runs when the PR is still open and is not staged for merging.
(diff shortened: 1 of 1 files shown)


saem (approved):
Good catch!

saem:
@alaviss pinged you on this in case there is a better way to implement it, otherwise the goal seems like the right way to go.

alaviss (commented):
This might stir up trouble once someone writes sample comments (ie. as part of a PR fixing some XML parser bug?) within a PR description.

With that said, other than moving all comments within the PR template to below the tail, I don't have any better solution.

.github/workflows/slash-command-merge.yml
```diff
@@ -52,6 +52,7 @@ jobs:
             }
 
             let [ body, body_tail ] = pr_info.body.split(/^---\s*$/m, 2);
+            body = body.replace(/<\!--.*?-->/sg, "");
```
saem:
```suggestion
            body = body.replace(/<\!--.*?-->/s, "");
```

my thinking is that if we drop the `g`, it will only replace the first occurence, avoiding the potential pitfall that @alaviss mentioned. Instead if we later add more `<!-- blah -->` sections those won't be covered, but having them in the middle is likely not the way to go anyway.

saem: @Clyybber bump

saem:
/merge

Thanks, @Clyybber

PR #913, opened 2023-09-17
vm: fix the profiler reporting wrong timings

zerbina:
The VM profiler (`--profilevm`) now properly reports accumulated time
differences again, instead of meaningless absolute times. In addition to
the fix, the table lookup within `leaveImpl` is optimized with
`mgetOrPut`.


zerbina:
This is quite the silly oversight -- apologies for me missing it.

saem (approved):
I really should have caught this in my review, apologies.

PR #915, opened 2023-09-17
sem: reject illegal `break`/`continue` statements

zerbina:
## Summary

Detect and report errors for `break` and `continue` statements where the
resulting control-flow would cross from compile-time into run-time
contexts or from one compile-time context into another.

**Breaking changes**:
- a `compiles` cannot be used within a `static` block/expression to
  detect whether a `break`/`continue` is valid outside of the block/
  expression

**Fixes:**
- the compiler crashing for illegal `continue` statements
- the compiler crashing for illegal labeled `break` statements

`return` and `yield` statements within static contexts are still not
rejected properly.

## Details

For checking whether an unlabeled `break` or `continue` statement is
valid, there are two counters in `TProcCon` that track whether analysis
happens inside a `block` or loop.

Since semantic analysis for expressions/statements intended for compile-
time execution happens with the same `TProcCon` as the rest of the
enclosing routine, `break` and `continue` statements where the target is
outside of the compile-time context are not rejected. Target-less
`nkContinueStmt` nodes then cause a crash when reaching `transf`, while
target-less unlabeled `nkBreakStmt` nodes are detected by `vmgen`, which
then raises a code generation error.

### Introduce `ExecutionCon`

Whether analysis currently happens within a loop or block is a property
of the *execution context*, which is something that is currently
represented by a `TProcCon`.

Since not only modules and routine definitions can start an execution
context, it makes sense to separate execution-context-related state from
`TProcCon`. The properties describing an *execution context* are moved
into the new type  `ExecutionCon` (the same nomenclature as for
`TProcCon` is used), which is stored in a separate stack
(`executionCons`) within `PContext`.

The moved properties include the loop counter (`nestedLoopCounter`),
block counter (`nestedBlockCounter`), case context stack
(`caseContext`), and static-context counter (`inStaticContext`). Since
an execution context is either static or not, `inStaticContext` is
turned into a flag (`ExecutionConFlag`).

There are other properties, such as `resultSym`, for which it could make
sense to make them part of the `ExecutionCon`, but this is left to future
work.

Same as with other stacks, the `ExecutionCon` stack also needs to be
saved/restored in `tryExpr`.

### Using `ExecutionCon`

To keep the previous behaviour of both modules and routine definitions
starting a new execution context, pushing and popping a `TProcCon`
(`pushProcCon`/`popProcCon`) also pushes or pops an `ExecutionCon`. 

In addition, a new execution is also pushed whenever a new compile-time
context starts:
- `semStaticStmt`; a `static: stmt` block statement
- `semStaticExpr`; a `static(expr)` expression
- `semNormalizedConst`; for a `const`'s initializer expression
- `semConstExpr`/`tryConstExpr`; all implicit constant expression,
  such as `when` conditions, `array`-type index expressions, etc. 

After the compile-time expression/statement is analyzed, the execution
context is popped from the stack again.

The context is created as a static context (`ecfStatic`) where the
`inStaticContext` counter was previously implemented. Using `ecfStatic`
in `semConstExpr` and `tryConstExpr` would make sense too, but that'd be
a more far-reaching change.

### Implicit vs. explicit contexts

A distinction is made between implicit and explicit contexts. The idea
here is that it should still be possible to use a `when compiles(...)`
for checking whether a statement/expression is valid in the surrounding
context; since a new execution is now started for constant expression
(the `when` condition), `break` or `continue` statements part of the
tested statement/expression would be treated as illegal.

Explicit execution contexts are opened where this is immediately
visible in the code:
- modules and non-template routine definitions
- `static` blocks
- coercions to compile-time evaluation (e.g., `static(...)`)

All other execution contexts are implicit.

A new rule is added to `compiles`: the AST is analyzed as if it were
part of the nearest enclosing explicit context. This keeps much of the
previous semantics intact.

For the implementation, `semCompiles` pops all `ExecutionCon` items
representing implicit contexts from the stack until the active execution
context is an explicit one. Once the `compiles` is done, the stack is
restored.

### Detecting illegal `break`/`continue`

`continue` and unlabeled `break` statements within compile-time context
and without a corresponding target are automatically detected by way of
each `ExecutionCon` having its own set of loop/block counter values.

However, checking whether a labeled `break`s is valid only takes into
account the owner of the label symbol: if it's the same as the owner of
the current `TProcCon`, the `break` is valid, otherwise it's not. Since
two executions context can be part of the same procedure definition,
labeled breaks crossing from one into aren't detected.

To be able to detect labeled `break` crossing execution contexts, an
`skLabel` symbol now stores a context ID, which is an index into the
`executionCons` stack.

On definition of a label, the label's context ID is set to the index of
the active `ExecutionCon`. When checking whether a labeled `break` is
valid, the label's context ID is compared to the active `ExecutionCon`
index: if both are the same, the `break` is valid, otherwise it's not.

### Cleanup

Since all invalid 'break' statements are rejected in sem and don't reach
`vmgen`, the `vmGenDiagCannotFindBreakTarget` error,
`adVmGenCannotFindBreakTarget` diagnostic, and
`rsemCannotFindBreakTarget` report are all obsolete and thus removed.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1629,11 +1629,8 @@ type
       bitsize*: int
       alignment*: int # for alignment
     of skLabel:
-      context*: int           ## disambiguates the context (e.g., the procedure)
-                              ## the label is part of. If the `context` value
-                              ## of a label differs from that of the current
-                              ## context, the label is not part of the current
-                              ## context
+      context*: int           ## the ID (i.e., index) of the execution context
+                              ## this label is part of
     else: nil
     magic*: TMagic
     typ*: PType
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1629,6 +1629,12 @@ type
       guard*: PSym
       bitsize*: int
       alignment*: int # for alignment
+    of skLabel:
+      context*: int           ## disambiguates the context (e.g., the procedure)
+                              ## the label is part of. If the `context` value
+                              ## of a label differs from that of the current
+                              ## context, the label is not part of the current
+                              ## context
     else: nil
     magic*: TMagic
     typ*: PType
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1009,7 +1009,7 @@ type
                                           # | TODO: these enum values duplicate
                                           # |       `VmGenDiagKind` vmgen enum
     adVmGenTooManyRegistersRequired       # |       defined in the `vmdef`
-    adVmGenCannotFindBreakTarget          # |       module. There should be a
+                                          # |       module. There should be a
     adVmGenNotUnused                      # |       way to cross-reference data
                                           # |       without introducing direct
                                           # |       import or type
@@ -1026,8 +1026,7 @@ type
 
   AstDiagVmGenError* = object
     case kind*: AstDiagVmGenKind:
-      of adVmGenTooManyRegistersRequired,
-          adVmGenCannotFindBreakTarget:
+      of adVmGenTooManyRegistersRequired:
         discard
       of adVmGenNotUnused,
           adVmGenCannotEvaluateAtComptime:
--- a/compiler/ast/report_enums.nim
+++ b/compiler/ast/report_enums.nim
@@ -225,7 +225,6 @@ type
     rvmOpcParseExpectedExpression
     rvmTooManyRegistersRequired
     rvmMissingImportcCompleteStruct
-    rvmCannotFindBreakTarget
     rvmNotUnused
     rvmUserError
     rvmTooLargetOffset
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -3003,9 +3003,6 @@ proc reportBody*(conf: ConfigRef, r: VMReport): string =
     else:
       assert false
 
-  of rvmCannotFindBreakTarget:
-    result = "VM problem: cannot find 'break' target"
-
   of rvmNotUnused:
     result = "not unused"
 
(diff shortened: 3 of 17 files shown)


zerbina:
I went through multiple iterations prior to settling with the current version.

My first idea was to (mis-)use `TProcCon` for this, by pushing a new `TProcCon` where an `ExecutionCon` is used now. There, the module symbol was used as the owner for the "wrapper" `TProcCon`, which, in addition to fixing the `break`/`continue` detection, also meant that `return` and `yield` statements within `static` expressions/statements are properly rejected.

However, the `owner` of a `TProcCon` is also used for other purposes, such as detecting illegal recursion in macros, which would have stopped working when using a different owner for the wrapper context. While that's easy to fix (by inheriting the owner of wrapper `TProcCon`s from their enclosing context), an additional problem with using `TProcCon` is that it also manages state unrelated to the execution context.

In the end, a dedicated type for managing the execution context state was the simplest and least invasive solution.

zerbina:
There's an on-going discussion regarding the `compiles` breaking change. Until that's settled, I'm changing the PR to a draft.

zerbina:
After some discussion with @saem and @Clyybber on Matrix, I've now introduced the notion of *implicit* and *explicit* execution contexts.

Prior to entering a `compiles`, all implicit contexts are popped from the stack, meaning that statement/expression passed to a `compiles` is analyzed as if part of the closest explicit context. In effect, this allows the `compiles` in the following snippet to still return 'true':

```nim
while true:
  when compiles(if true: continue):
    break
```

Clyybber (approved):
I'm not 100% sure that there won't be any surprising edge-cases we didn't think of with this new distinction of explicit contexts, but I can't think of any myself (perhaps `{.compiletime.}` procs, but those wouldn't be affected by this change anyways since breaks can't cross proc boundaries).

saem (approved):
Execution contexts seems like the right move, the removal of the VM errors have two key plus points:

- making certain invalid states unrepresentable, at least for parts of the compiler
- detecting issues earlier

Overall, this looks like a really good change.

PR #917, opened 2023-09-21
system: lower assigning-`new` directly

zerbina:
## Summary

Instead of special-casing the assigning-`new` (i.e., `new(x)`)
throughout the compiler and only turning it into an assignment during
code generation, directly turn it into an assignment within library
code. This simplifies the magic's implementation within the compiler.

**Breaking changes**:
- `new(result)` now enforces a void context, meaning that
  `(new(result); expr)` is no longer treated as an expression

## Details

* change the `new` magic into a procedure that returns a `ref`
  instead of accepting a `var` parameter
* the new `new` magic procedure is kept as a non-exported procedure.
  User code should continue to use the `new(typedesc)` procedure
* turn the `new` routine with the `var` parameter into a template
  that calls the new `new` magic and performs the assignment
* adjust `unsafeNew` in the same way that `new` is
* remove the obsolete `internalNew` procedure; `refc` support is already
  remove from the csources compiler

In order to support instantiated generic `ref` types, the assigning
`new` template cannot continue to use `var ref T`, but instead needs to
use `var T` where `T: ref`. This was previously not necessary because no
distinction between a normal `ref T` and one coming from a generic was
made at the point were the assignment was generated (i.e., during code
generation).

**Compiler changes/simplification**:
- remove `mNew` from the `FakeVarParams` set as it no longer has a `var`
  parameter
- remove the now obsolete `lowerNew` MIR pass. The pass was previously
  needed to ensure that the previous `ref` value was cleaned up, but
  the assignment rewriting automatically takes care of that now
- update the places treating `mNew` as an assignment; it no longer is
  one
- update the code generators
- since `new(result)` expands to `result = new()`, `new(result)` forces
  a void context without any special handling being required

The `system` changes are guarded with the new conditional symbol
`nimskullNoMagicNewAssign`. This is required for bootstrapping to keep
working.

### Misc

- update the `tmovebug.nim` test, which relied on `new(result)` not
  forcing a void context. The transformation that `sem` previously
  applied is now applied manually

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -126,7 +126,7 @@ const
   defaultAlignment* = -1
   defaultOffset* = -1
 
-  FakeVarParams* = {mNew, mInc, mDec, mIncl, mExcl,
+  FakeVarParams* = {mInc, mDec, mIncl, mExcl,
     mSetLengthStr, mSetLengthSeq, mAppendStrCh, mAppendStrStr, mSwap,
     mAppendSeqElem, mNewSeq, mReset, mShallowCopy, mDeepCopy, mMove,
     mWasMoved}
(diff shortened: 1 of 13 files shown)


saem (commented):
less special casing between compiler and stdlib! 🎉

PR #918, opened 2023-09-21
transf: lower all unlabeled `break` statements

zerbina:
## Summary

Fix the cases where unlabeled `break` statements were inserted by
`transf`, which means that unlabeled `break` statements now no longer
exist after the main `transf` pass. This allows for:

* removing unlabeled `break` support from the MIR and `CgNode` IR
* simplifying code generation and MIR processing
* simplifying the `closureiters` transformation

Replacing `skLabel` usage in the code generators also becomes easier.

## Details

Since they're directly associated with a `block`, labeled `break`
statements are easier to handle for analysis and code generation. For
this reason, `transf` already transforms all unlabeled `break`
statements into labeled ones, by:
- assigning labels to `block`s that don't have a user-provided one
- wrapping `while` and `for` statements with named `block`s

However, there were two occurrences where unlabeled `break` statements
(`(nkBreakStmt (nkEmpty))`) were *injected* by `transf`:
- for the exit of `for` statements of closure iterators (`liftForLoop`)
- by the `closureiters` transformation when turning `while cond` loops
  into `while true` loops (although not strictly part of `transf`)

As a consequence, unlabeled `break` reached all the way into the code
generators, requiring dedicated support by the MIR, MIR processing, and
code generation.

While `mirgen` could lower unlabeled `break` statements, the association
of a `while` to its surrounding injected `block` no longer exists, and
`mirgen` is generally not well suited for non-local (i.e., not applying
to a single node) transformations.

Instead, the occurrences where unlabeled `break` statements are inserted
are adjusted. For the `for` statement, the label of the wrapper-block is
passed to `liftForLoop`, where the injected `nkBreakStmt` can then use
it. However, for the other occurrence, a more complex solution is
needed.

Since the `while` lowering the `closureiters` pass conditionally applies
(which is where the unlabeled `break` is injected) is the same as the
one `mirgen` automatically applies when generating `repeat` statements,
the transformation from `mirgen.genWhile` is turned into a `PNode`-based
transformation that is applied in `transf.transformWhile`. Some extra
adjustments are required:
- `PNode` doesn't have a dedicated "scope" construct like the MIR does.
  It is emulated via a `nkBlockStmt`
- the closure-iterator transformation doesn't support `yield` statements
  within `if`/`elif` conditions, so directly placing the former `while`
  condition into an `if` condition would regress on `yield` support. Prior
  to  placing the expression into the `if` condition, `nkStmtListExpr`s
  are thus first unpacked into the loop body

Only `while true` loops exist after the main `transf` pass now.

```nim
  while cond:
    body
  # is transformed into:
  block label:
    while true:
      if not cond:
        break label
      block: # a separate scope is needed for the clean-up semantics
        body
```

### `closureiters` simplifications

Multiple simplifications are possible now that neither unlabeled
`break`s nor `while` loops with non-`true`-literal conditions exist
at the point when the `closureiters` pass is run.

* remove `transformBreaksAndContinuesInWhile`, which handled unlabeled
  `break`s and `continue` statements in `while` loops
* remove the now-obsolete statement-list lowering logic for
  `nkWhileStmt`s (`lowerStmtListExprs`)
* remove the `nkWhileStmt` condition handling from the to-basic-block
  transformation (`transformClosureIteratorBody`)
* remove the `nkWhileStmt` handling from preprocessing (`preprocess`)

### MIR simplifications

* remove the "unset" sentinel value support from `LabelId`; all breaks
  in the MIR are labeled now
* remove everything related to unlabeled MIR breaks from MIR processing
  (`mirexec` and `cgirgen`)

### Code generator simplifications

Each code generator required special handling for associating unlabeled
`break`s with their targeted construct (`while` or `block`), with
`cgen` and `jsgen` being affected the most. All of it is removed.

As a side effect, the output of `jsgen` for loops slightly improves: no
extra `Label:` is emitted around every `while`.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -271,7 +271,7 @@ func nextTempId(c: var TCtx): TempId =
   inc c.numTemps
 
 func nextLabel(c: var TCtx): LabelId =
-  result = LabelId(c.numLabels + 1)
+  result = LabelId(c.numLabels)
   inc c.numLabels
 
 # ----------- CodeFragment API -------------
@@ -1676,20 +1676,15 @@ proc gen(c: var TCtx, n: PNode) =
     gen(c, n.lastSon)
   of nkBreakStmt:
     var id: LabelId
-    case n[0].kind
-    of nkSym:
+    block search:
       let sym = n[0].sym
       # find the block with the matching label and use its ``LabelId``:
       for b in c.blocks.items:
         if b.label.id == sym.id:
           id = b.id
-          break
+          break search
 
-      assert id.isSome, "break target missing"
-    of nkEmpty:
-      id = NoLabel
-    else:
-      unreachable()
+      unreachable "break target missing"
 
     c.stmts.add MirNode(kind: mnkBreak, label: id)
   of nkVarSection, nkLetSection:
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1238,7 +1238,7 @@ proc genVarSection(c: var TCtx, n: PNode) =
 
 proc genWhile(c: var TCtx, n: PNode) =
   ## Generates the code for a ``nkWhile`` node.
-  assert isTrue(n[0]), "`n` wasn't properly stransformed"
+  assert isTrue(n[0]), "`n` wasn't properly transformed"
   c.stmts.subTree MirNode(kind: mnkRepeat):
     scope(c.stmts):
       c.gen(n[1])
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1240,26 +1240,11 @@ proc genVarSection(c: var TCtx, n: PNode) =
 
 
 proc genWhile(c: var TCtx, n: PNode) =
-  ## Generates the code for a ``nkWhile`` node. Because no conditional loops
-  ## exist in the MIR, the the condition is translated to a ``break`` statement
-  ## inside an ``if``
-  c.stmts.add MirNode(kind: mnkRepeat)
-
-  scope(c.stmts):
-    # don't emit a branch for ``while true: ...``
-    if not isTrue(n[0]):
-      forward(c): genx(c, n[0]) => notOp(c) # condition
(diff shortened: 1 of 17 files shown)


zerbina:
My plan is to eventually turn the `closureiters` pass into a MIR-based transformation, which, besides simplifying it and fixing the places where `yield` is not properly supported, would also allow for removing the statement-list-expression unpacking from `transf` again.

For the extraneous `block`s, I have a follow-up planned where `mirgen` will omit the `mnkBlock` (but not the `mnkScope`) for `nkBlockStmt`/`nkBlockExpr` nodes where the label doesn't have the `sfUsed` flag included. This will elide all `mnkBlock`s for blocks that are not used for control-flow.

compiler/mir/mirgen.nim
```diff
@@ -1237,27 +1237,12 @@ proc genVarSection(c: var TCtx, n: PNode) =
...
-      forward(c): genx(c, n[0]) => notOp(c) # condition
-      c.stmts.subTree MirNode(kind: mnkIf):
-        c.stmts.add MirNode(kind: mnkBreak)
-
-    # use a nested scope for the body. This is important for scope finalizers,
-    # as exiting the loop via the ``break`` emitted above (i.e. the loop
-    # condition is false) must not run the finalizers (if present) for the
-    # loop's body
+  ## Generates the code for a ``nkWhile`` node.
+  assert isTrue(n[0]), "`n` wasn't properly stransformed"
```
```suggestion
  assert isTrue(n[0]), "`n` wasn't properly transformed"
```

compiler/mir/mirgen.nim
```diff
@@ -1237,27 +1237,12 @@ proc genVarSection(c: var TCtx, n: PNode) =
...
+  c.stmts.subTree MirNode(kind: mnkRepeat):
     scope(c.stmts):
       c.gen(n[1])
 
-  c.stmts.add endNode(mnkRepeat)
```
saem, line 1259: ah, due to the simpler generation the need for bookending with the `endNode` drops too, cool.

saem (approved):
The increased regularity of the code is really good, it's silly looking at how much code was just devoted to carrying around that ambiguity.

PR #919, opened 2023-09-22
suggest: refactoring `outline` to parser based

bung87:
## Summary
* Remove `outline` processing from procedure `suggestSym`, change to
  parser based processing
* Outline is meant to `outline` the document, not compiled AST.

## Details
* Export `proc outline*(graph: ModuleGraph; fileIdx: FileIndex)` processing
  full document.
* Add  `flags*: set[SuggestFlag]` field to `Suggest`, allow mark symbol as 
  deprecated.
* Disable `nimsuggest/tests/tinclude.nim` which used for old implementation
  that require compiled included file.
* Add `nimsuggest/tests/toutline.nim` for testing new implementation.

---

## Notes for Reviewers
* part of #879 

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


zerbina:
Your PR messages are very much improving, good work!

Two comments that I have without having looked at the changes yet:
- the "tool: suggest x" title reads strange, as it sounds like something is suggested. The "suggest:" prefix is less ambiguous
- since this is a breaking change, please document the impact. Both the user-facing impact (that is, how are users of `nimsuggest` affected?) and the impact on the documentation need to be described

compiler/tools/suggest.nim
```diff
@@ -49,6 +49,10 @@ import
     types,
     typesrenderer,
     wordrecg,
+    syntaxes,
```
zerbina: For reading convenience, please keep the imports in ascending alphabetical order.

compiler/tools/suggest.nim
```diff
@@ -477,6 +481,115 @@ proc findTrackedSym*(g: ModuleGraph;): PSym =
     # xxx: the node finding should be specialized per symbol kind
     result = findTrackedNode(m.ast, g.config.m.trackPos)
 
+proc getSymNode(node: ParsedNode): ParsedNode =
+  result = node
+  if result.kind == pnkPostfix:
+    result = result[^1]
+  elif result.kind == pnkPragmaExpr:
+    result = getSymNode(result[0])
```
```suggestion
  case node.kind
  of pnkPostfix:    node[^1]
  of pnkPragmaExpr: getSymNode(node[0])
  of pnkIdent:      node
  else:             unreachable(node.kind)
```
Multiple things of note:
1. expression style is preferred
2. where possible `case` statements/expressions are preferred over `if` statements/expressions. Single line branches can be used if the bodies are short
3. `case` statements should be exhaustive whenever possible
4. procedures like this one should be very in strict in what they accept as input

Being very strict about the inputs and treating violations as hard internal errors helps detect violations of invariants early on.

compiler/tools/suggest.nim
```diff
@@ -477,6 +481,115 @@ proc findTrackedSym*(g: ModuleGraph;): PSym =
...
+
+proc pnkToSymKind(kind: ParsedNodeKind): TSymKind =
+  result = skUnknown
```
zerbina: The same applies here.

compiler/tools/suggest.nim
```diff
@@ -477,6 +481,115 @@ proc findTrackedSym*(g: ModuleGraph;): PSym =
...
+    colonDeprecated: bool
+  for s in n.sons:
+    identDeprecated = s.kind == pnkIdent and getName(s) == "deprecated"
+    colonDeprecated = s.kind == pnkExprColonExpr and getName(s[0]) == "deprecated" 
+    if identDeprecated or colonDeprecated:
+      sug.flags.incl SuggestFlag.deprecated
+
+proc parsedNodeToSugget(n: ParsedNode; originKind: ParsedNodeKind; module: string): Suggest =
+  if n.kind in {pnkError, pnkEmpty}: return
+  if n.kind notin {pnkProcDef..pnkVarTuple}: return
```
zerbina: Please use a named set here (e.g., `DefinitionNodes`). This makes it clear what the guard is about, and future additions to the `ParsedNodeKind` enums become easier as one can directly see what node kinds the set has to contain.

compiler/tools/suggest.nim
```diff
@@ -477,6 +481,115 @@ proc findTrackedSym*(g: ModuleGraph;): PSym =
...
+  new(result)
+  var token = getToken(n)
+  var name = ""
```
```suggestion
  var
    token = getToken(n)
    name  = ""
```
Please use var/let sections over single var/let sections. It's not a must, but aligning the horizontally aligning the '=' characters can help with readability.

compiler/tools/suggest.nim
```diff
@@ -477,6 +481,115 @@ proc findTrackedSym*(g: ModuleGraph;): PSym =
...
+
+  if n.kind in pnkRoutineDefs and n[pragmasPos].kind == pnkPragma:
+    processFlags(result, n[pragmasPos])
+  elif n[0].kind == pnkPragmaExpr and n[0][^1].kind == pnkPragma:
+    processFlags(result, n[0][^1])
+
+  if n.kind != pnkVarTuple:
+    var node: ParsedNode = getSymNode(n[0])
+    token = getToken(node)
```
zerbina, line 541:
```suggestion
    let node = getToken(getSymNode(n[0]))
```
Immutability is generally preferred over repeated assignments, as it makes the code easier to comprehend.

compiler/tools/suggest.nim
```diff
@@ -477,6 +481,115 @@ proc findTrackedSym*(g: ModuleGraph;): PSym =
...
+      when false:
+        if n.kind in pnkRoutineDefs and node.kind == pnkAccQuoted:
+          let identsLen = n[paramsPos].sons.len
+          for i in countup(1, identsLen - 1):
+            name.add getName(n[paramsPos][i][1])
+            if i != identsLen - 1:
+              name.add ","
+  else:
+    name.add "("
+    for c in n.sons:
```
```suggestion
    for c in n.items:
```
The `sons` terminology is phased out.

compiler/tools/suggest.nim
```diff
@@ -477,6 +481,115 @@ proc findTrackedSym*(g: ModuleGraph;): PSym =
...
+      if c.kind == pnkEmpty: break
+      name.add getName(c) & ","
+    name.add ")"
+  if name != "":
```
zerbina, line 558:
```suggestion

  if name != "":
```
Large indented blocks should be separated from following statements with an extra new line.

bung87:
the `outline` used in `nimsuggest.executeNoHooks` which is not exported procedure, I think it's fine.

compiler/tools/suggest.nim
```diff
@@ -489,7 +602,7 @@ proc executeCmd*(cmd: IdeCmd, file, dirtyfile: AbsoluteFile, line, col: int;
 
   if dirtyfile.isEmpty: msgs.setDirtyFile(conf, dirtyIdx, AbsoluteFile"")
   else: msgs.setDirtyFile(conf, dirtyIdx, dirtyfile)
-
+  if cmd == ideOutline: return
```
zerbina, line 607: I think it'd be better if the logic above is moved into a separate procedure (e.g., `markDirtyFile`), with `executeCmd` only being called if `cmd` is not `ideOutline`.

nimsuggest/nimsuggest.nim
```diff
@@ -222,6 +222,9 @@ proc executeNoHooks(cmd: IdeCmd, file, dirtyfile: AbsoluteFile, line, col: int;
       listUsages(graph, u)
     else:
       localReport(conf, conf.m.trackPos, reportSem(rsemSugNoSymbolAtPosition))
+  elif conf.ideCmd == ideOutline:
```
zerbina: Please use a `case` statement.

compiler/tools/suggest.nim
```diff
@@ -477,6 +481,115 @@ proc findTrackedSym*(g: ModuleGraph;): PSym =
...
+    while true:
+      parsedNode = parser.parseTopLevelStmt()
+      if parsedNode.kind == pnkEmpty:
+        break
+
+      if parsedNode.kind in Sections:
+        for node in parsedNode.sons:
+          suggestIt(node, parsedNode.kind)
+      else:
+        suggestIt(parsedNode, parsedNode.kind)
```
```suggestion
      case parsedNode.kind
      of pnkEmpty:
        break
      of Sections:
        for node in parsedNode.sons:
          suggestIt(node, parsedNode.kind)
      else:
        suggestIt(parsedNode, parsedNode.kind)
```

zerbina (changes requested):
I did a first pass over the changes, with the largest focus being on code style.

zerbina:
> the `outline` used in `nimsuggest.executeNoHooks` which is not exported procedure, I think it's fine.

It needs to be clear what changes. You disabled the `tinclude.nim` test, which is a direct indicator of user-facing behaviour changing significantly.

There's also missing a reason/motivation as to *why* the outline command is changed. That is, elaboration is missing on why `outline` should work on the document and not on the AST.

---

Regarding the `deprecated` detection and flag: I think it's not a good idea. Whether something is deprecated or not is, in my opinion, a property of a symbol, but no symbols exist at the stage where outlining now happens. The `deprecated` pragma can also be removed from a definition through macros.

zerbina:
Please also make sure that the code is documented. At least the routines, types, and constants should have a documentation comment attached.

bung87:
> > the `outline` used in `nimsuggest.executeNoHooks` which is not exported procedure, I think it's fine.
> 
> It needs to be clear what changes. You disabled the `tinclude.nim` test, which is a direct indicator of user-facing behaviour changing significantly.
> 
> There's also missing a reason/motivation as to _why_ the outline command is changed. That is, elaboration is missing on why `outline` should work on the document and not on the AST.
> 
> Regarding the `deprecated` detection and flag: I think it's not a good idea. Whether something is deprecated or not is, in my opinion, a property of a symbol, but no symbols exist at the stage where outlining now happens. The `deprecated` pragma can also be removed from a definition through macros.
* it's a weird idea outline the included file or expanded ast while users just want navigate between document sections
  also parser based allow get the result earlier.
* there is a concept of DocumentSymbols (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentSymbol), which contain a `tags` field.  it has `Deprecated` field

see the current LSP server screenshot, the vscode style the deprecated symbol _through line_
Screenshot 2023-09-23 at 15 40 17

zerbina:
I was rather tired yesterday, and thus my comments and review were only focused on surface-level details, apologies.

---

The direction this PR takes is, in my opinion, wrong. I agree that outlining should only report the symbols as they appear in the document (so no symbol coming from `include`s, macros, generic instantiation, etc.), but using a parser-based solution is not how this should (and can) be achieved.

After parsing, you only have *identifier*s, and trying to guess/approximate what symbols will result from them is cannot work reliably, especially with the amount of meta-programming on definitions that NimSkull allows for.

Phrased differently: I think that the AST-based processing needs to stay. In `suggestSym`, not retrieving the parent module for symbols declared within included code should be enough to achieve included symbols not being reported with the `outline` command.

> * it's a weird idea outline the included file or expanded ast while users just want navigate between document sections
>       also parser based allow get the result earlier.

I agree with the first half of your statement, but having result be produced faster is of little use when the results are incorrect.

> * there is a concept of DocumentSymbols (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentSymbol), which contain a `tags` field.  it has `Deprecated` field

I strongly agree that outlining should report whether a symbol is marked as deprecated. However, as I said, it's only possible to know which entities are deprecated once you have access to symbols (through looking for the `sfDeprecated` flag). The name `DocumentSymbols` also hints at this. 

Approximating this by looking for `deprecated` identifiers appearing within `pnkPragmaExpr` nodes is *not* a solution.

For example:
```nim
template drop() =
  discard

var def {.deprecated, drop.} = 1
```
Here, `def` is not something that should be produced by outlining.

bung87:
fair enough, even the example code rarely seen. don't know by what use case the user want to rewrite the pragms

zerbina:
> fair enough, even the example code rarely seen. don't know by what use case the user want to rewrite the pragms

In the end, it's only relevant whether something is allowed by the language. Also, while changing pragmas post-declaration might be rarely used (if ever), dropping a definition by way of a template/macro *is* a realistic use case. 

In addition, to name a few more cases, a parser-based solution cannot support:
- `when` statements
- `{.push deprecated.}`
- `{.pragma myDeprecated: deprecated.}`

PR #920, opened 2023-09-22
suggest: remove template `origModuleName`

bung87:
## Summary

Remove template `origModuleName` from `suggest`

## Details

It was introduced 9 years ago, originally a proc with special handling,
now just a field access template, which has an implementation shorter 
than its name.

### DIFF
--- a/compiler/tools/suggest.nim
+++ b/compiler/tools/suggest.nim
@@ -70,16 +70,11 @@ import
     pathutils
   ]
 
-
-
 when defined(nimsuggest):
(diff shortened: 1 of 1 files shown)


saem:
The PR body needs formatting fixed, lines should be wrapped to 72 characters, this needs fixing.

For PRs, stick with the Summary + Details format. Writing tip, write the Details section *first* with all the information and details, then write the Summary based on the Details. This makes writing all this far more easier.

PR #921, opened 2023-09-22
suggest: remove unused procs `ensureIdx/Seq[T]`

bung87:
Remove obsolete procedures `ensureIdx[T]` and `ensureSeq[T]` from the
`suggest` module. They were introduced 10 years ago, before default
initializers were implemented and are no longer used.

### DIFF
--- a/compiler/tools/suggest.nim
+++ b/compiler/tools/suggest.nim
@@ -533,12 +533,6 @@ proc findDefinition(g: ModuleGraph; info: TLineInfo; s: PSym; usageSym: var PSym
     else:
       usageSym = s
 
-proc ensureIdx[T](x: var T, y: int) =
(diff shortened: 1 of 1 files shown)


saem:
PR body lines need to be 72 chars long max.

saem:
I had to reformat the PR body, issues:

- file names, instead of module names (`suggest` instead of `suggest.nim`, the `nim` file extension will be changed in the future)
- early new line characters (well before 72 chars)
- trailing space characters

---

One thing that's not clear is "default initializers", zero based init has been around for a long time, and we don't have other sorts of initialization. Does it refer to zero based init or specifying defaults?

bung87:
yes, I refer to the `ensureSeq` problem.

PR #923, opened 2023-09-24
Linewrap PR description to 72 chars automatically

Clyybber:
## Summary
* Linewrap the PR description to 72 chars automatically when merging a
PR
* Codeblocks (```` ``` ````, `~~~` seperated or indented with 4 or more spaces, see
https://github.github.com/gfm/#indented-code-block ) are not linewrapped,
and codespans `like this` are treated as single words

## Details
* Remove the manual 72 char requirement and merge the top comment into
the comment below the section break
* Restrict the removal of html comment introduced in
https://github.com/nim-works/nimskull/commit/a862810f939f6f45ef8c54a135f39cfa060875a1
to non-codeblocks

### DIFF
--- a/.github/workflows/slash-command-merge.yml
+++ b/.github/workflows/slash-command-merge.yml
@@ -15,6 +15,10 @@ jobs:
     name: Trim description and merge
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            tools/ci_format_pr_description.js
       - id: pr-data
         name: Get PR data and trim PR body
         uses: actions/github-script@v6
@@ -51,50 +55,10 @@ jobs:
               return;
             }
 
-
-            function getSegments(rex, str) {
-              let segments = [];
-              let lastIndex = 0;
-              let match;
-              rex.lastIndex = 0; // In case there's a dangling previous search
-              while (match = rex.exec(str)) {
-                if (match.index > lastIndex) {
-                  segments.push([false, str.substring(lastIndex, match.index)]);
-                }
-                segments.push([true, match[0]]);
-                lastIndex = match.index + match[0].length;
-              }
-              if (lastIndex < str.length) {
-                segments.push([false, str.substring(lastIndex)]);
-              }
-              return segments;
-            }
-
-            const textWrapPRdescription = (text, maxLineLength) =>
-              getSegments(/(?:^ {0,3}(```|~~~)[^]*?^ {0,3}\1 *$)|^(?:\n {4,}.*)+/gm, text)
-                .map( ([isCodeBlock, segment]) =>
-                  isCodeBlock
-                  ? segment // Don't touch code blocks
-                  : segment.split('\n').map(line =>
-                      line.length <= maxLineLength
-                      ? line // Line isn't too long, nothing to do
-                      : getSegments(/(`+).*?\1/g, line)
-                          // Treat codespans as single words:
-                          .flatMap( ([isCodeSpan, segment]) => isCodeSpan ? segment : segment.split(' ') )
-                          .reduce(([newLine, lineLength], word) =>
-                            newLine
-                            // Don't touch trailing spaces:
-                            ? word.length > 0 && lineLength + 1 + word.length > maxLineLength
-                              ? [newLine + '\n' + word, word.length]
-                              : [newLine + ' ' + word, lineLength + 1 + word.length]
-                            : [newLine + word, lineLength + word.length]
-                          , ['', 0])[0]
-                    ).join('\n')
-                ).join('')
-
+            const formatPR = require('../../tools/ci_format_pr_description.js');
 
             let [ body, body_tail ] = pr_info.body.split(/^---\s*$/m, 2);
-            body = textWrapPRdescription(body, 72);
+            body = formatPR.lineWrapPRdescription(body, 72);
             body = body.trimEnd();
             body_tail = body_tail?.trimEnd() || '';
             // Only runs when the PR is still open and is not staged for merging.
--- a/.github/workflows/slash-command-merge.yml
+++ b/.github/workflows/slash-command-merge.yml
@@ -58,7 +58,7 @@ jobs:
             const formatPR = require('../../tools/ci_format_pr_description.js');
 
             let [ body, body_tail ] = pr_info.body.split(/^---\s*$/m, 2);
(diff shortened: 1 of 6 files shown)


.github/workflows/slash-command-merge.yml
```diff
@@ -51,8 +51,50 @@ jobs:
               return;
             }
 
+
+            function getSegments(rex, str) {
+              var segments = [];
+              var lastIndex = 0;
+              var match;
```
alaviss: Avoid `var` please, that stuff defines a global that is reachable from any scope.

Move the declarations outside if you need them preserved.

.github/workflows/slash-command-merge.yml
```diff
@@ -51,8 +51,50 @@ jobs:
...
+                      : getSegments(/(`+).*?\1/g, line)
+                          // Treat codespans as single words:
+                          .flatMap( ([isCodeSpan, segment]) => isCodeSpan ? segment : segment.split(' ') )
+                          .reduce(([newLine, lineLength], word) =>
+                            newLine
+                            // Don't touch trailing spaces:
+                            ? word.length > 0 && lineLength + 1 + word.length > maxLineLength
+                              ? [newLine + '\n' + word, word.length]
+                              : [newLine + ' ' + word, lineLength + 1 + word.length]
+                            : [newLine + word, lineLength + word.length]
```
alaviss: This is horrible code, sorry.

Use `filter` to remove the excessive ternary. Performance doesn't matter much over here.

alaviss (commented):
Some initial thoughts.

And I think we should move this out into its own file. I think inlining in YAML stops making sense once the script gets complex.

.github/workflows/slash-command-merge.yml
Clyybber: Lol, yeah I figured nested ternaries would be kinda hard to read. Blame it on js for not having if-expressions :p
Started out with just for loops and mostly imperative code and slowly converted it into this.
I don't see how you would replace these with filter though, maybe I'm missing something.

alaviss: Oh true, sorry.

You can write those as like actual functions instead of chaining expressions :P

saem: I like the ternaries, the shape of the code fits the logic of it. Bouncing up and down between function definitions and the logic doesn't seem like a win at all.

saem: I usually put the `?` and `:` as trailing on the line previous, so the reader knows they're entering into a ternary before leaving the current line. Helps reading flow, IMO.

tools/koch/koch.nim
```diff
@@ -24,6 +24,7 @@ import std/[json, os, strutils, parseopt, osproc, times]
   # Using `std/os` instead of `os` to fail early if config isn't set up properly.
   # If this fails with: `Error: cannot open file: std/os`, see
   # https://github.com/nim-lang/Nim/pull/14291 for explanation + how to fix.
+import ../compiler/utils/nodejs
```
Clyybber: Maybe it's better to put the tests in a `ci` category run by testament, as testament already imports compiler/utils/nodejs anyways..

.github/workflows/slash-command-merge.yml
Clyybber: Because of the nesting in this case, them trailing the previous line looks a bit confusing IMO:
```js
newLine ?
  word.length > 0 && lineLength + 1 + word.length > maxLineLength ?
    [newLine + '\n' + word, word.length] :
    [newLine + ' ' + word, lineLength + 1 + word.length] :
  [newLine + word, lineLength + word.length]
```

tools/koch/koch.nim
saem: IMO testament `ci` category would be better, it keeps `koch` that much slimmer.

tools/ci_format_pr_description.js
```diff
@@ -0,0 +1,192 @@
...
+          ? line // Line isn't too long, nothing to do
+          : getSegments(/(`+).*?\1/g, line)
+              // Treat codespans as single words:
+              .flatMap( ([isCodeSpan, segment]) => isCodeSpan ? segment : segment.split(' ') )
+              .reduce(([wrappedLine, lineLength], word) =>
+                wrappedLine
+                // Don't touch trailing spaces:
+                ? word.length > 0 && lineLength + 1 + word.length > maxLineLength
+                  ? [wrappedLine + '\n' + word, word.length]
+                  : [wrappedLine + ' ' + word, lineLength + 1 + word.length]
```
saem: did you consider assigning this to a temporary?

admittedly, it's slightly less expression-y.

saem (approved):
LGTM

My comments are non-blocking, happy to discuss further if you'd like.

zerbina:
I think the HTML comment stripping should stay in some form, or at least be reinstated again later. Without it, or something to take its place, each PR message would have to manually be inspected for possible HTML comments.

saem:
> I think the HTML comment stripping should stay in some form, or at least be reinstated again later. Without it, or something to take its place, each PR message would have to manually be inspected for possible HTML comments.

Is that blocking for this PR?

alaviss:
I think this PR is a good first step.

If we want to increase the feature set in the future, we can spin this workflow out into an action repo where we can even pull in a markdown parser to handle comments and such properly.

tools/koch/koch.nim
Clyybber: I put it into the `tests/tools` category, since the javascript file also sits under `tools`.

tools/ci_format_pr_description.js
Clyybber: Yeah, it would force one to use if statements with returns inside instead. Found it cleaner this way.

Clyybber:
I readded HTML comment removal outside of codeblocks and moved the test into the `tools` category. Should be ready to go.

tests/tools/tci_forrmat_pr_description.nim
```diff
@@ -0,0 +1,169 @@
+discard """
+description: "Test the PR description formatter"
+target: js
+"""
+
+import jsffi
```
saem:
```suggestion
import std/jsffi
```

saem: Not a hard blocker

saem (approved):
One minor suggestion, nonblocking. Otherwise it's got to go.

Clyybber:
Thanks for the reviews!

PR #925, opened 2023-09-26
sem: assign consistent `elif`/`else` node kinds

zerbina:
## Summary

Always assign the node kind for `elif`/`else` AST such that whether the
expression/statement version is used is consistent with the `if` AST. In
other words, a (valid) `nkIfStmt` node now only has `nkElifBranch`/
`nkElse` nodes as immediate sub-nodes, while a `nkIfExpr` node only has
`nkElifExpr`/`nkElseExpr` nodes as immediate sub-nodes.

This affects `typed` AST, and it also fixes an internal compiler error
with `if` statements within statement-list expressions, which was caused
by the node kind discrepancy.

## Details

* in `semIf`, transition branch nodes to their statement/expression
  version, if necessary. This makes whether the statement/expression
  version is used is consistent with `nkIfExpr`/`nkIfStmt`
* this fixes `semfold.foldInAst` (the constant-folding pass) skipping
  if-statement-within-statement-list-expressions. For these,
  `nkElifExpr` and `nkElseExpr` nodes appeared as sub-nodes to
  `nkIfStmt` AST, which lead to the do-nothing catch-all being taken
  (and folding thus being skipped)
* until the catch-all branch in `foldInAst` is removed, and in
  order to prevent silent regressions, both `nkElifExpr` and `nkElse`
  are now explicitly disallowed (`unreachable`) in `foldInAst`

### DIFF
--- a/compiler/sem/semfold.nim
+++ b/compiler/sem/semfold.nim
@@ -961,6 +961,9 @@ proc foldInAstAux(m: PSym, n: PNode, idgen: IdGenerator, g: ModuleGraph): Folded
     # don't process the symbol slots
     result = fromAst(n, n.len - 1)
     result.add foldInAstAux(m, n[^1], idgen, g)
+  of nkElifExpr, nkElseExpr:
+    # must not exist in statement AST
+    unreachable(n.kind)
   else:
     # XXX: type AST reaches here, but ideally this catch-all branch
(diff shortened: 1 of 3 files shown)


saem (approved):
Nice fix, the split in expr/stmt has some interesting edge cases.

PR #926, opened 2023-09-27
Fix merge workflow

Clyybber:
## Summary
* Fix the import path of the PR description formatter in the merge
workflow, as the path is treated as relative to the repo root and
not the workflow file 

Fixes the issue introduced in
https://github.com/nim-works/nimskull/commit/4cd20fac354d19a5f63803866234ca261c3c084c
and observed in
https://github.com/nim-works/nimskull/pull/925#issuecomment-1736448152

### DIFF
--- a/.github/workflows/slash-command-merge.yml
+++ b/.github/workflows/slash-command-merge.yml
@@ -55,7 +55,7 @@ jobs:
               return;
             }
 
-            const formatPR = require('../../tools/ci_format_pr_description.js');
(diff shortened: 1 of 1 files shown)


Clyybber:
Sorry for the trouble, this should fix the issue.
Since this has to be merged without `/merge` I cleaned
up the PR message manually.

Clyybber:
Merging as the issue blocks other PRs from being merged

PR #930, opened 2023-09-28
internal: remove obsolete `rsemWarnUnsafeCode`

bung87:
## Summary

Remove the unused `cstringCheck` procedure, together with the
`rsemWarnUnsafeCode` warning.

## Details

The warning was only produced by the `cstringCheck` procedure, to which
the only call was disabled. Checking for unsafe cstring conversions is
already performed in `sempass2.track`, so the unused `cstringCheck` can
be safely removed.


zerbina:
The removal is good, thanks.

Could you shorten the PR title and clean up the PR message a bit?

zerbina:
I've made a small change to the title, and also restructured/extended the PR body a bit.

PR #931, opened 2023-09-28
 fix(sem): crash on wrong symbol in definition slot

bung87:
## Summary

Fix the compiler either crashing or reporting no error when a symbol
of the wrong kind is placed into a definition position through a macro
or template.

## Details

When error-correcting, `newSymGNode` transitioned the symbol to the
expected kind with `transitionRoutineSymKind`. If the expected kind
is not that of a routine, this led to a `RangeDefect`, assuming that
the compiler was built with range checks.

Instead, a new symbol is now created, re-using the identifier of the
input one. Besides fixing the `RangeDefect` being raised, this also
prevents follow-up errors that may be caused by the flags, type, or
other state of the known-to-be-incorrect input symbol.

In addition, the `nkError` from `newSymGNode` was dropped in
`semIdentVis` when no export marker is present, resulting in no error
being reported there at all. The error is now `localReport`ed prior to
returning the recovery symbol.

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -431,7 +431,11 @@ proc newSymGNode*(kind: TSymKind, n: PNode, c: PContext): PNode =
         n.sym.owner = currOwner # xxx: modifying the sym owner is suss
         n
       else:
-        let recoverySym = newSym(kind, n.sym.name, nextSymId c.idgen, currOwner, info)
+        # a symbol with the wrong kind is transplanted into a definition
+        # position. Create a new symbol instead of a copy in order to prevent
+        # follow-up errors due to the symbol's flags, type, etc.
+        let recoverySym = newSym(kind, n.sym.name, nextSymId c.idgen,
+                                 currOwner, info)
         c.config.makeError(n, recoverySym, ExpectedKindMismatch)
   of nkIdent, nkAccQuoted:
     # xxx: sym choices qualify here, but shouldn't those be errors in
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -432,7 +432,7 @@ proc newSymGNode*(kind: TSymKind, n: PNode, c: PContext): PNode =
         n
       else:
         let recoverySym = copySym(n.sym, nextSymId c.idgen)
-        recoverySym.transitionRoutineSymKind(kind)
+        recoverySym.transitionSymKindCommon(kind)
         recoverySym.owner = currOwner
         c.config.makeError(n, recoverySym, ExpectedKindMismatch)
   of nkIdent, nkAccQuoted:
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -431,9 +431,7 @@ proc newSymGNode*(kind: TSymKind, n: PNode, c: PContext): PNode =
         n.sym.owner = currOwner # xxx: modifying the sym owner is suss
         n
       else:
-        let recoverySym = copySym(n.sym, nextSymId c.idgen)
-        recoverySym.transitionSymKindCommon(kind)
-        recoverySym.owner = currOwner
+        let recoverySym = newSym(kind, n.sym.name, nextSymId c.idgen, currOwner, info)
         c.config.makeError(n, recoverySym, ExpectedKindMismatch)
   of nkIdent, nkAccQuoted:
     # xxx: sym choices qualify here, but shouldn't those be errors in
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -688,7 +688,10 @@ proc semIdentVis(c: PContext, kind: TSymKind, n: PNode,
       c.config.semReportIllformedAst(
         n, "Expected two nodes for postfix expression, but found " & $n.len)
   else:
-    result = getDefNameSymOrRecover(newSymGNode(kind, n, c))
+    let sym = newSymGNode(kind, n, c)
+    if sym.kind == nkError:
+      localReport(c.config, sym)
+    result = getDefNameSymOrRecover(sym)
 
 proc semIdentWithPragma(c: PContext, kind: TSymKind, n: PNode,
                         allowed: TSymFlags): PSym =
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -690,6 +690,7 @@ proc semIdentVis(c: PContext, kind: TSymKind, n: PNode,
   else:
     let sym = newSymGNode(kind, n, c)
     if sym.kind == nkError:
+      # XXX: replace with propagating the error
       localReport(c.config, sym)
     result = getDefNameSymOrRecover(sym)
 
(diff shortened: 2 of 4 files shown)


zerbina:
Hm, I don't think introducing an unsafe procedure for arbitrary symbols kind conversion (the variant object fields are not properly copied) is a good idea. (I also think that the existing `transitionSymKindCommon` template shouldn't be exported, but that's a separate thing.)

Since `newSymGNode` copies the symbol anyway, I think a better solution would be creating a fresh symbol (via `newSym`) instead. That would discard the type, flags, and other information from the original symbol, but given that there's a mismatch between the expected and provided symbol kind (which is a rather severe error), discarding that state is likely the sanest thing to do.

compiler/sem/sem.nim
```diff
@@ -431,9 +431,7 @@ proc newSymGNode*(kind: TSymKind, n: PNode, c: PContext): PNode =
         n.sym.owner = currOwner # xxx: modifying the sym owner is suss
         n
       else:
-        let recoverySym = copySym(n.sym, nextSymId c.idgen)
-        recoverySym.transitionRoutineSymKind(kind)
-        recoverySym.owner = currOwner
+        let recoverySym = newSym(kind, n.sym.name, nextSymId c.idgen, currOwner, info)
```
```suggestion
        # a symbol with the wrong kind is transplanted into a definition
        # position. Create a new symbol instead of a copy in order to prevent
        # follow-up errors due to the symbol's flags, type, etc.
        let recoverySym = newSym(kind, n.sym.name, nextSymId c.idgen,
                                 currOwner, info)
```
1. lines should have a maximum length of 80 characters
2. the reason a new symbol is created instead of a copy is not immediately obvious -- a comment helps in making this clearer

zerbina (changes requested):
I think this is better, thanks.

Two things are still missing:
- the PR title needs to make it clear(er) that this a bugfix
- the PR summary needs to highlight what works now
- a test case is required (there already exists `tsymbol_must_match.nim` -- it can be extended)

bung87:
I updated the description, and added a test case, now compiles but the error won't show up.

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -57,3 +57,9 @@ replaceName(nskMacro):
   template t() = #[tt.Error
           ^ (SemSymbolKindMismatch)]#
     discard
+
+macro t1(): untyped =
+  let a = genSym(ident = "a")
+  result = newVarStmt(a, newLit 1)
+
+t1()
```
zerbina: This is not a proper test, as there's missing a check for the actual error. Look at the other test cases above for how the check should look like.

Other things that need to fixed/improved:
- explicitly document what the test tests for 
- don't depend on `genSym` defaulting to `nskLet`. The kind should be specified explicitly
- remove the `newLit 1` usage, it's not needed for the test and only introduces an unnecessary dependency

zerbina (changes requested):
How did you build the compiler you tested with? If it's a `-d:danger` compiler, then range checks are omitted, so no crash would appear prior to the fix.

Regarding the test, it needs to be adjusted. I've left a comment with the details.

---

As for the PR title and summary:
- for a fix, the title should focus on the language user's perspective, if possible. Right now, the focus is on implementation details
- the summary needs to describe the crash from the *user's* perspective, right now it describes it from the *compiler's* perspective. What can help here is asking yourself and trying to answer: what is it that one had to do that caused the compiler to crash?

Please adjust both accordingly, thanks.

tests/lang_callable/macros/tsymbol_must_match.nim
bung87: don't know what to say, the `transitionRoutineSymKind` introduced there without any document, now fix it will need much requirements.

bung87: hope someone can help me with this, no idea how to complete the test case like above code.

zerbina: > don't know what to say, the `transitionRoutineSymKind` introduced there without any document, now fix it will need much requirements.

I'm not sure I understand. Do you mean that you don't know how to document the test?

> hope someone can help me with this, no idea how to complete the test case like above code.

What are the things you have trouble with? I can try to help, but I'm not yet sure with what.

bung87: first I don't know how to make it produce the error then I can write the expected error in comment, second the  
`newVarStmt` need a value node, don't know how to get rid of `newLit`

bung87: so the error reported when sem routines.

zerbina: Unfortunately, this comment is not resolved yet.

A documentation comment for the test is missing (what exactly is being tested? the tests description applies, but the fact that the test is specifically for the wrong-symbol-in `var`-symbols-slot case should be highlighted).

In addition, the test as it is now no longer acts as a test for the fix, since with the previous logic in `newSymGNode`, the symbol would have been transitioned to a routine symbol, meaning that no range error would have happened.

> first I don't know how to make it produce the error then I can write the expected error in comment

The error happens when a symbol is transplanted into a symbol slot of a non-routine definition and the symbol. However, the error is indeed not reported at the moment.

> second the `newVarStmt` need a value node, don't know how to get rid of `newLit`

My original suggestion would have be to use the same approach as the tests for the routines use above, but `newLit` can also just stay for now.

zerbina:
The reason that you got no error when using `newVarStmt` is due to a bug in `semIdentVis`. There, the error is simply discarded if no export marker is used.

A separate PR is not necessary in this case, I think. You can just fix `semIdentVis` as part of this one, by `localReport`ing the error in the same way that the `nkPostfix` branch does it.

compiler/sem/semtypes.nim
```diff
@@ -688,7 +688,10 @@ proc semIdentVis(c: PContext, kind: TSymKind, n: PNode,
       c.config.semReportIllformedAst(
         n, "Expected two nodes for postfix expression, but found " & $n.len)
   else:
-    result = getDefNameSymOrRecover(newSymGNode(kind, n, c))
+    let sym = newSymGNode(kind, n, c)
+    if sym.kind == nkError:
+      localReport(c.config, sym)
```
zerbina, line 694:
```suggestion
      # XXX: replace with propagating the error
      localReport(c.config, sym)
```
The general direction is to remove `localReport`, meaning that when it is used, a comment should make it clear that this isn't "proper".

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -57,3 +57,12 @@ replaceName(nskMacro):
   template t() = #[tt.Error
           ^ (SemSymbolKindMismatch)]#
     discard
+
+macro t1(): untyped =
+  # test wrong symbol kind passed in non-routine context
+  # correctly report error
```
zerbina:
```suggestion
  # test that an error is correctly reported when inserting the wrong symbol
  # into a non-routine definition
```

bung87:
@zerbina  Thank you for review my PR with guidance and code edit suggestion, also feel free to edit the title and description, I tried you know.

zerbina:
Thank you for seeing this PR through.

> also feel free to edit the title and description, I tried you know.

Yep. I think it's visible, and I appreciate the effort.

zerbina (approved):

PR #932, opened 2023-09-30
`semFor` propagate iterator position error nodes

bung87:
## Summary

`semFor`  propagates errors that occur in the iterator position. They
are
now properly reported and `semFor` produces errors as required.

## Details

Updated `semFor` to propagate error nodes in the iterator position, as
well as appropriately wrap the semantically analysed  `nkFor`  if an
error
is present.

Added `adSemForExpectedIterator` to `AstDiagKind` to capture the error
as a node instead of using legacy reports `localReport`.


compiler/sem/semstmts.nim
```diff
@@ -1648,28 +1650,30 @@ proc semFor(c: PContext, n: PNode; flags: TExprFlags): PNode =
...
+        n[^2] = implicitIterator(c, "items", n[^2])
+        hasError = n[^2].isError
+      elif n.len == 4:
+        n[^2] = implicitIterator(c, "pairs", n[^2])
+        hasError = n[^2].isError
+      else:
+        localReport(c.config, n[^2], reportSem(rsemForExpectsIterator))
     result = semForVars(c, n, flags)
-  if result.kind == nkError:
+  if hasError or result.kind == nkError:
```
saem, line 1667: since `result` is the same as `n`, a single `var hasError = result[^2].kind == nkError` would work instead of changing each branch, this would allow for easy change of the `localReport` to an `nkError`, as well.

compiler/sem/semstmts.nim
```diff
@@ -1636,10 +1636,12 @@ proc isTrivalStmtExpr(n: PNode): bool =
   result = true
 
 proc semFor(c: PContext, n: PNode; flags: TExprFlags): PNode =
+  addInNimDebugUtils(c.config, "semFor", n, result, flags)
```
saem, line 1639: good, nice to expand coverage

saem (changes requested):
this change needs some tests.

reach out if you get stuck.

compiler/sem/semstmts.nim
bung87: see line 1657 `hasError` is used prevent `call[0]` cause `IndexDefect`

saem: Ah, good point.

tests/lang_callable/macros/tfor_ill_formed.nim
```diff
@@ -0,0 +1,25 @@
...
+tfor_ill_formed.nim(17, 14) Error: expression has no type: case true
+of true:
+  discard
+else:
+  discard
+'''
+"""
+import std/macros
+
+macro t1(): untyped =
```
saem: Can this not be a simple `for` loop rather than resorting to macros? Any `nkError` in the iterable position should do.

bung87: am not sure, here just ensure it goes to there, not supressed by other error

zerbina:
Hm, but there's no `IndexDefect` happening? When `semExprNoDeref` returns an `nkError`, `isCallExpr` will be 'false', and `call` is thus never indexed into.

I've tried with both your test case, and the reduced:
```nim
for x in doesntExist: # provokes an error
  discard
```
and neither fails.

If you're encountering an error `IndexDefect` as part of the LSP work, then it's not related to the error handling in `semFor`. Changing `var hasError = false` in `semForVars` to `var hasError = n[^2].kind == nkError` should be enough to propagate errors in the iterable expression slot.

saem:
I tried the following, which is a variant of what the macro did, and it failed appropriately without a crash:

```nim
for i in (if true: discard else: discard):
  discard
```

saem:
@bung87 do you have a stacktrace or something that shows the variant field access error?

bung87:
> Hm, but there's no `IndexDefect` happening? When `semExprNoDeref` returns an `nkError`, `isCallExpr` will be 'false', and `call` is thus never indexed into.
> 
> I've tried with both your test case, and the reduced:
> 
> ```nim
> for x in doesntExist: # provokes an error
>   discard
> ```
> 
> and neither fails.
> 
> If you're encountering an error `IndexDefect` as part of the LSP work, then it's not related to the error handling in `semFor`. Changing `var hasError = false` in `semForVars` to `var hasError = n[^2].kind == nkError` should be enough to propagate errors in the iterable expression slot.

it's in `not isCallExpr or call[0].kind != nkSym or
        call[0].sym.kind != skIterator`

zerbina:
> it's in `not isCallExpr or call[0].kind != nkSym or call[0].sym.kind != skIterator`

Yes, but if `call` is an `nkError`, then `isCallExpr` is 'false', and the entire rest of the `or` chain is short-circuited.

saem:
If it was an `nkError` then it would be a field not accessible error, if it's an index error that means it's a empty (invalid) call node. Furthermore, the `isCallExpr` would be `false`, and would not evaluate the rest of the boolean expression (McCarthy evaluation).

@bung87 can you do a `astrepr.debug` on `call` or `n` to see the shape of the ast?

bung87:
hmm, I can't reproduce but I encouter index error multiple times,  I just remind of it is a call with empty children.

bung87:
As it's not clear for now, I'll investigate later

bung87:
Found this is because LSP not properly handle global error reports come from like `checkMinSonsLen`, `checkMinSonsLen` which raise error in command line mode prevent compiler continue compiles, this may affect `nimsuggest` as well.

bung87:
now changed as a refactoring not a bugfix anymore.

tests/lang_callable/macros/tfor_ill_formed.nim
```diff
@@ -0,0 +1,19 @@
+discard """
+  description: '''
```
saem: this test is testing for statements not macros, it should be in `lang_stmts/for_stmt`

compiler/sem/semstmts.nim
```diff
@@ -1648,28 +1650,31 @@ proc semFor(c: PContext, n: PNode; flags: TExprFlags): PNode =
...
-    result = semForVars(c, n, flags)
   else:
+    if isCallExpr and isClosureIterator(call[0].typ.skipTypes(abstractInst)):
+      # first class iterator:
+      discard
+    elif not isCallExpr or call[0].kind != nkSym or
+        call[0].sym.kind != skIterator:
+      if n.len == 3:
+        n[^2] = implicitIterator(c, "items", n[^2])
+        hasError = n[^2].isError or hasError
```
saem: `hasError` is set in all three exhaustive branches, it should be possible to set it once at the end of all branches.

tests/lang_callable/tfor_ill_formed.nim
```diff
@@ -0,0 +1,19 @@
+discard """
```
saem: Why is this in `lang_callable`?

bung87: sorry, wrong copy paste file path.

compiler/sem/semstmts.nim
```diff
@@ -1648,28 +1650,29 @@ proc semFor(c: PContext, n: PNode; flags: TExprFlags): PNode =
...
-    discard # do nothing
+  if hasError or result.kind == nkError:
+    hasError = true
   elif n[^1].typ == c.enforceVoidContext:
     # propagate any enforced VoidContext:
     result.typ = c.enforceVoidContext
   elif efInTypeof in flags:
     result.typ = result.lastSon.typ
   closeScope(c)
+  if hasError: result = c.config.wrapError(result)
```
saem:
```suggestion
  if result.kind != nkError and hasError:
    result = c.config.wrapError(result)
```

otherwise we end up with unnecessary wrapping of errors.

compiler/sem/semstmts.nim
```diff
@@ -1648,28 +1650,29 @@ proc semFor(c: PContext, n: PNode; flags: TExprFlags): PNode =
...
+      elif n.len == 4:
+        n[^2] = implicitIterator(c, "pairs", n[^2])
+      else:
+        n[^2] = c.config.newError(n[^2], PAstDiag(kind: adSemForExpectedIterator))
+      hasError = n[^2].isError or hasError
     result = semForVars(c, n, flags)
-  if result.kind == nkError:
-    discard # do nothing
+  if hasError or result.kind == nkError:
+    hasError = true
```
saem: Is this right?

If `hasError` is `true`, then we really don't need to do anything, and if `result` is an error, then we only need to avoid the branches following this one? Is my understanding correct?

saem (changes requested):
found a couple items

compiler/sem/semstmts.nim
bung87: I think here is just avoiding go to `elif` branch like previous

saem: yes, but setting `hasError` when the `result` is an error (not a child node) isn't right. The convention is to set `hasError` if it has an error, not if it is one.

bung87: so just remain old code ?

compiler/front/cli_reporter.nim
```diff
@@ -3700,6 +3700,12 @@ func astDiagToLegacyReport(conf: ConfigRef, diag: PAstDiag): Report {.inline.} =
       kind: rsemExpectedOrdinal,
       ast: diag.wrongNode,
       typ: diag.nonOrdTyp)
+  of adSemForExpectedIterator:
+    semRep = SemReport(
+      location: some diag.location,
+      reportInst: diag.instLoc.toReportLineInfo,
+      kind: rsemForExpectsIterator,
+      ast: diag.wrongNode)
```
```suggestion
  of adSemForExpectedIterator:
    semRep = SemReport(
      location: some diag.location,
      reportInst: diag.instLoc.toReportLineInfo,
      kind: rsemForExpectsIterator,
      ast: diag.wrongNode)
```
This is unnecessary, as there's no special handling required for the diagnostic. Adding the `adSemForExpectedIterator` enum to the `of` branch in line 3293 is enough.

saem (approved):
The new code is better, nice.

I did have to spend a fair bit of time updating the commit message. The formatting was off and the summary contained an implementation description, instead of a description of behavioural change.

PR #933, opened 2023-09-30
fix(sem): crash due to ill-formed `if` AST

bung87:
## Summary

Fix the compiler crashing when encountering ill-formed `nkIfStmt`/
`nkIfExpr` AST and the maximum error count is larger than 1.

## Details

* in `semIf`, create an error node for the ill-formed branch instead of
  creating and reporting a `Report`
* this enables proper `nkError` propagation, which disables the second
  pass, preventing the invalid AST from being indexed into (this caused
  either an `IndexDefect` or `FieldDefect`)
* in addition, this also removes another use of immediate reporting via
  `handleReport`/`localReport`

### DIFF
--- a/tests/lang_callable/macros/tif_ill_formed.nim
+++ b/tests/lang_callable/macros/tif_ill_formed.nim
@@ -2,17 +2,16 @@ discard """
   description: '''
     Ensure that ill-formed if-statement AST created by macros is detected
   '''
-  cmd: "nim check $options $file"
+  cmd: "nim check --msgFormat:sexp --filenames=canonical --hints:off $options $file"
+  nimoutformat: sexp
   action: reject
-  nimout: '''
-tif_ill_formed.nim(13, 15) Error: illformed AST:  else: <<0th child missing for nkElseExpr >>
-'''
 """
 import std/macros
 
 macro t1(): untyped =
   result =
     nnkIfStmt.newTree(
-      newNimNode(nnkElseExpr))
+      newNimNode(nnkElseExpr)) #[tt.Error
+               ^ (SemIllformedAst)]#
 
 t1()
--- a/tests/lang_callable/macros/tif_ill_formed.nim
+++ b/tests/lang_callable/macros/tif_ill_formed.nim
@@ -12,6 +12,6 @@ macro t1(): untyped =
   result =
     nnkIfStmt.newTree(
       newNimNode(nnkElseExpr)) #[tt.Error
-               ^ illformed AST:  else: <<0th child missing for nkElseExpr >>]#
+                ^ illformed AST:  else: <<0th child missing for nkElseExpr >>]#
 
 t1()
--- a/tests/lang_callable/macros/tif_ill_formed.nim
+++ b/tests/lang_callable/macros/tif_ill_formed.nim
@@ -1,0 +1,16 @@
+discard """
+  cmd: "nim check $options $file"
+  action: reject
+  nimout: '''
+tif_ill_formed.nim(13, 15) Error: illformed AST:  else: <<0th child missing for nkElseExpr >>
+'''
+"""
+import std/macros
+
+macro t1(): untyped =
+  result = newTree(nnkIfStmt)
+  result.add(
+    newNimNode(nnkElseExpr)
+  )
+
+t1()
--- a/tests/lang_callable/macros/tif_ill_formed.nim
+++ b/tests/lang_callable/macros/tif_ill_formed.nim
@@ -8,9 +8,8 @@ tif_ill_formed.nim(13, 15) Error: illformed AST:  else: <<0th child missing for 
 import std/macros
 
 macro t1(): untyped =
-  result = newTree(nnkIfStmt)
-  result.add(
-    newNimNode(nnkElseExpr)
-  )
+  result =
+    nnkIfStmt.newTree(
+      newNimNode(nnkElseExpr))
 
 t1()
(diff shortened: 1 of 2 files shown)


zerbina (changes requested):
Thank you for fixing/improving the `nkError` propagation. Is this specific problem one that you noticed as part of the LSP-related work?  

---

As is, the change is not correct. A node must only be wrapped in a wrapper `nkError` if there's an `nkError` within, but this is not necessarily the case when `semReportIllformedAst` is called here.

To fix issue properly, you need to replace the `semReportIllformedAst` call with inserting an `nkError`, like so:

```nim
hasError = true
result[i] = c.config.newError(it, PAstDiag(kind: adSemIllformedAst))
```

In addition, could you add a test?

bung87:
yeah, recent bug fix all found during developing LSP, the server crash when outline compiler code.

tests/lang_callable/macros/tif_ill_formed.nim
```diff
@@ -0,0 +1,16 @@
...
+tif_ill_formed.nim(13, 15) Error: illformed AST:  else: <<0th child missing for nkElseExpr >>
+'''
+"""
+import std/macros
+
+macro t1(): untyped =
+  result = newTree(nnkIfStmt)
+  result.add(
+    newNimNode(nnkElseExpr)
+  )
```
zerbina:
```suggestion
  result =
    nnkIfStmt.newTree(
      newNimNode(nnkElseExpr))
```
A minor stylistic suggestion.

tests/lang_callable/macros/tif_ill_formed.nim
```diff
@@ -0,0 +1,16 @@
+discard """
+  cmd: "nim check $options $file"
```
zerbina:
```suggestion
  description: '''
    Ensure that ill-formed if-statement AST created by macros is detected
  '''
  cmd: "nim check $options $file"
```
This makes it a bit clearer that other `if` AST tests should also go here.

tests/lang_callable/macros/tif_ill_formed.nim
```diff
@@ -0,0 +1,16 @@
...
+  action: reject
+  nimout: '''
+tif_ill_formed.nim(13, 15) Error: illformed AST:  else: <<0th child missing for nkElseExpr >>
+'''
```
zerbina: Could you use inline annotations (`#[tt.Error`) instead? They make it easier to extend this test in the future, as one doesn't have to update all line numbers.

tests/lang_callable/macros/tif_ill_formed.nim
```diff
@@ -0,0 +1,17 @@
+discard """
+  description: '''
+    Ensure that ill-formed if-statement AST created by macros is detected
+  '''
+  cmd: "nim check --msgFormat:sexp --filenames=canonical --hints:off $options $file"
+  nimoutformat: sexp
```
zerbina:
```suggestion
  cmd: "nim check --hints:off $options $file"
  nimoutfull: "true"
```
The `sexp` mode and related testament support is a bit shaky at the moment, the general direction is to not use it for now.

tests/lang_callable/macros/tif_ill_formed.nim
```diff
@@ -0,0 +1,17 @@
...
+  action: reject
+"""
+import std/macros
+
+macro t1(): untyped =
+  result =
+    nnkIfStmt.newTree(
+      newNimNode(nnkElseExpr)) #[tt.Error
+               ^ (SemIllformedAst)]#
```
zerbina:
```suggestion
      newNimNode(nnkElseExpr)) #[tt.Error
               ^ illformed AST:  else: <<0th child missing for nkElseExpr >>]#
```

tests/lang_callable/macros/tif_ill_formed.nim
```diff
@@ -0,0 +1,17 @@
...
+  nimoutfull: "true"
+  action: reject
+"""
+import std/macros
+
+macro t1(): untyped =
+  result =
+    nnkIfStmt.newTree(
+      newNimNode(nnkElseExpr)) #[tt.Error
+               ^ illformed AST:  else: <<0th child missing for nkElseExpr >>]#
```
zerbina:
```suggestion
                ^ illformed AST:  else: <<0th child missing for nkElseExpr >>]#
```
That should fix the failure.

bung87:
well, look back into this, this is similar to https://github.com/nim-works/nimskull/pull/932, except that one may encounter `checkMinSonsLen`, `checkMinSonsLen` report global error prevent downstream errors, this one calls `semReportIllformedAst` which internally call `handleReport` directly, instead of `globalReport` or `localReport`, so the error affect rest part of procedure and downstream. as I can say they are both ill formed error but reported different way.

zerbina:
> internally call `handleReport` directly, instead of `globalReport` or `localReport`, so the error affect rest part of procedure and downstream

`localReport` also only calls `handleReport` with `doNothing`, which is the same thing that `semReportIllformedAst` does. A `globalReport` would have indeed prevented the crash, as it raises an `ERecoverableError`.

In general, the direction is that all calls to `globalReport`, `localReport`, `handleReport`, etc. should be replaced with `nkError` propagation, so that progress can be made despite ill-formed/illegal code. Turning all the `checkSonsLen`, etc. into `nkError` propagation would be quite the effort, however -- #508 was/is intended to help there.

zerbina (approved):

PR #934, opened 2023-09-30
internal: fix the meaning of up/down conversion

zerbina:
## Summary

The meaning of up- and down-conversions and the corresponding nodes was
inverted within the compiler. Up-conversion is changed to mean
"conversion to base type" (going up in the hierarchy), while down-
conversion is changed to mean "conversion to derived type" (going down
in the hierarchy).

This doesn't change program semantics, but the change is visible to
macros accepting typed AST, where `nkObjDownConv` and `nkObjUpConv` now
have the correct meaning.

## Details

The core of the problem was `inheritanceDiff` now doing what the
document comment said: the parent/base types of `a` were traversed
first (which checks whether `a` is a *subtype* of `b`), with each step
decrementing the counter (negative result). However, the procedure's
documentation said that `a` being a subtype of `b` results in a
*positive* return value, not a *negative*  one.

`inheritanceDiff` is fixed, and all callsites are adjusted:
* for all usage sites of `inheritanceDiff` and `safeInheritanceDiff`
  where the result is not used for a) producing `nkObjDownConv` or
  `nkObjUpConv` nodes, or b) only checking for an relation at all,
  the arguments are reversed
* the correct node kinds were used (although their meaning was
  inverted) when producing  `nkObjDownConv` / `nkObjUpConv`  (and their
code
  generator version), so no change is required there
* for `cgmeth.genConv`, whether only a down-conversion is allowed is
  adjusted at the callsites: when patching method calls, only down-
  conversions are allowed (sub-type to base), while when generating the
  dispatcher, only up-conversion (base to sub-type) are allowed

For the code generators, the implementations of `cgen.downConv` and
`cgen.upConv` are swapped, and `jsgen.upConv` is renamed to
`jsgen.downConv`.

Except for one occurrence in the previous `cgen.downConv`, one in
`closureiters`, and one in `wrapInHiddenAddr`, `nkObjDownConv`/
`nkObjUpConv` (and their code generator versions) are always used
together, so only the three mentioned occurrences need to be changed:
in each case, the node kind is replaced with its respective
counterparts.

### DIFF
--- a/compiler/ast/types.nim
+++ b/compiler/ast/types.nim
@@ -933,14 +933,14 @@ proc inheritanceDiff*(a, b: PType): int =
     x = skipTypes(x, skipPtrs)
     if sameObjectTypes(x, b): return
     x = x[0]
-    dec(result)
+    inc(result)
   var y = b
   result = 0
   while y != nil:
     y = skipTypes(y, skipPtrs)
     if sameObjectTypes(y, a): return
     y = y[0]
-    inc(result)
+    dec(result)
   result = high(int)
 
 proc commonSuperclass*(a, b: PType): PType =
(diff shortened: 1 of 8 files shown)


saem (approved):
While reading through this PR, I realized that I think of it as upgrade/downgrade and treat going to the parent a downgrade, so it never felt backwards. 😆

PR #935, opened 2023-10-01
mir: add a dedicated handle conversion operation

zerbina:
## Summary

Introduce a dedicated handle conversion operator (`mnkPathConv`) to the
MIR, leaving `mnkConv` to only represent conversions producing new (non
handle-like) values.

This clarifies the semantics around MIR-level conversions a bit, fixing
the internal-only issue of `mnkConv` and `mnkStdConv` being sometimes
treated as if they always represented handle conversion.

## Details

A value conversion (e.g., `int(f)` where `f` is a float value) produces
a new value -- the result cannot be assigned to. A *handle* conversion
produces a new *handle* to the same location as that of the input
value.

The distinction is important at the MIR level, as it affects ownership
and evaluation semantics. `mnkConv` was previously used to represent
both kinds of conversions, but since the type comparison required for
detecting the conversion kind is rather costly, `mnkConv` was so far
almost always treated as the one the context required.

The current structure of the MIR lets one get away with this, but it's
still incorrect, and it's also in the way of MIR improvements. In
addition,  `mnkStdConv`  always represents value conversions, and it
being
used/skipped where  `mnkConv`  is treated as a handle conversion is
always
wrong.

In order to remove the ambiguity, the new `mnkPathConv` operator is
introduced. Since it too represents a path/projection, it is grouped
together with the other path operations. For simplicity, `mnkPathConv`
also applies to first-class handle-like values (`ref`, `ptr`, `var`,
etc.) for now -- they are presently treated the same as L-value
conversions.

* add the `mnkPathConv` operation. Syntax-wise, it's a single-input,
  result-yielding operation
* add `mnkPathConv` support to the MIR construction DSL
* emit an `mnkPathConv` instead of `mnkConv` for object/ref up- and
  down-conversions
* emit a `mnkPathConv` instead of `mnkConv` for conversions between
  distinct types and their base (and vice versa)

The places where `mnkConv` was previously treated as a handle
conversions are changed to use `mnkPathConv`. `mirpasses.isRvalue`,
which had to conservatively exclude it previously, now includes the
`mnkConv` operation.

There's also an issue in `transf` that had to be fixed: all up- and
down- conversions are expected to use `nkObjUpConv` and `nkObjDownConv`
nodes past `transf` (see `transformConv`), but lowering an exception
handler branch (`transformExceptBranch`) emitted a raw
`nkHiddenSubConv` for down conversions.

### DIFF
--- a/compiler/backend/cgirgen.nim
+++ b/compiler/backend/cgirgen.nim
@@ -764,8 +764,8 @@ proc handleSpecialConv(c: ConfigRef, n: CgNode, info: TLineInfo,
     result = nil
 
 proc tbConv(cl: TranslateCl, n: CgNode, info: TLineInfo, dest: PType): CgNode =
-  ## Generates the IR for an expression that performs a type conversion for
-  ## `n` to type `dest`
+  ## Generates the ``CgNode`` IR for an ``mnkPathConv`` operation (handle
+  ## conversion).
   result = handleSpecialConv(cl.graph.config, n, info, dest)
   if result == nil:
     # no special conversion is used
@@ -1150,7 +1150,7 @@ proc tbInOut(tree: TreeWithSource, cl: var TranslateCl, prev: sink Values,
   of mnkCast:
     toValues newOp(cnkCast, info, n.typ, prev.single)
   of mnkConv:
-    toValues tbConv(cl, prev.single, info, n.typ)
+    toValues newOp(cnkConv, info, n.typ, prev.single)
   of mnkStdConv:
     let
       opr = prev.single
@@ -1214,6 +1214,8 @@ proc tbInOut(tree: TreeWithSource, cl: var TranslateCl, prev: sink Values,
 
   of mnkPathArray:
     toValues newExpr(cnkBracketAccess, info, n.typ, move prev.list)
+  of mnkPathConv:
+    toValues tbConv(cl, prev.single, info, n.typ)
   of mnkAddr:
     toValues newOp(cnkAddr, info, n.typ, prev.single)
   of mnkDeref:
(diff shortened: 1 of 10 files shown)


zerbina:
I'm unsure about why the `tdestructor_outer_scope.nim` test was failing with the MacOS CI. Going just by the line numbers reported in the C compiler errors, it looks like there was a lot of code missing within the `stdlib_system.c` file.

While I don't have access to a MacOS system, I did make a local build of the compiler with `-fsanitize=address` enabled, in order to rule out memory corruptions within the compiler that are caused by this PR. There were no problems reported, and on re-running the CI job, the test succeeded.

saem (commented):
Read all the changes, all good.

I'm re-reading the description and referencing the code to see if I can figure out a test or two as things were fixed and/or improved.

saem:
> I'm unsure about why the `tdestructor_outer_scope.nim` test was failing with the MacOS CI. Going just by the line numbers reported in the C compiler errors, it looks like there was a lot of code missing within the `stdlib_system.c` file.
> 
> While I don't have access to a MacOS system, I did make a local build of the compiler with `-fsanitize=address` enabled, in order to rule out memory corruptions within the compiler that are caused by this PR. There were no problems reported, and on re-running the CI job, the test succeeded.

I have a MacOS system did you want me to try and run the test?

zerbina:
> I'm re-reading the description and referencing the code to see if I can figure out a test or two as things were fixed and/or improved.

Hmm, there might actually be a case where an unnecessary destructor call previously was injected for `nil` literals. I'll look into whether that's the case.

> I have a MacOS system did you want me to try and run the test?

Yep, that'd be nice, thank you. The test that failed is the `lang_object/destructor/tdestructor_outer_scope.nim` test.

saem:
I ran the test and it passes, I tried:

- running the test via testment + a bootstrapped danger compiler
- running the test via testament + a debug temp compiler
- running the tests files manually via CLI using the debug temp compiler and manually checking the output

zerbina:
Thank you for the thorough testing. I think it's safe to treat the error as a spurious one, then.

PR #936, opened 2023-10-02
Refactor report hook

bung87:
## Summary
* The `structuredReportHook` procedure now takes a third parameter `TErrorHandling`. 
* The `structuredReportHook` previously only been used for certain cases, such as always returning `doNothing` when used with `nimsuggest`. In fact, we do not need `doDefault` or `doAbort` since they may unexpectedly call `quit`. However, we do need to properly handle `doRaise`, which is reported from `globalReport` and can prevent compilation in command line mode when it comes across an ill-formed AST or other errors that prevent code compilation. Therefore, we should reuse this logic to prevent `nimsuggest` from continuing to compile code that could cause crashes due to nil access or subsequent errors caused by ill-formed code.

## Details

---

## Notes for Reviewers
Earlier, the test case code would encounter a fatal error and fail to report it. 
Now it runs smoothly and successfully reports the error, so I believe this change has been addressed.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -3984,7 +3984,7 @@ proc legacyReportBridge*(conf: ConfigRef, diag: PAstDiag): Report =
   # TODO: will use conf soon
   astDiagToLegacyReport(conf, diag)
 
-proc reportHook*(conf: ConfigRef, r: Report): TErrorHandling =
+proc reportHook*(conf: ConfigRef, r: Report, rh: TErrorHandling): TErrorHandling =
   ## Default implementation of the report hook. Dispatches into
   ## `reportBody` for report, which then calls respective (for each report
   ## category) `reportBody` overloads defined above
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -276,10 +276,10 @@ proc quit(conf: ConfigRef; withTrace: bool) {.gcsafe.} =
       if stackTraceAvailable():
         discard conf.report(InternalReport(
           kind: rintStackTrace,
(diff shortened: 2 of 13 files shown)


zerbina:
Have you considered instead testing for `rsemIllformedAst` reports within the structured report hook implementation for `nimsuggest`, and on encountering it, forcing a `doRaise`? If I didn't make any mistakes, that should accomplish the same thing, while being much less invasive.

bung87:
`rsemIllformedAst` is only one kind of errors that reported from `globalReport`, from compiler developer perspective they first choose `globalReport` or `localReport` then pick one report kind, and the kinds can growing in future, so I cant proven that by using kinds.

zerbina:
I'm aware that `globalReport` is used for other errors too, but `rsemIllformedAst` is the only report where it's not so easy to remove the `globalReport` usage. As I mentioned in the other PR, the direction is *removing* `globalReport` and `localReport` (and reports in general) from the compiler.

Having said that, shifting the responsibility into the report hook (i.e., the user) is indeed not a good idea. Rather, I think that both `checkSonsLen` and `checkSonsMinLen` should invoke the new `reportAndForceRaise` procedure. This procedure would behave similar to `handleReport`, but instead of allowing nothing to happen, it would always call `raiseRecoverableError`, except if `doAbort` was returned by the hook. This ensures that `checkSonsLen` and `checkSonsMinLen` *always* unwind to the closest handler, without requiring any action from the hook.

bung87:
Create new PR here: https://github.com/nim-works/nimskull/pull/939

PR #937, opened 2023-10-02
fix(renderer): `hasCom` childless node crash

bung87:
## Summary

Fix compiler crash due to `renderer.hasCom` attempting to access `sons`
on `PNode`s without children.

## Details

This happened because `hasCom` was checking guarding with the
`nkEmpty..nkNilLit`  range, instead of constant  `nkWithoutSons` . The
prior
range didn't include  `nkNone`  and  `nkError` . The implementation now
uses
an exhaustive case to avoid such errors in the future.

### DIFF
--- a/compiler/ast/renderer.nim
+++ b/compiler/ast/renderer.nim
@@ -536,7 +536,7 @@ proc hasCom(n: PNode): bool =
   if n.comment.len > 0: return true
   case n.kind
   of nkWithoutSons: discard
-  else:
+  of nkWithSons:
     for i in 0..<n.len:
       if hasCom(n[i]): return true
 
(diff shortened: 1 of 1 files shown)


saem (changes requested):
A few things:

- the change needs a test case
- pr body should include a description of what was failing before

bung87:
* the problem is obvious to me
* said cause `FieldDefect`

compiler/ast/renderer.nim
```diff
@@ -535,7 +535,7 @@ proc hasCom(n: PNode): bool =
   if n.isNil: return false
   if n.comment.len > 0: return true
   case n.kind
-  of nkEmpty..nkNilLit: discard
+  of nkWithoutSons: discard
   else:
```
saem:
```suggestion
  of nkWithSons:
```

making it exhaustive will reduce the chance of missed branches in the future.

saem:
There is the exhausitiveness suggestion, otherwise this is good to go.

PS. I've updated the PR description, assuming the exhaustiveness change goes through.

saem (approved):

PR #938, opened 2023-10-03
`semProcAux` `hasError` cover `fixupInstantiatedSymbols`

bung87:
## Summary
* `semProcAux` `hasError` cover `fixupInstantiatedSymbols`
* Instantiated symbol's AST may be the `nkError` kind that cause`n[bodyPos]` raise `FieldDefect`

## Details
* `fixupInstantiatedSymbols` now returns `false` if instantiated symbol's AST is `nkError` kind during iteration

---

## Notes for Reviewers

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/seminst.nim
+++ b/compiler/sem/seminst.nim
@@ -119,15 +119,20 @@ proc instantiateBody(c: PContext, n, params: PNode, result, orig: PSym) =
     trackProc(c, result, result.ast[bodyPos])
     dec c.inGenericInst
 
-proc fixupInstantiatedSymbols(c: PContext, s: PSym) =
+proc fixupInstantiatedSymbols(c: PContext, s: PSym): bool =
+  result = true
   for i in 0..<c.generics.len:
     if c.generics[i].genericSym.id == s.id:
-      var oldPrc = c.generics[i].inst.sym
+      var 
+        oldPrc = c.generics[i].inst.sym
+        n = oldPrc.ast
+      if n.kind == nkError:
+        result = false
+        continue
       pushProcCon(c, oldPrc)
       pushOwner(c, oldPrc)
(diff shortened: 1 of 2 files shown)


zerbina (changes requested):
This is not correct, sorry.

There being an error in the instantiation of a generic procedure is **not** an error with the *definition*. Besides, the way you use `hasError` here is wrong: AST must only be wrapped in a wrapper error (`wrapError`) when there's an `nkError` somewhere within, but this is not the case here.

In general, please document in the PR message what exactly you're fixing (*what* is the error and *why* did it occur), and also provide a test where you have made sure that it reproduces the failure, thanks. I don't want to have to investigate these issues myself each time.

bung87:
by the words "not an error", here should continue iteration without marking anything ?

zerbina:
> by the words "not an error", here should continue iteration without marking anything ?

Yes. Instantiations with errors are reported when they're inserted into the module AST.

bung87:
is that a global error? because here `n[bodyPos]` may cause `FieldDefect`

zerbina:
Oh, my bad, I misremembered. `instantiateGeneric` reports the error (via `localReport`), `appendInstancedGenericRuntimeRoutines` doesn't.

bung87:
okay so here the error check is needed I guess, I also aware even similar kind of errors, they use `localReport` or `globalReport` that seems don't have certain rules. that's another thing.

zerbina:
To be clear, my original review comment still applies.

bung87:
okay, so this is the another case compiler continue compilation cause subsequent errors, the `appendInstancedGenericRuntimeRoutines` do `internalAssert` which should cause compiler abort,
whereas `structuredReportHook` may just return `doNothing`

PR #939, opened 2023-10-03
fix(sem): `nimsuggest` crashing for ill-formed AST

bung87:
## Summary

Fix `nimsuggest`, and other tooling that uses a custom structured
report hook, crashing when encountering ill-formed AST that has the
wrong number of nodes.

## Details

Many of the semantic analysis procedures first verify that the input
`PNode` has the right number of child nodes. This is done via the
`checkSonsLen` and `checkMinSonsLen` procedures, with the expectation
being that both raise an `ERecoverableError` (through `globalReport`)
when the check fails.

However, since reporting the error happens via `handleReport`, it's
possible that the error handling strategy is overridden with a mode 
(e.g., `doNothing`) that results in execution to continue without any
exception being raised. `nimsuggest` sets the error handling mode to
`doNothing`, which then resulted in `checkSonsLen` and
`checkMinSonsLen` not ending the active routine, usually leading to
`IndexDefect`s or other crashes.

As an interim solution, the `reportAndFail` procedure is introduced.
`reportAndForceRaise` works similar to similar to `handleReport`, but
it ensures that *at least* an exception is raised -- both
`checkSonsLen` and `checkMinSonsLen` use it to ensure that the calling
routine is exited.

### DIFF
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -664,7 +664,7 @@ template globalReport*(conf: ConfigRef, report: ReportTypes) =
   handleReport(
     conf, wrap(report, instLoc()), instLoc(), doRaise)
 
-proc reportAndForceRaise*(
+proc reportAndFail*(
   conf: ConfigRef, r: Report, reportFrom: InstantiationInfo) =
   ## Similar to `handleReport`, but, unless overridden with aborting
   ## (`doAbort`) by the structured report hook, always raises a recoverable
@@ -672,15 +672,25 @@ proc reportAndForceRaise*(
   var rep = r
   fillReportAndHandleVmTrace(conf, rep, reportFrom)
 
-  case conf.report(rep)
+  let userAction = conf.report(rep)
+  case userAction
   of doAbort:
     quit 1
-  of doRaise, doDefault, doNothing:
+  of doDefault:
+    let (action, trace) = errorActions(conf, rep, userAction)
+    case action
+    of doAbort:
+      quit 1
+    of doRaise, doNothing:
+      raiseRecoverableError("report")
+    of doDefault:
+      unreachable()
+  of doRaise, doNothing:
     raiseRecoverableError("report")
 
-template reportAndForceRaise*(
+template reportAndFail*(
   conf: ConfigRef; info: TLineInfo, report: ReportTypes) =
-  reportAndForceRaise(
+  reportAndFail(
     conf, wrap(report, instLoc(), info), instLoc())
 
 template localReport*(conf: ConfigRef; info: TLineInfo, report: ReportTypes) =
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -672,15 +672,14 @@ proc reportAndFail*(
   var rep = r
   fillReportAndHandleVmTrace(conf, rep, reportFrom)
 
-  let userAction = conf.report(rep)
-  case userAction
+  case conf.report(rep)
   of doAbort:
-    quit 1
+    quit(conf, false)
   of doDefault:
-    let (action, trace) = errorActions(conf, rep, userAction)
+    let (action, trace) = errorActions(conf, rep, doRaise)
     case action
     of doAbort:
-      quit 1
+      quit(conf, trace)
     of doRaise, doNothing:
       raiseRecoverableError("report")
     of doDefault:
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -600,10 +600,12 @@ proc report*(conf: ConfigRef, node: PNode): TErrorHandling =
   assert node.kind == nkError
   return conf.report(conf.astDiagToLegacyReport(conf, node.diag))
 
-proc fillReport(c: ConfigRef, r: var Report, reportFrom: InstantiationInfo) =
+proc fillReportAndHandleVmReport(c: ConfigRef, r: var Report, reportFrom: InstantiationInfo) =
   r.reportFrom = toReportLineInfo(reportFrom)
   if r.category in { repSem, repVM } and r.location.isSome():
     r.context = c.getContext(r.location.get())
+  if r.category == repVM and r.vmReport.trace != nil:
+    handleReport(conf, wrap(r.vmReport.trace[]), reportFrom)
 
 proc handleReport*(
     conf: ConfigRef,
@@ -611,9 +613,7 @@ proc handleReport*(
     reportFrom: InstantiationInfo,
     eh: TErrorHandling = doNothing) {.noinline.} =
   var rep = r
-  fillReport(conf, rep, reportFrom)
-  if rep.category == repVM and rep.vmReport.trace != nil:
-    handleReport(conf, wrap(rep.vmReport.trace[]), reportFrom)
+  fillReportAndHandleVmReport(conf, rep, reportFrom)
 
   let
     userAction = conf.report(rep)
@@ -662,10 +662,8 @@ proc reportAndForceRaise*(
   ## (`doAbort`) by the structured report hook, always raises a recoverable
   ## error.
   var rep = r
-  fillReport(conf, rep, reportFrom)
+  fillReportAndHandleVmReport(conf, rep, reportFrom)
 
-  if rep.category == repVM and rep.vmReport.trace != nil:
-    reportAndForceRaise(conf, wrap(rep.vmReport.trace[]), reportFrom)
   case conf.report(rep)
   of doAbort:
     quit 1
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -659,7 +659,7 @@ template globalReport*(conf: ConfigRef, report: ReportTypes) =
 proc reportAndForceRaise*(
   conf: ConfigRef, r: Report, reportFrom: InstantiationInfo) =
   ## always call `raiseRecoverableError` except if `doAbort` was returned
-  ## by the structuredReportHook.
+  ## by the `structuredReportHook`.
   var rep = r
   fillReport(rep, reportFrom)
 
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -600,10 +600,10 @@ proc report*(conf: ConfigRef, node: PNode): TErrorHandling =
   assert node.kind == nkError
   return conf.report(conf.astDiagToLegacyReport(conf, node.diag))
 
-proc fillReport(r: var Report, reportFrom: InstantiationInfo) =
+proc fillReport(c: ConfigRef, r: var Report, reportFrom: InstantiationInfo) =
   r.reportFrom = toReportLineInfo(reportFrom)
   if r.category in { repSem, repVM } and r.location.isSome():
-    r.context = conf.getContext(r.location.get())
+    r.context = c.getContext(r.location.get())
 
 proc handleReport*(
     conf: ConfigRef,
@@ -611,7 +611,7 @@ proc handleReport*(
     reportFrom: InstantiationInfo,
     eh: TErrorHandling = doNothing) {.noinline.} =
   var rep = r
-  fillReport(rep, reportFrom)
+  fillReport(conf, rep, reportFrom)
   if rep.category == repVM and rep.vmReport.trace != nil:
     handleReport(conf, wrap(rep.vmReport.trace[]), reportFrom)
 
@@ -661,7 +661,7 @@ proc reportAndForceRaise*(
   ## always call `raiseRecoverableError` except if `doAbort` was returned
   ## by the `structuredReportHook`.
   var rep = r
-  fillReport(rep, reportFrom)
+  fillReport(conf, rep, reportFrom)
 
   if rep.category == repVM and rep.vmReport.trace != nil:
     reportAndForceRaise(conf, wrap(rep.vmReport.trace[]), reportFrom)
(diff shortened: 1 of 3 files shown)


compiler/front/msgs.nim
```diff
@@ -654,6 +654,26 @@ template globalReport*(conf: ConfigRef, report: ReportTypes) =
   handleReport(
     conf, wrap(report, instLoc()), instLoc(), doRaise)
 
+proc reportAndForceRaise*(
```
zerbina: Please document this procedure. At least exported routines should always have a documentation comment describing what they do.

compiler/front/msgs.nim
```diff
@@ -654,6 +654,26 @@ template globalReport*(conf: ConfigRef, report: ReportTypes) =
...
+  conf: ConfigRef, r: Report, reportFrom: InstantiationInfo) =
+  var rep = r
+  rep.reportFrom = toReportLineInfo(reportFrom)
+  if rep.category in { repSem, repVM } and rep.location.isSome():
+    rep.context = conf.getContext(rep.location.get())
+
+  if rep.category == repVM and rep.vmReport.trace != nil:
+    reportAndForceRaise(conf, wrap(rep.vmReport.trace[]), reportFrom)
```
zerbina: This part is shared with `handleReport`. Please move it into a common procedure.

nimsuggest/tests/terror_handling.nim
```diff
@@ -0,0 +1,18 @@
+# Test ill formed AST will not causes crash and subsequent errors
```
zerbina:
```suggestion
# Ensure that AST not having the correct number of nodes doesn't cause
# crashes. This a regression test for `checkSonsMinLen` and
# `checkSonsLen` not aborting analysis.
```
It's a bit more precise, and it also highlights what exactly the issue is.

nimsuggest/tests/terror_handling.nim
```diff
@@ -0,0 +1,18 @@
...
+
+import std/macros
+
+macro t1(): untyped =
+  result = newNimNode(nnkForStmt)
+  result.add(ident("i")) 
+  result.add newNimNode(nnkCall)
+  result.add nnkDiscardStmt.newTree(newEmptyNode())
```
zerbina:
```suggestion
  newNimNode(nnkCall) # an call AST with not enough sub-nodes
```
It's simpler, and causes the same kind of failure.

compiler/front/msgs.nim
```diff
@@ -654,6 +654,26 @@ template globalReport*(conf: ConfigRef, report: ReportTypes) =
...
+  let
+    userAction = conf.report(rep)
+  case userAction
+  of doAbort:   quit 1
+  else: raiseRecoverableError("report")
```
zerbina:
```suggestion
  case conf.report(rep)
  of doAbort:
    quit 1
  of doRaise, doDefault, doNothing:
    raiseRecoverableError("report")
```
Use exhaustive `case` statements whenever possible.

Also, please be more thorough with code style. You don't have to align each branch body horizontally, but then don't use superfluous spaces.

zerbina (changes requested):
Thank you. Can the other PR be closed now?

The details section in the PR message is good, but the title and summary need to be adjusted. This PR represents a significant fix for `nimsuggest`, but it also affects `compiles`, both which needs to be highlighted.

Regarding the implementation, I've left some requests regarding code style, documentation, and test improvements.

nimsuggest/tests/terror_handling.nim
bung87: no, the purpose here is ensure for statement errors not be reported.

zerbina: The issue you're fixing is not at all related to `for` statements.

You didn't provide a reproducible case for the `IndexDefect` errors you encountered within `semFor`, but I'm very certain that this PR doesn't fix them. If the error would be caused by a macro-generated zero-length `nkCall` node, then the compiler would have already crashed in `semExpr` when analysing the `nkCall` node.

Don't get me wrong, I still think that making sure that `checkSonsLen` always at least raises an exception is a very important fix.

bung87: I still think it's needed, `newNimNode` call doesn't check the length. the test case covered before this PR, so I want ensure it will not happen again.

zerbina: Yes, `checkSonsLen` and `checkSonsMinLen` should always at least raise an error.

compiler/front/msgs.nim
```diff
@@ -654,6 +656,26 @@ template globalReport*(conf: ConfigRef, report: ReportTypes) =
   handleReport(
     conf, wrap(report, instLoc()), instLoc(), doRaise)
 
+proc reportAndForceRaise*(
+  conf: ConfigRef, r: Report, reportFrom: InstantiationInfo) =
+  ## always call `raiseRecoverableError` except if `doAbort` was returned
+  ## by the `structuredReportHook`.
```
zerbina:
```suggestion
  ## Similar to `handleReport`, but, unless overridden with aborting
  ## (`doAbort`) by the structured report hook, always raises a recoverable
  ## error.
```
This focuses more on the concrete behaviour, rather than on the implementation.

compiler/front/msgs.nim
```diff
@@ -654,6 +656,26 @@ template globalReport*(conf: ConfigRef, report: ReportTypes) =
...
+  var rep = r
+  fillReport(conf, rep, reportFrom)
+
+  if rep.category == repVM and rep.vmReport.trace != nil:
+    reportAndForceRaise(conf, wrap(rep.vmReport.trace[]), reportFrom)
```
zerbina: Reporting the stack-trace for a VM report must not forcefully raise. Please use `handleReport` here and then move the `if rep.category == repVM ...` into `fillReport`.

bung87: I have no idea what kind of vm report here, but doesn't `handleReport` could `quit` ? that might cause `nimsuggest` crash?

zerbina: Yes, `handleReport` can call `quit` (but so does `reportAndForceRaise`).

nimsuggest/tests/terror_handling.nim
zerbina: > the test case covered before this PR, so I want ensure it will not happen again.

I don't understand. As you can see by what the `chk` result is reported for (the invalid `nkCall` node), the `nnkForStmt` is unnecessary.

If you mean that the test should use `nnkCall.newNimNode()` instead of `nnkCall.newTree()`, then yep, that's better.

bung87: but before this PR, the test case will also cause `IndexDefect` in `semFor`, so my original attemption is not only testing the nnkCall error reported, but `checkMinSonsLen` raise and prevent subsequent errros been reproted in `semFor`

zerbina: > but before this PR, the test case will also cause `IndexDefect` in `semFor`

No, `semFor` doesn't crash with an `IndexDefect` prior to this PR, and yes, I've tested it. `nimsuggest` crashes in `semExpr`, when analysing the `nkCall` node, like I said above.

The original intention is not relevant for the final a test. The final test has to focus  on the actually fixed issue, and not be what the investigation into the issue started with. In other words, the test needs to be reduced as much as possible, and as it stands, your test is not. Please use the suggestion I provided -- I've also changed it to use `newNimNode`.

If `checkMinSonsLen` wouldn't raise then my suggested test would cause an `IndexDefect` crash, in the very same way that the current test does too.

Strictly speaking, each node kind would need to be tested here, but testing one is enough for now.

compiler/front/msgs.nim
```diff
@@ -600,18 +606,20 @@ proc report*(conf: ConfigRef, node: PNode): TErrorHandling =
   assert node.kind == nkError
   return conf.report(conf.astDiagToLegacyReport(conf, node.diag))
 
+proc fillReportAndHandleVmReport(c: ConfigRef, r: var Report, reportFrom: InstantiationInfo) =
```
zerbina:
```suggestion
proc fillReportAndHandleVmTrace(c: ConfigRef, r: var Report,
                                reportFrom: InstantiationInfo) =
```
The procedure doesn't handle VM reports in general, only the associated stack-trace report.

compiler/front/msgs.nim
```diff
@@ -600,18 +606,20 @@ proc report*(conf: ConfigRef, node: PNode): TErrorHandling =
...
     reportFrom: InstantiationInfo,
     eh: TErrorHandling = doNothing) {.noinline.} =
   var rep = r
-  rep.reportFrom = toReportLineInfo(reportFrom)
-  if rep.category in { repSem, repVM } and rep.location.isSome():
-    rep.context = conf.getContext(rep.location.get())
-
-  if rep.category == repVM and rep.vmReport.trace != nil:
-    handleReport(conf, wrap(rep.vmReport.trace[]), reportFrom)
+  fillReportAndHandleVmReport(conf, rep, reportFrom)
```
zerbina:
```suggestion
  fillReportAndHandleVmTrace(conf, rep, reportFrom)
```

compiler/front/msgs.nim
```diff
@@ -654,6 +662,25 @@ template globalReport*(conf: ConfigRef, report: ReportTypes) =
   handleReport(
     conf, wrap(report, instLoc()), instLoc(), doRaise)
 
+proc reportAndForceRaise*(
+  conf: ConfigRef, r: Report, reportFrom: InstantiationInfo) =
+  ## Similar to `handleReport`, but, unless overridden with aborting
+  ## (`doAbort`) by the structured report hook, always raises a recoverable
+  ## error.
+  var rep = r
+  fillReportAndHandleVmReport(conf, rep, reportFrom)
```
zerbina:
```suggestion
  fillReportAndHandleVmTrace(conf, rep, reportFrom)
```

compiler/front/msgs.nim
```diff
@@ -600,18 +606,20 @@ proc report*(conf: ConfigRef, node: PNode): TErrorHandling =
   assert node.kind == nkError
   return conf.report(conf.astDiagToLegacyReport(conf, node.diag))
 
+proc fillReportAndHandleVmReport(c: ConfigRef, r: var Report, reportFrom: InstantiationInfo) =
+  r.reportFrom = toReportLineInfo(reportFrom)
+  if r.category in { repSem, repVM } and r.location.isSome():
+    r.context = c.getContext(r.location.get())
+  if r.category == repVM and r.vmReport.trace != nil:
```
zerbina, line 615:
```suggestion

  if r.category == repVM and r.vmReport.trace != nil:
```
Block-like statements that don't have a strong relation to each other should, for readability, be separated by a new line.

zerbina (changes requested):
Sorry for not noticing this earlier. As is, the implementation ignores the default actions, which means that `--errorMax` is ignored by `reportAndForceRaise`.

You need to change `reportAndForceRaise` to first go through `errorActions` (refer to `handleReport` for how this should look like), before forcing `doNothing` to `doRaise`.

At this point, I think that `reportAndFail` would be a better name for the procedure.

bung87:
like this?
```nim
case conf.report(rep)
of doAbort:
  quit 1
of doDefault:
  let (action, trace) = errorActions(conf, rep, eh)
  case action
  of doAbort:
    quit 1
  of doRaise, doNothing:
    raiseRecoverableError("report")
of doRaise, doNothing:
  raiseRecoverableError("report")
```

compiler/front/msgs.nim
```diff
@@ -654,6 +664,35 @@ template globalReport*(conf: ConfigRef, report: ReportTypes) =
...
+
+  let userAction = conf.report(rep)
+  case userAction
+  of doAbort:
+    quit 1
+  of doDefault:
+    let (action, trace) = errorActions(conf, rep, userAction)
+    case action
+    of doAbort:
+      quit 1
```
zerbina:
```suggestion
  case conf.report(rep)
  of doAbort:
    quit(conf, false)
  of doDefault:
    let (action, trace) = errorActions(conf, rep, doRaise)
    case action
    of doAbort:
      quit(conf, trace)
```
Two things:
- the compiler's `quit` overload needs to be used
- for clarity, even though it doesn't make a difference at the moment, the requested error handling in the `doDefault` case should be `doRaise`

zerbina:
Hmm, I just noticed a much simpler and also more broad fix that also fixes `internalError` being turned into a no-op by the hook. The reporting subsystem is quite convoluted, and I have some trouble navigating it.

However, I don't want to block this PR any longer, and would thus make a follow-up PR myself. In the meantime, this fix works well enough. Sorry for all the detours, and thank you for working on this.

zerbina:
@bung87 I've updated the PR message and title, please take a look at it:
- I reworded the title so that it's a bit shorter 
- the summary gives a broad overview of what failed and what works now. For a tooling-related fix, this should focus on user-facing behaviour. The previous summary listed the implementation details, but didn't summarize the fixed issue

Once the remaining suggestion is applied, this PR is good to be merged.

bung87:
> Hmm, I just noticed a much simpler and also more broad fix that also fixes `internalError` being turned into a no-op by the hook. The reporting subsystem is quite convoluted, and I have some trouble navigating it.
> 
> However, I don't want to block this PR any longer, and would thus make a follow-up PR myself. In the meantime, this fix works well enough. Sorry for all the detours, and thank you for working on this.

I also think this can be simpler, eg. the condition can depends on `config.cmd` that give a default handling that doesn't need `TErrorHandling` returns from user side. but I didn't think further.

zerbina (approved):

PR #941, opened 2023-10-05
msgs: prevent downgrading of error handling

zerbina:
## Summary

Fix the structured report hook being able to downgrade or disable the
error handling as decided by the compiler. This improves stability of
`nimsuggest` when encountering certain errors, as reporting internal
errors or other errors where execution of caller has to stop now behave
as the callsites expect.

## Details

Except for the `doDefault` action, all other actions returned by the
structured report hook overrode the compiler-decided error handling in
`handleReport`. Since `errorActions` wasn't called in those cases, this
also meant that the error counter wasn't incremented properly.

As a consequence of the downgrading, `internalError`, `globalReport`,
and other error reporting that relied on execution of either the
compiler or current procedure to abort could be effectively disabled
through the report hook, which is what `nimsuggest` did, resulting in
execution to continue where it shouldn't.

For the fix, `handleReport` now always computes the error handling
based on the `eh` parameter (via `errorActions`) itself, and then picks
the more severe handling. This means that the structured report hook
can
still "upgrade" error handling to something more severe.

Some follow-up cleanup and refactoring is performed:
* `reportAndFail` is removed, and `checkSonsLen`/`checkMinSonsLen` use
  `globalReport` again
* `internalError`/`internalAssert` now always exit the currently
  executing routine, which is additionally verified by a call to
  `unreachable` after the `handleReport` call, and their
  documentation is updated to reflect that
* `handleReport` is documented

Note that `errorActions` can still downgrade the error handling as
specified by `handleReport`'s parameter, based on the active
configuration. However, fatal errors (see `isCompilerFatal`, which
includes  `rintUnreachable`  and  `rintAssert` ) still always results in
the
compiler/tool aborting.

### DIFF
--- a/compiler/front/msgs.nim
+++ b/compiler/front/msgs.nim
@@ -620,24 +620,38 @@ proc handleReport*(
     r: Report,
     reportFrom: InstantiationInfo,
     eh: TErrorHandling = doNothing) {.noinline.} =
+  ## Prepares the report `r` and passes it to the active report hook.
+  ## `eh` is currently only a suggestion, and it is sometimes ignored depending
+  ## on the currently active configuration.
   var rep = r
   fillReportAndHandleVmTrace(conf, rep, reportFrom)
 
-  let
-    userAction = conf.report(rep)
-    (action, trace) =
-      case userAction
-      of doDefault:
-        errorActions(conf, rep, eh)
-      else:
-        (userAction, false)
+  let userAction = conf.report(rep)
+  # ``errorActions`` also increments the error counter, so make sure to always
+  # call it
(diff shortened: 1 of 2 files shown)


saem (approved):
This is good for an intermediate fix.

Longer term, I think `doAbort` should be split in two, I left my thoughts in this matrix message: https://matrix.to/#/!MAXxvfCYnfYgZqSizw:matrix.org/$ibK0pzJR_5fAFqHAc9flb6cwjY9VxvsdIl5Xud0WNGk?via=libera.chat&via=matrix.org&via=envs.net

haxscramper (approved):

PR #942, opened 2023-10-05
rework the MIR (part 1)

zerbina:
## Summary

Start reworking the MIR by making it more imperative and changing it to
a fully expression/statement-based IR. This brings the MIR closer to
the `CgNode` IR, preparing for the eventual replacement of the latter
with the former.

As a side-effect of the IR changes, JavaScript FFI procedures that
rely on expression shapes rather than types in order to behave
correctly no longer work in most cases. In other words, FFI procedure
cannot be used like templates anymore.

Multiple bugs are fixed, but some regressions are also introduced
(although only temporarily). The most significant fixes are:
* the strict left-to-right, inside-to-outside evaluation order is now
  consistently enforced across all backends
* values moved into `sink` parameter or aggregate construction are no
  longer leaked when executing a following argument expression leaves
  the call/construction expression
* exiting a `ref`-object construction expression through unstructured
  control-flow no longer leaks the already-allocated heap cell
* calling `proc f(x: T, y: sink T): lent T` with `f(a, a)` (the
  arguments both name the same location) prevents `a` from being sunk
  into the parameter, thus preventing the value from being freed too
  early

### Overview

At the core of the rework's first part is replacing the previous stack-
machine-like representation of the MIR with a fully statement-and-
expression-based one that's very similar to the `CgNode` IR, but with
some important differences:
* the MIR (still) uses a flat in-memory representation rather than a
  `ref`-tree
* expressions in the MIR do not contain statements
* rvalue expressions (calls, constructions, conversions, casts, etc.)
  only appear as the source operand of assignments

Compared to the previous design, the new MIR is:
* easier to use for the code generators
* easier to extend
* more flexible and easier to transform/lower
* more similar to the structure of the AST, hopefully making it more
  accessible

Apart from the aforementioned changes, there are two other significant
changes compared to the previous MIR:
* the removal of argument blocks
* the removal of `mnkRegion` (i.e., structured temporaries)

Both have deep implications for the whole mid- and back-end stage, as
they make the MIR much more imperative, with correct evaluation order
now having to be enforced through the use of explicit temporaries.

Furthermore, the `CgNode` trees resulting from the MIR no longer
resemble the rough shape of the AST that is passed to the MIR stage,
meaning that the code generators cannot depend on complex expression
shapes for implementation of magics and optimizations anymore.

The updated `mir.rst` document details the new structure.

#### New MIR node kinds

All of the existing nodes stay (although with slight changes in
meaning, in some cases), but some new ones are added:
* `Alias`: name of run-time alias created during the MIR phase
* `Bind`, `BindMut`: statements with a similar function to `Def`, but
  for defining aliases rather than normal locations
* `ToSlice`: a dedicated operation for sequence-to-slice conversions;
  was previously part of `StdConv`

### AST to MIR translation

With the removal of argument blocks, `mirgen` is now responsible for
translating NimSkull's evaluation order:
* arguments passed to immutable parameters are captured in temporaries,
  except if the expression is a *pure* lvalue (i.e., always evaluates
  to the same value, independent of where it is evaluated)
* arguments passed to mutable `var` parameters (pass by-reference) are
  also captured, except if the expression is a *stable* lvalue (i.e.,
  always has the same address)

While capturing all values and lvalues in temporaries would work too,
it's more efficient to not create temporaries where not necessary,
which is what the above heuristics are fore.

Some lvalue expressions currently have side-effects (i.e., they may
raise an exception), which requires `mirgen` to capture them too.

In general, the focus of the AST -> MIR translation changes from "emit
optimal code first, correct/fixup later" to "emit correct code first,
optimize later".

#### Changes to `transf`

* fix the condition expression for a lowered `while` loop having no
  type
* drop `mRunnableExamples` in `transformCall` already, preventing the
  call from reaching `mirgen` in the first place
* transform to-`varargs` conversions to to-`openArray` conversions;
  `mirgen` needs to assign the conversion to a temporary, but `varargs`
  is not a valid type for a non-parameter location. Having `mirgen`
  produce a new type also doesn't work, as no `IdGenerator` is
  available there

#### Tree construction

* the mini DSL (`chain`, `eval`, etc.) previously implemented in
  `mirconstr` is removed and replaced by `MirBuilder`
* `MirBuilder` wraps two `MirBuffer`s (one final buffer and one staging
  buffer) and provides a structured interface for switching between
  them, appending to them, as well as moving nodes between them
* the double-buffering approach gets around the allocation of many
  short-lived `MirTree`s, at the cost of flexibility
* for convenience, allocation of temporaries is also managed through
  `MirBuilder`

`mirgen` uses a `MirBuilder` internally, wrapping the most used
routines in templates accepting a `TCtx`. The `generateCode` overload
previously accepting a `MirBuffer` now accepts a `MirBuilder`, allowing
for easier integration with AST -> MIR translation from outside of
`mirgen`.

### Changes to changesets

`Changeset` API changes:
* `replaceMulti` and `insert` now provide access to a `MirBuilder`
  instead of a raw node buffer
* the `changeTree` procedure is introduced. It's used for replacing the
  node starting a sub-tree without replacing the sub-nodes
* `getTemp` is removed; allocation of new temporaries is now managed
  through `MirBuilder`

Internally, the multi-node change recording no longer applies the same
source ID to all recorded nodes, leaving this to the `MirBuilder`. This
allows for greater flexibility in how `SourceId`s are attached when
recording multi-node changes.

### Path comparisons, data-flow analysis, and `injectdestructors`

* the `aliasanalysis` module is adapted to the MIR's new design
* instead of comparing two raw lvalue expressions, they're first
  compiled into a `Path` object via `computePath`
* this trades the amount of work during comparison for memory
  consumption and is usually more efficient, as one path staying the
  same across many comparisons is a common use case 
* MIR aliases are followed during compilation of a path

With the argument nodes (`mnkArg`, `mnkName`, and `mnkConsume`) no
longer signaling all usages within a MIR tree, emission of data-flow
instruction in `mirexec.computeDfg` becomes more complex. It also
means that `computeDfg` becomes the source-of-truth regarding what
effects relevant to data-flow analysis an expression/statement has.

For `analysis` and `injectdestructors`:
* the pre-pass (`analysis.computeValues`) for gathering expression
  roots and early ownership information is removed
* instead, the paths are compiled on-demand and a `PackedSet` is used
  for remembering what lvalues are owning, resulting in a simpler
  implementation and less memory usage, at the cost of analysis
  becoming slightly less efficient
* injecting the necessary temporaries for to-be-sunken arguments
  happens in `mirgen`; the now-obsolete `injectTemporaries` is removed
* both `solveOwnership` and `rewriteAssignment` are changed to center
  around data-flow instructions rather than around tree nodes
* injection of `markAcyclic` is removed for now; these calls were used
  for an untested and unfinished optimization that can only be enabled
  by modifying a system module (`orc`)
* rewriting of assignments and hook injection is adjusted to work with
  the new MIR

Two long-standing bugs are fixed as a side-effect:
* temporaries created as part of a call/construction expression are now
  always destroyed, even if the expression is not evaluated in full
  (refer to `ttuple_unpacking.nim`)
* lvalues passed to `lent` parameters (i.e., the first parameter of a
  `lent`-returning procedure) are now treated as usages, preventing
  them from being moved too eagerly and then causing use-after-free
  errors (refer to `tlate_use2.nim`)

### Changes to MIR passes

* the `ArgIter` type and associated routines are obsolete and thus
  removed
* both the in-place construction fixup (implemented as part of
  `preventRvo`) and the call argument fixup pass are made obsolete by
  the AST -> MIR translation being more correct now; both passes are
  removed

#### Optimization pass

* a per-procedure pass for eliding temporaries is introduced
* this was something originally planned for when the rest of the MIR
  rework is implemented, but tests showed that the slowdown for the VM
  target was unacceptable (i.e., order of magnitudes slower)
* the slowdown is due to the current lack of shallow copy support by
  both the VM and JS backend, with each assignment to a temporary
  resulting in a full copy

The pass works as follows:
* a list of single-use temporaries is collected
  * only temporaries that are initialized with an lvalue expression are
    considered
  * lvalues that are projections of temporaries (or direct references
    to a temporary) are ignored too. This optimization barrier prevents
    observable changes to program behaviour order from being changed 
  * lvalues of a select list of types are also ignored
* for each collected single-use temporary, data-flow analysis is used
  to detect whether the lvalue is possibly used for mutations between
  the creation of the temporary and the use thereof
* if no possible mutations are found, the temporary is eliminated
* because lvalue expression can currently have side-effects, they're
  not eliminated completely (so as to not interfere with evaluation
  order) but rather turned into aliases

### Changes to surrounding areas

Two magics were previously used in lvalue positions, which no longer
works longer supported (it was never explicitly supported, even
previously). Different solutions are used in both cases:

* `mAccessTypeEnv`: a new internal magic, `mCopyInternal`, is
  introduced, replacing the usage of `mAccessTypeEnv` in
  `liftdestructors`
* `mAccessEnv`: `mirgen` now lowers the magic into a tuple access
  (closures are effectively tuples). Until closure type are lowered
  prior to code generation, `cgen` has to detect and special-case these
  tuple accesses

The `mEcho` magic, which the code generators previously expected to
have a `cnkArrayConstr` node as the argument expression, needs to be
adjusted too: the strings are now passed to the call directly. For
convenience of code generators that need to create an array (i.e.,
`cgen`), the array type is passed along as the first argument.

#### Changes to MIR utility routines

* `utils.print` is renamed to `treeRepr`, changed to be more robust (it
  no longer raises on ill-formed trees), and its output is made more
  similar to that of `astrepr.treeRepr`
* a pretty-printer for the MIR is added (`render`, `exprToStr`,
  `stmtToStr`)

`--expandArc` now outputs the rendered MIR, rather than the `CgNode`
IR. The intention is to:
* make it easier to reason about the pass' output
* remove the dependency of `injectdestructors` on `cgirgen`

### MIR to `CgNode` translation

The translation becomes significantly simpler due to the removal of
argument blocks and regions, and the MIR now being much closer to the
`CgNode` IR, in terms of structure and semantics.

MIR aliases are translated into `var` and `lent` types, with their
usages wrapped in an automatic deference.

### Impact on the code generators

With some minor differences, the shape of the `CgNode` expressions/
statements resemble that of the MIR:
* only lvalue expressions and literals appear as call operands
* only lvalue expressions appear to the left of assignments and as
  index operands
* there are no statement-list expressions (i.e., statements nested in
  expressions)

A full rewrite of each code generator makes sense now, but is hold off
of until the MIR rework is complete. Some simplifications are still
applied:
* the immediate code paths handling rvalues are removed from the code
  generators
* management of temporaries created for holding call/construction
  expression results is removed from both `jsgen` and `vmgen` -- this
  fully handled at the MIR stage now

The changes in shape also mean that some optimizations stop working --
their implementations are kept but marked as currently being dead
code.

### Code generation fixes

Some of the exposed code generator bugs either only happen with the new
MIR or only happen in fringe cases -- they're fixed without dedicated
tests being added:
* `vmgen.genDef`: pointer-backend views being assigned to locals wasn't
  considered and `genAsgnSource` handled the conversion incorrectly
* `cgen.genOpenArrayConv`: the source type being a resolved concept
  wasn't considered
* `cgen.genOpenArrayConv`: the wrong type was checked for `var`-ness,
  leading to a string not being prepared for mutations when assigning
  it to a `var openArray[char]` location
* `cgen.genSlice`: `var openArray` types weren't considered, leading to
  the `openArray` type treated as the element type and thus incorrect
  code being generated

### Test changes

The more regular structure of the trees reaching the code generator
fixes many issues caused by the code generators glossing over some tree
structures. Tests that now work are:
* `tresults.nim`
* `t02_argument_passing_mutate_bug.nim`
* `topenarray_assign_2.nim`

`tcast.nim` also works for the JS target now, but this is only by
accident.

#### Shape-based code generation logic 

In order to ensure the correct ordering of side-effects, `mirgen`
captures some lvalue in aliases. This is a problem in two cases:
* it leads to the creation of aliases of `openArray`s, which the code
  generators do not support yet. `tsplit_into_seq` and
  `tsplit_into_table` are disabled
* it interferes with the detection of `cstring` element assignment in
  `jsgen`, meaning that `x[0] = 'a'` where `x` is a `cstring` is no
  longer a compile-time error

Both regressions will be fixed once MIR lvalues are free of side
effects.

#### Casting

A `cast[UncheckedArray](x)[y]` no longer works for disabling index
checks on arrays. This behaviour is going to be restored once index
checks are emitted by `mirgen`.

#### Others

* `tassignment.nim`: stops working for the C target due to RVO
  prevention not being triggered. Changing the `x` in `raise(x)`
  already made the test fail previously
* `tconstructor_with_concept_type.nim`: not a regression -- the test
  shouldn't have worked for the VM target in the first place

----

## To-Do
- [x] merge the `devel` branch once all prerequisite PRs are merged 
- [ ] collect and publish a performance comparison
- [x] write down and make a post detailing how the next steps look like
- [x] add tests for all the known bugs that are now fixed
- [x] clean up the changes
- [x] update outdated documentation comments
- [x] write a proper commit message

## Notes for reviewer
* a suggestion of how review might be easiest is provided
[here](https://github.com/nim-works/nimskull/pull/942#issuecomment-1839471818)
* things to look out for:
  * is the documentation, especially that of the MIR itself, easy enough
to understand?
  * stale documentation
  * things that could use more descriptive names


zerbina:
I've started over with the redesign, scrapping the new design the PR description still outlines at the time of writing (11.11.2023; **EDIT:** it no longer does).

There were a few major pain points I had with the old new design:
* it still included `Region`s, which significantly restrict procedure rewrites around control-flow operations (which is needed for a CPS transformation, for example)
* it didn't fully resolve the problems I encountered while working towards moving run-time check insertion into the MIR stage (refer to #830)
* changing the code generators to use the MIR in this form (replacing the `CgNode` IR with the MIR is the goal, eventually) is doable, but not very straightforward
* the structure discards a lot of important relation information present in the original AST and source code, which is a step back from the original MIR

Thinking about how to solve all of these issues, I always circled back to a more tree-based layout, and after some initial experiments, I've decided to go that route instead.

The new new design:
* fully tree-based, no more instruction-list-mixed-with-declarative-tree structure
* goto-based control-flow, but restricted in a way that control-flow analysis is not complicated and that translating to a target language without support for unstructured `goto` (JavaScript, WASM, etc.) is easily doable
* correct-by-construction, as opposed to materializing semantics after the initial MIR code production
* explicit location lifetimes instead of the declarative `Scope` mechanism
* no `Region` support. Transformation code has to emulate this feature where needed

One of the most significant downsides is the larger MIR code size when compared to the original new design. This can be remedied through improvements to the `MirNode` type:
* storing the type on each nodes becomes obsolete (8 byte removed)
* the link to the origin information needs to be stored as part of `MirNode` (4 byte added)
* the size of the tagged union can be reduced from 16 byte to 8
* the kind can be packed together with the operand, shrinking the size of what is currently the tagged union to 4 byte

In the end, the size of a `MirNode` would be 8 byte (packed kind+operand and source information ID).

## The plan

The original new design was a lot more similar to the original MIR, thus requiring no fundamental changes. With the new new design, this is not the case, and large parts of the MIR-related code will have to be completely rewritten.

I'm trying to split the whole rework into a series of PRs, so as to make review of it easier:
1. fully change to the tree-based layout while keeping the current high-level control-flow operations. Some temporary scaffolding is going to be required and the quality of the generated code may regress
2. shrink the size of a `MirNode` and improve the data structures in general (this will most likely happen in multiple PRs)
3. perform all run-time check injection with the MIR (this one too)
4. replace the control-flow constructs in the `CgNode` IR with control-flow instruction resembling the ones the MIR will use
5. replace the existing MIR control-flow operations with goto-based operations

## Current progress

The parts I've pushed so far are changes to how the control-flow graph computations and data-flow analysis work. This decouples the data-flow analysis from the MIR's layout and is thus necessary for the latter to be changed.

In addition, the structural changes to the MIR code are also outlined already.

Do note that the changes are of prototype-stage quality.

zerbina:
I've pushed all my most recent progress. The compiler bootstraps and, judging from local testing with the `arc` and `destructor` category, the `injectdestructors` pass seems to mostly work.

The `injectdestructors` pass was heavily tailored to the previous MIR, so a large amount of changes were necessary (and not everything works yet). With the changes so far, the pass does become slower (by around 50%), though this doesn't affect overall compile times much, as the pass only accounts for a small fraction of the total time. Some improvements are still missing, but it's unlikely that they'll impact much in terms of speed.

Another thing to note is that `mirgen` now relies heavily on temporaries in order to emit code with the correct semantics:
* arguments to calls are captured in temporaries
* call results are captured in temporaries
* array index operands are captured in temporaries
* etc.

This ensures left-to-right evaluation order for all backends, but it does result in both the size and efficiency of the generated C, JavaScript, and VM code regressing. In my local testing, the NimSkull part of the third bootstrap iteration now takes around 2 seconds longer (a slowdown of ~16%).

Using the MIR as the IR the code generators operate on and introducing an optimization pass that elides temporaries where possible will restore (and improve upon) the current quality, but both aren't things planned for this PR.

### What's still missing?

There are some move-analyser-related bugs that still need to be fixed and stack-traces are also wrong for the most part (I'm currently making a separate PR with some of the necessary preparations).

A lot of the documentations is stale and needs to be updated, especially the `mir.rst` document.

I'm looking into splitting some of the unrelated fixes out into separate PRs, but some of the bugs are fringe enough that I don't think the extra overhead is warranted.

zerbina:
Also, I've changed `--expandArc` to use a rendered version of the MIR instead of the `CgNode` IR. The `--expandArc`-using tests are already updated, meaning that one can look at them to get an idea of how the MIR code currently looks like.

zerbina:
Here are some highlights from the most recent push:
* the `mir.rst` document is updated
* lvalue expressions are now allowed in argument positions, allowing for much smaller and more
  concise trees
* a MIR-to-MIR optimization pass that eliminates unnecessary temporaries/copies, fixing the
  severe slowdowns for the VM and JavaScript backends

A few separate PRs for various fixes still need to be created. Once all are merged into `devel`
and `devel` is merged into this PR. I'll create and post the a before-vs-now compiler
performance comparison.

In general, this PR is close to completion now.

zerbina:
This PR only represents the first step in the rework of the MIR. The remaining steps require additional work that isn't directly related to the rework but still depend on the changes implemented here, so I've decided to split the whole rework into multiple PRs.

The remaning three steps, which I intend to work on immediately after this PR, are:

## Decoupling copies/moves from hooks

Copying and moving is currently coupled to lifetime hooks, but this is neither necessary nor a good idea.

The plan is to introduce three new MIR operators: `Move`, `Copy`, and `Sink`, all which appear to the right of assignments.

* `Move`: the assignment doesn't copy, and the source location is optionally reset (i.e., a   destructive move is performed) if a data-flow analysis deems it necessary
* `Copy`: create a full copy
* `Sink`: either a `Move` or a `Copy`. The move-analyzer is responsible for deciding what a   `Sink` can be collapsed to

`mirgen` is responsible for placing these operators, moving more behaviour-related decision making into the AST -> MIR translation step.

All lvalue assignments for non-scalar or custom-copy-behaviour-having types are going to use the new operands, except if they're shallow assignments.

As a consequence:
* `FastAsgn` becomes obsolete and can be removed
* the move-analyzer is decoupled from lifetime hooks
* `jsgen` becomes simpler and can generate more efficient code (by translating `Copy` to `nimCopy`)
* the `injectdestructors` becomes simpler and has make less decisions
* more control over where copies or moves are forced

## Goto-based control-flow

The current MIR control-flow constructs `Block`, `If`, `Case`, `Repeat`, `Try`, `Break`, `Raise`, and `Return` are going to be replaced with:
* `Label`: destination of a jump
* `If`: conditional jump. Does nothing if the condition evaluates to `true`, jumps to the given label otherwise. Forward control-flow only
* `Case`: selects a jump target based on the operand's value. Forward control-flow only
* `Loop`: unconditional jump-back to a previous label
* `Continue` (the name is not final): dynamic control-flow. Continues intra-procedure unwinding. Forward control-flow only

Each control-flow operation is restricted in a way such that translation to structured control-flow constructs is easily possible. For calls, there is going to be the `CheckedCall` variant, which stores a list of labels to jump to when the called procedure exits via exceptional control-flow.

Since the declarative `Scope` would only complicate analysis and processing at this point, it's going to be replaced with inserting a `LocEnd` (or similar) statement for signaling that a location goes out of scope.

As a consequence:
* more complex control-flow can be expressed within the MIR
* it becomes possible to fix the destructor injection bugs/issues that exist because of the limited control-flow constructs
* control-flow-driven transformations (e.g., a CPS transformation) become much easier to implement
* the MIR becomes a flat list of statements, allowing for easier traversal/processing
* multiple `Def` statements for a single name become possible, with the related workarounds becoming obsolete
* creating a `DataFlowGraph` becomes significantly less costly, which will make using data-flow analysis in more contexts feasible

## Decoupling destruction from hooks

Same as with moving and copying, destruction should also not be coupled to lifetime hooks.

`mirgen` is going to emit `Destroy` statements during the AST -> MIR translation. One MIR pass is going to eliminate unnecessary `Destroy` statements based on a data-flow analysis, and a follow-up lowering pass is going to lower the `Destroy` statements into `=destroy` hook calls.

This will:
* allow for better control over where to destroy
* make the `DefCursor` and `DefUnpack` statements obsolete
* allow for code generators to implement destructors themselves (`vmgen` is going to do this, at least in the interim)

zerbina:
When doing a full review of the changes, I'd suggest the following approach:
1. read the `Summary` and `Overview` of the PR message and skim the rest
2. read/review the `mir.rst` document. This should provide a broad overview of how the MIR looks like now
3. review the `mirtrees` module, which contains "future direction" comments outlining how things are going to change in detail
4. review the `mirconstr` module. Looking at the diff is most likely not going to be very useful, as the module's content changes to something completely different. This should provide a basic overview of the tools `mirgen` uses
5. review the `mirgen` module. This is the module with the most fundamental changes. Looking at the small changes in `transf` also makes sense here
6. for more context, take a look at what the remaining parts of the rework are [planned to look like](https://github.com/nim-works/nimskull/pull/942#issuecomment-1836765969)
7. review, in that order, the `mirexec`, `aliasanalysis`, `analysis`, and `injectdestructors` modules
8. take a look at the `--expandArc`-using tests (e.g., `topt_cursor`, `topt_no_cursor.nim`, etc.), to get an idea of what the resulting MIR code looks like
9. review the `mirpasses` module
10. review the `cgirgen` module
11. review the remaining modules part of the compiler, which only contain bugfixes and changes to some magic and their implementation
12. take a look at change tests
13. re-read the PR message in full

Alternatively, one can step through each individual commit, most of which have a detailed commit message themselves. Do note that there are multiple merge commits in there and that some commits were split out.

compiler/mir/mirtrees.nim
```diff
@@ -673,29 +539,41 @@ iterator pairs*(tree: MirTree): (NodePosition, lent MirNode) =
...
+  while tree[i].kind != mnkEnd:
+    yield (ArgKinds(tree[i].kind), tree.operand(i))
+    i = tree.sibling(i)
+
+func findDef*(tree: MirTree, n: NodePosition): NodePosition =
+  ## Finds and returns the first definition for the name of the temporary
+  ## at node `n`. No control-flow analysis is performed.
+  let expected = tree[n].temp
+  # first, unwind until the closest statement
+  var n = n
```
saem: can the local `n` not be replaced with `result`, or would that have the code read poorly?

compiler/sem/aliasanalysis.nim
```diff
@@ -64,75 +87,137 @@ func sameIndex*(a, b: MirNode): Ternary =
...
+
+func `[]`(p: Path, i: int): PathInstr =
+  # note: the instructions were added in reverse, which we account for
+  # here
+  if p.shortLen > 0: p.short[p.shortLen - 1 - i]
+  else:              p.long[^i]
+
+proc getRoot*(tree: MirTree, n: OpValue): OpValue =
+  ## If `n` points doesn't point to a path expression, returns `n`, the root
+  ## of the path otherwise. Aliases are followed.
```
saem:
```suggestion
  ## If `n` doesn't point to a path expression, returns `n`, the root of the
  ## path otherwise. Aliases are followed.
```

zerbina:
I've pushed the removal of various now-obsolete code-generator logic. While I originally wanted to do these as a separate PR, I figured that the cleanup belongs here, also helping with assessing the impact.

The PR description is slightly outdated by now -- I think it needs an overhaul.

compiler/mir/mirgen.nim
```diff
@@ -12,6 +11,30 @@
...
+## The values of rvalue operations, calls, and construction are first captured
+## in temporaries (as the MIR doesn't support them being used as, e.g., call
+## arguments directly). Lvalues that have side-effects (e.g., index errors)
+## are captured (together with their side-effects) via MIR alias. In general,
+## all call arguments are first assigned to a temporary, except if the
+## expression is stable (in case of by-name arguments) or pure (in case of
+## non-sink by-value arguments).
+##
+## Pure expression are those that don't have side-effect and always refer to
+## the exact same value, whereas stable expression are those that don't have
```
saem, line 29: Is another way of describe "stable expression" as the amount of memory allocated, if at all, is statically known + no side-effects?

doc/mir.rst
```diff
@@ -7,155 +7,185 @@ Mid-end intermediate representation
...
+             | Consume OPERAND
+
+  CALL_EXPR = Call <Proc> CALL_ARG ...   # a static call of the provided
+                                         # procedure with the given arguments
+            | Call LVALUE CALL_ARG ...   # indirect call
+            | Magic <Magic> CALL_ARG ... # a call of a magic procedure (i.e.,
+                                         # a procedure that is either going to
+                                         # be lowered into something else, or
+                                         # one of which of which the behaviour
+                                         # cannot be represented in the MIR)
```
saem:
```suggestion
                                         # one for which the behaviour cannot be
                                         # represented in the MIR)
```

might read a bit better

saem (approved):
The change is immediately large. 😆 I went through the commits as best as I could and then focused on the tests. I think the change is good!

compiler/mir/mirtrees.nim
zerbina: I think directly modifying `result` here is fine -- I'll change it.

compiler/mir/mirgen.nim
zerbina: Not quite. Expression `x` is stable when it has no side-effects and when `addr(x)` always evaluates to the same address, regardless of when `addr(x)` is evaluated. For example:
```nim
let p1 = new int
var p2 = new int
var x = (a: 1, b: 2)

discard x.a  # stable expression, as `x` is always at the same memory address 
discard p1[] # stable expression, as `p1` always stores the same pointer value
discard p2[] # not stable, as `p2` could change
```

Stable sounded like the right term to use here, but I'm open to more fitting ones.

saem: I think stable is as good a term as any.

Thanks for the further explanation, that helped.

compiler/mir/mirgen.nim
```diff
@@ -263,163 +243,134 @@ proc typeOrVoid(c: TCtx, t: PType): PType {.inline.} =
...
+    of tyArray:
+      # static arrays; stable when no index errors are possible
+      tree[n, 1].kind == mnkLiteral and isStable(tree, arr)
+    of tyUncheckedArray:
+      # cannot raise
+      isStable(tree, arr)
+    of tyString, tySequence, tyOpenArray, tyVarargs, tyCstring:
+      false # run-time arrays; could always raise
+    else:
+      unreachable()
```
alaviss, line 301: Isn't always `false` here a better choice? `unreachable` can help you discover paths to cover but since this is more or less an optimization(?) I think a "specialization" approach is better than trying to accurately pin-point all variants.

compiler/mir/mirgen.nim
```diff
@@ -434,43 +385,175 @@ func nameNode(s: PSym): MirNode =
...
+template buildStmt(c: var TCtx, k: MirNodeKind, body: untyped) =
+  c.builder.buildStmt(k, body)
+
+template buildMagicCall(c: var TCtx, m: TMagic, t: PType, body: untyped) =
+  c.subTree MirNode(kind: mnkMagic, magic: m, typ: t):
+    body
+
+func detectKind(tree: MirTree, n: NodePosition, sink: bool): ExprKind =
+  ## Detects the kind of expression `n` (with the originating from AST `e`)
+  ## represents. `sink` informs whether expression is used in a sink context.
```
alaviss, line 422: Can you elaborate a bit on the logic behind the `sink` boolean? In particular, why is `sink` not a modifier applied to the node (ie. wrapping `sink(node)`) but must be derived by the caller?

My personal hunch is that this could be fragile due to information loss in new code paths (ie. the writer didn't take this into account), but I could be very wrong here.

compiler/mir/mirtrees.nim
```diff
@@ -56,166 +56,138 @@ type
...
-    mnkSwitch ## ``switch(x, y)``; changes the active branch of the record-case
-              ## identified by the result of the ``pathVariant`` operation used
-              ## as the `x` operand. `y` is the new discriminator value
+    # future direction:
+    # store the type of the destination within each def, assignment, etc. and
+    # then remove the type field from ``MirNode``
+
+    mnkDef       ## marks the start of existence of a local, global, procedure,
+                 ## or temporary. Supports an optional intial value (except for
+                 ## procedure definitions)
```
alaviss, line 85: I think it might be better to specialize the Def-with-value variant to flatten the conditionals.

compiler/mir/mirtrees.nim
```diff
@@ -56,166 +56,138 @@ type
...
+                 ## The source expression must not be empty
+
+    mnkFastAsgn ## assignment that cannot be rewritten into copy, move, or
+                ## hook call
+    # future direction: same as with DefCursor, remove FastAsgn
+    mnkAsgn     ## normal assignment; the destination might store a value
+                ## already. Whether the source is copied or moved depends
+                ## on the expression
+    mnkInit     ## similar to ``mnkAsgn``, but with the guarantee that the
+                ## destination is empty
```
alaviss: I think "empty" is a bad qualifier here, and something like "not initialized" is better.

My guess with this type is that you want to handle the initialization case where destruction could be avoided, but it should not be a guarantee that it will be "empty" as in 0-init.

compiler/mir/mirtrees.nim
```diff
@@ -56,166 +56,138 @@ type
...
-    mnkStdConv    ## ``stdConv(x)``; a standard conversion.Produce a new value.
-                  ## Also used for to-slice (``openArray``) conversions, in
-                  ## which case the semantics are still fuzzy.
+    mnkAddr   ## create a pointer from the provided lvalue
+    mnkDeref  ## dereference a ``ptr`` or ``ref`` value
+
+    mnkView      ## create a first-class safe alias from an lvalue
+    mnkDerefView ## dereference a first-class safe alias
+
+    mnkStdConv    ## a standard conversion. Produce a new value.
```
alaviss, line 135: Is standard here just meant built-in?

alaviss (commented):
This is just a quick skim through the MIR modules. Most of the feedback here is just questions for me to get a better view on how things are done and are not required changes.

compiler/mir/mirgen.nim
zerbina: In a sense it's an optimization, yeah. Nonetheless, from my experience, being exhaustive via `unreachable` forces one to think about all possible variations and also acts as another layer of defense against types/expressions reaching where they should not.

Thus, I don't think `false` is the better choice.

compiler/mir/mirtrees.nim
zerbina: Essentially, yes. It's terminology I reused from `PNode` (i.e., `nkHiddenStdConv` vs. `nkConv`).

In any case, the distinction is not needed anymore, and I'm going to remove `mnkStdConv` in a follow-up.

compiler/mir/mirtrees.nim
zerbina: "empty" is not a guarantee about the bit pattern of the location, only that the location doesn't store a value (in the logical sense). Nonetheless, "not initialized" is indeed more clear.

compiler/mir/mirtrees.nim
zerbina: I would keep it as-is for now, as a separate node kind would introduce more special casing than it'd remove, at least at the moment.

Once the `mnkNone`-as-an-argument usage is removed, and `mnkDefUnpack` and `mnkDefCursor` are gone too, I'd take another look at splitting the node kind into two.

compiler/mir/mirgen.nim
zerbina: Hmm, I'm not sure I understand the questions. An expression itself doesn't know how or where it's used, only the usage site does.

If you mean that there should be an `nkSink` (or similar) node for annotating an expression as appearing in a context where it should be sunken (i.e., ideally moved, copied if not possible), then I strongly agree -- this information is something that should be inferred during semantic analysis, not by the AST-to-MIR translation.

alaviss: > If you mean that there should be an `nkSink` (or similar) node for annotating expression as appearing in a context where the should be sunken (i.e., ideally moved, copied if not possible), then I strongly agree -- this information is something that should be inferred during semantic analysis, not by the AST-to-MIR translation.

Yep, that's what I meant. I take it as this is a planned change?

zerbina: Yep, from my side, it is, although not as part of this PR.

compiler/mir/mirtrees.nim
zerbina: I've changed the comment to use "not initialized" rather than "empty".

PR #943, opened 2023-10-05
refactor(mir): use proper operand queries

zerbina:
## Summary

In the MIR-using passes, replace raw indexing into the tree with proper
operand queries. This makes more of the processing node-layout
agnostic, which is a preparation for reworking the MIR.

## Details

* rename `unaryOperand` to `operand`. It's shorter, and gets the same
  meaning across
* add `mnkView` to the `SingleInputNodes` set -- it was previously
  missing
* allow using values of `NodePosition` and `OpValue` index-types as the
  arguments to `operand`. Requiring a `Operation` index-type value only
  introduced noise
* replace raw single-operand lookup (indexing into the tree with
  `i - 1`) with usage of `operand`

Querying the second operand of a `mnkPathArray` is still done via the
raw-indexing approach (`i - 2`), as usage of `operand(tree, i, 1)`
would still be rather inefficient.

### DIFF
--- a/compiler/mir/analysis.nim
+++ b/compiler/mir/analysis.nim
@@ -257,22 +257,22 @@ func computeValuesAndEffects*(body: MirTree): Values =
       # the result is always owned
       start: Owned.yes
     of mnkAddr, mnkView, mnkPathPos, mnkPathVariant:
-      inheritDecay(i, i - 1)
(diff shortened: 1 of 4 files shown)


saem (approved):
the cleaner reading queries are my favourite part. :D

PR #944, opened 2023-10-06
fix(suggest): file level cmds don't leak internal syms

bung87:
## Summary

Suggest commands related to the whole file (i.e. outline), and not
cursor positions, no longer include symbols from inside template and
generic routines.

## Details

`semIdeForTemplateOrGenericCheck`  is used to detect whether the cursor
is within a generic routine or template body, before this change it
didn't validate which suggest command was invoked ( `ideCmd` ). This
resulted in  `safeSemExpr`  being run and potentially reporting symbols
from inside such routines. This led to commands such as  `outline` 
containing symbols from the routine bodies polluting module level
results.

`semIdeForTemplateOrGenericCheck`  has now been updated to check the
command against the newly added  `ideLocCmds`  constant, which includes
commands that require cursor position tracking ( `ideSug` ,  `ideCon` , 
`ideDef` ,  `ideUse` ,  `ideDus` ), preventing the unexpected analysis
and pollution of module level results.

A regression test was added to ensure that accidentally providing cursor
tracking information to module level commands doesn't change the
results.

### DIFF
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -308,7 +308,7 @@ type
 
 const 
   IdeLocCmds* = {ideSug, ideCon, ideDef, ideUse, ideDus}
-    ## source location awared IDE commands
+    ## IDE commands requiring source locations, related `MsgConfig.trackPos`
 
 template `[]`*(conf: ConfigRef, idx: FileIndex): TFileInfo =
   conf.m.fileInfos[idx.uint32]
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -307,7 +307,8 @@ type
       unreportedErrors*: OrderedTable[NodeId, PNode]
 
 const 
-  IdeLocCmds* = {ideSug, ideCon, ideDef, ideUse, ideDus, ideHighlight} ## source location awared IDE commands 
+  IdeLocCmds* = {ideSug, ideCon, ideDef, ideUse, ideDus}
+    ## source location awared IDE commands
 
 template `[]`*(conf: ConfigRef, idx: FileIndex): TFileInfo =
   conf.m.fileInfos[idx.uint32]
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -306,6 +306,9 @@ type
     when defined(nimDebugUnreportedErrors):
       unreportedErrors*: OrderedTable[NodeId, PNode]
 
+const 
+  IdeLocCmds* = {ideSug, ideCon, ideDef, ideUse, ideDus, ideHighlight} ## source location awared IDE commands 
+
 template `[]`*(conf: ConfigRef, idx: FileIndex): TFileInfo =
   conf.m.fileInfos[idx.uint32]
 
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -167,7 +167,9 @@ proc deltaTrace(stopProc, indent: string, entries: seq[StackTraceEntry])
 template semIdeForTemplateOrGenericCheck(conf, n, requiresCheck) =
   # we check quickly if the node is where the cursor is
   when defined(nimsuggest):
-    if n.info.fileIndex == conf.m.trackPos.fileIndex and n.info.line == conf.m.trackPos.line:
+    if conf.m.trackPos.col != -1 and
+       n.info.fileIndex == conf.m.trackPos.fileIndex and
+       n.info.line == conf.m.trackPos.line:
       requiresCheck = true
 
 template semIdeForTemplateOrGeneric(c: PContext; n: PNode;
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -164,21 +164,20 @@ proc deltaTrace(stopProc, indent: string, entries: seq[StackTraceEntry])
     echo:
       "$1| $2 $3($4)" % [indent, $e.procname, $e.filename, $e.line]
 
-template semIdeForTemplateOrGenericCheck(conf, n, requiresCheck) =
-  # we check quickly if the node is where the cursor is
+template semIdeForTemplateOrGenericCheck(conf, n, cursorInBody) =
+  # use only for idetools support; detecting cursor in generic or template body
+  # if so call `semIdeForTemplateOrGeneric` for semantic checking
   when defined(nimsuggest):
-    if conf.ideCmd in {ideSug, ideCon} and
+    if conf.ideCmd in IdeLocCmds and
        n.info.fileIndex == conf.m.trackPos.fileIndex and
        n.info.line == conf.m.trackPos.line:
-      requiresCheck = true
+      cursorInBody = true
 
 template semIdeForTemplateOrGeneric(c: PContext; n: PNode;
-                                    requiresCheck: bool) =
-  # use only for idetools support; this is pretty slow so generics and
-  # templates perform some quick check whether the cursor is actually in
-  # the generic or template.
+                                    cursorInBody: bool) =
+  # provide incomplete information for idetools support in generic or template
   when defined(nimsuggest):
-    if c.config.cmd == cmdIdeTools and requiresCheck:
+    if c.config.cmd == cmdIdeTools and cursorInBody:
       #if optIdeDebug in gGlobalOptions:
       #  echo "passing to safeSemExpr: ", renderTree(n)
       discard safeSemExpr(c, n)
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -167,7 +167,7 @@ proc deltaTrace(stopProc, indent: string, entries: seq[StackTraceEntry])
 template semIdeForTemplateOrGenericCheck(conf, n, requiresCheck) =
   # we check quickly if the node is where the cursor is
   when defined(nimsuggest):
-    if conf.m.trackPos.col != -1 and
+    if conf.ideCmd in {ideSug, ideCon} and
        n.info.fileIndex == conf.m.trackPos.fileIndex and
        n.info.line == conf.m.trackPos.line:
       requiresCheck = true
--- a/nimsuggest/tests/toutline_generic.nim
+++ b/nimsuggest/tests/toutline_generic.nim
@@ -1,0 +1,11 @@
+from typetraits import supportsCopyMem
+
+proc tgeneric[T](x: var T) =
+  when supportsCopyMem(T):
+    discard
+
+discard """
+$nimsuggest --tester $file
+>outline $path/toutline_generic.nim
+outline;;skProc;;toutline_generic.tgeneric;;*;;3;;5;;"";;100
+"""
--- a/nimsuggest/tests/toutline_generic.nim
+++ b/nimsuggest/tests/toutline_generic.nim
@@ -9,7 +9,7 @@ template templ() =
   var x = 0 #[!]#
   return # <- when no cursor is provided, this triggered `x` being reported
 
-# try both with and without a specified cursor position
+# try both with and without a specified cursor position, they should be the same
 
 discard """
 $nimsuggest --tester $file
--- a/nimsuggest/tests/toutline_generic.nim
+++ b/nimsuggest/tests/toutline_generic.nim
@@ -1,11 +1,27 @@
-from typetraits import supportsCopyMem
+# Regression test for symbols within template and generic bodies being
+# wrongfully reported by `outline`.
 
-proc tgeneric[T](x: var T) =
-  when supportsCopyMem(T):
-    discard
+proc generic[T]() =
+  var x = 0 #[!]#
+  return # <- when no cursor is provided, this triggered `x` being reported
+
+template templ() =
+  var x = 0 #[!]#
+  return # <- when no cursor is provided, this triggered `x` being reported
+
+# try both with and without a specified cursor position
 
 discard """
 $nimsuggest --tester $file
+>outline $1
+outline;;skProc;;toutline_generic.generic;;*;;4;;5;;"";;100
+outline;;skTemplate;;toutline_generic.templ;;*;;9;;9;;"";;100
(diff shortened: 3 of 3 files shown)


zerbina:
The more elaborate PR message helps with understanding your reasoning better, which thus makes review easier, nice.

compiler/sem/sem.nim
```diff
@@ -167,7 +167,9 @@ proc deltaTrace(stopProc, indent: string, entries: seq[StackTraceEntry])
 template semIdeForTemplateOrGenericCheck(conf, n, requiresCheck) =
   # we check quickly if the node is where the cursor is
   when defined(nimsuggest):
-    if n.info.fileIndex == conf.m.trackPos.fileIndex and n.info.line == conf.m.trackPos.line:
+    if conf.m.trackPos.col != -1 and
```
```suggestion
    if conf.m.trackPosAttached and
```

nimsuggest/tests/toutline_generic.nim
```diff
@@ -0,0 +1,11 @@
+from typetraits import supportsCopyMem
+
+proc tgeneric[T](x: var T) =
```
zerbina: The tests needs to be reduced further. It shouldn't depend on `supportsCopyMem`, which is unrelated to the problem.

bung87:
I aware of that, just it take times, as it's much different from making PR in nim repo.

zerbina (changes requested):
Testing whether the cursor column is `-1` relies too much on details decided far away from here, which introduces hidden dependencies making the code harder to reason about and change.

It's more robust to check whether a cursor is attached at all (whole project or whole file operations don't attach a cursor).

compiler/sem/sem.nim
bung87: while looking at the `lexer.nim`  where `trackPosAttached` become `true` under condition `L.config.ideCmd in {ideSug, ideCon}`, does this mean `semIdeForTemplateOrGeneric` only used for these commands?

nimsuggest/tests/toutline_generic.nim
bung87: can't think how to detect it invoke `safeSemExpr` or not.

compiler/sem/sem.nim
zerbina: D'oh, you're right, I completely misunderstood the meaning of `trackPosAttached`, sorry.

Hm, I think the original intention with `semIdeForTemplateOrGeneric` was to allow *suggestions* within templates/generics (basic highlighting, at least of the unresolved symbols, is already taken care of through `onMarkUsed` -> `suggestSym`), so making that explicit, via:

```suggestion
    if conf.ideCmd in {ideSug, ideCon} and
```

should be fine here. @saem: What are your thoughts on this?

nimsuggest/tests/toutline_generic.nim
zerbina: You force the circumstances that previously led to `semIdeForTemplateOrGenericCheck` setting `requiresCheck` to true, and then ensure that a local within the generic is not reported as part of the `outline` command.

compiler/sem/sem.nim
bung87: I still think with col validation is better, as their line value are default to 0, we don't know it's default value or really cursor at first line. I'd like use these two conditions all

zerbina: That's a very indirect test. `semIdeForTemplateOrGeneric` doesn't care about the track position's column value nor should it -- it's interested in whether the cursor has meaning.

Testing for 'A' directly is better than testing for 'B', where 'B' is *currently* implied by 'A'. Both may yield the same result at the time, but testing for 'B' introduces the hidden expectation that 'A' continues to imply 'B'.

bung87: well,  I'll think it just test `trackPos` is valid

zerbina: Could you elaborate? I don't understand what you mean.

zerbina: Or, asked differently, why do you think that the `trackPos.col != -1` test is better than `ideCmd in {ideSug, ideCon}`?

saem: the list of commands where we need a cursor might be a bigger list: usage, definition, definition + usage, suggest, consider_parameter. I believe we _must_ have a cursor for all of those, so if the cursor isn't valid we should not act in the first place. Assuming that's correct, I think cursor validation (`trackPos` etc) should happen early, if we don't have it for the relevant command then we don't even bother executing the command. After that then `semIdeForTemplateOrGeneric` should test based on the command.

Thoughts from both of you about the above?

saem:
@bung87 the PR message is really helpful, thank you!

compiler/sem/sem.nim
bung87: > Or, asked differently, why do you think that the `trackPos.col != -1` test is better than `ideCmd in {ideSug, ideCon}`?

I'll use them all, it's not about which one is better, `trackPos.col != -1` is just check `trackPos` is valid.

nimsuggest/tests/toutline_generic.nim
bung87: no idea how to do that, as `semIdeForTemplateOrGeneric` used for `ideSug, ideCon`, how I test it under `outline`

compiler/sem/sem.nim
zerbina: > I'll use them all, it's not about which on e is better, `trackPos.col != -1` is just check `trackPos` is valid.

Like @saem said, if `trackPos` is not valid, the command shouldn't be executed in the first place.

bung87: okay

nimsuggest/tests/toutline_generic.nim
zerbina: `semIdeForTemplateOrGenericCheck` only wrongfully set `requiresCheck` to true when there's either:
* a cursor is placed in the body
* no cursor is provided to the command and there's a node with missing line information in the body

The most simple way to (currently) force a node with missing line information is to use a statement that has an optional slot filled with `nkEmpty` by the parser.

I've made a separate suggestion for how to improve the test.

nimsuggest/tests/toutline_generic.nim
```diff
@@ -0,0 +1,11 @@
...
+  when supportsCopyMem(T):
+    discard
+
+discard """
+$nimsuggest --tester $file
+>outline $path/toutline_generic.nim
+outline;;skProc;;toutline_generic.tgeneric;;*;;3;;5;;"";;100
+"""
```
zerbina:
```suggestion
# Regression test for symbols within template and generic bodies being
# wrongfully reported by `outline`.

proc generic[T]() =
  var x = 0 #[!]#
  return # <- when no cursor is provided, this triggered `x` being reported

template templ() =
  var x = 0 #[!]#
  return # <- when no cursor is provided, this triggered `x` being reported

# try both with and without a specified cursor position

discard """
$nimsuggest --tester $file
>outline $1
outline;;skProc;;toutline_generic.generic;;*;;4;;5;;"";;100
outline;;skTemplate;;toutline_generic.templ;;*;;9;;9;;"";;100

>outline $2
outline;;skProc;;toutline_generic.generic;;*;;4;;5;;"";;100
outline;;skTemplate;;toutline_generic.templ;;*;;9;;9;;"";;100

>outline $path/toutline_generic.nim
outline;;skProc;;toutline_generic.generic;;*;;4;;5;;"";;100
outline;;skTemplate;;toutline_generic.templ;;*;;8;;9;;"";;100
"""
```

zerbina:
The changes are good, but the PR message needs to be adjusted a bit:
- the title and summary should focus less on the implementation and more on the change in behaviour (generic/template bodies no longer being sometimes outlined)
- the details section should mention under what circumstances the problem appeared (you can use the test to base this on)

I can help with phrasing and grammar, but the basic structure and information needs to be there.

bung87:
updated, thank you!

zerbina:
While updating the PR message, I thought about whether this would a regression for `use`, `def`, and `highlight` in templates/generics (these commands also need the `semExpr` call), and I think it would.

However, the workings of the three commands are a bit shaky with regards to templates/generics, so maybe the regression is okay. What are your thought's on this, @saem and @bung87?

bung87:
I still have no knowledge why this need a seperate sem step `semExpr`, while `use` may works most of time as we place `suggestSym` in many slots, `def` will not work as said before. `highlight` never used in `nimlsp`. maybe need a test trigger the issue.

saem:
> However, the workings of the three commands are a bit shaky with regards to templates/generics, so maybe the regression is okay. What are your thought's on this, @saem and @bung87?

Honestly, I'm a bit confused, and it could be that the other changes are elsewhere.

My mental model:

- early within `suggest` if we don't have a valid cursor then cursor requiring commands are never run
  - error handling for this is upto `nimsuggest` or `nimskulllsp` if/as required
- that `semIdeForTemplateOrGenericCheck` should:
  - for cursor requiring commands, assume a valid cursor and set `requiredCheck` appropriately
  - for 'all usage'/global commands set `requiredCheck` to `true`
  - for commands that are neither set `requiredCheck` to `false`

Example commands (non-exhaustive):
- require cursors: suggest, consider param, usage, definition, definition + usage
- global: highlight and module modified

Thoughts/Questions:

- Right now it seems the set of commands we're guarding on is too narrow in `semIdeForTemplateOrGenericCheck`
- Is the early avoidance of running cursor requiring commands outside the scope of this PR?
- Usage is so very broken for me that there really isn't much to regress; while definition partially works and I use it often

bung87:
well, if this intent to work for two kinds of commands then can't guard on cursor outside, the suggest tool run single command one time.

zerbina:
> My mental model:

Thank you for the overview. My mental model matches yours.

> * Right now it seems the set of commands we're guarding on is too narrow in `semIdeForTemplateOrGenericCheck`

Yep, I think so too, and that's what I was trying to raise with my initial question.

> * Is the early avoidance of running cursor requiring commands outside the scope of this PR?

Personally, I'd say so, yeah.

compiler/front/options.nim
```diff
@@ -306,6 +306,9 @@ type
     when defined(nimDebugUnreportedErrors):
       unreportedErrors*: OrderedTable[NodeId, PNode]
 
+const 
+  IdeLocCmds* = {ideSug, ideCon, ideDef, ideUse, ideDus, ideHighlight} ## source location awared IDE commands
```
zerbina: Two things:
* `ideHighlight` is not a command that uses a cursor position 
* the line is too long

saem:
> well, if this intent to work for two kinds of commands then can't guard on cursor outside, the suggest tool run single command one time.

Just to make sure I understand, the reason we can't have early avoidance of running the command is not so much the avoidance part not working, but suggest runs once and unless something is marked dirty it won't run again, it'll simply use cached results. The changing of the cursor position doesn't invalidate cache.

So if we had the following module:

```nim
template a(thing: untyped) =
  thing.#[1]#

proc b[T](thing: T) =
  thing.#[2]#
```

And we ran two `sug` in a row, without any edits in between, at positions `#[1]#` and `#[2]#`, the `semIdeForTemplateOrGenericCheck` would only come into play for `#[1]#` and we'd only get cached results for `#[2]#`?

bung87:
no, the dirty file thing you described only apply to https://github.com/nim-works/nimskull/pull/946 that `ideUse, ideDus` require pre processing `allUsages`, guard on outside will like `ideCmd` in `IdeLocCmds` then validate the `col`, so the col either valid or invalid when valid invoke `semIdeForTemplateOrGeneric`, on the other hand will not.

compiler/front/options.nim
```diff
@@ -306,6 +306,10 @@ type
     when defined(nimDebugUnreportedErrors):
       unreportedErrors*: OrderedTable[NodeId, PNode]
 
+const 
+  IdeLocCmds* = {ideSug, ideCon, ideDef, ideUse, ideDus}
+    ## source location awared IDE commands
```
saem:
```suggestion
    ## IDE commands requiring source locations, related `MsgConfig.trackPos`
```

compiler/sem/sem.nim
```diff
@@ -164,19 +164,20 @@ proc deltaTrace(stopProc, indent: string, entries: seq[StackTraceEntry])
...
 template semIdeForTemplateOrGeneric(c: PContext; n: PNode;
-                                    requiresCheck: bool) =
-  # use only for idetools support; this is pretty slow so generics and
-  # templates perform some quick check whether the cursor is actually in
-  # the generic or template.
+                                    cursorInBody: bool) =
+  # provide incomplete information for idetools support in generic or template
   when defined(nimsuggest):
-    if c.config.cmd == cmdIdeTools and requiresCheck:
+    if c.config.cmd == cmdIdeTools and cursorInBody:
```
saem, line 180: I like the `cursorInBody` name, it's much better than `requiersCheck`

nimsuggest/tests/toutline_generic.nim
```diff
@@ -0,0 +1,27 @@
...
+
+proc generic[T]() =
+  var x = 0 #[!]#
+  return # <- when no cursor is provided, this triggered `x` being reported
+
+template templ() =
+  var x = 0 #[!]#
+  return # <- when no cursor is provided, this triggered `x` being reported
+
+# try both with and without a specified cursor position
```
saem:
```suggestion
# try both with and without a specified cursor position, they should be the same
```

saem (commented):
Couple of minor phrasing items.

Once you've updated the PR body to the latest ping me, thanks.

bung87:
@saem updated.

saem:
It's no where close to done, but I made a single change of moving the second bullet point from the _Summary_ to the _Details_. The _Summary_ should be short, focusing on the user facing impact. Also it seems like you wrote the _Summary_ first, instead write the _Details_ first, then summarize it.

PR #945, opened 2023-10-06
internal: reduce surface area of fatal reports

zerbina:
## Summary

Change all reporting of fatal `InternalReport` errors to use
`internalError` and remove unused `InternalReport` report kinds
representing fatal errors. This hides another report dependency, and
all
existing reporting of fatal internal reports now going through a single
procedure makes it easier to remove the `Report` usage later on.

## Details

* change `globalReport` and `localReport` usage with `rintUnreachable`
  to use `internalError`
* remove the now-unused `internalError` overloads accepting an
  `InternalReportKind`
* three occurrences in the `types` module are replaced with using
  `unreachable`, removing the construction of a partial `ConfigRef`
* the `rintUnknown`, `rintIce`, and `rintFatal` report kinds are
  removed -- they were unused, and it's unlikely that they'll become
  used again

### DIFF
--- a/compiler/ast/types.nim
+++ b/compiler/ast/types.nim
@@ -35,9 +35,6 @@ import
     modulegraphs,
   ]
 
-# TODO: switch to internalError/Assert or better yet emit the appropriate
-#       diagnostic/event/telemetry data instead, then drop this dependency
-from compiler/ast/reports_internal import InternalReport
 from compiler/ast/report_enums import ReportKind
 
 from compiler/ast/reports_sem import SemReport,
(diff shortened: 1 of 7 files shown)


compiler/ast/types.nim
saem, line 1: I believe `reports_internal` import can now be dropped?

saem (approved):
There is a good amount of reduction in noise with this change, and yeah I think it'll make reworking things all that much easier going forward.

compiler/ast/types.nim
zerbina: Yep, it can. Good call, thanks.

saem:
dat code churn ratio 🤩

PR #947, opened 2023-10-07
vm: convert `slurp` and `staticRead` to vmops

saem:
## Summary

Convert `slurp` and `staticRead` in the `system` module from magics to
VM callback operations.

## Details

Removal of the magic requires creating a conditional compilation step,
here `nimskullReworkStaticExec` conditional symbol was resued (see:
`condsym`). Two definitions of `slurp` and `staticRead` now exist, one
with the magic and one without. Without this, bootstrapping would no
longer work.

Also marked `slurp` as deprecated, `staticRead` is a far better name.
The reason for the deprecation is that `slurp` is just not as clear as
`staticRead`.

Removed `mSlurp` magic enum, and the `opcSlurp` VM opcode.

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -617,8 +617,7 @@ proc isSinkType*(t: PType): bool {.inline.} =
   t.kind == tySink
 
 const magicsThatCanRaise* = {
-  mNone, mSlurp, mParseExprToAst, mParseStmtToAst, mEcho,
-  mChckRange }
+  mNone, mParseExprToAst, mParseStmtToAst, mEcho, mChckRange }
 
 proc canRaiseConservative*(fn: PNode): bool =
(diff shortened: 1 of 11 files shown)


zerbina (approved):
Looks good to me; another compiler-specific operation decoupled from the core VM :D

Out of curiosity, what's the reason for keeping the magic and opcode? To be clear, I have no objections against them staying as part of this PR.

saem:
> Out of curiosity, what's the reason for keeping the magic and opcode? To be clear, I have no objections against them staying as part of this PR.

I was going to remove them in a separate step, but I guess I could blow them away in one go.

zerbina:
Got it, thanks.

Your call. I'm fine with both approaches.

saem:
I removed it as a separate commit, but as part of this PR.

PR #948, opened 2023-10-07
msgs: more decision making in handleReport

zerbina:
## Summary

* fix hints/warnings promoted to errors always counting towards the
  maximum error count, even if the hint/warning is currently disabled
* fix disabled warnings and hints being reported by `nimsuggest chk`

## Details

Processing of a `Report` previously worked as follows:
1. `handleReport` pre-processes the report
2. the report is passed to the report hook
3. the hook outputs the report if it determines that the report should
   be rendered and displayed
4. `handleReport` determines the error handling strategy
5. error handling is performed

This has two problems:
1. the responsibility of deciding whether a report should be output is
   left to the hook, which also means that the hook needs access to all
   state necessary to make the decision
2. error handling always takes place, even if the report is disabled

Report handling is split up, and the responsibilities are shifted:
1. if a report is disabled (`isEnabled` returns false), `handleReport`
   is a no-op
2. deciding writability is now the responsibility of `handleReport`
   (i.e., the compiler), with the report hook only invoked if the
   report is eligible for being displayed
3. computing writability no longer takes into account whether the
   report is enabled (since writeability is only decided *after* it is
   known that a report is enabled)
4. `handleReport` now handles the "forcefully enabled" case, by
   temporarily overriding the `writeln` hook with something that always
   echoes the message (this mirrors the previous behaviour of the
   report hook implementations)

In short, more responsibility regarding reports shift into a single
place (`handleReport`), with the report hook now only being invoked for
reports that should be displayed. Centralizing `Report` handling is a
step towards eventually removing/replacing `Report`s.

Since the the report hook used by `nimsuggest` doesn't consider
writeability ( `writabilityKind` ), this fixes all produced warnings
being
reported by `chk`, even those that are disabled.

Finally, disabled reports no longer triggering error handling fixes
disabled warnings promoted to errors being able to abort compilation.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -3981,17 +3981,8 @@ proc reportHook*(conf: ConfigRef, r: Report): TErrorHandling =
   ## `reportBody` for report, which then calls respective (for each report
   ## category) `reportBody` overloads defined above
   assertKind r
-  let wkind = conf.writabilityKind(r)
-  # debug reports can be both enabled and force enabled. So adding a case here
-  # is not really useful, since report writability kind does not necessarily
-  # dictate how it is written, just whether it can/must/cannot be written.
-  # xxx: the above convoluted comment brought to you by _glaring_ design flaws!
-  if wkind == writeDisabled:
-    return
-  else:
-    if wkind == writeForceEnabled:
-      echo conf.reportFull(r)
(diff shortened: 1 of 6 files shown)


saem (approved):
Nice fix, and also glad it cuts away more of the report hook mess.

PR #949, opened 2023-10-08
make the `in` magic work the same across backends

zerbina:
## Summary

Make range-checks of the left-hand operand in built-in `in` calls
affect the operations result instead of causing a `RangeDefect`: if the
operand is not in the value range expected by the `set`, the call
returns 'false', regardless of the selected backend and whether a
literal set construction is used or not.

The `ContainsConv` warning that warned about the previous behaviour is
removed.

## Details

* lower `x in y` where `x` requires a range check (`nkChckRange`) into
  `a <= x and x <= y and x in y` in `mirgen`
* the lowering has to happen in `mirgen`, as non-local control-flow
  needs to be fixed at MIR generation and the otherwise emitted
  `chckRange` call has control-flow effects
* remove the `mChckRange` handling from `ccgexprs.genInOp`
* remove the `ContainsConv` warning and the associated test
* add a test for the magic's new behaviour regarding range checks


zerbina:
The `tunittestpass.nim` test failure is due to `check` relying on a `typeKind` implementation bug, where the result of `typeKind` for a `NimNode` without type information is dependent on the bytecode `vmgen` emits.

I'll make a separate fix, both for `typeKind` and `unittest.check`.

zerbina:
I've merged the `devel` branch into this branch, fixing the test failure, as expected.

As described, applying the optimization-turned-semantic-extension in `mirgen` is necessary for consistent behaviour, but I'm not too fond of this. Since the compiler itself depends on the quirk, removing the latter would require finding and removing all problematic `in` usages first, but maybe that's a better solution?

saem:
> As described, applying the optimization-turned-semantic-extension in `mirgen` is necessary for consistent behaviour, but I'm not too fond of this. Since the compiler itself depends on the quirk, removing the latter would require finding and removing all problematic `in` usages first, but maybe that's a better solution?

I'm rushed today so I've not done a good job of internalizing the particulars of the quirk, but I think it might be better, if possible, to put a warning in place so we can do an audit and remove the sites of reliance.

saem:
I re-read the test, this time more carefully following the rules breakdown (super helpful!), I get it now.

yeah, that's busted and besides runtime warnings I don't see how else we can ferret them out, hmm.

zerbina:
> yeah, that's busted and besides runtime warnings I don't see how else we can ferret them out, hmm.

@saem: Sorry for the late reply. I've been thinking about this some, and I'm now fairly certain that removing this special case is indeed better. I'd approach this as follows (with each step being a separate PR):
* extend `semMagicsAfterOverloadResolution` with a check + associated warning where the range check would later be skipped
* audit all warnings in the repository (compiler, tools, and tests) and replace all problematic occurrences  
* remove the range-check skipping in `cgen`

Thoughts?

saem:
> * extend `semMagicsAfterOverloadResolution` with a check + associated warning where the range check would later be skipped

Ah, I didn't think this would work because the value isn't know until runtime, but we only really need to pay attention to the types I suppose. That was a missing piece for me, nice!

The overall plan sounds awesome, and if there are a lot of occurrences the second step can be spread across more people.

zerbina:
@Clyybber recently suggest a different approach: instead of reporting a warning for potentially dangerous code (like what is now implemented in the `devel` branch), the defect-raising range check could be integrated into the `in` operation itself.

That is:
```nim
var x = -1
x in {0, 1}
```
would be lowered into:
```nim
var x = -1
0 <= x and x <= 65536 and x in {0, 1}
```

---

I think that this is a good idea, as I believe it's the more intuitive behaviour for `in` (as opposed to the code being effectively undefined behaviour). The behaviour across backends becomes consistent too.

I've changed this PR to now do as described above, with `mirgen` lowering the range check into an `and` operation while ensuring that the set still appears to be still always be fully evaluated.

Since the warning is obsolete now, I've removed it again.

saem (approved):
The new semantics are far more sensible.

compiler/mir/mirgen.nim
```diff
@@ -712,6 +712,91 @@ proc genMacroCallArgs(c: var TCtx, n: PNode, kind: TSymKind, fntyp: PType) =
...
+                  # the 'in' element operand:
+                  chain(c): opParam(0, x.typ) => arg()
+                chain(c): magicCall(mInSet, n.typ) => arg()
+              c.stmts.add MirNode(kind: mnkInit)
+
+    # generate: ``tmp``
+    c.tempNode(n.typ, res)
+  else:
+    # the operation eligible for being turned into not an ``if`` chain. Emit a
+    # generic magic call
```
Clyybber:
```suggestion
    # the operation is not eligible for being turned into an ``if`` chain. Emit a
    # generic magic call
```

Clyybber (approved):

PR #950, opened 2023-10-08
fix(unittest): check not capturing operands

zerbina:
## Summary

Fix the value of some operand expressions to `not`, `==`, etc. within
`check` calls not being captured, which would result in incorrect
`check` failure/success.

### Issue description

For displaying the value of the operands in case of a test failure,
the `check` macro "captures" the values of the operands prior to
evaluating the condition expression. This is implemented in
`inspectArgs` by assigning the operand expressions to temporaries and
then accessing the temporaries.

Type expressions cannot be assigned to run-time variables, so they were
guarded against, by testing for whether the nodes have `ntyTypeDesc`
as the type kind. However, since the AST passed to `check` is
untyped, this cannot work, but so far it did, due to an implementation
bug of `typeKind`, where the value returned for `typeKind` called on a
`NimNode` that has no type results in undefined behaviour.

In this specific case, the bug resulted in the `notin {ntyTypeDesc}`
expressions always to always evaluate to `false`, which in turn
disabled capturing for some expressions.

### The solution

The decision of whether to assign the expression can only happen when
the expressions are typed. A two-phase macro could be used, but that
would require larger changes to `check`, so a different solution is
chosen: instead of an assignment, `inspectArgs` emits a call to the new
`materialize` template, passing along a generated `nksTemplate` symbol.
To access the materialized (or not) value, `inspectArgs` emits calls
with the generated template symbol.

When the `materialize` calls are later expanded, a template with the
passed-long symbol is produced. Using two `materialize` overloads, one
for `typedesc`s and for everything else, doesn't work at the moment: if
the value expression is erroneous, no `materialize` template is
invoked, which would leave the gensym'ed symbol in an improper state,
later causing the compiler to crash when it attempts to use it during
overload resolution.

In addition, for improved clarity, `inspectArgs` is changed to use a
`case` statement.

### DIFF
--- a/lib/pure/unittest.nim
+++ b/lib/pure/unittest.nim
@@ -683,6 +683,15 @@ proc untype(arg: NimNode): NimNode =
     for n in arg:
       result.add untype(n)
 
+template materialize(val: untyped, name: untyped) =
+  ## Helper template for ``check``. Captures `val` and makes it accessible
+  ## through a template named `name`.
+  when val is typedesc:
+    template name(): untyped = val
+  else:
+    let tmp = val
+    template name(): untyped = tmp
+
 macro check*(conditions: untyped): untyped =
   ## Verify if a statement or a list of statements is true.
   ## A helpful error message and set checkpoints are printed out on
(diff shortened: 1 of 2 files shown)


saem (approved):
Nice catch.

wrt longer term solution, I think stuff like this should be more generics + comptime (similar to how one might do it in Zig) than macro heavy.

PR #951, opened 2023-10-08
CI: use strict mode for compiling

zerbina:
## Summary

Extend `nimStrictMode` to error on unused imports, and enable
`nimStrictMode` when compiling the compiler and core tools during CI
runs.

The idea is to have this help with catching bugs and preventing
leftover debug code being merged, while not complicating local
development. Over time, `nimStrictMode` should promote more warnings
and hints to errors.

## Details

* treat `UnusedImport` warnings as an error when `nimStrictMode` is
  enabled
* remove the obsolete `runCI` koch command
* introduce the new `all-strict` koch command, which does the same
  as `all`, except that it enables strict mode
* when building the compiler and tools in strict mode, use an
  `--errorMax` of 3, so that compilation doesn't abort after the first
  error
* run `all-strict` instead of `all` in the CI

Due to a bug, the current csources compiler doesn't properly support
warnings/hints promoted to errors, so `nimStrictMode` is turned into a
no-op when `nimKochBootstrap` is defined (this define is set for the
first iteration of bootstrapping). This is a temporary workaround,
which can be removed after the next csources update.

### DIFF
--- a/tools/koch/koch.nim
+++ b/tools/koch/koch.nim
@@ -66,9 +66,8 @@ Boot options:
                            for bootstrapping
 
 Commands for core developers:
-  runCI                    bootstrap the compiler and build tools for release,
-                           in the context of continuous integration (CI),
-                           e.g.: github actions
+  all-strict               bootstrap the compiler and build tools for release,
+                           using strict mode for compiling
   docs [options]           generates the full documentation
   csource -d:danger        builds the C sources for installation
   pdf                      builds the PDF documentation
@@ -244,12 +243,12 @@ type
     Windows
     Unix
 
-proc buildReleaseBinaries() =
+proc buildReleaseBinaries(args = "") =
   ## Build binaries needed for creating a release
   # Boot the compiler
-  kochExec("boot -d:danger")
+  kochExec("boot -d:danger " & args)
   # Build the tools
-  buildTools()
+  buildTools(args)
 
 proc binArchive(target: BinArchiveTarget, args: string) =
   ## Builds binary archive for `target`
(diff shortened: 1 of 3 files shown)


tools/koch/koch.nim
```diff
@@ -529,32 +531,13 @@ proc testTools(cmd: string) =
...
-  # xxx: even though this is the main bottleneck, we could speedup the rest via batching with `--batch`.
-  # BUG: with initOptParser, `--batch:'' all` interprets `all` as the argument of --batch, pending bug #14343
-  nimexecFold("Run tester", "c -r --putenv:NIM_TESTAMENT_REMOTE_NETWORKING:1 -d:nimStrictMode testament/testament $# all" % batchParam)
-
-  testTools(cmd)
-
-  when not defined(bsd):
-    # the BSDs are overwhelmed already, so only run this test on the other machines:
-    kochExecFold("Boot Nim ORC", "boot -d:release --gc:orc --lib:lib")
+proc runCI() =
```
saem: might be a better name given the repurposing -- soft suggestion.

```suggestion
proc ciBuild() =
```

saem (approved):
I like the change overall, especially as it drops a bunch of confusing deadcode along the way.

tools/koch/koch.nim
```diff
@@ -66,7 +66,9 @@ Boot options:
                            for bootstrapping
 
 Commands for core developers:
-  runCI                    runs continuous integration (CI), eg: github actions
+  runCI                    bootstrap the compiler and build tools for release,
+                           in the context of continuous integration (CI),
+                           e.g.: github actions
```
alaviss: I think we should call this `all-strict` in this case.

tools/koch/koch.nim
```diff
@@ -529,32 +531,13 @@ proc testTools(cmd: string) =
...
-  when not defined(bsd):
-    # the BSDs are overwhelmed already, so only run this test on the other machines:
-    kochExecFold("Boot Nim ORC", "boot -d:release --gc:orc --lib:lib")
+proc runCI() =
+  ## Builds the compiler and tooling in the context of continuous integration
+  ## (=CI).
+  # enable strict mode in CI, which turns some warnings and hints into
+  # errors. In addition, don't abort after the first error
+  kochExec("boot -d:danger -d:nimStrictMode --errorMax:3")
+  buildTools("-d:nimStrictMode --errorMax:3")
```
alaviss: Instead of this, I think having buildReleaseBinaries taking flags is a better idea. That'd allow the build target logic to stay in one place.

zerbina:
I've renamed the command to `all-strict` and made it use `buildReleaseBinaries`.

alaviss (approved):
LGTM

saem:
whoops, didn't refresh the page with comments properly, now I'm curious how this might work.

alaviss:
You don't really have to do that, the script is tolerant to multiple requests.

But maybe I'll have to improve the feedback when you do it twice somehow

saem:
it seemed to have gotten stuck eventually, and when I checked branches there was nothing in the merge queue.

alaviss:
I think you found a Github issue. This PR should be in merge queue now instead of being stuck in auto-merge.

PR #952, opened 2023-10-09
`ideCmd` guard on `suggestExpr`

bung87:
## Summary
* limit `suggestExpr` and `suggestExprNoCheck` call to `{ideSug, ideCon, ideDef}`
* `suggestExpr` wraps around `suggestExprNoCheck`, `suggestExprNoCheck` is only
  work for `ideSug`, `ideCon`, `ideDef`, should not be called when running in other IDE commands.

## Details

---

## Notes for Reviewers

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -1773,7 +1773,7 @@ proc builtinFieldAccess(c: PContext, n: PNode, flags: TExprFlags): PNode =
   # work without now. template/tsymchoicefield doesn't like an early exit
   # here at all!
   when defined(nimsuggest):
-    if c.config.cmd == cmdIdeTools:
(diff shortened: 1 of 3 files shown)


saem (changes requested):
there are some basics that need fixing:

- use a documented constant
- the guard itself should be centralized rather than repeated everywhere

Besides performance what else goes wrong when `suggestExpr` is called for other commands?

bung87:
`compiler/sem/semexprs.nim` line `1778` `n[1].info` may cause `IndexDefect`

bung87:
hmm, after recent report system change, `checkSonsLen` should work as expected, I think this is not needed.

PR #953, opened 2023-10-11
fix(suggest): project parents non-project modules

bung87:
## Summary

`suggest` correctly falls back to treating the project module as the
parent module for out of project files/modules.

## Details

Prior to this change, querying the parent module for a file that is not
the project or a dependent module in the the project (via `import` or
`include`), resulted in a returning a file index of `0`. This resulted
in the parent being treated as the config file (typically first to get
indexed).

After this change, parent module querying, via `parentModule`, where no
parent is found results in an `InvalidFileIdx`. Since this is used as
part of suggest we continue to make progress by assuming the project
module is the parent module and continue semantic analysis.

As part of this change a regression test was added to ensure this
continues to work.

### DIFF
--- a/nimsuggest/tests/tisolated.nim
+++ b/nimsuggest/tests/tisolated.nim
@@ -1,8 +1,11 @@
-# Regression test for server initialazed with project file but query another
-# file that not indexed with project's graph. this covers below use cases
-# 1. a new file under project directory structure not be imported nor included
-#    by compiled files.
-# 2. a file that used by explicitly importing in another project.
+# Regression test for querying a file/module that is not the project file/module
+# or part of the transitive closure of all dependencies (via import or include).
+# In such a case the queried file is not indexed within the project's graph.
+#
+# Covered use cases:
+# 1. a new file under project directory structure not imported nor included by
+#    compiled files.
+# 2. a file that's used by explicitly importing in another project.
 
 discard """
 $nimsuggest --tester $file
--- a/nimsuggest/tests/tisolated.nim
+++ b/nimsuggest/tests/tisolated.nim
@@ -1,0 +1,11 @@
+# Regression test for server initialazed with project file but query another
+# file that not indexed with project's graph. this covers below usa cases
+# 1. a new file under project directory structure not be imported nor included
+#    by compiled files
+# 2. a file that used by explicitly importing in another project.
+
+discard """
+$nimsuggest --tester $file
+>def $path/tinclude.nim:7:14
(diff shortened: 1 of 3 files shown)


saem (changes requested):
Please add a test.

bung87:
I debuged this locally and test it through LSP, the isolated module `compiler/vm/compilerbridge.nim` initialized with project file `vm.nim` , then the `parentModule` procedure assuming given module whether indexed with project file via import or include, no, it does't, and the issue can be evidenced without this context, inside the `parentModule` procedure `else` branch `getOrDefault` default to `0` . so for testing this, need a fake project and initilazed with project file and query an isolated module. but I don't think this really necessary.

saem:
I don't see how it could fail outside of a suggest context, because that'd be the only time we'd get a module outside the project + transitive closure of dependent modules. Am I missing something?

There are similar tests already in `nimsuggest/tests` adding another one doesn't seem like a big deal, unless there is a simpler way to test it.

bung87:
as said it's _isolated_, a file belongs to a project in directory structure doesn't proven it been indexed in project's `graph`, in other word it either in a new graph, or rebuild the exists graph.

saem:
That's only half the condition, the file has to be queried for in order to actually trigger things.

Regardless, it still needs a test, this would have easily been caught with a test initially and prevented any regression.

nimsuggest/tests/tisolated.nim
```diff
@@ -0,0 +1,11 @@
+# Regression test for server initialazed with project file but query another
+# file that not indexed with project's graph. this covers below use cases
+# 1. a new file under project directory structure not be imported nor included
+#    by compiled files.
+# 2. a file that used by explicitly importing in another project.
```
saem: I updated the phrasing, I'm pretty sure I got the meaning right.

```suggestion
# Regression test for querying a file/module that is not the project file/module
# or part of the transitive closure of all dependencies (via import or include).
# In such a case the queried file is not indexed within the project's graph.
#
# Covered use cases:
# 1. a new file under project directory structure not imported nor included by
#    compiled files.
# 2. a file that's used by explicitly importing in another project.
```

Regression test looks good.

saem:
I've updated the PR body, just waiting to resolve the documentation update.

saem (changes requested):
Suggested some grammar fixes, if they're good to go, then I can merge this PR once they're in.

saem (approved):

PR #957, opened 2023-10-13
refactor(sem): clean up `hasUnresolvedParams`

bung87:
## Summary

Remove disabled code from the `hasUnresolvedParams` procedure. The
code has been disabled for 7 years and is unlikely to become used
again.

## Details

* remove the disabled code block
* remove the unused `flags: TExprFlags` parameter

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -52,7 +52,7 @@ proc semOperand(c: PContext, n: PNode, flags: TExprFlags = {}): PNode =
 
   if result.typ != nil:
     # XXX tyGenericInst here?
-    if result.typ.kind == tyProc and hasUnresolvedParams(result, {efOperand}):
+    if result.typ.kind == tyProc and hasUnresolvedParams(result):
       result = c.config.newError(n, PAstDiag(kind: adSemProcHasNoConcreteType))
     elif result.typ.kind in {tyVar, tyLent}:
       result = newDeref(result)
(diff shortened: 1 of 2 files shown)


zerbina:
I'm not sure I agree with their removals. They do convey additional information, and they also provide a customization point (e.g., for placing an assertion). For example, I think that `hasUnresolvedParams(s)` (the name is not entirely correct) conveys more information to the reader than a "raw" `tfUnresolved in s.flags` check.

In any case, a routine having a name longer than its implementation is by itself not a reason for removal.

To be clear, I have no problem with the disabled code being removed; removing the procedures as a whole is what I feel uneasy about.

bung87:
I agree with your points, but still problem remains `template rangeHasUnresolvedStatic` give the connection to reader that seems `range` type has special detection of `unresolved`, but it doesn't.  `proc hasUnresolvedParams` accept extro`TExprFlags` that been used in disabled code. which also gives reader unnecessary association

zerbina:
> still problem remains `template rangeHasUnresolvedStatic` give the connection to reader that seems `range` type has special detection of `unresolved`

What do you mean? I'm not sure I understand.

> `proc hasUnresolvedParams` accept extro`TExprFlags` that been used in disabled code. which also gives reader unnecessary association

Yep, the unused flag is confusing -- it should be removed together with the disabled code.

bung87:
followed your suggestions, I messed up git history, sorry.

zerbina (approved):
@bung87 I've reworded the PR message a bit and shortened the title. Once you've taken a look at the updated message and title, please ping me and I'll start the merge.

bung87:
@zerbina Thank you! nice!

PR #958, opened 2023-10-13
suggest: display name include generic parameters

bung87:
## Summary

Display names include generic parameters, resulting in generic parameter
information
being displayed for routines in IDE `outline` panel.

## Details

*  `Suggest.name`  type changed from  `ptr string`  to  `string` 
avoiding extra indirection
*  inside  `proc symToSuggest`   `Suggest.name`  includes generic
parameters for generic routines


saem (approved):
Nice change, and the PR message is easy to follow (good effort, they're improving 🎉).

I'll make some minor edits and merge this.

PR #960, opened 2023-10-13
stdlib: simplify the `genEnumCaseStmt` macro

zerbina:
## Summary

Reduce the amount of macro features the `genEnumCaseStmt` statement
macro relies on, by leveraging overload resolution and non-macro
compile-time evaluation. Using less advanced macro features helps with
the eventual re-design of the macro API.

## Details

The `genEnumCaseStmt` is used for generating a string-parsing case
statement. There were multiple problems with its implementation:
1. it relied on the type API, of which the future is uncertain
2. it duplicated the compiler's enum-to-string logic
3. it used a `static` procedural value as a parameter, which are
   currently very under-specified

The implementations is improved as follows:
* split `genEnumCaseStmt` into an outer template and inner macro
* prevent passing arbitrary `typedesc`s by using a constraint on the
  `typedesc`
* the list of enum values plus their corresponding string values are
  collected via a compile-time procedure (`collect`) instead of via
  type introspection
* for getting the tuple list into the auxiliary macro, the `collect`
  call is coerced to a static expression and the auxiliary macro uses a
  `typed` parameter
* the auxiliary macro doesn't require any type introspection; it only
  needs to iterate over the array construction expression passed to it
* the `enumFullRange` macro is moved above the `genEnumCaseStmt`
  template; the `collect` procedure needs it in order to support enums
  with holes

The immediate `genEnumCaseStmt` implementation is now independent of
the type API, enum type AST, and the compiler's enum-to-string
translation. A downside is that the "ambiguous enum" error no longer
points to the problematic field in the type.

In addition to the refactoring, the documentation of `genEnumCaseStmt`
is also slightly improved.

### DIFF
--- a/lib/std/enumutils.nim
+++ b/lib/std/enumutils.nim
@@ -10,73 +10,61 @@
 import std/macros
 from std/typetraits import OrdinalEnum, HoleyEnum
 
+macro enumFullRange(a: typed): untyped =
+  newNimNode(nnkCurly).add(a.getType[1][1..^1])
+
 # xxx `genEnumCaseStmt` needs tests and runnableExamples
 
-macro genEnumCaseStmt*(typ: typedesc, argSym: typed, default: typed,
-            userMin, userMax: static[int], normalizer: static[proc(s :string): string]): untyped =
-  # Generates a case stmt, which assigns the correct enum field given
-  # a normalized string comparison to the `argSym` input.
-  # string normalization is done using passed normalizer.
-  # NOTE: for an enum with fields Foo, Bar, ... we cannot generate
-  # `of "Foo".nimIdentNormalize: Foo`.
-  # This will fail, if the enum is not defined at top level (e.g. in a block).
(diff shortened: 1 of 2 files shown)


lib/std/enumutils.nim
```diff
@@ -10,73 +10,61 @@
...
+  result = nnkCaseStmt.newTree(newCall(normalizer, argSym))
+  for it in list.items:
+    expectKind it, nnkTupleConstr
+
+    let fStr = it[1].strVal
+    if fStr notin seen:
+      result.add nnkOfBranch.newTree(it[1], it[0])
+      seen.add fStr
+    else:
+      error("Ambiguous enums cannot be parsed, field " & $fStr &
```
saem, line 37: this is a reasonable compromise, a move towards pure enums everywhere will likely deal with this issue.

saem (approved):
this is a nice change, the fact that we can start pushing these facilities in the compiler is really promising. 🎉

PR #961, opened 2023-10-14
fix(sigmatch): AST corruption of bracket calls

zerbina:
## Summary

Fix explicit bracket calls (`[](...)`) and bracket expressions in
generic procedures or templates causing AST corruption.

A concrete example where this caused problems is when a bracket call is
used as the argument to an `untyped` parameter of an overloaded
template where there exist overloads with typed parameters but none
of them matches.

### Problem description

For bracket calls, the overload set contains the `mArrGet` or `mArrPut`
magics -- these always match.

For `mArrGet`  and `mArrPut` candidates, `sigmatch.matches` assigns the
*original* call AST, which is a problem when the candidate is the one
picked in the end: `semResolvedCall` expects a call AST that it can
freely modify, but since the original AST is used, the modifications
show up in unintended places (such as on the AST for earlier processed
candidates).

Due to bracket expression (`nkBracketExpr`) rewriting performed for the
bodies of templates and generic procedures, bracket expressions within
those were also affected.

### The solution

In order to uphold the expectation that the immediate call AST of a
`TCandidate`  can be modified if the candidate represents a match,
change
`matches` to create a shallow copy of the call node for a `mArrGet`/
`mArrPut` match.

### DIFF
--- a/tests/lang_callable/overload/tbracket_call_corruption.nim
+++ b/tests/lang_callable/overload/tbracket_call_corruption.nim
@@ -6,7 +6,7 @@ proc op(x: int) =
   discard
 
 template overload(x: untyped) =
-  # this is the overload that is picked. It's definition has to come
+  # this is the overload that is picked. Its definition has to come
   # first, so that it's processed first during overload resolution
   op(x) # this failed with an "expected `int` but got `T`" error
 
--- a/tests/lang_callable/overload/tbracket_call_corruption.nim
+++ b/tests/lang_callable/overload/tbracket_call_corruption.nim
@@ -1,0 +1,30 @@
+discard """
+  description: "Regression test for `[]` calls causing AST corruption"
+"""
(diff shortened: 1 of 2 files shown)


tests/lang_callable/overload/tbracket_call_corruption.nim
```diff
@@ -0,0 +1,30 @@
+discard """
+  description: "Regression test for `[]` calls causing AST corruption"
+"""
+
+proc op(x: int) =
+  discard
+
+template overload(x: untyped) =
+  # this is the overload that is picked. It's definition has to come
```
saem:
```suggestion
  # this is the overload that is picked. Its definition has to come
```

saem (approved):
One _very_ soft suggestion. Kinda surprised how long this big has lurked around.

PR #963, opened 2023-10-14
refactor(lookups): simpler identifier error handling

zerbina:
## Summary

Simplify identifier error handling in `qualifiedLookUp`, removing a
duplication of how errors for potentially quoted identifier are
produced.

## Details

* replace the `isNotFound` test and subsequent `nkError` creation with
  directly using the `errNode` returned by `considerQuotedIdent`
* remove the now obsolete `errorExpectedIdentifier` procedure

There is a small difference in the "wrong node" selection now, but the
previous logic in `qualifiedLookUp` was incorrect (the erroneous node
is the `nkAccQuoted`, not its element), so this is an improvement in
error precision.

### DIFF
--- a/compiler/sem/lookups.nim
+++ b/compiler/sem/lookups.nim
@@ -782,20 +782,6 @@ proc lookUp*(c: PContext, n: PNode): PSym =
   when false:
     if result.kind == skStub: loadStub(result)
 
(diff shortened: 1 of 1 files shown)


saem (approved):
nice ratio

PR #964, opened 2023-10-15
cgir: split up `cnkBracketAccess`

zerbina:
## Summary

Split the `cnkBracketAccess` operation into array-like access
(`cnkArrayAccess`) and tuple access (`cnkTupleAccess`). This reduces
the amount of dispatching on type kinds in the code generators.

## Details

* add the `cnkTupleAccess` and `cnkArrayAccess` nodes/operations; the
  `cnkBracketAccess` one is removed
* update all usage sites of the now-removed `cnkBracketAccess`
* translate all MIR positional field access into `cnkTupleAccess`; this
  ensures that tuple access always uses the same code generation logic,
  and it also allows for removing the tuple handling from `cgen`'s
  record-access logic (at least for now)
* adjust the `mOffsetOf` lowering in `cgen` to consider tuple access
* remove obsolete (`tyRef` and `tyPtr`) and sometimes unnecessarily
  broad (`abstractVarRange` instead of `abstractVar`) type-skipping
  from the code generation logic for array-like access

### DIFF
--- a/compiler/backend/ccgcalls.nim
+++ b/compiler/backend/ccgcalls.nim
@@ -41,7 +41,7 @@ proc reportObservableStore(p: BProc; le, ri: CgNode) =
         # it through a procedure that eventually raises is also an observable
         # store
         return inTryStmt and sfUsedInFinallyOrExcept in p.body[n.local].flags
-      of cnkFieldAccess, cnkBracketAccess, cnkCheckedFieldAccess:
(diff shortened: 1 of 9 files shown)


saem (approved):
The greater linearity of control flow is nice and I like the fact that we're not "losing" information of earlier analysis by storing it in a more generalized `cnkBracketAccess`.

PR #965, opened 2023-10-16
render generic parameters for proc types

bung87:
## Summary

Render a procedural type's generic parameters instead of
`[*missing parameters*]`. Whether required generic parameters
weren't provided is nothing the type-renderer knows about.

## Details

Use `renderTree` on the type owner's generic parameter slot AST. The
`tfUnresolved` flag can only be present when a `tyProc` type
belongs to generic routines, as other proc type definitions
themselves cannot be parameterized (only the generic type can be),
so no additional check for whether `owner` is a routine symbol is
necessary.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -651,7 +651,7 @@ type
     tfPacked
     tfHasStatic
     tfGenericTypeParam
-    tfImplicitTypeParam
+    tfImplicitTypeParam  ## generic or concept inferred from formal parameter
     tfInferrableStatic
     tfConceptMatchedTypeSym
     tfExplicit        ## for typedescs, marks types explicitly prefixed with the
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -651,7 +651,8 @@ type
     tfPacked
     tfHasStatic
     tfGenericTypeParam
-    tfImplicitTypeParam  ## generic or concept inferred from formal parameter
+    tfImplicitTypeParam ## the type parameter was lifted from the routine's
+                        ## formal parameters
     tfInferrableStatic
     tfConceptMatchedTypeSym
     tfExplicit        ## for typedescs, marks types explicitly prefixed with the
--- a/compiler/ast/typesrenderer.nim
+++ b/compiler/ast/typesrenderer.nim
@@ -388,7 +388,15 @@ proc typeToString*(typ: PType, prefer: TPreferedDesc = preferName): string =
                  else: "proc "
               else:
                 "proc "
-      if tfUnresolved in t.flags: result.add "[*missing parameters*]"
+      if isGenericRoutineStrict(t.owner):
+        let params = t.owner.ast[genericParamsPos]
+        let len = params.safeLen
+        var genericParams = if len > 0: "[" else: ""
+        for i in 0 ..< len:
+          genericParams.add getPIdent(params[i]).s
+          if i < len - 1: genericParams.add(", ")
+        if len > 0: genericParams.add "]"
+        result.add genericParams
       result.add "("
       for i in 1..<t.len:
         if t.n != nil and i < t.n.len and t.n[i].kind == nkSym:
--- a/compiler/ast/typesrenderer.nim
+++ b/compiler/ast/typesrenderer.nim
@@ -388,7 +388,7 @@ proc typeToString*(typ: PType, prefer: TPreferedDesc = preferName): string =
                  else: "proc "
               else:
                 "proc "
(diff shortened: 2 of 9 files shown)


zerbina:
> * `[*missing parameters*]` is helpful when reporting errors, however it should
>       be replaced by precise error point out why the parameter is unresolved.

The `tfUnresolved` flag for a `tyProc` indicates that the type has freestanding generic parameters (i.e., ones that don't appear in the `tyProc` itself). It is unrelated to errors, so there's nothing to replace the rendering with.

compiler/ast/ast_types.nim
```diff
@@ -651,7 +651,7 @@ type
     tfPacked
     tfHasStatic
     tfGenericTypeParam
-    tfImplicitTypeParam
+    tfImplicitTypeParam  ## generic or concept inferred from formal parameter
```
zerbina:
```suggestion
    tfImplicitTypeParam ## the type parameter was lifted from the routine's
                        ## formal parameters
```
Implicit type parameters are not inferred -- they're lifted.

compiler/ast/typesrenderer.nim
```diff
@@ -154,6 +154,24 @@ proc rangeToStr(n: PNode): string =
   assert(n.kind == nkRange)
   result = valueToString(n[0]) & ".." & valueToString(n[1])
 
+proc genericParamsToStr*(n: PNode): string =
```
zerbina: `renderer.renderTree` already supports rendering `nkGenericParams` trees -- that should be used instead.

compiler/ast/typesrenderer.nim
```diff
@@ -388,7 +406,8 @@ proc typeToString*(typ: PType, prefer: TPreferedDesc = preferName): string =
                  else: "proc "
               else:
                 "proc "
-      if tfUnresolved in t.flags: result.add "[*missing parameters*]"
+      if isGenericRoutineStrict(t.owner):
+        result.add genericParamsToStr(t.owner.ast[genericParamsPos])
```
zerbina:
```suggestion
        result.add renderTree(t.owner.ast[genericParamsPos])
```

bung87:
I didn't mean that replace the rendering, eg. `tests/errmsgs/t5167_4.nim` `proc (x: string)` is mismatched indeed, `proc [B](x: int)` mismatched because there's error (missing generic params)

zerbina (changes requested):
Nice. I think rendering the parameters instead of the not-very-descriptive `*missing parameters*` is better. The parameter rendering should not be duplicated, however.

What's your opinion on only rendering the generic parameters when there are freestanding ones? Always rendering them looks to be a bit noisy.

compiler/ast/ast_types.nim
bung87: can we mention `generic or concept` here?

zerbina: I don't see why. `generic` doesn't make sense in this case, and concepts aren't anymore special than other generic types.

zerbina: What do you think of the following?
```suggestion
    tfImplicitTypeParam ## the type parameter was lifted from a generic type
                        ## used by a formal parameter
```

bung87:
> What's your opinion on only rendering the generic parameters when there are freestanding ones? Always rendering them looks to be a bit noisy.

am not sure, in some cases the generic param symbol same as formal param type, in `tests/lang_callable/closure/tinvalidclosure.nim` the `proc [T](x: int)` point out it is generic procedure and doesn't have duplicates

compiler/ast/ast_types.nim
bung87: ok

zerbina:
> > What's your opinion on only rendering the generic parameters when there are freestanding ones? Always rendering them looks to be a bit noisy.
> 
> am not sure, in some cases the generic param symbol same as formal param type, in `tests/lang_callable/closure/tinvalidclosure.nim` the `proc [T](x: int)` point out it is generic procedure and doesn't have duplicates

That's a bug with `instantiateProcType`. The `tyProc` type's owner is not updated to the instantiated generic routine.

zerbina (approved):

zerbina:
@bung87 I've reworded and improved the PR message and title a bit.

In the future, please make sure to note down non-obvious but important details in the details section. It's not immediately clear why no additional check of the `owner` symbol is required.

Once you've taken a look at updated PR message and title and are okay with them, please ping me and I'll start the merge.

bung87:
yeah, `isGenericRoutineStrict` used in my previous commit, so I thought no need mention intermediate code, `tfUnresolved` only available in generic routines.

PR #966, opened 2023-10-16
refactor(sem): remove `fixupTypeAfterEval`

zerbina:
## Summary

Remove the obsolete `fixupTypeAfterEval` procedure and all its usages.

## Details

`fixupTypeAfterEval` was originally required for ensuring the value AST
returned by the VM is properly typed, but with the introduction of
`vmcompilerserdes` (https://github.com/nim-works/nimskull/pull/242),
this became obsolete.

### Removed usages

Normal procedures cannot return `untyped`, `typed`, or `typeDesc`
values, meaning that the  `fixupTypeAfterEval`  in  `evalAtCompileTime` 
was
always a no-op.

For `static` expression, `typeDesc` values are possible, but since the
evaluated result coming out of `evalStaticExpr` for those is always a
`nkType` AST, for which `semExpr` is a no-op, `fixupTypeAfterEval` was
also a no-op there.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -1068,28 +1068,6 @@ proc fixVarArgumentsAndAnalyse(c: PContext, n: PNode): PNode =
 
 include semmagic
 
-proc fixupTypeAfterEval(c: PContext, evaluated, eOrig: PNode): PNode =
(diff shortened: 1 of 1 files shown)


saem (approved):
nice deduction, it really didn't seem like a no-op

saem:
FYI: typo edit to the PR body.

PR #968, opened 2023-10-16
internal: remove the double-meaning of `mMove`

zerbina:
## Summary

Don't use a special `move` magic form for the `seq`/`string` sink
implementation, but instead emit the required code directly. This moves
decision making out the code generators, fixes a MIR-level issue, and
is a preparation for MIR improvements.

## Details

For the `=sink` implementation of `seq`s and `string`s,
`liftdestructors` emitted a special `mMove` call that the C code
generator then expanded into the C equivalent of:

```nim
if not samePayload(a, b):
  =destroy(a)
a = b
```

The problem is with how the generated `move` call looked:
`move(a, b, =destroy(a))`. The `CgNode` IR for the call has to have the
same shape, but this required special-casing in `mirgen` and the
resulting MIR code also doesn't properly represent the actual
behaviour (the `=destroy(a)` call is not unconditionally evaluated),
which could interfere with MIR analysis.

In order to correctly represent the behaviour at the MIR level, the
`liftdestructors` lifting emits the above code directly, without
overloading the `mMove` magic. Since how the payload comparison works
is currently backend dependent, the new internal-only `mSamePayload`
magic is introduced, which the C generator then lowers into a
comparison of the payload pointers.


saem (approved):
IIUC we're no longer conditionally dispatching based on the len/shape of the call, instead we lower it once with the correct shape and there is a single unambiguous magic that `ccgexprs` can then further specialize.

zerbina:
> IIUC we're no longer conditionally dispatching based on the len/shape of the call, instead we lower it once with the correct shape and there is a single unambiguous magic that `ccgexprs` can then further specialize.

Yep, that's exactly what happens.

PR #970, opened 2023-10-18
suggest: only `outline` current file symbols

bung87:
## Summary
* Change the `outline` command to only list definitions from a
  *document* (instead of from a *module*)
* This means that symbols from `include`d modules are no longer
  listed by the command

## Details
* When deciding whether to include a symbol in the list, don't search
  for the outermost module -- only inspect the symbol's immediate file
  index
* Adjust `nimsuggest/tests/tinclude.nim`

Two refactors are applied to `suggestSym`:
* Add the `symToSuggestGlobal` template for replacing the
  `suggestSym(isLocal=false)` calls sharing most of their arguments.
  This removes duplicated code and also makes the lines shorter
* The main `if` statement is replaced with a `case` statement, as
  this makes the code clearer and shorter

### DIFF
--- a/compiler/tools/suggest.nim
+++ b/compiler/tools/suggest.nim
@@ -199,6 +199,12 @@ proc symToSuggest(g: ModuleGraph; s: PSym, isLocal: bool, section: IdeCmd, info:
                     else:
                       getTokenLenFromSource(g.config, s.name.s, infox)
 
+template symToSuggestGlobal(g: ModuleGraph; s: PSym, section: IdeCmd,
+                            info: TLineInfo;
+                            useSuppliedInfo = false): Suggest =
+  symToSuggest(g, s, isLocal = false, section, info, 100,
+               PrefixMatch.None, inTypeContext = false, 0, useSuppliedInfo)
+
 proc suggestResult(conf: ConfigRef; s: Suggest) =
   if not isNil(conf.suggestionResultHook):
     conf.suggestionResultHook(s)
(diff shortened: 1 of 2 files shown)


compiler/tools/suggest.nim
```diff
@@ -502,19 +502,9 @@ proc suggestSym*(g: ModuleGraph; info: TLineInfo; s: PSym; usageSym: var PSym; i
...
-        parentModule = s.owner
-      while parentModule != nil and parentModule.kind == skModule:
-        parentFileIndex = parentModule.info.fileIndex
-        parentModule = parentModule.owner
-
-      if parentFileIndex == conf.m.trackPos.fileIndex:
-        suggestResult(conf, symToSuggest(g, s, isLocal=false, ideOutline, info, 100, PrefixMatch.None, false, 0))
+    elif conf.ideCmd == ideOutline and isDecl and info.fileIndex == conf.m.trackPos.fileIndex:
+      if "`gensym" in s.name.s: return # prevent template generated symbol show up, xxx: avoid string comparison 
+      suggestResult(conf, symToSuggest(g, s, isLocal=false, ideOutline, info, 100, PrefixMatch.None, false, 0))
```
zerbina: Please clean up this code, it's hard to read. Lines should have a maximum length of 80 characters.

zerbina (changes requested):
I think the change for `gensym`s should go into a separate PR (and also include tests), but apart from that, the implementation looks good to me.

compiler/tools/suggest.nim
```diff
@@ -495,26 +495,20 @@ proc suggestSym*(g: ModuleGraph; info: TLineInfo; s: PSym; usageSym: var PSym; i
...
-        parentFileIndex = info.fileIndex # assume we're in the correct module
-        parentModule = s.owner
-      while parentModule != nil and parentModule.kind == skModule:
-        parentFileIndex = parentModule.info.fileIndex
-        parentModule = parentModule.owner
-
-      if parentFileIndex == conf.m.trackPos.fileIndex:
+    of ideHighlight:
+      if info.fileIndex == conf.m.trackPos.fileIndex:
+        suggestResult(conf, symToSuggest(g, s, isLocal=false, ideHighlight, info, 100, PrefixMatch.None, false, 0))
```
```suggestion
        suggestResult(conf):
          symToSuggest(g, s, isLocal=false, ideHighlight, info, 100,
                       PrefixMatch.None, false, 0)
```

bung87: added `symToSuggestGlobal` and use that now.

zerbina:
@bung87 The changes look good, and I think that the `case` statement reads much better.

Please also mention the `case` change in the details section. While there exists a dedicated commit for the change (which is good), the merged commit will be a squashed one, meaning that the information is not available without looking at the PR on GitHub.

zerbina (approved):
I've rephrased the PR message a bit, putting more emphasis on what changes. Once you've taken a look, please ping me and I'll start the merge process.

bung87:
@zerbina Thank you! I've read the changes.

PR #971, opened 2023-10-18
ci: use treeless clones

alaviss:
## Summary
With the release of actions/checkout@v4.1.0, partial clone is now
supported.

This feature allows us to clone only the latest files but with all of
git history attached, giving us full commit tags and the lower cost of 
`--depth=1` .

At the time of writing, this dropped CI clone time from over 60s to just
3-9s.

For a primer on treeless clones, see:
https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,6 +118,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          filter: tree:0
 
       - name: Enable annotations
         run: echo "::add-matcher::.github/nim-problem-matcher.json"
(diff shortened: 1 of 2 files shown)


alaviss:
/merge

Finally faster clones are in

alaviss:
/merge

Fixed permission for the app

zerbina:
Using the [action](https://github.com/nim-works/nimskull/actions/runs/6553638541) for the merge of the commit this PR is based on as reference, cloning the repository previously took around 60 - 100 seconds for each job, while it now only takes between 3 to 10 seconds!

Very cool, thank you!

**Edit:** I didn't refresh the page prior to publishing the comment, sorry.

PR #973, opened 2023-10-18
fix the `ConversionConv` warnings

zerbina:
## Summary

Fix all `ConversionConv` warnings in the compiler and standard library
and most of the warnings in the tests (some are false positives). This
is a preparation for disabling the range check omission for `in` calls.

## Details

This is a best-effort fix, the warnings were found by:
* searching and going over all `in`, `notin`, and `contains` calls
  that use a set construction expression
* searching for potential usage in macros by looking for `"in"`,
  `"notin"`, and `"contains"` usage
* temporarily turning the warning into an error and running the whole
  test suite

A lot of problematic `in` usages were found and resolved, but it's
possible that some remain.

How the problematic `in` calls are resolved depends on the usage:
* if the element operand's range fits into the set, the constructed set
  is coerced to the right type (which removes the implicit conversion)
* `in` calls using single-slice sets (`{a..b}`) are changed to use the
  slice directly
* `in` calls with small sets use `or` chains (which is also what the
  code would currently be lowered to in the end)
* others usages are changed to use `case` statements

### DIFF
--- a/compiler/ast/reports.nim
+++ b/compiler/ast/reports.nim
@@ -243,40 +243,44 @@ template reportHere*[R: ReportTypes](report: R): R =
 
     tmp
 
+template `in`[T](val: T, r: typedesc[range]): bool =
+  let v = val
+  v >= low(r) and v <= high(r)
+
 func wrap*(rep: sink LexerReport): Report =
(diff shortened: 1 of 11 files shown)


saem (approved):
code reads better for it.

PR #976, opened 2023-10-18
fix(sigmatch): built-in type-class constraint issue

zerbina:
## Summary

Fix overload resolution failing for procedures with signatures like
`proc p[A, B: <built-in-type-class>](a: A, b: B)` where the types used
as the arguments are not equal.

## Details

For type relation (`typeRel`) analysis between a type and a generic
type parameter constraint, binding types to type-classes is disabled
(the `trDontBind` flag is passed), but the logic for
`tyBuiltInTypeClass` ignored the request and always bound the actual
type to the type-class.

Since in `[A, B: <built-in-type-class>]`, the `tyBuiltInTypeClass`
`PType` instance used as the constraints is the same, once `A` was
bound, `B` effectively acted as being constrained to `A`, which caused
a type relation failure when the type passed to `B` is not equal to
that passed to `A`.

The `typeRel` logic for `tyBuiltInTypeClass` now respects the
`trDontBind` flag, fixing the issue. However, not all built-in type-
classes use the `tyBuiltInTypeClass` type at the moment (fixing that is
outside the scope of this commit), so for those that don't, the issue
persists.

### DIFF
--- a/tests/lang_types/typeclass/ttypeclass_multi_use.nim
+++ b/tests/lang_types/typeclass/ttypeclass_multi_use.nim
@@ -49,7 +49,7 @@ doAssert not compiles(test({1}, {'2'}))
 doAssert test(range[0..1](0), range[3..4](3)) == 8
 doAssert test((ptr int)(nil), (ptr float)(nil)) == 9
 doAssert test((ref int)(nil), (ref float)(nil)) == 10
-doAssert not compiles(test(@[1], @[""]))
+doAssert not compiles(test(@[1], @[""])) # == 11
 doAssert test((proc(a: int))(nil), (proc(a: float))(nil)) == 12
-doAssert not compiles(test1(arr1, arr2))
+doAssert not compiles(test1(arr1, arr2)) # == 13
 doAssert not compiles(test1((ptr UncheckedArray[int])(nil), (ptr UncheckedArray[float])(nil)))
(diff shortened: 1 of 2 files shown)


Clyybber (approved):

tests/lang_types/typeclass/ttypeclass_multi_use.nim
```diff
@@ -0,0 +1,55 @@
...
+doAssert not compiles(test(arr1, arr2))
+doAssert test((1,), ("",)) == 5
+doAssert test(Object1(), Object2()) == 6
+doAssert not compiles(test({1}, {'2'}))
+doAssert test(range[0..1](0), range[3..4](3)) == 8
+doAssert test((ptr int)(nil), (ptr float)(nil)) == 9
+doAssert test((ref int)(nil), (ref float)(nil)) == 10
+doAssert not compiles(test(@[1], @[""]))
+doAssert test((proc(a: int))(nil), (proc(a: float))(nil)) == 12
+doAssert not compiles(test1(arr1, arr2))
```
saem:
```suggestion
doAssert not compiles(test(@[1], @[""])) # == 11
doAssert test((proc(a: int))(nil), (proc(a: float))(nil)) == 12
doAssert not compiles(test1(arr1, arr2)) # == 13
```

very soft suggestion, just a tad bit easier to draw the correspondence.

saem (approved):
One very soft suggestion, feel free to merge without it.

PR #977, opened 2023-10-19
Reduce TNode size by removing unused node flags

Clyybber:
## Summary
* Removed two node flags unused after the VM rework, which reduced the
size of  `TNodeFlags`  to 2 bytes
* Moved the  `id`  field below  `typ`  because  `typ`  requires an
alignment of 8 bytes, while  `id`  has a size of 8 bytes
* The two changes above combined result in reducing the size of  `TNode` 
from 48 to 40 bytes on a x86_64 system

## Details
* More node flags can be removed/moved from  `TNodeFlag` , but that
won't reduce the size of  `TNode`  further (unless they are reduced to
only 8 node flags), so I'll make a seperate PR for those

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1533,8 +1533,9 @@ type
         discard
 
   TNode*{.final, acyclic.} = object # on a 32bit machine, this takes 32 bytes
+                                    # on a 64bit machine, this takes 40 bytes
     typ*: PType
-    id*: NodeId
+    id*: NodeId  # placed after `typ` field to save space due to field alignment
     info*: TLineInfo
     flags*: TNodeFlags
     case kind*: TNodeKind
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -604,8 +604,6 @@ type
     nfDotField  ## the call can use a dot operator
     nfDotSetter ## the call can use a setter dot operarator
     nfExplicitCall ## `x.y()` was used instead of x.y
-    nfIsRef     ## this node is a 'ref' node; used for the VM
-    nfIsPtr     ## this node is a 'ptr' node; used for the VM
     nfFromTemplate ## a top-level node returned from a template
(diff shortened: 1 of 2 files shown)


zerbina:
Ah, very nice catch. If you move the `id` field below `info` field, that should get the size down to 32 bytes.

Clyybber:
I moved it below the `info` field initially, as I had the same thought, but it doesn't have an effect on the size, since `info` has a size of 8 but only an alignment requirement of 4.
Reducing it to 32 bytes is impossible unfortunately because TNodeSeq is 16 bytes on it's own.
On the upside that means there's some unused space in the other variant branches into which the nfBaseX flags can be moved into (doing that in the seperate PR)

compiler/ast/ast_types.nim
```diff
@@ -1535,8 +1533,8 @@ type
         discard
 
   TNode*{.final, acyclic.} = object # on a 32bit machine, this takes 32 bytes
-    id*: NodeId
     typ*: PType
+    id*: NodeId
```
```suggestion
    id*: NodeId  # placed after `typ` field to save space due to field alignment
```

soft suggestion, just in case someone gets excited and wants to reorder the fields -- aka future me.

compiler/ast/ast_types.nim
```diff
@@ -1535,8 +1533,8 @@ type
         discard
 
   TNode*{.final, acyclic.} = object # on a 32bit machine, this takes 32 bytes
```
zerbina, line 1535:
```suggestion
  TNode*{.final, acyclic.} = object # on a 32bit machine, this takes 32 bytes
                                    # on a 64bit machine, this takes 40 bytes
```
Only a very soft suggestion, feel free to ignore.

saem (approved):
one soft suggestion to reduce the chance of future regression.

zerbina (approved):

saem:
Nice change, I'm curious if we'll notice a difference in CI times.

Clyybber:
Thanks for the reviews/suggestions!

PR #979, opened 2023-10-19
tools: port `dust` a source module minimizer

saem:
## Summary

Add a compiler error message minimizer executable to the compiler
distribution, called  `dust` . Given a compiler error message, 
`dust $compiler_options $module_to_reduce`  will reduce said module's
code in place until a minimal module that generates the same error is
produced.

## Details

`dust`  is a new executable that ships with NimSkull, that minimizes a
module down to the smallest failing module. This is useful for those
submitting bugs, allowing for easy reduction in test cases size, or
those looking to reduce failing code to better understand it.

It can be run by calling  `dust <file_to_minimize.nim>` , it will then
reduce the file to the minimal failing case. This is an inplace edit --
so be sure to save a copy if you want to keep the original.

In addition to the introduction of  `dust` , a dedicated  `tester`  also
exists meant to facilitate testing of the tool, it's a simple driver for
the  `dust`  executable.

### Porting Details

`dust`  was originally written by @disruptek, this ported version has a
few key differences:

1. It works on NimSkull and ships with the compiler, ensuring we're
always reducing with a like for like compiler
2. The search algorithm and AST mutators are much faster (greater than
70x)
3. Includes a `tester` that can be run via CI


tools/dust/spec.nim
```diff
@@ -0,0 +1,84 @@
...
+  compiler / ast / [ lineinfos, renderer, ast ]
+
+import hashing
+
+const
+  unsunny* = {nkCharLit..nkUInt64Lit} +
+             {nkFloatLit..nkFloat128Lit} +
+             {nkStrLit..nkTripleStrLit} +
+             {nkSym, nkIdent}
+
```
Clyybber:
```suggestion
```
I believe this should essentially be `nkWithoutSons` barring `nkEmpty, nkNone, nkType, nkNilLit, nkError` which I don't think are omitted on purpose and is also used in this way with `notin` below.

tools/dust/spec.nim
```diff
@@ -0,0 +1,84 @@
...
+proc count*(r: Remains): int = len(r.signatures)
+
+proc pop*(remains: var Remains): PNode =
+  assert len(remains) > 0, "pop from empty remains"
+  result = pop(remains.attempts).node
+
+proc size*(n: PNode): int =
+  assert not n.isNil
+  result = 1
+  if n.kind notin unsunny:
```
Clyybber:
```suggestion
  if n.kind in nkWithSons:
```

tools/dust/mutate.nim
```diff
@@ -0,0 +1,71 @@
...
+    result = shallowCopy n
+    for i, child in pairs(n):
+      result.sons[i] = transform(child, fn)
+
+proc rebuild*(n: PNode; fn: Transformer): PNode =
+  assert not n.isNil
+  result = fn(n)
+  if cast[uint](result) == cast[uint](n):
+    result = shallowCopy n
+    if n.kind notin unsunny:
```
Clyybber:
```suggestion
    if n.kind in nkWithSons:
```

tools/dust/mutate.nim
```diff
@@ -0,0 +1,71 @@
...
+  result = fn(n)
+  if result.isNil:
+    result = shallowCopy n
+    for i, child in pairs(n):
+      result.sons[i] = transform(child, fn)
+
+proc rebuild*(n: PNode; fn: Transformer): PNode =
+  assert not n.isNil
+  result = fn(n)
+  if cast[uint](result) == cast[uint](n):
```
zerbina:
```suggestion
  if result == n:
```
The equality operator for `PNode` is not overloaded, so this should do the same.

tools/dust/mutate.nim
```diff
@@ -0,0 +1,71 @@
...
+type
+  Transformer = proc (n: PNode): PNode
+
+proc transform*(n: PNode; fn: Transformer): PNode =
+  assert not n.isNil
+  result = fn(n)
+  if result.isNil:
+    result = shallowCopy n
+    for i, child in pairs(n):
+      result.sons[i] = transform(child, fn)
```
zerbina:
```suggestion
      result[i] = transform(child, fn)
```

tools/dust/mutate.nim
```diff
@@ -0,0 +1,71 @@
...
+
+proc rebuild*(n: PNode; fn: Transformer): PNode =
+  assert not n.isNil
+  result = fn(n)
+  if cast[uint](result) == cast[uint](n):
+    result = shallowCopy n
+    if n.kind notin unsunny:
+      result.sons.setLen 0
```
zerbina:
```suggestion
      result.setLen 0
```

tools/dust/mutate.nim
```diff
@@ -0,0 +1,71 @@
+import
+  compiler / modules / modulegraphs,
+  compiler / ast / [ lineinfos, renderer, ast, astalgo, ]
+
+import spec
+import hashing
+
+proc node(k: TNodeKind): PNode = PNode(kind: k)
```
zerbina:
```suggestion
proc node(k: TNodeKind): PNode = newNode(k)
```
This makes sure that the node has a valid `info` and `id` field.

tools/dust/spec.nim
```diff
@@ -0,0 +1,84 @@
...
+  if not n.isNil:
+    remains.add newAttempt(n, sig)
+
+proc next*(remains: Remains): PNode =
+  if len(remains) > 0:
+    result = remains.attempts[^1].node
+
+proc massageMessage*(s: string): string =
+  result = s.splitLines()[0]
+  for c in {';', ':'}:
```
zerbina:
```suggestion
  for c in [';', ':']:
```
It's a bit more efficient.

tools/dust/mutate.nim
```diff
@@ -0,0 +1,71 @@
+import
+  compiler / modules / modulegraphs,
```
zerbina:
```suggestion
```

tools/dust/pass.nim
```diff
@@ -0,0 +1,39 @@
...
+    # yield the original node
+    result = n
+  else:
+    if c.ignore:
+      # we already yielded the program, so just return nil
+      result = nil
+    else:
+      # yield the next available permutation (once)
+      result = pop(remains)
+      assert result.kind == nkStmtList
```
zerbina:
```suggestion
```
If the module only consists of a single line, this assertion would fail.

tools/dust/pass.nim
```diff
@@ -0,0 +1,39 @@
...
+      c.ignore = true
```
zerbina: I'm not entirely sure, but since the process procedure is called for every fragment, wouldn't this mean that everything beyond and including the first declarative statement would be thrown away?

tools/dust/pass.nim
zerbina, line 1: This module seems unused? That is, neither `remains` nor `dustPass` seem to be used anywhere.

tools/dust/boring.nim
```diff
@@ -0,0 +1,163 @@
...
+import
+  compiler / ast / [
+    idents,
+    lineinfos,
+  ],
+  compiler / front / [
+      cmdlinehelper,
+      commands,
+      condsyms,
+      main,
```
zerbina:
```suggestion
```

tools/dust/boring.nim
```diff
@@ -0,0 +1,163 @@
...
+      msgs,
+      nimconf,
```
zerbina:
```suggestion
```

zerbina (commented):
I did a first pass and left some minor style suggestions.

tools/dust/spec.nim
saem: Good call, I added a todo item to audit dust code to use the latest nimskull bits.

I just went for getting it to compile rather than combing through it.

tools/dust/pass.nim
saem: you're correct, it seems to be implementing the pass interface, it might prove to be useful if we decide to head down the `nim minimize foo.nim` route, but if we don't do that this PR, I'd opt for dropping it.

tools/dust/pass.nim
saem: yeah, seeing as this code seems to be unused, it looks to be an incomplete implementation.

tools/dust/boring.nim
```diff
@@ -0,0 +1,186 @@
...
+
+proc parentDir(file: AbsoluteFile): AbsoluteDir =
+  result = AbsoluteDir(file) / RelativeDir".."
+
+proc loadConfig*(graph: ModuleGraph; fn: AbsoluteFile) =
+  ## use the ident cache to load the project config for the given filename
+  var result = graph.config
+
+  #excludeAllNotes(result, hintConf)
+  #excludeAllNotes(result, hintLineTooLong)
```
saem: Re-add if possible

saem:
Dust now works and successfully reduces the broken code in `mminimizeme.nim`, it's a bit slow, but that's probably fixable as sem gets faster (removal of unnecessary wrapping etc).

Another thing that will likely help a fair bit is a breadth first search and early elimination of the search space, but I'm not sure how to implement that (besides the breadth first) just yet.

saem:
Currently dust just up and quits, running the compiler on the test file emits an error as you'd expect (`no return type declared`).

But when running `dust` (`./bin/dust tools/dust/tests/mminimizeme.nim`), the following error is report is generated:

```nim
(category: Internal, internalReport: (msg: "wanted: tyInt got: tyProxy", kind: rintUnreachable, location: none(TLineInfo), reportInst: (file: "magicsys.nim", line: 103, col: 14), reportFrom: (file: "magicsys.nim", line: 103, col: 14)))
other error: (category: Internal, internalReport: (msg: "", kind: StackTrace, trace: dust.nim                 dust
dust.nim                 dust
boring.nim               compile
sem.nim                  semStmtAndGenerateGenerics
semexprs.nim             semExpr
semstmts.nim             semStmtList
semexprs.nim             semExpr
semstmts.nim             semConstLetOrVar
semstmts.nim             semNormalizedLetOrVar
semexprs.nim             semExprWithType
semexprs.nim             semExprCheck
semexprs.nim             semExpr
semdata.nim              setIntLitType
semdata.nim              getIntLitType
magicsys.nim             getSysType
msgs.nim                 doInternalUnreachable
msgs.nim                 quit
, location: none(TLineInfo), reportInst: (file: "msgs.nim", line: 282, col: 20), reportFrom: (file: "", line: 0, col: 0)))
```

This indicates that an internal assert being hit because we're not appropriately guarding for it in the compiler. The only problem with that hypothesis is that I can't replicate the issue with `--errormax:100`.

zerbina:
> This indicates that an internal assert being hit because we're not appropriately guarding for it in the compiler. The only problem with that hypothesis is that I can't replicate the issue with --errormax:100.

The error you've shown can only mean one thing, namely that a symbol named `int` cannot be found in the system module's interface. This causes `getSysSym` (called by `getSysType`) to return an `skError` symbol with type `tyError` (an alias for `tyProxy`), hence the assertion failure.

With a properly-processed `system` module, this error is effectively impossible, which hints at the module graph being faulty, or something with system processing going wrong.

Having said that, I've tried to reproduce your error by running `./bin/dust tools/dust/tests/mminimizeme.nim` locally, but:
* the error reported for the module (with just `nim`) is `undeclared identifier: "result"`, not `no return type declared`
* `dust` correctly minimizes the module without crashing

Changing `result = 10` to `return 10` (so that the error is `no return type declared`) also doesn't help; `dust` is still able to minimize the module properly.

zerbina:
I've looked into some improvements to speed up `dust`, which I think could be of interest to you.

## Mutation Producer

The mutation producer (`mutations`) as is creates a lot of ASTs that are grammatically invalid (e.g., an `nkDotExpr` with only 1 child), which is a problem because:
* it adds unnecessary work
* the AST renderer doesn't consider grammatically invalid trees in a lot of cases, leading to crashes

I've implemented a grammar-aware mutation producer (it doesn't handle the full grammar yet), which only creates legal ASTs. In addition, it also performs some reductions that the node-based one didn't, such as replacing `if a: b` with `b`.

<details>
  <summary>Patch</summary>

  ```diff
diff --git a/tools/dust/mutate.nim b/tools/dust/mutate.nim
index 07de3b2fd..01f19ac39 100644
--- a/tools/dust/mutate.nim
+++ b/tools/dust/mutate.nim
@@ -4,67 +4,164 @@ import
 import spec
 import hashing
 
-proc node(k: TNodeKind): PNode = newNode(k)
+const
+  FixedSize = {
+      nkDiscardStmt, nkCast, nkConv, nkAsgn, nkAddr, nkReturnStmt,
+      nkYieldStmt, nkDotExpr, nkPar, nkBreakStmt, nkContinueStmt, nkTypeDef,
+      nkObjectTy, nkRefTy}
+    ## syntax with a fixed number of children
+  WithBody = callableDefs + {nkElifExpr, nkElifBranch, nkElse, nkElseExpr,
+      nkWhileStmt, nkForStmt, nkBlockExpr, nkBlockStmt, nkPragmaBlock, nkOfBranch}
+    ## syntax with a fixed number of children, where the last child represents
+    ## a body (statement or expression)
+  WithBodyReplaceable = callableDefs + {nkWhileStmt, nkForStmt, nkBlockExpr,
+      nkBlockStmt, nkPragmaBlock}
+    ## syntax that may be replace with its body (i.e., the last child)
+  ListLike = {nkTupleConstr, nkVarSection, nkLetSection, nkConstSection,
+      nkTypeSection, nkStmtList, nkStmtListExpr}
+    ## syntax that must always have at least one element
+  IfLike = {nkIfExpr, nkIfStmt, nkWhenStmt}
+    ## syntax that has branches (e.g., `nkElse`, `nkElifBranch`, etc.) as children
 
-type
-  Transformer = proc (n: PNode): PNode
+proc numSteps*(n: PNode): int =
+  ## Computes the number of distinct single modifications (i.e., removing a
+  ## child node, replacing a child node, etc.) that can be performed
+  ## for AST `n` (including its sub-trees).
+  proc count(n: PNode): int =
+    for it in n.items:
+      result += numSteps(it)
 
-proc transform*(n: PNode; fn: Transformer): PNode =
-  assert not n.isNil
-  result = fn(n)
-  if result.isNil:
-    result = shallowCopy n
-    for i, child in pairs(n):
-      result[i] = transform(child, fn)
+  case n.kind
+  of nkWithoutSons:
+    # no steps are possible
+    result = 0
+  of nkIdentDefs:
+    # the identdefs must always have at least one element
+    result = count(n) + (if n.len > 3: n.len - 2 else: 0) + ord(n[^1].kind != nkEmpty)
+  of nkConstDef:
+    # similar to an identdefs, but the value expression cannot be removed
+    result = count(n) + (if n.len > 3: n.len - 2 else: 0)
+  of FixedSize:
+    result = count(n)
+  of WithBody:
+    result = count(n) + ord(n[^1].kind != nkEmpty) + ord(n.kind in WithBodyReplaceable)
+  of nkCaseStmt, nkRecCase:
+    # the selector must be kept, but replacing with branch bodies is possible
+    result = count(n) + (n.len - 1) + (n.len - 1)
+  of nkCallKinds - {nkPostfix}, nkCurlyExpr, nkBracketExpr:
+    # the first node must always stay
+    result = count(n) + n.len - 1
+  of ListLike:
+    # no removals are possible when the list has a single element
+    result = count(n) + (if n.len > 1: n.len else: 0)
+  of IfLike:
+    # the statement/expression can also be reduced by replacing it with
+    # one of the bodies
+    result = count(n) + (if n.len > 1: n.len else: 0) + n.len
+  of nkPostfix:
+    # the postfix can be replaced with its body
+    result = count(n) + 1
+  of nkPragmaExpr:
+    result = count(n) + 1
+  else:
+    # the rest can be fully reduced
+    result = count(n) + n.len
 
-proc rebuild*(n: PNode; fn: Transformer): PNode =
-  assert not n.isNil
-  result = fn(n)
-  if result == n:
-    result = shallowCopy n
-    if n.kind in nkWithSons:
-      result.sons.setLen 0
-      for child in items(n):
-        let rebuilt = rebuild(child, fn)
-        if not rebuilt.isNil:
-          result.add rebuilt
+proc reduced*(n: PNode, index: var int): PNode =
+  ## Computes the tree corresponding to a step.
+  if index < 0:
+    return n
 
-proc removeIndex(n: PNode; index: int): PNode =
-  var index = index
-  let z = size(n)
-  proc remover(n: PNode): PNode =
-    if index == 0:
-      result = nil
-    else:
-      result = n
+  template doStep(): bool =
+    let res = index == 0
     dec index
-  result = rebuild(n, remover)
+    res
+
+  proc reduceAll(n: PNode, index: var int): PNode =
+    # TODO: only copy on write
+    result = shallowCopy(n)
+    for i, it in n.pairs:
+      result[i] = reduce(it, index)
+
+  proc reduceWithRemove(n: PNode, start: int, index: var int): PNode =
+    result = newNode(n.kind)
+    for i in 0..<start:
+      result.add reduce(n[i], index)
+    for i in start..<n.len:
+      if not doStep():
+        result.add reduce(n[i], index)
+
+  case n.kind
+  of nkWithoutSons:
+    result = n
+  of nkIdentDefs, nkConstDef:
+    result = reduceAll(n, index)
+    if n.len > 3 and index >= 0:
+      if index < n.len - 2:
+        # remove the given identifier
+        result.delSon(index)
+      index -= (n.len - 2)
 
-proc discardIndex(n: PNode; index: int): PNode =
-  var index = index
-  proc discarder(n: PNode): PNode =
-    if index == 0:
-      result = nkDiscardStmt.node
+    if n.kind == nkIdentDefs and n[^1].kind != nkEmpty and doStep():
+      # remove the value expression
+      result[^1] = newNode(nkEmpty)
+  of nkCaseStmt, nkRecCase:
+    if index < n.len - 1:
+      result = n[1 + index][^1]
+      index -= (n.len - 1)
     else:
-      dec index
-  result = transform(n, discarder)
+      index -= (n.len - 1)
+      result = reduceWithRemove(n, 1, index) # the selector is always present
+  of nkCallKinds - {nkPostfix}, nkBracketExpr, nkCurlyExpr:
+    result = reduceWithRemove(n, 1, index) # the selector is always present
+  of WithBody:
+    result = reduceAll(n, index)
+    if n[^1].kind != nkEmpty:
+      if doStep():
+        # remove the body
+        result[^1] = nkDiscardStmt.newTree(nkEmpty.newNode())
 
-proc emptyIndex(n: PNode; index: int): PNode =
-  var index = index
-  proc emptier(n: PNode): PNode =
-    if index == 0:
-      result = nkEmpty.node
+    if n.kind in WithBodyReplaceable and doStep():
+      # replace with the body
+      result = result[^1]
+  of FixedSize:
+    result = reduceAll(n, index)
+  of ListLike:
+    if n.len > 1:
+      result = reduceWithRemove(n, 0, index)
+    else:
+      result = reduceAll(n, index)
+  of IfLike:
+    if index < n.len:
+      # replace with the body of the given branch
+      result = n[index][^1]
+      index -= n.len
+    else:
+      index -= n.len
+      if n.len > 1:
+        result = reduceWithRemove(n, 0, index)
+      else:
+        result = reduceAll(n, index)
+  of nkPostfix:
+    if doStep():
+      result = n[1]
+    else:
+      result = reduceAll(n, index)
+  of nkPragmaExpr:
+    if doStep():
+      result = n[0] # replace with the inner expression
     else:
-      dec index
-  result = transform(n, emptier)
+      result = reduceAll(n, index)
+  else:
+    result = reduceWithRemove(n, 0, index)
 
 iterator mutations*(n: PNode): PNode =
-  for i in 1 ..< size(n):
-    yield removeIndex(n, i)
-    yield emptyIndex(n, i)
-    yield discardIndex(n, i)
+  for i in 0 ..< numSteps(n):
+    var index = i
+    yield reduce(n, index)
+    assert index < 0 # sanity check
 
 proc init*(remains: var Remains; n: PNode) =
   let h = hashNode(n)
   if h notin remains:
-    remains.add(n, h)
\ No newline at end of file
+    remains.add(n, h)
  ```
  
</details>

## Driver

The current driver tries all candidates/attempts in the order they're pushed to the list, stopping only once the whole search space has been exhausted.

When for all `a` and `b` where `property(a) == true and property(b) == true and a < b` it holds that reducing `b` will yield `a` at some point, picking the smallest mutation for which the property holds and using it as the new best significantly reduces the amount of ASTs that need to be tried.

I've made a patch implementing this:

<details>
  <summary>Patch</summary>

  ```diff
diff --git a/tools/dust/dust.nim b/tools/dust/dust.nim
index d7b1bad48..34a3c6224 100644
--- a/tools/dust/dust.nim
+++ b/tools/dust/dust.nim
@@ -25,8 +25,10 @@ from compiler / front / cli_reporter import reportFull, legacyReportBridge
 from compiler / front / msgs import defaultDiagHandler
 
 import std/options as std_options # due to legacy reports stupidity
+import std/[sets, algorithm]
 
 import spec
+import hashing
 import boring
 import mutate
 
@@ -102,9 +104,6 @@ proc dust*(filename: AbsoluteFile): ErrorCode =
     best = toPNode(parseString(readFile(filename.string),
                        cache = cache, config = config, line = 0,
                        filename = filename.string))
-    score = size(best)
-    remains.add best
-    assert len(remains) > 0
 
   # if the semcheck passes, we have nothing to do
   if config.errorCounter == 0:
@@ -116,39 +115,64 @@ proc dust*(filename: AbsoluteFile): ErrorCode =
 
   # make note of the expected number of errors
   let expected = config.errorCounter
+  var seen: HashSet[SigHash]
+
+  while true:
+    var remains: seq[(SigHash, PNode)]
+    # gather all possible, not-yet-tried mutations:
+    for mutant in mutations(best):
+      let hash = hashNode(mutant)
+      if hash notin seen:
+        remains.add (hash, mutant)
+  
+    if remains.len == 0:
+      # there are none; we're done
+      break
+
+    # sort by their score. The one with the lowest score has to come first
+    sort(remains, proc(a, b: auto): int =
+      calculateScore(config, a[1]) - calculateScore(config, b[1]))
 
-  while len(remains) > 0:
-    echo rendered
-    echo "remaining: ", len(remains), " best: ", score
-    let node = pop(remains)
-
-    semcheck:
-      try:
-        writeFile(filename.string, $node)
-      except IndexError:
-        echo "cheating to get around rendering bug"
-        continue
-
-    # extra errors are a problem
-    if config.errorCounter > expected:
-      echo "(unexpected errors)"
-    # if we didn't unhook the errors,
-    # it means we didn't find the error we were looking for
-    elif config.structuredReportHook != dustReportHook:
-      echo "(uninteresting errors)"
-    # i guess this node is a viable reproduction
-    else:
-      let z = calculateScore(config, node)
-      if z < score:
-        echo "(new high score)"
-        best = node
-        score = z
-        rendered = $best
-      for mutant in mutations(node):
-        remains.add mutant
-
+    echo best
+    echo "----- current score: ", calculateScore(config, best)
+
+    var found = PNode nil
+    # try all candidates, starting with the smallest one
+    for (hash, node) in remains.items:
+      seen.incl(hash)
+
+      semcheck:
+        try:
+          writeFile(filename.string, $node)
+        except IndexError:
+          echo "cheating to get around rendering bug"
+          continue
+
+      # extra errors are a problem
+      if config.errorCounter > expected:
+        echo "(unexpected errors)"
+      # if we didn't unhook the errors,
+      # it means we didn't find the error we were looking for
+      elif config.structuredReportHook != dustReportHook:
+        echo "(uninteresting errors)"
+      # i guess this node is a viable reproduction
+      else:
+        # found a viable tree
+        found = node
+        break
+    
+    if found.isNil:
+      # none of the candidates reproduces the property; we're done
+      break
+
+    # note: it's possible, and valid, for the new best to have the same score
+    # as the previous best. This is because some reductions (e.g., replacing
+    # an `nkIntLit` with a `nkEmpty`) don't change the number of nodes
+    best = found
+    
   if not best.isNil:
     debug best
+    score = calculateScore(config, best)
     echo "=== minimal after ", counter, "/", remains.count, " semchecks; scored ", score
     echo best
     writeFile(filename.string, $best)
  ```

</details>

---

With both patches applied, `dust` is able to reduce the current test very quickly (~2 seconds for me).

saem:
> > This indicates that an internal assert being hit because we're not appropriately guarding for it in the compiler. The only problem with that hypothesis is that I can't replicate the issue with --errormax:100.
> 
> The error you've shown can only mean one thing, namely that a symbol named `int` cannot be found in the system module's interface. This causes `getSysSym` (called by `getSysType`) to return an `skError` symbol with type `tyError` (an alias for `tyProxy`), hence the assertion failure.

Yup, that's exactly what's happening, when I output all errors (and more servere) I got an error for system not being found in the `.nimble` directory. I suspect there is something off locally, but I'm not sure what exactly. I have merged in the latest devel so it's confusing as to why that's happening. As soon as I provide `--lib:./lib` on the command line to `dust` it correctly minimizes the input module.

> With a properly-processed `system` module, this error is effectively impossible, which hints at the module graph being faulty, or something with system processing going wrong.
> 
> Having said that, I've tried to reproduce your error by running `./bin/dust tools/dust/tests/mminimizeme.nim` locally, but:
> 
> * the error reported for the module (with just `nim`) is `undeclared identifier: "result"`, not `no return type declared`
> * `dust` correctly minimizes the module without crashing

Yeah, I had a slightly different version of the test in order to see if that might reduce the time to minimize the input module, so I had the wrong message listed, that's fine. It's correctly finding the error.

> Changing `result = 10` to `return 10` (so that the error is `no return type declared`) also doesn't help; `dust` is still able to minimize the module properly.

Yup, it's something to do with my locally compiled dust (`time ./koch.py tools --lib:lib -d:debug --stacktrace -u:leanCompiler`), when run (`time ./bin/dust tools/dust/tests/mminimizeme.nim`) searching for the system library in the following path: `/home/saem/.nimble/lib/system.nim` (there is a message for it).

Fuller error messages from my weird `dust` environment:

```nim
Error: cannot open '/home/saem/.nimble/lib/system.nim'
/home/saem/Development/nim/nimskull/tools/dust/tests/mminimizeme.nim(1, 18) Error: system needs: 'int'
Fatal: Internal unreachable code executed - magicsys.nim(103, 15) (wanted: tyInt got: tyProxy) should never be called.
dust.nim(0)     dust
dust.nim(0)     dust
boring.nim(0)   compile
modules.nim(0)  compileProject
modules.nim(0)  compileModule
passes.nim(0)   processModule
passes.nim(0)   processTopLevelStmt
sem.nim(0)      myProcess
sem.nim(0)      semStmtAndGenerateGenerics
semstmts.nim(0) semStmt
semexprs.nim(0) semExprNoType
semexprs.nim(0) semExpr
semstmts.nim(0) semStmtList
semexprs.nim(0) semExpr
semstmts.nim(0) semConstLetOrVar
semstmts.nim(0) semNormalizedLetOrVar
semexprs.nim(0) semExprWithType
semexprs.nim(0) semExprCheck
semexprs.nim(0) semExpr
semdata.nim(0)  setIntLitType
semdata.nim(0)  getIntLitType
magicsys.nim(0) getSysType
msgs.nim(0)     doInternalUnreachable
msgs.nim(0)     quit
```

saem:
> I've looked into some improvements to speed up `dust`, which I think could be of interest to you.

I've run the new driver patch alone (see below) and it goes from 2m 23s to less than 20s, and that's with a debug build. 🚀 👍🏽 

> ## Mutation Producer
> The mutation producer (`mutations`) as is creates a lot of ASTs that are grammatically invalid (e.g., an `nkDotExpr` with only 1 child), which is a problem because:
> 
> * it adds unnecessary work
> * the AST renderer doesn't consider grammatically invalid trees in a lot of cases, leading to crashes
> 
> I've implemented a grammar-aware mutation producer (it doesn't handle the full grammar yet), which only creates legal ASTs. In addition, it also performs some reductions that the node-based one didn't, such as replacing `if a: b` with `b`.

I think you accidentally posted the dust search patch here, instead of the mutation producer (a patch against `mutate.nim`).

saem:
The CI failure is misleading, the compiler (wrapped by dust) is quitting the process with an exit code 1, which happens to correspond to the exit code of no file for dust itself, so it's being interpretted as such. It's almost certainly dust having issues with finding the standard libary.

zerbina:
I've gathered some suggestions, but they all span multiple lines and files, so I'm leaving them as patches. The second one depends on the first one; the last two are independent.

### Fix config file processing

Currently, the config files are processed twice (and the first time it happens is also before the first command-line processing pass took place), which could is both unnecessary and could cause issues.

The patch removes the `loadConfig` procedure and moves the extra option/note configuration into `setup`.

(This is an improved version of the patch I sent on Matrix.)

<details>
  <summary>0001-fix-config-loading.patch</summary>

  ```diff
diff --git a/tools/dust/boring.nim b/tools/dust/boring.nim
index ba51cdb2c..739a278c9 100644
--- a/tools/dust/boring.nim
+++ b/tools/dust/boring.nim
@@ -33,9 +33,6 @@ from compiler / front / commands import procSwitchResultToEvents,
                                         cliEventLogger,
                                         showMsg
 
-const
-  NimCfg* {.strdefine.} = "nim".addFileExt "cfg"
-
 template excludeAllNotes(config: ConfigRef; n: typed) {.used.} =
   config.notes.excl n
   when compiles(config.mainPackageNotes):
@@ -134,6 +131,8 @@ proc compile*(graph: ModuleGraph) =
   if config.libpath notin config.searchPaths:
     config.searchPaths.add config.libpath     # make sure we can import
 
+  initDefines(config.symbols)
+
   config.setErrorMaxHighMaybe                 # for now, we honor errorMax
   defineSymbol(config, "nimcheck")            # useful for static: reasons
 
@@ -152,35 +151,10 @@ proc setup*(cache: IdentCache; config: ConfigRef; graph: ModuleGraph): bool =
     processCmdLineAndProjectPath(prog, config, argv)
     result = loadConfigsAndProcessCmdLine(prog, cache, config, graph, argv)
 
-proc parentDir(file: AbsoluteFile): AbsoluteDir =
-  result = AbsoluteDir(file) / RelativeDir".."
-
-proc loadConfig*(graph: ModuleGraph; fn: AbsoluteFile) =
-  ## use the ident cache to load the project config for the given filename
-  var result = graph.config
-
-  #excludeAllNotes(result, hintConf)
-  #excludeAllNotes(result, hintLineTooLong)
+    #excludeAllNotes(result, hintConf)
+    #excludeAllNotes(result, hintLineTooLong)
 
-  initDefines(result.symbols)
-
-  let compilerPath = AbsoluteFile findExe"nim"
-  result.prefixDir = parentDir(compilerPath) / RelativeDir".."
-  result.projectPath = parentDir(fn)
-
-  when false:
-    let cfg = fn.string & ExtSep & "cfg"
-    if fileExists(cfg):
-      if not readConfigFile(cfg.AbsoluteFile, graph.cache, result):
-        raise newException(ValueError, "couldn't parse " & cfg)
-  else:
-    let cwd = getCurrentDir()
-    setCurrentDir $result.projectPath
-    try:
-      discard loadConfigs(NimCfg.RelativeFile, graph.cache, result)
-    finally:
-      setCurrentDir cwd
-
-  incl result, optStaticBoundsCheck
-  excl result, optWarns
-  excl result, optHints
\ No newline at end of file
+    # force enable/disable some options
+    incl config, optStaticBoundsCheck
+    excl config, optWarns
+    excl config, optHints
diff --git a/tools/dust/dust.nim b/tools/dust/dust.nim
index 8d181aa43..6eafcc319 100644
--- a/tools/dust/dust.nim
+++ b/tools/dust/dust.nim
@@ -45,17 +45,14 @@ template semcheck(body: untyped) {.dirty.} =
   config.diagHandler = msgs.defaultDiagHandler
   config.astDiagToLegacyReport = cli_reporter.legacyReportBridge
   graph = newModuleGraph(cache, config)
-  graph.loadConfig(filename)
 
-  # perform boring setup of the config using the cache
-  if not setup(cache, config, graph):
+  # perform boring setup of the config and graph (command line parsing, config
+  # file loading, etc.)
+  if not setup(cache, config, graph, args):
     return ErrorCode.setupError
 
   config.verbosity = compVerbosityMin   # reduce spam
 
-  # create a new module graph
-  #graph = newModuleGraph(cache, config)
-
   body
   registerPass graph, semPass           # perform semcheck
   compile graph                         # run the compile
  ```
</details>

### Improve command-line processing

The overarching goal of this suggestion is to make `dust` behave more as if it were a compiler sub-command. It also fixes some problem with missing configuration setup.

In detail:
* use `check` as the internal command (instead of `c`)
* fix target-specific configuration (i.e., GC, exception mode) not being set up properly
* fix using a backed other than `c` not being supported
* always treat the first command-line argument (that's not a switch or option) as the file, meaning that, e.g., `dust c test.nim` no longer works
* handle the "main file" requirement via `wantMainModule` and remove the extra project file discovery in `dust.nim`. The main procedure now takes the command-line argument list as input, not a filepath

<details>
  <summary>0002-improve-command-line-processing.patch</summary>

  ```diff
diff --git a/tools/dust/boring.nim b/tools/dust/boring.nim
index 739a278c9..fbbadab24 100644
--- a/tools/dust/boring.nim
+++ b/tools/dust/boring.nim
@@ -8,8 +8,6 @@ import std/times
 import std/os
 import std/parseopt
 
-from std/strutils import endsWith
-
 import
   compiler / ast / [
     idents,
@@ -19,7 +17,6 @@ import
       cmdlinehelper,
       # commands,
       condsyms,
-      msgs,
       options,
       optionsprocessor,
   ],
@@ -30,8 +27,9 @@ import
   compiler / utils / pathutils
 
 from compiler / front / commands import procSwitchResultToEvents,
-                                        cliEventLogger,
-                                        showMsg
+                                        cliEventLogger
+
+from compiler/front/main import customizeForBackend
 
 template excludeAllNotes(config: ConfigRef; n: typed) {.used.} =
   config.notes.excl n
@@ -43,20 +41,10 @@ template excludeAllNotes(config: ConfigRef; n: typed) {.used.} =
 proc processArgument(pass: TCmdLinePass; p: OptParser;
                      argsCount: var int; config: ConfigRef): bool =
   if argsCount == 0:
-    if p.key.endsWith(".nim"):
-      config.setCmd cmdCompileToC
-      config.projectName = unixToNativePath(p.key)
-      config.arguments = cmdLineRest(p)
-      result = true
-    elif pass != passCmd2: setCommandEarly(config, p.key)
-  else:
-    if pass == passCmd1: config.commandArgs.add p.key
-    if argsCount == 1:
-      # support UNIX style filenames everywhere for portable build scripts:
-      if config.projectName.len == 0 and config.inputMode == pimFile:
-        config.projectName = unixToNativePath(p.key)
-      config.arguments = cmdLineRest(p)
-      result = true
+    # the first argument is the file
+    config.projectName = unixToNativePath(p.key)
+    config.arguments = cmdLineRest(p) # consume the rest
+    result = true
   inc argsCount
 
 proc cmdLine(pass: TCmdLinePass, cmd: openArray[string]; config: ConfigRef) =
@@ -64,6 +52,10 @@ proc cmdLine(pass: TCmdLinePass, cmd: openArray[string]; config: ConfigRef) =
   var p = initOptParser(cmd)
   var argsCount = 0
 
+  # the 'check' command (which best approximates dust's operation) is implied
+  if pass != passCmd2:
+    setCommandEarly(config, "check")
+
   config.commandLine.setLen 0  # some bug
   while true:
     next(p)
@@ -95,23 +87,6 @@ proc cmdLine(pass: TCmdLinePass, cmd: openArray[string]; config: ConfigRef) =
       if processArgument(pass, p, argsCount, config):
         break
 
-  when false:
-    if pass == passCmd2:
-      if {optRun, optWasNimscript} * config.globalOptions == {} and
-          config.arguments.len > 0 and
-          config.command.normalize notin ["run", "e"]:
-        rawMessage(config, errGenerated, errArgsNeedRunOption)
-
-proc helpOnError(config: ConfigRef) =
-  const
-    Usage = """
-  dust [options] [projectfile]
-
-  Options: Same options that the Nimskull compiler supports.
-  """
-  showMsg(config, Usage)
-  msgQuit 0
-
 proc reset*(graph: ModuleGraph) =
   ## reset the module graph so it is ready for recompilation
   # we're not dirty if we don't have a fileindex
@@ -136,25 +111,34 @@ proc compile*(graph: ModuleGraph) =
   config.setErrorMaxHighMaybe                 # for now, we honor errorMax
   defineSymbol(config, "nimcheck")            # useful for static: reasons
 
+  customizeForBackend(graph, config, backendC)# use C as the default target
+
   graph.suggestMode = true                    # needed for dirty flags
   compileProject graph                        # process the graph
 
-proc setup*(cache: IdentCache; config: ConfigRef; graph: ModuleGraph): bool =
-  proc noop(graph: ModuleGraph) {.used.} = discard
-  let prog = NimProg(supportsStdinFile: true,
-                     processCmdLine: cmdLine) #, mainCommand: mainCommand)
+proc setup*(cache: IdentCache; config: ConfigRef; graph: ModuleGraph,
+            argv: openArray[string]): bool =
+  let prog = NimProg(
+    supportsStdinFile: false,
+    processCmdLine: cmdLine
+  )
   initDefinesProg(prog, config, "dust")
-  if paramCount() == 0:
-    helpOnError(config)
+  processCmdLineAndProjectPath(prog, config, argv)
+  result = loadConfigsAndProcessCmdLine(prog, cache, config, graph, argv)
+
+  #excludeAllNotes(result, hintConf)
+  #excludeAllNotes(result, hintLineTooLong)
+
+  # force enable/disable some options
+  incl config, optStaticBoundsCheck
+  excl config, optWarns
+  excl config, optHints
+
+proc wantMainModule*(config: ConfigRef): bool =
+  ## Sets the main module to the file whose path was provided on the command
+  ## line, returning false if this isn't possible (because there's no path).
+  if config.projectFull.isEmpty:
+    result = false
   else:
-    let argv = getExecArgs()
-    processCmdLineAndProjectPath(prog, config, argv)
-    result = loadConfigsAndProcessCmdLine(prog, cache, config, graph, argv)
-
-    #excludeAllNotes(result, hintConf)
-    #excludeAllNotes(result, hintLineTooLong)
-
-    # force enable/disable some options
-    incl config, optStaticBoundsCheck
-    excl config, optWarns
-    excl config, optHints
+    modules.wantMainModule(config)
+    result = true
diff --git a/tools/dust/dust.nim b/tools/dust/dust.nim
index 6eafcc319..0a9894dbc 100644
--- a/tools/dust/dust.nim
+++ b/tools/dust/dust.nim
@@ -51,6 +51,10 @@ template semcheck(body: untyped) {.dirty.} =
   if not setup(cache, config, graph, args):
     return ErrorCode.setupError
 
+  # make sure there's a main module provided
+  if not wantMainModule(config):
+    return ErrorCode.fileNotProvided
+
   config.verbosity = compVerbosityMin   # reduce spam
 
   body
@@ -70,7 +74,7 @@ proc dustReportHook(conf: ConfigRef, report: Report): TErrorHandling =
   #   echo conf.reportFull(report)
   doDefault
 
-proc dust*(filename: AbsoluteFile): ErrorCode =
+proc dust*(args: openArray[string]): ErrorCode =
   var
     graph: ModuleGraph
     cache: IdentCache
@@ -99,9 +103,9 @@ proc dust*(filename: AbsoluteFile): ErrorCode =
   # in the first pass, we add the program to our cache
   semcheck:
     # basically, just taking advantage of cache and config values...
-    best = toPNode(parseString(readFile(filename.string),
+    best = toPNode(parseString(readFile(config.projectFull.string),
                        cache = cache, config = config, line = 0,
-                       filename = filename.string))
+                       filename = config.projectFull.string))
 
   # if the semcheck passes, we have nothing to do
   if config.errorCounter == 0:
@@ -140,7 +144,7 @@ proc dust*(filename: AbsoluteFile): ErrorCode =
 
       semcheck:
         try:
-          writeFile(filename.string, $node)
+          writeFile(config.projectFull.string, $node)
         except IndexError:
           echo "cheating to get around rendering bug"
           continue
@@ -172,27 +176,18 @@ proc dust*(filename: AbsoluteFile): ErrorCode =
     score = calculateScore(config, best)
     echo "=== minimal after ", counter, "/", remains.count, " semchecks; scored ", score
     echo best
-    writeFile(filename.string, $best)
+    writeFile(config.projectFull.string, $best)
 
 when isMainModule:
-  from std/strutils import strip
-
-  if paramCount() > 0 and paramStr(paramCount()).strip() != "":
-    let
-      file = paramStr(paramCount())
-      absFile = toAbsolute(file, AbsoluteDir(getCurrentDir()))
-      code = dust(absFile)
-    case code
-    of ErrorCode.setupError:
-      echo "crashing due to error during setup"
-    of ErrorCode.noError:
-      echo "error: " & file & " passes the semcheck"
-    of ErrorCode.fileNotProvided:
-      echo "args: ", commandLineParams()
-      echo "supply a source file to inspect"
-    of ErrorCode.success:
-      echo "success: " & file
-    quit ord(code)
-  else:
+  let args = getExecArgs()
+  let code = dust(args)
+  case code
+  of ErrorCode.setupError:
+    echo "crashing due to error during setup"
+  of ErrorCode.noError:
+    echo "error: module passes the semcheck"
+  of ErrorCode.fileNotProvided:
     echo "supply a source file to inspect"
-    quit ord(ErrorCode.fileNotProvided)
+  of ErrorCode.success:
+    echo "success"
+  quit ord(code)
  ```

</details>

### Remove unused APIs from `spec`

With the new driver logic, the `Remains` and `Attempt` APIs are not used anymore, meaning that they can be removed.

<details>
  <summary>0003-remove-unused-apis.patch</summary>

  ```diff
diff --git a/tools/dust/dust.nim b/tools/dust/dust.nim
index 0a9894dbc..a3f14dbe8 100644
--- a/tools/dust/dust.nim
+++ b/tools/dust/dust.nim
@@ -83,7 +83,6 @@ proc dust*(args: openArray[string]): ErrorCode =
     best: PNode
     counter = 0
     score: int
-    remains: Remains
 
   result = ErrorCode.success
 
@@ -174,7 +173,7 @@ proc dust*(args: openArray[string]): ErrorCode =
   if not best.isNil:
     debug best
     score = calculateScore(config, best)
-    echo "=== minimal after ", counter, "/", remains.count, " semchecks; scored ", score
+    echo "=== minimal after ", counter, " semchecks; scored ", score
     echo best
     writeFile(config.projectFull.string, $best)
 
diff --git a/tools/dust/mutate.nim b/tools/dust/mutate.nim
index c8f417476..8ad6bc1d6 100644
--- a/tools/dust/mutate.nim
+++ b/tools/dust/mutate.nim
@@ -1,10 +1,6 @@
 import
   compiler / ast / [ lineinfos, renderer, ast, ]
 
-import spec
-import hashing
-
-
 const
   DynamicSize = {nkBracket, nkCurly, nkRecList, nkTupleTy, nkPragma,
       nkGenericParams}
@@ -172,8 +168,3 @@ iterator mutations*(n: PNode): PNode =
     var index = i
     yield reduce(n, index)
     assert index < 0 # sanity check
-
-proc init*(remains: var Remains; n: PNode) =
-  let h = hashNode(n)
-  if h notin remains:
-    remains.add(n, h)
diff --git a/tools/dust/spec.nim b/tools/dust/spec.nim
index b33903ec9..4659c80c0 100644
--- a/tools/dust/spec.nim
+++ b/tools/dust/spec.nim
@@ -2,32 +2,13 @@ import std/sets
 
 import
   compiler / modules / modulegraphs,
-  compiler / ast / [ lineinfos, renderer, ast ]
-
-import hashing
+  compiler / ast / [ lineinfos, renderer, ast_types, ast_query ]
 
 type
-  Attempt = object
-    node: PNode
-    sig: SigHash
-    size: int
-
-  Remains* = object
-    signatures: HashSet[SigHash]
-    attempts: seq[Attempt]
-
   DustContext* = ref object of PPassContext
     mainIndex*: FileIndex
     ignore*: bool
 
-proc len*(r: Remains): int = len(r.attempts)
-
-proc count*(r: Remains): int = len(r.signatures)
-
-proc pop*(remains: var Remains): PNode =
-  assert len(remains) > 0, "pop from empty remains"
-  result = pop(remains.attempts).node
-
 proc size*(n: PNode): int =
   assert not n.isNil
   result = 1
@@ -35,36 +16,4 @@ proc size*(n: PNode): int =
     for child in items(n.sons):
       inc result, size(child)
 
-proc contains*(remains: Remains; h: SigHash): bool =
-  h in remains.signatures
-
-proc contains*(remains: Remains; n: PNode): bool =
-  hashNode(n) in remains
-
-proc newAttempt(n: PNode): Attempt =
-  var n = copyTree(n)
-  result = Attempt(node: n, sig: hashNode(n), size: size(n))
-
-proc newAttempt(n: PNode; sig: SigHash): Attempt =
-  var n = copyTree(n)
-  assert hashNode(n) == sig
-  result = Attempt(node: n, sig: sig, size: size(n))
-
-proc add(remains: var Remains; a: Attempt) =
-  if a.size > 1:
-    if not containsOrIncl(remains.signatures, a.sig):
-      remains.attempts.add a
-
-proc add*(remains: var Remains; n: PNode) =
-  if not n.isNil:
-    remains.add newAttempt(n)
-
-proc add*(remains: var Remains; n: PNode; sig: SigHash) =
-  if not n.isNil:
-    remains.add newAttempt(n, sig)
-
-proc next*(remains: Remains): PNode =
-  if len(remains) > 0:
-    result = remains.attempts[^1].node
-
-export `$`
\ No newline at end of file
+export `$`
  ```

</details>

### Fix some mutator issues

The mutation producer is missing basic handling for some syntax.

In addition, to be more robust in the face of future syntax additions, the default behaviour for syntax should be "keep all children" instead of "allow removing children".

<details>
  <summary>0004-make-mutator-more-robust.patch</summary>

  ```diff
diff --git a/tools/dust/mutate.nim b/tools/dust/mutate.nim
index 2a41af627..c8f417476 100644
--- a/tools/dust/mutate.nim
+++ b/tools/dust/mutate.nim
@@ -6,25 +6,28 @@ import hashing
 
 
 const
-  FixedSize = {
-      nkDiscardStmt, nkCast, nkConv, nkAsgn, nkAddr, nkReturnStmt,
-      nkYieldStmt, nkDotExpr, nkPar, nkBreakStmt, nkContinueStmt, nkTypeDef,
-      nkObjectTy, nkRefTy}
-    ## syntax with a fixed number of children
-  WithBody = callableDefs + {nkElifExpr, nkElifBranch, nkElse, nkElseExpr,
-      nkWhileStmt, nkForStmt, nkBlockExpr, nkBlockStmt, nkPragmaBlock, nkOfBranch}
+  DynamicSize = {nkBracket, nkCurly, nkRecList, nkTupleTy, nkPragma,
+      nkGenericParams}
+    ## syntax where an arbitrary number of children is allowed (including none)
+  WithBody = callableDefs +
+      {nkElifExpr, nkElifBranch, nkElse, nkElseExpr, nkWhileStmt, nkForStmt,
+       nkBlockExpr, nkBlockStmt, nkPragmaBlock, nkOfBranch}
     ## syntax with a fixed number of children, where the last child represents
     ## a body (statement or expression)
-  WithBodyReplaceable = callableDefs + {nkWhileStmt, nkForStmt, nkBlockExpr,
-      nkBlockStmt, nkPragmaBlock}
+  WithBodyReplaceable = callableDefs +
+      {nkWhileStmt, nkForStmt, nkBlockExpr, nkBlockStmt, nkPragmaBlock}
     ## syntax that may be replace with its body (i.e., the last child)
-  ListLike = {nkTupleConstr, nkVarSection, nkLetSection, nkConstSection,
-      nkTypeSection, nkStmtList, nkStmtListExpr}
+  FixedFirstChild = {nkCall, nkCurlyExpr, nkBracketExpr, nkObjConstr} -
+      {nkPostFix}
+    ## syntax where the child has to always be present
+  ListLike = {nkTupleConstr, nkTableConstr, nkVarSection, nkLetSection,
+      nkConstSection, nkTypeSection, nkStmtList, nkStmtListExpr,
+      nkFormalParams, nkBindStmt, nkMixinStmt, nkEnumTy, nkUsingStmt,
+      nkImportStmt}
     ## syntax that must always have at least one element
-  IfLike = {nkIfExpr, nkIfStmt, nkWhenStmt}
+  IfLike = {nkIfExpr, nkIfStmt, nkWhenStmt, nkRecWhen}
     ## syntax that has branches (e.g., `nkElse`, `nkElifBranch`, etc.) as children
 
-
 proc numSteps*(n: PNode): int =
   ## Computes the number of distinct single modifications (i.e., removing a
   ## child node, replacing a child node, etc.) that can be performed
@@ -39,18 +42,20 @@ proc numSteps*(n: PNode): int =
     result = 0
   of nkIdentDefs:
     # the identdefs must always have at least one element
-    result = count(n) + (if n.len > 3: n.len - 2 else: 0) + ord(n[^1].kind != nkEmpty)
+    result = count(n) + (if n.len > 3: n.len - 2 else: 0) +
+      ord(n[^1].kind != nkEmpty)
   of nkConstDef:
     # similar to an identdefs, but the value expression cannot be removed
     result = count(n) + (if n.len > 3: n.len - 2 else: 0)
-  of FixedSize:
-    result = count(n)
+  of DynamicSize:
+    result = count(n) + n.len
   of WithBody:
-    result = count(n) + ord(n[^1].kind != nkEmpty) + ord(n.kind in WithBodyReplaceable)
+    result = count(n) + ord(n[^1].kind != nkEmpty) +
+      ord(n.kind in WithBodyReplaceable)
   of nkCaseStmt, nkRecCase:
     # the selector must be kept, but replacing with branch bodies is possible
     result = count(n) + (n.len - 1) + (n.len - 1)
-  of nkCallKinds - {nkPostfix}, nkCurlyExpr, nkBracketExpr:
+  of FixedFirstChild:
     # the first node must always stay
     result = count(n) + n.len - 1
   of ListLike:
@@ -60,15 +65,16 @@ proc numSteps*(n: PNode): int =
     # the statement/expression can also be reduced by replacing it with
     # one of the bodies
     result = count(n) + (if n.len > 1: n.len else: 0) + n.len
+  of nkCommand:
+    result = count(n) + (if n.len > 2: (n.len - 1) else: 0)
   of nkPostfix:
     # the postfix can be replaced with its body
     result = count(n) + 1
   of nkPragmaExpr:
     result = count(n) + 1
   else:
-    # the rest can be fully reduced
-    result = count(n) + n.len
-
+    # everything else has to keep all their children
+    result = count(n)
 
 proc reduce*(n: PNode, index: var int): PNode =
   ## Computes the tree corresponding to a step.
@@ -114,8 +120,8 @@ proc reduce*(n: PNode, index: var int): PNode =
     else:
       index -= (n.len - 1)
       result = reduceWithRemove(n, 1, index) # the selector is always present
-  of nkCallKinds - {nkPostfix}, nkBracketExpr, nkCurlyExpr:
-    result = reduceWithRemove(n, 1, index) # the selector is always present
+  of FixedFirstChild:
+    result = reduceWithRemove(n, 1, index)
   of WithBody:
     result = reduceAll(n, index)
     if n[^1].kind != nkEmpty:
@@ -125,8 +131,8 @@ proc reduce*(n: PNode, index: var int): PNode =
     if n.kind in WithBodyReplaceable and doStep():
       # replace with the body
       result = result[^1]
-  of FixedSize:
-    result = reduceAll(n, index)
+  of DynamicSize:
+    result = reduceWithRemove(n, 0, index)
   of ListLike:
     if n.len > 1:
       result = reduceWithRemove(n, 0, index)
@@ -143,6 +149,11 @@ proc reduce*(n: PNode, index: var int): PNode =
         result = reduceWithRemove(n, 0, index)
       else:
         result = reduceAll(n, index)
+  of nkCommand:
+    if n.len > 2:
+      result = reduceWithRemove(n, 1, index)
+    else:
+      result = reduceAll(n, index)
   of nkPostfix:
     if doStep():
       result = n[1]
@@ -154,7 +165,7 @@ proc reduce*(n: PNode, index: var int): PNode =
     else:
       result = reduceAll(n, index)
   else:
-    result = reduceWithRemove(n, 0, index)
+    result = reduceAll(n, index)
 
 iterator mutations*(n: PNode): PNode =
   for i in 0 ..< numSteps(n):
  ```
</details>

saem:
I've applied the above patches (thank you very much, @zerbina), which brought about all the indicated improvements, and also no longer produces the error where the standard library isn't triggered and `dust` can run without specifying `--lib:./lib`.

Going to do another pass to see what else can be gotten rid of.

compiler/sem/semcall.nim
```diff
@@ -551,7 +551,7 @@ proc semResolvedCall(c: PContext, x: TCandidate,
     result[0] = newSymNode(finalCallee, getCallLineInfo(result[0]))
     if x.fauxMatch == tyError:
       # at least one argument expression was erroneous
-      result = c.config.wrapError(result)
+      discard "already reported"
```
zerbina: Unfortunately, removing the error wrapping is currently not correct and results in macro and template invocations with erroneous arguments being expanded, which allows `nkError`, `tyError`, and `skError` to reach into macros, as well as causing some templates expanding indefinitely.

For example:
```nim
type Test = object
  field: int

template field(x: Test): int =
  x.field

field(missing)
```

will keep recursively expanding the template until either `sem` crashes with a stack overflow or the maximum number of nested template expansions is reached.

However, as you noted, unconditionally using `wrapError` is also wrong. I can look into making a separate fix for this problem.

tools/dust/boring.nim
```diff
@@ -0,0 +1,125 @@
...
+  let prog = NimProg(
+    supportsStdinFile: false,
+    processCmdLine: cmdLine
+  )
+  initDefinesProg(prog, config, "dust")
+  processCmdLineAndProjectPath(prog, config, argv)
+  result = loadConfigsAndProcessCmdLine(prog, cache, config, graph, argv)
+
+  # force enable/disable some options
+  incl config, optStaticBoundsCheck
```
zerbina:
```suggestion
```
Static bound checks don't really work properly (at the moment), and having them enabled leads to most code using array/seq access not compiling.

tools/dust/hashing.nim
```diff
@@ -0,0 +1,10 @@
+include compiler/sem/sighashes # because too few things are exported
+
+const
+  considerAll = {ConsiderFlag.low .. ConsiderFlag.high}
+
+proc hashNode*(n: PNode; flags: set[ConsiderFlag] = considerAll): SigHash =
```
zerbina: Could you move this procedure into `sighashes`, rename it to `hashTree`, replace all callsites (there should only be one, in `dust.nim`), and then remove the `hashing` module? This would get rid of the include.

Alternatively, you could also use `import {.all.}`, but I think having this procedure in `sighashes` is preferable.

tools/dust/tester.nim
```diff
@@ -0,0 +1,73 @@
+import experimental/cmdline
+from compiler/utils/pathutils import AbsoluteFile
```
zerbina:
```suggestion
```
The import is unused.

tools/dust/tester.nim
```diff
@@ -0,0 +1,73 @@
...
+from std/os import walkDir, PathComponent
+from std/osproc import execCmdEx
+from std/strutils import endsWith
+import dust
```
zerbina: Importing `dust` (and pulling in large parts of the compiler) just for the `ErrorCode` is too much, I think.

Could you move the `ErrorCode` enum into `spec` (not ideal, but it's a small, common module, with little dependencies, and the enum at least somewhat fits in there) and than replace the import?

tools/dust/tester.nim
```diff
@@ -0,0 +1,73 @@
...
+proc main(testsDir: string = "./tools/dust/tests") =
+  var
+    success = 0
+    error = 0
+    failed = 0
+
+  for pc, file in walkDir(testsDir):
+    if pc != PathComponent.pcFile:
+      continue
+    if file.endsWith(".nim"):
```
zerbina:
```suggestion
    if pc == PathComponent.pcFile and file.endsWith(".nim"):
```

tools/dust/tester.nim
```diff
@@ -0,0 +1,73 @@
...
+      case code
+      of ErrorCode.success:
+        echo "success: " & file
+        inc success
+      of ErrorCode.fileNotProvided:
+        echo "error: " & file & " with code " & $code
+        inc error
+      of ErrorCode.setupError:
+        echo "error: " & file & " with code " & $code
+        inc error
```
zerbina:
```suggestion
      of ErrorCode.fileNotProvided, ErrorCode.setupError:
        echo "error: " & file & " with code " & $code
        inc error
```

zerbina (commented):
The changes looks good quite good. I left a few comments about outstanding problems and about some things that I think should be changed.

Separately, to be in line with the style in the rest of the codebase, could you:
- [x] merge all imports into a single section at the top of the file
- [x] remove the spaces after and before `/` in import paths 
- [x] add a small doc comment at the top of each new module, detailing their purpose
- [x] change the procedure-level doc comments to use third-person singular and to start with a capital letter and end with a dot

Finally, the `dust` executable isn't shipped in the realease archive right now. You need to register the executable in `compiler/compiler.ini`, in both the `[WinBin]` and `[UnixBin]` sections.

compiler/sem/semcall.nim
saem: Thanks, I'll revert the change here (I've added a note to the todo list).

```suggestion
      result = c.config.wrapError(result)
```

Undo `wrapError` removal, as it causes various issues if done in isolation.

saem:
Thanks for the additional review, @zerbina -- I added the items you mentioned to the todo list of the PR to go through them.

I'll poke through some more changes tomorrow, I should be done with the latest round ideally before the weekend.

P.S. For some reason the batches didn't work or I made a mistake, had to bring in the stragglers separately.

zerbina:
> P.S. For some reason the batches didn't work or I made a mistake, had to bring in the stragglers separately.

Hm, I created the suggestions the same way I usually do, but when looking at the final preview prior to submitting the review, the line associations did look off. I chalked this up to a rendering bug, but - given that applying the suggestions didn't work properly - seems like something else went wrong ...

tools/dust/hashing.nim
saem: Agreed, it's better in `sighashes` -- done.

tools/dust/tester.nim
saem: All done.

compiler/sem/sighashes.nim
```diff
@@ -103,6 +106,15 @@ proc hashTree(c: var MD5Context, n: PNode; flags: set[ConsiderFlag]) =
   of nkWithSons:
     for i in 0..<n.len: hashTree(c, n[i], flags)
 
+
+proc hashTree*(n: PNode; flags: set[ConsiderFlag] = considerAll): SigHash =
+  ## Compute the hash of an AST node (`n`) using the given flags.
```
zerbina:
```suggestion
  ## Computes the hash of an AST node (`n`) using the given flags.
```

tools/dust/boring.nim
```diff
@@ -0,0 +1,118 @@
...
+  if argsCount == 0:
+    # the first argument is the file
+    config.projectName = unixToNativePath(p.key)
+    config.arguments = cmdLineRest(p) # consume the rest
+    result = true
+  inc argsCount
+
+
+proc cmdLine(pass: TCmdLinePass, cmd: openArray[string]; config: ConfigRef) =
+  ## Parse the command-line into the config.
```
zerbina:
```suggestion
  ## Parses the command-line into the config.
```

tools/dust/boring.nim
```diff
@@ -0,0 +1,118 @@
...
+          config.cliEventLogger(e)
+    of cmdArgument:
+      config.commandLine.add " "
+      config.commandLine.add p.key.quoteShell
+      if processArgument(pass, p, argsCount, config):
+        break
+
+
+proc compile*(graph: ModuleGraph) =
+  ## Compile a module graph.
```
zerbina:
```suggestion
  ## Compiles a module graph.
```

tools/dust/dust.nim
```diff
@@ -0,0 +1,190 @@
+## The main dust driver, orchestrating the AST reduction process.
+
+import std/os
```
zerbina:
```suggestion
```

tools/dust/dust.nim
```diff
@@ -0,0 +1,190 @@
...
+  compiler/front/options,
+  compiler/modules/modulegraphs,
+  compiler/sem/[
+    passes,
+    sem,
+    sighashes,
+  ],
+  compiler/utils/[astrepr,],
+  std/options as std_options, # due to legacy reports stupidity
+  std/[sets, algorithm,],
```
zerbina:
```suggestion
```

tools/dust/dust.nim
```diff
@@ -0,0 +1,190 @@
+## The main dust driver, orchestrating the AST reduction process.
+
+import std/os
+
+{.define(nimcore).}
+
+import
+  compiler/ast/[
```
zerbina, line 12:
```suggestion
  std/[
    algorithm,
    os,
    sets
  ],
  compiler/ast/[
```
To have all `std` imports in one place.

tools/dust/dust.nim
```diff
@@ -0,0 +1,190 @@
...
+  mutate
+
+# legacy reports stupidity
+from compiler/ast/reports import Report, location, kind
+from compiler/front/cli_reporter import reportFull, legacyReportBridge
+from compiler/front/msgs import defaultDiagHandler
+
+
+template semcheck(body: untyped) {.dirty.} =
+  ## Perform the complete setup and compilation process.
```
zerbina:
```suggestion
  ## First performs the complete setup, then runs `body`, and finally runs
  ## sem-checking on the graph.
```
Soft suggestion. I think this is a bit clearer.

zerbina (approved):
Looks good, thank you! `dust` is a nice addition to the set of compiler tools, providing a good basis for further minimization work.

I left a few minor suggestions for missed import cleanup and doc-comment adjustments.

As for your question about adding relevant documentation, I think things are fine as they are, at least for now. Having some document in `docs` describing what `dust` does and how its internals work is possible, of course, but without some automated job for flagging out-of-date documentation, I believe such document would only end up on the pile of stale and outdated pieces of out-of-band documentation.

saem:
sigh... I'm guessing it didn't run the leak sanitizer locally when I did ran the tools tests via `koch`, oh well, I'll have to figure that out and see what to make of the errors.

saem:
I presently can't get a local reproduction, so while I wait to hear back from others about how I might do that I took a stab in the dark and enabled `--gc:orc` for `dust` in case that's the issue, on the basis that the `vmrunner` requires it as well.

---

Reproduction steps I've tried:

- install liblsan (previously i got a linker failure until I installed it)
- compiled `dust` with all the flags from `leak.yml`
- ran `koch.py testtools` just like `leak.yml`
- result: I didn't get any failure spam
- my os is Fedora 43 workstation, if that helps

tools/koch/koch.nim
```diff
@@ -183,6 +183,12 @@ proc buildTool(toolname, args: string) =
   copyFile(dest="bin" / splitFile(toolname).name.exe, source=toolname.exe)
 
 proc buildTools(args: string = "") =
+  let dustArgs = if "-d:leakTest" in args: ""
+                 else: args
+                 ## dust doesn't support leak testing until `TAstDiag` is no
+                 ## longer marked `.acyclic` or ideally removed
+  nimCompileFold("Compile dust", "tools/dust/dust.nim",
+                 options = "-d:release --gc:orc $# $#" % [defineSourceMetadata(), dustArgs])
```
saem: @zerbina FYI, I've removed `dust` from leak checking, the `.acyclic` pragma on `TAstDiags` seems to be the culprit.

saem: I've removed this after merging in the removal of `.acyclic` from `TAstDiag`.

PR #980, opened 2023-10-19
Cull some node flags and remove the {.shallow.} noop

Clyybber:
## Summary
* Removed  `nfNone`  as it was unused outside of the unused  `hasCyclic` 
proc
* Removed  `nfBase2|8|16`  and instead use an enum field in the
respective PNode variant ( `nkIntLiterals`  and  `nkFloatLiterals` )
* Removed the no-op pragma  `{.shallow.}`  and its associated typeflag 
`tfShallow` 

## Details
* Because  `sons*: TNodeSeq`  in  `TNode`  is 16 bytes big anyways, this
leaves us 8 bytes of unused space in the  `nkIntLiterals`  and 
`nkFloatLiterals`  variants, which can thus be used for an enum field of
type  `NumericalBase`  without increasing the size of  `TNode` .
* Separate fields for the  `nkIntLiterals`  and  `nkFloatLiterals` 
variants are used because variants can't share fields yet
* Because  `ast/trees.sameTree`  checks the node flags for equality and
those determined the base of a literal, checks for the introduced base
fields have been added to it in order to not change the behaviour 
* Should the removed  `hasCyclic`  proc be needed in the future it
should be easy to re-implement without using a node flag for storing
whether a node has already been visited, instead using a Set or 
`IntSet` .
* The  `nfAllFieldsSet`  constant/alias of  `nfBase2`  is unused as
there is no code that checks for it
* I considered adding  `{.pragma: shallow.}`  to  `system`  but I think
its usage is not that common, considering it was underspecified and not
memory safe.

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -472,10 +472,10 @@ template copyNodeImpl(dst, src, processSonsStmt) =
   case src.kind
   of nkIntLiterals:
     dst.intVal = src.intVal
-    dst.intLitBase= src.intLitBase
+    dst.intLitBase = src.intLitBase
   of nkFloatLiterals:
     dst.floatVal = src.floatVal
-    dst.floatLitBase= src.floatLitBase
+    dst.floatLitBase = src.floatLitBase
   of nkSym: dst.sym = src.sym
   of nkIdent: dst.ident = src.ident
   of nkStrLit..nkTripleStrLit: dst.strVal = src.strVal
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -470,8 +470,12 @@ template copyNodeImpl(dst, src, processSonsStmt) =
       echo "GOES TO ", dst.id
       writeStackTrace()
   case src.kind
-  of nkCharLit..nkUInt64Lit: dst.intVal = src.intVal
-  of nkFloatLiterals: dst.floatVal = src.floatVal
+  of nkIntLiterals:
+    dst.intVal = src.intVal
+    dst.intLitBase= src.intLitBase
+  of nkFloatLiterals:
+    dst.floatVal = src.floatVal
+    dst.floatLitBase= src.floatLitBase
   of nkSym: dst.sym = src.sym
   of nkIdent: dst.ident = src.ident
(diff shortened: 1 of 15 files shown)


compiler/ast/ast.nim
```diff
@@ -471,8 +470,12 @@ template copyNodeImpl(dst, src, processSonsStmt) =
       echo "GOES TO ", dst.id
       writeStackTrace()
   case src.kind
-  of nkCharLit..nkUInt64Lit: dst.intVal = src.intVal
-  of nkFloatLiterals: dst.floatVal = src.floatVal
+  of nkIntLiterals:
+    dst.intVal = src.intVal
+    dst.intLitBase= src.intLitBase
```
zerbina:
```suggestion
    dst.intLitBase = src.intLitBase
```

compiler/ast/ast.nim
```diff
@@ -471,8 +470,12 @@ template copyNodeImpl(dst, src, processSonsStmt) =
...
+  of nkFloatLiterals:
+    dst.floatVal = src.floatVal
+    dst.floatLitBase= src.floatLitBase
```
zerbina:
```suggestion
    dst.floatLitBase = src.floatLitBase
```

zerbina (approved):
Looks good to me, and I like it. I left two minor style related suggestions.

For your info, I also made some small fixes to the PR message.

compiler/ast/ast.nim
Clyybber: Thanks, must have hit backspace accidentally :D

compiler/ast/ast_types.nim
```diff
@@ -1541,8 +1536,12 @@ type
     case kind*: TNodeKind
     of nkCharLit..nkUInt64Lit:
       intVal*: BiggestInt
+      intLitBase*: NumericalBase
     of nkFloatLit..nkFloat128Lit:
       floatVal*: BiggestFloat
+      floatLitBase*: NumericalBase
+        # Once case branches can share fields this can be unified with
+        # intLitKind above
```
Clyybber, line 1544:
```suggestion
        # intLitBase above
```

saem (approved):

PR #981, opened 2023-10-20
suggest: refactor `proc suggestResult`

bung87:
## Summary
* `proc suggestResult` takes `ModuleGraph` as first argument instead of `ModuleGraph`'s field `config`
* Avoiding access `config` field in its usages, make code line shorter.

## Details
* Additionally remove obsolete `when defined(nimsuggest)` statement upper `proc listUsages`.

---

## Notes for Reviewers

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


zerbina (changes requested):
Widening the interface of a procedure for the sake of shortening the code at the callsites is not acceptable, sorry.

If the amount of code duplication due to parameter passing would be larger, a proxy template would make sense, but the improvements in this case are so small that I do not think that this is warranted.

bung87:
ok,  get it

PR #982, opened 2023-10-20
Include nkCommentStmt in nkWithoutSons

Clyybber:
## Summary
*  `nkCommentStmt`  never had sons, so it should be included in 
`nkWithoutSons` 

## Details
*  `compiler/ast/trees.sameTree`  now checks the comments of 
`nkCommentStmt` s for equality, although this doesn't actually have an
effect on anything as that proc is unused.

### DIFF
--- a/compiler/ic/packed_ast.nim
+++ b/compiler/ic/packed_ast.nim
@@ -139,7 +139,9 @@ proc addModuleId*(tree: var PackedTree; s: ModuleId; info: PackedLineInfo) =
 proc addSymDef*(tree: var PackedTree; s: SymId; info: PackedLineInfo) =
   tree.nodes.add PackedNode(kind: nkSym, operand: int32(s), info: info)
 
-proc isAtom*(tree: PackedTree; pos: int): bool {.inline.} = tree.nodes[pos].kind <= nkNilLit
+proc isAtom*(tree: PackedTree; pos: int): bool {.inline.} =
+  let kind = tree.nodes[pos].kind
+  kind <= nkNilLit or kind == nkCommentStmt
 
 proc copyTree*(dest: var PackedTree; tree: PackedTree; n: NodePos) =
   # and this is why the IR is superior. We can copy subtrees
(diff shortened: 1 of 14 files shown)


compiler/ic/packed_ast.nim
```diff
@@ -139,7 +139,9 @@ proc addModuleId*(tree: var PackedTree; s: ModuleId; info: PackedLineInfo) =
 proc addSymDef*(tree: var PackedTree; s: SymId; info: PackedLineInfo) =
   tree.nodes.add PackedNode(kind: nkSym, operand: int32(s), info: info)
 
-proc isAtom*(tree: PackedTree; pos: int): bool {.inline.} = tree.nodes[pos].kind <= nkNilLit
+proc isAtom*(tree: PackedTree; pos: int): bool {.inline.} =
+  let kind = tree.nodes[pos].kind
+  kind <= nkNilLit or kind == nkCommentStmt
```
saem, line 144: It's probably a bug in the making of we don't make this a comparison with `nkWithoutSons`

compiler/sem/evaltempl.nim
```diff
@@ -91,10 +91,7 @@ proc evalTemplateAux(templ, actual: PNode, c: var TemplCtx, result: PNode) =
     # in the AST that would confuse it (bug #9432), but only if we are not in a
     # "declarative" context (bug #9235).
     if c.isDeclarative:
-      var res = copyNode(c, templ, actual)
-      for i in 0..<templ.len:
-        evalTemplateAux(templ[i], actual, c, res)
-      result.add res
+      result.add copyNode(c, templ, actual)
```
saem, line 94: Good catch

saem (approved):
One suggestion for good measure

compiler/ic/packed_ast.nim
Clyybber: Yeah I think so too, I'm making a seperate PR that does a sweep across the codebase that replaces these kinds of checks with the respective sets so I figured I'd leave this one as minimal as can be, as there are issues with running the testsuite with a debug compiler, so I want to hold off changes like this that could subtly influence behaviour until that's fixed

Clyybber:
Thanks for the review!

PR #983, opened 2023-10-20
Fix wrapError crash in semobjconstr

Clyybber:
## Summary
*  `wrapError`  inside  `sem/semobjconstr`  would fail an assertion
because there was no  `nkError`  node inserted when the object
constructor was missing fields, instead a  `localReport`  was used, but 
`wrapError`  was still called at the end of the procedure

## Details
* This never got caught by CI because the CI compiler is compiled with 
`-d:danger` /assertions disabled.


saem (approved):
nice catch.

PR #984, opened 2023-10-20
testament: don't error on directories without tests

Clyybber:
## Summary
This makes  `testament`  not crash on directories that don't contain
tests to run as there's no good reason to consider that an error

## Details
These can linger around after checking out an old branch that has
different test categories and running the tests there, because git will
not delete the directories when switching the branch as they will
contain binaries from the tests.


saem (approved):
Sounds good to me.

PR #985, opened 2023-10-21
sem: improve type class support for built-ins

zerbina:
## Summary

* the `array`, `set`, `seq`, `openArray`, `varargs`, and
  `UncheckedArray` constraints can now be used on multiple generic
  parameters without restricting the parameters to all be bound the
  same type
* type-classes of the `array`, `set`, `range`, `Ordinal`, `openArray`,
  `varargs`, and `UncheckedArray` built-in types can now be used as
  return types and as routine parameter types (making the routine
  implicitly generic)
* all type-classes for built-in generic types (except `typedesc`) are
  now bind-once. The manual already stated that they were, but reality
  didn't reflect that
* `lent`, `range`, `Ordinal`, `openArray`, `varargs`, and
  `UncheckedArray` are promoted to proper type classes -- it was
  already possible to use them as such, but this wasn't documented
* errors when using built-in type-classes in the body of non-generic
  tuples and objects are reported earlier
* it is now possible to create aliases of all built-in type-classes
  except `typedesc`, `object`, `enum`, `tuple`
* `getType`, `getTypeImpl`, and `getTypeInst` now work for all
  `tyBuiltInTypeClass` types. The compiler would previously crash for
  some of them

Breaking changes:
* `sink` cannot be used as a constraint or type-class anymore
* consistent with `ptr`, `ref`, and so on, the `array`, `set`, `seq`,
  `openArray`, `varargs`, `lent`, `UncheckedArray` type classes no
  longer match against anything else besides the type class. For
  example, `seq is seq[auto]` now returns false

The goal is to make their behaviour and implementation more consistent
in order for them to more amendable to changes in the future.

### Previous implementation

* using a built-in generic type defined in the `system` module resulted
  in a type like, for example, `array[I, T]`
  (`(Array (GenericParam I) (GenericParam T))`)
* when these types were used a routine parameter type on an otherwise
  non-generic routine, they were rejected by `typeAllowed` (due to the
  generic parameter)
* these types weren't designated as being meta-types, so meta-type
  detection using `isMetaType` missed them
* when used as generic parameter constraints, their type representation
  allowed most of them to work without special handling (but this
  caused issues when the type was used as a constraint multiple times,
  as binding of generic parameters is enabled when matching against
  constraints)
* a `tyNone` marker type was added to the base instances of `tyRange`
  and `tyOrdinal`, which allowed `typeRel` to detect these instances
  and special case them
* the `seq` type was special cased in `liftParamType` and a
  `tyBuiltInTypeClass` lifted from it

### Reworked implementation

* all built-in generic types that have no parameters specified are
  treated as a type class (`tyBuiltInTypeClass`)
* lifting the `tyBuiltInTypeClass` happens immediately after analysing
  a stand-alone type identifier/symbol (in `semTypeNode`)
* `semTypeNode` performs the lifting via the new `liftTypeClass`
* `liftTypeClass` takes a type as input and, if required it's a raw
  magic generic type, produces a `tyBuiltInTypeClass`. The generic
  magic type is stored as the first children 
* for the bind-once behaviour to work, the type needs an identifier in
  `liftParamType`. For `tyBuiltInTypeClass`es of `system`-defined
  types, the identifier is taken from magic type's symbol, while for
  the other kinds an identifier is looked up using the `TSpecialWord`
  corresponding to the type
* this way of implementing the lifting prevents the creation of extra
  symbols

So that types other than `Ordinal`, `openArray`, and `varargs` continue
to match against the respective type classes, the `typeRel` logic for
`tyBuiltInTypeClass` matches the actual type against the raw generic
type stored in the type-class type (but with all type binding
disabled).

`sigmatch` is adjusted accordingly:
* the special handling of `range` type-classes is moved to the
  `tyBuiltInTypeClass` branch
* `tyNone` detection for `tyOrdinal` is removed (it's obsolete now)

There were also two problems with structure of `tyBuiltInTypeClass` not
being accounted for properly, but they remained mostly hidden due to
the seldom use of the type:
* `hashType` was missing dedicated handling, which would lead to the
  compiler crashing when attempting to hash certain
  `tyBuiltInTypeClass`
* `mapTypeToAst` was also missing special missing handling

Multiple clean-ups and improvements are applied:
* adding `tyNone` markers to the `tyRange` and `tyOrdinal` base
  instances is obsolete and thus removed
* standalone `sink` usage is now rejected by `liftTypeClass`. Since
  `sink` is a special type modifier only allowed for routine
  parameters, it makes no sense to allow it as a constraint or type-
  class. A corresponding `SemReport` kind for the new error is added 
* the workaround for in `fitNodePostMatch` is cleaned up
* the user-facing rendering of `tyBuiltInTypeClass` types is improved

### DIFF
--- a/compiler/ast/typesrenderer.nim
+++ b/compiler/ast/typesrenderer.nim
@@ -264,6 +264,7 @@ proc typeToString*(typ: PType, prefer: TPreferedDesc = preferName): string =
     of tyBuiltInTypeClass:
       result = case t.base.kind
         of tyVar: "var"
+        of tyLent: "lent"
         of tyRef: "ref"
         of tyPtr: "ptr"
         of tySequence: "seq"
@@ -271,11 +272,16 @@ proc typeToString*(typ: PType, prefer: TPreferedDesc = preferName): string =
         of tySet: "set"
         of tyRange: "range"
         of tyDistinct: "distinct"
+        of tyOrdinal: "Ordinal"
         of tyProc: "proc"
+        of tyEnum: "enum"
         of tyObject: "object"
         of tyTuple: "tuple"
         of tyOpenArray: "openArray"
-        else: typeToStr[t.base.kind]
+        of tyVarargs: "varargs"
+        of tyUncheckedArray: "UncheckedArray"
+        of {low(TTypeKind)..high(TTypeKind)} - tyBuiltInTypeClasses:
+          "<illegal type: " & $t.base.kind & ">"
     of tyInferred:
       let concrete = t.previouslyInferred
       if concrete != nil: result = typeToString(concrete)
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -56,10 +56,10 @@ proc liftTypeClass(c: PContext, typ: PType, prev: PType,
 
   let isMagic = typ.sym != nil and typ.sym.magic != mNone
 
-  case result.kind
-  of tyOrdinal, tyRange, tySequence, tySet, tyArray, tyLent, tyOpenArray,
-     tyVarargs, tyUncheckedArray:
-    if isMagic:
+  if isMagic:
+    case result.kind
+    of tyOrdinal, tyRange, tySequence, tySet, tyArray, tyLent, tyOpenArray,
+       tyVarargs, tyUncheckedArray:
       # this is the "raw", uninstantiated magic type -> it's
       # a built-in type class
       result = newOrPrevType(tyBuiltInTypeClass, prev, c)
@@ -67,12 +67,11 @@ proc liftTypeClass(c: PContext, typ: PType, prev: PType,
       # add the generic type, but don't propagate the flags
       result.sons = @[typ]
 
-  of tySink:
-    if isMagic:
+    of tySink:
       localReport(c.config, info, SemReport(kind: rsemSinkIsNotATypeClass))
       result = newOrPrevType(tyError, prev, c)
-  else:
-    discard
+    else:
+      discard
 
 proc semEnum(c: PContext, n: PNode, prev: PType): PType =
   if n.len == 0: return newConstraint(c, tyEnum)
--- a/compiler/sem/semtypes.nim
+++ b/compiler/sem/semtypes.nim
@@ -1237,6 +1237,25 @@ template shouldHaveMeta(t) =
   c.config.internalAssert tfHasMeta in t.flags
   # result.lastSon.flags.incl tfHasMeta
 
+proc getTypeIdent(cache: IdentCache, typ: PType): PIdent =
+  ## Returns the identifier for built-in types that may be used as
+  ## type classes.
+  case typ.kind
+  of tyObject:    cache.getIdent($wObject)
+  of tyEnum:      cache.getIdent($wEnum)
+  of tyTuple:     cache.getIdent($wTuple)
+  of tyProc:      cache.getIdent($wProc)
+  of tyDistinct:  cache.getIdent($wDistinct)
+  of tyVar:       cache.getIdent($wVar)
+  of tyPtr:       cache.getIdent($wPtr)
+  of tyRef:       cache.getIdent($wRef)
+  of tySequence, tyOrdinal, tyRange, tySet, tyLent, tyArray, tyOpenArray,
+      tyVarargs, tyUncheckedArray:
+    # these use a symbol, take the identifier from there
+    typ.sym.name
+  of {low(TTypeKind)..high(TTypeKind)} - tyBuiltInTypeClasses:
+    unreachable(typ.kind)
+
 proc addImplicitGeneric(c: PContext; typeClass: PType, typId: PIdent;
                         info: TLineInfo; genericParams: PNode;
                         paramName: string): PType =
@@ -1323,17 +1342,7 @@ proc liftParamType(c: PContext, procKind: TSymKind, genericParams: PNode,
 
   of tySequence, tySet, tyArray, tyOpenArray,
      tyVar, tyLent, tyPtr, tyRef, tyProc:
-    # XXX: this is a bit strange, but proc(s: seq)
-    # produces tySequence(tyGenericParam, tyNone).
-    # This also seems to be true when creating aliases
-    # like: type myseq = distinct seq.
-    # Maybe there is another better place to associate
-    # the seq type class with the seq identifier.
-    if paramType.kind == tySequence and paramType.lastSon.kind == tyNone:
-      let typ = c.newTypeWithSons(tyBuiltInTypeClass,
-                                  @[newTypeS(paramType.kind, c)])
-      result = addImplicitGeneric(c, typ, paramTypId, info, genericParams, paramName)
-    else:
+    if true:
       for i in 0..<paramType.len:
         if paramType[i] == paramType:
           globalReport(c.config, info, reportTyp(rsemIllegalRecursion, paramType))
@@ -1410,6 +1419,11 @@ proc liftParamType(c: PContext, procKind: TSymKind, genericParams: PNode,
 
   of tyUserTypeClasses, tyBuiltInTypeClass, tyCompositeTypeClass,
      tyAnd, tyOr, tyNot:
+    if not anon and paramTypId == nil and paramType.kind == tyBuiltInTypeClass:
+      # for efficiency, built-in type-classes don't use a symbol by
+      # default. We fetch the identifier here.
+      paramTypId = getTypeIdent(c.cache, paramType.base)
+
     result = addImplicitGeneric(c,
         copyType(paramType, nextTypeId c.idgen, getCurrOwner(c)), paramTypId,
         info, genericParams, paramName)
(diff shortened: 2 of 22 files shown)


compiler/ast/typesrenderer.nim
```diff
@@ -264,18 +264,24 @@ proc typeToString*(typ: PType, prefer: TPreferedDesc = preferName): string =
     of tyBuiltInTypeClass:
       result = case t.base.kind
         of tyVar: "var"
+        of tyLent: "lent"
```
saem, line 267: my thinking might be too shallow, but why allow `lent` if we don't allow `sink` or vice versa?

compiler/sem/semtypes.nim
```diff
@@ -47,6 +47,33 @@ proc reportMeta(c: PContext; info: TLineInfo; t: PType) =
...
+      result.flags.incl tfCheckedForDestructor
+      # add the generic type, but don't propagate the flags
+      result.sons = @[typ]
+
+  of tySink:
+    if isMagic:
+      localReport(c.config, info, SemReport(kind: rsemSinkIsNotATypeClass))
+      result = newOrPrevType(tyError, prev, c)
+  else:
+    discard
```
Clyybber: Could be turned inside-out, although it put's the focus more on the `isMagic` check:
```suggestion
  if isMagic:
    case result.kind
    of tyOrdinal, tyRange, tySequence, tySet, tyArray, tyLent, tyOpenArray,
       tyVarargs, tyUncheckedArray:
      # this is the "raw", uninstantiated magic type -> it's
      # a built-in type class
      result = newOrPrevType(tyBuiltInTypeClass, prev, c)
      result.flags.incl tfCheckedForDestructor
      # add the generic type, but don't propagate the flags
      result.sons = @[typ]

    of tySink:
      localReport(c.config, info, SemReport(kind: rsemSinkIsNotATypeClass))
      result = newOrPrevType(tyError, prev, c)
    else:
      discard
```

compiler/sem/sigmatch.nim
```diff
@@ -1733,15 +1728,35 @@ typeRel can be used to establish various relationships between types:
...
+
+      if targetKind == effectiveArgType.kind and
+         not effectiveArgType.isEmptyContainer:
+        # same type kind -> match
+        result = isGeneric
       else:
-        return isNone
+        case targetKind
+        of tyOrdinal, tyOpenArray, tyVarargs:
+          # these type classes of these act the same as if using, for example,
```
```suggestion
          # the type classes of these act the same as if using, for example,
```

compiler/sem/sigmatch.nim
```diff
@@ -1733,15 +1728,35 @@ typeRel can be used to establish various relationships between types:
...
+          # ``openArray[T]``. The generic type is already stored as the base
+          # type -- but don't bind anything
+          result = typeRel(c, f.base, a, {trDontBind})
+        of tyRange:
+          if a.kind == tyRange:
+            result = isGeneric
+        of tyArray:
+          # XXX: empty array types match for the array type class, but this is
+          #      inconsistent with, for example, sequence types
```
saem: clarifying for myself, this is where a literal `[]` would be typed as an "empty array" and that would match `array`  the type class?

tests/lang_types/typeclass/tbuiltin_type_classes_bind_once.nim
```diff
@@ -0,0 +1,55 @@
+discard """
+  description: '''
+    All built-in type classes are *named*, and they thus have bind-once
+    behaviour. ``typedesc`` is an exception.
```
saem, line 4: IMO, the manual's claim of bind once seems like an unprincipled decision and instead one of convenience wrt implementation bugs at the time. I'm not firm on this, but I can't think of a good reason to have it work this way.

I don't think any adjustments need to be made per my comment above, I'd prefer to discuss it further before settling on something.

Briefly, I think the following are saying different things:

```nim
# type classes
proc foo(a: array, b: array) # two params, each of some type satisfying `array`
proc foo(a, b: array) # same as above, as I believe pragmas require a copy of the type ast during normalization

# array[T]/array[T:any]
proc foo[T](a: array[T], b: array[T]) # only one specific `array[T]`
proc foo[T](a, b: array[T]) # same as above, even though `array[T]` is copied per param during normalization due to pragmas, `T` only binds once
proc foo[T: any](a: array[T], b: array[T]) # same as above, this is technically the most "expanded" version of the three
```

Again, I don't think this PR should necessarily change its direction.

compiler/sem/semtypes.nim
saem: just my two cents, feel free to ignore, being `isMagic` forward is a good thing as it seems to be the first fact that we need to establish in order to "do the right thing".

compiler/sem/sigmatch.nim
saem: further observation: In hindsight after my comment/observation with respect to bind-once behaviour of type classes, I think this behaviour makes a certain amount of sense, as an empty array with an unbound contained type (`any`) does very much still satisfy the type class.

saem (approved):
This is fantastic, the mechanics seem more clear, with fewer edge cases, and the exhaustiveness of the tests alone really help clarify what should and shouldn't happen.

compiler/sem/semtypes.nim
```diff
@@ -2256,12 +2297,16 @@ proc semTypeNode(c: PContext, n: PNode, prev: PType): PType =
...
-      if alias != nil:
-        result = alias
+      let lifted = liftTypeClass(c, s.typ, prev, n.info)
+      if lifted != s.typ:
+        result = lifted
       else:
+        let alias = maybeAliasType(c, s.typ, prev)
+        if alias != nil:
+          return alias
+
```
Clyybber: Could be turned into
```nim
    else:
      let lifted = liftTypeClass(c, s.typ, prev, n.info)
      if lifted != s.typ or prev == nil:
        result = lifted
      else:
        let alias = maybeAliasType(c, s.typ, prev)
        if alias != nil:
          return alias
```
which would make it more similar to the branching below: https://github.com/nim-works/nimskull/pull/985/files#diff-edd10e83e69d8a646ce08fedd7dd9d2487fa08045c3fc3f53c4114b683930a71R2326-R2337

Clyybber (approved):

compiler/ast/typesrenderer.nim
zerbina: Since `var` is allowed as a type class, my thinking was that `lent` should be allowed too, as it may also be used as the type of a field, local, or result variable.

`sink` only annotates a parameter as having full ownership of the value, and the type cannot appear outside of parameter definitions. Outside of destruction injection and move analysis, it is as if `sink` (`tySink`) doesn't exist, whereas this is not the case for `lent`.

---

At present, both `var` and `lent` have the problem that `x is var` and `x is lent` checks don't work due to `semExprWithType` always inserting a hidden dereference, but this is fixable/changeable.

compiler/sem/sigmatch.nim
zerbina: > clarifying for myself, this is where a literal `[]` would be typed as an "empty array" and that would match `array` the type class?

Yep.

> I think this behaviour makes a certain amount of sense

I also think it makes sense, although there is the problem that `proc p(x: array)` used like `p([])` would cause the compiler to crash. In addition, the behaviour is inconsistent with `seq`: a `@[]` (`(Sequence (Empty))`) doesn't match the `seq` type class.

In any way, I just noticed that `tySet` also needs to be special-cased in the same way that `tyArray` is.

tests/lang_types/typeclass/tbuiltin_type_classes_bind_once.nim
zerbina: Once case to consider could be the following:
```nim
proc p(a: array): array # the return type is the same as the parameter's type
```
I'm not entirely sure, but this could be useful.

Anyhow, I don't think I'd object to removing the "bind once" behaviour. For now, my goal was to make the language work as specified in the manual.

compiler/ast/typesrenderer.nim
saem: Thanks for the detailed explanation, something is a bit weird about it, but that's well outside the scope of this pr. I'll do some more poking around and see if I can come up with some further progress/thoughts.

tests/lang_types/typeclass/tbuiltin_type_classes_bind_once.nim
saem: > Once case to consider could be the following:
> 
> ```nim
> proc p(a: array): array # the return type is the same as the parameter's type
> ```
> 
> I'm not entirely sure, but this could be useful.

Yup, it's a bit convenient but it might be better if the return type was `typeOf(x)`. Type classes and implicit generics are much weirder the more I think about them.

> Anyhow, I don't think I'd object to removing the "bind once" behaviour. For now, my goal was to make the language work as specified in the manual.

That makes sense, it's also why I don't think this pr should change because of it.

zerbina:
Thank you for the review, @Clyybber and @saem!

I've addressed the review comments and fixed the empty set literal (`{}`) not matching for a `set` type class (this would have been a regression).

PR #986, opened 2023-10-22
Semfold type expressions into nkType nodes

Clyybber:
## Summary
* Fold type expressions into  `nkType`  nodes in  `semfold`  so that the
later stages don't have to deal with the unwieldy type AST

## Details
* This fixes a debug/release compiler crashing on 
`tests/vm/taccess_checks.nim`

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -106,6 +106,27 @@ const
 
   nkSymChoices* = {nkClosedSymChoice, nkOpenSymChoice}
 
+  nkTypeExprs* = {
+    nkTypeOfExpr,
+    nkObjectTy,
+    nkTupleTy,
+    nkTupleClassTy,
+    nkTypeClassTy,
+    nkStaticTy,
+    nkRefTy,
+    nkPtrTy,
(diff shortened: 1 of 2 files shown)


compiler/ast/ast_query.nim
```diff
@@ -106,6 +106,27 @@ const
...
+    nkVarTy,
+    nkConstTy,
+    nkMutableTy,
+    nkDistinctTy,
+    nkProcTy,
+    nkIteratorTy,
+    nkSharedTy,
+    nkEnumTy,
+    nkStmtListType,
+    nkBlockType
```
saem, line 127: I wonder if we should not include these under expressions?

saem (approved):
One item to consider, but otherwise I'm good with merging it.

compiler/ast/ast_query.nim
Clyybber: Yeah, it doesn't matter much because I planned on removing them anyways; `nkBlockType` isn't used and `nkStmtListType` can be replaced with `nkStmtListExpr` I think.

PR #995, opened 2023-10-24
fix(sem): missed `discard` in `try/finally` errors

saem:
## Summary

A missing `discard` in a `try/finally` no longer crashes the compiler
and emits an error as it should.

## Details

The `discardCheck` used to ensure expressions are used or `void` did not
traverse `try` appropriately. Instead of looking at the last expression
in the last `except` or `try` branch, it would look at the last branch,
even if that was a `finally`. This led to incomplete data being sent for
error reporting, which ultimately led to an NPE.

Along with the fix a regression test has been added.

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -307,15 +307,15 @@ proc semTry(c: PContext, n: PNode; flags: TExprFlags): PNode =
     if n[0].isError:
       return wrapError(c.config, n)
     for i in 1 ..< n.len:
-      n[i] = discardCheck(c, n[i], flags)
-      if n[i].isError:
+      n[i][^1] = discardCheck(c, n[i][^1], flags)
+      if n[i][^1].isError:
         return wrapError(c.config, n)
     if typ == c.enforceVoidContext:
       result.typ = c.enforceVoidContext
   else:
     if n.lastSon.kind == nkFinally:
-      n[^1] = discardCheck(c, n[^1], flags)
-      if n[^1].isError:
+      n[^1][^1] = discardCheck(c, n[^1][^1], flags)
+      if n[^1][^1].isError:
         return wrapError(c.config, n)
     n[0] = fitNode(c, typ, n[0], n[0].info)
     for i in 1..last:
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -141,9 +141,9 @@ proc discardCheck(c: PContext, n: PNode, flags: TExprFlags): PNode =
             case m[^1].kind
             of nkFinally:
               m[^2]
-            of skipForDiscardable - nkFinally:
-              m.lastSon
-            of nkAllNodeKinds - skipForDiscardable:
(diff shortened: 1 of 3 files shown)


compiler/sem/semstmts.nim
```diff
@@ -135,7 +135,18 @@ proc discardCheck(c: PContext, n: PNode, flags: TExprFlags): PNode =
...
-        m = m.lastSon
+        case m.kind
+        of nkTryStmt:
+          m =
+            case m[^1].kind
+            of nkFinally:
+              m[^2]
+            of skipForDiscardable - nkFinally:
+              m.lastSon
+            of nkAllNodeKinds - skipForDiscardable:
```
```suggestion
            of nkExceptBranch:
              m[^1]
            of nkAllNodeKinds - {nkFinally, nkExceptBranch}:
```
This adds another layer of security against unexpected AST.

compiler/sem/semstmts.nim
```diff
@@ -296,16 +307,15 @@ proc semTry(c: PContext, n: PNode; flags: TExprFlags): PNode =
     if n[0].isError:
       return wrapError(c.config, n)
     for i in 1 ..< n.len:
-      n[i][^1] = discardCheck(c, n[i][^1], flags)
-      if n[i][^1].isError:
+      n[i] = discardCheck(c, n[i], flags)
+      if n[i].isError:
```
```suggestion
      n[i][^1] = discardCheck(c, n[i][^1], flags)
      if n[i][^1].isError:
```

compiler/sem/semstmts.nim
```diff
@@ -296,16 +307,15 @@ proc semTry(c: PContext, n: PNode; flags: TExprFlags): PNode =
...
         return wrapError(c.config, n)
-
     if typ == c.enforceVoidContext:
       result.typ = c.enforceVoidContext
   else:
     if n.lastSon.kind == nkFinally:
-      n[^1][^1] = discardCheck(c, n[^1][^1], flags)
-      if n[^1][^1].isError:
+      n[^1] = discardCheck(c, n[^1], flags)
+      if n[^1].isError:
```
```suggestion
      n[^1][^1] = discardCheck(c, n[^1][^1], flags)
      if n[^1][^1].isError:
```

zerbina (commented):
The node indexing changes in `semTry` need to be reverted, otherwise the following:
```nim
try:
  discard
except:
  1
```
is no longer reported as being incorrect. This is because the type of both `nkExceptBranch` and `nkFinally` is always nil, meaning that the `discardCheck` call will do nothing.

tests/lang_exprs/tdiscardcheck_try_finally.nim
```diff
@@ -0,0 +1,16 @@
+discard """
+description: "Ensure discard checks are done for try expressions"
+errormsg: "expression '1' is of type 'int literal(1)' and has to be used (or discarded)"
+line: 11
```
zerbina:
```suggestion
line: 14
```
This fixes the test failure.

saem:
@zerbina I added an extra test to avoid regressing on the example provided -- nice catch.

PR #997, opened 2023-10-26
sem: make return-type inference work consistently

zerbina:
## Summary

Inference of a routine's return type now works the same as inference in
the case of `var x: MetaType = y`. This means that generic converters
are now considered when inferring the return type.

Three bugs are fixed:
* return-type inference via `yield` no longer ignores the specified
  generic/meta return type
* the instance of a generic `and`/`or`/`not` type-class (e.g.,
  `Constraint[int]` where `type Constraint[T] = T or float`) now works
  as a constraint on `var`/`let`/`const` definitions and for the
  return type
* a concept type used as the return type is no longer replaced with a
  non-concept type when a `result` assignment exists in the body

## Details

Both `semAsgn` and `semYield` didn't implement the inference properly.
`semAsgn` only performed a `sigmatch`-based type comparison and on
success used the type of the source expression as the return type,
while `semYield` fully ignored the specified meta-type.

For symmetry with `var`, `let`, and `const` definitions, fitting the
expression with the inferrable return type is done via
`inferWithMetatype`, which, compared to `fitNode`, ensures that the
expression uses the resolved meta type instead of the expression's
original type.

**Main changes:**

* don't treat resolved concept types as allowing for inference in
  `resultTypeIsInferrable` -- they're already resolved
* move the return-type inference logic into the new `inferReturnType`
  procedure, which is used by both `semAsgn` and `semYield`
* use `inferWithMetatype` for fitting the source expression into the 
  result when the result type is still a meta type
* the `auto` type, which at the time of inference is represented via a
  `tyUntyped`, requires special handling
* use `inferWithMetatype` for fitting the expression in a macro/
  template using a generic and/or meta return type
* extend `inferWithMetatype` to handle `tyGenericInstance`s, which
  fixes instantiation failures
* `tyBuiltInTypeClass` types no longer reach into
  `fitNodePostMatch`, so the associated workaround is removed

**Error handling changes**:
* don't disable the void context enforcement for result assignments
  when the right-hand expression is erroneous
* infer the return type as `tyError` when the first analysed result
  assignment has an error

Both changes are intended to reduce error cascading.

**Misc changes:**

* don't mutate the input AST in `semYield`
* change `semYieldVarResult` to not create another copy -- its input
  is now already a production
* remove the obsolete `skMacro` check in `semAsgn` -- since
  https://github.com/nim-works/nimskull/pull/620, the return type of
  a macro is always `NimNode`

### DIFF
--- a/tests/lang/s02_core/s07_type_inference/t01_return_type_inference.nim
+++ b/tests/lang/s02_core/s07_type_inference/t01_return_type_inference.nim
@@ -1,0 +1,36 @@
+discard """
+  description: '''
+    The specified return type of a non-generic routines may be generic, in
+    which case the concrete type is inferred from the body.
+  '''
+"""
+
+## Using 'auto' as the return type enables unconstrained return type
+## inference. The real return type is taken from the source expression
+## of the first analysed assignment to the `result` variable.
+
+proc p(): auto =
+  result = 1
+
+doAssert typeof(p()) is int
+
+## The return type specified on the procedure definition may also be a
+## type-class of a built-in generic type, in which case it acts as a
+## constraint.
+
+proc p2(): tuple =
+  result = (1, 2)
+
+doAssert typeof(p2()) is (int, int)
+
+## Return type inference is also available with iterators.
+
+iterator iter(): auto =
+  yield 1
+
+doAssert typeof(iter()) is int
+
+iterator iter2(): tuple =
+  yield (1, 2)
+
+doAssert typeof(iter2()) is (int, int)
--- a/tests/lang/s02_core/s07_type_inference/t01_return_type_inference.nim
+++ b/tests/lang/s02_core/s07_type_inference/t01_return_type_inference.nim
@@ -1,6 +1,6 @@
 discard """
   description: '''
-    The specified return type of a non-generic routines may be generic, in
+    The specified return type of non-generic routines may be generic, in
     which case the concrete type is inferred from the body.
   '''
 """
--- a/tests/lang/s02_core/s07_type_inference/t01_return_type_inference_issues.nim
+++ b/tests/lang/s02_core/s07_type_inference/t01_return_type_inference_issues.nim
@@ -1,0 +1,20 @@
+discard """
+  knownIssue: '''
+    Standalone composite type-classes used as return types always result in
+    an error
+  '''
+"""
+
+type Generic[T] = (T, T)
+
+# currently fails with: "cannot instantiate"
+
+proc p3(): Generic =
+  result = ("", "")
+
+doAssert typeof(p3()).T is string
+
+iterator iter3(): Generic =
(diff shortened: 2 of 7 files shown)


tests/lang/s02_core/s07_type_inference/t01_return_type_inference.nim
```diff
@@ -0,0 +1,36 @@
+discard """
+  description: '''
+    The specified return type of a non-generic routines may be generic, in
```
saem:
```suggestion
    The specified return type of non-generic routines may be generic, in
```

tests/lang/s02_core/s07_type_inference/t01_return_type_inference.nim
```diff
@@ -0,0 +1,36 @@
...
+proc p(): auto =
+  result = 1
+
+doAssert typeof(p()) is int
+
+## The return type specified on the procedure definition may also be a
+## type-class of a built-in generic type, in which case it acts as a
+## constraint.
+
+proc p2(): tuple =
```
saem, line 21:
```suggestion
proc p2(): tuple =
  ## equivalent to `proc p2[R: tuple](): R =`
```

nice to have, just read slightly more spec like.

tests/lang/s02_core/s07_type_inference/t01_return_type_inference_issues.nim
```diff
@@ -0,0 +1,20 @@
+discard """
+  knownIssue: '''
+    Standalone composite type-classes used as return types always result in
```
saem, line 3: a proactive test is very nice.

tests/lang/s02_core/s07_type_inference/t02_return_type_inference_converters.nim
```diff
@@ -0,0 +1,39 @@
+discard """
+  description: '''
+    Generic converters are considered when inferring the return type.
```
saem, line 3: mind blown that this works.

saem (approved):
I like it.

My suggestions are very soft/minor.

tests/lang/s02_core/s07_type_inference/t01_return_type_inference.nim
zerbina: They're not equivalent, as `proc p2[R: tuple](): R` requires the call to specify the type at the callsite (no return type inference would be used then) whereas for `proc p2(): tuple` this is neither required nor possible.

The original changes included that `p2` wouldn't be a generic routine in the first place (it doesn't have to be one), but that seemed like something better changed separately.

saem: Ah, because it's in the signature so inference based on the body wouldn't happen.

PR #998, opened 2023-10-26
fix(sem): sigmatch matches check for errors before updating call node

saem:
## Summary
* what changed and how?
* why are we changing it?

## Details
* anything else

---

## Notes for Reviewers

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


zerbina:
Ah, so it's the known "template/macro with `untyped` parameter has overloads with typed parameter" problem. A reduced test case:
```nim
# ``test`` has to be an overloaded routine, where the overload with the non-
# untyped parameter has to be defined **first**

proc test(arg: string) =
  discard

template test(arg: untyped) =
  for it in arg:
    discard

# two overloads of ``test2`` have to exist, with one being an iterator while
# the other is not. The return types don't matter as long as the non-iterator
# doesn't return anything that matches `string` or something for which an
# ``items`` iterator exists.
iterator test2(): int =
  yield 0

proc test2(): bool =
  discard

test(test2())
```

saem:
Current minimal case to reproduce the existing issue:

```nim
type
  Foo = enum
    foo1
    foo2
    foo3

const someFoos = {foo2 .. foo3}

doAssert set[Foo](someFoos) == {foo2, foo3}
```

saem:
Minor update: That took a long while, but the issue is starting to become more clear. `matchesAux`'s `noMatchAux` template, which is used to signal errors, relies on `n`, the call's input AST, to contain the call arg, which is not the case after this change. This proc is likely going to require a fair bit of refactoring to get this to work properly.

saem:
This PR has been superceded by this one: https://github.com/nim-works/nimskull/pull/1102

PR #999, opened 2023-10-26
vm: simpler closure support

zerbina:
## Summary

Replace the VM's closure support with a simpler and much more generic
implementation. This makes the VM agnostic of how closures are
represented in-memory, thus giving the core language control over their
representation.

## Details

The VM previously only supported closures to be represented as a
procedure address + `ref` pair. That significantly restricts the
core language and effectively prevents support for stack-allocated
closure environments.

Closure support is now reduced to a calling convention: if the invoked
procedure is marked as using the `.closure` calling convention, the
real number of arguments is the number of arguments specified + 1.

### VM changes

* remove the `akCallable` type kind and everything related to it
* remove the `opcAccessEnv` and `opcWrClosure` opcodes and
  corresponding implementation
* `FuncTableEntry` only stores whether the procedure uses the closure
  calling convention -- the `envParamType` field is removed

### Type translation

A closure type is translated to a tuple type equivalent to
`(proc, ref RootObj)` (where `proc` is a procedure type). Since
`tyProc` with `callConv=ccClosure` can mean both "closure" and
"procedure type", all type translation routines now accept an
additional `noClosure` parameter that specifies which interpretation to
use.

The `ref RootObj` type is cached such that `vmtypegen` doesn't have to
re-generate it each time it's needed. However, the `ref RootObj` cannot
be set up in `initCtx`, as:
* the `vmtypegen` module isn't available there
* the `RootObj` type might not have been defined yet

Instead, the new `initRootRef` procedure is expected to be called prior
to translating closure types. The direction is to lower closure types
during the MIR phase, which will allow for removing the aforementioned
type translation additions and workarounds again.

### Code generator changes

For calls where the callee is a closure, `vmgen` allocates
arguments + 1 registers and unpacks the closure (i.e., the tuple): the
procedure address goes into the usual callee register while the
environment ref goes into the last register. If at run-time the callee
isn't a procedure using the `.closure` calling convention (which means
that the closure value represents a thunk), the trailing register is
ignored.

Multiple adjustments and additions are required now that the VM no
longer knows about closures:
* the `mIsNil` and `mEqProc` magics are implemented in bytecode. In the
  future, these magics should be lowered during the MIR phase
* `opcAccessEnv` is implemented in bytecode
* `opcWrClosure` is also implemented in bytecode. Because the
  type of the environment value expression is wrong (it's either
  `tyNil`) or a sub-type of `ref RootObj`, the handle has to be
  converted to a `ref RootObj` prior to being written to the
  destination
* in the context of the callee, the environment parameter has to be
  converted to the expected type first, as the caller is allowed (and
  does) pass it as a super-type `ref`

### De-/serialization

* create a thunk when putting a raw symbol into a register where a
  closure is expected (`putIntoReg`) -- this replaces the improper
  workaround
* update `serialize` and `deserialize` to treat closures as a tuple
  with two elements. `deserializeRef` is extended to support
  deserializing a full cell without having to know the cell's type at
  the callsite

### DIFF
--- a/compiler/vm/compilerbridge.nim
+++ b/compiler/vm/compilerbridge.nim
@@ -173,20 +173,22 @@ proc putIntoReg(dest: var TFullReg; jit: var JitState, c: var TCtx, n: PNode,
     dest.initLocReg(typ, c.memory)
     c.serialize(n, dest.handle)
   of tyProc:
-    # XXX: a hack required to uphold some expectations. Procedural types as
-    #      static macro arguments are underspecified
-    let pt =
-      if t.callConv == ccClosure and n.kind == nkSym:
-        # Force the location to be of non-closure type. This breaks other
-        # assumptions!
-        n.sym.typ
+    let val =
+      if t.callConv == ccClosure:
+        # do what ``transf`` would do and turn the expression into a closure
+        # construction
+        case n.kind
+        of nkSym:
+          newTreeIT(nkClosure, n.info, t, [n, newNode(nkNilLit)])
+        of nkClosure, nkNilLit:
+          n
+        else:
+          unreachable()
       else:
-        t
+        n
 
-    let typ = c.getOrCreate(pt)
-    dest.initLocReg(typ, c.memory)
-    c.serialize(n, dest.handle, pt)
-
+    dest.initLocReg(c.getOrCreate(t), c.memory)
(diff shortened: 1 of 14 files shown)


saem (approved):
Generalizing out closure support looks good, I like how it simplifies the opcodes.

PR #1000, opened 2023-10-27
refactor(sem): remove duplicated type assignment

zerbina:
## Summary

De-duplicate the handling of assigning a symbol's type by re-using the
unused `setVarType`.

## Details

Both `semNormalizedLetOrVar` and `semNormalizedConst` handle the type
assignment for the symbol in the same way, so it makes sense to move
the implementation into a shared procedure. The unused `setVarType`
procedure is repurposed for this and renamed to `setSymType`.


saem (approved):
Good eye, I might have just deleted it. 😅

PR #1001, opened 2023-10-27
refactor(sem): indirect operation analysis

zerbina:
## Summary

Refactor analysis and processing of indirect operation AST
(`semIndirectOp`). The immediate processing now no longer modifies the
input AST.

## Details

* don't modify the callee expression slot. Use a temporary variable
  instead
* move all callee node error handling into a single place
* type callee expressions without a type (i.e., a statement) as
  `tyVoid`. This removes the need to guard against nil and allows for
  using a single `case` statement for the main processing
* introduce the `copyNodeWithKids` routine for performing the common
  shallow copy operation on nodes

In addition, the call to `semOpAux` in the `tyTypeDesc` case is
removed; `semConv` and `semObjConstr` analyze all operands themselves.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -1252,7 +1252,9 @@ proc semIndirectOp(c: PContext, n: PNode, flags: TExprFlags): PNode =
 
   assert n.kind in nkCallKinds
 
-  let prc = n[0]
+  var callee = PNode(nil)
+    ## the analysed AST from the callee position
+
   case n[0].kind
   of nkDotExpr:
     # the callee position is a dotExpr (`a.b`) which could be:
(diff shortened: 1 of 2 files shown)


compiler/sem/semexprs.nim
```diff
@@ -1305,37 +1303,47 @@ proc semIndirectOp(c: PContext, n: PNode, flags: TExprFlags): PNode =
...
-    n[0] = semExpr(c, n[0], {efInCall})
-    if n[0] != nil and n[0].isErrorLike:
-      result = wrapError(c.config, n)
-      return
+    callee = semExpr(c, n[0], {efInCall})
 
-  if n[0].typ != nil and n[0].typ.kind in {tyVar, tyLent}:
+  # error handling:
+  if callee.kind == nkError:
+    result = copyNodeWithKids(n)
```
saem, line 1313: I can see us getting a fair bit of mileage out of this simple addition.

saem (approved):
One less misbehaving procedure, typing consistency with void, and a nice utility procedure to boot. 🤩

PR #1002, opened 2023-10-30
fix(sem): issues related to subscript analysis

zerbina:
## Summary

Refactor subscript analysis to not mutate the input AST, which, as a
side-effect, fixes two issues:
* fix the `[]` and `[]=` operator overloads for pointer-to-array-like
  or ref-to-array-like types being ignored
* fix a compiler crash on macro-generated `nnkDerefExpr` AST where the
  operand is not a `ref` or `ptr` expression

In addition:
* the error message for an out-of-bounds tuple access now includes the
  provided index and tuple length
* nullary generic routines can now be instantiated outside of callee
  expressions using the `routine[]` syntax

## Details

* make `semDerefEval` a standalone procedure that expects a sem'ed
  pointer-like expression and always returns a non-nil AST
* change `semDeref` to return an error when the expression is not a
  valid deref expression

The analysis of subscript operations (`semSubscript`) is restructured
and changed:
* explicit generic instantiations are tested for first. This render the
  `tyStatic` handling obsolete and enables support for instantiating
  nullary generic routines outside of callee positions
* don't fold the result of a type instantiation into a symbol node.
  Doing so was inconsistent with what the `tyTypeDesc` branch does
  (which was the branch previously taken in most cases)
* `semDerefEval` is dispatched to directly (without going through
  `semDeref`), but only if the dereference operator is available for
  the type
* the array-like and tuple subscript analysis is changed to only insert
  dereference operations when the subscript operator is a built-in one.
  This fixes `[]` and `[]=` operator overloads for pointer-to-array-
  like types being ignored
* analysis for the array-like and tuple `[]` operator is changed to not
  mutate the input AST
* processing of `macroOrTempl[...]` expressions is still incorrect, but
  fixing the problem is outside the scope of this commit

Instead of returning 'nil' when the subscript operator is not a built-
in one, `semSubscript` now returns a `(Call (Ident '[]') ...)` AST.
This is necessary for keeping the already-analyzed argument
expressions without having to modify the input AST. The callsites of
`semSubscript` are adjusted accordingly.

Finally, tests and regression tests are added for the fixes and
changes.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -1872,10 +1872,35 @@ proc buildOverloadedSubscripts(n: PNode, ident: PIdent): PNode =
   result.add(newIdentNode(ident, n.info))
   for s in n: result.add s
 
+proc semDerefEval(c: PContext, n: PNode): PNode =
+  ## Evaluates a ``nkDerefExpr`` expression where the operand was already
+  ## analyzed, returning either an error or `n` with the correct type
+  ## assigned.
+  result = n # `n` may be mutated directly
+
+  let
+    target = n[0]
+    targetAsConstExpr = getConstExpr(c.module, target, c.idgen, c.graph)
+    semmedTarget =
+      if targetAsConstExpr != nil:
+        targetAsConstExpr
+      else:
+        target
+
+  result[0] = semmedTarget
+  result.typ = elemType(target.typ.skipTypes({tyAlias, tyGenericInst, tySink}))
+
+  case semmedTarget.kind
+  of nkError:
+    result = c.config.wrapError(result)
+  of nkNilLit:
+    result = c.config.newError(result):
+      PAstDiag(kind: adSemDisallowedNilDeref)
+  else:
+    discard "all good"
+
 proc semDeref(c: PContext, n: PNode): PNode =
-  ## analyse deref such as `foo[]`, when given an nkBracketExpr converts it to
-  ## an nkDerefExpr and evals it, if given an nkDerefExpr evals it, nkError is
-  ## passed through.
+  ## Analyses a dereference expression such as ``foo[]`` and evaluates it.
   ##
   ## the behaviour is different if the deref target is a constant expression 
   ## (static) or not (dynamic).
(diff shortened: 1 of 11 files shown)


compiler/sem/semexprs.nim
```diff
@@ -1872,205 +1872,220 @@ proc buildOverloadedSubscripts(n: PNode, ident: PIdent): PNode =
...
+        # only insert the dereferences when the built-in overload matched
+        result[0] = makeDeref(target)
+
   of tyTypeDesc:
     # The result so far is a tyTypeDesc bound
     # a tyGenericBody. The line below will substitute
     # it with the instantiated type.
-    result = n
-    result.typ = makeTypeDesc(c, semTypeNode(c, n, nil))
+    # XXX: error propagation is missing
```
saem, line 2047: Should this `localReport` on an `nkError` in the meantime?

compiler/sem/semexprs.nim
```diff
@@ -1872,205 +1872,220 @@ proc buildOverloadedSubscripts(n: PNode, ident: PIdent): PNode =
...
-          kind: rsemInvalidTupleSubscript,
-          ast: n[1]))
+    if n.len == 2:
+      # [] operator for tuples requires constant expression:
+      result[1] = semRealConstExpr(c, n[1])
+
+      case skipTypes(result[1].typ, {tyGenericInst, tyAlias, tySink, tyRange,
+                                     tyOrdinal}).kind
+      of tyInt..tyInt64:
+        # XXX: unsigned integers should probably be allowed too...
```
saem, line 2057: Yup.

saem (approved):
The logic is much easier to follow now, looking at the old code, I wouldn't be surprised is plenty more bugs lay in wait.

compiler/sem/semexprs.nim
zerbina: There's nothing to `localReport` here, unfortunately. In case of an error, `semTypeNode` already `localReport`s internally and only returns a `tyError` with no AST.

saem: Better than the alternative of "leaking" the error node, I suppose.

PR #1003, opened 2023-10-31
CI: treat `XDeclaredButNotUsed` as an error

zerbina:
## Summary

* fix all `XDeclaredButNotUsed` warnings in the code bases of the
  compiler and tooling
* enable the hint for the compiler and associated tooling
* promote the hint to an error when strict mode (`-d:nimStrictMode`) is
  enabled

The aim is to prevent obsolete code from being merged or being left
behind after a cleanup/refactor.

## Details

The warnings are fixed through a mixture of approaches:
* unused routines and types that were unused for a long time
  and where it's unlikely that they'll become used again are removed
* unused locals with side-effect-less initializer are removed as a
  whole
* unused locals with side-effect-ful initializer expressions are
  replaced with `discard` statements
* symbols injected by templates are marked as `.used`
* currently unused routines and locals meant for debugging are kept and
  marked as `.used`
* some unused locals are not removed but instead made use of

A few places are of special note:
* the unused `a0` local in `semTemplBody` was assigned the result of
  a redundant `semTemplBody` call. The call does have side-effects, but
  thanks to in-place AST modifications the duplicate processing didn't
  cause problems
* the `assign` procedure in the `checked_ast` module was meant from the
  start to be exported
* there are multiple routines in the `suggest` module that are unused
  when building the compiler itself. `XDeclaredButNotUsed` hints are
  now disabled for the module when not building with the `nimsuggest`
  define

Finally, disabling the `XDeclaredButNotUsed` is removed from all
configuration files and the hint is promoted to an error when
`nimStrictMode` is enabled.

### DIFF
--- a/compiler/sem/nilcheck.nim
+++ b/compiler/sem/nilcheck.nim
@@ -286,7 +286,7 @@ proc history(map: NilMap, index: ExprIndex): seq[SemNilHistory] =
 
 
 proc symbol(n: PNode): Symbol
-func `$`(map: NilMap): string
+func `$`(map: NilMap): string {.used.}
 proc reverseDirect(map: NilMap): NilMap
 proc checkBranch(n: PNode, ctx: NilCheckerContext, map: NilMap): Check
 proc hasUnstructuredControlFlowJump(n: PNode): bool
@@ -311,6 +311,7 @@ proc symbol(n: PNode): Symbol =
   # echo "symbol ", n, " ", n.kind, " ", result.int
 
 func `$`(map: NilMap): string =
+  # intended for debugging
   var now = map
   var stack: seq[NilMap] = @[]
   while not now.isNil:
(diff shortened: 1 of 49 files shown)


compiler/sem/nilcheck.nim
```diff
@@ -358,7 +351,8 @@ proc namedSetsDebugInfo(ctx: NilCheckerContext, map: NilMap): string =
     result.add("} ")
   result.add("\n")
 
-proc namedMapAndSetsDebugInfo(ctx: NilCheckerContext, map: NilMap): string =
+proc namedMapAndSetsDebugInfo(ctx: NilCheckerContext,
+                              map: NilMap): string {.used.} =
```
saem: Non-blocker: should debug routines be marked as `{.deprecated: "do not use in final build".}`, similar to `astrepr`?

saem (approved):
One non-blocking item, otherwise this is good to merge. It'll be nice having tools take on this workload.

compiler/sem/nilcheck.nim
zerbina: Hm, yeah, that sounds like a good idea -- I'll change it.

PR #1004, opened 2023-11-02
fix wrong C code for `addr` of `ref` conversions

zerbina:
## Summary

Fix taking the address of a conversion of a `ref` (either explicitly
through the use of `addr` or implicitly as part of iterator inlining)
resulting in illegal C code being generated in some cases.

## Details

Emitting `&x->Sup` for up-conversions of `ref` values is wrong when the
expression needs to have its address taken, so the code generator
emits a cast of the `&x` instead.

Whether to the use this form was based on the operand expression, not
on how the resulting expression is used, and since the used analysis
(`compat.isLvalue`) is itself flawed, the wrong from was sometimes
used (e.g., `let` globals and `lent` dereferences).

### Fixing the issue

What code to generated for `ref` conversion is now decided based on
whether the expression is required to be an l-value, with the callsite
communicating this information this to `upConv` and `downConv` via the
new `lfWantsLvalue` flag.

The places where the flag needs to be set is where an expression has
its address taken directly or is used as the destination of an
assignment.

### Follow-on issues

The now-proper handling surfaced two pre-existing issues:
* `cgmeth.genConv` generated `(ObjUpConv (HiddenAddr x))` AST instead
  of the correct `(HiddenAddr (ObjUpConv x))` AST
* `cgen.downConv` treated generic object instance conversions as ref-
  conversion due to missing a `skipTypes` call

Both pre-existed issues are also fixed.

### Misc

The C code generated for up/down ref conversions is slightly improved:
`addrLoc` is now used for the operand and the destination loc is marked
as indirect. This prevents a redundant deref/take-address pair in some
cases.

In addition, discarding redundant up-conversion are now already skipped
in `mirgen`.

### DIFF
--- a/tests/misc/titer_inlining_regression.nim
+++ b/tests/misc/titer_inlining_regression.nim
@@ -1,0 +1,30 @@
+discard """
+  targets: "c js vm"
+  description: '''
+    Regression test for a C code generator bug triggered indirectly by
+    iterator inlining
+  '''
+"""
+
+# it's important for `Base` to be both a generic ref type
+type
+  Base[T] = ref object of RootObj
+  Sub     = ref object of Base[int]
+
+iterator iter(x: Base[int]): int =
+  # the iterator's body doesn't matter
+  discard
+
+proc get(x: Sub): lent Sub =
+  # a procedure that returns the borrowed parameter
+  x
+
+proc test() =
+  let v = Sub()
+  # an implicit up-conversion is inserted by the compiler for the iterator
+  # call's argument, which combined with the value coming from dereferencing
+  # a `lent` view resulted in a C code generator error
+  for it in iter(get(v)):
+    discard
+
+test()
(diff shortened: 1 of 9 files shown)


zerbina:
That should do. The starting implementation wasn't thorough enough (not propagating the flag through conversions, ignoring assignments) and there were also two pre-existing issues that surfaced with the more proper implementation.

tests/misc/titer_inlining_regression.nim
```diff
@@ -0,0 +1,30 @@
+discard """
+  targets: "c js vm"
+  description: '''
+    Regression test for a C code generator bug triggered indirectly by
+    iterator inlining
+  '''
+"""
+
+# it's important for `Base` to be both a generic ref type
```
saem:
```suggestion
# it's important for `Base` to be both a generic and a ref type
```

very soft suggestion

saem (approved):
cool, this should unblock @alaviss.

PR #1005, opened 2023-11-03
sem: unify the post-match expression processing

zerbina:
## Summary

Use the same post-match expression processing for all receiver contexts
and fix statement-list expressions ending in an empty container (e.g.,
`@[]`) crashing the compiler.

## Details

The expression AST produced by `paramTypesMatchAux` (which is the
procedure ultimately called for arguments during overload resolution
and by `fitNode`) needs to be post-processed (implicit conversions
eagerly evaluated, empty container types fixed, etc.).

This processing was implemented in two places, `fixAbstractType` (call
arguments and assignments) and `fitNodePostMatch` match (all other
receiver contexts), with both implementing the processing differently
and thus producing different AST for the same input.

So that the AST in receiver contexts is consistently processed and for
there to be a single implementations, the processing is unified, moved
into `fitNodePostMatch`, and improved.

### Reworked implementation

1. the processing is changed to a case statement, making the logic much
   easier to follow. Raw `nkTupleConstr` and `nkCurly` ASTs not wrapped
   in an implicit conversion are no longer processed
2. the argument of an inserted hidden call conversion is now also
   processed. This step was missing from both previous
   implementations
3. applying the conversion is moved into a separate procedure. The
   implementation is mostly the same as that of `skipHiddenSubConv` and
   `takeType`, but with to-tuple and to-set conversions always
   applied (this is inherited from `fixAbstractType`)
4. for statement-list expression, the conversion is applied to the
   trailing expression and the resulting type then propagated back to
   the statement-list expressions. This fixes inconsistent typing and
   statement-list expression ending in empty container values later
   crashing the compiler (due to them containing `tyEmpty`) 
5. `nkHiddenStdConv` are, if possible, also applied directly in
   `fitNodePostMatch`

Effects on typed AST:
* the type of `nkStmtListExpr` is the same as that of its last element
  after fitting (both would previously differ when a conversion was
  applied)
* implicit conversions are consistently collapsed
* after the post-match fitting, only expressions part of compound
  expressions (e.g., `if a: [] else: []`) use empty container types
  (e.g., `(Set Empty)`, `(Seq Empty`), which is something that should
  to be changed in future work

### Subsequent simplifications/changes

* remove the empty container type fixup in `transf.transformConv` (via
  `takeType`); the fixup always takes place during semantic analysis
  now
* remove `takeType`, it was fully integrated into `applyConversion`
* remove the wrong `checkSonsLen` call in `fixVarArgumentsAndAnalyse`.
  Error nodes may reach there, and since `nkError` doesn't have a
  `sons` field, this led to `FieldDefect` errors
* renamed `fixAbstractType` to `fitArgTypesPostMatch` to better
  reflect what the procedure does and when it should be used, while
  keeping it inline with the naming of `fitNodePostMatch`

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -822,11 +822,10 @@ proc semArrayConstr(c: PContext, n: PNode, flags: TExprFlags): PNode =
     if hasError:
       result = c.config.wrapError(result)
 
-proc fixAbstractType(c: PContext, n: PNode): PNode =
-  ## Takes the production AST of a call and performs the post-match node
-  ## fitting on it. Errors are passed through.
-  # XXX: the procedure's name doesn't make much sense anymore...
-  addInNimDebugUtils(c.config, "fixAbstractType", n, result)
+proc fitArgTypesPostMatch(c: PContext, n: PNode): PNode =
+  ## Takes the production AST of a call and performs the post-match type
+  ## fitting on the argument nodes. Errors are passed through.
+  addInNimDebugUtils(c.config, "fitArgTypesPostMatch", n, result)
   result = n # `n` is already a production
 
   case n.kind
@@ -1185,7 +1184,7 @@ proc afterCallActions(c: PContext; n: PNode, flags: TExprFlags): PNode =
   else:
     semFinishOperands(c, result)
     activate(c, result)
-    result = fixAbstractType(c, result)
+    result = fitArgTypesPostMatch(c, result)
     result = fixVarArgumentsAndAnalyse(c, result)
     if callee.magic != mNone:
       result = magicsAfterOverloadResolution(c, result, flags)
@@ -1395,7 +1394,7 @@ proc semIndirectOp(c: PContext, n: PNode, flags: TExprFlags): PNode =
         else:
           afterCallActions(c, result, flags)
     else:
-      result = fixAbstractType(c, result)
+      result = fitArgTypesPostMatch(c, result)
       result = fixVarArgumentsAndAnalyse(c, result)
   else:
     discard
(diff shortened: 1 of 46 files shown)


zerbina:
Huh, the test seems to have uncovered a rather severe `jsgen` bug. I'll make a separate PR with a fix.

Apart from that, the PR is now ready for review.

zerbina:
I'll split the `sigmatch` change out into a separate PR.

compiler/sem/semexprs.nim
```diff
@@ -822,38 +822,27 @@ proc semArrayConstr(c: PContext, n: PNode, flags: TExprFlags): PNode =
     if hasError:
       result = c.config.wrapError(result)
 
-proc isArrayConstr(n: PNode): bool {.inline.} =
-  n.kind == nkBracket and n.typ.skipTypes(abstractInst).kind == tyArray
-
 proc fixAbstractType(c: PContext, n: PNode): PNode =
```
```suggestion
proc postMatchCallFitArgTypes(c: PContext, n: PNode): PNode =
```

Not required for this PR, just thought of it given the comment.

saem (approved):
My one comment is entirely non-blocking. The change itself brings more consistency, which is good.

compiler/sem/semexprs.nim
zerbina: Thank you for the suggestion. For symmetry with `fitNodePostMatch`, I've used `fitArgTypesPostMatch`.

saem: The new name is much better. 🎉

saem:
To avoid clobbering any potential edits to the PR body, I'll leave the suggestion here. I'd add the following bulletpoint, or something like it, to the end of the commit message part of the PR:

* Renamed `fixAbstractType` to `fitArgTypesPostMatch` to better reflect what the procedure does and when it should be used, while keeping it inline with the naming of `fitNodePostMatch`

zerbina:
Good call, thank you for the reminder -- I would've forgotten about updating the PR message.

PR #1006, opened 2023-11-03
fix(sem): template/macro issues with `var`/`lent`

### LINKED ISSUE #328: Wrong behaviour at runtime with templates and `var`/`lent` tuples

zerbina:
## Summary

* fix the result of `lent` returning macros and templates being
  assignable when the final body of the template/macro ends in a
  mutable lvalue expression
* fix wrong code being generated for templates and macros that had a
  concrete tuple return type and where the body was of `var`/`lent`
  tuple type (all backends were affected)

In terms of how the result can be used, macros and templates with
`var`/`lent` return types now behave the same as the other routines do.

Fixes https://github.com/nim-works/nimskull/issues/328.

## Details

The problem was that the expansion of a template/macro using a concrete
type as the return type wasn't typed correctly:
* `semExpr` doesn't insert a hidden deref for `var`/`lent`-typed
  expressions, leading to a `var`/`lent`-typed expression being passed
  to `fitNode`
* `semExpr` doesn't enforce that the expression is not `void`-typed
  (i.e., a statement)
* `fitNode` alone doesn't handle `var` and `lent` target types
  correctly

This led to semantics inconsistent with `var`/`lent`-returning
procedures and code generation problems.

For fixing the issues:
* `semExpr` is replaced with `semExprWithType`
* `fitNode` is replaced with `fitNodeConsiderViewType`

Two templates in the `results` module can now use concrete types as the
return type again.

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -702,13 +702,13 @@ proc semAfterMacroCall(c: PContext, call, macroResult: PNode,
         retType = generateTypeInstance(c, paramTypes,
                                        macroResult.info, retType)
 
-      result = semExpr(c, result, flags)
+      result = semExprWithType(c, result, flags)
       if resultTypeIsInferrable(retType):
         # this is a "return type inference" scenario. There's no return type
         # to infer, but the expression still needs to use the proper type
         result = inferWithMetatype(c, retType, result)
       else:
-        result = fitNode(c, retType, result, result.info)
+        result = fitNodeConsiderViewType(c, retType, result, result.info)
   dec(c.config.evalTemplateCounter)
   discard c.friendModules.pop()
 
--- a/lib/experimental/results.nim
+++ b/lib/experimental/results.nim
@@ -200,11 +200,8 @@ func `$`*[E](self: Result[void, E]): string =
   if self.isOk: "Ok()"
   else: "Err(" & $self.e & ")"
 
-# XXX: the `value` templates only use `untyped` in order to work around a
-#      compiler bug (https://github.com/nim-works/nimskull/issues/328).
-#      Once the bug is fixed, `T` should be used as the return type again
-template value*[T, E](self: Result[T, E]): untyped = get self
-template value*[T, E](self: var Result[T, E]): untyped = get self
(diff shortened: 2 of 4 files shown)


saem (approved):
Nice, the consolidations are making it much easier to make big consistency improvements like this.

PR #1007, opened 2023-11-03
front/nimconf: add exported version of readConfigFile

alaviss:
## Summary
This API is used by [nimph](https://github.com/disruptek/nimph) to read
configuration files for modification.

## Details
* This is a part of the effort to get nimph running with nimskull

### DIFF
--- a/compiler/front/nimconf.nim
+++ b/compiler/front/nimconf.nim
@@ -107,7 +107,6 @@ type
   CancelConfigProcessing = object of CatchableError
     ## internal error used to halt processing
 
-
 # ---------------- configuration file parser -----------------------------
 # we use Nim's lexer here to save space and work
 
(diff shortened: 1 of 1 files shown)


compiler/front/nimconf.nim
```diff
@@ -502,4 +501,10 @@ proc loadConfigs*(
                              stopOnError: stopOnError)
   parser.loadConfigs(cfg, cache)
 
-  
+proc readConfigFile*(
+  filename: AbsoluteFile, cache: IdentCache,
+  conf: ConfigRef, evtHandler: NimConfEvtWriter
+): bool {.inline.} =
```
saem, line 507:
```suggestion
): bool {.inline.} =
  # created and exported for `nimph`
```

saem (commented):
Minor suggestion nonblocking

disruptek:
Probably a good stop-gap measure, but ideally, we expose everything we need via json and reduce that list where possible.

alaviss:
> Probably a good stop-gap measure, but ideally, we expose everything we need via json and reduce that list where possible.

How would that work? Like adding a `parsecfg` command or something?

PR #1008, opened 2023-11-03
lib/parseopt: require explicit parameters list

alaviss:
Now that  `getopt`  no longer automatically pulls in command line
arguments, it makes no sense to keep the default "empty" parameter.

Included is an update to documentation about the current behavior and a
fix to nimgrep to be able to read command line parameters.

Fixes 1a72373f3f2a58709a7f87c3374a107565bf1338

### DIFF
--- a/lib/pure/parseopt.nim
+++ b/lib/pure/parseopt.nim
@@ -405,14 +405,12 @@ iterator getopt*(p: var OptParser): tuple[kind: CmdLineKind, key,
     if p.kind == cmdEnd: break
     yield (p.kind, p.key, p.val)
 
-iterator getopt*(cmdline: seq[string] = @[],
-                  shortNoVal: set[char] = {}, longNoVal: seq[string] = @[]):
+iterator getopt*(cmdline: seq[string],
+                 shortNoVal: set[char] = {}, longNoVal: seq[string] = @[]):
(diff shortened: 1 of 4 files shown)


lib/pure/parseopt.nim
```diff
@@ -405,14 +405,12 @@ iterator getopt*(p: var OptParser): tuple[kind: CmdLineKind, key,
...
-iterator getopt*(cmdline: seq[string] = @[],
-                  shortNoVal: set[char] = {}, longNoVal: seq[string] = @[]):
+iterator getopt*(cmdline: seq[string],
+                 shortNoVal: set[char] = {}, longNoVal: seq[string] = @[]):
             tuple[kind: CmdLineKind, key, val: string] =
   ## Convenience iterator for iterating over command line arguments.
   ##
-  ## This creates a new `OptParser<#OptParser>`_. If no command line
-  ## arguments are provided, the real command line as provided by the
-  ## `os` module is retrieved instead.
```
saem, line 415: The retrieval still happens, as that code doesn't seem to have been changed?

alaviss: That already happened due to https://github.com/nim-works/nimskull/commit/1a72373f3f2a58709a7f87c3374a107565bf1338, I'm just fixing the docs to reflect that.

PR #1009, opened 2023-11-04
fix(js): wrong code for `discard` of address-like

zerbina:
## Summary

Fix incorrect JavaScript code being generated for `discard`s of
`pointer` values, `ptr`-to-scalar values, or address-of field/array-
element expressions.

## Details

* emit both the `address` part and `res` part for address expression
  (and not only the index)
* use `lineF` directly, slightly reducing the amount of string
  allocations
* replace using `var _ = ...` with using parenthesis when turning the
  expression into JavaScript statements; it's less code to emit

### DIFF
--- a/compiler/backend/jsgen.nim
+++ b/compiler/backend/jsgen.nim
@@ -2522,8 +2522,11 @@ proc gen(p: PProc, n: CgNode, r: var TCompRes) =
   of cnkFastAsgn: genFastAsgn(p, n)
   of cnkVoidStmt:
     genLineDir(p, n)
-    gen(p, n[0], r)
-    r.res = "var _ = " & r.res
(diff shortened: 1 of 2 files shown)


zerbina:
This PR is currently blocked by yet another `jsgen` issue: the code generator handles locals (which includes the `result` variable) of `pointer` type incorrectly.

saem (approved):

zerbina:
Despite having immediately looked at the implement of `jsgen.rdLoc` prior to using it here, I still somehow mixed up how it works with `cgen.rdLoc`. Anyhow, the problem is fixed properly now, and I've also changed the generated code to use parenthesis instead of the `var _ = ...`.

PR #1013, opened 2023-11-04
fix a destruction bug for global-backed locals

### LINKED ISSUE #1010: destructor double-destroy due to break within loop at top-level
### Example

```nim
type
  X = object
    destroyed: bool

proc `=destroy`(x: var X) =
  doAssert not x.destroyed
  x.destroyed = true

for i in 1..2:
  if i == 2:
    break
  let c = X()
  echo c
```

### Actual Output

```
(destroyed: false)
iterators_1.nim(92)      test
test.nim(6)              =destroy
assertions.nim(38)       failedAssertImpl
assertions.nim(28)       raiseAssert
fatal.nim(48)            sysFatal
Error: unhandled exception: test.nim(6, 3) `not x.destroyed`  [AssertionDefect]
```

### Expected Output

```
(destroyed: false)
```

### Additional Information

If the block is in a `main()` proc, the double-free does not occur.

Some observation:

- The `c` local was not reset on loop start when the code is at top-level, but was reset when the code is in a proc.
- The variable `c` was promoted to global, even though it is bound by the loop's scope.

### References

* Found while testing `nimph`.

zerbina:
## Summary

Fix stale state in some cases reaching destructors for locals defined
in loops that are themselves not part of a routine.

Fixes https://github.com/nim-works/nimskull/issues/1010.

## Details

Locations defined within loops outside of routines are globals
internally. Globals so far didn't use the default-initialized-at-scope-
entry behaviour that routine-level locals use, which caused their
destructors observing them in a not-reset state in certain situations.

For ensuring proper default-initialization:
1. the pass for removing 'def's of nested globals is removed
2. patching the symbols of globals after the `injectdestructors` pass
   has run still needs to happen after the, so the patching logic is
   duplicated into a dedicated pass 
4. removing the 'def's of globals is performed during the MIR ->
   `CgNode` translation. This allows for `cgirgen` to emit a default-
   initialization for globals that need it

The facility already used for locals (i.e. moving the definition when
it's not placed immediately within a MIR scope) is also used for
globals.

This does **not** fix the fundamental issue of cleanup also being
applied to not-yet-alive globals, but it makes the behaviour between
globals and locals consistent.

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -348,15 +348,11 @@ proc preprocess*(conf: BackendConfig, prc: PSym, graph: ModuleGraph,
 
 proc process*(prc: var Procedure, graph: ModuleGraph, idgen: IdGenerator) =
   ## Applies all applicable MIR passes to the procedure `prc`.
-  rewriteGlobalDefs(prc.body.tree, prc.body.source, outermost = true)
+  rewriteGlobalDefs(prc.body.tree, prc.body.source)
 
   if shouldInjectDestructorCalls(prc.sym):
     injectDestructorCalls(graph, idgen, prc.sym,
                           prc.body.tree, prc.body.source)
-
-  rewriteGlobalDefs(prc.body.tree, prc.body.source, outermost = false)
-  # XXX: ^^ this is a hack. See the documentation of the routine for more
-  #      details
 
   let target =
     case graph.config.backend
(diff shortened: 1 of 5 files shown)


zerbina:
For fixing the mentioned fundamental issue, more progress towards improving the MIR has to be made first. The rework from #942 is the first step towards resolving this.

zerbina:
Okay, the test failures are fixed now. A positive side-effect of this PR is that I had to look into imported globals again, which made their meaning and how they should behave much clearer.

saem (approved):

PR #1017, opened 2023-11-05
evaltempl: unwrap default static[T] parameters

### LINKED ISSUE #1015: semfold unreachable error trying to convert static[bool] to int
### Example

Minimized from `lib/pure/terminal.nim`

```nim
proc takeInt(i: int) = discard

template x(b: static[bool] = false): untyped =
  takeInt b.int

x()
```

### Actual Output

```
nim.nim(154)             nim
main.nim(524)            mainCommand
main.nim(214)            commandCompileToC
modules.nim(212)         compileProject
modules.nim(128)         compileModule
passes.nim(275)          processModule
passes.nim(102)          processTopLevelStmt
sem.nim(905)             myProcess
sem.nim(835)             semStmtAndGenerateGenerics nkStmtList 371800 /home/leorize/documents/sources/nimskull/test.nim(6, 1)
semfold.nim(964)         foldInAst
semfold.nim(921)         foldInAstAux
semfold.nim(852)         foldConstExprAux
semfold.nim(911)         foldConstExprAux
semfold.nim(801)         getConstExpr
semfold.nim(504)         foldConv
idioms.nim(42)           unreachableImpl
idioms.nim(36)           unreachableImpl
assertions.nim(28)       raiseAssert
fatal.nim(48)            sysFatal
Error: unhandled exception: semfold.nim(504, 50) unreachable: tyStatic [AssertionDefect]
```

## Additional information

* Blocking [disruptek/balls](https://github.com/disruptek/balls)
* This happens with *any* enum value, not just `bool`

alaviss:
## Summary
Fixed a case where  `static[T]`  values leaked from templates to backend
and other parts of sem causing compiler crashes.

## Details
Most sem code expects  `tyStatic`  to have been resolved by the time it
hits them. But it seems that this was not processed for template
parameters, so add some code to handle it there.

Fixes https://github.com/nim-works/nimskull/issues/1015

### DIFF
--- a/tests/lang_callable/template/template_issues.nim
+++ b/tests/lang_callable/template/template_issues.nim
@@ -296,3 +296,11 @@ block: # bug #12595
     discard {i: ""}
 
   test()
+
+block: # https://github.com/nim-works/nimskull/issues/1015
+  proc takeInt(i: int) = discard
+
+  template x(b: static[bool] = false): untyped =
+    takeInt b.int
+
+  x()
--- a/tests/lang_callable/template/template_issues.nim
+++ b/tests/lang_callable/template/template_issues.nim
@@ -297,10 +297,10 @@ block: # bug #12595
 
   test()
 
-block: # https://github.com/nim-works/nimskull/issues/1015
-  proc takeInt(i: int) = discard
+block static_parameter_with_default:
+  # using a `static` template parameter as the argument to a conversion
+  # crashed the compiler
+  template x(b: static[bool] = false) =
+    discard int(b)
 
-  template x(b: static[bool] = false): untyped =
-    takeInt b.int
-
-  x()
+  x() # the default value for the parameter must be used
--- a/compiler/sem/evaltempl.nim
+++ b/compiler/sem/evaltempl.nim
@@ -17,7 +17,8 @@ import
     idents,
     renderer,
     errorhandling,
-    errorreporting
+    errorreporting,
+    types
   ],
   compiler/front/[
     options,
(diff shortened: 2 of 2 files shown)


tests/lang_callable/template/template_issues.nim
```diff
@@ -296,3 +296,11 @@ block: # bug #12595
...
 
   test()
+
+block: # https://github.com/nim-works/nimskull/issues/1015
+  proc takeInt(i: int) = discard
+
+  template x(b: static[bool] = false): untyped =
+    takeInt b.int
+
+  x()
```
zerbina:
```suggestion
block static_parameter_with_default:
  # using a `static` template parameter as the argument to a conversion
  # crashed the compiler
  template x(b: static[bool] = false) =
    discard int(b)

  x() # the default value for the parameter must be used
```
The intention is to only use the language features that are necessary and to document highlight the important parts, making navigating the test easier.

zerbina (commented):
The change looks good to me. I left a suggestion about how to improve the test case and bring it in line with the current style.

Could you reword title and PR message to focus more on what works now? The current body should go into the details section.

zerbina:
The issue is a bit hard to fit into the title, and I think the most recent one is fine. For the summary, I think it should include a mention that crashes and other misbehaviour were the result of `tyStatic` reaching where it shouldn't.

alaviss:
> The issue is a bit hard to fit into the title, and I think the most recent one is fine. For the summary, I think it should include a mention that crashes and other misbehaviour were the result of `tyStatic` reaching where it shouldn't.

Hopefully this works

zerbina (approved):
Yep, it does. Thank you!

PR #1018, opened 2023-11-05
fix(sem): error with `var`/`lent`-returning templates

zerbina:
## Summary

Fix using templates and macros with a `var T` or `lent T` return type
leading to compiler crashes or wrong code being generated. Both the C
and the VM backend were affected.

## Details

When the receiving type is a `var` or `lent` type, `sigmatch` produced
implicit conversions to said type. These are wrong at the type-level,
as a conversion cannot turn a non-view value into a view. Due a
separate issue with `semAfterMacroCall`, this led to semantically
invalid expressions being produced after expansion of the
aforementioned templates.

The `var` and `lent` modifiers are now dropped from the produced
conversions, both fixing the concrete problem with templates and
allowing for the removal of several workarounds in the compiler. 

Reporting a warning for narrowing object conversions where information
is lost relied on the `var` modifier in conversion. The test is
replaced with looking for the problematic conversion in receiver
contexts (except for `discard`, as the value isn't used there).

Finally, a general test is added for ensuring that templates returning
`var` or `lent` types behave the same as a normal procedure would. They
currently don't, but most of the test at least compiles and runs now.

### DIFF
--- a/compiler/backend/cgirgen.nim
+++ b/compiler/backend/cgirgen.nim
@@ -691,22 +691,13 @@ proc wrapInHiddenAddr(cl: TranslateCl, n: CgNode): CgNode =
   ## Restores the ``cnkHiddenAddr`` around lvalue expressions passed to ``var``
   ## parameters. The code-generators operating on ``CgNode``-IR depend on the
   ## hidden addr to be present
-  let inner =
-    if n.kind == cnkStmtListExpr: n[^1] else: n
-
-  result =
-    if n.typ.skipTypes(abstractInst).kind != tyVar:
-      newOp(cnkHiddenAddr, n.info, makeVarType(cl.owner, n.typ, cl.idgen), n)
-    elif inner.kind == cnkObjUpConv and
-         inner.operand.typ.kind != tyVar:
-      # TODO: ``nkHiddenSubConv`` nodes for objects (which later result
-      #       in ``cnkObjUpConv`` nodes) are in some cases
-      #       incorrectly typed as ``var`` somewhere in the compiler
-      #       (presumably during sem). Fix the underlying problem and remove
-      #       the special case here
-      newOp(cnkHiddenAddr, n.info, n.typ, n)
-    else:
-      n
+  if n.typ.skipTypes(abstractInst).kind != tyVar:
+    newOp(cnkHiddenAddr, n.info, makeVarType(cl.owner, n.typ, cl.idgen), n)
+  else:
+    # XXX: is this case ever reached? It should not be. Raw ``var`` values
+    #      must never be passed directly to ``var`` parameters at the MIR
+    #      level
+    n
 
 proc genObjConv(n: CgNode, a, b, t: PType): CgNode =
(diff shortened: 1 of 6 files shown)


compiler/backend/cgirgen.nim
```diff
@@ -691,22 +691,13 @@ proc wrapInHiddenAddr(cl: TranslateCl, n: CgNode): CgNode =
...
-      #       the special case here
-      newOp(cnkHiddenAddr, n.info, n.typ, n)
-    else:
-      n
+  if n.typ.skipTypes(abstractInst).kind != tyVar:
+    newOp(cnkHiddenAddr, n.info, makeVarType(cl.owner, n.typ, cl.idgen), n)
+  else:
+    # XXX: is this case ever reached? It should not be. Raw ``var`` values
+    #      must never be passed directly to ``var`` parameters at the MIR
+    #      level
```
saem, line 699: I'm guessing you didn't feel comfortable making this an `unreachable` just yet?

tests/lang_callable/varres/ttemplate_var_lent_res.nim
```diff
@@ -0,0 +1,47 @@
...
+
+  # note: `v` being an unnamed tuple is deliberate, as it is
+  # intended to ensure that the implicit conversion in the template
+  # doesn't cause problems
+  var v: (int,) = (0,)
+  # test full assignment and read access:
+  borrowVar(v) = (x: 1)
+  doAssert borrowVar(v) == (x: 1)
+  # test with field write and read:
+  borrowVar(v).x = 2
```
saem, line 37: mostly a note for myself, but having this work by clobbering the type is a bit mind bending and possibly dangerous.

saem (approved):
one non-blocking question, otherwise good to go.

compiler/backend/cgirgen.nim
zerbina: Yep, I believe that there are still some cases where this is violated that the test suite doesn't catch.

tests/lang_callable/varres/ttemplate_var_lent_res.nim
zerbina: What do you mean by "clobbering the type"?

saem: If i'm following the expansion correctly, `borrowVar(v)` is just `v`. `v`'s type is `var (,int)`, but by way of the template it's `var tuple`, which due to assignment(?) is treated as `var (, x: int)`, allowing for `borrowVar(v).x`, even though `v` originally didn't have an `x` field.

zerbina: Ah, that you mean. Unnamed tuples are currently implicitly convertible to named ones that have the same structure, but the type of `v` stays the same. The expanded code essentially looks like this:
```nim
Tuple(v).x = 2
```

Relying on this behaviour is not necessary for the test, so I've now changed `v` to be of type `Tuple`.

zerbina:

PR #1019, opened 2023-11-05
fix(sem): crash on missing explicit initial value

### LINKED ISSUE #1016: SIGSEGV while sem-ing an uninitialized requiresInit variable
### Example

```nim
type X {.requiresInit.} = object

var y: X
```

### Actual Output

```
nim.nim(154)             nim
main.nim(524)            mainCommand
main.nim(214)            commandCompileToC
modules.nim(212)         compileProject
modules.nim(128)         compileModule
passes.nim(275)          processModule
passes.nim(102)          processTopLevelStmt
sem.nim(905)             myProcess
sem.nim(823)             semStmtAndGenerateGenerics nkStmtList 371735 test.nim(1, 5)
semstmts.nim(3404)       semStmt
semexprs.nim(1451)       semExprNoType
semexprs.nim(3830)       semExpr nkStmtList 371735 test.nim(1, 5)
semstmts.nim(3300)       semStmtList nkStmtList 371735 test.nim(1, 5)
semexprs.nim(3834)       semExpr nkVarSection 371736 test.nim(3, 0)
semstmts.nim(1352)       semConstLetOrVar nkVarSection 371736 test.nim(3, 0)
semstmts.nim(931)        semNormalizedLetOrVar
```

### Expected Output

Some kind of errors about not declaring the value of `y`, or no error because `y` is never used.

zerbina:
## Summary

Fix the compiler crashing on `var x: Y` statements where `Y` is a
`.requiresInit`-using object that does not have any fields requiring
initialization.

Fixes https://github.com/nim-works/nimskull/issues/1016.

## Details

* add a dedicated error diagnostic and report for diagnosing
  defining a `.requiresInit` object variable without providing an
  initial value
* for consistency, a message similar to that of the error regarding
  `distinct` types is used
* handle the "`.requiresInit` object with no fields requiring
  initialization" case in `defaultConstructionError`. `nil` was
  returned previously in that case, leading to an NPE
* the procedure is changed to not use `result`, preventing similar
  errors in the future (missing `result` assignment)

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1155,7 +1155,7 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       result = "Array length can't be negative, but was " & $r.countMismatch.got
 
     of rsemObjectDoesNotHaveDefaultValue:
-      result = "The type '$1' doesn't have a default value" % r.typ.render
+      result = "The type '$1' requires an initial value" % r.typ.render
 
     of rsemDistinctDoesNotHaveDefaultValue:
       result = "The $1 distinct type doesn't have a default value." % r.typ.render
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1154,6 +1154,9 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemArrayExpectsPositiveRange:
       result = "Array length can't be negative, but was " & $r.countMismatch.got
 
+    of rsemObjectDoesNotHaveDefaultValue:
+      result = "The type '$1' doesn't have a default value" % r.typ.render
+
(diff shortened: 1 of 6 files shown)


compiler/front/cli_reporter.nim
```diff
@@ -1154,6 +1154,9 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemArrayExpectsPositiveRange:
       result = "Array length can't be negative, but was " & $r.countMismatch.got
 
+    of rsemObjectDoesNotHaveDefaultValue:
+      result = "The type '$1' doesn't have a default value" % r.typ.render
```
saem:
```suggestion
      result = "The type '$1' requires an initial value" % r.typ.render
```

saem (approved):
Non-blocking suggestion, I can change the messages in one fell swoop later.

PR #1021, opened 2023-11-06
sem: consolidate macro pragma processing

zerbina:
## Summary

Merge the separate implementations of macro/template pragma processing
into a single one. The benefits:
* less duplicated code
* macro and template annotations are ensured to work the same for all
  types of definitions
* a single place to change the behaviour in

This changes:
* routine AST passed to macro/template pragmas keeps the `nnkPragma`
  node, even if empty
* ill-formed accent quoted expressions preceding a macro/template
  pragma don't prevent the transform from taking place

## Details

Routines, types, and var/let/const definitions all had their own
implementation, but their only difference was in where the AST was
modified and whether a full copy was created or not.

There's nothing that requires fundamentally different processing for
each type of definition, so the processing is consolidated:
* the new `tryMacroPragma` procedure processes a single pragma,
  returning either the raw transform or nil
* the new `semAnnotation` processes a pragma list (`nkPragma`), trying
  each item with `tryMacroPragma`
* sem'ing of routines, types, and var/let/const definitions use
  `semAnnotation`

So that the macro/template pragma call can be temporarily removed from
the definition's pragma list (which is at different tree positions for
each definition AST), the address of the node is passed to
`semAnnotation`.

Differences compared to the previous implementation:
* empty `nkPragma` nodes are kept. This was previously only the case
  for var/let/const annotations
* instead of the manual `semMacroExpr` call, `afterCallActions` is
  used. This is a step towards removing the bare `semMacroExpr` calls

Ideally, `semAnnotation` would always return the raw transformation
result and leave sem'ing it to the callsite, but this is not yet
possible.

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -480,6 +480,90 @@ proc copyExcept(n: PNode, i: int): PNode {.inline.} =
   for j in 0..<n.len:
     if j != i: result.add(n[j])
 
+proc tryMacroPragma(c: PContext, pragmas: ptr PNode, i: int,
+                    operand: PNode, flags: TExprFlags): PNode =
+  ## If the `i`-th pragma in the `pragmas` list is a macro/template pragma,
+  ## applies it to the untyped `operand` AST and returns the transform (or an
+  ## error). `nil` is returned otherwise.
+  ##
+  ## `pragmas` needs to point to the ``nkPragma`` AST within the `operand`
+  ## AST. `flags` are the flags passed to the after-call expansion.
+  ##
+  ## Example:
+  ##
+  ## ..code-block:: nim
+  ##
+  ##   let l {.m(a, b), rest.} # the operand AST
+  ##   # is transformed into
+  ##   m(a, b, let l {.rest.})
+  ##
+  let n = pragmas[]
+  assert n.kind == nkPragma
+  let
+    prag         = n[i]
+    isCallSyntax = prag.kind in nkPragmaCallKinds and prag.len >= 1
+    key =
+      if isCallSyntax: prag[0]
+      else:            prag
+
+  # we only want to process macro pragmas (ast transforms)
+  # xxx: @saem: this dance to filter through them is another sign that pragmas
+  #      shouldn't be associated to symbols as the syntax hints
+  if whichPragma(prag) == wInvalid and key.kind in nkIdentKinds:
+    # a custom pragma as opposed to a built-in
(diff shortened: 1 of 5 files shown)


compiler/sem/semstmts.nim
```diff
@@ -480,6 +480,90 @@ proc copyExcept(n: PNode, i: int): PNode {.inline.} =
...
+
+  # we only want to process macro pragmas (ast transforms)
+  # xxx: @saem: this dance to filter through them is another sign that pragmas
+  #      shouldn't be associated to symbols as the syntax hints
+  if whichPragma(prag) == wInvalid and key.kind in nkIdentKinds:
+    # a custom pragma as opposed to a built-in
+    let (ident, err) = considerQuotedIdent(c, key)
+    if err != nil:
+      # don't report the error yet, a macro/template pragma could
+      # still make the AST well-formed
```
saem: This would then be an ordering violation? where any future macro pragma(s) that make it valid, if it exists, should have come first?

The biggest drawback I can think of is a spelling mistake in one pragma might go unreported because a second pragma clobbered it, leading to unexpected semantics.

The benefit of reporting the error, is that we can assume all pragma prior to the current `i` to have been fully applied, simplifying the concept of progress.

saem (commented):
The non-reporting of identifier errors might be quite problematic, see inline comment for more details. Happy to discuss it further, in case I'm misunderstanding something.

compiler/sem/semstmts.nim
zerbina: I'm not sure I understand the question with the ordering violation, but with the spelling mistake part I do somewhat agree.

Other invalid AST in pragma lists is already ignored during the early annotation processing, so my thinking was that making an exception for invalid accent quotations is unexpected/inconsistent.

zerbina: Your edit helped, thanks. Hm, yeah, I see what you mean. The perspective I made the change from is that macro pragmas take precedence over anything else, operating on the unprocessed definition AST.

saem: Sorry it wasn't clear from the get go, I was more tired than I realized and ended up with some word salad. Glad to hear the edit helped clarify things.

saem (approved):

saem:
/cancel

Just realized the PR body wasn't updated just yet.

zerbina:
> Just realized the PR body wasn't updated just yet.

Yep, sorry, I got sidetracked with something else. It's updated now.

PR #1023, opened 2023-11-06
fix(sem): don't execute static calls too early

### LINKED ISSUE #1022: mutations in macros to compile-time variable is not observable via another compile-time function
### Example

```nim
var x {.compileTime.} = 10
proc foo(): int {.compileTime.} = x

macro echoFoo(): untyped =
  inc x
  inc x
  inc x
  echo foo()
  echo x

echoFoo()
```

### Actual Output

```
10
13
```

### Expected Output

```
13
13
```

### Additional information

* Found while working on [disruptek/balls](https://github.com/disruptek/balls).

zerbina:
## Summary

Fix early execution/folding of calls to compile-time-only procedures
with side-effects not being disabled within macros and `.compileTime`
procedures. It was already disabled for `static` and `const`
contexts.

Fixes https://github.com/nim-works/nimskull/issues/1022.

## Details

* use `inCompileTimeOnlyContext` for detecting whether within a
  compile-time-only context. This also considers wether the current
  routine is compile-time-only
* refactor the existing test (`tcompiletimesideeffects.nim`). The
  `echo`-based testing is replaced with `doAssert` and all compile-
  time-only contexts are tried, instead of only `const`

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -1121,7 +1121,7 @@ proc evalAtCompileTime(c: PContext, n: PNode): PNode =
 
     # only attempt to fold the expression if doing so doesn't affect
     # compile-time state
-    if ecfStatic notin c.execCon.flags or sfNoSideEffect in callee.flags:
+    if not inCompileTimeOnlyContext(c) or sfNoSideEffect in callee.flags:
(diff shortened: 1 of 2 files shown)


saem (approved):
The testing is really thorough, nice.

PR #1024, opened 2023-11-07
mir: improve the `Changeset` API

zerbina:
## Summary

Remove the cursor tracking and require the modification position to be
manually provided when recording a change. This makes the API less
stateful and also significantly reduces the time and memory cost of
creating and `Changeset`.

## Details

**Concrete changes:**
* the `seek` and `position` routine are removed
* the `pos` and `tree` field are removed
* all routines that need to perform a tree traversal now have a
  `MirTree` parameter
* all routines for recording changes now have a `NodePosition`
  parameter

The reason for the high creation overhead was that the `Changeset`
carried a full copy of the MIR tree it recorded the changes for. This
was necessary for providing automatic cursor movement, which was
intended for efficiently recording changes to nodes following
immediately after each other.

However, there was only a single use of the automatic cursor movement
functionality, and it's unlikely that this is going to change.

### DIFF
--- a/compiler/mir/mirbridge.nim
+++ b/compiler/mir/mirbridge.nim
@@ -107,18 +107,17 @@ proc rewriteGlobalDefs*(body: var MirTree, sourceMap: var SourceMap;
         let
           sym = restoreGlobal(body[def].sym)
           typ = sym.typ
-        changes.seek(i)
         if depth > 1:
           # don't rewrite the def, but still patch the symbol if requested
           if patch:
-            changes.seek(i + 1)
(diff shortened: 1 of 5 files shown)


zerbina:
For me, when bootstrapping the compiler with `-d:release`, the change resulted in a decrease in compile time of around 0.2 seconds for the third iteration. This is a reduction of approximately 1%.

That may seem small, but profiling showed `initChangeset` (because of the full copy) to be among the top routines that time was spent in during the MIR processing.

Clyybber (approved):

saem:
/merge

Sorry, accidentally pressed the close button while scrolling on my phone.

PR #1027, opened 2023-11-18
mir: store `SourceId`s directly on nodes

zerbina:
## Summary

Don't use a separate storage for `SourceId` and instead store the ID
directly in `MirNode`. This is simpler, more flexible, and it prepares
for changes to the MIR's data representation.

## Details

The `SourceId` associated with a `MirNode` was previously stored in a
separate sequence, with the connection between both happening through
the index within their storing sequences (struct-of-arrays approach).

This was intended for maximum flexibility in what data representation
the source mappings use, but this potential was not realized nor is it
likely that the ID-to-struct approach is going to change.

Upcoming changes to the MIR are going to require changing/keeping meta-
data attachments on single MIR nodes when recording changesets, which
is something not easy to support ergonomically with the decoupled
storage, so the ID is stored directly in `MirNode`.

Due to the current alignment requirements of `MirNode`, this can be
done without affecting the size of the type.

**Changes:**
* change `SourceMap` to only store `SourceId` -> `PNode` mappings and
  adjust the API accordingly
* un-export `SourceMap`'s internal fields
* move `SourceId` to the `mirtrees` module
* add the `info` field to `MirNode`
* rename the `CodeFragment` type to `MirBuffer`, which better reflects
  its purpose/usage. The approach of attaching the `SourceId` in a
  deferred manner is kept, with `MirBuffer` keeping track of what nodes
  still need to be update via the `cursor` field
* export parts of the `MirBuffer` API for use in `backends` and change
  the `MirTree`-accepting `generatedCode` overload to accept a
  `MirBuffer`
* updating the attachments is still done through the `Changeset` API,
  but it now happens directly after recording a change rather than when
  applying a changeset. This is both simpler and more efficient

Documentation is updated where necessary.

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -146,8 +146,8 @@ template add[T](x: Deque[T], elem: T) =
 func append*(dest: var MirFragment, src: sink MirFragment) =
   ## Appends the code from `src` to `dest`. No check whether the resulting
   ## code is semantically valid is performed
+  dest.source.merge(src.tree, src.source)
   dest.tree.add src.tree
-  dest.source.append src.source
 
 # ----- private and public API for ``Queue`` -----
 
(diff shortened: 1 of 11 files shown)


saem (approved):
Going through the change, I wonder if the general guidance for DOD approaches should be to keep metadata association inline with the data itself. With the exception being where it's very sparse and the churn in primary data is minimal.

PR #1028, opened 2023-11-18
fix(C): code generation issue with `finished`

zerbina:
## Summary

Fix assigning the result of a `finished` call to a local either being
treated as a no-op or resulting in C compiler errors. Only the C
backend was affected.

## Details

`genBreakState` didn't handle the put-into-destination case properly:
instead of emitting an assignment, the already initialized `d` was
re-initialized as an expression.

When the call was used at the right side of an assignment, this led to
no code being generated for the assignment (leaving the assignment
destination as is). When the call was used on the right side of a
`var`/`let` statement that wasn't rewritten into an assignment, the
C name of the local was overwritten with the expression, causing all
subsequent assignments to the local resulting in invalid C code.

The `genBreakState` procedure is moved to the `ccgexprs` module (where
the magic-related code generation logic resides) and changed to place
its result into the destination loc via `putIntoDest` (which, if
necessary, emits a C assignment).

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -1722,6 +1722,25 @@ proc genSlice(p: BProc; e: CgNode; d: var TLoc) =
     localReport(p.config, e.info, "invalid context for 'toOpenArray'; " &
       "'toOpenArray' is only valid within a call expression")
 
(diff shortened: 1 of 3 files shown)


saem (approved):
Fix and regression test look good.

PR #1029, opened 2023-11-21
fix a move-analyzer bug

zerbina:
## Summary

Fix a bug with the move analyzer that caused some sinks to be
erroneously turned into copies rather than moves. This didn't affect
run-time correctness or cause memory leaks, but it did led to
unnecessary copies and errors for non-copyable types.

## Details

The `resume` template in `mirexec.traverse` only exits the enclosing
loop, which is a wrong when called from within the inner loop. In
effect, `return`, `break`, and other unstructured exits were treated as
fall-through.

In the abort case, `resume` now ensures that the outer loop is exited
too. This surfaced a second bug: the `wasMoved` injection didn't
consider destructors that were automatically lifted into `finally`
clauses (because some other local within the same scope escaped), thus
leading to double frees.

Marking all destructors for a scope requiring a `finally` as appearing
in a finalizer is now done in `computeDestructors`, making this
information available to `needsReset`.

### DIFF
--- a/compiler/sem/injectdestructors.nim
+++ b/compiler/sem/injectdestructors.nim
@@ -226,13 +226,20 @@ type
     ## Entity dictionary. Stores all entities relevant to destructor
     ## injection and the move analyser
 
+  DestroyEntry = tuple
+    scope: NodePosition ## the position of the enclosing 'scope' node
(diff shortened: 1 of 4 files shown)


saem (approved):
The move to a named tuple (`DestroyEntry`) is easier to follow, as well.

PR #1030, opened 2023-11-23
mirexec: rework CFG computation

zerbina:
## Summary

Rework how the control-flow graph for a MIR body is computed, reducing
the amount of time taken for the `injectdestructors` pass by ~14%. The
CFG stays the same, only its creation changes.

## Details

Instead of adding both the `fork`/`goto`/`loop` and `join` instruction
when processing a control-flow statement and then relying on a `sort`
call to bring the list into the correct order, a stack-based approach
is used.

Only the opening instruction (`fork`, `goto`, or, for loops, `join`) is
added when processing the start of a structured control-flow block,
with the corresponding `join` or `loop` instruction added when the
`End` node is processed.

This eliminates both the extra MIR tree traversals (which profiling
showed to be where most of the time was spent) and the `sort` call.

### DIFF
--- a/compiler/sem/mirexec.nim
+++ b/compiler/sem/mirexec.nim
@@ -22,9 +22,8 @@
 import
   std/[
     algorithm,
+    options,
     packedsets,
-    sets,
-    tables
   ],
   compiler/ast/[
     ast_types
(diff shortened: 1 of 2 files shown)


compiler/sem/mirexec.nim
```diff
@@ -142,93 +141,106 @@ func computeCfg*(tree: MirTree): ControlFlowGraph =
...
-    let target =
-      if isLoop:
-        assert a > b, "a loop CFG edge must describe backwards control-flow"
-        # a loop edge can't cross the boundary of a finalizer
-        b
+      ifs: seq[InstrPos]
+        ## stack of instruction positions for the currently open
+        ## structured control-flow blocks (if, loop, and regions).
+      blocks: seq[tuple[label: LabelId, hidden: bool]]
+        ## stack of targets for forward, mergin control-flow. `label` and
```
saem:
```suggestion
        ## stack of targets for forward, merging control-flow. `label` and
```

compiler/sem/mirexec.nim
```diff
@@ -142,93 +141,106 @@ func computeCfg*(tree: MirTree): ControlFlowGraph =
...
+        ## `hidden` are mutually exclusive.
```
saem: could `hidden` then be reduced to the presence of a `default(LabelId)` value?

compiler/sem/mirexec.nim
```diff
@@ -142,93 +141,106 @@ func computeCfg*(tree: MirTree): ControlFlowGraph =
...
-               isLoop: bool): JoinPoint =
-    let target =
-      if isLoop:
-        assert a > b, "a loop CFG edge must describe backwards control-flow"
-        # a loop edge can't cross the boundary of a finalizer
-        b
+      ifs: seq[InstrPos]
+        ## stack of instruction positions for the currently open
+        ## structured control-flow blocks (if, loop, and regions).
+      blocks: seq[tuple[label: LabelId, hidden: bool]]
```
saem:
```suggestion
      strctrdBlkStack: seq[InstrPos]
        ## stack of instruction positions for the currently open
        ## structured control-flow blocks (if, loop, and regions).
      mrgTrgts: seq[tuple[label: LabelId, hidden: bool]]
```

Possibly alternative names based on the doc comments... they might be too wordy/not quite right. Just popped into my head, feel free to ignore.

saem (approved):
Nice, I like how declarative things are with the template based dsl (`emit/join/findBlock/etc`).

compiler/sem/mirexec.nim
zerbina: The default value for `LabelId` is a valid label, so that wouldn't work.

However, since `std/options` is now already imported, I think I'll change the tuple into an `Option`.

**EDIT**: I've replaced the tuple with an `Option[LabelId]`.

compiler/sem/mirexec.nim
zerbina: I've replaced `ifs` with `structStack` and kept `blocks` for now.

zerbina:

PR #1031, opened 2023-11-25
MIR: refactor the data-flow analysis

zerbina:
## Summary

Change the data-flow analysis implementation used by the MIR phase
(`mirexec`) to rely less on the MIR's tree structure, preparing for
changes to the latter. This also speeds up the traversal and simplifies
the implementation.

## Details

The MIR data-flow analysis used a hybrid approach so far: `mirexec`
provided control-flow graph traversal routines for traversing the basic
blocks in a MIR tree, with the data-flow relevant effects directly
computed from the MIR nodes during traversal.

### Problems

This approach worked well, but it has the problem of necessitating that
MIR nodes appear in the order their data-flow effects happen, which
is what ultimately led to the current instruction-like design of the
MIR. An exception was already made for calls and assignments (refer to
`computeValuesAndEffects`), where mutation effects have to be applied
when the *call/assignment* node is visited, not when the argument's
*tag* node (which comes earlier) is visited.

### Revised implementation

In order to make evolving the MIR's layout possible without having to
change the data-flow analysis algorithms, the data-flow effects are now
encoded directly in the control-flow graph, making it a data-flow
graph.

**Data-flow operations**:
* `opDef`: corresponds to `ekReassign` effects and is also used for
  definitions (`mnkDef`)
* `opUse`: represents all lvalue usages that are not consumes
* `opConsume`: corresponds to `mnkConsume` and other consume contexts
* `opKill`, `opMutate`, `opInvalidate`: correspond to the
  `EffectKind` of the same name
* `opMutateGlobal`: represents a mutation of an unspecified global

The data-flow instruction are emitted together with the control-flow
instructions in `computeCfg`.

**Changes:**
* move the `ConsumeCtx`, `UseContext`, `OpsWithEffects` sets into
  `mirtrees`
* instead of iterating over the tree and having to map the positions to
  control-flow instructions, the `traverse`, `traverseReverse`, and
  `traverseFromExits` iterators only need to iterate the instruction
  list (which renders the `tree` parameters obsolete)
* the iterators return the data-flow operation kind plus expression
  directly

**Changes to the usage sites:**
* change the `traverse` iterator to start at `start` rather than at
  `start + 1`. This is more intuitive and easier to implement. The
  transitive callsites `isLastRead`, `isLastWrite`, `computeOwnership`,
  and `needsReset` are adjusted accordingly, by applying the offset
  themselves
* support `start` pointing outside the provided `span` in `traverse`,
  making the API easier to use (e.g., no extra guard logic is needed)
* remove the special handling for `mnkDef` statements from
  `analysis.computeAlive` -- the information is now encoded in the
  data-flow graph
* remove the now-obsolete effect tracking for `Call`, `Asgn`, etc.
  nodes
* the `tmir_exec.nim` is updated to work with the `mirexec` changes

`ControlFlowGraph` is renamed to `DataFlowGraph`, `computeCfg` to
`computeDfg`, and the associated documentation is updated to reflect
the shift in purpose.

### Additional benefits

* computing the data-flow operations in a single place (i.e., during
  data-flow graph create) means that downstream users don't have to
  duplicate the logic
* the traversal routines no longer need a `MirTree`, making it much
  easier to generalize them into something that can be used without the
  MIR

### DIFF
--- a/compiler/sem/mirexec.nim
+++ b/compiler/sem/mirexec.nim
@@ -1,19 +1,17 @@
-## This module implements algorithms for traversing a control-flow graph (=CFG)
-## in forward or backward direction. In the abstract, it means visiting the
-## nodes in topological-order (forward) or post-order (backward), with special
-## handling for loops.
+## Implements a data-flow graph (=DFG) representation plus algorithms for
+## traversing the graph in both forward or backward direction. In the
+## abstract, this means visiting the nodes in topological-order (forward)
+## or post-order (backward), with special handling for loops.
 ##
 ## In the compiler, this is mainly used for control- and/or data-flow
-## analysis, meant to propagate properties through the graph or to to answer
+## analysis, meant to propagate properties through the graph or to answer
 ## questions such as: "is point A connected to point B?", "is X initialized on
 ## all paths?", etc.
 ##
-## Before traversing a control-flow graph, one has to be created via
-## ``computeCfg`` first. Instead of a pointer-based graph structure, the graph
-## is represented via a linear list of instruction.
-##
-## TODO: expand this section with how the algorithms used by for- and backward
-##       traversal work
+## A DFG is built via the `computeCfg <#computeCfg,MirTree>`_ routine. Instead
+## of a pointer-based graph structure, the graph is represented as a sequence
+## of instructions encoding all data- and control-flow relevant properties
+## for a piece of code.
 
 # XXX: with a small to medium amount of work, the algorithms and
 #      ``ControlFlowGraph`` can be generalized to work independent from the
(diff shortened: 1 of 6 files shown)


compiler/sem/mirexec.nim
```diff
@@ -677,18 +722,19 @@ iterator traverseReverse*(tree: MirTree, c: ControlFlowGraph,
...
       # the start
       s.time = high(Time)
 
   exit = s.active
 
-iterator traverseFromExits*(tree: MirTree, c: ControlFlowGraph,
-                            span: Slice[NodePosition], exit: var bool
-                           ): (NodePosition, lent MirNode) =
+iterator traverseFromExits*(c: ControlFlowGraph, span: Slice[NodePosition],
+                            exit: var bool): (Opcode, OpValue) =
```
saem, line 738: Do you think it's worth creating a named type for `(Opcode, OpValue)`?

compiler/sem/mirexec.nim
```diff
@@ -55,12 +68,16 @@ type
       dest: JoinId
     of opJoin:
       id: JoinId
+    of opUse, opDef, opConsume, opInvalidate, opMutate, opKill, opMutateGlobal:
```
saem: Not sure if this is a win, but if the constant `DataFlowOps` is moved up, then it could be used here.

saem (approved):
I left two minor comments, I don't think any thing needs to change.

compiler/sem/mirexec.nim
zerbina: I originally decide against doing so due to the `type` section having to be split into two, but not duplicating the set is probably better.

compiler/sem/mirexec.nim
zerbina: Hm, given that all callsites unpack the tuple right away, I don't think it's worth it, but I'm also not strongly against it.

PR #1032, opened 2023-11-26
internal: fix `skConst` translation in `mirgen`

zerbina:
## Summary

Fix top-level `skConst` symbols being translated to `mnkGlobal` MIR
nodes.

## Details

The usage-symbol translation in `mirgen` (`nameNode`) treated all
symbols marked with the `sfGlobal` flag as globals, but since `skConst`
symbols defined at the top-level are also marked with the flag, they
too were translated to globals.

This incorrect translation didn't cause any problems, as:
* `injectdestructors` ignored the globals, since no `mnkDef` exists for
  them in the analyzed body
* both `mnkConst` and `mnkGlobal` are translated to `cnkSym`, so the
  code generators weren't affected

`nameNode` is changed to use a case statement where the `sfGlobal`
flag is only considered for the relevant symbol kinds.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -419,14 +419,18 @@ proc genEmpty(c: var TCtx, n: PNode): EValue =
   result = EValue(typ: c.graph.getSysType(n.info, tyVoid))
 
 func nameNode(s: PSym): MirNode =
-  if sfGlobal in s.flags:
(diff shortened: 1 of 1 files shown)


saem (approved):
Ah, so we were consider them to be runtime globals prior to this change, as opposed to simply compile time global constants?

zerbina:
> Ah, so we were consider them to be runtime globals prior to this change, as opposed to simply compile time global constants?

Yep, that's what effectively happened.

PR #1034, opened 2023-11-27
internal: improve the `mirexec` API

zerbina:
## Summary

Refactor the `traverse` iterators to accept instructions spans/indices
and provide translation routines for node -> instruction translation.
This slightly simplifies the implementation, allows for instruction-
level addressing, and it reduces the amount of node-position ->
instruction translation.

## Details

It's not necessary for the `traverse` iterator to perform the node
position -> instruction translation, and since multiple instructions
may be associated with a single node, this also limited the
granularity at which traversal could start/end to nodes.

The `Subgraph` type is added, representing a sub-graph (i.e.,
instruction span) within the data-flow graph. Computing the `Subgraph`
and translating a node to an instruction position are done via
`subgraphFor` and `find`. All three `traverse` iterators are changed
to accept both a `Subgraph` and `InstrPos`. 

The change in API is propagated through the routines in the `analysis`
module, allowing for performing the node span -> `Subgraph` translation
only once for each analyzed location (instead of at the start of each
traversal). Instead of the enclosing `mnkScope`'s sub-tree,
`EntityInfo` stores the pre-computed `Subgraph` -- `getAliveRange` is
renamed to `getAliveGraph` and returns the subgraph directly.

Finally, the `traverse` iterators are changed to return a
`DataFlowOpcode` rather than an `Opcode` value, eliminating unnecessary
`case` branches when matching over the value.

### DIFF
--- a/compiler/mir/analysis.nim
+++ b/compiler/mir/analysis.nim
@@ -214,11 +214,10 @@ func computeValues*(body: MirTree): Values =
       discard "leave uninitialized"
 
 func isAlive*(tree: MirTree, cfg: DataFlowGraph, v: Values,
-             span: Slice[NodePosition], loc: LvalueExpr,
-             pos: NodePosition): bool =
-  ## Computes if the location named by `loc` does contain a value at `pos`
-  ## (i.e. is alive). The performed data-flow analysis only considers code
-  ## inside `span`
+             span: Subgraph, loc: LvalueExpr, start: InstrPos): bool =
+  ## Computes whether the location named by `loc` does contain a value (i.e.,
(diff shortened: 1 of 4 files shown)


zerbina:
Does anybody still want to review this PR? It's blocking the cleanup pass in #942.

saem:
I'm reviewing this now.

saem (approved):
Moving to the more fine grained addressing (instructions) is likely best for the lower level of abstraction API. If consumers need any sort of guarantee around addressing at the node level or node boundaries, they can trivially be wrapped. Makes sense.

PR #1035, opened 2023-11-29
nim secret (Fix to work again)

nixfreak:
When running `nim secret` the default compiler is backendC and should be backendVm.

## Summary
* what changed and how?
Added logic to compilation environment to disable ```backendC``` when calling ```nim secret``` and instead change to ```backendVm```
* why are we changing it? 

> Okay, the root of the problem is that the compilation environment is configured for as if the C backend were used. This leads to various semantic errors, and while that still shouldn't crash the compiler, it unfortunately does. Anyhow, since the REPL is effectively a backend itself, the customizeForBackend call at main.nim:510 needs to be disabled when the REPL is used (indicated bycmdInteractive).
commandInteractive (the procedure setting up the REPL) then needs to call customizeForBackend itself, but with backendVm rather than backendC.

## Details
* Credit goes to @zerbina:matrix.org

---

## Notes for Reviewers

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/front/main.nim
+++ b/compiler/front/main.nim
@@ -505,7 +505,7 @@ proc mainCommand*(graph: ModuleGraph) =
 
   ## command prepass
   if conf.cmd == cmdCrun: conf.incl {optRun, optUseNimcache}
-  if conf.cmd notin cmdBackends + {cmdTcc, cmdNimscript}:
+  if conf.cmd notin cmdBackends + {cmdTcc, cmdNimscript, cmdInteractive}:
     customizeForBackend(graph, conf, backendC)
   if conf.outDir.isEmpty:
     # doc like commands can generate a lot of files (especially with --project)
@@ -660,7 +660,9 @@ proc mainCommand*(graph: ModuleGraph) =
     wantMainModule(conf)
     commandView(graph)
     #msgWriteln(conf, "Beware: Indentation tokens depend on the parser's state!")
-  of cmdInteractive: commandInteractive(graph)
+  of cmdInteractive:
+    customizeForBackened(graph, conf, backendNimVm)
+    commandInteractive(graph)
   of cmdNimscript:
     if conf.inputMode == pimFile and not fileExists(conf.projectFull):
       localReport(conf, InternalReport(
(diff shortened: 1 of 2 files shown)


compiler/front/main.nim
```diff
@@ -660,7 +660,9 @@ proc mainCommand*(graph: ModuleGraph) =
     wantMainModule(conf)
     commandView(graph)
     #msgWriteln(conf, "Beware: Indentation tokens depend on the parser's state!")
-  of cmdInteractive: commandInteractive(graph)
+  of cmdInteractive:
+    customizeForBackened(graph, conf, backendNimVm)
```
zerbina, line 664:
```suggestion
    customizeForBackend(graph, conf, backendNimVm)
```

zerbina (changes requested):
Thank you for creating the PR, having `nim secret` operational again makes quickly trying out some code much easier. I left a suggestion for fixing the typo with procedure name.

So that future regression don't go by unnoticed, could you add a basic test for the REPL? The test (e.g., `trepl.nim`) should go into the `tests/compiler` directory and do the following:
* invoke the compiler with the `secret` command
* ensure that the three arrows appear without any errors
* send an `echo ...` statement to the process
* verify that the echoed string appears in the process' output
* send a `quit()` statement to the process
* verify that the process terminates

This ensures that the REPL starts and can be shutdown again.

---

Both the title and PR body need some adjustments:
* since it's a user-facing bugfix, the title should broadly summarize what the fix does (e.g., "make the `nim secret` command work again")
* what's currently in the Summary section should be part of the Details section, and the quote should be shortened a bit
* finally, the Summary section then summarizes the Details section, which in the case of this PR could be a slightly more detailed version of the title I suggested

If okay with you, I can also update both the title and PR body to be in line with
the general PR style.

nixfreak:
Looking into how to test nim secret via nim files.

zerbina:
@nixfreak: How's the progress on this PR?

zerbina:
Superseded by #1351.

PR #1036, opened 2023-11-30
fix(cgen): run-time crash when modifying a `seq`

zerbina:
## Summary

Fix a crash when trying to add to, set the length of, or assign to a
sequence part of a tuple or array location that was initialized with a
fully constant tuple or array construction expression.

## Details

Due to `seq`s not supporting copy-on-write behaviour like strings do,
fully constant seq construction, which are lifted into C constants by
the C code generator have to be copied (via their `=copy` hook). As
an optimization, empty seq construction were exempt from this, as
they're not lifted into C constants due to an unrelated `cgen` issue.

When an empty seq construction reaches the code generator as part of
a tuple/array construction (e.g., `(TupleConstr (ArrayConstr))`), the
tuple/array construction was lifted into a C constant, which then
caused an memory access violation when adding to, setting the length
of, or assigning to the `seq` stored in the variable that the tuple/
array was assigned to.

The crash happened because the seq's payload was marked with the
`strlitFlag`, which the `add` and `setLen` procedures don't account
for, treating the sequences as having enough capacity and then
trying to write into its non-writeable buffer.

### The fix

The special casing of empty `seq` constructions is removed, meaning
that they're now also treated as non-moveable values during hook
injection -- translating `nkBracket` in `cgirgen.translateLit`
becomes obsolete too.

### Alternatives

Alternatively, `add`, `grow`, and `setLen` could take the `strlitFlag`
into account, but this alone wouldn't allow for removing the
`nkBracket` workaround.

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -2157,10 +2157,11 @@ proc expr(p: BProc, n: CgNode, d: var TLoc) =
     else:
       genSetConstr(p, n, d)
   of cnkArrayConstr:
-    let kind = skipTypes(n.typ, abstractVarRange).kind
-    if isDeepConstExpr(n) and (kind == tySequence or n.len != 0):
+    # XXX: constructions of empty seqs should be lifted into C constants too,
+    #      but that currently causes collisions and thus C compiler errors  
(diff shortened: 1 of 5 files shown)


saem (approved):
It's better in that things aren't smeared across layers.

PR #1038, opened 2023-11-30
fix(c): `toOpenArray` issues with `array` operand

zerbina:
## Summary

Fix two C code generator issues with `toOpenArray`:
* passing a dereferenced `ref array` or `lent array` to `toOpenArray`
  no longer results in an internal compiler error
* bound checks are now correctly performed for dereferenced `ptr array`
  and `var array` `toOpenArray` arguments

## Details

The C code generator translates pointer-to-array types (e.g.,
`ref array[T]`, `ptr array[T]`, etc.) to `ptr T` and then omits the
deference. As a consequence, the type of C expression (`TLoc`) returned
by `ccgexpres.genDeref` is still a `ptr array[T]`, etc., but
`genOpenArraySlice` didn't account for this.

Pointer-like types are now skipped by both `genOpenArraySlice` and
`genBoundsCheck` (which is only used by `genOpenArraySlice`) fixing the
issues that arose from pointer-like types being ignored.

The case statement in `genBoundsCheck` is made exhaustive too, in order
to prevent bound checks from being silently omitted.

There was also a test (`toverload_bracket_accessor_template.nim`) that
relied on the bound check to be omitted. Since the test is for
overloading and doesn't depend on `ptr array` being used, it is changed
to use `ptr UncheckedArray` instead.

### DIFF
--- a/compiler/backend/ccgcalls.nim
+++ b/compiler/backend/ccgcalls.nim
@@ -147,7 +147,7 @@ proc genOpenArraySlice(p: BProc; q: CgNode; formalType, destType: PType): (Rope,
   # but first produce the required index checks:
   if optBoundsCheck in p.options:
     genBoundsCheck(p, a, b, c)
-  let ty = skipTypes(a.t, abstractVar+{tyPtr})
+  let ty = skipTypes(a.t, abstractVar+{tyPtr, tyRef, tyLent})
   let dest = getTypeDesc(p.module, destType)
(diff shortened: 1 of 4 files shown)


zerbina:
The type skipping is a bit of a hack, but it will fortunately become obsolete in the future,
as:
* inserting the bound checks will happen as a MIR pass
* `openArray` will be lowered into a tuple prior to code generation, at least for the C backend

saem (approved):

PR #1040, opened 2023-12-01
js: improve cast-to-`pointer` support

zerbina:
## Summary

Support casting `ref` of aggregates, `proc`, and scalar types to
`pointer` values and the pointer back to the original type, for the JS
backend. Using the result of these casts previously either worked,
misbehaved at run-time, or caused an internal compiler error.

## Details

The `jsgen.genCast` procedure didn't account for all non-fat-pointer
types in `genCast`. If the `TCompRes` resulting from such a cast
reached `genVarInit`, the assertion for the `etyBaseIndex` handling
triggered.

As an interim solution, `genCast` now handles all to- and from-
`pointer` casts the same:
* non-`etyBaseIndex` values are cast into `pointer` by treating the
  expression as the address and `null` as the index (already happened
  for `etyObject`)
* `pointer` values are cast into non-`etyBaseIndex` values by using the
  address part as the expression (the inverse of the to-`pointer` cast)
* `pointer` to fat-pointer casts and vice versa are no-ops

This implementation means that casting `A` to `pointer` and the
`pointer` to `B` is undefined behaviour for the JS target, but that was
already the case before the change.

### Tests

Test coverage for casts is increased by:
* adding a test for casting `ref`-of-scalars to `pointer` and back
* adding a test for casting `ref`-of-aggregate to `pointer` and back
* adding a test for casting an `int` to `pointer` and back

The not-yet-working `misc/tcast.nim` is marked as a known issue for the
JS target.

### DIFF
--- a/compiler/backend/jsgen.nim
+++ b/compiler/backend/jsgen.nim
@@ -2374,7 +2374,7 @@ proc genCast(p: PProc, n: CgNode, r: var TCompRes) =
   if toUint and (fromInt or fromUint):
     let trimmer = unsignedTrimmer(dest.size)
     r.res = "($1 $2)" % [r.res, trimmer]
-  elif toInt:
+  elif toInt and (fromInt or fromUint):
     if fromInt:
       return
     elif fromUint:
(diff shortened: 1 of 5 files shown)


saem (approved):
That's a nice and simple fix, and the test coverage should pay off in spades.

zerbina:
/merge

It looks like merging failed because of a network error during installation the Linux runner's dependencies.

PR #1042, opened 2023-12-01
ci: try `knownIssue` tests

zerbina:
## Summary

Enable execution of `knownIssue` during CI test runs. This
automatically alerts about `knownIssue` tests starting to work, meaning
that one no longer has to check this manually.

Existing `knownIssue` tests that already work are updated or adjusted.

## Details

Execution of `knownIssue` tests is enabled by passing `--tryFailing` to
testament when testing in CI.

Multiple existing `knownIssue` tests already work, some also use the
`knownIssue` incorrectly. They're resolved in order for the test suite
to still pass.

Working tests:
* `tfragment_gc.nim` and `tfragment_alloc.nim` were both disabled due
  to out-of-memory failures caused by 4 GB allocations. The minimum
  amount of RAM provided by runners is currently 7 GB (according to the
 
[documentation](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners)),
  meaning that the tests work
* `t02_argument_passing_mutate_bug.nim`: has always been working with
  the VM target
*  `t01_var_openarray.nim` : fixed by
https://github.com/nim-works/nimskull/pull/756
*  `tviews_in_object.nim` : fixed by
https://github.com/nim-works/nimskull/pull/769
*  `tdouble_evaluation_bug.nim`  fixed by
https://github.com/nim-works/nimskull/pull/828
* `t04_type_inference_fail_colon_expr.nim`: fixed by
  https://github.com/nim-works/nimskull/pull/1005 and replaced by
  `lang_exprs/tempty_typed_expressions.nim`

Tests that have been working prior to the most recent csources update
(~9 months ago):
* `t08_varargs_regular_subtype.nim`
* `tforloop_tuple_unpacking.nim`
* `tdata_serdes.nim`
* `tsequtils.nim`
* `tresults.nim` (for the JS target)

Tests that require specification changes:
* `t01_borrow_literal.nim`: the test tests for a compilation error, but
  was missing the `reject` specification
* `t07_addr_borrow_fail.nim`: the current language specification says
  that `addr x` is a path expression, but also that `ptr lent T` is not
  a view type. Given the unclear specification, the test is changed
  into testing for an error rather than success
* `t06_varargs_converter_clash.nim`: the test intends to ensure
  success, but the `errormsg` key means it does the opposite. The
  `errormsg` key is removed and turned into a comment instead
* `tmove_from_lvalue_conversion.nim`: doesn't rely on `--gc:orc` and
  already works with the JS and VM targets
* `thttpclient.nim`: the issue is with the test itself. It is disabled
  in addition to the `knownIssue` marker

Running `knownIssue` tests in the CI represents progress towards
https://github.com/nim-works/nimskull/issues/61.


saem (commented):
I think if we're enabling try failing then perhaps `isJoinable` should be updated to return false if there is a known issue matching the target?

zerbina:
> I think if we're enabling try failing then perhaps `isJoinable` should be updated to return false if there is a known issue matching the target?

`isJoinableSpec` already returns false for a spec that has a `knownIssue`, although not on a per-target basis. That is, if a spec has at least one `knownIssue`, joining a run created from the spec into the megatest is disabled.

saem:
> > I think if we're enabling try failing then perhaps `isJoinable` should be updated to return false if there is a known issue matching the target?
> 
> `isJoinableSpec` already returns false for a spec that has a `knownIssue`, although not on a per-target basis. That is, if a spec has at least one `knownIssue`, joining a run created from the spec into the megatest is disabled.

Whoops, I read it quickly and didn't pick up on that, on which case all good.

PR #1044, opened 2023-12-02
fix(sem): some caught exceptions being leaked

zerbina:
## Summary

Fix exceptions caught by `T as e` handler branches within closure
iterators where `T` is an object type being leaked when the handler
is entered multiple times.

## Details

* add the `makeRefType` procedure to `semdata`, for creating a `ref`
  type of a given type while properly setting the `tfHasAsgn` flag
* use `makeRefType` in `semTry` for creating the ref type of the non-
  ref exception type
* also set the type of symbol *node* for the `T as e` part; this was
  previously missing, leaving the node's type as nil
* take the type for the conversion expression created during lowering
  of the handler branch from the symbol rather than creating a new and
  improper `ref` type via `toRef`
* remove the `ast.toRef` procedure; it's unused now and also not
  generally applicable (due to the missing `tfHasAsgn` handling)

### The problem

The lowering of `except T as e:` produces a
`let e = T(getCurrentException())` statement, using `T`'s type as
the the right-hand expression's type. If `T` wasn't a `ref` type
already, a new `tyRef` type was first created via `ast.toRef`.

The `tyRef` created by `toRef` doesn't have the `tfHasAsgn` flag set,
however, and since no more automatic hook lifting takes place after
`transf`, the ref type had no attached hooks and thus no sink or copy
hook calls are injected for assignments it's part of.

This is not a problem for the when `e` is not lifted into an object, as
the local is still destroyed and an initial assignment without a hook
call is correct. However, when `e` is lifted into an object and
assigned to multiple times, the missing `=sink` call means that the
`ref` already stored in the field is not freed, leaking the heap cell.

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -586,14 +586,6 @@ proc toVar*(typ: PType; kind: TTypeKind; idgen: IdGenerator): PType =
     result = newType(kind, nextTypeId(idgen), typ.owner)
     rawAddSon(result, typ)
 
-proc toRef*(typ: PType; idgen: IdGenerator): PType =
-  ## If ``typ`` is a tyObject then it is converted into a `ref <typ>` and
-  ## returned. Otherwise ``typ`` is simply returned as-is.
-  result = typ
-  if typ.skipTypes({tyAlias, tyGenericInst}).kind == tyObject:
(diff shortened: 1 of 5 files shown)


saem:
fixed a minor type in the pr message

saem (approved):

PR #1045, opened 2023-12-02
fix(vm): assigning string literals to `openArray`

zerbina:
## Summary

Fix accessing the element of a first-class `openArray` resulting in a
access violation run-time errors or wrong results, if the `openArray`
was initialized from a string literal. Only the VM backend was
affected.

## Details

Initial assignments to local `openArray` bindings used a handle copy
(`FastAsgnComplex`), which works if the source is a non-temporary
location. If it isn't, the handle stored for the `openArray` local
points to a freed or reused memory cell after the temporary source
register is reclaimed.

Loading a string literal allocates a temporary string location, and
thus they were affected by this issue.

For fixing the issue, a handle copy is only created when the source
register is the register backing a local -- per the language rules, the
source lvalue has to outlive the `openArray` binding, so this is safe.
If the source is some temporary register, it is promoted to the
register backing the local.

### DIFF
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -2841,15 +2841,30 @@ proc genDef(c: var TCtx; a: CgNode) =
           s   = a[0].local
           typ = a[0].typ
         if true:
-          let reg = setSlot(c.prc, s)
           if a[1].kind == cnkEmpty:
             # no initializer; only setup the register (and memory location,
             # if used)
+            let reg = setSlot(c.prc, s)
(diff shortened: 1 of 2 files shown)


saem:
FYI: Updated PR body to fix typo.

saem (approved):

zerbina:
The `tfragment_alloc.nim` test that was enabled yesterday appears to be flaky. This occurrence might be an outlier, although I suspect it's not -- further investigation is most likely required.

PR #1046, opened 2023-12-03
cgen: don't emit unnecessary type definitions

zerbina:
## Summary

Don't emit the full C type definition for a type when assigning a `ref`
or `ptr` of said type, reducing the amount of generated C code and thus
the amount of work for both the NimSkull and C compiler.

## Details

The full C type definition of the pointed-to type was requested on
assignment (`ccgstmts.genAsgn`) of the pointer-like type, but this is
unnecessary, as C doesn't require the pointed-to type to be a complete
type when assigning the pointer.

It's likely that this is a leftover from when `genDeref` didn't request
the type definition of the pointed-to type. Requesting the type
definition is removed, as translating the left and right side of the
assignment already pulls in the relevant type definitions.

### DIFF
--- a/compiler/backend/ccgstmts.nim
+++ b/compiler/backend/ccgstmts.nim
@@ -803,7 +803,6 @@ proc genAsgn(p: BProc, e: CgNode) =
     let le = e[0]
     let ri = e[1]
     var a: TLoc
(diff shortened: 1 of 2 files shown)


saem (approved):
This reminds me of how much thinking nimskull's partial/unknown still needs. 😅

PR #1047, opened 2023-12-03
system: explicit `<`/`<=` overloads for `cstring`

zerbina:
## Summary

Introduce an explicit overload of the `<` and `<=` system operators for
`cstring`, removing the reliance on `cstring` being implicitly
convertible to `pointer`.

## Details

`cstring` is currently always implicitly convertible to `pointer`,
which meant that the `LePtr` and `LtPtr` magic overloads were chosen so
far (through introduction of an implicit conversion for the arguments).

This is a problem for the upcoming MIR rework, as there the result of
each operation has to be assignable to a temporary. Since `jsgen`
treats the `cstring`-to-`pointer` conversion as a no-op, the assignment
would result in a typing error within the JS code generator.

As a preparation for both the MIR rework and for revisiting the
semantics of cstrings, explicit `<` and `<=` overloads are introduced
for `cstring`. Using imported overloads when the JS backend is enabled
wouldn't work, as that'd prevent the comparison operators from being
used in compile-time(-only) contexts.

The introduction is backwards compatible with the csources compiler, so
no new conditional symbol is required. For `tset_with_range.nim` the
number of mismatching overloads in the error message is updated.

### DIFF
--- a/lib/system/comparisons.nim
+++ b/lib/system/comparisons.nim
@@ -80,7 +80,7 @@ proc `<=`*[T](x, y: set[T]): bool {.magic: "LeSet", noSideEffect.} =
 proc `<=`*(x, y: bool): bool {.magic: "LeB", noSideEffect.}
 proc `<=`*[T](x, y: ref T): bool {.magic: "LePtr", noSideEffect.}
 proc `<=`*(x, y: pointer): bool {.magic: "LePtr", noSideEffect.}
-proc `<=`*(x, y: cstring): bool {.magic: "LtPtr", noSideEffect.}
+proc `<=`*(x, y: cstring): bool {.magic: "LePtr", noSideEffect.}
 
 proc `<`*[Enum: enum](x, y: Enum): bool {.magic: "LtEnum", noSideEffect.}
 proc `<`*(x, y: string): bool {.magic: "LtStr", noSideEffect.} =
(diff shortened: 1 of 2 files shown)


lib/system/comparisons.nim
```diff
@@ -80,6 +80,7 @@ proc `<=`*[T](x, y: set[T]): bool {.magic: "LeSet", noSideEffect.} =
 proc `<=`*(x, y: bool): bool {.magic: "LeB", noSideEffect.}
 proc `<=`*[T](x, y: ref T): bool {.magic: "LePtr", noSideEffect.}
 proc `<=`*(x, y: pointer): bool {.magic: "LePtr", noSideEffect.}
+proc `<=`*(x, y: cstring): bool {.magic: "LtPtr", noSideEffect.}
```
saem: Shouldn't this be `LePtr`?

saem (commented):
Spotted a probable error, maybe I'm misunderstanding something.

lib/system/comparisons.nim
zerbina: It should be! Good catch.

zerbina:
```suggestion
proc `<=`*(x, y: cstring): bool {.magic: "LePtr", noSideEffect.}
```

saem (approved):
LGTM

zerbina:
The `tfragment_alloc.nim` test is failing with the MacOS runner again. I'll try to take a closer at the issue in the next week.

PR #1050, opened 2023-12-06
Link to `style_guide.html` in `intern.rst`

nixfreak:
## Summary

`nep1.html` was renamed to `style_guide.html`, but the reference
to it in `intern.rst` wasn't updated, resulting in a broken link.
This is fixed now.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -143,7 +143,7 @@ git clone https://github.com/nim-works/nimskull.git
 cd nimskull
 ./koch.py boot -d:release
 ./koch.py tools -d:release
-**Optional** Build all nimSkull documentation
+ Build all nimSkull documentation
 ./koch.py docs -d: release
 ```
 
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -143,6 +143,8 @@ git clone https://github.com/nim-works/nimskull.git
 cd nimskull
 ./koch.py boot -d:release
 ./koch.py tools -d:release
+ Build all nimSkull documentation
+./koch.py docs -d: release
```
zerbina:
```suggestion
```
An unrelated change part of the other PR.

zerbina (changes requested):
Thank you for fixing documentation issues. The documentation generator doesn't check for stale links or anchors at the moment (but it really should), so it's easy for links to start pointing to non-existent documents.

The title is good, but the PR message needs to not contain parts of the message template.

Also, for improved readability, could you wrap the names of the files within the PR message and title in backward quotes? That'll make them appear like `this`, which is slightly easier to read. Thanks.

readme.md
Clyybber: Commited the suggestion now, since "Maintainers are allowed to edit this pull request." is enabled on this PR.

zerbina (approved):

PR #1059, opened 2023-12-19
vm: move thread-local state into `VmThread`

zerbina:
## Summary

Internal-only refactoring. Move all state associate with a single
thread of execution into `VmThread`. This means that it's now possible
to create multiple threads from a single `TCtx` and run them
concurrently.

## Details

* store the stack-frame data, exception handling state, and loop
  iteration counter in `VmThread` rather than `TCtx`
* remove the `frame` field from `TCtx`; it's never read from
* remove `currentExceptionB`; it has been unused for a long time
* `currentExceptionA` is renamed to just `currentException`
* update the usage sites accordingly
* pass the executed thread to `rawExecute`

The restriction of only one `VmThread` being able exists at a time is
thus lifted, paving the way for eventual multi-threading support. In
the shorter-term, `rawExecute` having access to the `VmThread` object
means that adding new thread-local state doesn't require adding new
types to `vmdef`.

### DIFF
--- a/compiler/vm/compilerbridge.nim
+++ b/compiler/vm/compilerbridge.nim
@@ -348,7 +348,7 @@ proc execute(jit: var JitState, c: var TCtx, start: int, frame: sink TStackFrame
         "non-static stmt evaluation must produce a value, mode: " & $c.mode
       let reg =
         if r.reg.isSome:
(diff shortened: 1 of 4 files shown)


saem (approved):
Very cool, that's a big unblock for a small change.

PR #1060, opened 2023-12-21
sem: clarify `static[T](x)` semantics

zerbina:
## Summary

Implement `static[T](x)` as being semantically equivalent to
`static(T(x))`. As a consequence:
* the `x` expression in the coercion is now always evaluated at
  compile-time, with an error being reported if it's not
* `static[T](U)` (where `U` is a type expression) is not legal anymore.
  It was previously treated as a type construction of `static[U]`

In typed expression AST, `static int` is now represented as
`(StaticTy (Sym "int"))` instead of
`(Command (Sym "static") (Ident "int))`.

## Details

* reorder the `tyStatic` handling in `semConv` such that `static[T]` is
  handled separately from `static`
* interpret and sem a `static[T](x)` conversion expression as
  `(StaticExpr (Conv (x)))`
* don't evaluate `tyTypeDesc`-typed expression with the VM in
  `semStaticExpr`
* don't treat `nkType` without a `tyTypeDesc` type as a valid type
  expression (because it's not)
* in `semConv`, emit a `nkStaticTy` tree for `static` type
  constructions, properly reflecting the meaning of the expression
* `semStaticType` cannot be used in `semConv` anymore, as `evaluated`
  is no longer an `nkType` node. Re-semming the sem'ed type expression
  leads to errors at the moment
* update the comment regarding `tyTypeDesc` expressions in `mirgen`

Besides clarifying the expression's semantics, this also removes the
costly and unnecessary VM invocation (`transf` -> MIR -> VM) for
`static` type constructions.

In addition, `semStaticExpr` now tries to fold the expression first,
further reducing unnecessary VM invocations.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1874,7 +1874,6 @@ proc genWithDest(c: var TCtx, n: PNode; dest: Destination) =
   else:
     gen(c, n)
 
-
 proc generateCode*(graph: ModuleGraph, options: set[GenOption], n: PNode,
                    code: var MirBuffer, source: var SourceMap) =
   ## Generates MIR code that is semantically equivalent to the expression or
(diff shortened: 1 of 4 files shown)


saem:
_I need to think things through outloud, please bear with me..._

Previous/what I think the meanings should have been:

1. `static` a type class for all `static[T]`
2. `static[int]` a type class for any compile time `int`
3. `static[int](_)` either:
    1. a conversion of type `_` to a `static[int]`, or
    2. a construction/conversion of a `static[int]` typed value
4. `static(int(_))`
    1. a static context conversion of a comptime known value `_`, and
    2. if `_` is a type then it should be an error... i think

I think the main issue is point `3.1` which also doesn't make sense from a syntax consistency/conceptual sense.

---

_think with a basic principle in mind_

`static[T] rest`, or `static rest`, really just means, evaluate the `rest` in a static context, the type of any such evaluation should be a `static[T]`, where `T` is the type of `rest`. This is the simplest form, I can think of.

So `static[int](rest)` can really only meant `static int(rest)`, in an expression position at least. `static int(char)` is silly, but doesn't necessarily have to be illegal... leaving it as an error for now isn't the end of the world.

---

OK, colour me convinced that the change makes sense, going to review the implementation.

saem (approved):

zerbina:
/merge

Thanks for the review, @saem.

PR #1063, opened 2023-12-22
ci: update artifact actions

zerbina:
## Summary

Update the artifact actions:
* https://github.com/actions/download-artifact to version 4 (was 3)
* https://github.com/actions/upload-artifact to version 4 (was 3)

Both updates require changes to the CI workflow, namely that each
release binary needs to use a separate name now.

## Details

The new versions use version 2 of the
[@actions/artifact](https://github.com/actions/toolkit/tree/main/packages/artifact)
package, which doesn't support uploading to the same artifact more than
once during a single workflow run.

This is resolved by creating two "release binaries" and "binaries from
source archives" artifacts (one for each target), with the platform
name as the suffix. For retrieving them, the download action's new
`pattern` and `merge-multiple` features are used. The `publisher` job
is adjusted too.

### DIFF
--- a/.github/actions/download-compiler/action.yml
+++ b/.github/actions/download-compiler/action.yml
@@ -9,7 +9,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/download-artifact@v3.0.2
+    - uses: actions/download-artifact@v4
       with:
         name: compiler ${{ runner.os }}
         path: "${{ runner.temp }}"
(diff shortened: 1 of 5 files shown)


alaviss (approved):

alaviss:
/merge

We will have to monitor the moment this PR hits `devel` to see if the publisher still works correctly.

I have verified that the download-artifact action we are using have been updated to use the new artifact API, so this should work.

alaviss:
Looks like the artifacts are extracted into subfolders:

https://github.com/nim-works/nimskull/actions/runs/7306273713/job/19910814438#step:5:28

PR #1065, opened 2023-12-23
cgir: use dedicated node kind per symbol kind

zerbina:
## Summary

Remove the `cnkSym` node kind and replace it with a dedicated kind for
each entity `cnkSym` previously represented. This brings the `CgNode`
IR closer to the MIR.

## Details

The `CgNode` IR groups symbols of constants, globals, fields, and
procedures under a single node kind, requiring dispatching based
on the `TSym.kind`. Which entity some logic expects is also not
expressible without a separate assertion.

In order to:
* reduce the amount of dispatching in the code generators
* bring the `CgNode` IR closer to the MIR
* prevent bugs caused by symbol kinds not being considered

the `CgNode` IR is changed to also use a dedicated node kind per
symbol.

Each usage of `cnkSym` is replaced with usage of the node kind(s)
that the respective code generator logic expects, which is context
dependent. A positive effect of this is that it is now clear which
entities are processed where.

Finally, a small cleanup is performed in `ccgstmts.genAsmOrEmitStmt`:
only the field symbol handling of the `cnkSym` branch is kept, for
all other symbol kinds, the else statement suffices.

### DIFF
--- a/compiler/backend/ccgcalls.nim
+++ b/compiler/backend/ccgcalls.nim
@@ -15,7 +15,7 @@ proc canRaiseDisp(p: BProc; n: CgNode): bool =
   ## control-flow, otherwise 'false'. If panics are disabled, this also
   ## includes all routines that are not certain magics, compiler procs, or
   ## imported.
-  if n.kind == cnkSym and {sfNeverRaises, sfImportc, sfCompilerProc} * n.sym.flags != {}:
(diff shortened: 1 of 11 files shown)


saem (approved):
I was going to suggest a prefix of `cnkSymX` for each of the dedicated symbol kinds, but since cgir will be subsumed soon, it's not worth it.

Also, your hypothesis was correct as the more explicit kind handling seems to have done a better job at surfacing issues. 🎉

PR #1066, opened 2023-12-25
mir: unify call syntax

zerbina:
## Summary

Make the MIR's call expression syntax more regular by not treating
`Magic` as a call-like node. Magic calls are now represented as
`(Call (<Magic>) ...)`, with the `Magic` node treated like a callee.

## Details

Saving a node per magic call doesn't outweigh the additional complexity
of accounting for the special tree shape (no callee slot) of
magic calls when analyzing call expressions.

The `Magic` node is now treated as a node allowed in the callee slot of
a `Call` tree, removing the special shape and thus simplifying
processing and analysis. More importantly, this allows for specifying
whether the *call* of a magic can raise -- this was previously a
property of the *magic*, not of the call thereof.

**Changes:**
- don't treat `mnkMagic` as a call-like node (and remove all usages of
  it as such)
- add a `[]` sub-node access operator overload that accepts an
  `OpValue`
- update `mirpasses.preventRvo` and `mirpasses.search` to account for
  the different magic call syntax
- add the `buildCheckedMagicCall` template to `mirgen`; it's used for
  emitting magic calls that are tagged as potentially raising
- add the `buildDefectMagicCall` template to `mirgen`; it's used for
  emitting magic calls that are only tagged as potentially raising if
  panics are disabled

Since in `mirgen` the `mChckRange` call is now emitted via
`buildDefectMagicCall`, compiler-injected range checks are correctly
treated as not raising an exception when panics are enabled
(`--panics:on`).

### DIFF
--- a/compiler/backend/cgirgen.nim
+++ b/compiler/backend/cgirgen.nim
@@ -621,16 +621,11 @@ proc callToIr(tree: TreeWithSource, cl: var TranslateCl, n: MirNode,
               cr: var TreeCursor): CgNode =
   ## Translate a valid call-like tree to the CG IR.
   let info = cr.info
-  case n.kind
-  of mnkMagic:
-    assert n.typ != nil
-    result = newExpr(cnkCall, info, n.typ)
-    result.add newMagicNode(n.magic, info)
-  of mnkCall:
-    result = newExpr(cnkCall, info, n.typ)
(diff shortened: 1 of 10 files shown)


saem (approved):
Definitely reads more linearly, nice.

PR #1068, opened 2023-12-26
internal: support multiple lifetimes for entities

zerbina:
## Summary

Update the `injectdestructors` pass to support an entity (global or
local) being defined multiple times within a MIR body. A workaround for
this previous shortcoming is now obsolete and thus removed.

### Background

No new global variables must be introduced after module lowering. For
this reason, `transf` doesn't introduce replicas for symbols of global
variables when inlining the same `for`-loop body multiple times
(because of more than one `yield` statement).

This, however, is a problem for the `injectdestructors` pass, as it, so
far, only supported an location existing once (i.e., there only being a
single 'def' for it). As a workaround, `transf` had to introduce
temporary replicas of the symbol, where are then replaced with the
originating-from symbol after the `injectdestructors` pass is over.

### The problem

The workaround is scattered across multiple layers and all new
processing that touches symbols of globals needs to account for the
fact that the symbol might not be the "correct" one.

### Solution

The `injectdestructors` pass is updated to support entity locations'
having multiple lifetimes within a body:
* associate an entity with a list of `EntityInfo` rather than only a
  single one (each one representing a single lifetime/scope)
* replace `getAliveGraph` with `findScope`, which looks up the entity's
  scope corresponding to the given data-flow instruction

A downside is that more allocations are performed due to the usage of
`seq`, but this is going to be resolved by injecting destructors
during AST -> MIR translation.

Creating a temporary copy of the symbol and later restoring it is
removed.

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -333,14 +333,11 @@ proc preprocess*(conf: BackendConfig, prc: PSym, graph: ModuleGraph,
 
 proc process*(prc: var Procedure, graph: ModuleGraph, idgen: IdGenerator) =
   ## Applies all applicable MIR passes to the procedure `prc`.
-  rewriteGlobalDefs(prc.body.tree, prc.body.source, patch=false)
+  rewriteGlobalDefs(prc.body.tree, prc.body.source)
 
   if shouldInjectDestructorCalls(prc.sym):
     injectDestructorCalls(graph, idgen, prc.sym,
                           prc.body.tree, prc.body.source)
-
-  # restore the correct symbols for globals:
-  patchGlobals(prc.body.tree, prc.body.source)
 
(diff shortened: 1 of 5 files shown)


saem (approved):
Would `lifetime` be a better name than `livetime`?

Otherwise looks good to me.

zerbina:
> Would `lifetime` be a better name than `livetime`?

Yep. Not only is it a better name, it's also correct, whereas `livetime` is not -- I'm not sure why I used `livetime` in the first place.

A "live" variable (or location) usually refers to a variable that still stores a value and is going to be read from again, thus meaning that `livetime` refers to the time span where a variable/location is "live". However, `EntityInfo` has nothing to do with whether or when the location is "live", only with the time of its *existence*.

PR #1069, opened 2023-12-27
vm: remove the `opcLdStrIdxAddr` opcode

zerbina:
## Summary

The opcode is fully redundant with `LdArrAddr`, and the specialization
provides little to no benefit given that taking the address of a
string's element is an uncommon operation.

## Details

* remove the `opcLdStrIdxAddr` opcode and its implementation
* `vmgen` emits an `opcLdArrAddr` instruction instead

Note that creating a `var` or `lent` view of a string's element did not
use `LdStrIdxAddr` previously. The dedicated opcode is a leftover from
the time where strings used a special in-VM representation.


saem (approved):
Yay, fewer opcodes

PR #1070, opened 2023-12-27
koch: remove self-exec and dead Windows code

alaviss:
## Summary
It doesn't make a lot of sense to use self-exec to handle
forward-declaration problems.

The old Windows archive building code has been obsolete after the
niminst rework, and appears to be the only other user of this feature.
Thus it is pruned as well.

As a bonus, this fixes  `koch.py all`  on Windows as there is no longer
a need to replace the running executable.


zerbina (commented):
Could you also remove the `kochExe` global? It's unused now.

Otherwise this looks good to me.

PR #1076, opened 2024-01-01
mir: separate index/bound checks from array access

zerbina:
## Summary

* code generators don't decide index and bound check placement -- the
  AST -> MIR translation does
* data-flow analysis considers index and bound errors, fixing
  destructors not being run when the only possibly raised exception is
  an `IndexDefect`
* bound checks are performed for `toOpenArray` calls when using the JS
  backend
* MIR semantics: allow mutations of projection from aliases created
  with `Bind`

### Translation changes

A statement like `arr[i] = (inc i; 1)` was previously translated to:
```
def _1 = i
bind_mut _1 = arr[_1]
inc(name i)
_2 = 1
```
It is now translated to:
```
def _1 = i
chckIndex(arg arr, arg _1)
inc(name i)
arr[_1] = 1 
```

Similarily, `call(toOpenArray(a, lo, hi))` is now translated to:
```
chckBounds(arg a, arg lo, arg hi)
def _1 = toOpenArray(arg a, arg lo, arg hi)
call(arg _1)
```

### Consequences

* `mnkPathArray` never has side-effects now
* `mnkToSlice` never has side-effects now
* exceptional control-flow arising from index and bound checks encoded
  in the MIR, making it accessible for data-flow analysis 
* decision making regarding the initial placement of index and bound
  checks happens in a single place instead of in each code generator
* code generators no longer need to query the local `optBoundChecks`
  option

This is progress towards:
* an optimization pass that eliminates redundant index and bound checks
* making all lvalue expressions free of side-effects 
* all possible exceptional control-flow being encoded in the MIR
* restoring local `.push`/`.pop` support for all checks

### Implementation details

* index and bound checks are represented in the MIR via the new
  internal `mChckIndex` and `mChckBounds` magics
* `isPure` and `isStable` analysis is simplified, now that
  `mnkPathArray` is free of side-effects
* the operand in the array slot of an `nkBracketExpr` may be either an
  rvalue, lvalue, or literal. The new `mirgen.capture` procedure
  implements the logic for correctly capturing the operand  
* `jsgen` now implements code generation for bound checks

**Changes to the VM:**
* the `opcIndexCheck` opcode is introduced; it implements the index
  check (`mChckIndex`)
* the index check logic is removed from all handle-computation-only
  operation; the access checks of memory load/store operations ensures
  that memory safety is not compromised. Operations that write directly
  to a memory location (`opcWrStrIdx` and `opcWrArr`) still need to
  perform the index check 

### MIR semantics

For for both efficiency and implementation simplicity, aliases created
through the `Bind` statement support indirect mutations (e.g.,
`x.a = 1`), but not direct assignments or being passed to a mutable
by-name parameter.

Without this adjustment, `mirgen.capture` would need to know whether a
projection (e.g., `x[1]` or `x.a`) of the alias is used for mutation.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -526,7 +526,7 @@ proc genRd(c: var TCtx, n: PNode; consume=false): Value =
     result = captureInTemp(c, f, consume)
 
 proc capture(c: var TCtx, n: PNode): Value =
-  ## If not not a stable lvalue expression, captures the result of the
+  ## If not a stable lvalue expression, captures the result of the
   ## expression `n` in a temporary.
   ## * rvalue expression are captured in temporaries
   ## * lvalue expressions are captured as non-assignable aliases
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -79,6 +79,9 @@ import
   compiler/front/[
     options
   ],
+  compiler/sem/[
+    ast_analysis
+  ],
   compiler/utils/[
     containers,
     idioms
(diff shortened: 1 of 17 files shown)


compiler/mir/mirgen.nim
```diff
@@ -528,6 +525,29 @@ proc genRd(c: var TCtx, n: PNode; consume=false): Value =
     # either an rvalue or some non-pure lvalue -> capture
     result = captureInTemp(c, f, consume)
 
+proc capture(c: var TCtx, n: PNode): Value =
+  ## If not not a stable lvalue expression, captures the result of the
```
```suggestion
  ## If not a stable lvalue expression, captures the result of the
```

IIUC, mutable (unstable) lvalues require capture, so the second `not` was a typo.

saem (approved):
The mir is really paying off helping consolidate many aspects of the code generators, this being one of them.

I think I found a typo, otherwise this is good to go.

compiler/mir/mirgen.nim
zerbina: Yep, it's a typo -- good catch.

zerbina:
/merge

PR #1077, opened 2024-01-03
vmgen: faster output and improved `openArray` support

zerbina:
## Summary

* significantly reduce the number of run-time copies and temporaries,
  by employing in-place construction and the return-value optimization
* fix constructing array or tuple values that have `openArray`
  elements/fields crashing the VM

This is an intermediate change, aimed at making the performance of
`vmgen` output less sensitive to temporaries created during the MIR
phase.

## Details

The VM already supported in-place construction for the largest part,
but the code generator wasn't making use of it.

For each assignment, a temporary to hold the source operand's value was
allocated, the source expression loaded into the temporary, and then
the temporary was copied (moving is not yet supported by the VM) into
the actual destination location.

All rvalues (calls, conversions, casts, etc.) being assigned to MIR
temporaries first exacerbated the problem.

### VM changes

* add the dedicated `opcReset` operation; it replaces the inefficient
  `opcLdNull` + `opcWrLoc` pattern
* restore the `opcAsgnConst` operation, it's now used for assigning
  constants directly to locations

### `vmgen` changes

* assigning to in-memory locations allways goes through the new
  `putIntoLoc` procedure
* `putIntoLoc` handles the necessary conversion, i.e.,:
  * writing register-stored source values to the in-memory destination
  * converting handle-backed single-location views to an address value
    first
  * ensuring that the `openArray` conversion takes place (implemented
    via `vmgen.asgnOpenArray`)
* similar to `putIntoLoc`, `genAsgnToLocal` handles the necessary
  conversion for locals. Locals may be backed by a register rather than
  an in-memory location, which needs to be accounted for
* the `ToSlice` operation is handled from the assignment's side, as the
  behaviour depends on whether or not the destination is a local
  variable

`openArray` handling now being part of `putIntoLoc`/`genAsgnToLocal`
makes the assignment of `openArray`s work properly in all the different
assignment situations.

Not all code generation paths accounted for in-place construction yet,
with the most notable being:
* literals: simple constants not fitting into a register are placed
  into the destination via `opcAsgnConst`
* calls: if not part of the initial assignment, a temporary location +
  copy afterwards is required for potentially raising calls
  (`cnkCheckedCall`), otherwise modification to the result might be
  observable in case the call raises
* object/set constructions: the destination is reset first, as the
  construction might not overwrite the full location

To signal that initialization of a local is in progress, the
`slotNoValue` slot kind is introduced. It's temporarily used as the
slot kind for a local being initialized.

### Issues

Top-down code generation for assignemnts results in convoluted code and
also prevents code generator optimizations.

A better architecture would be to emit the assigments from the *source
expressions*, by passing along the necessary information about the
destination to `gen` (this is what `cgen` does).

The problem: this requires a much larger refactoring, which would be
redundant with the already planned full-rewrite of `vmgen`.


saem (approved):

zerbina:
Code generation for `cnkStringToCstring` and `cnkCstringToString` did not consider the RVO case, which was caught by the `tcstring.nim` test -- it's fixed now.

PR #1078, opened 2024-01-03
vm: fix spurious access violation errors

zerbina:
## Summary

Fix an internal VM bug that led to seemingly random access violation
errors being reported. The error oftentimes went away when reordering
the code or inserting dummy code somewhere else.

## Details

The following bytecode triggered the error:
```
LdNull r1 ... # setup a memory location and store its handle in r1
FastAsgnComplex r2 r1 # copy the *handle* to r2
LdNullReg r1 ...      # this frees the previously allocated location.
                      # The handle in r2 is now dangling
... # set up a callee in r3
IndCallAsgn r2 r3 1   # reported
```

which could, depending on the state of the register allocator, result
from:
```nim
block:
  let a: seq[int]
  let b: openArray[int] = a
let y: int
let x: int = f()
```

This is because `IndCallAsgn` validates the handle in the destination
register, raising an error if doesn't point to a valid location.
However, the destination register is not required to hold a valid
handle (the referenced memory location is not accessed by
`IndCallAsgn`), so the access check is incorrect -- it's removed.

### Missing access check

While the access check is wrong for calls of procedures implemented in
bytecode, it is required for calls of callbacks, as those are not
guaranteed to validate that the handle referencing the result location
is valid.

If the call's result doesn't fit into a register, and the destination
register is thus required to hold a valid handle, an access check is
now performed.

The missing access check didn't compromise safety for code running at
compile-time or on the VM target, as `vmgen` currently always sets up a
full temporary for call results.

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -1905,8 +1905,8 @@ proc rawExecute(c: var TCtx, t: var VmThread, pc: var int): YieldReason =
           # setup register that will store the result
           if not loadEmptyReg(regs[ra], retType, c.debug[pc], c.memory):
             # allocating the destination location is the responsibility of
-            # ``vmgen``
-            discard
+            # ``vmgen``, but we still have to make sure its accessible
+            checkHandle(regs[ra])
 
         # We have to assume that the callback makes use of its parameters and
         # thus need to validate them here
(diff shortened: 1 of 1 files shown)


zerbina:
A regression test would be nice to have, but implementing a proper one is not easy. While trying to coerce `vmgen` into generating an instruction sequence that triggers the error would work, the test would stop being a regression test as soon as any detail regarding code generation changes.

Alternatively, implementing an assembler and integrating it with the `asm` statement would allow for manually emitting the problematic instruction sequence, but that'll likely take a week of effort, blocking this PR and, with it, further progress on the MIR/backend rework.

saem (approved):

PR #1080, opened 2024-01-05
remove obsolete VM-related reports

zerbina:
## Summary

Remove both the `rvmGlobalError` and `rvmCallingNonRoutine` report.

## Details

* `rvmGlobalError` is not used anymore
* `rvmCallingNonRoutine` was only used by `execProc` (compiler API) for
  reporting about the violation of a parameter's precondition. This
  doesn't warrant a dedicated report kind -- an internal error is used
  instead
* the `VmEventKind` and `AstDiagVmKind` values corresponding to
  `rvmCallingNonRoutine` are removed too

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -947,7 +947,6 @@ type
     adVmUserError
     adVmUnhandledException
     adVmCannotCast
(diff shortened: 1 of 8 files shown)


saem (approved):
That's a lot less stuff. 🎉

PR #1081, opened 2024-01-07
internal: all magic procedures use `mnkMagic`

zerbina:
## Summary

Translate all magics that are implemented by the compiler to `mnkMagic`
nodes during AST -> MIR translation. This simplifies parts of the
backend processing and is preparation for referring to procedures via
an ID during the MIR phase.

### Architectural changes

* all procedures that identify as a magic one but are implemented in 
  user code are translated to `mnkProc` nodes (which uses `PSym`) --
  all others are translated to `mnkMagic` nodes (which used `TMagic`)
* the set of magic procedures that are not compiler-implemented is
  backend-specific, so it needs to be provided from outside `mirgen`

As a consequence, this means that all `mnkProc` nodes in MIR code refer
to "real" procedures (i.e., procedures implemented in user code).

### Implementation details

* `mirgen.generateCode` now accepts a `TranslationConfig` instance
  (which replaces the `set[GenOption]` parameter)
* the set of non-compiler-implemented magics is passed to `mirgen` via
  `TranslationConfig.magicsToKeep`
* the `noMagic` parameter is removed from the `backends.process`
  iterator -- the information is now stored as part of `BackendConfig`
  (via `TranslationConfig`)

### Simplifications

1. the `deps` iterator doesn't need to know about which procedures it
   has to ignore
2. `cgirgen` can treat all `mnkProc` nodes reaching it as references to
   "real" procedure, and thus doesn't have to inspect the `TSym.magic`
   field

Since `cgirgen` now only omits the `cnkHiddenAddr` wrapping for
`mnkMagic`-using procedure calls, `cgen` doesn't have to restore the
`cnkHiddenAddr` for `mNewSeq`, `mSetLengthSeq`, and `mAppendSeqElem` 
magic calls (all three are "real" procedures with the C backend).

### Problem with `vmgen`

The `mNGetSize`, `mNGetType`, and `mNLineInfo` magics are overloaded by
name, which requires `vmgen` to have access to the original symbols
used in the call expression.

For now, the magics are included in the `MagicsToKeep` set, so that
the symbols reach `vmgen`. As a consequence, a VM function table entry
is now unnecessarily reserved for them.

### Misc

`--expandArc` output changes due to `mnkMagic` nodes being rendered
using a name derived from that of the `TMagic` value.


compiler/vm/vmgen.nim
```diff
@@ -170,11 +170,13 @@ const
...
-                   mAbsI, mDotDot}
+  MagicsToKeep* = {mIsolate, mNHint, mNWarning, mNError, mMinI, mMaxI,
+                   mAbsI, mDotDot, mNGetType, mNSizeOf, mNLineInfo}
     ## the set of magics that are kept as normal procedure calls and thus need
-    ## an entry in the function table. For convenience, the ``mNone`` magic is
-    ## also included
+    ## an entry in the function table.
+    # XXX: mNGetType, mNGetSize, and mNLineInfo *are* real magics, but their
+    #      symbols must reach here for disambiguation. This needs to be solved
+    #      differently
```
zerbina, line 189: As a follow-up PR, in order to remove this workaround again, I'm going to move resolving the magic into `mirgen`, by adding an `int` argument representing the picked variant to the calls.

For example:
```nim
getOffset(n)
# would be translated to:
NSizeOf(arg n, arg 2)
```

compiler/mir/mirgen.nim
```diff
@@ -123,6 +123,14 @@ type
     goIsCompileTime ## whether the code is meant to be run at compile-time.
                     ## Affects handling of ``.compileTime`` globals
 
+  TranslationConfig* = object
+     ## Extra configuration for the AST -> MIR translation.
+     options*: set[GenOption]
+     magicsToKeep*: set[TMagic]
+      ## magic procedures that need to referenced via their symbols, either
```
```suggestion
      ## magic procedures that need to be referenced via their symbols, either
```

saem (commented):
`TranslationConfig` nicely decouples things, very cool.

I found one minor typo, the meaning is clear regardless.

PR #1082, opened 2024-01-07
tests: enable the `exceptions` category for all targets

zerbina:
## Summary

Clean up the category and enable it for all targets by default. This
significantly increases test coverage of exceptions for the JS and VM
targets.

## Details

**General changes/cleanup:**
* references to issues like `#1234` are replaced with proper GitHub
  links
* the `targets` key is removed for tests that should be run on all
  targets
* usage of `cmd` is replaced with usage of `matrix` (allowing for
  multi-target testing)
* `write(stdout, ...)` is replaced with `echo` in order to support the
  JS and VM targets
* execution of `static:` blocks is replaced with using the VM target,
  when the test is not specific to compile-time execution
* tests that don't work but should are marked as `knownIssue`

**Specific changes:**
* where not necessary for the test, `getCurrentExceptionMsg` is
  replaced with `getCurrentException().msg`, as the former is not yet
  supported by the VM
* `texceptions2.nim` is removed -- it's redundant with `texception.nim`
* `texception_message_null_byte.nim` is turned into a normal test that
  doesn't rely on self-execution


zerbina:
The `texception_message_null_byte.nim` test works for me locally (Windows), but fails in the CI. It's probably an issue with the null byte and testament's output comparison -- I'll investigate.

saem (approved):
Change itself looks good, testament issue aside.

PR #1083, opened 2024-01-08
testament: support `\0` byte in output comparison

zerbina:
## Summary

Neither the program output nor the comparison string are cut off at the
first null byte anymore.

## Details

Output comparisons in testament used `strutils.contains`, which uses
`strutils.find`, which, by default, uses `c_strstr`.

`c_strstr` operates on null-terminated strings, and thus stopped at the
first occurrence of one, even if the end of the NimSkull string hasn't
been reached yet.

Testament now uses a custom `contains` implementation based on C's
`memcmp`, so that the `\0` byte doesn't terminate the search.

### DIFF
--- a/testament/testament.nim
+++ b/testament/testament.nim
@@ -10,12 +10,15 @@
 ## Testament runs tests for the compiler.
 
 import std/[
-  strutils, pegs, os, osproc, streams, json, parseopt, browsers,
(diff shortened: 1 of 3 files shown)


zerbina:
While `strutils.find` could have been fixed instead, it's a subtle breaking change, and I don't want to track down and verify all usages of `find` at this time.

saem (approved):

PR #1085, opened 2024-01-11
backends: refactor the `process` iterator

zerbina:
## Summary

Rework the procedure processing of the `backend.process` iterator such
that it's easier to reason about, extend, and adjust.

## Details

The core part of the `backend.process` iterator (which implements the
common backend pre-processing) was so far structured around procedures
being the unit of work. That is, procedures are added to a queue, from
which they're then picked and processed (which might result in new
procedures being queued).

Progress events are reported by `yield`ing a `BackendEvent` back to the
caller. In order to reduce the amount duplicated *written* code,
templates have to be used (so that `yield` works).

The problem with this architecture is that:
* it's rigid; introducing new events or changing their order is not
  easily done
* the templates obfuscate control flow and make the `process` iterator
  hard to reason about for the reader
* there's a lot of physical code duplication (after template expansion)

### New architecture

* instead of around procedures, the main loop is now structured around
  *steps*
* processing of a procedure is split into multiple steps (pre-
  processing, translation to MIR, etc.), with each step placed into the
  queue
* processing a constant (currently only scanning the expression) is a
  dedicated step too, preparing for translating constants to a MIR
  representation
* another way to look at a *step* it is that it represents a
  continuation

### `process` API changes

* a `bekDiscovered` event is reported for each discovered entity,
  instead of only for batches. This is simpler to implement for the
  event processor(s)
* the `bekConstant` event is introduced; it is reported once a constant
  was processed and all its dependencies are available

### Implementation details

* `ProcedureIter` is renamed to `WorkQueue`
*  a step is represented by the new `WorkItem` type
* `preprocess` only runs `transf` and emits follow-up steps; it 
* the to-MIR translation is part of the public `process` procedure now;
  the procedure is renamed to `translate`
* the `reportProgress` template is turned into the standalone
  `pushProgress` procedure
* the `Procedure` type is removed; it's unused now
* `jsbackend` and `cbackend` emit the definition for constant in
  response to the `bekConstant` event, not when the constant is first
  discovered

Since reporting and queuing pending entities (implemented by the
`flush` iterator) happens in between most steps, no dedicated step is
added for this -- the `flush` iterator is simply run after the
dispatcher.

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -49,20 +49,6 @@ type
     tree*: MirTree
     source*: SourceMap
 
-  Procedure* = object
-    sym*: PSym
-    case isImported*: bool ## wether it's a .dynlib procedure
-    of false:
-      body*: MirFragment
-        ## the procedure's body
-
-      globals*: seq[PNode]
-        ## the unprocessed identdefs of globals defined as part of the
-        ## procedure's body. Due to how ``transf`` handles inlining, this
-        ## list can contain duplicates
-    of true:
-      discard
-
   BackendConfig* = object
     ## Configuration state altering the operation of the ``process``
     ## iterator.
@@ -72,11 +58,6 @@ type
     noImported*: bool
       ## if ``true``, indicates that a procedure with a body should not be
       ## treated as imported, even if it's marked as such
-
-  ProcedureIter = object
-    queued: Deque[tuple[prc: PSym, fr: FileIndex]]
-      ## procedures that are queued for code-generation
-    config: BackendConfig
 
   Queue*[T] = object
     ## Combines a sequence of items with a "read" cursor.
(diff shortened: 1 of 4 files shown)


zerbina:
One goal of rewriting the C code generator is to make `.inline` procedure management in `cbackend` work without help from `backends`. This would eliminate the need for the "queued from" tracking, which complicates the `process` iterator and its surroundings by an significant amount.

Beyond that, extracting globals and threadvars from procedure bodies, as well as producing the init/teardown fragment for them, should happen prior to all the backends processing, which would simplify the `backends` module further.

---

It's possible that these changes might result in the architecture as implemented here to become unnecessary again, but for now, splitting procedure processing into discrete steps makes the processing simpler.

saem (approved):

PR #1086, opened 2024-01-12
omit index checks when no index error is possible

zerbina:
## Summary

If all possible index operand values are valid indices, no index check
is required at run-time. The C code generator already implemented
this, but only for zero-based `array`s.

The optimization is now part of `mirgen`, making it available to all
backends and with all `array` types.


saem (approved):
That's a nice little optimization.

PR #1091, opened 2024-01-14
sem: `semFor` no longer mutates input AST

saem:
## Summary

Stop mutating the input AST when semantically analysing  `for`  loops in 
`semFor` .

## Details

Previously,  `semFor`  would mutate the input AST parameter  `n` ,
resulting in partial contribution to call analysis corruption during
overload resolution. Now  `semFor`  does a cursory examination of the 
`for`  loop and dispatches to either  `semForVars`  or  `semForFields` 
based on whether it's a standard  `for`  over non-type values or over a
type's fields, respectively.

As part of this change and issue was discovered in  `semAfterMacroCall`
, which was not handling errors passed via  `tyError` , this has been
updated by having it use  `semTypeNode2` . This meant a minor change in
error message for meta-routines returning a type ( `typedesc` ), from 
`expression has no type`  to 
`type expected, but expression has no type` , which is somewhat more
informative; tests have been updated accordingly.

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -735,6 +735,8 @@ proc semAfterMacroCall(c: PContext, call, macroResult: PNode,
         let err = newError(c.config, result, PAstDiag(kind: adSemExpressionHasNoType))
         localReport(c.config, err)
         result = newSymNode(errorSym(c, result, err))
+      elif typ.kind == tyError and typ.n != nil and typ.n.kind == nkError:
+        result = typ.n
       else:
         result.typ = makeTypeDesc(c, typ)
     else:
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -122,6 +122,7 @@ proc fitNode(c: PContext, formal: PType, arg: PNode; info: TLineInfo): PNode
 proc changeType(c: PContext; n: PNode, newType: PType, check: bool): PNode
 
 proc semTypeNode(c: PContext, n: PNode, prev: PType): PType
+proc semTypeNode2(c: PContext, n: PNode, prev: PType): PNode
 proc semStmt(c: PContext, n: PNode; flags: TExprFlags): PNode
 proc semOpAux(c: PContext, n: PNode): bool
 proc semParamList(c: PContext, n, genericParams: PNode, kind: TSymKind): PType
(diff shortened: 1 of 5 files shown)


compiler/sem/semstmts.nim
```diff
@@ -1599,34 +1599,36 @@ proc semFor(c: PContext, n: PNode; flags: TExprFlags): PNode =
...
-        n[^2] = implicitIterator(c, "items", n[^2])
+        result[^2] = implicitIterator(c, "items", result[^2])
       elif n.len == 4:
-        n[^2] = implicitIterator(c, "pairs", n[^2])
+        result[^2] = implicitIterator(c, "pairs", result[^2])
       else:
-        n[^2] = c.config.newError(n[^2], PAstDiag(kind: adSemForExpectedIterator))
-      hasError = n[^2].isError or hasError
-    result = semForVars(c, n, flags)
+        result[^2] = c.config.newError(n[^2],
```
```suggestion
        result[^2] = c.config.newError(call,
```
This should restore the previous error behaviour (the `tfor_ill_formed.nim` test failure).

zerbina (commented):
For improved clarity and an, in my opinion, better flow, I think `semFor` should not set up the production AST (i.e., `copyNodeWithKids`) but instead leave this to `semForVars` and `semForFields`. The structure I'm envisioning is:
```nim
var call = semExprNoDeref(c, n[2], ...)
if call.kind == ...:
  ...
if is_for_fields:
  result = semForFields(c, n, call)
else:
  # the `implicitIterator` patching goes here
  result = semForVars(c, n, call)
```

This way, `semFor` is only responsible for dispatching to the correct analysis procedure, there are less `result` assignments, and the current `semForFields`-related compiler crash should be fixed too.

In any case, I think this restructuring is non-blocking -- it can also be applied separately.

saem:
> For improved clarity and an, in my opinion, better flow, I think `semFor` should not set up the production AST (i.e., `copyNodeWithKids`) but instead leave this to `semForVars` and `semForFields`. The structure I'm envisioning is

That approach makes sense, it became easier to see how dispatch is predicated on `call` after the change thus far for me. I'll give that a go prior to finalizing this.

saem:
The current failure is due to `semTypeOf2` and type evaluation in general relies upon mutating the passed in AST if we want `nkError` nodes to not be dropped on the floor (the produciton isn't incorporated back into the queried AST).

compiler/sem/sem.nim
```diff
@@ -735,6 +735,8 @@ proc semAfterMacroCall(c: PContext, call, macroResult: PNode,
         let err = newError(c.config, result, PAstDiag(kind: adSemExpressionHasNoType))
         localReport(c.config, err)
         result = newSymNode(errorSym(c, result, err))
+      elif typ.kind == tyError and typ.n != nil and typ.n.kind == nkError:
+        result = typ.n
```
saem: Not particularly happy about this, as there are plenty of other usages of `semTypeNode` that could also have the same sort of issue. It's not clear if `semTypeNode` should do a `localReport` or callsites should check things. They all seem flawed in their own way.

zerbina: I think you could use `semTypeNode2`, which is meant for the type-AST/normal-AST edge, for the whole `tyTypeDesc` branch, like so:
```nim
      if result.kind == nkStmtList: result.transitionSonsKind(nkStmtListType)
      result = semTypeNode2(c, result)
      if result.kind != nkError:
        result.typ = makeTypeDesc(c, result.typ)
```
The error for the "no type" case would be a different one, `adSemTypeExpected` vs. `adSemExpressionHasNoType`, although I'd argue that the former is more correct, since here a *type expression* is expected rather than just a *typed expression*.

`semTypeNode2` seems to be missing the `typ.n != nil` check, however -- that'd need to be fixed.

saem: Thanks, that's a fair bit cleaner. I can fix any tests that break due to message changes, and I agree that the message is more correct.

saem:
Switched to using `semTypeNode2` and also went the dispatcher route for `semFor`. Things were clean on my end so just need CI to confirm at this point.

PR #1092, opened 2024-01-14
debug: improve compiler tracing line info

saem:
## Summary

Improve compiler tracing line info, previously wrong line information
would be tracked uselessly pointing back to  `debugutils` . Now line
information is accurate, although there is a minor regression for some
inter-trace stacktrace printing emitting the entire stack instead of the
interval.

## Details

In addition to the above, tracing for call analysis, procs using 
`TCandidate`  now emit a  `to`  node for tracking mutations to input
AST.

Coverage for a few semantic analysis procs have been added as these
changes were the result of an attempt at overload call analysis
debugging.

### DIFF
--- a/compiler/utils/debugutils.nim
+++ b/compiler/utils/debugutils.nim
@@ -336,6 +336,8 @@ proc generateOutput(conf: ConfigRef, r: CompilerTrace): string =
               result.add render(s.candidate.calleeSym)
             field("call")
             result.add render(s.candidate.call)
+            field("to node")
+            result.add render(s.node)
   of compilerTraceLine:
     let ind = repeat("  ", r.ctraceData.level)
     var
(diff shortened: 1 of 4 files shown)


compiler/utils/debugutils.nim
```diff
@@ -493,7 +495,7 @@ template addInNimDebugUtilsAux(conf: ConfigRef; prcname: string;
         if indentLevel != 0: # print a delta stack
           # try to print only the part of the stacktrace since the last time,
           # this is done by looking for any previous calls in `debugUtilsStack`
-          {.line.}: # stops the template showing up in the StackTraceEntries
+          {.line:instantiationInfo(-2, true).}: # stops the template showing up in the StackTraceEntries
```
zerbina, line 498:
```suggestion
          {.line: instLoc(-3).}: # stops the template showing up in the StackTraceEntries
```
A soft suggestion -- it's a bit shorter and also respects the `excessiveStackTrace` option. The depth should be correct, but I haven't tried it.

zerbina (commented):
Looks good to me. For fixing/preventing the stack-trace regression, I think changing the `{.line.}` in line 531 to `{.line: instLoc(-3).}` should be enough.

compiler/utils/debugutils.nim
saem: I tried it and it didn't work (kept showing `debugutils` instead of the actual trace), I also ran into issues with `excessiveStackTrace` entires being `false`.

The various `line` pragma need to be consolidated, I think they're interfering with each other, I'll do that in a future PR.

PR #1093, opened 2024-01-15
mir: introduce `MirBody`

zerbina:
## Summary

The `MirBody` type represents a self-contained piece of MIR code,
usually the body of a procedure. It replaces passing around loose
`MirTree` and `SourceMap` values (a `MirBody` is now used instead).

## Details

* both `MirFragment` and `TreeWithSource` are also replaced by
  `MirBody`
* for more flexibility with imports, the type is not made part of
  `mirtrees` but instead has its own module: `mirbodies`
* the `backends.append` routine was unused; it's removed

`MirBody` serves the same purpose for the MIR as `cgir.Body` does for
the `CgNode` IR.

### DIFF
--- a/compiler/mir/mirbodies.nim
+++ b/compiler/mir/mirbodies.nim
@@ -1,0 +1,32 @@
+## Implements the ``MirBody`` type and basic routines for querying and
+## modifying it.
+
+import
+  compiler/ast/[
+    ast_types
+  ],
+  compiler/mir/[
+    mirtrees,
+    sourcemaps
+  ]
+
+type
+  MirBody* = object
+    ## A ``MirBody`` represents a self-contained piece of MIR code. This can
+    ## either be:
+    ## - the full body of a procedure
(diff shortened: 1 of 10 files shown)


compiler/mir/mirbodies.nim
```diff
@@ -0,0 +1,32 @@
...
+    ## A ``MirBody`` represents a self-contained piece of MIR code. This can
+    ## either be:
+    ## - the full body of a procedure
+    ## - the partial body a procedure
+    ## - a standalone statement/expression (currently supported for compile
+    ##   time code execution)
+    ##
+    ## In each case, ``MirBody`` stores all the local data referenced and
+    ## needed by the body's MIR code. It also store additional information
+    ## associated with a body, such as how far the lowering is along.
```
zerbina, line 24: It doesn't do so at the moment, but it's going to eventually.

saem (approved):

PR #1094, opened 2024-01-15
mir: separate type checks from object conversion

zerbina:
## Summary

* placement of object conversion checks is now the responsibility of
  the AST -> MIR translation
* destructors are now called when the only exceptional control-flow
  within a scope comes from object-conversion checks 
* exceptional control-flow from object-conversion checks is now visible
  to data-flow analysis
* object conversion checks are now available with the JS backend

## Details

An object conversion, like:
```nim
var x: ref Base
discard (ref Sub)(x)
```

is now translated to, assuming that object conversion checks are
enabled:

```
def_cursor _1: ref Base = x
def _2: bool = isNil(arg _1)
def _3: bool = not(arg _2)
if _3:
  chckObj(arg _1, arg type(Sub)) (raises)
discard x.(ref Sub)
```

As a consequence:
*  `mnkPathConv` expressions never have side-effects now
* the exceptional control-flow is visible to data-flow analysis
* placement of object checks works the same across all backends

### `mirgen` implementation

* object conversion checks are only enabled for non-pure `ref object`
  and `ptr object` types
* this matches the behaviour of `cgen`
* as an optimization, up-/down-conversion sequences (e.g.,
  `x.Base.Sub1.Base.Sub2`) are collapses into a single `mnkPathConv`
  operation, with, at most, one check emitted 
* object conversion are special projections that, compared to the other
  projections, do require using a `BindMut` for `x` in cases such as
  `A(x) = nil`. At the moment, this requires an awkward special-casing
  of `nkObjDownConv`/`nkObjUpConv` in `genLvalueOperand`

### Implementation in the code generators

* `cgen` re-uses most of its existing converion check implementation
* `jsgen` implements `mChckObj` with a call to the pre-existing
  `chckObj` run-time procedure
* for the VM, the conversion checks are moved from the `opcObjConv`
  operation into the dedicated `opcCheckObj` operation. `vmgen` uses
  the latter to implement `mChckObj`

### Future work

All MIR lvalue expressions are now free of side effects, allowing for
various MIR pass improvements. So far, passes had to assume that all
lvalue expression potentially raise exceptions.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -545,7 +545,7 @@ proc capture(c: var TCtx, n: PNode; sink=false): Value =
       captureInTemp(c, f, sink=false)
     else:
       # if used in a sink context, the alias must support potentially
-      # destructive moves throught it, which requires a ``BindMut``
+      # destructive moves through it, which requires a ``BindMut``
       captureName(c, f, mutable=sink)
 
 proc genOperand(c: var TCtx, n: PNode; sink = false) =
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -753,9 +753,13 @@ proc genVariantAccess(c: var TCtx, n: PNode) =
             c.use x
 
 proc genObjConv(c: var TCtx, n: PNode, sink, dest: bool) =
-  ## Given an ``nkObjDownConv`` or ``nkObjUpConv`` AST, unwraps all . Only one
-  ## ``mnkPathConv`` is emitted, back-and-forth conversions are all dropped
-  ## (but checks are still generated for them).
+  ## Given an ``nkObjDownConv`` or ``nkObjUpConv`` AST, folding away all
+  ## immediately nested up- and down-conversion nodes. That is, only one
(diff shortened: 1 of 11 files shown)


zerbina:
I've originally used `buildMagicCall` instead of `buildDefectMagicCall`, which meant that the exceptional control-flow wasn't actually visible to data-flow analysis, hence the test failure.

The test failure now is because of a `vmgen` bug with lvalue conversion. Those now appear in argument positions, which `vmgen` doesn't properly account for. I'll make a separate PR with the fix.

compiler/mir/mirgen.nim
```diff
@@ -545,7 +544,9 @@ proc capture(c: var TCtx, n: PNode): Value =
       # don't create an alias for openArrays, a copy of the view suffices
       captureInTemp(c, f, sink=false)
     else:
-      captureName(c, f, mutable=false)
+      # if used in a sink context, the alias must support potentially
+      # destructive moves throught it, which requires a ``BindMut``
```
```suggestion
      # destructive moves through it, which requires a ``BindMut``
```

saem (approved):
LGTM

One very minor typo, it's still clear what is meant.

PR #1099, opened 2024-01-17
fix(sem): don't fold away `proc` casts

zerbina:
## Summary

Don't fold away `cast`s to `proc` type where the operand is a
constant expression -- the backend might need to generate code for
the cast operation.

This fixes C compiler errors when casting `proc` types and using a C
compiler that disallows incompatible function pointer types.

## Details

Folding the `nkCast` allowed casting procedure literals to different
procedural types with the VM backend, even though `cast` is not
implemented for `tyProc` there.

`vmgen` now implements to-`tyProc` casts for literals, by loading the
procedure literal without modifications. This is equivalent to what
happened previously.

### DIFF
--- a/compiler/sem/semfold.nim
+++ b/compiler/sem/semfold.nim
@@ -905,7 +905,7 @@ proc foldConstExprAux(m: PSym, n: PNode, idgen: IdGenerator, g: ModuleGraph): Fo
   # are not folded if they have a comment in the first position
   let exprIsPointerCast = n.kind in {nkCast, nkConv, nkHiddenStdConv} and
                           n.typ != nil and
-                          n.typ.kind == tyPointer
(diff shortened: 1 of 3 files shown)


zerbina:
As expected, `vmgen` now reports an error because it doesn't support `proc` to `proc` casts. Fortunately, the cast-literal-into-type can be trivially implemented in the code generator. I'll fix it tomorrow.

zerbina:
I've adjusted `vmgen` already, fixing the test failures.

A somewhat cleaner solution would be to treat the cast as a non-constant expression in `getConstExpr`, but that would be a far more invasive change (and `const c = cast[Proc](1)` would stop working), so I decided against it for the time being.

saem (approved):
FYI: @alaviss

PR #1101, opened 2024-01-17
koch, niminst: move csources bundle tag location

alaviss:
## Summary

Previously, we stored the csources bundle tag using the confusing name
"csources-release" and stored within the build/csources folder.

However, this caused some undesirable issues:

- If this file become stored within the csources git repository, it
will
  disable all csources versioning logic in koch.py.
- The old name "csources-release" did not adequately describe what the
  file is for.

This commit move the bundle tag to `build/csources-bundled-version`,
which now accurately describe the intent of the file and move the
ownership of "bundled" state away from csources.

## Details
This was brought about by
https://github.com/nim-works/csources_v1/pull/3 introducing 
`csources-release`  into csources_v1 repository, disabling koch
automatic csources version management.

A few notes:

* Moving to  `build/`  means that this file is now generated during
archive generation. To prevent that from silently turning off koch
csources logic for developers working on this, I opted to move the
archive output to  `build/archive/` .

### DIFF
--- a/koch.py
+++ b/koch.py
@@ -72,8 +72,8 @@ class Bootstrap:
     # The location of build configuration
     ConfigPath: ClassVar = NimSource / "config" / "build_config.txt"
 
-    # The location of bootstrap source release file
-    SourceReleasePath: ClassVar = Source / "csources-release"
+    # The location of bootstrap source bundle state
+    SourceReleasePath: ClassVar = NimSource / "build" / "csources-bundled-version"
 
     def __init__(self) -> None:
         """Load configuration and initialize the object"""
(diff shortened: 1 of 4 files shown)


koch.py
```diff
@@ -73,7 +73,7 @@ class Bootstrap:
     ConfigPath: ClassVar = NimSource / "config" / "build_config.txt"
 
     # The location of bootstrap source release file
```
saem: This comment should be updated too?

saem (approved):
Minor update might be required in a comment.

Question is which csources should be copied now if we want to update them in the future?

alaviss:
> Minor update might be required in a comment.
> 
> Question is which csources should be copied now if we want to update them in the future?

you can just copy it straight from a source tarball after this since we pruned the troublesome file from it

alaviss:
/merge

Fixed the patterns used by repro test

PR #1102, opened 2024-01-18
sem: preserve untyped arguments in call dispatch

saem:
## Summary

Untyped arguments are no longer overwritten by their typed variety
during overload resolution (call corruption), fixing overload resolution
where overload candidates with untyped parameters were tested after ones
with typed parameters in the same argument positions.

## Details

Previously, during overload resolution the original call ast would be
updated incrementally with semantically analysed arguments as matches
were attempted. This was problematic when a typed formal parameter was
checked before an  `untyped`  parameter. This would result in the
argument being analysed and overwriting the original untyped AST in the
call.

Now we make a shallow copy, with the same child nodes, of the call AST,
which allows us to keep a reference to the original untyped AST. This is
then used whenever an overload resolution callee candidate has an
untyped formal parameter.

A regression test has been added to avoid this happening in the future.

### DIFF
--- a/compiler/sem/sigmatch.nim
+++ b/compiler/sem/sigmatch.nim
@@ -2826,8 +2826,10 @@ proc matchesAux(c: PContext, n, nOrig: PNode, m: var TCandidate, marker: var Int
     f = if m.callee.kind != tyGenericBody: 1
         else: 0
       ## index to iterate over formal parameters
+    operand: PNode
+      ## current prepared operand/argument 
     arg: PNode
-      ## current prepared argument
+      ## current prepared and param type matched argument
     formalLen = m.callee.n.len
     formal = if formalLen > 1: m.callee.n[1].sym else: nil
       ## current routine parameter
@@ -2844,6 +2846,8 @@ proc matchesAux(c: PContext, n, nOrig: PNode, m: var TCandidate, marker: var Int
       
       formal = m.callee.n[f].sym
       incl(marker, formal.position)
+
+      operand = n[a] # initialize to current arg in case of early `noMatch`
 
       case n[a].kind
       of nkHiddenStdConv:
@@ -2911,9 +2915,7 @@ proc matchesAux(c: PContext, n, nOrig: PNode, m: var TCandidate, marker: var Int
 
         arg = n[a][1]
       else:
-        let operand = prepareOperand(c, formal.typ, n[a][1], nOrig[a][1])
-          ## analysed operand, if it's an error the issue is based on the
-          ## formal type and not the actual callsite operand.
+        operand = prepareOperand(c, formal.typ, n[a][1], nOrig[a][1])
 
         case operand.kind
         of nkError:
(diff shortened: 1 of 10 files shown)


compiler/sem/sigmatch.nim
```diff
@@ -41,7 +41,8 @@ import
   ],
   compiler/utils/[
     debugutils,
-    idioms
+    idioms,
+    astrepr
```
```suggestion
    idioms
```
This looks to be a leftover from debugging?

zerbina:
The culprit here are the `n[a] = operand` assignments.

Given the following:
```nim
template call(a: int)
template call(a: untyped)
template call(a: float)

call((let v = 1.0; v)) # expr is of type float
```
What happens is:
1. `call(int)` is matched against
2. `expr` is typed (`prepareOperand`) and the result cached in `n[1]`
3. no match is found
4. `call(untyped)` is matched against
5. `prepareOperand` returns the original, untyped AST and the previously cached typed node in `n[1]` is overwritten with the untyped AST
6. the overload matches
7. `call(float)` is matched against
8. `n[1]` (the first argument) has no type, so `prepareOperand` runs semantic analysis again. `v` was already added to the symbol table, so analysis reports an error

In short, the result of `prepareOperand` must only be cached if the formal type is not `tyUntyped`.

saem:
> The culprit here are the `n[a] = operand` assignments.
> 
> Given the following:
> 
> ```nim
> template call(a: int)
> template call(a: untyped)
> template call(a: float)
> 
> call((let v = 1.0; v)) # expr is of type float
> ```
> 
> What happens is:
> 
> 1. `call(int)` is matched against
> 2. `expr` is typed (`prepareOperand`) and the result cached in `n[1]`
> 3. no match is found
> 4. `call(untyped)` is matched against
> 5. `prepareOperand` returns the original, untyped AST and the previously cached typed node in `n[1]` is overwritten with the untyped AST
> 6. the overload matches
> 7. `call(float)` is matched against
> 8. `n[1]` (the first argument) has no type, so `prepareOperand` runs semantic analysis again. `v` was already added to the symbol table, so analysis reports an error
> 
> In short, the result of `prepareOperand` must only be cached if the formal type is not `tyUntyped`.

Thanks for the analysis, but I think the net result should be that after the overload is chosen, the arguments that are untyped should remain as such in `n`; which means I either need to:
- go back and ensure that's the case after choosing an overload
- have an untyped (`nOrig`) and a typed (`???`) stores for arguments, and then swap them in and out of `n` as required

zerbina:
> * have an untyped (`nOrig`) and a typed (`???`) stores for arguments, and then swap them in and out of `n` as required

But doesn't `n` already represent the typed store for the argument? Previous overloads are not affected by changes to `n`, since each match uses its own AST (in `m.call`).

saem:
> > * have an untyped (`nOrig`) and a typed (`???`) stores for arguments, and then swap them in and out of `n` as required
> 
> But doesn't `n` already represent the typed store for the argument? Previous overloads are not affected by changes to `n`, since each match uses its own AST (in `m.call`).

I thought about it some more and you're right. If anything ends up using `n` after matching instead of grabbing the result out of `m.call` then that's a bug I should fix. 😆 

Resulting in:

- `nOrig` stores untyped AST
- `n` stores typed AST
- `m.call` the final call, some admixture of the above as required by the call

zerbina (approved):
Very nice. Also, `untyped` parameters already support constraints, so making `checkConstraint` use the untyped AST is already necessary.

PR #1104, opened 2024-01-18
compiler, ci: remove acyclic and run leak tests

alaviss:
## Summary
Local tests have shown that the compiler does create cycles and leak
memory during compilation. As such we are dropping all acyclic
annotations and prepare the stage to either live without them or have
them reintroduced over time.

## Details
Aside from dropping acyclic annotations, this commit prepares the
compiler for future work on memory optimizations with these changes:

- Cycle collection is disabled for the compiler itself. This is because
the compilation command is not a long-running task and we value speed
over memory usage here.

- nimsuggest is set to collect cycles only after running a command. This
should keep memory usage low during extended usage while allowing the
compilation phases to stay at top speed.

- A CI pipeline has been setup to test the compiler using LeakSanitizer,
making sure that leaks are caught before they are introduced.

Some performance numbers due to these changes:

Benchmarked via: 
`hyperfine -- "bin/nim c --forceBuild --compileOnly --verbosity:0 --hints:off --warnings:off compiler/nim.nim"`

            Time             Ratio
    Before  9.155s ± 0.044s  1.00
    After   9.599s ± 0.039s  1.05

### DIFF
--- a/.github/workflows/leak.yml
+++ b/.github/workflows/leak.yml
@@ -8,9 +8,7 @@ on:
 jobs:
   leaktest:
     # Skip this for PRs
-    #
-    # FIXME: uncomment this before merge
-    # if: github.event_name != 'pull_request'
+    if: github.event_name != 'pull_request'
 
     name: Compiler and tools leak test
     runs-on: ubuntu-latest
--- a/.github/workflows/leak.yml
+++ b/.github/workflows/leak.yml
@@ -1,0 +1,42 @@
+name: Test compiler and tools for memory leaks
+
+on:
+  # This is for passing required checks
+  pull_request:
+  merge_group:
+
+jobs:
+  leaktest:
+    # Skip this for PRs
+    #
+    # FIXME: uncomment this before merge
+    # if: github.event_name != 'pull_request'
+
+    name: Compiler and tools leak test
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          filter: tree:0
+
+      - name: Enable annotations
+        run: echo "::add-matcher::.github/nim-problem-matcher.json"
+
+      - name: Build release binaries with LeakSanitizer
+        run: |
+          ./koch.py all-strict \
+            -d:useMalloc \
+            --passC:-fsanitize=leak \
+            --passC:"-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer" \
+            --passL:-fsanitize=leak \
+            --debuginfo \
+            --linedir
+
+      - name: Run tools tests
+        run: ./koch.py testTools
+
+      # NOTE: We don't run the full test suite here as self-bootstrap
+      # and tools testsuite can be considered enough to rat out memory leaks
+      # within the compiler itself.
--- a/compiler/ast/ast_parsed_types.nim
+++ b/compiler/ast/ast_parsed_types.nim
@@ -226,7 +226,7 @@ type
     base*: NumericalBase
     literal*: string
 
-  ParsedNodeData*{.final, acyclic.} = object
+  ParsedNodeData*{.final.} = object
     # TODO: replace token fields with indexing into a token sequence, this
     #       should also address line info tracking.
     comment*: string       # TODO: replace with an index into a token stream
--- a/compiler/ast/ast_parsed_types.nim
+++ b/compiler/ast/ast_parsed_types.nim
@@ -226,6 +226,7 @@ type
     base*: NumericalBase
     literal*: string
 
+  # TODO: This type should be marked acyclic
   ParsedNodeData*{.final.} = object
     # TODO: replace token fields with indexing into a token sequence, this
     #       should also address line info tracking.
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -901,7 +901,7 @@ type
 
 
 type
-  TIdObj* {.acyclic.} = object of RootObj
+  TIdObj* = object of RootObj
     itemId*: ItemId
   PIdObj* = ref TIdObj
 
@@ -1256,7 +1256,7 @@ type
     adSemDeprecatedCompilerOptArg   # warning promoted to error
 
   PAstDiag* = ref TAstDiag
-  TAstDiag* {.acyclic.} = object
+  TAstDiag* = object
     # xxx: consider splitting storage type vs message
     # xxx: consider breaking up diag into smaller types
     # xxx: try to shrink the int/int128 etc types for counts/ordinals
@@ -1561,7 +1561,7 @@ type
           adSemDefNameSymIllformedAst:
         discard
 
-  TNode*{.final, acyclic.} = object # on a 32bit machine, this takes 32 bytes
+  TNode*{.final.} = object # on a 32bit machine, this takes 32 bytes
                                     # on a 64bit machine, this takes 40 bytes
     typ*: PType
     id*: NodeId  # placed after `typ` field to save space due to field alignment
@@ -1642,7 +1642,7 @@ type
 
   PInstantiation* = ref TInstantiation
 
-  TScope* {.acyclic.} = object
+  TScope* = object
     depthLevel*: int
     symbols*: TStrTable
     parent*: PScope
@@ -1650,7 +1650,7 @@ type
 
   PScope* = ref TScope
 
-  TSym* {.acyclic.} = object of TIdObj # Keep in sync with PackedSym
+  TSym* = object of TIdObj # Keep in sync with PackedSym
     ## proc and type instantiations are cached in the generic symbol
     case kind*: TSymKind
     of routineKinds - {skMacro}:
@@ -1728,7 +1728,7 @@ type
     attachedTrace,
     attachedDeepCopy
 
-  TType* {.acyclic.} = object of TIdObj
+  TType*  = object of TIdObj
     ## types are identical only if they have the same id; there may be multiple
     ## copies of a type in memory! Keep in sync with PackedType
     kind*: TTypeKind          ## kind of type
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1642,6 +1642,7 @@ type
 
   PInstantiation* = ref TInstantiation
 
+  # TODO: This type should be marked acyclic
   TScope* = object
     depthLevel*: int
     symbols*: TStrTable
--- a/compiler/ast/idents.nim
+++ b/compiler/ast/idents.nim
@@ -22,6 +22,7 @@ import
 
 type
   PIdent* = ref TIdent
+  # TODO: This type should be marked acyclic
   TIdent* = object
     id*: int ## unique id; use this for comparisons and not the pointers
     s*: string
--- a/compiler/ast/idents.nim
+++ b/compiler/ast/idents.nim
@@ -22,7 +22,7 @@ import
 
 type
   PIdent* = ref TIdent
-  TIdent*{.acyclic.} = object
+  TIdent* = object
     id*: int ## unique id; use this for comparisons and not the pointers
     s*: string
     next*: PIdent ## for hash-table chaining
--- a/compiler/backend/cgir.nim
+++ b/compiler/backend/cgir.nim
@@ -169,7 +169,7 @@ type
   LocalId* = distinct uint32
     ## Identifies a local within a procedure.
 
-  CgNode* = ref object
+  CgNode* {.acyclic.} = ref object
     ## A node in the tree structure representing code during the code
     ## generation stage. The "CG" prefix is short for "code generation".
     info*: TLineInfo
--- a/compiler/backend/cgir.nim
+++ b/compiler/backend/cgir.nim
@@ -169,7 +169,7 @@ type
   LocalId* = distinct uint32
     ## Identifies a local within a procedure.
 
-  CgNode* {.acyclic.} = ref object
+  CgNode* = ref object
     ## A node in the tree structure representing code during the code
     ## generation stage. The "CG" prefix is short for "code generation".
     info*: TLineInfo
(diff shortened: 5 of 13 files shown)


alaviss:
As an aside, here are the numbers with ORC cycle collector enabled for the compiler.

```
        Time             Ratio
Before  9.155s  ± 0.044s  1.00
After   9.599s  ± 0.039s  1.05
ORC on  11.417s ± 0.150s  1.25
```

compiler/ast/ast_parsed_types.nim
```diff
@@ -226,7 +226,7 @@ type
     base*: NumericalBase
     literal*: string
 
-  ParsedNodeData*{.final, acyclic.} = object
+  ParsedNodeData*{.final.} = object
```
zerbina, line 230:
```suggestion
  ParsedNodeData*{.final, acyclic.} = object
```

compiler/ast/ast_types.nim
```diff
@@ -1638,15 +1638,15 @@ type
 
   PInstantiation* = ref TInstantiation
 
-  TScope* {.acyclic.} = object
+  TScope* = object
```
zerbina, line 1646:
```suggestion
  TScope* {.acyclic.} = object
```

compiler/ast/idents.nim
```diff
@@ -22,7 +22,7 @@ import
 
 type
   PIdent* = ref TIdent
-  TIdent*{.acyclic.} = object
+  TIdent* = object
```
zerbina, line 26:
```suggestion
  TIdent*{.acyclic.} = object
```

compiler/backend/cgir.nim
```diff
@@ -169,7 +169,7 @@ type
   LocalId* = distinct uint32
     ## Identifies a local within a procedure.
 
-  CgNode* {.acyclic.} = ref object
+  CgNode* = ref object
```
zerbina:
```suggestion
  CgNode* {.acyclic.} = ref object
```

compiler/sem/semdata.nim
```diff
@@ -59,7 +59,7 @@ type
 
   POptionEntry* = ref TOptionEntry
   PProcCon* = ref TProcCon
-  TProcCon* {.acyclic.} = object ## procedure context; also used for top-level
+  TProcCon* = object ## procedure context; also used for top-level
```
zerbina, line 63:
```suggestion
  TProcCon* {.acyclic.} = object ## procedure context; also used for top-level
```

compiler/sem/semtypinst.nim
```diff
@@ -121,7 +121,7 @@ proc cacheTypeInst(c: PContext; inst: PType) =
   addToGenericCache(c, gt.sym, inst)
 
 type
-  LayeredIdTable* {.acyclic.} = ref object
+  LayeredIdTable* = ref object
```
zerbina, line 125:
```suggestion
  LayeredIdTable* {.acyclic.} = ref object
```

compiler/utils/btrees.nim
```diff
@@ -16,7 +16,7 @@ const
   Mhalf = M div 2
 
 type
-  Node[Key, Val] {.acyclic.} = ref object
+  Node[Key, Val]  = ref object
```
zerbina, line 20:
```suggestion
  Node[Key, Val] {.acyclic.} = ref object
```

zerbina (changes requested):
Thank you for the PR. Having an automatic leak test for the compiler and tools is a very welcome addition, being another line of defense against bugs with destructor injection and/or the cycle collector.

As for removing `.acyclic` from all types, I strongly disagree. Most reference types used in the compiler are indeed acyclic, even though it doesn't look like it from their type definitions.

Only `TSym`, `TType`, `TNode`, `TIdObj`, and `ModuleGraph` should form cycles, all others do not, and if they actually do, then that's a bug, which ASAN would catch, that needs to be fixed.

alaviss:
> As for removing `.acyclic` from all types, I strongly disagree. Most reference types used in the compiler are indeed acyclic, even though it doesn't look like it from their type definitions.

While I agree that these annotations might return, I think it's best that we have this PR bring us back to a "correct" baseline, then reintroduce the annotations in later PRs.

This makes sure that we have a documented trail of progress as well as information about why those types are acyclic documented within the compiler commit history.

I'm a strong believer of make it work, make it correct then make it fast, so I'd prefer each steps to be documented.

As an aside, it seems like this change brought out some allocator bugs which caused tcompilerapi to crash, so we will have to look into that first.

zerbina:
> I'm a strong believer of make it work, make it correct then make it fast, so I'd prefer each steps to be documented.

But removing all `.acyclic` annotations is a step back, in that it *hides* potential issues. If, for example, `ParsedDataNode` forms a cycle, then that's flat out wrong, but it's no longer caught now. 

Also, while it's true that the other types have no dedicated commit that I know of, `CgNode` *does* have a dedicated PR that marked it as `.acyclic`.

zerbina:
> As an aside, it seems like this change brought out some allocator bugs which caused tcompilerapi to crash, so we will have to look into that first.

I've looked into it, and it's not a bug with the allocator, rather a fundamental issue with how the cycle collector handles destruction of cells. It's a variation of the issue with `.acyclic` I mentioned -- I'll create the issue today.

alaviss:
> But removing all `.acyclic` annotations is a step back, in that it _hides_ potential issues. If, for example, `ParsedDataNode` forms a cycle, then that's flat out wrong, but it's no longer caught now.

Right now, `{.acyclic.}` is an optimization, not an indicator for correctness. While this PR introduces a process to verify these using sanitizers, I would not consider it reliable on any platforms other than Linux. As far as we know of right now, any of these could have not met the assertion, as they were not tested until now.

Which is why I think we should move ahead slow and steady, reintroduce these annotations with reasoning in code in future PR to contain those changes with a vote of confidence from CI.

Ideally we add a mode in ORC to verify that "acyclic" types don't exhibit cycles so local users can test it easily, but that's work for the future.

Furthermore, as you noted, there are issues with how the runtime handles acyclic as well, so it doesn't inspire confidence that acyclic is safe to apply without proper reasoning and testing.

zerbina:
> Right now, `{.acyclic.}` is an optimization, not an indicator for correctness

I somewhat disagree. I see it as an assertion made by the programmer to the compiler that the type isn't really cyclic even though it looks like it. While the compiler has to trust this assertion, although it could - as you said - verify it at runtime, the leak tests act as a strong verification of assumption now. 

> Furthermore, as you noted, there are issues with how the runtime handles acyclic as well

That's not what I said. There are issue with how the runtime handles freeing of cells, but that's not directly related to `.acyclic`.

---

As a compromise, what do you think about the following:
* the `.acylic` pragma is not removed for `CgNode` (refer to https://github.com/nim-works/nimskull/pull/992)
* I fix the cycle collector issue
* I, after this PR is merged, make a PR that reinstates the `.acyclic` pragma in all cases I mentioned

alaviss:
> As a compromise, what do you think about the following

Isn't this how I was advocating everything to be done? :P

Sans the CgNode ofc.

But sure, I'll re-instantiate acyclic for that one.

saem:
> Furthermore, as you noted, there are issues with how the runtime handles acyclic as well, so it doesn't inspire confidence that acyclic is safe to apply without proper reasoning and testing.

Besides `TNode` and friends that are not in fact `acyclic`, the others ones are in terms of their intention. So we can apply `acyclic` based on the behaviour of the implementation or based on intention. I believe @zerbina's perspective is that we should do it by intention and yours is by behaviour.

Going by behaviour has the plus side of having the correct net effect, while taking into account the current implementation's short comings. The downside is that the implementation is a moving target and so is the code that's using it, then we've got two moving targets, plus some tests which may or may not provide a sufficient spec (with failures marked as known issues). Also, since this stuff is lots of threading the needle, one has to actively remember where all the dragons lie, which isn't easy if there isn't directly focused effort on the collector & destructors.

Going by the intention has the plus side of testing the interface as it should be, and there is only one moving target the implementation, while the compiler code and tests together act as a spec (higher likelihood of it being more complete). The downside of course is that we might have some serious issues with the implementation so we _might_ have to compromise on intention correct code, but that's not guaranteed.

I personally favour the intention approach, because that exercises things with and without `acyclic`, and we can note leaving those types as such in this PR. Not to mention, it's easier to remember to take something away, than it is to re-add something. Finally, since working on arc/orc is more indirect work right now based on as the need arises, rather than dedicated effort, the behaviouar based approach is more onerous in terms of having to remember all sorts of edge cases whenever usage gets revisited.

zerbina:
> > As a compromise, what do you think about the following
> 
> Isn't this how I was advocating everything to be done? :P

It's not how I understood your suggestion. You didn't mention `CgNode` and you also said "slow and steady", which I consider a single PR to be the opposite of.

.github/workflows/leak.yml
```diff
@@ -0,0 +1,42 @@
...
+  # This is for passing required checks
+  pull_request:
+  merge_group:
+
+jobs:
+  leaktest:
+    # Skip this for PRs
+    #
+    # FIXME: uncomment this before merge
+    # if: github.event_name != 'pull_request'
```
saem, line 13:
```suggestion
    if: github.event_name != 'pull_request'
```

Leaving this as a suggestion so it's easier to remember.

saem:
```suggestion
    if: github.event_name != 'pull_request'
```

Leaving this as a suggestion so it's easier to remember.

alaviss:
> I somewhat disagree. I see it as an assertion made by the programmer to the compiler that the type isn't really cyclic even though it looks like it. While the compiler has to trust this assertion, although it could - as you said - verify it at runtime, the leak tests act as a strong verification of assumption now.

I viewed this one through the lens of the current state of the language to an user, hence usage of the word "right now", as there is no facilities for verifying these assertions, making them more or less an optimization.

The only solution at the moment is to verify via an external provider (LeakSanitizer in this case). If we want this to be treated as a proper assertion, then I believe testing facilities must be provided and documented.

compiler/front/main.nim
```diff
@@ -519,6 +519,17 @@ proc mainCommand*(graph: ModuleGraph) =
   when defined(nimDebugUnreportedErrors):
     addExitProc proc = echoAndResetUnreportedErrors(conf)
 
+  when defined(gcOrc):
```
saem, line 522: would disabling gc be better in `nim.nim`, that way it's restricted to the compiler binary, rather than anything that pulls in `main`?

saem (commented):
I'd favour not removing `acyclic` where it actually is valid.

compiler/front/main.nim
alaviss: Going by the code, this looks more like the main proc to me as you kinda become the `nim` frontend when you use this, but I don't object moving to `nim.nim` if needed.

saem: It's a soft suggestion, just feels symmetric with the change in `nimsuggest.nim`.

zerbina:
There was a bit of back and forth on this topic now. What's the agreed-on plan now?

If the correct `.acyclic` are removed, then the places I mentioned should at least be marked with a comment to note that `.acyclic` needs to be restored.

alaviss:
> If the correct `.acyclic` are removed, then the places I mentioned should at least be marked with a comment to note that `.acyclic` needs to be restored.

sure, I can do that

alaviss:
I still think that these annotations should be restored separately from this PR. For two reasons:

- As mentioned before, this PR brings us back to a "correct" (read: leakless) state.
- In the restore PR there would be more space to do things like adding an explainer on why is something acyclic, which I think should be done for all types marked as such.

saem:
> I still think that these annotations should be restored separately from this PR. For two reasons:
> 
> * As mentioned before, this PR brings us back to a "correct" (read: leakless) state.
> * In the restore PR there would be more space to do things like adding an explainer on why is something acyclic, which I think should be done for all types marked as such.

If you can run with the above plan then I'm good with it, one less thing for zerbina or I to keep in mind.

zerbina (approved):
Okay, then I think that's settled.

---

While a task for a later PR, the compiler should also run `GC_fullCollect` in strategic place. Besides leaking the cycles, the root list is also keeps growing, creating another logical memory leak (reintroducing `.acyclic` annotations will mitigate this somewhat).

While I do value speed a bit more, I also value keeping memory usage low during normal compilation.

nimsuggest/nimsuggest.nim
```diff
@@ -207,6 +207,16 @@ proc listEpc(): SexpNode =
...
 
+proc runGc() =
+  ## Runs a GC collection pass. This procedure will also collect any cycles assuming ORC is in use.
+  ##
+  ## When nimsuggest is being used as a library, no GC mode change will be performed.
+  when isMainModule and defined(gcOrc):
+    GC_enableOrc()
+  GC_fullCollect()
+  when isMainModule and defined(gcOrc):
+    GC_disableOrc()
```
alaviss: @zerbina do you know if I need to toggle orc state to run the cycle collector?

zerbina: Enabling ORC prior to `GC_fullCollect` is not necessary, but if you want to keep the cycle collector disabled afterwards, the `GC_disableOrc` is needed.

nimsuggest/nimsuggest.nim
```diff
@@ -207,6 +207,16 @@ proc listEpc(): SexpNode =
     methodDesc.add(docstring)
     result.add(methodDesc)
 
+proc runGc() =
+  ## Runs a GC collection pass. This procedure will also collect any cycles assuming ORC is in use.
+  ##
+  ## When nimsuggest is being used as a library, no GC mode change will be performed.
+  when isMainModule and defined(gcOrc):
+    GC_enableOrc()
```
alaviss:
```suggestion
```

alaviss:
/merge

I suppose we are all happy with this.

Let's hit the gas then. @zerbina I suppose we will be counting on you for the acyclic restoration.

PR #1105, opened 2024-01-18
ci: put release archive and metadata in the same prefix

alaviss:
## Summary
https://github.com/nim-works/nimskull/pull/1101 introduced a regression
where the archive is placed in a subfolder within the uploaded artifact.
This happened due to the metadata being one level above the archive,
which broke publisher.

## Details
This PR puts the metadata in the same subfolder as the archive so that
the artifact uploader correctly place them in the same level.


saem (approved):

alaviss:
/merge

Manually verified and the layout is correct.

PR #1106, opened 2024-01-18
fix(orc): cycle collection causing heap corruption

zerbina:
## Summary

In case there was a reference cycle that was kept alive by a non-
traced cell, and the non-traced cell was kept alive by a garbage
reference cycle, running the cycle collector made the collector
recurse.

A non-traced cell refers to a managed heap cell that's not scanned by
the cycle collector because its not cyclic (including due to the use
of `.acyclic`).

Depending on the circumstances, the recursion would either cause a
stack overflow or a heap corruption. The bug is now fixed.

## Details

The fundamental problem is that a call to `free` can add new roots to
the global `roots` list, either via automatic destructor or through
user-defined destructors.

Since neither the `roots` list nor root threshold are reset prior to
the cleanup phase, a call to `registerCycle` would trigger a
`collectCycles` recursion, which then ran trial deletion for all
already scanned roots again.

This would then cause a heap corruption if one of the roots was freed
already.

### The solution

Prior to the cleanup phase, all processed roots are removed from the
`roots` list. In addition, a reentrance guard (`collectorLock`) is
added that prevents recursion into `collectCycles`/`partialCollect`,
even if explicitly requested via `GC_fullCollect`.

So that a single full collection frees *all* reference cycles, even
those only discovered during a run of the cleanup phase,
`collectCycles` runs the core cycle collection in a loop until no more
new roots were discovered during cleanup.

### DIFF
--- a/tests/arc/tcycle_test.nim
+++ b/tests/arc/tcycle_test.nim
@@ -24,11 +24,11 @@ proc init() =
   # create two cycles and link them together via `Bridge`
   var a = Cycle1(self: Cycle1())
   a.self.self = a
-  var b = Cycle2(self: Cycle2())
-  b.self.self = b
+  var c = Cycle2(self: Cycle2())
+  c.self.self = c
 
   a.self.bridge = new Bridge
-  a.self.bridge.other = b
+  a.self.bridge.other = c
   # the following structure now exists:
   # A -> B -> Bridge -> C -> D
   # ^    |              ^    |
--- a/tests/arc/tcycle_test.nim
+++ b/tests/arc/tcycle_test.nim
@@ -1,0 +1,58 @@
+discard """
+  description: '''
+    A regression test for a bug with the cycle collector, where two reference
+    cycles linked together by a non-cyclic reference type trigger a memory
+    corruption
+  '''
+  targets: "c"
+  matrix: "-d:useMalloc"
+  valgrind: true
+"""
+
+type
+  Bridge = object
+    other: Cycle2
+  Cycle1 = ref object
+    bridge: ref Bridge
+      ## bridges the two cyclic types. This must be a `ref`, otherwise tracing
+      ## would visit the `other` field and the issue would not be triggered
+    self: Cycle1
+  Cycle2 = ref object
+    self: Cycle2
+
+proc init() =
+  # create two cycles and link them together via `Bridge`
+  var a = Cycle1(self: Cycle1())
+  a.self.self = a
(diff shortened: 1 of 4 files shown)


zerbina:
The compiler could also emit a `=trace` hook, together with `nimTraceRef` calls, for all types that are not cyclic themselves but store `ref`s to cyclic types. This would allow the cycle collector to trace "through" them.

However, this is a somewhat orthogonal solution, as while it would also fix `tcyclic_test` and `tcyclic_test_acyclic_annotation`, it would not solve the general problem of the cycle collector having trouble with new roots registered during the free phase, which is somethings that's still possible through user-defined destructors.

tests/arc/tcycle_test_acyclic_annotation.nim
```diff
@@ -0,0 +1,49 @@
...
+    self: Cycle
+
+proc init() =
+  # create two cycles and link them together via `Bridge`
+  var a = Cycle(self: Cycle())
+  a.self.self = a
+  var b = Cycle(self: Cycle())
+  b.self.self = b
+
+  b.self.bridge = Bridge(other: a)
```
saem:
```suggestion
  var c = Cycle(self: Cycle())
  c.self.self = c

  a.self.bridge = Bridge(other: c)
```

the names line up with the diagram a bit better, I believe, as `B` and `D` are unamed (created inline).

tests/arc/tcycle_test.nim
```diff
@@ -0,0 +1,58 @@
...
+
+proc init() =
+  # create two cycles and link them together via `Bridge`
+  var a = Cycle1(self: Cycle1())
+  a.self.self = a
+  var b = Cycle2(self: Cycle2())
+  b.self.self = b
+
+  a.self.bridge = new Bridge
+  a.self.bridge.other = b
```
saem:
```suggestion
  var a = Cycle1(self: Cycle1())
  a.self.self = a
  var c = Cycle2(self: Cycle2())
  c.self.self = c

  a.self.bridge = new Bridge
  a.self.bridge.other = c
```

Again, just to line up better with the diagram naming convention below.

saem (approved):
LGTM

The variable rename suggestions in the tests are soft suggestions.

alaviss (approved):
LGTM

Just have to fix the tests and this can go in

zerbina:
/merge

Thank you for the review, @saem and @alaviss.

alaviss:
/merge

homebrew went brr

PR #1107, opened 2024-01-19
osproc: don't allocate memory during fork-exec

alaviss:
## Summary

Previously,  `osproc`  would attempt to allocate memory during fork-exec
via the use of  `findExe`  to find programs in  `PATH` . This causes
memory allocation during this sensitive period.

This PR replaces the use of  `findExe` , which was employed due to 
`execvpe`  not being ubiquitous, with  `execvpe` , now that it is widely
available.

## Details

There are a couple problems with allocating memory during fork-exec:

- This is extremely unsafe as this code is also used in  `vfork()`  case
(although disabled by default, thankfully), of which the use of an
allocator would completely mess up the parent process.
- This is super slow, since  `fork()`  has to perform CoW on pages that
are touched by the allocation procedure.
- A deadlock can occur in rare cases due to interaction with the global
allocator lock. This problem was observed in disruptek/balls.

With regard to substituting `findExe` with `execvpe`:

* This API has been available in uClibc since 2013, musl libc since 2014
and Haiku since 2015, removing the need to special case against those
systems.
* We can expect behaviors closer to how other programs perform  `PATH` 
searches. It is expected that there won't be any regression from this
change.

### DIFF
--- a/lib/pure/osproc.nim
+++ b/lib/pure/osproc.nim
@@ -1183,10 +1183,8 @@ elif not defined(useNimRtl):
       discard fcntl(data.pErrorPipe[writeIdx], F_SETFD, FD_CLOEXEC)
 
       if (poUsePath in data.options):
-        when defined(uClibc) or defined(linux) or defined(haiku):
-          # uClibc environment (OpenWrt included) doesn't have the full execvpe
-          let exe = findExe(data.sysCommand)
-          discard execve(exe.cstring, data.sysArgs, data.sysEnv)
+        when not defined(macosx):
+          discard execvpe(data.sysCommand.cstring, data.sysArgs, data.sysEnv)
         else:
           # MacOSX doesn't have execvpe, so we need workaround.
           # On MacOSX we can arrive here only from fork, so this is safe:


saem:
I'm a little confused, is the new incoming code dangerous or is that in reference to the old code being replaced?

Would really like a `Summary` and `Details` section, separating user impact and change details, respectively.

alaviss:
I forgot to write the lead-in :p

saem (approved):
The other way to fix it would be to know/limit how much memory `findExe` needs, pre-allocate that, and then do the search? The upside for this being more backwards compatability, but also sounds like a pain.

alaviss:
> The other way to fix it would be to know/limit how much memory `findExe` needs, pre-allocate that, and then do the search? The upside for this being more backwards compatability, but also sounds like a pain.

The search itself is wrong due to potential TOCTOU races (between finding a binary and executing it), so I'd like to avoid it if possible.

Fixing this the proper way (ie. not using libc goodies) is just really hard. I had one stashed in nim-sys, but it requires much of the infrastructure I built for that project.

PR #1110, opened 2024-01-19
relax a couple (terminal) func side-effects

disruptek:
## Summary
I noSideEffect'd a couple raises because they don't compile as `func` otherwise and I don't want to spend the morning tracking down and un-func'ing everything.

## Notes for Reviewers
Maybe we shouldn't use `func` until we understand what it does and why it exists.

### DIFF
--- a/compiler/utils/idioms.nim
+++ b/compiler/utils/idioms.nim
@@ -33,7 +33,8 @@ func unreachableImpl(str: string, loc: IInfo) {.noinline, noreturn.} =
     if str.len > 0: " unreachable: "
     else:           " unreachable"
   msg.add str
-  raiseAssert(msg)
+  {.cast(noSideEffect).}:
+    raiseAssert(msg)
 
 func unreachableImpl(e: enum, loc: IInfo) {.noinline, noreturn.} =
   ## A bit more efficient than ``unreachable($e)``, due to the stringication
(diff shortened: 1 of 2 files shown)


compiler/utils/idioms.nim
```diff
@@ -33,7 +33,8 @@ func unreachableImpl(str: string, loc: IInfo) {.noinline, noreturn.} =
     if str.len > 0: " unreachable: "
     else:           " unreachable"
   msg.add str
-  raiseAssert(msg)
+  {.cast(noSideEffect).}:
```
saem, line 36: This is an internal compiler func, where are you running into issues with this one? I might be able to make it go away.

lib/system.nim
```diff
@@ -2698,7 +2698,9 @@ when compileOption("rangechecks"):
     ## Helper for performing user-defined range checks.
     ## Such checks will be performed only when the `rangechecks`
     ## compile-time option is enabled.
-    if not cond: sysFatal(RangeDefect, "range check failed")
+    if not cond:
+      {.cast(noSideEffect).}:
+        sysFatal(RangeDefect, "range check failed")
```
alaviss, line 2703: If `sysFatal` is not classified as `noSideEffect` then I think that's a bug

zerbina:
The compiler doesn't treat `raise` itself as a side-effect, so that should not be an issue. What options are you compiling with when you're seeing an error?

zerbina:
With #1187, `sysFatal` no longer has side effects under `--panics:on`, which should make the workarounds implemented here obsolete.

PR #1111, opened 2024-01-19
sem: ill-formed sym in `for` produces an error

saem:
## Summary

Ill-formed  `for`  loops with an error in the sym position, in
particular produced by meta routines now correclty produce an error.

## Details

Prior to this fix the error would not be reported because it was not
added to the production that was ultimately wrapped in an  `nkError` .
This would result in a compiler crash due to assertion failure, wrapping
AST in an error without an error present therein. Finally, the new
approach no longer mutates the input AST.

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -1434,7 +1434,7 @@ proc symForVar(c: PContext, n: PNode): PSym =
       semmedNode[0] = resultNode
       semmedNode[1] = pragma
 
-  if resultNode.kind == nkError or hasPragma and n[1].kind == nkError:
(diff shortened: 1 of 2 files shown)


saem:
OK, this time it should work, I promise... unless it doesn't. 😆

zerbina (approved):
Apart from fixing the improper wrapping, this also fixes another AST corruption in the case where the pragma has an error (the input AST was previously modified, and now it isn't).

saem:
> Apart from fixing the improper wrapping, this also fixes another AST corruption in the case where the pragma has an error (the input AST was previously modified, and now it isn't).

Thanks, I noted this in the PR details.

Issue #1113, opened 2024-01-19
type-checking issue with dereferencing expression within templates


alaviss:
The title is a mouthful but I have no better description

### Example

This one affects only ARC:

```nim
type
  O = object

iterator iter(x, y: O): O = discard

template stuff() =
  for _ in iter((ref O)()[], O()):
    discard

stuff()
```

This one affects both GCs:

```nim
type
  O = object

proc foo(x, y: O) = discard

template stuff() =
  foo((ref O)()[], O())

stuff()
```

### Actual Output

First example:

```
$ nim c --gc:arc test.nim
test.nim(7, 16) Error: type mismatch: got <test.O, test.O>
but expected one of:
iterator iter(x, y: O): O
  first type mismatch at position: 1
  required type for x: O
  but expression 'ref O()[]' is of type:

expression: iter(ref O()[], O())
```

Second example:

```
test.nim(9, 6) Error: type mismatch: got <test.O, test.O>
but expected one of:
proc foo(x, y: O)
  first type mismatch at position: 1
  required type for x: O
  but expression 'ref O()[]' is of type:

expression: foo(ref O()[], O())
```

### Expected Output

This should not error

### Additional Information

* ~~Only occurs with `--gc:arc`~~
* Minimized from https://github.com/nim-works/nimskull/blob/7e7226f1904360c2e14c8f57fc9e0f61e5483d58/lib/pure/streamwrapper.nim#L101-L104
* Does not occur if used outside of templates, or if `{.dirty.}` templates are used.

zerbina:
Resolved by #1116, but was not closed automatically.

PR #1115, opened 2024-01-19
streamwrapper: create an alias before using fields

alaviss:
This serves as a workaround for https://github.com/nim-works/nimskull/issues/1113 as that bug blocks software using osproc from compiling with ARC.

---

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


zerbina:
The problem looks to be causes by an AST corruption. If it's a simple fix, I think I'd prefer the fix rather than implementing a workaround.

zerbina:
Yep, it was indeed an AST corruption -- a PR is incoming :D

alaviss:
Closed in favor of #1116

Issue #1075, opened 2023-12-31
run tests with all targets by default


zerbina:
## Summary

Run all tests for all targets by default, with non-working combinations marked as a known issue. 

## Motivation

Test coverage for the VM and JavaScript backends is still lacking, making it hard to track how well both backends support the language.

Since testing with all targets is not the default, it's also easy to miss enabling a test for all targets. The extra `targets: "c js vm"` specification also introduces additional visual noise.

Finally, since `knownIssue` tests are tried automatically during CI runs, this ensures that bug fixes don't go by unnoticed.

## How to approach
1. pick a test category (i.e., a directory from the `tests` folder) that isn't already enabled for all targets [here](https://github.com/nim-works/nimskull/blob/12f0aa45abab91601f53cc79854a27376c403d0f/testament/testament.nim#L1111)
2. enable it for all targets
3. if a test fails when run for a target where it should succeed, mark the test as a `knownIssue`  for that target
4. (optionally) if a test fails and there's no GitHub issue already documenting the problem, create one and link to it from the test

Doing this category-by-category allows means that:
* review of the changes is easier, as the amount of changes is smaller
* multiple people can work on the overall task in parallel (by picking different categories)
* improved test coverage is available earlier

For an example of how such a PR could/would look like, refer to #1118 or #1082.

If you'd like to work on this issue, please ping @zerbina.

PR #1118, opened 2024-01-19
tests: run more tests with all targets by default

ghost:
Summary
=======
Enabled all test targets for some tests, fixed tests where necessary.
Better test coverage for JS and VM.

Details
=======
- Reordered the target category strings to be in alphabetical order to
make later adjustments easier.
- The  `targets`  key is removed for all tests except one to disable a
target where the target doesn't make
   sense.
-  `write(stdout, ...)`  is replaced with  `echo ...`  to support the JS
and VM targets.
- Some  `static: `  blocks have been removed and replaced with the VM
target where not specific to compile-time execution.
- Tests that fail but should work are marked as `knownIssue`.

part of https://github.com/nim-works/nimskull/issues/1075

### DIFF
--- a/tests/align/talign.nim
+++ b/tests/align/talign.nim
@@ -1,6 +1,5 @@
 discard """
 ccodeCheck: "\\i @'NIM_ALIGN(128) NI mylocal1' .*"
-targets: "c"
 labels: "pragma alignment generic"
 description: '''
   . First one is is for Azure. The keyword ``alignof`` only exists in ``c++11``
@@ -9,6 +8,7 @@ description: '''
   . From https://github.com/nim-lang/Nim/issues/13122
     {.align.} pragma is not applied if there is a generic field
 '''
+knownIssue.vm: "VM does not support casting a ptr to a uint"
 """
 
 import globalalignas
--- a/tests/align/talign.nim
+++ b/tests/align/talign.nim
@@ -8,6 +8,7 @@ description: '''
   . From https://github.com/nim-lang/Nim/issues/13122
     {.align.} pragma is not applied if there is a generic field
 '''
+targets: "!js"
 knownIssue.vm: "VM does not support casting a ptr to a uint"
 """
 
--- a/tests/assert/tassert_c.nim
+++ b/tests/assert/tassert_c.nim
@@ -1,10 +1,10 @@
 discard """
-  cmd: "nim $target $options --excessiveStackTrace:off $file"
+  matrix: "--excessiveStackTrace:off"
   output: '''true'''
   knownIssue.js vm:'''
-  JavaScript produces a different stacktrace output
-  message than the c target.
-  VM does not support 'getStackTrace' on an Exception.
+    JavaScript produces a different stacktrace output message than the C
+    target.
+    VM does not support 'getStackTrace' on an Exception.
   '''
 """
 
--- a/tests/assert/tassert_c.nim
+++ b/tests/assert/tassert_c.nim
@@ -1,11 +1,16 @@
 discard """
   cmd: "nim $target $options --excessiveStackTrace:off $file"
   output: '''true'''
+  knownIssue.js vm:'''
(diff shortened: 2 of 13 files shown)


tests/assert/tassert_c.nim
```diff
@@ -1,11 +1,16 @@
 discard """
   cmd: "nim $target $options --excessiveStackTrace:off $file"
   output: '''true'''
+  knownIssue.js vm:'''
+  JavaScript produces a different stacktrace output
+  message than the c target.
+  VM does not support 'getStackTrace' on an Exception.
```
zerbina:
```suggestion
    JavaScript produces a different stacktrace output message than the C
    target.
    VM does not support 'getStackTrace' on an Exception.
```

tests/align/talign.nim
```diff
@@ -1,6 +1,5 @@
 discard """
 ccodeCheck: "\\i @'NIM_ALIGN(128) NI mylocal1' .*"
-targets: "c"
```
zerbina, line 3: The JS target has no concept of alignment, and I don't think it ever will.

You can use negative form, `targets: "!js"`, to disable only the JS target while still inheriting the remaining default targets.

tests/assert/tassert_c.nim
```diff
@@ -1,11 +1,16 @@
 discard """
   cmd: "nim $target $options --excessiveStackTrace:off $file"
```
zerbina:
```suggestion
  matrix: "--excessiveStackTrace:off"
```
If you come across a `cmd` specification that doesn't change the command that's run (e.g., `check`), you can replace it with a single-item `matrix`. This makes it clearer for both `testament` and the reader what's happening.

tests/assign/tobject_assign.nim
```diff
@@ -3,6 +3,9 @@ discard """
     Regression test for assigning to an uninitialized `seq` slot that
     has a type header
   '''
+  knownIssue.vm: '''
+  Cannot 'importc' proc 'localtime' in 'times.nim' at compile time
```
zerbina:
```suggestion
    'localtime' is not yet supported with the VM backend
```
The error message as reported by the compiler is misleading, mentioning compile-time while the problem is not at all related to compile-time evaluation.

**Edit:** I've replaced the backticks with single quotes.

tests/assign/toverload_asgn2.nim
```diff
@@ -1,7 +1,7 @@
 discard """
   output: '''i value 88
 2aa'''
-    knownIssue: "possible issue where result is not treated as var Concrete in passAround"
+  knownIssue.c js vm: "possible issue where result is not treated as var Concrete in passAround"
```
zerbina:
```suggestion
  knownIssue: "possible issue where result is not treated as var Concrete in passAround"
```
Since the bug is unrelated to the selected backend (it's a bug in semantic analysis), keeping the `knownIssue` unspecific is a bit clearer.

zerbina (changes requested):
A very well-made PR. Thank you very much!

I left a few suggestions regarding the wording and styling of `knownIssue`s, as well as some additional requests.

Also, a small request regarding the PR description: could you use a listing for the items in the PR's 'Details' section? That'd make it a bit easier to parse.

tests/assign/tobject_assign.nim
ghost: I noticed you use backticks instead of single quotes like I did. Should I generally use backticks in such a case for overall consistency or does it not matter?

zerbina: That's a habit of mine from writing documentation comments, where names are wrapped in backticks.

For test specifications, I think it doesn't matter, although it should be locally consistent within a single `knownIssue` or `description` item.

zerbina (approved):
Very good work, thanks again. Also, good call on reordering the strings in `defaultTargets`.

PR #1121, opened 2024-01-20
compiler: enable ORC when running the leak test

zerbina:
## Summary

Even when ORC is disabled at run-time, cells potentially part of
reference cycles are added to potential-cycle-root list, preventing
the leak sanitizer from detecting cycles not being free.

The `leakTest` define is now passed to the compiler when leak-
testing, with ORC enabled when the define is present.

### DIFF
--- a/.github/workflows/leak.yml
+++ b/.github/workflows/leak.yml
@@ -26,6 +26,7 @@ jobs:
         run: |
           ./koch.py all-strict \
             -d:useMalloc \
+            -d:leakTest \
             --passC:-fsanitize=leak \
             --passC:"-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer" \
             --passL:-fsanitize=leak \
--- a/compiler/front/main.nim
+++ b/compiler/front/main.nim
@@ -519,7 +519,7 @@ proc mainCommand*(graph: ModuleGraph) =
   when defined(nimDebugUnreportedErrors):
     addExitProc proc = echoAndResetUnreportedErrors(conf)
 
-  when defined(gcOrc):
+  when defined(gcOrc) and not defined(leakTest):
     # Compilation is currently very taxing on ORC due to frequent
     # creations and destructions of ref objects with potential cycles.
     #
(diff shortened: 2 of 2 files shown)


alaviss:
How about we set it to run at exit? Because even now a single bootstrap cycle takes 70s for leak tests, so I'm not sure if we want to have it running for the entirety of compilation.

Though, being merge CI means time doesn't matter as much.

alaviss (approved):
Actually I just looked at the numbers again and they looks fairly normal, so I don't think there's much to do here.

zerbina:
> How about we set it to run at exit?

Sure, that works too.

> Because even now a single bootstrap cycle takes 70s for leak tests

Hm, do you mean locally for you? In the CI, the (NimSkull part) of the second and third iteration only take ~27 seconds each (the number comes from [here](https://github.com/nim-works/nimskull/actions/runs/7596243536/job/20689735177)).

saem (approved):
tl;dr: I think we should merge as is

My understanding is that if we were to add it at the end we'd put a `GC_enableOrc()` at the end of `mainCommand` and also force a collection, in the presence of the `leakTest` define.

If it's that tiny, then that's tempting for faster CI. Although, and my memory could be playing tricks on me, but I don't believe we've got leak detection testing in our regular tests. So maybe we're better off deferring the optimization until we've got more core tests in place; I could just be paranoid.

alaviss:
> In the CI, the (NimSkull part) of the second and third iteration only take ~27 seconds each (the number comes from [here](https://github.com/nim-works/nimskull/actions/runs/7596243536/job/20689735177)).

Yea that's why I revised later that the numbers looked alright.

I must have imagined the 70s figure because I couldn't find it anywhere now. Though a forced-collection might be useful since I don't think we perform full collect at the end of the program by default.

alaviss:
/merge

Since we have 2 approvals already, let's get this in.

alaviss:
/merge

Just some GitHub issues, should be fine now with another go.

PR #1123, opened 2024-01-20
ci: add Windows builds using MinGW toolchain

### LINKED ISSUE #276: Add Windows binaries to GitHub Actions

alaviss:
## Summary
Implement Windows 64-bit testing and builds using MinGW toolchain, as
well as providing Windows compatibility patches

### Toolchain and dependencies
* DLLs are built within CI from Microsoft's vcpkg repository as it
appears to be the most reliable source for pinned library versions out
there.
* vcpkg uses custom triplets hosted in this repository to ensure
reproducible builds
* Toolchain is sourced from
https://github.com/niXman/mingw-builds-binaries, which provides bare
toolchains without extras like gdb.
* The target is Windows Universal C Runtime, the new C99 compatible
standard library bundled with Windows 10.

### CI changes
* Download/Upload compiler actions has been rewritten in Python to make
sure that they are compatible across OSes.
* Added a new `setup-mingw` action to setup a MinGW environment.
* Added Windows to the matrix using `windows-2022` image.

### Git tree changes
* Added configuration to force LF line endings on all platforms. This
makes sure that Windows users always has the same source on disk as
everyone else.

### Compiler changes
* Added missing resources required to compile  `koch`  from source
archive.
* On Windows,  `nim`  and  `nimsuggest`  now have a stack size of 8MiB.
This avoids stack overflow when compiling complex programs, and make
sure that the stack size is similar to Linux and macOS.
* On Windows, PE timestamps are removed for release builds using GCC.
This makes sure that release builds are reproducible.

### Standard library changes
*  `posix_utils`  examples are disabled on Windows as they could not be
compiled there.
*  `pcre`  and  `sqlite`  now prefers canonical DLL names on Windows.
This makes it easier to use built DLLs from repositories like Conan or
vcpkg.

### Tests changes
*  `nimsuggest` 's tester now normalizes all path endings to Unix-style 
`/` .
* Removed  `tfdleak_multiple`  as the "flaky" status no longer applies
due to changes to  `tfdleak` .
*  `tvmprofiler`  got an increased time range to account for Windows CI
being relatively slower than the rest.
* Tests that requires POSIX imports are disabled.

Closes https://github.com/nim-works/nimskull/issues/276


saem:
Minor update to the PR summary to indicate that it's for 64-bit. Reviewing the rest of the PR now.

saem (dismissed):
LGTM

saem (approved):
IMO, this change is about as lean as it gets. If we broke things out further, then we'd be making changes without Windows CI in place. Alternatively, we'd be blocked on Windows CI because it needs some of these changes.

saem:
PR description looks good.

saem:
Not going to merge until we get a chance to test the windows binaries that are built from this CI flow.

tests/arc/timportedobj.nim
```diff
@@ -1,6 +1,7 @@
 discard """
   cmd: "nim c --gc:arc $file"
   action: "compile"
+  disabled: "windows"
```
zerbina, line 4: This test is not POSIX specific, but it can stay disabled on Windows for now. I'll make a follow-up PR that removes the `posix` module dependency.

zerbina (approved):
Looks good to me. Thank you for adding Windows CI support!

Since this closes #276, could you add a "Closes" line for it to the PR description?

> As far as I know of, @zerbina uses MinGW on Windows, so we are targeting the most tested toolchain.

Yep, I use MinGW on Windows.

alaviss:
/merge

nimgrep is working on windows with the bundled pcre, so :shipit:

tests/arc/timportedobj.nim
zerbina: The follow-up PR: #1149.

PR #1124, opened 2024-01-21
sem: clarify macro pragma with bad inputs error

### LINKED ISSUE #1108: confusing error message with typed macro on proc with semantic error
In this example, the error got shifted to the typed macro symbol within the proc pragma instead of being about the invalid symbol.

### Example

```nim
import macros

macro bar(x: typed): untyped =
  x

proc foobar() {.bar.} =
  totallyInvalidSym()
```

### Actual Output

```
test.nim(6, 17) Error: invalid pragma: bar
```

### Expected Output

```
test.nim(7, 3) Error: undeclared identifier: 'totallyInvalidSym'
```

### Additional Information

* Found during port of cps to nimskull
* The error can be found with a higher `--errorMax`

```
$ nim -v

Nimskull Compiler Version 0.1.0-dev.21160 [linux: amd64]

Source hash: 7e7226f1904360c2e14c8f57fc9e0f61e5483d58
Source date: 2024-01-19

active boot switches: -d:release
```

saem:
## Summary

Typed macro pragmas now return a clear error message when provided bad
input AST (i.e.  `nkError` ).

## Details

The majority of changes were in  `sigmatch.matchesAux` , it now provides
the precise argument that was used for matching against the formal
parameter. In addition, when determine if there is an error at the
current parameter position it takes into account whether the formal
parameter allows  `nkError`  (i.e. is  `typed` ). This also resulted in
some minor changes, largely improvements, to error messages reported, so
the impacted test was updated ( `tsigmatch` ).

The primary impact is now when error AST is passed to a  `typed`  macro,
the programmer no longer receives an error stating that the the pragma
is invalid, instead the error from the input AST is reported. This is
true for  `typed`  and  `varargs[typed]`  parameters.

Additional fixes that were made as part of this:

-  `sigmatch.paramTypesMatchAux`  was updated to apply  `typeRel`  and
match index score changes when receiving an error node for a 
`typed/varargs[typed]`  formal parameter
-  `evaltempl.evalTemplateArgs`  wasn't scanning all args, the index in
the loop was hardcoded to  `1` , instead of  `i` , a typo
-  `evaltempl.evalTemplateArgs`  did not recursively scan args for 
`nkError` , this meant that  `varargs[typed]`  were ignored, this was
replaced with the  `walkErrors`  iterator
-  `errorhandling.buildErrorList`  couldn't handle  `nil`  nodes in its
traversal, this led to NPE when scanning for errors in template args

Fixes https://github.com/nim-works/nimskull/issues/1108


zerbina:
Reading over the changes, is the intention to allow erroneous AST as `typed` arguments? That is, is the following intended to compile now?
```nim
macro m(x: typed) =
  discard

m(missingIdent)
```
I haven't tried it, but I believe this would compile with the changes as implemented here. Previously, it would fail.

saem:
> Reading over the changes, is the intention to allow erroneous AST as `typed` arguments? That is, is the following intended to compile now?
> 
> ```nim
> macro m(x: typed) =
>   discard
> 
> m(missingIdent)
> ```
> 
> I haven't tried it, but I believe this would compile with the changes as implemented here. Previously, it would fail.

Yeah, it definitely seems that way, but what actually happens is that it allows the match (`csMatch` instead of `csNoMatch`), but when the macro/template is actually evaluated (`evalTemplateArgs`) we get an error. I feel timing-wise it makes sense.

```nim
# here is the output from a local run

foo.nim(7, 4) template/generic instantiation of `m` from here
foo.nim(7, 5) Error: undeclared identifier: 'missingIdent'
candidates (edit distance, scope distance); see '--spellSuggest':
 (6, 4): 'nnkIdent' [enumField declared in nimskull/lib/core/macros.nim(50, 24)]
nimskull/compiler/sem/lookups.nim(791, 35) compiler msg instantiated here [MsgOrigin]
nimskull/compiler/sem/evaltempl.nim(156, 11) compiler report submitted here [MsgOrigin]
nimskull/compiler/nim.nim(154)             nim
nimskull/compiler/nim.nim(89)              handleCmdLine
nimskull/compiler/front/main.nim(539)      mainCommand
nimskull/compiler/front/main.nim(491)      compileToBackend
nimskull/compiler/front/main.nim(214)      commandCompileToC
nimskull/compiler/modules/modules.nim(212) compileProject
nimskull/compiler/modules/modules.nim(128) compileModule
nimskull/compiler/sem/passes.nim(275)      processModule
nimskull/compiler/sem/passes.nim(102)      processTopLevelStmt
nimskull/compiler/sem/sem.nim(950)         myProcess
nimskull/compiler/sem/sem.nim(868)         semStmtAndGenerateGenerics
nimskull/compiler/sem/semstmts.nim(3357)   semStmt
nimskull/compiler/sem/semexprs.nim(1441)   semExprNoType
nimskull/compiler/sem/semexprs.nim(3821)   semExpr
nimskull/compiler/sem/semstmts.nim(3253)   semStmtList
nimskull/compiler/sem/semexprs.nim(3733)   semExpr
nimskull/compiler/sem/semexprs.nim(2926)   semWhen
nimskull/compiler/sem/semexprs.nim(3821)   semExpr
nimskull/compiler/sem/semstmts.nim(3253)   semStmtList
nimskull/compiler/sem/semexprs.nim(3664)   semExpr
nimskull/compiler/sem/semexprs.nim(1417)   semDirectOp
nimskull/compiler/sem/semexprs.nim(1190)   afterCallActions
nimskull/compiler/sem/sem.nim(775)         semMacroExpr
nimskull/compiler/sem/evaltempl.nim(156)   evalTemplateArgs
nimskull/compiler/front/msgs.nim(944)      handleReport
nimskull/compiler/front/msgs.nim(679)      handleReport
nimskull/compiler/front/msgs.nim(285)      quit
FAILURE
```

saem:
Whoops, line numbers are off in the test, I'll fix that tomorrow.

zerbina:
> but when the macro/template is actually evaluated (`evalTemplateArgs`) we get an error. I feel timing-wise it makes sense.

Ah, I forgot about macro/template argument preparation not letting `nkError` through.

At first I thought this flow to be somewhat non-intuitive. However, since `typed` just means "typed AST", not failing the match does indeed make sense timing-wise, I think.

compiler/sem/sigmatch.nim
```diff
@@ -2796,6 +2798,10 @@ proc matchesAux(c: PContext, n, nOrig: PNode, m: var TCandidate, marker: var Int
         m.error.firstMismatch.kind = kVarNeeded
         noMatch()
 
+  template acceptsTyped(typ: PType): bool =
+    typ != nil and (typ.kind == tyTyped or
+                    typ.kind == tyVarargs and typ[0].kind == tyTyped)
```
```suggestion
    m.calleeSym != nil and m.calleeSym.kind in {skMacro, skTemplate} and
      typ != nil and (typ.kind == tyTyped or
                      typ.kind == tyVarargs and typ[0].kind == tyTyped)
```
This is not very pretty, but it does prevent `varargs[typed]` used outside of templates/macros (`echo`, for example) from treating an error as acceptable.

At the moment, the error would be leaked, and only later wrapped via `foldInAst`.

zerbina (commented):
Given the following:
```nim
macro m(x: varargs[typed]) =
  discard

m(missingIdent)
```
the following AST leaves `resolveOverloads` (I've also verified this locally):
```
(Call (Sym "m") (HiddenStdConv (Empty) (Bracket (Error)) 
```

Since `evalTemplateArgs` doesn't account for nested errors, the erroneous AST is passed through to the macro/template (which in the above example means that it's effectively discarded).

I think the best solution for now would be to have `evalTemplateArgs` also look for errors within `nkHiddenStdConv` nodes of `tyVarargs[tyTyped]` type.

I've left a suggestion with how a second problem with `varargs[typed]` could be resolved.

saem:
> At first I thought this flow to be somewhat non-intuitive. However, since `typed` just means "typed AST", not failing the match does indeed make sense timing-wise, I think.

Ah, I found a problem, if there is also an overload present `macro m(x: untyped) = discard`, then it will show an ambiguous call error. I think I have to slightly lower the score for a template or macro match with an error argument.

saem:
Hmm, the ambiguous call is actually not an issue from the perspective of what used to happen. In the case of a valid symbol, such as a `int`, in place of `missingIdent`, an ambiguous call error would be issued. Now in the error scenario, the `untyped` version is used.

So first, need to decide if the new behaviour is better and if so, that should be a separate PR, IMO. To answer the first question, is the new behaviour better, here it is outlined:

1. if an `untyped` overload exists:
   1. given an `nkError`, match the `untyped` overload and proceed
   2. given a non-`nkError`, treat the `typed` overload as more specific and use that
2. if an `untyped` overload does not exist:
   1. given an `nkError`, report the error message
   2. given a non-`nkError`, `typed` wins

The `1.x` and `2.ii` branches are the new behaviour, with `1.i` covered in this PR. The main downside I can think of this approach is that it sorta cements the current notion that `untyped` can match last, but we know that due to compile time effects it must match first, in the long run.

```nim
# shows 1.1
import std/macros

macro m(x: typed): untyped =
  newLit "typed"

macro m(x: untyped): untyped =
  newLit "untyped"

doAssert "untyped" == m(missingIdent)
```

```nim
# shows 1.2
import std/macros

macro m(x: typed): untyped =
  newLit "typed"

macro m(x: untyped): untyped =
  newLit "untyped"

doAssert "typed" == m(int)
```

compiler/sem/sigmatch.nim
saem: Thank you, fixed this locally.

saem: `echo`'s definition is a bit abusive, we're treating `typed` almost like a `dynamic` keyword, where it's not just like an `any`, which binds to a single type, but each element is bound separately.

Mostly just noting an interesting case for the future.

compiler/ast/errorhandling.nim
```diff
@@ -175,7 +175,8 @@ proc buildErrorList(config: ConfigRef, n: PNode, errs: var seq[PNode]) =
     errs.add n
   else:
     for i in 0..<n.len:
-      buildErrorList(config, n[i], errs)
+      if n[i] != nil:
```
zerbina, line 178: Hm, what's the AST that necessitates this guard?

compiler/sem/sigmatch.nim
```diff
@@ -3070,14 +3080,14 @@ proc matchesAux(c: PContext, n, nOrig: PNode, m: var TCandidate, marker: var Int
...
+              discard "assigned below"
             else:
               if formal.typ.kind != tyUntyped:
                 n[a] = operand
-              arg = paramTypesMatch(m, formal.typ, operand.typ, operand)
-              
-              if arg.isNil(): # invalid arg
-                noMatch()
+
+            arg = paramTypesMatch(m, formal.typ, operand.typ, operand)
```
```suggestion
            arg = paramTypesMatch(m, formal.typ, operand.typ, operand)
```
Removes the trailing white-space.

compiler/sem/sigmatch.nim
```diff
@@ -2919,13 +2924,13 @@ proc matchesAux(c: PContext, n, nOrig: PNode, m: var TCandidate, marker: var Int
 
         case operand.kind
         of nkError:
-          arg = operand
+          discard "assigned below"
```
zerbina, line 2927: It's a pre-existing issues (so no change required), but since the typed storage (`n`) is not updated with the error, the remaining overloads expecting typed parameters, will re-sem the AST, resulting in potentially different errors.

zerbina (approved):
Looks good to me.

The extra complexity to have an error match `untyped` better than `typed` is not worth it for now, I think -- the temporary solution you went with (ambiguity) is fine, in my opinion.

compiler/ast/errorhandling.nim
saem: The `nkArgList` that's created has a bunch of nils in it, it was being triggered by `tsugar`. Downside of a macro API that allows nils/malformed nodes.

IIRC, this snippet from `tsugar` for example causes the issue:
```nim
      proc call5(f: (int {.noSideEffect.} -> int)): int = f(42)
      doAssert call5(x {.noSideEffect.} => x + 1) == 43

# results in:
N:ArgList nid:1618474
0 <nil>
1 <nil>
2 N:Pragma nid:1618475
  0 N:Ident nid:1618476 "noSideEffect" ident.id:104 
```

compiler/sem/sigmatch.nim
saem: Yup, good analysis. It definitely seemed fishy that I wasn't assigning the analysed result, but I didn't get a chance to think through all the consequences. I'll fix that in a separate PR, which combined with this change should slim down `matchesAux` a bit.

compiler/ast/errorhandling.nim
zerbina: > Downside of a macro API that allows nils/malformed nodes.

The macro API doesn't allow for `nil` nodes (assigning `nil` or leaving a `NimNode` uninitialized automatically turns it into a `nkNilLit` node), which hints at a severe bug with the VM.

saem: Looking at it again, I think it stems from `evalTemplateArgs`, which creates a `nkArgList` node and then collects the various parameters within, but I'm not sure how it's getting `nil` values in there. I'll have to investigate that after this PR.

compiler/sem/sigmatch.nim
saem: Follow-on work done in this PR: https://github.com/nim-works/nimskull/pull/1133

compiler/ast/errorhandling.nim
saem: Follow-on work here: https://github.com/nim-works/nimskull/pull/1135

PR #1126, opened 2024-01-21
fix: setup emulated TLS before anything else

zerbina:
## Summary

Move setup of emulated thread-local storage (=TLS) before anything
else. For platforms that require loading `io` procedure from a dynamic
library (via `.dynlib`), this fixes programs immediately crashing at
start-up when using `--tlsEmulation:on`. Windows is one such platform.

## Details

The `initThreadVarsEmulation` call is now emitted at the very start of
the program entry point (`NimMain`), instead of at the end of the
`systemDatInit` procedure.

Modules `import`ed - instead of `include`d - by `system` (such as `io`)
run their module operations before `systemDatInit`. If a thread-local
variable (such as the stack-trace frame pointer or `nimInErrorMode`
global) is accessed by them to thread-local variables, the program
would crash with a nil access.

### DIFF
--- a/compiler/backend/cbackend.nim
+++ b/compiler/backend/cbackend.nim
@@ -67,12 +67,13 @@ import
     containers,
     idioms,
     pathutils,
+    platform,
     ropes
   ]
 
 import std/options as std_options
 
-from compiler/ast/ast import id, newNode
+from compiler/ast/ast import id, newNode, newTree, newSymNode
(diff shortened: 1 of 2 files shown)


saem (approved):
LGTM

alaviss (approved):
It's already passed Windows CI, so let's get it in.

PR #1127, opened 2024-01-21
compiler: restore `.acyclic` annotation where possible

zerbina:
## Summary

Mark cyclic types that cannot form reference cycles at run-time with
`.acyclic`. This reduces both the pressure on the cycle collector as
well as reference counting overhead for the types in question.

## Details

* a comment is added to types for which it's not obvious why they
  cannot form reference cycles
* `ModulfGraph` and `ConfigRef` aren't marked with `.acyclic`, since
  they both store non-final `ref` types or closures, which can
  introduce valid reference cycles

### DIFF
--- a/.github/workflows/leak.yml
+++ b/.github/workflows/leak.yml
@@ -2,12 +2,14 @@ name: Test compiler and tools for memory leaks
 
 on:
   # This is for passing required checks
-  push:
   pull_request:
   merge_group:
 
 jobs:
   leaktest:
+    # Skip this for PRs
+    if: github.event_name != 'pull_request'
+
     name: Compiler and tools leak test
     runs-on: ubuntu-latest
 
--- a/.github/workflows/leak.yml
+++ b/.github/workflows/leak.yml
@@ -2,14 +2,12 @@ name: Test compiler and tools for memory leaks
 
 on:
   # This is for passing required checks
+  push:
   pull_request:
   merge_group:
 
 jobs:
   leaktest:
-    # Skip this for PRs
-    if: github.event_name != 'pull_request'
-
     name: Compiler and tools leak test
     runs-on: ubuntu-latest
 
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1562,7 +1562,9 @@ type
         discard
 
   TNode*{.final.} = object # on a 32bit machine, this takes 32 bytes
-                                    # on a 64bit machine, this takes 40 bytes
+                           # on a 64bit machine, this takes 40 bytes
+    # NOTE: don't mark as `.acyclic`. User-created AST might form cycles, and
+    # nodes can also form reference cycles with ``PSym`` and ``PType``
     typ*: PType
     id*: NodeId  # placed after `typ` field to save space due to field alignment
     info*: TLineInfo
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1256,7 +1256,7 @@ type
     adSemDeprecatedCompilerOptArg   # warning promoted to error
 
   PAstDiag* = ref TAstDiag
-  TAstDiag* = object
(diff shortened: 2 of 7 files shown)


.github/workflows/leak.yml
```diff
@@ -2,14 +2,12 @@ name: Test compiler and tools for memory leaks
 
 on:
   # This is for passing required checks
+  push:
```
zerbina:
```suggestion
```

.github/workflows/leak.yml
```diff
@@ -2,14 +2,12 @@ name: Test compiler and tools for memory leaks
...
   pull_request:
   merge_group:
 
 jobs:
   leaktest:
-    # Skip this for PRs
-    if: github.event_name != 'pull_request'
-
```
zerbina: This needs to be restored prior to merging.

compiler/sem/semdata.nim
```diff
@@ -59,9 +59,8 @@ type
 
   POptionEntry* = ref TOptionEntry
   PProcCon* = ref TProcCon
-  # TODO: This type should be marked acyclic
-  TProcCon* = object ## procedure context; also used for top-level
-                                 ## statements
+  TProcCon* {.acyclic.} = object
+    ## procedure context; also used for top-level statements
```
alaviss, line 63: You should add a reason for why this is acyclic.

compiler/sem/semtypinst.nim
```diff
@@ -121,8 +121,7 @@ proc cacheTypeInst(c: PContext; inst: PType) =
   addToGenericCache(c, gt.sym, inst)
 
 type
-  # TODO: This type should be marked acyclic
-  LayeredIdTable* = ref object
+  LayeredIdTable* {.acyclic.} = ref object
```
alaviss, line 124: Same as above

compiler/utils/btrees.nim
```diff
@@ -16,8 +16,7 @@ const
   Mhalf = M div 2
 
 type
-  # TODO: This type should be marked acyclic
-  Node[Key, Val]  = ref object
+  Node[Key, Val] {.acyclic.} = ref object
```
alaviss, line 19: Same thing

compiler/ast/ast_types.nim
```diff
@@ -1256,7 +1256,7 @@ type
     adSemDeprecatedCompilerOptArg   # warning promoted to error
 
   PAstDiag* = ref TAstDiag
-  TAstDiag* = object
+  TAstDiag* {.acyclic.} = object
```
alaviss, line 1259: Not sure about this one, given that it's stored inside a `PNode` that could harbor a cycle.

zerbina: If a diagnostic stores a tree that references the diagnostic itself then that's a severe bug.

compiler/utils/btrees.nim
zerbina: It's a node in a balanced *tree*. I figured that no cycles being possible in a node tree is self-explanatory.

compiler/ast/ast_types.nim
alaviss: Perhaps a note then?

compiler/sem/semdata.nim
zerbina: Hm, why? The `next` field is explicitly documented as being used for *stacking*.

compiler/sem/semtypinst.nim
saem: Isn't it fair to say here that it's for `nextLayer` and that's a pretty obvious linked list (non-circular)?

saem (approved):

alaviss (approved):

compiler/ast/ast_types.nim
zerbina: I'll add a note.

PR #1128, opened 2024-01-22
niminst: use 7-Zip to create zip archives 

alaviss:
## Summary

This pull switches from using Info-ZIP to 7-Zip for creating Zip
archives. This makes sure that the build works correctly on Windows
regardless of locale.

## Details

While we originally moved away a while back, there are a few issues with
Info-ZIP:

- It uses "legacy" APIs on Windows, and is not Unicode aware on such
systems. This makes it unable to ingest the file list on Windows
depending on locale.
- It is not as widely available on Windows systems as 7-Zip
- Development of Info-ZIP has largely stopped, so the issues we found
are unlikely to be fixed.

The original concerns that lead to moving away from this was that
Info-ZIP has explicit support for omitting some metadata. Reexamining
7-Zip, I found that:

- 7-Zip does not support these extra metadata in the ancient version
used by most Linux distros. Later versions of 7-Zip supports these, but
have them disabled by default.
- While there are some problems with filename encoding if locale does
not support UTF-8, this problem seems to be isolated to Unix and can be
solved by forcing the locale to  `C.utf8` , which is done here.

### DIFF
--- a/.github/workflows/reproducible.yml
+++ b/.github/workflows/reproducible.yml
@@ -30,7 +30,7 @@ jobs:
     #               varies based on outside environment.
 
     # Skip this for PRs
-    if: github.event_name != 'pull_request'
+    #if: github.event_name != 'pull_request'
 
     strategy:
       fail-fast: false
--- a/.github/workflows/reproducible.yml
+++ b/.github/workflows/reproducible.yml
@@ -30,7 +30,7 @@ jobs:
     #               varies based on outside environment.
 
     # Skip this for PRs
-    #if: github.event_name != 'pull_request'
+    if: github.event_name != 'pull_request'
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/reproducible.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     #               varies based on outside environment.
 
     # Skip this for PRs
-    if: github.event_name != 'pull_request'
+    #if: github.event_name != 'pull_request'
```
saem:
```suggestion
    if: github.event_name != 'pull_request'
```

So we remember to change it back.

saem (approved):
If this works please update the PR message (summary/detail), then merge away.

alaviss:
This doesn't work, 7-zip couldn't handle locale changes.

alaviss:
Let's revive this. Info-ZIP can't do Unicode on Windows.

alaviss:
I'll merge this once Windows CI is green.

alaviss:
/merge

Windows CI is finally green :tada:

PR #1129, opened 2024-01-22
shellrunner: add Windows support

alaviss:
## Summary
This pull request adds support for Windows to `std/shellrunner`.

## Details
Most of the code is already there, having originated from osproc. Only a
new API has to be added to osproc to expose the process handle for
Windows.

In the future the process handle API could be expanded to POSIX systems
where pidfd is available (ie. Linux, FreeBSD).


saem (approved):
Cool, and tests are in place thanks because `tshellrunner` doesn't have any platform restrictions. 🎉

PR #1130, opened 2024-01-22
testament: don't close Process before reading exit code

alaviss:
## Summary

On Windows Testament now correctly captures the exit code from commands
(compile, test run, etc).

## Details

On platforms like Windows closing a process before reading the exit code
invalidates the exit code, returning 0, which would falsely signify a
success.

Instead, use  `defer`  to make sure that the object is kept alive until
the end of scope.

### DIFF
--- a/testament/testament.nim
+++ b/testament/testament.nim
@@ -400,6 +400,7 @@ proc callNimCompiler(cmd: string): CompilerOutput =
   verboseCmd(cmd)
   var p = startProcess(command = cmd,
                        options = {poStdErrToStdOut, poUsePath, poEvalCommand})
+  defer: close(p)
   let outp = p.outputStream
   var foundSuccessMsg = false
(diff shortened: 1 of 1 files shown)


testament/testament.nim
```diff
@@ -400,6 +400,7 @@ proc callNimCompiler(cmd: string): CompilerOutput =
   verboseCmd(cmd)
   var p = startProcess(command = cmd,
                        options = {poStdErrToStdOut, poUsePath, poEvalCommand})
+  defer: close(p)
```
saem, line 404:
```suggestion
  # windows requires reading the the exit code (below) prior to closing
  defer: close(p)
```

saem (approved):
LGTM, I left a soft suggestion explaining why we defer in a comment.

I also updated the PR message, please take a read before merging, thanks.

PR #1131, opened 2024-01-22
tests: avoid importing posix on non-POSIX systems

alaviss:
## Summary

Importing posix or modules that are meant to work on POSIX systems will make tests unable to compile on non-POSIX platforms (ie. Windows) due to missing required C headers.

---

## Notes for Reviewers
* Cherry picked from #1123 
* I'm not sure if importing `posix` is required for `tconcepts_issues` (ie. due to unwanted interactions), but I haven't seen complaints from CI so far, nor was there any references to symbols defined in posix.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/tests/lang_experimental/concepts/tconcepts_issues.nim
+++ b/tests/lang_experimental/concepts/tconcepts_issues.nim
@@ -31,7 +31,7 @@ Meow
 joinable: false
 """
 
-import macros, typetraits, os
+import macros, typetraits, os, posix
 
 
 block t5983:
--- a/tests/lang_experimental/concepts/tconcepts_issues.nim
+++ b/tests/lang_experimental/concepts/tconcepts_issues.nim
@@ -31,7 +31,7 @@ Meow
 joinable: false
 """
 
-import macros, typetraits, os, posix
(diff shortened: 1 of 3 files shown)


tests/lang_experimental/concepts/tconcepts_issues.nim
```diff
@@ -31,7 +31,7 @@ Meow
 joinable: false
 """
 
-import macros, typetraits, os, posix
+import macros, typetraits, os
```
zerbina: Until the test is properly reduced, I think the `posix` import has to stay, and the test needs to be disabled for Windows. Refer to https://github.com/nim-lang/nim/issues/6782, from which the test was taken.

When you do so, could you also add a note stating that the `t6782` case needs further reduction?

alaviss: It might be better if you create an issue for that.

I'll drop this commit from the branch.

saem (approved):
Looks good with the change for the concepts test dropped.

PR #1132, opened 2024-01-22
mir: use IDs for more entities

zerbina:
## Summary

Use ID types for representing procedures, globals, and constants during
the mid- and backend phase, replacing `PSym`. This is significant
progress towards:
- a more data-oriented design
- decoupling `sem` and the backend

Together with the introduction of `MirEnv`, this also paves the way for
further MIR and backend improvements.

#### General architecture

- procedures, named constants (i.e., `const`), and globals (both normal
  ones and `.threadvar`s) are all identified with a dedicated ID type
  in the mid- and back-end

- `MirEnv` stores the all non-local, non-ephemeral MIR-related data
  - at the moment, these are the `SymbolTable`s, which map the
    aforementioned IDs to the data associated with the entities (name,
    type, flags, etc.)
  - the data is still represented with `PSym`

- `MirEnv` supports a simple snapshot/checkpoint mechanism for
  iterating over the newest additions or rewinding the environment
- the `MirEnv` is a non-`ref` type; it needs to be passed to all
  routines that wish to modify it or lookup entity-related information
- all non-local entities that the MIR wants to reference must be added
  to the `SymbolTable`s
  - a `PSym` can be added directly to a `SymbolTable`, which returns
    the ID
  - the ID is used to lookup the entity-related data later

#### MIR data types

- `MirNode` uses `ProcedureId`,  `ConstId`, and `GlobalId`, for
  referring to the respective entities, replacing `PSym`
- local entities still use `PSym`, but this is going to change in the
  future

#### AST -> MIR translation

- `mirgen` adds symbols to the environment as it needs their IDs
  - this renders separately scanning the `MirTree`s obsolete,
    `backends` only needs to look for additions to the `MirEnv`
- `mirgen` correctly translates top-level globals itself. `vmjit`
  previously had to scan the `def`s of them for the purpose of
  discovery, necessitating a separate pass
  (`mirbridge.rewriteGlobalDefs`)
- the callee of a VM-evaluated template call (i.e., `getAst(templ())`)
  is translated to an `mnkLiteral` instead of an `mnkProc`

#### Common backend logic (`backends`)

- `backends` uses the checkpoint facility of `MirEnv` to keep track of
  environmental changes; the `Queue` facility is no longer needed and
  thus removed

- `BackendEvent` uses the ID types for identifying procedures and
  constants
- `registerLate` is replaced with the `setModuleOverride` facility
- for backend-like processing that doesn't use the `process` iterator
  (e.g., `vmjit`), the `discover` iterator is provided
- procedures that need incremental code generation are marked with the
  `sfForward` flag by the `process` iterator. This prevents them from
  being queued for normal code generation
- apart from passing around IDs instead of `PSym`s, `backends` stays
  mostly the same

#### General code generation

- same as `MirNode`, `CgNode` also uses ID types instead of `PSym`
- a `MirEnv` instance is owned by the backends' respective context
  type (`cgendata.BModuleList`, `jsgen.PGlobals`, `vmgen.CodeGenCtx`)
- in general, the ID types are passed around instead of `PSym`

#### `cgen`

- code-generator information for procedures, constants, and globals now
  use a `SeqMap` instead of the legacy `PSym`-based `SymbolMap`
- `registerLateProc` replaces the manual access to the `extra` list
- `useHook` replace the manual access to the `hooks` list

#### `jsgen`

- constants and globals also use `Loc` now, making the `PSym` overloads
  for `addrLoc` and `accessLoc` obsolete
- procedures use a dedicated `SeqMap` for associating the mangled name
- only field symbols still use the `PGlobals.names` table
- the `definedGlobals` overload accepting an `openArray` was never
  used; it's removed

#### `vmgen` and `vmjit`

- the entities' ID directly corresponds to their VM table index
  - the `symToIndexTbl` still has to be kept for now, as some non-code-
    generation facilities still depend on it

- basic evaluatibility tests for the code (such as rejecting access to
  non-compile-time globals, use of forwarded procedures, etc.) now
  happens in a dedicated pass (`vmjit_checks`) prior to the MIR phase
  - it's a better architecture, decoupling more JIT-mode related checks
    from the code generator
  - it's also necessary now that globals are eagerly added to the
    environment, regardless of whether they're accessible according to
    the language rules

- the `discoverGlobalsAndRewrite` pass is removed; it's obsolete now
  that `mirgen` performs both actions

#### Code-gen orchestration

- all orchestrators are adapted to the `BackendEvent` changes, as well
  as the changes to their respective code generators
- `cbackend` uses `ProcedureId` instead of `int` (a `PSym.id` value)
  for the `.inline` procedure handling
- `vmbackend` uses `ProcedureId` and `GlobalId` instead of `LinkIndex`
  as the key for its list of procedures and globals 

#### Debug rendering

- both `treeRepr` and the pretty-printer for `MirTree`/`MirNode` are
  updated
- if no `MirEnv` is provided to the pretty-printer, it renders the raw
  ID values instead of looking up their user-provided names
  - this allows using the pretty-printer in situations where no
    `MirEnv` is available

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -4,7 +4,7 @@ import
   std/[
     deques,
     dynlib, # for computing possible candidate names
-    intsets
+    tables
   ],
   compiler/ast/[
     ast,
@@ -23,6 +23,7 @@ import
     mirbodies,
     mirbridge,
     mirconstr,
+    mirenv,
     mirgen,
     mirpasses,
     mirtrees,
@@ -56,31 +57,19 @@ type
       ## if ``true``, indicates that a procedure with a body should not be
       ## treated as imported, even if it's marked as such
 
-  Queue*[T] = object
-    ## Combines a sequence of items with a "read" cursor.
-    data: seq[T]
-    progress: int
-      ## remembers the position until which the items have already
-      ## been processed
-
   DiscoveryData* = object
     ## Bundles all data needed during the disovery of alive, backend-relevant
     ## entities.
-    # XXX: this type does too much. It acts as both a symbol table and
-    #      communication interface. Eventually, it should be split up.
-    seen: IntSet
-      ## remembers the symbol ID of all discovered entities
-
-    procedures*: Queue[PSym]
-    constants*: Queue[PSym]
-    globals*: Queue[PSym]
-    threadvars*: Queue[PSym]
-
+    progress: EnvCheckpoint
+      ## tracks how much of the environment was already processed (e.g.,
+      ## translated, scanned, etc.)
     libs*: seq[LibId]
       ## all dynamic libraries that the alive graph depends on
 
-    additional: seq[tuple[m: FileIndex, prc: PSym]]
-      # HACK: see documentation of the procedure that appends to this list
+    overrides: Table[ProcedureId, FileIndex]
+      ## maps a procedure to the module it needs to be queued with.
+      ## If not overriden, a procedure is queued with the module it's
+      ## discovered from.
 
   BackendEventKind* = enum
     bekDiscovered## a new entity was discovered during MIR processing. This
@@ -105,18 +94,19 @@ type
 
     case kind*: BackendEventKind
     of bekDiscovered:
-      entityId*: int
-        ## the position of the symbol within its ``DiscoveryData`` table
-      entity*: PSym
+      entity*: MirNode
         ## a reference to the discovered entity
     of bekModule:
       discard
     of bekConstant:
-      cnst*: PSym
-        ## the symbol of the constant the event is about
+      cnst*: ConstId
+        ## the ID of the constant
     of bekPartial, bekProcedure, bekImported:
+      id*: ProcedureId
+        ## the ID of the procedure
       sym*: PSym
         ## the symbol of the procedure the event is about
+        ## XXX: only here for convenience, remove it once feasible
       body*: MirBody
 
   WorkItemKind = enum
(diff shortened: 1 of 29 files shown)


compiler/backend/jsgen.nim
```diff
@@ -355,6 +362,13 @@ proc ensureMangledName(p: PProc, s: PSym): lent string =
 
   result = p.g.names[s.id]
 
+proc ensureMangledName(p: PProc, id: ProcedureId): lent string =
+  if id notin p.g.procs:
+    # the mangled named hasn't been generated yet
```
```suggestion
    # the mangled name hasn't been generated yet
```

compiler/vm/vmjit.nim
```diff
@@ -135,8 +122,15 @@ proc updateEnvironment(c: var TCtx, data: var DiscoveryData) =
     #       allocated with `allocConstantLocation` inside `serialize` here
     c.serialize(sym.ast, handle)
 
-    c.complexConsts[i] = handle
+    c.complexConsts.add handle
 
+template preCheck(env: MirEnv, n: PNode) =
+  ## Verifies that `n` can be build and run in JIT mode. If not, aborts the
```
```suggestion
  ## Verifies that `n` can be built and run in JIT mode. If not, aborts the
```

compiler/vm/vmjit_checks.nim
```diff
@@ -0,0 +1,92 @@
+## Implements a validation pass for code running at compile-time or in
+## NimScript/REPL contexts.
+##
+## Future direction: this validation needs to be decoupled from the JIT and
+## moved into the semantical analyisis layer.
```
saem:
```suggestion
## moved into the semantic analyisis layer.
```

compiler/backend/backends.nim
```diff
@@ -843,93 +790,101 @@ iterator process*(graph: ModuleGraph, modules: var ModuleList,
...
+  ## workaround, try to use it as little as possible.
+  discovery.overrides[id] = module
+
+# ----- routines for manual implementation of the backend processing -----
+
+iterator discover*(env: var MirEnv, progress: EnvCheckpoint
+                  ): tuple[s: PSym, n: MirNode] =
+  ## Returns all entities - in an unspecified order - added to `env` since the
+  ## `progress` checkpoint was created. Procedures referenced from constants
+  ## also added to the environment and returnd.
```
saem:
```suggestion
  ## also added to the environment and returned.
```

saem (approved):
Yay, I can see the DOD emerging. :D

Found a few minor typos, nothing particularly pressing.

PR #1133, opened 2024-01-23
sem: avoid excess argument analysis for overloads

saem:
## Summary

Arguments in a call being analysed during overload resolution that
produce an error are no longer repeatedly analysed per overload
candidate.

## Details

Updated  `sigmatch.matchesAux`  to assign the semantically analysed
result to  `n` , the typed AST storage, even upon error. Previously,
errors were not assigned to  `n` , which resulted in repeatedly
analysing the same AST over and over again (performance). Overall there
should be almost no perceptable change.

The one difference is error messages when accidentally putting a space
between a prefix routine call and the parenthesis now emits a tuple
construction error.


zerbina (approved):
The code being more linear makes it a lot easier to read and reason about.

PR #1134, opened 2024-01-23
fix(mirgen): loop-initialized closure locals not being destroyed

zerbina:
## Summary

Fix locals defined within a loop and closed over by an inner procedure
(which lifts them into the closure environment) not properly destroyed
when control-flow reaches the `let` or `var` statement again.

The compiler crashing for `let (a, b) = x` statements where `a`
or `b` are closed over is fixed too.

## Details

Except for `let :env.a = b` statements, `mirgen` treated all `let`/
`var` statements for closure locals as *initializing* assignments (by
emitting an `mnkInit` statement).

This is correct for closure locals defined outside of loops, but not
for those defined within loops, as control-flow possibly reaches the
assignment multiple times, but location having been destroyed in
between, (the closure environment is the same through the whole
procedure invocation).

Since `mnkInit` assignments are (correctly) turned into raw
assignments, the old value wasn't destroyed on re-assignment. `mirgen`
now keeps track of whether a statement is enclosed by a loop (via the
`inLoop` counter), and if the statement is, emits a normal assignment
(`mnkAsgn`) instead of an initializing one.

### Compiler crash

For `let (a, b) = x`, `mirgen.genVarTuple` always tried to access the
identifiers nodes as if they were symbols. If they're closed over, this
results in a field defect, since then they're `nkDotExpr` nodes. The
type is now taken from the node itself.

The added test also covers this.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1561,7 +1561,7 @@ proc genVarTuple(c: var TCtx, n: PNode) =
       # generate the assignment:
       c.buildStmt mnkInit:
         genOperand(c, lhs)
-        c.subTree MirNode(kind: mnkPathPos, typ: lhs.sym.typ,
+        c.subTree MirNode(kind: mnkPathPos, typ: lhs.typ,
                           position: i.uint32):
           c.use val
 
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -144,6 +144,8 @@ type
     sp: SourceProvider
 
     numLabels: int ## provides the ID to use for the next label
+    inLoop: int
+      ## > 0 if the current statement/expression is part of a loop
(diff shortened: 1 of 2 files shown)


alaviss (approved):
Looks good, just a quick question:

> There was also a separate problem with `mirgen.genVarTuple`, where
the `.sym` field of `lhs` was accessed even if it's an `nkDotExpr`
(resulting in a field defect error). The type is now taken from the
node itself.

Are the added tests covering this as well?

zerbina:
> Are the added tests covering this as well?

Yep, they do (the `let (o2, i) = ...` ones, to be precise). The tests are actually how I discovered the bug.

The PR description is a bit unclear in this regard, I'll update it.

zerbina:
/merge

I've updated the relevant parts of the commit message.

PR #1135, opened 2024-01-24
sem: `auto`/inferred lambdas generate valid AST

saem:
## Summary
`auto` parameter and return types resulted in inferred lambda AST being
generated with `nil`s in the parameter position. This is now no longer
the case and the AST generated is well formed.

## Details

`seminferredlambda` was incorrectly copying the AST out of the `tyProc`
type into the parameter position; that AST isn't well formed. This
resulted in `nil` nodes being inserted for the parameter and return in
the output AST. When traversing it for errors, this would generate a
NPE
during `buildErrorList`. A `nil` checking guard was put in place in
order to work around the problem.
see: https://github.com/nim-works/nimskull/pull/1124

Now the AST is well formed, and the check has been removed.
Additionally, some associated changes have been made:

- tracing coverage expansion to diagnose the issue and confirm the fix
- noting potentially the same issue in `liftdestructors`

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -2182,7 +2182,7 @@ proc semInferredLambda(c: PContext, pt: TIdTable, n: PNode): PNode {.nosinks.} =
   n[genericParamsPos] = c.graph.emptyNode
   # for LL we need to avoid wrong aliasing
   n[paramsPos] = newNodeI(nkFormalParams, n[paramsPos].info, n.typ.n.len)
-  for i, p in n.typ.n:
+  for i, p in n.typ.n.pairs:
     n[paramsPos][i] =
       case i
       of 0: # return type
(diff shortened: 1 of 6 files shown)


compiler/sem/semstmts.nim
```diff
@@ -2180,15 +2181,21 @@ proc semInferredLambda(c: PContext, pt: TIdTable, n: PNode): PNode {.nosinks.} =
   n[namePos].sym = s
   n[genericParamsPos] = c.graph.emptyNode
   # for LL we need to avoid wrong aliasing
-  let params = copyTree n.typ.n
-  n[paramsPos] = params
+  n[paramsPos] = newNodeI(nkFormalParams, n[paramsPos].info, n.typ.n.len)
+  for i, p in n.typ.n:
```
```suggestion
  for i, p in n.typ.n.pairs:
```
A very soft suggestion. I think the iterator being explicit helps readability.

zerbina (approved):
Looks good to me.

Also, thanks for making the `TType.node` documentation a proper list, it really was hard to read.

PR #1137, opened 2024-01-24
mir: don't use `mnkPNode` for `except` branches

zerbina:
## Summary

Translate `except T as e` (where `T` is an imported exception) to
`(Except (Local "e") ...)` instead of `(Except (PNode ...) ...)`. This
removes a usage of `mnkPNode`, and it also makes the existence of the
local visible at the MIR level.

## Details

* `mnkPNode` was only intended as a transition helper
* instead of a `cnkBinding`, the code generators (only `jsgen` at the
  moment) can identify the imported exception handler by the
  `cnkLocal` node

### DIFF
--- a/compiler/backend/cgirgen.nim
+++ b/compiler/backend/cgirgen.nim
@@ -338,18 +338,11 @@ proc tbExceptItem(tree: MirBody, cl: var TranslateCl, cr: var TreeCursor
                  ): CgNode =
   let n {.cursor.} = get(tree, cr)
   case n.kind
-  of mnkPNode:
-    assert n.node.kind == nkInfix
-    assert n.node[1].kind == nkType
-    assert n.node[2].kind == nkSym
-    # the infix expression (``type as x``) signals that the except-branch is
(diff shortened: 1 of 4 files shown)


zerbina:
The goal is to eventually use the (not yet available) `(Except (Type) ... (Local) ...)` form for all exceptions. Besides simplifying the syntax, this would also remove one usage of `getCurrentException` (having to track the current exception in a user-accessible way is the source of a significant amount of complexity in the backend).

saem (approved):

saem:
/merge

---

> this would also remove one usage of `getCurrentException` (having to track the current exception in a user-accessible way is the source of a significant amount of complexity in the backend).

I've been thinking about this for some time, and I struggle to see much advantage of not having matching based `except` blocks and dropping the whole `getCurrentException` approach. Maybe it made more sense when there were all sorts of exception handling strategies, but I'm not sure what the draw is from a user perspective at this point.

zerbina:
/merge

The CI run failure looks to have been caused by spurious network issue.

PR #1138, opened 2024-01-24
fix(vmbackend): run-time errors with non-zero-based sets

zerbina:
## Summary

Fix a bug with the VM backend's constant data serialization where
having and using a constant `set` with a non-zero lower bound resulted
in:
* set operations on the set not working as expected
* runner crashes at startup

## Details

The de-serialization logic in `vmrunner` expects zero-based `set`
construction elements, but the serialization logic in `packed_env`
didn't ensure that the element values are zero-based.

`PackedEncoder` now requires access to a `ConfigRef`, which is used for
subtracting the lower bound from all element values (making them zero
based).

### DIFF
--- a/compiler/vm/packed_env.nim
+++ b/compiler/vm/packed_env.nim
@@ -21,7 +21,11 @@ import
   compiler/ast/[
     ast_query,
     ast_types,
-    lineinfos
+    lineinfos,
+    types
+  ],
+  compiler/front/[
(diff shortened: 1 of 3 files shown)


saem (approved):
Carrying the `ConfigRef` around is a bit of a bummer, but the refactor for that is well out of scope for this bug fix.

PR #1139, opened 2024-01-25
mir: separate constants from constant data

zerbina:
## Summary

In the mid- and backend, there's now a clear distinction being made
between *constants* and *constant data*. A *constant* (i.e., a `const`)
links a name/interface with *constant data*, while *constant data* is
just the raw data.

Storage of constant data is now also decoupled from `TSym.ast`,
preparing for the introduction of a dedicated constant data IR.

#### General architecture

* same as before, constant data is represented by a `PNode` AST
* every complete `PNode` AST representing constant data is identified
  by the new `DataId`
* all constant data the mid- and backend know about is stored in
  `MirEnv` (in a `DataTable`)
* the association between a `DataId` and the data is *bi-directional*
  (a `DataId` maps to a constant expression and a constant expression
  maps to a `DataId`)
* each user-defined constant (`ConstId`) is associated with a `DataId`
* when discovering user-defined constants, `backends.process` and
  `backends.discover` register the data with the environment and link
  it with the constant

#### Rationale

Separating constants from their data allows two constants to share the
same `PNode` AST internally (at least in the backend). It also makes it
easy for the mid-end to create new constant data (e.g., by lifting
constant expression from procedure bodies) itself -- no name nor
interface has to be associated with the data in this case.

#### Implementation

* `DataTable` is a customized version of `BiTable`
* the `PNode` ASTs representing the constant expression are compared by
  their structure
* using structural comparison means that `Obj(a: 0)` and `Obj(b: 0)`
  are treated as being different, even though they represent the same
  data. While not critical, a normalization pass is going be needed
  here, eventually
* the `DataTable` also supports the basic rewind mechanism employed by
  `MirEnv`
* as an interim solution, a separate table is used for mapping
  constants to their body (`MirEnv.bodies`)

#### Code generation

* instead of querying `TSym.ast`, `cgen` and `jsgen` lookup the body
  for constant through `MirEnv`
* constant data in the VM maps directly to the `DataTable` now (that
  is, each item in `DataTable` maps to a complex VM constant). `vmgen`,
  `vmjit`, and `vmbackend` are adjusted accordingly

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -20,6 +20,7 @@ import
     options
   ],
   compiler/mir/[
+    datatables,
     mirbodies,
     mirbridge,
     mirconstr,
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -818,6 +818,7 @@ iterator process*(graph: ModuleGraph, modules: var ModuleList,
       # future direction: the body of the constant (i.e., the value
       # expression) will be transformed to its MIR representation here
       discoverFromValueAst(env, env[item.cnst].ast)
+      env.setData(item.cnst, env.data.getOrPut(env[item.cnst].ast))
       # we cannot report (i.e., yield) right away, the discovered dependencies
       # have to be reported first
       queue.prepend(module, WorkItem(kind: wikReportConst, cnst: item.cnst))
@@ -879,8 +880,10 @@ iterator discover*(env: var MirEnv, progress: EnvCheckpoint
   ## processing -- don't use it in conjunction with the ``process`` iterator.
   for id, it in since(env.constants, progress.consts):
     # first discover and report the procedures referenced by the constant's
-    # data, allowing
+    # data. This ensures that the callsite can rely on all the constant's
+    # dependencies existing in the environment
     discoverFromValueAst(env, astdef(it))
+    env.setData(id, env.data.getOrPut(astdef(it)))
     yield (it, MirNode(kind: mnkConst, cnst: id))
 
   for id, it in since(env.procedures, progress.procs):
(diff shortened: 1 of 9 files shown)


compiler/backend/backends.nim
```diff
@@ -879,8 +880,10 @@ iterator discover*(env: var MirEnv, progress: EnvCheckpoint
   ## processing -- don't use it in conjunction with the ``process`` iterator.
   for id, it in since(env.constants, progress.consts):
     # first discover and report the procedures referenced by the constant's
-    # data, allowing
+    # data. This ensures that the callsite can rely on all the constant's
+    # dependencies existing in the environment
```
zerbina, line 885: There's no change in behaviour here, I just forgot to finish the sentence in the original PR.

saem (approved):
LGTM

Is the idea that eventually `DataTable` might get some out into a small module of its own?

zerbina:
> Is the idea that eventually `DataTable` might get some out into a small module of its own?

Yep, that's the plan, although thinking about it again, it might be a better idea to move it into its own module already. I often tend to put too many related-in-usage things into the same module, which is often detrimental in the longer term.

saem:
> Yep, that's the plan

Sounds good.

> although thinking about it again, it might be a better idea to move it into its own module already. I often tend to put too many related-in-usage things into the same module, which is often detrimental in the longer term.

Up to you, I think it's fine to merge. I personally think modules are decision encapsulation boundaries, so I don't get too fussed unless the public interface is starting to get ugly/too big. Could just as well pull it out in the next PR.

zerbina:
/merge

I've moved the `DataTable` to its own module.

PR #1140, opened 2024-01-26
mirgen: lift constant expressions

zerbina:
## Summary

Introduce anonymous constants to the MIR and use them to lift fully
constant expressions into separate constants during AST -> MIR
translation.

This speeds up the compiler, improves efficiency of the generated
code, and also fixes a memory leak affecting the C backend.

### Anonymous constants

Anonymous constants are constants that don't have a name/interface
attached -- they're data only. `ConstId` represents both user-defined
and anonymous constants, by using the value of the most-significant bit
to distinguish between them: if the bit is set, the lower 31 bit
represent a `DataId`, if it's not set, the `ConstId` is the ID of a
user-defined constant.

`toConstId`, `isAnon`, and `extract` are the procedure for
introspection and creation of anonymous constants. `MirEnv.dataFor`
transparently returns the `DataId` for anonymous constants.

### Constant expression lifting

* constant `set` constructions are always lifted (mirroring what both
  `cgen` and `jsgen` did)
* non-empty, constant-folded `seq` constructions (`nkBracket`) are
  always lifted (mirroring what `cgen` did)
* non-empty constant tuple, object, and array construction are lifted
  only when used in a non-sink context (e.g., as an argument to a non-
  `sink` parameter)
* due to a limitation with `DataTable`, hidden conversions being part
  of an expression make it non-constant for now (achieved with a custom
  `isDeepConstExpr` procedure)

Lifting constant construction appearing in `sink` contexts would alter
semantics, as constants have to copied from (which might result in a
`=copy` call).

Since `mirgen` now does the lifting of `seq` construction, the
workaround of treating MIR `seq` constructions (`mnkConstr` of type
`tySequence`) as constants (because `cgen` lifted them into immutable
C constants) is no longer needed and thus removed.

### Fixes

* with the C backend, using constant-evaluated constructions of
  `seq[T]` where `T` is an aggregate type - `static: @[(1, 2)]`, for
  example - led to a copy and the original `seq` never being freed
  (refer to `tconst_seq_leak`). Removal of the workaround in
  `injectdestructors` fixes this
* `MirEnv.rewind` used the wrong checkpoint for rewinding the `bodies`
  table, which could result in compiler crashes
* `tsplit_into_seq` works now, but not because the underlying issue
  is fixed. Constants `seq` construction not requiring copies only
  means that the problematic `=copy` hook is not used

### `cgen`

* lifting of constant expressions (`exprComplexConst`, `genSetNode`) is
  removed; `mirgen` does this now
* the `dataCache` table is removed. Its most prominent user was the
  expression lifting, and it largely duplicates `DataTable`. The other
  users of `dataCache` use different strategies now:
	* caching default values uses a `SigHash`-based table
	* string values use a dedicated `StrNode` table (using `string`
    directly would require copying the string)
	* `nil` closure literals are expanded to a local C `const`
* anonymous constants are *defined* (not declared) in each C file
  they're used in

### `jsgen`

* lifting of constant `set` construction is removed (it's done by
  `mirgen` now)
* a JavaScript global is emitted for every anonymous constant on use

### `vmgen`

* the code generator and backend already use `DataId` everywhere, so no
  change is necessary
* the `packed_env` packing logic for `nkNilLit` has to work around
  unexpanded `nil` closure literals, by expanding the literal into a
  tuple construction itself. Unexpanded `nil` closure literals didn't
  reach there previously, but now they do

### Performance

The changes have a significant positive impact on:
* compiler speed
* efficiency of the generated code (all backends are affected)

The main reasons for both are that:
* lifting constant expressions in `mirgen` fixes the performance
  regression with field checks introduced by
  `78082ead0e2c5d15a1928359b7dd81dc48925d7a`, where C constants were
  copied into locals, which neither GCC nor MSVC are able to optimize
  away
* constant expression are now also lifted when using the VM and
  JavaScript backends

Compiler speed also improves because of to the reduction in MIR code,
as less MIR code means:
* less nodes that MIR passes need to scan/inspect/analyze/etc.
* less nodes that the code generators need to generate code for

Benchmarking the compiler compiling the compiler at
`3eb7c736384de0b62c88c0dcedbb910fce47f190`  (with 
`hyperfine --warump 1 "nim c --compileOnly --verbosity:0 --hints:off --warnings:off compiler/nim.nim"`
)
yields:
```
before: 14.54 s ± 0.14  (100%)
now:    13.65 s ± 0.05  ( 94%)
```

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -602,20 +602,6 @@ func discoverFrom*(env: var MirEnv, decl: PNode) =
     else:
       unreachable(n.kind)
 
-func discoverFromValueAst(env: var MirEnv, ast: PNode) =
-  ## Discovers new routines from `ast`, which is an AST representing a value
-  ## construction expression.
-  case ast.kind
-  of nkSym:
-    let s = ast.sym
-    if s.kind in routineKinds:
-      discard env.procedures.add(s)
-  of nkWithSons:
-    for n in ast.items:
-      discoverFromValueAst(env, n)
-  of nkWithoutSons - {nkSym}:
-    discard "nothing to do"
-
 func queue(queue: var WorkQueue, id: ProcedureId, prc: PSym, m: FileIndex) =
   ## If eligible for processing and code generation, adds `prc` to
   ## `queue`'s queue.
@@ -818,7 +804,7 @@ iterator process*(graph: ModuleGraph, modules: var ModuleList,
       # scan the constant for its dependencies
       # future direction: the body of the constant (i.e., the value
       # expression) will be transformed to its MIR representation here
-      discoverFromValueAst(env, env[item.cnst].ast)
+      scanExpr(env, env[item.cnst].ast)
       env.setData(item.cnst, env.data.getOrPut(env[item.cnst].ast))
       # we cannot report (i.e., yield) right away, the discovered dependencies
       # have to be reported first
@@ -883,7 +869,7 @@ iterator discover*(env: var MirEnv, progress: EnvCheckpoint
     # first discover and report the procedures referenced by the constant's
     # data. This ensures that the callsite can rely on all the constant's
     # dependencies existing in the environment
-    discoverFromValueAst(env, astdef(it))
+    scanExpr(env, astdef(it))
     env.setData(id, env.data.getOrPut(astdef(it)))
     yield (it, MirNode(kind: mnkConst, cnst: id))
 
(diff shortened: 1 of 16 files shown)


saem (approved):
This was really easy to read, partly I'm getting more familiar with the MIR, but there are plenty of broad simplifications happening.

zerbina:
The `tsplit_into_seq` test was unexpectedly successful.

Investigating why this is the case showed that it's a side-effect of `var x = @[]` not expanding to a call of the seq's `=copy` hook. In the case of the `tsplit_into_seq` test, the lifted copy hook was responsible for the code generation failure, meaning that the test works now (even though the code generator limitation underlying the previous failure still exists).

PR #1142, opened 2024-01-27
fix(vm): access check not failing

zerbina:
## Summary

Fix a bug with the VM's pointer-to-cell mapping that allowed out-of-
bounds reads or writes of empty `seq`s to crash the VM, instead of
resulting in a proper memory access error.

## Details

The fundamental problem was with unclear responsibilities: callers of
`mapPointerToCell` and `mapInteriorPointerToCell` expected both to
handle `nil` pointers, while the procedures expected their callers to
do so.

At the moment, the `mapX` procedure consider free cells (for which
`VmCell.p == nil`), which means that trying to map `nil` pointer values
could result in a valid `CellId` being returned! A `LocHandle` having a
non-`-1` `CellId` means that only the cell's data pointer (`nil`, in
this case) and the user pointer need to match (which in this specific
case, they did) for the access checks to not reject the access.

The dangerous `LocHandle` could only be created through indexing
into a `seq`, as the `makeLocHandle` overload for arbitrary user-
pointers properly guarded against `nil` pointers. If the VM was built
with assertions enabled, merely assigning such handle to a register
(incorrectly) resulted in an assertion failure.

When the VM is built with assertions disabled, reading from or writing
to the location resulted in the VM crashing with a segmentation
fault / `nil` dereference error.

The incorrect assertion (forbidding `nil` handles in registers is
wrong) is removed, and `mapPointerToCell`/ `mapInteriorPointerToCell`
now always return a lookup failure for `nil` pointers, fixing the
two problems.

Two tests for situations where a crash would previously happen are
added.

### DIFF
--- a/tests/vm/tempty_seq_access_bug.nim
+++ b/tests/vm/tempty_seq_access_bug.nim
@@ -1,0 +1,36 @@
+discard """
+  description: '''
+    Regression test for an internal issue with the VM's pointer-to-cell
+    mapping that allowed
+  '''
+  targets: vm
+  matrix: "--boundChecks:off"
+  outputsub: "trying to access a location outside of the VM's memory"
+  exitcode: 1
+"""
+
+# this test is written against internal implementation details of the VM as it
+# worked when this test was written. The current VM could differ.
+
+proc test() =
+  # test in a procedure so that the locations have local lifetime
+
+  # allocate six memory cells, 3 for the seqs, 3 for their payloads:
+  var
+    a = newSeq[int](1)
+    b = newSeq[int](1)
+    c = default(seq[int]) # empty seq
+
+  # free two cells (by overwriting the seqs with empty ones):
+  a.newSeq(0)
+  b.newSeq(0)
+
+  proc write(i: var int) =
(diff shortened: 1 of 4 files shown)


zerbina:
It's only tangentially related to this PR, but in the future, I want to move the VM to use its own virtual address space -- at the moment, all user-visible pointers are addresses in the host's address space.

Since the VM has no control over the host address returned by the allocator the VM allocates memory with, the access checks are limited in how they work. At the moment, a `O(n)` linear search over all memory cells is used by the pointer-to-cell mapper, which is quite inefficient.

tests/vm/tempty_seq_access_bug.nim
```diff
@@ -0,0 +1,36 @@
+discard """
+  description: '''
+    Regression test for an internal issue with the VM's pointer-to-cell
+    mapping that allowed
```
saem:
```suggestion
    mapping that allowed out-of-bound writes.
```

saem (commented):
Nice.

I noticed one incomplete description in the tests, otherwise looks good.

PR #1143, opened 2024-01-28
transf: improve `case` statement transform

zerbina:
## Summary

Improve the `transf` logic for `case` statements. It now drops
unused `of` branches (those without labels) and replaces `case`
statements that - once the other transformations took place - only
consist of an `else` branch.

## Details

* `mirgen` previously took care of dropping unreachable `of` branches,
  but `transf` is better suited for doing so
* a `case` statement/expression like `case x else: body` is turned into
  `(discard x; block: body)`
  * the `discard` makes sure that side-effects of `x` are evaluated and
    that the usage of `x` is preserved
  * the `block` makes sure that the scope of `body` doesn't change

Since `closureiters` takes place after the main `transf` pass but
before `mirgen`, it now also benefits from the elision of unused
`of` branches (i.e., less code to process).

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1752,24 +1752,12 @@ proc genCase(c: var TCtx, n: PNode, dest: Destination) =
   ## straightforward
   assert isEmptyType(n.typ) == not dest.isSome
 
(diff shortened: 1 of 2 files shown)


saem (approved):
That's a nice improvement.

PR #1146, opened 2024-01-30
pragmas: unknown hint/warning name is not an error

zerbina:
## Summary

By default, providing the name of an unknown hint or warning to the 
`hint` ,  `warning` ,  `hintAsError` , or  `warningAsError`  pragmas is
now a warning rather than error. This makes it easier to write forward-
and backward-compatible code.

Two new warnings are introduced for this:  `UnknownHint`  and 
`UnknownWarning` .

## Details

Apart from reporting a warning rather than an error, the  `processNote` 
procedure, which is responsible for processing, for example, 
`hint[A]: on`  pragmas is improved:

* the procedure's documentation is improved, now describing what it does
and expects
* the boolean expression is always evaluated, even when the name of the
hint/warning is not valid
* an input AST modification is removed and proper error wrapping is
used
*  `handleNote`  doesn't capture  `n` , removing the unnecessary
allocation of a closure environment

### Tests

Two tests for the new warnings are added, and the existing 
`tinvalidpragma`  test is updated to also test the malformed pragma
expression situations, increasing test coverage of  `processNote`  's
error handling.

### DIFF
--- a/compiler/ast/report_enums.nim
+++ b/compiler/ast/report_enums.nim
@@ -772,6 +772,8 @@ type
     rsemUntypedParamsFollwedByMoreSpecificType
     rsemBindDeprecated
     rsemObservableStores       = "ObservableStores"
+    rsemUnknownHint            = "UnknownHint"
+    rsemUnknownWarning         = "UnknownWarning"
     rsemUseOfGc                = "GcMem" # last !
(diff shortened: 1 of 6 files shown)


saem (approved):
Nice, this should make things a little less painful for folks.

PR #1147, opened 2024-01-30
disallow string range as `of` label

zerbina:
## Summary

Using ranges of strings (e.g.,  `"a".."c"` ) in  `of`  branches wasn't
rejected, but since both the C and JavaScript backends expect them to
not exist, it's likely that string ranges being allowed is
unintentional. An error is now reported for string ranges used as  `of` 
labels.

## Details

* the `rsemStringRangeNotAllowed` report is added
* the callers of  `semBranchRange`  doesn't implement error propagation
yet, so  `localReport`  is used

### Tests

* the `tcasestmt` test is changed to work with all targets
* the float slice-list test from  `vm/tcasestmt`  is merged into the
general  `tcasestmt`  test


saem:
Fixed a minor typo in the PR body, reviewing the change.

saem (approved):
Error message looks good and the combined tests with extra coverage is much better.

PR #1148, opened 2024-01-30
vm: implement `case string` in bytecode

zerbina:
## Summary

Remove the dedicated  `string`  slice-list matching support from the VM
and replace it with an in-bytecode implementation. This gives more
control to the code generator and removes the last internal usage of 
`cnstString` , preparing for the latter's removal.

## Details

An  `of`  branch of a  `case`  matching over a string is now translated
into a sequence of  `LdConst` + `EqStr` + `TJmp`  (one per label),
replicating the behaviour previously implemented by the  `Branch` 
opcode.

While the bytecode implementation is a bit slower and also results in
slightly larger VM executables, having less high-level string operations
in the VM makes it easier to evolve.

The string slice-list matching is removed from the  `Branch`  opcode
implementation, and  `cnstSliceListStr`  together with all its usages is
removed.


saem (approved):
Dropping the specialization saves us a bit of code, as well.

PR #1150, opened 2024-01-31
bundle the `vmrunner` with releases

zerbina:
## Summary

Without the `vmrunner` executable, the runner has to be built manually
in order to run `.nimbc` executables -- the `nim vm -r` command also
doesn't work.

The `vmrunner` executable is now part of the standard installation.


alaviss (approved):

saem (approved):
nice addition.

PR #1153, opened 2024-02-01
std/io, os: stop `newWideCString` buffer overflow

alaviss:
## Summary

Fix buffer overflow in the  `std/io`  and  `std/os`  modules related to
incorrect usage of  `newWideCstring` .

## Details

`newWideCString` 's binary form, 
`proc newWideCString*(source: cstring, L: int): WideCStringObj` ,
results in buffer overflows within the  `std/io`  and  `std/os` 
modules. This change revises the erroneous usage of  `newWideCString` ,
with an empty string ( `""` ) as the  `source`  parameter in these
modules, with the single parameter version for preallocating a wide
string.

The 2-parameter form of newWideCString is meant to take a  `cstring` 
and its  `length` , then convert them to wide form. As such this length
must be the exact length of the given string.

To prevent incorrect usage in the future, this PR also documents the
various  `newWideCString`  variants.

### DIFF
--- a/lib/system/widestrs.nim
+++ b/lib/system/widestrs.nim
@@ -121,9 +121,12 @@ iterator runes(s: cstring, L: int): int =
     yield result
 
 proc newWideCString*(size: int): WideCStringObj =
+  ## Create an empty `WideCStringObj` with the given buffer size in
+  ## UTF-16 code units.
   createWide(result, size + 1) # +1 for the zero terminator
 
 proc newWideCString*(source: cstring, L: int): WideCStringObj =
+  ## Take a `cstring` and its length `L` in bytes, then returns a copy in wide form.
   createWide(result, L + 1) # +1 for the zero terminator
   var d = 0
   for ch in runes(source, L):
@@ -144,11 +147,13 @@ proc newWideCString*(source: cstring, L: int): WideCStringObj =
   result[d] = Utf16Char(0)
 
 proc newWideCString*(s: cstring): WideCStringObj =
+  ## Returns a copy of the string `s` in wide form. `s` should be NUL-terminated.
   if s.isNil: return nullWide
 
   result = newWideCString(s, s.len)
 
 proc newWideCString*(s: string): WideCStringObj =
+  ## Returns a copy of the string `s` in wide form.
   result = newWideCString(cstring s, s.len)
 
 proc `$`*(w: WideCString, estimate: int, replacement: int = 0xFFFD): string =
--- a/lib/pure/os.nim
+++ b/lib/pure/os.nim
@@ -1373,13 +1373,13 @@ when not defined(nimscript):
     elif defined(windows):
       var bufsize = MAX_PATH.int32
       when useWinUnicode:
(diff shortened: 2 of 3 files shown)


saem (commented):
OK, so here is my take of the changes and broader context of `newWideCString`:

- `newWideCString(cstring, int): WideCStringObj`, which uses `iterator runes` internally has a buffer overflow issue
  - basically `L` should be less than the length of `sources`, or
  - `runes` should check for null terminators in its looping
- this change switches away from using the problematic two parameter version to the safer single parameter version
  - using the two parameter version was likely an attempt to optimize/preallocate

Do I have that right?

alaviss:
> `newWideCString(cstring, int): WideCStringObj`, which uses `iterator runes` internally has a buffer overflow issue
> 
>     * basically `L` should be less than the length of `sources`, or
> 
>     * `runes` should check for null terminators in its looping

I think the current behavior is correct, as it treats `L` as the sole source of truth for buffer size.

> this change switches away from using the problematic two parameter version to the safer single parameter version
> 
>     * using the two parameter version was likely an attempt to optimize/preallocate

Yep this is correct.

saem:
> I think the current behavior is correct, as it treats `L` as the sole source of truth for buffer size.

Meh, I think the API is problematic and it's too easy to make a mistake, not to mention the lack of docs. But all that stuff is out of scope for this PR.

Thanks for the clarifications, I'll take a crack at the PR message.

saem (commented):
Revised the PR message, take a look. After having revised it, I think it's fair to say we should put in at least a minimal doc comment for the binary version of `newWideCString`. Something that captures the lesson we learned here as part of this PR.

alaviss:
> After having revised it, I think it's fair to say we should put in at least a minimal doc comment for the binary version of `newWideCString`

Do you want that in this PR or do we do a follow up?

saem:
> > After having revised it, I think it's fair to say we should put in at least a minimal doc comment for the binary version of `newWideCString`
> 
> Do you want that in this PR or do we do a follow up?

I think it should be part of this PR, sorry I meant to specify it but got lost in editing.

lib/system/widestrs.nim
```diff
@@ -144,11 +147,13 @@ proc newWideCString*(source: cstring, L: int): WideCStringObj =
   result[d] = Utf16Char(0)
 
 proc newWideCString*(s: cstring): WideCStringObj =
+  ## Returns a copy of the string `s` in wide form. `s` should be NUL-terminated.
```
saem, line 150:
```suggestion
  ## Returns a copy of the string `s` in wide form. `s` should be NULL-
  ## terminated.
```

null, as in null byte, is double 'l'; it's a small thing, as the meaning is still clear.

alaviss: `NUL` is the correct spelling for the ASCII null character. Apparently it's an [abbreviation for null](https://en.wikipedia.org/wiki/ASCII#Control_code_chart).

saem: LoL, I guess it needed abbreviating.

PR #1157, opened 2024-02-02
backend: use MIR for representing constant expressions

zerbina:
## Summary

Instead of `PNode`, use a subset of the MIR for representing constant
expression in the mid- and back-end. Apart from removing a significant
amount of `PNode` usage from the backend, using a dedicated IR for
constant expression paves the way for applying transformation and
lowering passes to them.

### Constant expression IR

* a subset of the MIR is used (refer to the `Constant Expressions`
  section in `doc/mir.rst`)
* in the subset, recursion is allowed
* using a subset of the MIR allows for re-using most of the existing
  MIR facilities (changesets, traversal routines, etc.)

### Translation

* the translation is implemented as a single-pass traversal over the
  constant expression AST
* since procedures referenced by constant expression are now registered
  with the `MirEnv` as they're encountered, the separate scanning pass
  (`scanExpr`) is obsolete
* the translation also handles `nkHiddenStdConv` and `nkHiddenSubConv`
  nodes, making the custom `mirgen.isDeepConstExpr` overload, which
  treated these nodes as non-constant, obsolete
* the constant-expression-to-`CgNode` translation (`compat.translate`)
  is updated to operate on the MIR constant expression

### VM serialization

The VM itself is not affected, but the logic for initializing memory
locations directly from constant expressions is. Initializing a VM
memory location directly from a MIR constant expression is implemented
by the new `vmserialize` module.

Apart from the loading of constant data in `vmjit.updateEnvironment`
(which must use `vmserialize`, since it now gets passed a `MirTree`),
there were two other situations where a VM memory location is
initialized from a `PNode` expression:
1. for `static` macro parameter initialization (`putIntoReg`)
2. for setting the value of a global via the compiler API
  (`setGlobalValue`)

Both are changed to first translate the `PNode` AST to a MIR constant
expression, which the register or memory location is then initialized
with. This has the benefits that:
- initializing VM locations from constant expression works the same
  everywhere
- procedures referenced from the AST are automatically registered with
  the `MirEnv` and JIT environment
- the `vmcompilerserdes` procedures for initializing a VM location from
  `PNode` AST become obsolete

Since they're now obsolete, the `vmcompilerserdes` serialization
routines are all removed. This has an important implication: a
`LinkerData` instance doesn't have to be carried around by `vmdef.TCtx`
anymore. Nothing is removed here yet -- the cleanup is left to follow-
up work.

Finally, the serialization of constant expressions in `packed_env` is
also updated to operate directly on MIR constant expressions.

### Misc

* speed up `mirconstr.pop` in the full-buffer-to-empty-buffer case
  (which happens for `constDataToMir`)

### DIFF
--- a/compiler/mir/datatables.nim
+++ b/compiler/mir/datatables.nim
@@ -94,13 +94,10 @@ proc cmp(a, b: ConstrTree): bool =
     return false
 
   # both trees are known to have the same length at this point
-  var i = 0
   # traverse both trees simultaneously and look for a divergence:
-  while i < a.len:
+  for i in 0..<a.len:
     if a[i] != b[i]:
       return false # divergence -> not equal
-
-    inc i
 
   # the cursor reached the end; the trees are equal
   result = true
--- a/compiler/mir/datatables.nim
+++ b/compiler/mir/datatables.nim
@@ -1,5 +1,5 @@
 ## Implements a bi-directional table that associates constant data
-## (represented by ``PNode`` AST) with a ``DataId`` and vice versa.
+## (represented by a subset of the MIR) with a ``DataId`` and vice versa.
 
 import
   std/[
@@ -20,7 +20,7 @@ import
   ]
 
 type
-  ConstrTree* = PNode
+  ConstrTree* = MirTree
   DataTable* = object
     ## A bi-directional table that associates constant expressions with an ID.
     vals: Store[DataId, ConstrTree]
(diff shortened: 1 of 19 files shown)


compiler/mir/datatables.nim
```diff
@@ -32,40 +32,78 @@ type
...
+    return false
+
+  # both trees are known to have the same length at this point
+  var i = 0
+  # traverse both trees simultaneously and look for a divergence:
+  while i < a.len:
+    if a[i] != b[i]:
+      return false # divergence -> not equal
+
+    inc i
```
Clyybber: could use a for loop here
```suggestion
  # traverse both trees simultaneously and look for a divergence:
  for i in 0..<a.len:
    if a[i] != b[i]:
      return false # divergence -> not equal
```

Clyybber (approved):

compiler/mir/datatables.nim
zerbina: Oh, nice catch.

saem (approved):

PR #1162, opened 2024-02-03
fix(vm): multiple memory leak issues

zerbina:
## Summary

Fix multiple severe memory leaks with the VM implementation, reducing
the memory usage and run time of complex and allocation-heavy
programs/macros/static-blocks.

## Details

In each case, the VM's allocator still knew about the cells and kept a
reference to the host memory, meaning that a `malloc`-level leak
detector wouldn't have caught them.

The problems:
1. ref-counted cells never reached a ref-count of zero, meaning that
   `ref`s were never freed in the VM
2. even if the ref-count would have reached zero, during standalone
   execution (via `vmrunner`), garbage ref-counted cells were never
   destroyed/freed
3. the `NodeToReg` implementation didn't clean up the destination
   register
4. the `NGetLineInfo` implementation didn't clean up the destination
   register

### Improper ref-counting

Ref-counted cells (allocated through `heapNew`) start with a ref-count
value of 1. Since the initial assignment to the destination location
used `asgnRef`, the ref-count was incremented again, preventing it from
ever reaching zero.

Instead of `asgnRef`, the destination location is first cleaned up with
`resetLocation` and then the `ref` value is assigned manually.

### Missing garbage cell clean-up

Ref-counted cells aren't destroyed and freed immediately but are first
added to the `pending` list. This list was never processed during VM
execution (only when disposing a `VmThread`), meaning that the garbage
cells accumulated indefinitely.

The `New` opcode now processes the `pending` list once its size exceeds
a fixed threshold (currently 128). In addition, `cleanUpPending` is
changed to process the list back-to-front (instead of front-to-back),
reducing the amount of `seq` resizing when new cells are added to the
list during cleanup.

### Missing register clean-up

Registers have to be cleaned up (via `cleanUpReg`) prior to changing
the kind, otherwise cells owned by owning handles (`rkLocation`
registers) aren't destroyed and freed.

Both `NodeToReg` and `NGetLineInfo` were missing this, resulting in,
depending on the register layout, cells never being freed.

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -990,7 +990,11 @@ proc rawExecute(c: var TCtx, t: var VmThread, pc: var int): YieldReason =
       if regs[rb].kind in {rkLocation, rkHandle}:
         checkHandle(regs[rb])
         if regs[rb].handle.typ.kind in RegisterAtomKinds:
-          loadFromLoc(regs[ra], regs[rb].handle)
+          let h = regs[rb].handle
+          # retrieve the handle *before* cleaning up the register, since `ra`
+          # and `rb` may point to the same register
+          cleanUpReg(regs[ra], c.memory)
+          loadFromLoc(regs[ra], h)
         else:
           unreachable() # vmgen issue
       else:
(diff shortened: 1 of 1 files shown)


zerbina:
The direction - reaching back to the initial rework of the VM - is to make `TFullReg` a "plain old data-type", with clean up of "stack" memory driven by the code generator through the use of dedicated opcodes. While `vmgen` is not quite there yet, it is getting closer.

saem (approved):

saem:
Outside the scope of this PR maybe a simple query routine for the VM to report it's memory usage statistics could be used to test. Something like:

```nim
  let snapshot = vmrunner.memoryStats
  # do stuff
  doAssert vmrunner.memoryStat == snapshot
```

PR #1164, opened 2024-02-03
mir: add unchecked signed-integer arithmetic ops

zerbina:
## Summary

Which signed integer arithmetic operation are checked for overflow is
now made explicit within the MIR. This:
* makes the exceptional control-flow arising from overflow checks
  visible to data-flow analysis, fixing destructors sometimes not being
  run
* moves decision-making regarding overflow checks out of the code
  generators and into a central place (the AST -> MIR translation)

## Details

In order to give the code generators more freedom over how they
implement the overflow checks, no separate magics that implement the
checks are introduced.

### MIR additions

* add built-in signed integer arithmetic unary operators (negation)
* add built-in signed integer arithmetic binary operators (addition,
  subtraction, multiplication, division, and remainder)
* the behaviour of the built-in operators regarding over- and
  underflow, as well as division-by-zero, is undefined

### General architecture

* when overflow checks are enabled, signed integer arithmetic
  operations are kept as magic calls (using `mnkCheckedCall` if panics
  are disabled)
* when overflow checks are disabled, signed integer arithmetic
  operations are translated to use the MIR built-in operators
* signed integer arithmetic operators kept as magic calls are
  unconditionally translated to checked operations by the code
  generators
* the built-in arithmetic operators are unconditionally translated to
  unchecked operations (if available) by the code generators
* the overflow check behaviour of `abs` is still controlled
  individually by each code generator

### AST-to-MIR translation

* `mSucc` (successor) and `mPred` (predecessor) are lowered into normal
  addition and subtraction operators, respectively. This was previously
  left to the code generators

### MIR-to-CGIR translation

* built-in operations equivalent to the ones added to the MIR are added
  to the `CgNode` IR
* the MIR's arithmetic operations have a one-to-one correspondence with
  the `CgNode` ones

### Code generation

* `vmgen` generates `Addu`, `Subu`, and `Mulu` for unchecked signed
  additions, subtractions, and multiplications. For a lack of unchecked
  versions, negation, division, and remainder still use the checked
  opcodes
* `cgen` and `jsgen` generate the same code for the unchecked built-in
  operators as they previously did for unchecked arithmetic magics
* all three code generators ignore `mSucc` and `mPred`

### Other

* `mirexec` treats the built-in arithmetic operations operands as
  usages
* the MIR built-in arithmetic operators are rendered as how they would
  look like in the high-level language (`+`, `*`, `div`, etc.)


zerbina:
The `abs` behaviour as written down in the `toverflow_abs.nim` test was discussed [here](https://matrix.to/#/!MAXxvfCYnfYgZqSizw:matrix.org/$r2dqBMiiyQxcuvn7YbvlHT2zm3fgQ17E5mbGdEb9FxA).

tests/lang_objects/destructor/tdestruction_when_checks_failed.nim
```diff
@@ -76,4 +76,12 @@ block object_conversion_checks:
...
\ No newline at end of file
+    test(RootRef()) # provoke a object-conversion-check failure
+
+block signed_integer_overflow_check:
+  proc test(a: int32) =
+    var obj = Object() # obj stores a value that needs to be destroyed
+    discard 1'i32 + a
+
+  runTest:
+    test(high(int32)) # provok an overflow-check failure
```
saem:
```suggestion
    test(high(int32)) # provoke an overflow-check failure
```

minor type, nbd.

saem (approved):
Found one very minor typo, nbd. Otherwise, the code is easy to read.

PR #1165, opened 2024-02-04
mir: separate NaN checks from FP arithmetic

zerbina:
## Summary

Decouple NaN checks from floating point arithmetic by emitting a
dedicated magic call during AST -> MIR translation.

This makes the exceptional control-flow arising from NaN checks visible
at the MIR level, which fixes locals not being destroyed if their scope
is left via an exception raised by a NaN check and there's no other
unstructured control-flow leaving the scope.

## Details

* if NaN checks are enabled, `mirgen` emits a call to the `mChckNaN`
  magic after each FP arithmetic operation (add, sub, mul, div)

* `mChckNaN` is a new internal-only unary magic

* `cgen` implements the magic in the same way it previously implemented
  NaN checks; NaN checks are no longer emitted manually for FP
  arithmetic

* `jsgen` and `vmgen` treat the `mChckNaN` as a no-op for now (both
  were already missing NaN check support previously)


compiler/backend/ccgexprs.nim
```diff
@@ -1763,6 +1760,15 @@ proc genMagicExpr(p: BProc, e: CgNode, d: var TLoc, op: TMagic) =
     linefmt(p, cpsStmts, "$1 = ($2)($3);$n", [a.r, typ, rdLoc(b)])
   of mChckRange:
     genRangeChck(p, e, d)
+  of mChckNaN:
+    var a: TLoc
+    initLocExpr(p, e[1], a)
+    # XXX: using a comparison could result in a floating point exception to
+    #      be raised (i.e., the flag to be set in the fenv).
+    #      ``__builtin_isnan``, which doesn't exhibit this behaviour, could be
```
saem: Any reason not to do this now?

saem (approved):
Quick question about not using the built-in now, but it's not really a blocker, feel free to merge.

compiler/backend/ccgexprs.nim
zerbina: It felt like something unrelated to the rest of the PR, but you're right, there's not really a reason against it -- I'll change it here.

zerbina: I've only changed the comment for now. The `__builtin_isnan` built-in is only available for GCC, Clang, and some other compilers aiming for GCC compatibility, but not for MSVC, and the only portable way of testing for NaNs that I've found would be inspecting the bit pattern (the C `isnan` function only became standard with C99).

Since `mChckNaN` is only intended for testing values resulting from arithmetic operations (which can never be signaling NaNs), I've changed the `mChckNaN` documentation to mention that signaling NaN behaviour is undefined.

PR #1167, opened 2024-02-06
fix(vmgen): wrong `range` value being read

zerbina:
## Summary

Fix an incorrect/invalid value being read from locations of unsigned
`range` type. Only the VM backend was affected, and only when the value
stored in the location had the most-significant bit set.

## Details

With the VM, all unsigned values require masking after being read from
a memory location.  `genRegLoad`  emits the necessary masking
instruction,
but `isUnsigned` doesn't consider range types (a separate bug), meaning
that values of type `tyRange` with an unsigned base type were not
masked.

* all code generation for narrowing/masking instructions is changed to
  skip `tyRange` types
* the custom narrowing implementation used for `mShrI` code generation
  is merged into `genNarrow`
* the duplicated unsigned integer masking is moved to a dedicated
  procedure (`genNarrowUnsigned`)

There was also the problem of the types' `size` field being accessed
directly; since a type's size is computed on demand, the value in
`size` might not be valid. `getSize` is used instead.

### DIFF
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -1362,7 +1362,7 @@ proc genNarrowUnsigned(c: var TCtx; info: TLineInfo, typ: PType,
   ## Only masks the value in `dest` (with ``opcNarrowU``) if `typ` is an
   ## unsigned integer type.
   let t = skipTypes(typ, IrrelevantTypes + {tyRange})
-  if isUnsigned(t) and (let size = getSize(c.config, typ); size < 8):
+  if isUnsigned(t) and (let size = getSize(c.config, t); size < 8):
     c.gABC(info, opcNarrowU, dest, TRegister(size * 8))
 
 proc genBinaryABCnarrow(c: var TCtx; n: CgNode; dest: var TDest; opc: TOpcode) =
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -1350,12 +1350,9 @@ proc genNarrow(c: var TCtx; n: CgNode; dest: TRegister; sNarrow = opcNarrowS) =
       else:             sNarrow
     c.gABC(n, op, dest, TRegister(size*8))
 
-proc genNarrowU(c: var TCtx; n: CgNode; dest: TDest) =
-  let t = skipTypes(n.typ, IrrelevantTypes + {tyRange})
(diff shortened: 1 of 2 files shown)


compiler/vm/vmgen.nim
```diff
@@ -1337,22 +1337,33 @@ proc genBinaryABC(c: var TCtx; n: CgNode; dest: var TDest; opc: TOpcode) =
...
 
-proc genNarrow(c: var TCtx; n: CgNode; dest: TDest) =
-  let t = skipTypes(n.typ, abstractVar-{tyTypeDesc})
-  # uint is uint64 in the VM, we we only need to mask the result for
-  # other unsigned types:
-  if t.kind in {tyUInt8..tyUInt32} or (t.kind == tyUInt and t.size < 8):
-    c.gABC(n, opcNarrowU, dest, TRegister(t.size*8))
-  elif t.kind in {tyInt8..tyInt32} or (t.kind == tyInt and t.size < 8):
-    c.gABC(n, opcNarrowS, dest, TRegister(t.size*8))
+proc genNarrow(c: var TCtx; n: CgNode; dest: TRegister; sNarrow = opcNarrowS) =
```
saem, line 1340:
```suggestion
proc genNarrow(c: var TCtx; info: TLineInfo; dest: TRegister; sNarrow = opcNarrowS) =
```

This could be a line info, instead of `n: CgNode`, in which case `genNarrowU` and `genNarrowUnsigned` could be simplified out or become inline `proc`s which specify the opcode and possibly carry some minimal logic.

saem (approved):
One soft stylistic suggestion that might reduce some code, otherwise looks good.

compiler/vm/vmgen.nim
zerbina: Only `genNarrowU` can be simplified away this way; `genNarrowUnsigned` deviates from the other narrow procedures in that it generates nothing for signed types.

I'll replace `genNarrowU`.

saem: I'm not sure how I feel about it, as it could be too clever, but `ocpRet` could potentially be used as a sentinel value to indicate "just return the value if signed"?

It's probably not a great idea.

zerbina: > but `ocpRet` could potentially be used as a sentinel value to indicate "just return the value if signed"?

Yep, that would work, although I think I'd rather leave `genNarrowUnsigned` be for the moment.

PR #1170, opened 2024-02-06
only allow re-raise in scope of `except`

zerbina:
## Summary

Treat a re-raise (i.e., `raise` without operand) as a compile-time
error when used outside of the *scope* of an `except` clause -- re-
raising from within a `finally` clause that's within the scope of an 
`except` clause is disallowed too.

This brings the re-raise feature into a "safe" state, where it can be
extended from again if wanted/needed.

## Details

If a re-raise is allowed from anywhere, it is possible to handle an in-
flight exception from within a `finally` clause:

```nim
try:
finally:
  try:
    raise # re-raise the in-flight exception
  except:
    # handle the exception
    discard
```

Since the exception is handled, control-flow would have to "fall out"
of the `finally` clause, but this violates the expectations of control-
flow analysis in the compiler as well as those of the code generators.
There, the expectation is that if a `finally` clause was entered
through exceptional control-flow, raising continues when the end of
the `finally` is reached.

Until it's clear if, how, and under what circumstances this should
work, a `finally` clause handling the in-flight exception is prevented
by restricting re-raise statements to the scope of `except` clauses.
`sempass2` implements the detection and error reporting.

The manual is updated accordingly and tests for the new behaviour are
added.

### Usage

In the compiler's repository, re-raising from outside the scope of an
`except` clause was only done by two tests. For one of them, the re-
raise can safely be replaced with a normal `raise`.

The other test explicitly tested the "re-raise in finally" case. It is
removed for now.

### DIFF
--- a/tests/exception/tdisallowed_reraise.nim
+++ b/tests/exception/tdisallowed_reraise.nim
@@ -1,0 +1,62 @@
+discard """
+  description: '''
+    Ensure that re-raise statements are rejected where required by the
+    language specification.
+  '''
+  cmd: "$nim check --hints:off $options $file"
+  action: reject
+  nimoutFull: true
+  nimout: '''
+tdisallowed_reraise.nim(26, 5) Error: an exception can only be re-raised within the scope of an except, with no finally in-between
+tdisallowed_reraise.nim(20, 3) Error: an exception can only be re-raised within the scope of an except, with no finally in-between
+tdisallowed_reraise.nim(39, 5) Error: an exception can only be re-raised within the scope of an except, with no finally in-between
+tdisallowed_reraise.nim(50, 7) Error: an exception can only be re-raised within the scope of an except, with no finally in-between
+'''
+"""
+
+
+block standalone_reraise:
+  # a standalone re-raise statement is disallowed
+  raise
+
+block reraise_in_procedure:
+  proc reraise() =
+    # the re-raise statement doesn't appear within an ``except`` clause's
+    # scope
+    raise
+
+  try:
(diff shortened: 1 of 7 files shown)


tests/exception/tdisallowed_reraise.nim
```diff
@@ -0,0 +1,62 @@
...
+  # no error is reported here
+  try:
+    raise CatchableError.newException("")
+  except:
+    try:
+      # a re-raise within a try's scope is allowed, if the try is within the
+      # scope of an except
+      raise
+    finally:
+      discard
```
saem, line 62:
```suggestion
block reraise_in_nested_try:
  # no error is reported here
  try:
    raise CatchableError.newException("")
  except:
    try:
      # a re-raise within a try's scope is allowed, if the try is within the
      # scope of an except
      raise
    finally:
      discard

block reraise_in_nested_finally_try:
  # no error is reported here
  try:
    discard
  finally:
    try:
      raise CatchableError.newException("")
    except:
      # a re-raise within an except within a finally, but not directly
      # within, is allowed.
      raise
```

I think this adds a bit of coverage, and directly for the second point as stated in the manual.

saem (approved):
Minor suggestion for additional test coverage, but perhaps it's sufficiently covered in `tfinally6.exceptionInFinally` -- feel free to ignore it.

tests/exception/tdisallowed_reraise.nim
zerbina: I'll add it.

PR #1171, opened 2024-02-07
fix unsigned `range` arithmetic causing overflow error

zerbina:
## Summary

Fix incrementing or decrementing (with `inc`, `dec`, `+=`, or `-=`) to
cause an overflow error for unsigned `range` types when the value would
wrap around.

The regression was introduced by commit
db5dcbffa2b580edb040b5d7d77ec96d6dfe082e.

## Details

`isUnsigned` didn't skip `tyRange` types, meaning that range types were
not treated as unsigned if their base type is an unsigned int. The
`isUnsigned` procedure is used:
1. in `getOrdValue` for forcing the node's integer value to be treated
   as unsigned (problematic usage)
2. in `mirgen.genMagic` for `mInc` and `mDec` translation (problematic
   usage)
3. in `semfold.getConstExpr` (unproblematic usage; the type is never a
   `tyRange`)
4. in `vmgen` for deciding the opcode for narrowing (unproblematic
   usage; range types are skipped prior)

For `getOrdValue`, the result of `isUnsigned` only makes a difference
values of unsigned type where the node kind is not a `nkUIntLitX` node.
The node kind seems to match the signed-ness of the type in most cases,
so `isUnsigned` not considering `tyRange` didn't lead to problems in
practice.

For `mirgen.genMagic`, `isUnsigned` is used to decide whether or not
checked signed integer arithmetic needs to be used for `mInc` or
`mDec`. Since `tyRange` was never treated as unsigned, this meant that
values of unsigned range type were erroneously subjected to overflow
checks.

`isUnsigned` now skips `tyRange` types, fixing the issue. A regression
test for the incorrect overflow checking is added.

### DIFF
--- a/compiler/ast/types.nim
+++ b/compiler/ast/types.nim
@@ -97,7 +97,7 @@ proc isPureObject*(typ: PType): bool =
   result = t.sym != nil and sfPure in t.sym.flags
 
 func isUnsigned*(t: PType): bool {.inline.} =
-  t.skipTypes(abstractInst + {tyEnum}).kind in {tyChar, tyUInt..tyUInt64}
+  t.skipTypes(abstractRange + {tyEnum}).kind in {tyChar, tyUInt..tyUInt64}
 
 proc getOrdValue*(n: PNode; onError = high(Int128)): Int128 =
   let k =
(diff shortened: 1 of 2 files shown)


saem (approved):
Nice fix.

Takeaway for me is that I need to think about ranges when I'm thinking about numbers/ordinal types.

PR #1175, opened 2024-02-08
tests: organize lang spec and generic related tests

saem:
## Summary
* what changed and how?
* why are we changing it?

## Details
* anything else

---

## TODO
- [ ] shift focus towards generics and keep going until tired

## Notes for Reviewers
* test re-org is messy stuff, no real narrative

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


saem:
I'm re-running the tests, I can't replicate the failure locally.

zerbina:
Huh, that's strange. `tjsonutils.nim` consistently fails for the Windows and MacOS CIs, but not on the Linux CI. The only reason for this I can think of is that moving around, removing, or adding new tests changed the make-up of the megatest in a way that caused a latent bug to surface.

saem:
> Huh, that's strange. `tjsonutils.nim` consistently fails for the Windows and MacOS CIs, but not on the Linux CI. The only reason for this I can think of is that moving around, removing, or adding new tests changed the make-up of the megatest in a way that caused a latent bug surfacing.

Yup, that's what I'm thinking too. Locally, I get js failures for `tjsonutils` on line 119. My best guess is it has something to do with the define `nimPreviewJsonutilsHoleyEnum`. It's specified in `tests/nim.cfg` and `jsonutils.toJson` conditionally compiles on it within generic code. I wonder if it's being instantiated one way and then differently at some other time.

I'm not quite sure why `fromJson` is failing in CI on line 286, the only thing they have in common is `isNamedTuple` but it's been around for a while.

Hmm, given the common denominator is `isNamedTuple` right now, then I'll start by investigating that I suppose.

saem:
Just a small update:

In terms of test ordering and population, the only difference is linux includes (`tests/misc/tsizeof4.nim`), while macos m1 does not -- it's disabled on arm64.

So I get local failures if I run `tjsonutils` for the js target, but no issues when doing a full run across all targets with megatest. I'm going to see if I can sort out the local failure issue and why it's not failing when run in a megatest context. Hopefully that'll help shed some light on things.

saem:
i haven't quite figured out how, but there seems to be a js codegen issue for `tjsonutils:119`. the passed in parameter seems to have field positions swapped so the values are assigned incorrectly.

Here is a snippet from the generated js:

```js
var Data5 = {jsonNodeMode: 1, enumMode: 0};
var Data6 = {jsonNodeMode: 2, enumMode: 0};
var Data7 = {enumMode: 2, jsonNodeMode: 0};
```

This corresponds to initializations of `ToJsonOptions`, the third item has the fields in the correct order. The first two the fields are in the incorrect order (not really a big deal), but the values are assigned incorrectly, `jsonNodeMode` should be `0` in both cases, while `enumMode` should have `1|2` assigned.

Just gotta track down where this is happening in jsgen.

Update: after some further debugging the issue might be earlier, in `mir` or `transf`.

saem:
With the above issue fixed in the following [PR](https://github.com/nim-works/nimskull/pull/1198#event-11841206905), thanks to @zerbina, going to see if that also resolves the issue encountered in CI.

PR #1176, opened 2024-02-09
std: `compilesettings` show CLI compile/link flags

saem:
## Summary

`compilersettings`  now includes compiler and linker options that are
set via the command-line  `--passc/passl`  flags.

## Details

Prior to this change only pragma based compiler and linker options were
shown, via  `{.passC: "...".}`  or  `{.passL:"...".}` . Now  `vmops` ,
which implements the compile time querying, uses compiler and linker
option construction procedures from the  `options`  module to fulfill
the query.

The compiler and linker options procedures in  `options`  are newly
introduced, taking over logic previously carried out in the  `extccomp` 
module, which orchestrates backend C compiler and linker operations. As
part of this change  `extccomp`  has been updated to also use these
procedures so there is one way and place where these options are
constructed. Additionally, the   `tcompilesetting`  test has been
updated to ensure these continue to work.

Note: the  `compilesettings`  query for compile options does _not_
include module specific compile options, set via 
`{.localPassc: "...".}`  pragma. This could be implemented, but that
would likely require further changes, such as introducing a new query to
get all file compiler options vs file specific.

### DIFF
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -394,6 +394,11 @@ passStrTableField macrosToExpand
 passStrTableField arcToExpand
 
 proc getCompileOptionsStr*(conf: ConfigRef): string =
+  ## Returns the combination of the current C compile options and the
+  ## global `--passC` options (combined in that exact order).
+  ##
+  ## Global `--passC` options already present in the current C compile
+  ## options are *not* include again.
   result = conf.compileOptions
 
   for option in conf.compileOptionsCmd:
@@ -402,6 +407,8 @@ proc getCompileOptionsStr*(conf: ConfigRef): string =
       result.add option
 
 proc getLinkOptionsStr*(conf: ConfigRef): string =
+  ## Returns the combination of the current C linker options and the global
+  ## `--passL` options (combined in that exact order).
   conf.linkOptions & " " & conf.linkOptionsCmd.join(" ")
 
 proc defineSymbol*(conf: ConfigRef, symbol: string, value: string = "true") =
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -393,6 +393,16 @@ passStrTableField symbols
 passStrTableField macrosToExpand
 passStrTableField arcToExpand
 
+proc getCompileOptionsStr*(conf: ConfigRef): string =
+  result = conf.compileOptions
+
+  for option in conf.compileOptionsCmd:
+    if strutils.find(result, option, 0) < 0:
+      if result.len == 0 or result[^1] != ' ': result.add " "
+      result.add option
+
+proc getLinkOptionsStr*(conf: ConfigRef): string =
+  conf.linkOptions & " " & conf.linkOptionsCmd.join(" ")
 
 proc defineSymbol*(conf: ConfigRef, symbol: string, value: string = "true") =
   conf.symbolsSet(symbol, value)
--- a/tests/vm/tcompilesetting.nim
+++ b/tests/vm/tcompilesetting.nim
@@ -13,7 +13,8 @@ template main =
   doAssert querySetting(backend) == "c"
   doAssert fileExists(libPath.querySetting / "system.nim")
   doAssert "-fmax-errors=4" in querySetting(compileOptions)
-  doAssert "-u dummysymboldoesnotexist" in querySetting(linkOptions), querySetting(linkOptions)
+  when not defined(macos) and not defined(windows):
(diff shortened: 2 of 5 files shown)


zerbina (commented):
Regarding the impact:
1. in addition to the pragma-specified options, `compilerOptions` previously also contained the `<ccompiler>.options.always` options (it still does)
2. the `compileOptions` query result now includes the `<ccompiler>.options.speed|debug|size` options
3. the `linkOptions` query result now include the linked C libraries

Number 2 is fine, I think, but number 3 I'm not so sure about, given that the libraries to link against aren't really options, in that sense.

As an alternative, `vmops` could append the options itself, like so:
```nim
# compile options
result = conf.compileOptions
for option in conf.compileOptionsCmd:
  if find(result, option, 0) < 0:
    result.add " " & option

# link options
result = conf.linkOptions & " " & conf.linkOptionsCmd.join(" ") 
```
This would also remove the `extccomp` dependency, at the cost of a small duplication of logic. It's only a soft suggestion, however, so feel free to ignore.

saem:
Hmm, maybe the query should be in `options` and only cover combining pragma and cli options, as the split storage is a bit more of a `ConfigRef` detail.

---

Separately, wrt the failing tests I guess clang is pickier about undefined symbols, I need to find another "no-op" like option.

compiler/front/options.nim
```diff
@@ -393,6 +393,16 @@ passStrTableField symbols
 passStrTableField macrosToExpand
 passStrTableField arcToExpand
 
+proc getCompileOptionsStr*(conf: ConfigRef): string =
+  result = conf.compileOptions
```
zerbina, line 402:
```suggestion
  ## Returns the combination of the current C compile options and the
  ## global `--passC` options (combined in that exact order).
  ##
  ## Global `--passC` options already present in the current C compile
  ## options are *not* include again.
  result = conf.compileOptions
```

compiler/front/options.nim
```diff
@@ -393,6 +393,16 @@ passStrTableField symbols
...
+
+  for option in conf.compileOptionsCmd:
+    if strutils.find(result, option, 0) < 0:
+      if result.len == 0 or result[^1] != ' ': result.add " "
+      result.add option
+
+proc getLinkOptionsStr*(conf: ConfigRef): string =
+  conf.linkOptions & " " & conf.linkOptionsCmd.join(" ")
```
zerbina, line 412:
```suggestion
  ## Returns the combination of the current C linker options and the global
  ## `--passL` options (combined in that exact order).
  conf.linkOptions & " " & conf.linkOptionsCmd.join(" ")
```

tests/vm/tcompilesetting.nim
```diff
@@ -1,9 +1,9 @@
 discard """
-cmd: "nim c --nimcache:build/myNimCache --nimblePath:myNimblePath $file"
+matrix: "--nimcache:build/myNimCache --nimblePath:myNimblePath --passc:'-fmax-errors=4' --passl:'-u dummysymboldoesnotexist'"
```
zerbina:
```suggestion
matrix: "--nimcache:build/myNimCache --nimblePath:myNimblePath --passc:'-fmax-errors=4' --passl:'-Xlinker --defsym -Xlinker dummysymboldoesnotexist=0x0'"
```
I cannot test it with the XCode toolchain, but I think that this should work.

zerbina (approved):
Nice, I think having the combination logic in `options` is indeed better. I've left two suggestion regarding documentation of the new procedures.

Looking for a safe replacement of `-u`, I found `--defsym` (create a global symbol storing the provided address), which might work with the linker from the Xcode toolchain (it does with the MinGW `ld`, at least).

tests/vm/tcompilesetting.nim
saem: LooooooL, no idea why I thought guarding the compiler query would change the outcome of the test considering the cli params. I'm a bit distracted today. 😮‍💨 

I'll get things building again on my mac (so slow) and see if this works there.

compiler/front/options.nim
saem: Thank you for suggesting these docs, I totally blanked on the writing docs for public definitions.

saem:
I'm restricting linker flag testing to linux only for now, this is turning out to be a bear. 😬

saem:
/merge

huh, not sure what's with the kill 9, in that last run trying again.

PR #1177, opened 2024-02-10
cgen: fix unnecessary default initialization of locals

zerbina:
## Summary

A local of certain types was unnecessarily default-initialized even if
it was assigned immediately (that is, `var x = y`). This doesn't change
observable behaviour, but it does speed up the generated code a bit.

## Details

* only emit default-initialization for definitions when there's no
  initial-value expression
* remove the `isAssignedImmediately` procedure. It's only used a
  single time, and its general name might cause (mis-)use for purposes
  it's not intended for  
* remove the outdated comment in `cgen.initLocalVar`

### DIFF
--- a/compiler/backend/ccgstmts.nim
+++ b/compiler/backend/ccgstmts.nim
@@ -15,13 +15,6 @@ const
   stringCaseThreshold = 8
     # above X strings a hash-switch for strings is generated
 
-proc isAssignedImmediately(conf: ConfigRef; n: CgNode): bool {.inline.} =
-  # note: the destination can be considered assigned immediately even if the
-  # source is an exception-raising RVO call. This is because the call happens
-  # *before* the definition, meaning that nothing on the exceptional control-
-  # flow path must observe (read or write) the destination
-  n.kind != cnkEmpty
-
 proc inExceptBlockLen(p: BProc): int =
   for x in p.nestedTryStmts:
     if x.inExcept: result.inc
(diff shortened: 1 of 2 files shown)


compiler/backend/ccgstmts.nim
```diff
@@ -16,13 +16,11 @@ const
...
-  if isInvalidReturnType(conf, n.typ):
-    # var v = f()
-    # is transformed into: var v;  f(addr v)
-    # where 'f' **does not** initialize the result!
-    return false
-  result = true
+  # note: the destination can be considered assigned immediately even if the
+  # source is an exception-raising RVO call. This is because the call happens
+  # *before* the definition, meaning that nothing on the exceptional control-
+  # flow path must observe (read or write) the destination
```
saem: Here "exceptional control flow path" is exclusive of both prior to an exception or in a finally? Basically, one of those latter two needs to write to the destination in order for it to be initialized?

saem (approved):
A minor question, otherwise it looks good.

saem:
/merge

Thought about it a bit more and I don't think the answer to the questions would change anything about this PR, can be discussed in a non-blocking fashion.

zerbina:
@saem: Sorry, I was away for a moment. I've also thought about it some more and I do think the comment is a bit unclear, which is partly due to where it's located (it's in a procedure named `isAssignedImmediately` and talks about "definition").

I'll move it somewhere else and will also try to make it a bit clearer.

saem:
> @saem: Sorry, I was away for a moment. I've also thought about it some more and I do think the comment is a bit unclear, which is partly due to where it's located (it's in a procedure named `isAssignedImmediately` and talks about "definition").
> 
> I'll move it somewhere else and will also try to make it a bit clearer.

Thanks, I re-read it and figured the code wouldn't really change so started the merge. Totally cool if you want to cancel it and restart it after the change.

PR #1179, opened 2024-02-11
vm: use contiguous register storage

zerbina:
## Summary

This is an internal-only change. The VM implementation now uses a
register sequence per *thread* rather than per *stack frame*, reducing
the overhead of entering and exiting a procedure for code running in
the VM.

### VM interface

* allocating the initial stack frame and registers is part of 
`initVmThread`  now
* `compilerbridge` is adjusted accordingly; `execute` now takes a
  `VmThread`, instead of a `TStackFrame`, as input so that the
  callsites can still initialize the parameter registers

### VM architecture

* `VmThread` stores the `seq[TFullReg]` with all registers
* instead of their own `seq[TFullReg]`, each stack frame
  (`TStackFrame`) remembers the index where its register slice begins
* the register sequence is a stack: invoking a procedure grows it,
  while leaving a procedure shrinks it again
* the `.cursor`-based mechanism for keeping a low-overhead reference to
  the current frame's register slice is replaced with `Registers`,
  which is a length + `ptr UncheckedArray` pair (an unsafe first-class
  `openArray`, effectively)
* instead of by an `IndexDefect` being raised, out-of-bounds register
  access is reported via the `VmEvent` facility. This means that out-
  of-bounds register access resulting from ill-formed bytecode no
  longer crashes the VM

Using a single sequence per thread means that there's no more `seq`
allocation/destruction overhead per procedure call, which also means
less pressure on the host's allocator.

### Misc

* a leftover `debugEcho` from some earlier change in `packed_env` is
  removed

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -2912,12 +2912,16 @@ proc rawExecute(c: var TCtx, t: var VmThread, pc: var int): YieldReason =
 
 proc `=copy`*(x: var VmThread, y: VmThread) {.error.}
 
-proc initVmThread*(c: var TCtx, pc: int, frame: sink TStackFrame): VmThread =
-  ## Sets up a ``VmThread`` instance that will start execution at `pc`.
-  ## `frame` provides the initial stack frame.
+proc initVmThread*(c: var TCtx, pc: PrgCtr, numRegisters: int,
+                   sym: PSym): VmThread =
+  ## Sets up a `VmThread <#VmThread>`_ instance that will start execution at
+  ## `pc` and that has `numRegisters` as the initial amount of registers.
+  ## `sym` is the symbol to associate the initial stack-frame with. It may be
+  ## nil.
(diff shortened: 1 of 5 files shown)


compiler/vm/vm.nim
```diff
@@ -192,6 +200,33 @@ proc createStackTrace*(
 
   assert result.stacktrace.len() <= recursionLimit # post condition check
 
+func initFrom(regs: seq[TFullReg], start: int): Registers =
+  ## Creates a register list view covering `at..regs.high`.
```
```suggestion
  ## Creates a register list view covering `start..regs.high`.
```

I believe this is what was meant.

saem (approved):
Small suggestion for what I believe was a parameter rename in docs.

This is a really nice improvement.

compiler/vm/vm.nim
zerbina: Yep, it is, thanks.

PR #1180, opened 2024-02-11
sem: correctly show errors from `defer` node body

saem:
## Summary

Errors in a defer node body are now correctly reported, no longer
producing misleading  `'defer' takes a 'void' expression`  messages.

## Details

A `defer`  with an error node body is no longer treated as an expression
and having the underlying error supressed. Instead the body error is
correctly reported. The  `defer`  AST is also correctly wrapped in an
error node.

The implementation of `defer` semantic analysis has changed a fair bit:

1. it now checks for well formedness (number of children, and reports
illegal AST when not conforming)
2. the body is now checked via  `semStmt` , which appropriately handles
discard checks
3. as a consequence of the above, the specialized  `void expression` 
message is no longer required and assocated reports have been removed

To avoid regressions in any of these areas a relatively comprehensive
test sutie for  `defer`  statements was added testing all of these
facets.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -3878,22 +3878,14 @@ proc semExpr(c: PContext, n: PNode, flags: TExprFlags = {}): PNode =
   of nkDefer:
     if c.currentScope == c.topLevelScope:
       localReport(c.config, n, reportSem rsemUnexpectedToplevelDefer)
-    let res = semExpr(c, n[0])
-    case res.kind
+    checkSonsLen(n, 1, c.config)
+    result = copyNodeWithKids(n)
+    result[0] = semStmt(c, n[0], {})
+    case result[0].kind
     of nkError:
-      result = copyNodeWithKids(n)
-      result[0] = res
       result = c.config.wrapError(result)
     else:
-      if res == n[0]: # no change
-        discard
-      else:
-        result = copyNodeWithKids(n)
-        result[0] = res
-
-      # TODO: convert to nkError
-      if not res.typ.isEmptyType and not implicitlyDiscardable(res):
-        localReport(c.config, n, reportSem rsemExpectedTypelessDeferBody)
+      discard
   of nkMixinStmt: discard
   of nkBindStmt:
     if c.p != nil:
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -3878,9 +3878,22 @@ proc semExpr(c: PContext, n: PNode, flags: TExprFlags = {}): PNode =
   of nkDefer:
     if c.currentScope == c.topLevelScope:
       localReport(c.config, n, reportSem rsemUnexpectedToplevelDefer)
-    n[0] = semExpr(c, n[0])
-    if not n[0].typ.isEmptyType and not implicitlyDiscardable(n[0]):
-      localReport(c.config, n, reportSem rsemExpectedTypelessDeferBody)
+    let res = semExpr(c, n[0])
+    case res.kind
+    of nkError:
+      result = copyNodeWithKids(n)
+      result[0] = res
+      result = c.config.wrapError(result)
+    else:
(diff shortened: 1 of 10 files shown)


compiler/sem/semexprs.nim
```diff
@@ -3878,9 +3878,22 @@ proc semExpr(c: PContext, n: PNode, flags: TExprFlags = {}): PNode =
...
+    else:
+      if res == n[0]: # no change
+        discard
+      else:
+        result = copyNodeWithKids(n)
+        result[0] = res
+
+      # TODO: convert to nkError
+      if not res.typ.isEmptyType and not implicitlyDiscardable(res):
+        localReport(c.config, n, reportSem rsemExpectedTypelessDeferBody)
```
```suggestion
    checkSonsLen(n, 1, c.config)
    result = copyNodeWithKids(n)
    result[0] = semStmt(c, n[0], {})
    case result[0].kind
    of nkError:
      result = c.config.wrapError(result)
    else:
      discard
```
Two additional notes:
* a new node is currently always required, otherwise the `nfSem` inclusion at the end of `semExpr` would mutate input AST
* the `checkSonsLen` call fixes:
  * the compiler crashing with an out-of-bounds access if a macro provides an empty `nkDefer` node
  * `nkDefer` AST with more than one child node not being rejected

tests/lang_stmts/defer/tdefer_cannot_be_top_level.nim
```diff
@@ -0,0 +1,8 @@
+discard """
+  description: "Defer cannot be decalred at the top level"
```
zerbina:
```suggestion
  description: "Defer cannot be declared at the top level"
```

zerbina (commented):
I think it would be better if `nkDefer` requests a statement instead via `semStmt`. This would have the following benefits:
* the same error message for the `defer` body as is used with all other statement contexts (i.e., "expression needs to be used or discarded")
* `nkDefer` analysis is not special anymore
* proper "implicitly discardable" handling: an call with an implicitly discardable result needs to be wrapped in an `nkDiscardStmt`, but this is currently missing (`semStmt` would take care of this via `discardCheck`)

I left a suggestion with the change implementing the above (minus the removal of `rsemExpectedTypelessDeferBody`). It could also be done as a follow-up, I think.

saem:
Thanks for the review, I'll update the PR accordingly, couldn't do it today as I was under the weather.

zerbina:
> Thanks for the review, I'll update the PR accordingly, couldn't do it today as I was under the weather.

No worries :).

Also, I worded my review a bit too strong -- it's really only a soft suggestion. Besides the missing copy in case the node didn't change, I think the change is totally fine, and using `semStmt` can always be done separately.

saem:
> Also, I worded my review a bit too strong -- it's really only a soft suggestion. Besides the missing copy in case the node didn't change, I think the change is totally fine, and using `semStmt` can always be done separately.

It's cool, I like the changes, consistency FTW.

tests/lang_stmts/defer/tdefer_malformed_many_children.nim
```diff
@@ -0,0 +1,16 @@
+discard """
+  description: "Empty `defer` body is not allowed (macro input)."
```
zerbina:
```suggestion
  description: "`defer` must have exactly one child node (macro input)."
```

zerbina (approved):
Cool, that's one less report and a lot more test coverage for `defer`. The description for the "many children" test seems to be the same as for the "no child node" test, and you might want to update the PR description, but looks good otherwise.

tests/lang_stmts/defer/tdefer_malformed_many_children.nim
saem: yup, that was a copy pasta fail. 😆

saem:
Thanks for all the reviews, I'm going to do a once over of the PR body and then send it off to merge.

PR #1182, opened 2024-02-12
fix compiler crash with macros and enum `$`

zerbina:
## Summary

Fix the compiler crashing when returning AST containing a call to the
compiler-generated `$` for an enum type from a macro. The crash only
occurred if the problematic callee is a resolved symbol (not a raw
identifier).

## Details

`sigmatch.initCallCandidate` searches for the module a routine is
owned by, by traversing the owners while skipping generic routines. If
a routine is marked with the `sfFromGeneric` flag, `skipGenericOwner`
unconditionally skips one `owner`. This is because the instantiation of
a generic routine has the generic routine it was instantiated from as
its immediate owner.

The compiler-generated enum-to-string procedure is also marked with
`sfFromGeneric`, but it has the owner of the enum type as the `owner`,
and not a generic routine. If the owner of the enum type is a module
(`skModule`), this meant that `initCallCandidate` skipped over the
`skModule` symbol, eventually reaching the end of the chain and
crashing with an NPE.

Since in the case of `genEnumToStrProc`, the generic procedure
(`EnumToStr` from the `dollars` module) might not be available yet, no
generic symbol can be used as the owner. To fix the problem, the
synthesized procedure is now no longer marked with the `sfFromGeneric`
flag.

### DIFF
--- a/compiler/ast/enumtostr.nim
+++ b/compiler/ast/enumtostr.nim
@@ -11,7 +11,10 @@ import
 
 
 proc genEnumToStrProc*(t: PType; info: TLineInfo; g: ModuleGraph; idgen: IdGenerator): PSym =
-  result = newSym(skProc, getIdent(g.cache, "$"), nextSymId idgen, t.owner, info)
+  # the compiler-generated enum-to-string procedure is treated as an
+  # instantiation of the generic ``mEnumToStr`` magic
+  let owner = getSysMagic(g, info, "$", mEnumToStr)
+  result = newSym(skProc, getIdent(g.cache, "$"), nextSymId idgen, owner, info)
 
   let dest = newSym(skParam, getIdent(g.cache, "e"), nextSymId idgen, result, info)
   dest.typ = t
@@ -51,15 +54,7 @@ proc genEnumToStrProc*(t: PType; info: TLineInfo; g: ModuleGraph; idgen: IdGener
   # treat the symbol as a magic so that further processing can easily detect
   # auto-generated enum-to-string procedures
   result.magic = mEnumToStr
-  result.flags.incl {sfNeverRaises}
-  # XXX: the synthesized procedure could be seen as coming from a generic
-  #      routine (the generic ``mEnumToStr`` magic procedure from the system
-  #      module), but for the ``sfFromGeneric`` flag to be included, the owner
-  #      of the symbol would have to be the originating-from generic procedure,
-  #      which isn't possible at the moment, given that sem might not have
-  #      processed the declaration (meaning no symbol) of said magic procedure
-  #      when ``genEnumToString`` is called
-  #result.flags.incl {sfFromGeneric}
+  result.flags.incl {sfFromGeneric, sfNeverRaises}
 
 proc searchObjCaseImpl(obj: PNode; field: PSym): PNode =
   case obj.kind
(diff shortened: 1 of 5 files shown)


zerbina:
The only possible problem with not marking the procedure with `sfFromGeneric` I could find is that `object` or `enum` types owned by the enum-to-string procedure don't get hashed properly (refer to [`hashTypeSym`](https://github.com/nim-works/nimskull/blob/bd28145a33ab7540d08258b0fd306c65947c3f02/compiler/sem/sighashes.nim#L69-L71)). This is not a problem in practice, however, as the generated enum-to-string procedure neither defines nor instantiates any `object` or `enum` types.

saem (commented):
This is one way to do it, another would be to enforce that the generic magic, `EnumToStr` is already processed or to treat that as an error (bug in `std`). That's sorta better, but also a pain if someone wants to use a custom standard library.

I'm ok with merging this as is, but that's mostly because I'm not sure how to differentiate which approach is better.

zerbina:
> another would be to enforce that the generic magic, `EnumToStr` is already processed or to treat that as an error (bug in `std`).

That's what I went with first (because I think it's the better solution), and while it's simple to implement in the compiler, it doesn't work without changing `system`. The `bool` type is defined as an `enum`, which means that the the `EnumToStr` magic has to be defined prior.

I've thought about it some more, and I believe that moving the `EnumToStr` definition from `dollars.nim` to just above the `basic_types.nim` include in `system.nim` would work. That's not super pretty, but it would get the job done until the `system` module is restructured and magic type forward definitions are allowed (the latter which would remove the need to have `EnumToStr` defined prior to the initial `bool` definition.

I'll give moving `EnumToStr` a try and see what the CI says.

zerbina:
Well, looks like it works. The out-of-place procedure definition, and the `mixin`s it requires, is a bit unfortunate, but I still think that this is the better solution.

compiler/ast/enumtostr.nim
```diff
@@ -11,19 +11,21 @@ import
 
 
 proc genEnumToStrProc*(t: PType; info: TLineInfo; g: ModuleGraph; idgen: IdGenerator): PSym =
-  result = newSym(skProc, getIdent(g.cache, "$"), nextSymId idgen, t.owner, info)
+  # the compiler-generated enum-to-string procedure is treated as an
+  # instantiation of the generic ``mEnumToStr`` magic
+  let owner = getSysMagic(g, info, "$", mEnumToStr)
+  result = newSym(skProc, getIdent(g.cache, "$"), nextSymId idgen, owner, info)
```
saem, line 17: Yup, this feels a lot better in terms of the name of the proc and the actual implementation all matching up, with how everything else works.

lib/system.nim
```diff
@@ -52,6 +52,15 @@ type
   typed* {.magic: Stmt.}         ## Meta type to denote an expression that
                                  ## is resolved (for templates).
 
+# XXX: the ``EnumToStr`` magic has to be defined here so that the ``bool``
+#      definition can compile
```
saem, line 56: This feels better, because the cost of the complexity of the `system` module's current arrangement is now paid for by the `system` module.

lib/system.nim
```diff
@@ -1970,6 +1979,7 @@ template newException*(exceptn: typedesc, message: string;
                        parentException: ref Exception = nil): untyped =
   ## Creates an exception object of type `exceptn`, initializes it's `name`
   ## and sets its `msg` field to `message`. Returns the new exception object.
+  mixin `$`
```
saem, line 1982: Better a `mixin` than an oddity in the compiler, side-benefit is whenever we figure out what, if anything, should obviate `mixin`s, this can addressed with then.

saem (approved):

saem:
/merge

The new approach is much nicer, IMO. Thanks for revising it.

PR #1184, opened 2024-02-13
mir: introduce the proto-MIR

zerbina:
## Summary

The proto-MIR is the precursor to the MIR, acting as an intermediate
target for expressions during the AST-to-MIR translation. It is used
for the analysis previously performed on MIR expressions, allowing
`mirgen` to focus more on syntax-related translation.

Semantics don't change, and there there's no deliberate change to how
constructs are translated, although there are some minor differences
(improvements) in the produced MIR.

### Background

`mirgen` previously performed translation, semantics-related decision
making (e.g., where to inject run-time checks), and attribute
inheritance (e.g., ownership of constructed values) all in one go. For
deciding where and how to inject temporaries, the just-produced MIR was
analyzed (`isPure`, `isStable`, and `detectKind`).

While efficient in terms of memory and time usage, this approach leads
to a fairly complex `mirgen`, with the performed analysis also limited
by the MIR itself.

### The Proto-MIR

The proto-MIR (short: PMIR) is an IR separate from the MIR that only
covers a subset of existing expressions from the AST (calls,
constructions, and complex expressions, are not covered). It represents
expressions in a way such that translating them to the MIR requires as
little conditional logic as possible:
* it includes all lvalue-expression-related run-time checks
* tuple, array, and field access are separated
* `when nimvm` expressions are resolved
* purity of index and dereference operands is already ensured
* some necessary lowering is already applied

#### Data representation

* expressions are represented as *chains* of *items* (not trees of
  nodes)
* the sub-expressions forming an expression (only used for index
  operands at the moment) are part of a single `seq`
* each item stores the originating-from AST node (which still carries
  some information required by the later translation), the type, plus
  item-specific data
* all fields of `ref` type are marked as `.cursor` in order to get
  around the unnecessary ref operations (the objects are guaranteed to
  outlive the PMIR items)

#### Proto-MIR implementation

* translating expression AST to a proto-MIR expression is done via
  `exprToPmir`
* the `exprToMir` implementation is more or less a copy of
  `mirgen.genx`, but with PMIR items being emitted rather than MIR
  nodes
* the `isPure`, `isStable`, and `detectKind` (renamed to `classify`)
  analysis procedures from `mirgen` are moved to the `proto_mir` module
  and adjusted to work with the PMIR
* in contrast to the previous `isPure` implementation, the new one is
  exhaustive and also treats 'let' globals and various rvalue-yielding
  operations as pure

### Integration into `mirgen`

* whenever an expression AST is to be translated, it is first
  translated to PMIR
* the `wantX` routines are called to ensure that the expressions meet
  the contextual requirements (stable, pure, etc.)
* for expressions, `mirgen` only implements the PMIR-to-MIR step (with
  `genx`)

### Minor differences in MIR output

When field checks are disabled, a nested variant access expression such
as `a.f` is now translated to:
```
discard a.case1.case2.case3.f
```
instead of:
```
bind _1 = a.case1
bind _2 = _1.case2
discard _2.case3.f
```

This is because the operand of unchecked variant access is now never
(explicitly) captured, whereas `mirgen.genVariantAccess` previously
always captured the operand for the sake of a simpler implementation.

### DIFF
--- a/compiler/mir/proto_mir.nim
+++ b/compiler/mir/proto_mir.nim
@@ -124,11 +124,14 @@ type
       sym* {.cursor.}: PSym
     of pirTupleAccess:
       pos*: uint32
-    of pirFieldAccess, pirVariantAccess, pirCheckedVariantAccess:
+    of pirFieldAccess, pirVariantAccess:
       field* {.cursor.}: PSym
     of pirArrayAccess, pirSeqAccess, pirCheckedArrayAccess, pirCheckedSeqAccess:
       index*: int
         ## the item index of the index operand expression
+    of pirCheckedVariantAccess:
+      nodeIndex*: int
+        ## the check's index in the originating-from ``nkCheckedFieldAccess``
     of pirCheckedObjConv:
       check* {.cursor.}: PType
         ## the type to emit a run-time check with
@@ -549,15 +552,13 @@ proc exprToPmir(c: TranslateCtx, result: var seq[ProtoItem], n: PNode, sink: boo
         else:                          check[2].sym
 
       if optFieldCheck in c.options:
-        # note: for a checked access, the symbol of the actually accessed
-        # field, not that of the discriminator, is used
-        result.add ProtoItem(orig: check, typ: n[0][0].typ, keep: kLvalue,
-                             kind: pirCheckedVariantAccess, field: field)
+        result.add ProtoItem(orig: n, typ: n[0][0].typ, keep: kLvalue,
+                             kind: pirCheckedVariantAccess, nodeIndex: i)
       else:
-        result.add ProtoItem(orig: check, typ: n[0][0].typ, keep: kLvalue,
+        result.add ProtoItem(orig: n, typ: n[0][0].typ, keep: kLvalue,
                              kind: pirVariantAccess, field: discr)
 
-    result.addField n[0], pirFieldAccess, field
+    result.addField n, pirFieldAccess, field
 
   of nkObjDownConv, nkObjUpConv:
     # the object conversion might need a run-time check
--- a/compiler/mir/proto_mir.nim
+++ b/compiler/mir/proto_mir.nim
@@ -686,7 +686,12 @@ proc exprToPmir(c: TranslateCtx, result: var seq[ProtoItem], n: PNode, sink: boo
     recurse(selectWhenBranch(n, c.pickVm), sink)
   of nkStmtListExpr:
     recurse(n.lastSon, sink)
-    node pirStmtList
+    # HACK: inherit the type from the child node. This prevents incorrectly
+    #       typed array constructions (a sem bug) from appearing to work
+    #       correctly (see
+    #       tests/lang_exprs/tempty_typed_expressions_issues.nim)
+    result.add ProtoItem(orig: n, typ: result[^1].typ, kind: pirStmtList)
+    #node pirStmtList
   of nkPragmaBlock:
     # the pragma is uninteresting here, just skip it
     recurse(n.lastSon, sink)
--- a/compiler/mir/proto_mir.nim
+++ b/compiler/mir/proto_mir.nim
@@ -227,7 +227,7 @@ func isPure(e: seq[ProtoItem], n: int): bool =
     e[n].typ.kind != tySink
   of pirConst, pirLiteral, pirProc, pirConstExpr:
     true
-  of pirLocal:
+  of pirLocal, pirGlobal:
     # let bindings are pure, but only if they don't have a destructor (in
     # which case they're movable)
     e[n].sym.kind in {skLet, skForVar} and not hasDestructor(e[n].typ)
--- a/compiler/mir/proto_mir.nim
+++ b/compiler/mir/proto_mir.nim
@@ -89,11 +89,11 @@ type
     pirMatLvalue ## materialize the lvalue into a temporary reference
 
   Keep* = enum
-    kDontCare  ## can capture by value or lvalue; it doesn't matter
-    kLvalue    ## a capture must preserve lvalue-ness
-    kMutLvalue ## a capture must preserve lvalue-ness and the created
-               ## reference must be usable as an immediate assignment
-               ## destination
+    kDontCare  ## lvalue-ness doesn't matter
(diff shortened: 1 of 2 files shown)


zerbina:
The name "Proto-MIR" was originally a working title. I think it fits well, but I'm also open to suggestion for better names.

compiler/mir/proto_mir.nim
```diff
@@ -0,0 +1,718 @@
...
+    # --- materialization
+
+    pirMat       ## materialize the value into an owning temporary
+    pirMatCursor ## materialize the value into non-owning temporary
+    pirMatLvalue ## materialize the lvalue into a temporary reference
+
+  Keep* = enum
+    kDontCare  ## can capture by value or lvalue; it doesn't matter
+    kLvalue    ## a capture must preserve lvalue-ness
+    kMutLvalue ## a capture must preserve lvalue-ness and the created
```
saem:
```suggestion
    kCaptureAny  ## can capture by value or lvalue; it doesn't matter
    kCaptureLval    ## a capture must preserve lvalue-ness
    kCaptureMutLval ## a capture must preserve lvalue-ness and the created
```

Just going based off the docs, haven't really thought about how it would read in context. This is a very soft suggestion, feel free to ignore it.

compiler/mir/proto_mir.nim
```diff
@@ -0,0 +1,718 @@
...
+  of pirStmtList:
+    classify(e, i - 1)
+
+func classify*(e: seq[ProtoItem]): ExprKind =
+  ## Returns the kind of the given proto-MIR expression.
+  classify(e, e.high)
+
+func isPure(e: seq[ProtoItem], n: int): bool =
+  ## Returns whether the expression at `n` is a pure expression. An expression
+  ## is pure evaluating it doesn't depend on mutable state. In other words, a
```
saem:
```suggestion
  ## is pure if evaluating it doesn't depend on mutable state. In other words, a
```

compiler/mir/proto_mir.nim
```diff
@@ -0,0 +1,718 @@
...
+    # note: the AST-to-PMIR tranlsation made sure that the index operand is
+    # pure
+    isPure(e, n - 1)
+  of pirMatLvalue:
+    # depends on the lvalue
+    isPure(e, n - 1)
+  of pirMat, pirMatCursor:
+    # the materialized-into temporary is never assigned to
+    true
+  else:
```
saem: Very soft suggestion, but perhaps a `pirNeverPureKinds` to replace the `else` and make this exhaustive is worth it?

compiler/mir/proto_mir.nim
```diff
@@ -0,0 +1,718 @@
...
+    false
+
+func isStable(e: seq[ProtoItem], n: int): bool =
+  ## Returns whether the run-time address of the lvalue expression `n` is
+  ## always the same, regardless of when the address is computed.
+  case e[n].kind
+  of pirGlobal, pirParam, pirLocal, pirConst, pirConstExpr, pirMat,
+     pirMatCursor, pirMatLvalue:
+     # location names (which includes constant expressions) and materialized
+     # expressions are never change identity
```
saem:
```suggestion
     # expressions never change identity
```

compiler/mir/proto_mir.nim
```diff
@@ -0,0 +1,718 @@
...
+  ##
+  ## Only fully constant, non-empty aggregate or set constructions are
+  ## treated as constant expressions.
+  if not sink and n.len > ord(n.kind == nkObjConstr) and isDeepConstExpr(n):
+    result.add ProtoItem(orig: n, typ: n.typ, kind: pirConstExpr)
+  elif kind != pirSetConstr:
+    result.add ProtoItem(orig: n, typ: n.typ, kind: kind)
+    result[^1].owning = sink
+  else:
+    result.add ProtoItem(orig: n, typ: n.typ, kind: kind)
```
saem:
```suggestion
  elif kind == pirSetConstr:
    result.add ProtoItem(orig: n, typ: n.typ, kind: kind)
  else:
    result.add ProtoItem(orig: n, typ: n.typ, kind: kind)
    result[^1].owning = sink
```

Very soft suggestion, but I find avoiding negations tends to read slightly better.

saem (approved):
The comments are minor, a few typo/grammar suggestions and soft stylistic suggestions. I couldn't think of a better name than Proto-MIR.

Things are a bit cleaner with Proto-MIR, as the amount one needs to consider/understand at any given time is smaller. As alluded to in the PR message, the bigger wins are in follow-on work, so I'm hoping the analysis/codegen benefits become more pronounced then.

mrgaturus:
what about naming module as `mirproto.nim` to keep consistency between other mir modules naming?

compiler/mir/proto_mir.nim
zerbina: Yeah, the docs conflict with the names. My thinking was to use declarative names (i.e., what properties are expected / need to be kept) in order to make the enum more broadly applicable (in the context of the proto-MIR).

I'll update the documentation to not mention capturing.

compiler/mir/proto_mir.nim
zerbina: A dedicated constant is probably a bit too much (I currently don't foresee it being used anywhere else), but the `case` should nevertheless still be exhaustive.

zerbina:
> what about naming module as `mirproto.nim` to keep consistency between other mir modules naming?

My thinking was/is that the proto-MIR is not directly part of or relates to the MIR in the way the other modules do, so I deliberately not used the same nomenclature as for the other modules, in order to signify the `proto_mir` module's outlier status.

> so I'm hoping the analysis/codegen benefits become more pronounced then.

I do plan (and hope) for this to be the case. Admittedly, it took me quite a bit of effort to convince myself that introducing yet-another IR is worth it, but the two other things I experimenting with:
1. having `genx` return the information to propagate back to the caller
2. passing information about the destination's syntax and semantics expectations into `genx` and then figure out what to do within (rather than having the caller do it)

were, in my opinion, not better, as they both increased the size and complexity of the already unwieldy `mirgen`.

The immediate benefit after this PR is that it removes a large amount of duplicated decision-making from part 2 of the MIR rework. This is what motivated the proto-MIR in the first place.

PR #1185, opened 2024-02-14
sem: `discardable` routines are discarded in `try`

### LINKED ISSUE #1178: discardable ignored in try/finally block
### Specification

The `discardable` pragma is ignored in `try` statements when the `except` clause is missing and a `finally` clause terminates the statement.

### Example

```nim
proc foo(): int {.discardable.} =
  42

try:
  foo()
finally:
  discard
```

### Actual Output
```
Error: expression 'foo()' is of type 'int' and has to be used (or discarded)
```
### Expected Output
```
```
### Additional Information

I've encountered errors with `discardable` which may have been in other contexts, so when debugging this I'd probably start with looking at the pragma and then move on to looking at `try`, though I think `try` was modified recently, so...  :shrug:

```
Nimskull Compiler Version 0.1.0-dev.21220 [linux: amd64]

Source hash: c604a5d6cb1548bff241bf7ccc8b5181182effbd
Source date: 2024-02-11
```

saem:
## Summary

Routines marked with the  `discardable`  pragma are now correctly
discarded inside  `try` / `except` / `finally`  expressions.

## Details

Previously the  `discardCheck`  which handled this did not correctly
traverse  `nkTryStmt`  when checking for  `implicitlyDiscardable` 
expressions.

Along with fixing the traversal, such traversals in an error case are no
longer repeated. Instead the traversal is done once, and only once. The 
`implicitlyDiscardable`  procedure is now simplified to only examine the
input node and no longer attempt traversal, which is now the
responsibility of the caller.

In addition to the fix, a test case has been added to ensure that this
behaviour is preserved, see  `trydiscardcheck_try_discardable` .

Fixes https://github.com/nim-works/nimskull/issues/1178

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -116,9 +116,10 @@ proc fixNilType(c: PContext; n: PNode) =
 
 # start `discard` check related code
 
-proc implicitlyDiscardable(n: PNode): bool =
-  var n = n
-  while n.kind in skipForDiscardable: n = n.lastSon
+func implicitlyDiscardable(n: PNode): bool =
+  ## true if a node, `n`, is implicitly discardable, meaning it's a control
+  ## flow ending statement in a block (`return`, `break`, `raise`, etc) or a
+  ## routine call to a routine type marked as `{.discardable.}`.
   result = n.kind in nkLastBlockStmts or
            (isCallExpr(n) and n[0].kind == nkSym and
            sfDiscardable in n[0].sym.flags)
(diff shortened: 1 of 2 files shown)


compiler/sem/semstmts.nim
```diff
@@ -116,9 +116,10 @@ proc fixNilType(c: PContext; n: PNode) =
 
 # start `discard` check related code
 
-proc implicitlyDiscardable(n: PNode): bool =
-  var n = n
-  while n.kind in skipForDiscardable: n = n.lastSon
+func implicitlyDiscardable(n: PNode): bool =
+  ## true if a node, `n`, is implicitly discardable, meaning it's a control
+  ## flow ending statement in a block (`return`, `break`, `raise`, etc) or a
+  ## routine call to a routine type marked as `{.discardable.}`.
```
zerbina, line 122:
```suggestion
  ## routine call to a routine marked as `{.discardable.}`.
```
A soft suggestion. Since the result being discardable is not a property of the procedure's type, mentioning its type could be confusing.

zerbina (approved):
Looks good to me, `implicitlyDiscardable` being more atomic in what it does is better.

compiler/sem/semstmts.nim
saem: I know we mark it on the symbol, but the symbol does multiple duties, one of which is a stand-in for the type of the procedure.

Pragmas are a bit funky because they commingle ones that apply to the AST, definition/backend, type, annotations, etc.

So I thought it sorta like a type property based on:
- expressions must be used in the language
- expressions are types + operational semantics
- void has nothing to use and unit are auto-consumed

Since the last property is describing things based on types, I thought it a type property.

zerbina:
> I got what seems to be an entirely unrelated error in `tfragment_alloc.nim`, where occupied memory exceeded expectations

I had to [disable](https://github.com/nim-works/nimskull/pull/1055) that test a while back in MacOS CI, because it kept failing. There was some conclusion I came to as to why this happened, but I forgot whether it was because of a latent bug with the allocator or something else.

compiler/sem/semstmts.nim
zerbina: > I know we mark it on the symbol, but the symbol does multiple duties, one of which is a stand-in for the type of the procedure.

My suggestion was more in the context of the implementation side of things (should've made that clearer, sorry), but I agree that whether a value is automatically discardable makes more sense as a type property.

I do think it's more a property of the *return* type rather than directly one of the *procedure* type. In any case, I don't think it's a problem if the comment stays as it is.

saem:
> I had to [disable](https://github.com/nim-works/nimskull/pull/1055) that test a while back in MacOS CI, because it kept failing. There was some conclusion I came to as to why this happened, but I forgot whether it was because of a latent bug with the allocator or something else.

If it annoys me enough I'll dive in deeper.

PR #1186, opened 2024-02-14
fix(sem): type-less nodes crashing the compiler

zerbina:
## Summary

Fix  `nkEmpty`  and  `nkType`  nodes without a type (which can be
created by a macro) appearing in some type contexts (such as the type of
a conversion) crashing the compiler. A "type expected" error is now
reported.

## Details

`semTypeNode`  assumed that both  `nkEmpty`  and  `nkType`  nodes have a
type assigned already, but this is not true for nodes created by macros.
If the caller of  `semTypeNode`  tried to access the returned type, an
NPE ensued.

It is now verified that both nodes have a type assigned, and if they do
not, a "type expected" user error is reported. An assertion is added to 
`semTypeNode`  to alert about erroneously missing types in the future.

There was one situation where an untyped  `nkEmpty`  node is legal: in
the body slot of a magic type definition such as 
`type set[T]{.magic: "Set".}` .  `typeSectionRightSidePass`  now detects
this case itself and doesn't call  `semTypeNode`  when the body slot
legally contains an  `nkEmpty`  node.

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -1967,10 +1967,11 @@ proc typeSectionRightSidePass(c: PContext, n: PNode) =
       # ``static type``
       rawAddSon(s.typ, newTypeS(tyNone, c))
       s.ast = a
-      inc c.inGenericContext
-      var body = semTypeNode(c, a[2], nil)
-      dec c.inGenericContext
-      if body != nil:
+      # magic type definitions are allowed to not have a body on the right
+      if s.magic == mNone or a[2].kind != nkEmpty:
+        inc c.inGenericContext
+        var body = semTypeNode(c, a[2], nil)
+        dec c.inGenericContext
         body.sym = s
         body.size = -1 # could not be computed properly
         s.typ[^1] = body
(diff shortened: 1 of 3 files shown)


saem:
Perhaps we should disallow `newNimNode` for these more internal kinds (e.g.: `nnkType`)? They're not really part of the syntax, even accounting for desugaring, they're at most semantic markers.

compiler/sem/semstmts.nim
```diff
@@ -1967,10 +1967,11 @@ proc typeSectionRightSidePass(c: PContext, n: PNode) =
       # ``static type``
       rawAddSon(s.typ, newTypeS(tyNone, c))
       s.ast = a
-      inc c.inGenericContext
-      var body = semTypeNode(c, a[2], nil)
-      dec c.inGenericContext
-      if body != nil:
+      # magic type definitions are allowed to not have a body on the right
+      if s.magic == mNone or a[2].kind != nkEmpty:
```
saem, line 1971: This is a reasonable approach, at best we could make `semTypeNode` treat this as a `tyNone`, or something, but that could a) lead to other issues given the current sloppiness in the compiler, and b) might not be all that principled and still require special casing here.

saem (approved):

saem:
Looks like a low likelihood transient failure in `tvmprofiler`, can probably re-run that after the remaining tasks complete and should be good to go.

PR #1189, opened 2024-02-15
fix(sem): return statements not being re-sem'ed

zerbina:
## Summary

Return statements of the form
`(ReturnStmt (Asgn (Sym "result") (...)))` weren't re-analyzed and
validated again after macro expansion, allowing macros to create
potentially ill-formed AST that crashes the compiler. This is now
fixed.

## Details

* if the `result` symbol in the assignment is correct, the `nkAsgn` is
  analyzed again, instead of being passed through
* the `setVar` inner procedure unnecessarily allocated a closure. This
  is fixed and the procedure is explicitly marked as a `.nimcall`


saem (approved):
I guess the lesson is that `nfSem` is about the only guard we have between two callsites in the compiler, if at any point any part of the AST might have gone through the macro layer.

PR #1191, opened 2024-02-16
don't emit header includes for unused imported globals

### LINKED ISSUE #1125: header include generated for unused imported variables
### Example

`x.nim`:

```nim
var x* {.importc, header: "does-not-exist".}: int
```

`test.nim`:

```nim
import x
```

### Actual Output

```
$ nim c test.nim

@mx.nim.c:8:10: fatal error: does-not-exist: No such file or directory
```

### Expected Output

Success

### Additional Information

* Found via Windows CI docgen for `posix_utils.nim`
* I'm not sure if this can even be classified as a bug

### LINKED ISSUE #1190: stdlib_*.nim.c files generated incompatible with my platform
Trying to compile a simple program for the `avr` architecture; works fine with nim, but with nimskull I end up with a handful of `stdlib_*.nim.c` files, one of which includes a header that 's not available for this particular architecture:

```
/home/ico/.cache/nimskull/main_r/stdlib_fenv.nim.c:8:10: fatal error: fenv.h: No such file or directory
 #include <fenv.h>
          ^~~~~~~~
compilation terminated.
```
I have no short and simple repro, but the code showing the problem is at https://github.com/zevv/cps-baremetal/tree/master

zerbina:
## Summary

Imported globals are now only included in the build if they're used
anywhere in alive code, fixing C headers being unnecessarily pulled in
for them.

Closes https://github.com/nim-works/nimskull/issues/1190.
Closes https://github.com/nim-works/nimskull/issues/1125.

## Details

An imported 'let' or 'var' was so far added to the (logical) struct of
the module the global is part of, but this is wrong! The definition of
an imported global doesn't imply the definition of a location owned by
the NimSkull program, rather it makes an external name available with
the given name and type.

**Changes:**
- don't add globals to the module struct if they're marked with
  `sfImportc`
- introduce `symbolToPmir` for querying the PMIR/MIR kind of a symbol
  without adding it to the MIR environment (`nameNode` does this)
- change `mirgen.genLocDef` to only use `nameNode` (which registers the
  symbol with the environment) if the statement constitutes an
  assignment
- add a specification test for the new behaviour

Imported globals are now only registered with the MIR environment (and
subsequently have their definition/include emitted) when they're part
of the alive program.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1243,10 +1243,10 @@ proc genLocDef(c: var TCtx, n: PNode, val: PNode) =
     s = n.sym
     hasInitializer = val.kind != nkEmpty
     sink = sfCursor notin s.flags
-    node = nameNode(c, s)
+    kind = symbolToPmir(s)
 
   c.builder.useSource(c.sp, n)
-  if node.kind == mnkGlobal and c.scopeDepth == 1:
+  if kind == pirGlobal and c.scopeDepth == 1:
     # no 'def' statement is emitted for top-level globals
     if hasInitializer:
       genAsgn(c, true, sink, n, val)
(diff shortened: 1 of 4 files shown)


saem (approved):
It took me a bit to convince myself (mostly tired), but it makes sense.

Even if alive, due to an assignment/read/write, it remains an external location made available under a certain name/type and not a NimSkull program location. The lazy inclusion of the header works nicely with the rest of conditional compilation, neat.

PR #1192, opened 2024-02-16
sem: re-sem all call arguments in macro output 

### LINKED ISSUE #1181: calls to procedures using type generated within macro fails if the macro output is passed to an another typed macro
### Example

```nim
import macros

macro modfn(n: typed): untyped =
  result = newStmtList()
  let sym = ident"O"
  result.add:
    nnkTypeSection.newTree:
      nnkTypeDef.newTree(
        sym,
        newEmptyNode(),
        nnkObjectTy.newTree(newEmptyNode(), newEmptyNode(), newEmptyNode())
      )
  let newprc = copyNimTree n
  newprc.params.add:
    nnkIdentDefs.newTree(
      ident"o",
      sym,
      newEmptyNode()
    )
  result.add newprc
  result.add:
    nnkCall.newTree(newprc.name):
      nnkCall.newTree(sym)

# if used outside of `expandMacros`, it will work
# proc foo() {.modfn.} = discard
expandMacros:
  proc foo() {.modfn.} = discard

```

### Actual Output

```
test.nim(28, 16) Error: type mismatch: got <test.O>
but expected one of:
proc foo(o: O)
  required type for o: O
  but expression 'O()' is of type: O

expression: foo(O())
```

### Expected Output

This should compile normally

### References

* Minimized from https://github.com/nim-works/cps/pull/316

### LINKED ISSUE #1183: calls introduced by macros does not exhibit automatic typedesc conversion
### Example

```nim
import macros

proc foo(T: typedesc) = discard

macro emit(): untyped =
  result = nnkCall.newTree(
    bindSym"foo", bindSym"int"
  )

emit()
```

### Actual Output

```
lib/core/macros.nim(618, 22) Error: type mismatch: got <int>
but expected one of:
proc foo(T: typedesc)
  required type for T: typedesc
  but expression 'int' is of type: int

expression: foo(int)
```

### Expected Output

This should work, as users could write `foo(int)` by hand and obtain an AST like:

```
Call
  Sym "foo"
  Sym "int"
```

### Possible Solution

* Workaround typically used in macros is to manually create `typedesc` nodes, but this does not work all the time.

zerbina:
Call operands are now re-analyzed when they're not marked as having been analyzed (via the `nfSem` flag) -- having a type or not is no longer relevant. This improves stability of inserting typed AST verbatim into macro output, since expression are now properly re-typed.

There is a lot of analysis that is *destructive* to the shape of AST, meaning that it's oftentimes not possible to discard the types of a typed expression, re-analyze it, and expect the correct type. For this reason, some special handling is required for some AST throughout sem.

Fixes https://github.com/nim-works/nimskull/issues/1183.
Fixes https://github.com/nim-works/nimskull/issues/1181.

---

## To-Do
* [x] (partially) keep the shape of array construction like `[1: x, ...]`
* [ ] fix the remaining problems and make all existing tests succeed again
* [ ] split out changes where possible
* [ ] add regression tests for the fixed issues
* [ ] write a proper commit message

## Notes for Reviewers
* still in progress, but the general direction should be visible

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


compiler/sem/sigmatch.nim
```diff
@@ -2608,7 +2612,7 @@ proc prepareOperand(c: PContext; formal: PType; a, aOrig: PNode): PNode =
   if formal.kind == tyUntyped:
     assert formal.len != 1
     result = aOrig
-  elif a.typ.isNil:
+  elif nfSem notin a.flags:
     # XXX This is unsound! 'formal' can differ from overloaded routine to
     # overloaded routine!
```
saem, line 2638: Soft suggestion, but you can remove this if you'd like.

I don't think this is valid anymore, because of `untyped` vs non-`untyped` argument isolation, the soon to be included `nfSem` flag guards, and the removal of the `tyIterable` hack a while back.

Varriount:
> There is a lot of analysis that is destructive to the shape of AST, meaning that it's oftentimes not possible to discard the types of a typed expression, re-analyze it, and expect the correct type

Out of curiosity, *why* is this the case? Is it just how the implementation was originally written? I would normally expect analysis procedures to be fairly side-effect free (at least with regards to their inputs).

zerbina:
> Out of curiosity, _why_ is this the case? Is it just how the implementation was originally written?

Most likely because it was simpler to implement. If typed AST doesn't have to be transformable back into exactly the AST is originated from, one has to put a less thought into the design of the semantic analysis phase.

> I would normally expect analysis procedures to be fairly side-effect free (at least with regards to their inputs).

Analysis here refers to the `semX` procedures of semantic analysis, which have a large amount of side-effects, many of them irreversible. A lot of work has went into removing input mutations from the `semX` procedure, so at least the input aspect has gotten a lot better.

With "destructive to the shape", I'm referring to things like the following:
```nim
type MyInt = distinct int
const C = MyInt(1)

discard C
# before analysis and typing, the untyped AST of the discard looks like
# this: `(DiscardStmt (Ident "C"))`
# the typed AST (i.e., the AST after analysis and typing) looks like this:
# `(DiscardStmt (IntLit 1))`
# Throwing away all types would mean that the information of the int literal
# being a `MyInt` is irreversibly lost 
```

saem:
> > There is a lot of analysis that is destructive to the shape of AST, meaning that it's oftentimes not possible to discard the types of a typed expression, re-analyze it, and expect the correct type
> 
> Out of curiosity, _why_ is this the case? Is it just how the implementation was originally written? I would normally expect analysis procedures to be fairly side-effect free (at least with regards to their inputs).

Because of things like automatically converting or dereferencing things, we need to indicate that somehow. There are also desugaring things, like `let a, b, c, d = 10`, getting converted to 4 individual `let`, but those are all largely benign; we could leave it alone, but then processing macro pragmas would become seriously problematic. There are also various ambiguities that need to be resolved, like `foo.bar` is that a field access or a call. Furthermore, there are some additional challenges, such as abstraction, which goes beyond simple desugaring, such as `for` loop conversion to `while`; although this isn't exposed to macros today, the inability to "pierce the veil" is actually problematic for things like the CPS library (they want to be able to see the `for` expansion). You can apply that last bit to lambda lifting, too and a variety of transform.

Now lots of this _is_ the current implementation's problem, but some of it isn't. The `typed` AST itself needs a grammar, because the fundamental way the compiler and macros should talk is through _syntax_, this is leveraging one of the big lessons of homoiconic languages (lisp, racket, etc), without altering the AST in key areas, this isn't possible.

PR #1195, opened 2024-02-16
rework the MIR (part 2)

zerbina:
## Summary

This introduces the `copy`, `move`, and `sink` *assignment modifiers*,
which make it explicit where and how ownership is transferred with
assignments. An assignment with an lvalue expression as the source
operand and no modifier is a *shallow copy*, replacing the need for
`fastAsgn`. Previously, the exact semantics of an assignment where an
implicit property of every assignment.

Besides slightly simplifying the `injectdestructors` pass and data-flow
graph construction, the semantics of assignments being properly encoded
in the MIR is a pre-requisite for many MIR-based optimization passes.
More generally, it's a significant step towards decoupling value
ownership semantics from lifetime hooks.

Semantics of existing code don't (intentionally) change, this is an
internal-only rework.

### MIR

* the `mnkCopy`, `mnkMove`, and `mnkSink` nodes are added, representing
  the modifiers. Syntax-wise, they may only appear directly in
  assignment source slots
* the `mnkFastAsgn` node is obsolete and removed
* `copy` and `move` assignments are *final* . That is, a `copy`
  assignment cannot become a `move` assignment, and the inverse is also
  not possible
* only `sink` assignments are non-final. Collapsing them into either a
  `copy` or `move` assignment is the focus of the move analyzer

### AST-to-MIR Translation

`mirgen` performs the initial placement of assignment modifier, using a
proto-MIR-based analysis. The decision-making works as follows:
1. when the destination is owning (i.e., all non-`.cursor` locals/
   globals/fields):
	1. if a non-destructive move is definitely possible, `move` is used
	2. if a move is never possible (e.g., because the source is a
     cursor), `copy` is used
	3. if whether a move is possible depends on a data-flow analysis,
     `sink` is used
	4. if a destructive move is possible, `sink` is used
	5. if the source expression is an rvalue expression returning an
     owning value (e.g., a call, a construction, etc.), and the
     assignment is an initial assignment, an in-place assignment is
     used
	6. for types with custom copy/sink/destroy behaviour, and non-initial
     assignments, the rvalue is first assigned to a temporary and then
     `move`d into the destination
2. when the destination is non-owning, no modifier is used

Except for the rule 1.6, placement of the modifiers is independent on
whether lifetime hooks are involved. 

In addition, all rvalue expressions (except for `cast`) are now treated
as returning *owning* values. What this means, in effect, is that
temporaries now properly use `def` instead of `def_cursor`. As the
moment, whether `def` or `def_cursor` is used for non-lifetime-hook
using types has no practical effect, but the new behaviour is
technically correct, while the previous one was not.

### Move Analyzer

Operation of the move analyzer changes slightly: instead of looking for
all `opConsume` data-flow instructions, it only looks for those
corresponding to `sink` assignments. Since whether a move or copy is
used is only relevant for lifetime-hook-using types, the `isLastRead`
analysis continues to only be performed for assignments of those types.

Compared to before, the DFG is now updated with the move analyzer
results, removing the need to pass them to each of the following
analysis routine.

### Assignment Rewriting

Rewriting of assignment now only looks for assignments with modifiers,
all assignments without a modifier are ignored. `move`s are exclusively
turned into `=sink` calls (never destructive moves), `copy`s into
`=copy`. `sink` is either turned into a copy, move, or destructive
move. Only the implementation is different, effective program behaviour
doesn't change.

### MIR-to-CGIR Translation

* `move` assignments and *shallow copy* assignments are translated to
  `cnkFastAsgn`
* all other assignments translate to `cnkAsgn`
* the small optimization preventing definitions being moved to the
  start of their scope is expanded to also apply to `def`s of
  temporaries without lifetime hooks, now that `def` is used for
  temporaries more often

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1166,26 +1166,35 @@ proc genReturn(c: var TCtx, n: PNode) =
 
   c.add MirNode(kind: mnkReturn)
 
-proc genAsgnSource(c: var TCtx, e: PNode, sink: bool) =
+proc genAsgnSource(c: var TCtx, e: PNode, status: set[DestFlag]) =
   ## Generates the MIR code for the right-hand side of an assignment.
-  ## The value is captured in a temporary if necessary for proper
-  ## destruction.
-  var e = exprToPmir(c, e, sink, false)
-  if not sink:
+  ## `status` provides the information necessary to decide what assignment
+  ## modifiers to use and whether a temporary is required.
+  ##
+  ## If not an initial assignment, and lifetime hooks are present, a temporary
+  ## is introduced for rvalue expressions that return owning values:
+  ##
+  ##   def _1 = get()
+  ##   dest = move _1
+  ##
+  ## This is necessary for the later hook injection, which triggers on
+  ## assignment modifiers, to work.
+  var e = exprToPmir(c, e, dfOwns in status, false)
+  if dfOwns in status:
+    wantOwning(e, dfEmpty notin status and hasDestructor(e.typ))
+  else:
     wantShallow(e)
+
   genx(c, e, e.high)
 
 proc genAsgn(c: var TCtx, dest: Destination, rhs: PNode) =
   assert dest.isSome
-  let owns = dfOwns in dest.flags
   let kind =
-    if owns:
-      if dfEmpty in dest.flags: mnkInit
-      else:                     mnkAsgn
-    else:                       mnkFastAsgn
+    if dfEmpty in dest.flags: mnkInit
+    else:                     mnkAsgn
   c.buildStmt kind:
     c.use dest.val
-    c.genAsgnSource(rhs, sink = owns)
+    c.genAsgnSource(rhs, dest.flags)
 
 proc unwrap(c: var TCtx, n: PNode): PNode =
   ## If `n` is a statement-list expression, generates the code for all
@@ -1225,16 +1234,20 @@ proc genAsgn(c: var TCtx, isFirst, sink: bool, lhs, rhs: PNode) =
     genWithDest(c, rhs, initDestination(dest, isFirst, sink))
   else:
     let kind =
-      if sink:
-        if isFirst: mnkInit
-        else:       mnkAsgn
-      else:         mnkFastAsgn
+      if isFirst: mnkInit
+      else:       mnkAsgn
+
+    var status: set[DestFlag]
+    if sink:
+      status.incl dfOwns
+    if isFirst:
+      status.incl dfEmpty
 
     c.buildStmt kind:
       # ``genLvalueOperand`` ensures that unstable lvalue
       # expressions are captured
       genLvalueOperand(c, lhs, true)
-      genAsgnSource(c, rhs, sink)
+      genAsgnSource(c, rhs, status)
 
 proc genLocDef(c: var TCtx, n: PNode, val: PNode) =
   ## Generates the 'def' construct for the entity provided by the symbol node
@@ -1267,7 +1280,9 @@ proc genLocDef(c: var TCtx, n: PNode, val: PNode) =
     c.buildStmt (if sfCursor in s.flags: mnkDefCursor else: mnkDef):
       c.add nameNode(c, s)
       if hasInitializer:
-        genAsgnSource(c, val, sink)
+        genAsgnSource(c, val):
+          if sink: {dfEmpty, dfOwns}
+          else:    {dfEmpty}
       else:
         c.add MirNode(kind: mnkNone)
 
@@ -1319,7 +1334,7 @@ proc genVarTuple(c: var TCtx, n: PNode) =
     let val = c.allocTemp(initExpr.typ)
     c.buildStmt mnkDefUnpack:
       c.use val
-      genx(c, initExpr, consume = true)
+      genAsgnSource(c, initExpr, {dfEmpty, dfOwns})
 
     # generate the unpack logic:
     for i in 0..<numDefs:
@@ -1778,6 +1793,18 @@ proc genx(c: var TCtx, e: PMirExpr, i: int) =
         Destination(isSome: true, val: tmp, flags: {dfOwns, dfEmpty})
 
     c.use tmp
+  of pirCopy:
+    c.buildOp mnkCopy, n.typ:
+      recurse()
+  of pirMove:
+    c.buildOp mnkMove, n.typ:
+      recurse()
+  of pirSink, pirDestructiveMove:
+    # a destructive move is currently not translated into a move + wasMoved,
+    # but rather into a sink, which is then, if necessary, later turned into
+    # a destructive move
+    c.buildOp mnkSink, n.typ:
+      recurse()
   of pirMat, pirMatCursor:
     let f = c.builder.push: recurse()
     # only materialize a temporary if the expression is not already a
@@ -1858,7 +1885,7 @@ proc gen(c: var TCtx, n: PNode) =
                           field: dest[^1].field):
           genx(c, dest, dest.len - 2)
 
-        genAsgnSource(c, n[1], false) # the source operand
+        genAsgnSource(c, n[1], {dfOwns}) # the source operand
     else:
       # a normal assignment
       genAsgn(c, false, true, n[0], n[1])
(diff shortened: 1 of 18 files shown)


compiler/mir/proto_mir.nim
```diff
@@ -286,15 +291,107 @@ func isStable(e: seq[ProtoItem], n: int): bool =
...
+    else:       kind
+
+  case e[i].kind
+  of pirParam:
+    if e[i].typ.kind == tySink:
+      pirSink
+    else:
+      pirCopy
+  of pirLocal, pirGlobal:
+    if sfCursor in e[i].sym.flags:
```
saem, line 312: Any `importc` location isn't owned by the NimSkull program, so shouldn't that necessitate a copy or are we leaving this up to the move analyser when collapsing the `sink`?

compiler/mir/proto_mir.nim
```diff
@@ -286,15 +291,107 @@ func isStable(e: seq[ProtoItem], n: int): bool =
   else:
     unreachable(e[n].kind)
 
+func ownershipOp(e: seq[ProtoItem], i: int): ProtoItemKind =
+  ## Infers and returns the best fitting operation to retrieve an owning
+  ## value from the given *lvalue*.
+  proc decayMove(kind: ProtoItemKind): ProtoItemKind {.inline.} =
```
saem:
```suggestion
  func decayMove(kind: ProtoItemKind): ProtoItemKind {.inline.} =
```

Soft suggestion, but figured a `func` is both sufficient and in keeping with the outer `func ownershipOp`.

compiler/mir/proto_mir.nim
```diff
@@ -286,15 +291,107 @@ func isStable(e: seq[ProtoItem], n: int): bool =
...
+    pirCopy
+  of pirStmtList, pirMatLvalue:
+    ownershipOp(e, i - 1)
+  else:
+    # cannot be part of an lvalue expression sequence
+    unreachable(e[i].kind)
+
+func wantOwning*(e: var seq[ProtoItem], forceTemp: bool) =
+  ## Makes sure `e` produces an owning value. If `forceTemp` is true, a
+  ## temporary is materialized even if the expression would already produces
```
saem:
```suggestion
  ## temporary is materialized even if the expression would already produce
```

compiler/mir/proto_mir.nim
```diff
@@ -286,15 +291,107 @@ func isStable(e: seq[ProtoItem], n: int): bool =
...
+  of OwnedRvalue:
+    var i = e.high
+    while e[i].kind == pirStmtList:
+      dec i
+    case e[i].kind
+    of pirMat:
+      e.add pirMove
+    of pirComplex:
+      # watch out! try-finally expressions can have exceptional control-flow
+      # that forces the destination temporary to have be destroyed in a
```
saem:
```suggestion
      # that forces the destination temporary to have to be destroyed in a
```

compiler/mir/mirgen.nim
```diff
@@ -1166,26 +1169,35 @@ proc genReturn(c: var TCtx, n: PNode) =
...
-  var e = exprToPmir(c, e, sink, false)
-  if not sink:
+  ## `status` provides the information necessary to decide what assignment
+  ## modifiers to use and whether a temporary is required.
+  ##
+  ## If not an initial assignment, and lifetime hooks are present, a temporary
+  ## is introduced for rvalue expressions that return owning values:
+  ##
+  ##   def _1 = get()
+  ##   dest = move _1
```
saem, line 1181: So this is the extra temporary for non-initial assignment. Is the idea that in the future after hook injection happens, the code generator could look for this pattern and not generate the temporary if it's not required?

compiler/mir/mirgen.nim
```diff
@@ -1225,16 +1237,20 @@ proc genAsgn(c: var TCtx, isFirst, sink: bool, lhs, rhs: PNode) =
...
-        else:       mnkAsgn
-      else:         mnkFastAsgn
+      if isFirst: mnkInit
+      else:       mnkAsgn
+
+    var status: set[DestFlag]
+    if sink:
+      status.incl dfOwns
+    if isFirst:
+      status.incl dfEmpty
```
saem, line 1247: Longer term, outside of this PR/whenever it's convenient, I would suggest staying away from boolean parameters ("boolean trap"), and favour sets, or as alternative always pass via named parameters.

compiler/mir/mirgen.nim
```diff
@@ -1331,9 +1349,13 @@ proc genVarTuple(c: var TCtx, n: PNode) =
       # generate the assignment:
       c.buildStmt (if isInit: mnkInit else: mnkAsgn):
         genOperand(c, lhs)
-        c.subTree MirNode(kind: mnkPathPos, typ: lhs.typ,
-                          position: i.uint32):
-          c.use val
+        # the temporary tuple is only used for unpacking; it can always be
+        # moved out of. The temporary tuple is not destroyed, so no
+        # destructive move is required
+        c.buildTree mnkMove, lhs.typ:
```
saem, line 1357: Here we're relying on the move analyser to insert a copy in case something accesses the tuple/rhs after the assignment?

saem (approved):
A few small suggestions, but overall the change looks good.

compiler/mir/proto_mir.nim
zerbina: That's a very good question. While it might be a bit restrictive, I do think it'd be better to necessitate a copy, like you said, at least for now. Ultimately, it depends on what assumptions the compiler should be allowed to make about an `importc`'ed location.

When the expectations/requirements become clearer, the necessitated copy could be loosened, so that:
```nim
var external {.importc.}: seq[int]
external = @[1, 2, 3]
var other = external # this could move, since `external` is immediately reassigned after 
external = @[2, 3, 4]
var other2 = external # cannot move, since the compiler doesn't know how/where/if
                      # `external` is used afterwards
```

Since forcing a copy would be a change in semantics, I'd do this as a follow-up, if that's okay with you.

compiler/mir/mirgen.nim
zerbina: Yep, that's the plan.

compiler/mir/mirgen.nim
zerbina: Not quite. By way of how the translation works, the field of the temporary is guaranteed to never be used afterwards.

For context, this code is responsible for the translation of `var (a, b) = f()`. Previously, the following MIR was produced by `mirgen`:
```
def_unpack _1 = get()
def a = _1.0
def b = _1.1
```
`def_unpack` is a special kind of definition (read: hack) that makes `_1` visible to the move analyzer, but disables destructor injection.

With the explicit modifiers, `mirgen` can now produce:
```
def_unpack _1 = get()
def a = move _1.0
def b = move _1.1
```
effectively doing the work the move analyzer previously had to do. Since all the parts are guaranteed to be moved out of `_1`, and no control-flow is possible during the unpacking, `_1` doesn't have to be destroyed, and thus a normal move (instead of a destructive move) is fine.

compiler/mir/proto_mir.nim
saem: Thanks for the detailed answer, the change in semantics should not be part of this PR, it's really about the refactor

compiler/mir/mirgen.nim
saem: Ah, got it. I was thinking of the following, which I should have added to the initial comment, `var someTuple = ...; var (a,b) = someTuple; ...`. here `someTuple` could be used afterwards so I was trying to suss out when/where we do a defensive copy, my thinking was the move analyser because that's when we'd know if the assignment was the last read.

zerbina: Ah, I see. Thank you for clarifying.

> my thinking was the move analyser because that's when we'd know if the assignment was the last read.

Yep, you're thinking thinking is correct. For the `var (a,b) = someTuple` case, `mirgen` produces the following:
```
def_unpack _1 = sink someTuple
...
```
leaving the work of figuring out whether to copy or move the tuple the move analyzer. The relevant part in `mirgen` is [here](https://github.com/nim-works/nimskull/pull/1195/files/faf45f05831e803a723fa0276fb2f0d7bdc52c59#diff-dc0794500e842ec99b3d7e1f3da2cb00fa6ceb7f88f89b428b7df50766f5bca8R1340)

I'll adjust the comment to make this relationship clearer.

zerbina:
/merge

Thank you for the review and questions, @saem.

PR #1196, opened 2024-02-17
cgen: fix undefined behaviour with object conversion 

### LINKED ISSUE #1193: unconditional pointer member access is generated in C for inheritance
### Example

```nim
type
  O = ref object of RootObj

proc foo(o: ref RootObj) =
  echo "some side-effect"

var o: O = nil
foo(o)
```

### Actual Output

```
$ nim r -d:release -d:useMalloc --passC:-fno-omit-frame-pointer --passC:-mno-omit-leaf-frame-pointer --passC:-fsanitize=undefined --passL:-fsanitize=undefined --debuginfo --linedir test.nim

test.nim:8:5: runtime error: member access within null pointer of type 'struct tyObject_OcolonObjectType___31J4fwfXSYbejl8P6EY8zw'
some side-effect
some side-effect
```

### Expected Output

```
some side-effect
```

### Possible Solution
Don't generate unconditional access

### Additional Information

Generated C code:

```c
	o__test_5 = ((tyobject_ocolonobjecttype___31j4fwfxsybejl8p6ey8zw*) nim_nil);
	_1 = &o__test_5->sup; // Unconditional NULL dereference
	foo__test_3(_1);
```

While most compiler will generate rather benign code for this (ie. just a pointer offset), this is still undefined behavior.

It's not high priority, but it prevents effective usage of undefined behavior sanitizer with nimskull.

zerbina:
## Summary

Fix up-conversions of `nil` `ref object` or `ptr object` values
invoking undefined behaviour at the C level.

Fixes https://github.com/nim-works/nimskull/issues/1193.

## Details

Up conversions of pointers or `ref`s were translated into a
dereference + address-of sequence (e.g., `&a->Sup`), which invokes UB
if the pointer is null.

Instead, the pointer is now directly cast to the super type. According
to the C89 standard (6.5.2.1 "Structure and union specifiers"), "A
pointer to a structure object, suitably converted, points to its
initial member", so this cast is legal.

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -1943,10 +1943,15 @@ proc upConv(p: BProc, n: CgNode, d: var TLoc) =
                 a.storage)
     # an indirection is used:
     d.flags.incl lfIndirect
+  elif isRef:
+    # using ``&(x->Sup)`` is undefined behaviour when x is null, so the
+    # pointer has to be cast instead
+    putIntoDest(p, d, n,
+                "(($1) ($2))" % [getTypeDesc(p.module, n.typ), rdLoc(a)])
   else:
-    var r = rdLoc(a) & (if isRef: "->Sup" else: ".Sup")
+    var r = rdLoc(a) & ".Sup"
     for i in 2..inheritanceDiff(src, dest): r.add(".Sup")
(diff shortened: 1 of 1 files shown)


saem (approved):
Regression tests would be good, but I'm guessing the dependency would need a bit of test infra work to detect if it's available. Alaviss might have a more comprehensive plan he wants to put into place, so that can be separate.

PR #1199, opened 2024-02-18
sem: template `typed` param only  in using pos

saem:
## Summary

Disallow  `typed`  parameter template substitution in defining
positions, restricting them to using position substitution only.

## Details

Updated  `semtempl`  to only allow  `untyped`  parameters to be inserted
into definition positions during template parameter substitution. This
is based on lessons learned following the initial changes to template
parameter substitution rules in:
483726f957da5660268d0f72120ec234509a7b0c via the PR
https://github.com/nim-works/nimskull/pull/605

Along with this change the substitution test has been expanded and
manual updated.


tests/lang_callable/template/template_usage_substitution_typed_params.nim
```diff
@@ -0,0 +1,32 @@
+discard """
+  description: '''
+Templates `typed` and non-AST parameters do not replace identifiers in new
+symbol definition positions. Meaning a template parameter that is not `untyped`
```
zerbina:
```suggestion
Except for `untyped`, template parameters do not replace identifiers in new
symbol definition positions. Meaning a template parameter that is not `untyped`
```
I think this is bit easier to understand. It's a soft suggestion, feel free to ignore.

compiler/sem/semtempl.nim
```diff
@@ -259,7 +259,7 @@ func isTemplParam(c: TemplCtx, n: PNode): bool {.inline.} =
 func definitionTemplParam(c: TemplCtx, n: PNode): bool {.inline.} =
   ## True if `n` is an AST typed (`typed`/`untyped`) parameter symbol of the
   ## current template
```
zerbina:
```suggestion
  ## True if `n` is an `untyped` parameter symbol of the current template.
```

zerbina (approved):
Looks good to me, and I agree that this is the more intuitive behaviour.

For your info: I've fixed a few typos in the PR description.

tests/lang_callable/template/template_usage_substitution_typed_params.nim
saem: Yup, definitely easier to read. Thanks.

saem:
Thanks for the review and suggestions, @zerbina.

PR #1201, opened 2024-02-19
refactor: remove `Values` type from `analysis`

zerbina:
## Summary

Replace the `Values` usages in `injectdestructors` with directly using
a `PackedSet`, and remove the `Values` type.

### DIFF
--- a/compiler/sem/injectdestructors.nim
+++ b/compiler/sem/injectdestructors.nim
@@ -175,11 +175,15 @@ type
     needsFinally: bool  ## whether the destructor needs to be placed in a
                         ## 'finally' clause
 
+  Moves = PackedSet[OpValue]
+    ## A set storing the operands of all sinks that were collapsed into
+    ## moves.
+
   AnalysisResults = object
     ## Bundled-up immutable state needed for assignment rewriting. Since
     ## they're immutable, ``Cursor``s are used in order to not copy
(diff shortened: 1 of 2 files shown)


compiler/sem/injectdestructors.nim
```diff
@@ -175,11 +175,15 @@ type
...
+  Moves = PackedSet[OpValue]
+    ## A set storing the operands of all sinks that were collapsed into
+    ## moves.
+
   AnalysisResults = object
     ## Bundled-up immutable state needed for assignment rewriting. Since
     ## they're immutable, ``Cursor``s are used in order to not copy
     # XXX: ideally, view types (i.e. ``lent``) would be used here
-    v: Cursor[Values]
+    moves: Cursor[Moves]
```
saem, line 186:
```suggestion
    moved: Cursor[Moves]
```

**Very** soft suggestion, but they happened in the past, although by that token, `destroy` should be renamed to `destroyed`. Honestly, I could go either way. The rename would undoubtely have some cascade renames.

saem (approved):
I'm good to merge it as is, the suggestion around the naming of `moves` is unlikely to make things that much more clear.

compiler/sem/injectdestructors.nim
zerbina: I think I'd keep `moves` for now. My thinking behind not using past tense was/is that the set tells which sinks are moves.

PR #1204, opened 2024-02-20
mir: separate hook injection from move analyzer

zerbina:
## Summary

Make turning copy and move assignments into `=copy` and `=sink` calls a
standalone MIR pass; semantics don't change, but higher quality MIR
code is produced. This is progress towards fully decoupling lifetime
hooks from the move analyzer / destructor injection.

### General architecture

* the `injectdestructors` pass/module only collapses `sink` assignments
  and updates the MIR tree accordingly
* turning copy assignments into `=copy` and sink assignments into
  `=sink` calls is done by the new `injecthooks` pass/module
* reporting errors and warnings regarding hooks is also made part of
  the `injecthooks` pass
* the `injecthooks` is run after the `injectdestructors` pass
* handling of `--expandArc` is moved to `backends.process`, so that it
  can happen after hook injection

### `injectdestructors` Pass

* assignments are not replaced with `=copy` or `=sink` calls
* `expandAsgn` is renamed to `specializeAsgn`
* `sink` assignments are turned into either `copy`, `move`, or
  destructive `move` assignments (same as before)
* assignments to locations that don't yet store a value are turned into
  *initializing* assignments (this is necessary for the later `=sink`
  injection to work)
* compared to before, *all* assignments are processed by
  `specializeAsgn`, not only those for locations with lifetime hooks
* `=destroy` hooks are still injected

### `injecthooks` Pass

* the pass looks for `move` and `copy` assignment modifiers
  * if the involved type has lifetime hooks, the assignment is
     replaced with a call to the appropriate hook
  * the same injection rules as used previously apply
* the error detection and reporting logic is moved over - without
  change - from `injectdestructors`
* an adapter procedure to the legacy pass managements is provided
  (needed by `backends`)

With less surrounding complexity, more effort is spent on using less
temporaries for the hook call injection:
* no temporary is used for the destination operand
* no temporary is used for the source operand when it's guaranteed to
  not alias the destination. This is the case for:
  * `move` assignments
  * define with `copy` (e.g., `def x = copy y`)
  * initial assignments with `copy` (e.g., `x := copy y`)

Less temporaries means less work for the pass eliminating unnecessary
temporaries.

### Technical Correctness

* `mirconstr.inline` is only usable with lvalue expressions and always
  creates a non-owning temporary (instead of an owning temporary)
* the remaining usage of `mirconstr.inline` in
  `injectdestructors.lowerBranchSwitch`, where an owning temporary is
  required, is replaced with using `wrapTemp`

### DIFF
--- a/compiler/mir/injecthooks.nim
+++ b/compiler/mir/injecthooks.nim
@@ -193,7 +193,7 @@ proc injectHooks*(body: MirBody, graph: ModuleGraph, env: var MirEnv,
         # with:
         #   def_cursor _1 = a.b
         #   =copy(name x, arg _1)
-        # XXX: the temporary could be emitted in more cases by using proper
+        # XXX: the temporary could be omitted in more cases by using proper
         #      alias analysis
         changes.replaceMulti(tree, stmt, bu):
           let tmp = bu.inline(tree, src)
--- a/compiler/mir/injecthooks.nim
+++ b/compiler/mir/injecthooks.nim
@@ -1,0 +1,259 @@
+## Implements the MIR pass for replacing copy and move assignments with the
+## ``=copy`` or ``=sink`` hook (if available for the type).
+##
+## Future direction: injection of ``=destroy`` hooks also needs to happen
+## here.
+
+import
+  compiler/ast/[
+    ast_query,
+    ast_types,
+    lineinfos
+  ],
+  compiler/front/[
+    msgs,
+    options
+  ],
+  compiler/mir/[
+    mirbodies,
+    mirchangesets,
+    mirconstr,
+    mirenv,
+    mirtrees
+  ],
+  compiler/modules/[
+    modulegraphs
+  ],
+  compiler/utils/[
+    idioms
+  ]
+
+# XXX: reports are a code smell meaning data types are misplaced
+from compiler/ast/reports_sem import SemReport
+from compiler/ast/report_enums import ReportKind
+
+# XXX: temporary dependency until destroy hooks are injected here
+from compiler/sem/injectdestructors import getOp, genDestroy, buildVoidCall
+
+from compiler/sem/liftdestructors import boolLit, cyclicType
+
+type
+  LocalDiagKind = enum
+    ldkPassCopyToSink       ## a copy is introduced in a consume context
+    ldkUnavailableTypeBound ## a type-bound operator is requested but not
+                            ## available
+
+  LocalDiag = object
+    ## A temporary diagnostic representation that is later turned into a
+    ## ``SemReport``
+    pos: NodePosition ## the location of the report
+    case kind: LocalDiagKind
+    of ldkUnavailableTypeBound:
+      op: TTypeAttachedOp
+    of ldkPassCopyToSink:
+      discard
+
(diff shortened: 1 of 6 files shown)


compiler/mir/injecthooks.nim
```diff
@@ -0,0 +1,259 @@
...
+  compiler/utils/[
+    idioms
+  ]
+
+# XXX: reports are a code smell meaning data types are misplaced
+from compiler/ast/reports_sem import SemReport
+from compiler/ast/report_enums import ReportKind
+
+# XXX: temporary dependency until destroy hooks are injected here
+from compiler/sem/injectdestructors import getOp, genDestroy, buildVoidCall
```
saem, line 36: Was going to ask, thanks for clarifying.

compiler/mir/injecthooks.nim
```diff
@@ -0,0 +1,259 @@
...
+        #   =copy(name x, arg a.b)
+        changes.replaceMulti(tree, stmt, bu):
+          genCopy(bu, graph, env, op, tree, dest, true):
+            bu.emitFrom(tree, src)
+      of mnkAsgn:
+        # the source and destination could overlap. Replace ``x = copy a.b``
+        # with:
+        #   def_cursor _1 = a.b
+        #   =copy(name x, arg _1)
+        # XXX: the temporary could be emitted in more cases by using proper
```
saem:
```suggestion
        # XXX: the temporary could be omitted in more cases by using proper
```

I'm guessing you mean 'omitted', because alias analysis would tell us whether the source and destination do in fact overlap.

compiler/sem/injectdestructors.nim
```diff
@@ -667,118 +598,57 @@ proc expandAsgn(tree: MirTree, ctx: AnalyseCtx, ar: AnalysisResults,
...
+      # the value cannot be moved, turn the sink into a copy
       c.changeTree(tree, operator):
-        MirNode(kind: mnkMove, typ: tree[operator].typ)
+        MirNode(kind: mnkCopy, typ: tree[operator].typ)
 
+    if tree[stmt].kind == mnkAsgn and not isAlive:
+      # the assignment initializes the location
+      c.changeTree(tree, stmt): MirNode(kind: mnkInit)
+  else:
+    # it's a move or move already, so nothing to change there
```
saem:
```suggestion
    # it's a move or destructive move already, so nothing to change there
```

Not sure if 'destructive' is what you meant, but IIRC, mir/proto-mir can generate either if safe, and will do so instead of emiting a `sink` for injectdestructors to resolve. That could be the case if the destination had no value to begin with or perhaps the location was reset at that point (`isAlive == false`).

saem (commented):
A few minor doc typos, but looks good otherwise.

compiler/sem/injectdestructors.nim
zerbina: Ah, good catch -- a copy & past mistake. I meant to use "copy" here.

zerbina:
```suggestion
    # it's a move or copy already, so nothing to change there
```

PR #1206, opened 2024-02-24
mir: separate destructor from `=destroy` injection

zerbina:
## Summary

Introduce the new `destroy` MIR operation and use it to decouple
destructor injection from the injection of `=destroy` hooks. Injection
of lifetime hooks is now fully decoupled from the move-analyzer /
destructor-injection pass. This is an internal-only change.

## Details

* introduce the `mnkDestroy` node kind / operator
* instead of `=destroy` hook calls, the `injectdestructors` pass only
  injects `mnkDestroy` operations
* the `injecthooks` pass then replaces `mnkDestroy` operations with
  `=destroy` calls
* the `getOp` and `genDestroy` procedure are moved to `injecthooks`
  without change
* injection of `=destroy` calls for variant objects (part of the 
  `mnkSwitch` lowering) is still part of the `injectdestructors` pass
* since `mnkDestroy` operations are replaced right away, both `mirexec`
  and `cgirgen` (MIR-to-CGIR translation) don't handle them yet

### DIFF
--- a/compiler/backend/cgirgen.nim
+++ b/compiler/backend/cgirgen.nim
@@ -741,6 +741,8 @@ proc stmtToIr(tree: MirBody, env: MirEnv, cl: var TranslateCl,
       res.kids.addIfNotEmpty stmtToIr(tree, env, cl, cr)
   of mnkScope:
     toSingleNode scopeToIr(tree, env, cl, cr)
+  of mnkDestroy:
+    unreachable("a 'destroy' that wasn't lowered")
   of AllNodeKinds - StmtNodes:
     unreachable(n.kind)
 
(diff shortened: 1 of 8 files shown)


saem (approved):
IIUC this will be used to push `mnkDestroy` injection right upto the AST to MIR boundary. Is the intention to shift the actual call injection deeper into backends, so we can do more backend specific things or will they remain unreachable node kinds for the foreseeable future?

zerbina:
> IIUC this will be used to push `mnkDestroy` injection right upto the AST to MIR boundary

Yep, that's correct.

> Is the intention to shift the actual call injection deeper into backends, so we can do more backend specific things or will they remain unreachable node kinds for the foreseeable future?

I think the actual call injection is going to stay roughly where it is, though it might be partially disabled for the VM backend. Data-flow graph creation (`mirexec`) will have to implement the node kind in the shorter term, while `cgirgen`, in the near future, most likely won't.

saem:
/merge

Thanks for the clarification.

PR #1207, opened 2024-02-25
sem: varargs now have and conform to a specification

saem:
## Summary
* created a battery of tests specifying `vararg` parameters and to a lesser extent the pragma
* updated the implementation for varargs argument and parameter handling to conform to the aforementioned tests
* updated the docs to describe the latest behaviour and indicate future direction

## Details
* anything else

---

## Notes for Reviewers
* the plan for this PR is to merge in tests and the rework of `sigmatch` in one go
  * reason for doing both simultaneously is because the spec itself needs to be vetted through implementing
* created initial as a non-draft PR to ensure the existing tests/knownIssues are at least green in CI
* rough plan, in case you have feedback on the approach:
  - [x] get initial tests in place
  - [ ] add docs to tests clarifying the intention from the notes remaining in `tvarargs`
  - [ ] initial revision of docs (`manual.rst`) to indicate new behaviour and future direction
  - [ ] rewrite `sigmatch.matchesAux` which contains the parameter handling
        - [ ] any related updates/changes
  - [ ] write a proper PR title and message

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


saem:
One of the things blocking this PR is that the core [matchesAux](https://github.com/nim-works/nimskull/blob/bf52661c0df03f9fb65354e85f0b76183a425934/compiler/sem/sigmatch.nim#L2786) loop needs to work on a per formal parameter basis instead of per argument as it is now. This allows for the progress of matching against formals to be controlled more precisely.

As part of that I believe one of the things that needs to happen is to move default parameter handling/setting to happen in `matchesAux` as well, instead of in `matches`, and possibly `matchesGenericParams`.

PR #1209, opened 2024-02-26
cgen: better default initialization for `result`

zerbina:
## Summary

Use a MIR pass for handling the default-initialization of `result`
variables for the C backend. This allows using the MIR's more precise
data-flow analysis, allowing for broader omission of `result` variable
initialization.

## Details

* introduce the `injectResultInit` MIR pass
  * it's only used with the C backend for now
  * if a data-flow analysis deems it required, a `result := default()`
    assignment is placed at the start of the body
* remove the `CgNode`-based `result`-assignment analysis from `cgen`

### Data-flow Analysis Bug

A small bug with the `mirexec.traversal` (forward traversal) routine
was discovered and fixed: the `exit` flag was erroneously set to true
when the main path was aborted on a data-flow instruction that's the
last instruction in the subgraph. The `exit` flag must only be true
when the end of the subgraph is reached, so this is wrong.

A regression test for the `traverse` routine is added.

This bug only affected the `injectdestructors.needsReset` optimization
(unnecessary but otherwise harmless `wasMoved` calls were injected
because of it) and `injectResultInit` (unnecessary initialization was
emitted).

### DIFF
--- a/compiler/backend/cgen.nim
+++ b/compiler/backend/cgen.nim
@@ -730,117 +730,6 @@ proc closureSetup(p: BProc, prc: PSym) =
   linefmt(p, cpsStmts, "$1 = ($2) ClE_0;$n",
           [rdLoc(p.locals[n.local]), getTypeDesc(p.module, ls.typ)])
 
-func containsResult(n: CgNode): bool =
-  result = false
-  case n.kind
-  of cnkAtoms - {cnkLocal}:
-    discard "ignore"
-  of cnkLocal:
-    if n.local == resultId:
-      result = true
-  of cnkWithOperand:
-    result = containsResult(n.operand)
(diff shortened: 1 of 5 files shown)


saem (approved):

saem:
/merge

FYI: I made a small edit to the PR body, there was a minor typo.

saem:
/merge

fragmentation test is flaky, might need to disable it or bump the value, but just in case it's transient trying again.

saem:
Hmm, I would simply disable the test on windows for now and we can reenable it in the future once we have more of a handle on how this stuff should behave.

zerbina:
/merge

I'm giving it one more go. If it fails again, I'll disable the test here.

PR #1212, opened 2024-02-28
make `.computedGoto` pragma a no-op

zerbina:
## Summary

The `.computedGoto` pragma is now a no-op pragma usable in pragma
statements everywhere. It's seldomly useful, and since the current
implementation complicates the rework of the C code generator,
dedicated C code generator support is removed for now.

## Details

* remove the `.computedGoto` implementation from `cgen`
* remove `compat.flattenStmts`, `ccgutils.stmtsContainsPragma`, and
  `ccgutils.getPragmaStmt`; they were only used by the `.computedGoto`
  implementation
* as a consequence, `cnkPragmaStmt` is no longer used; the node kind is
  removed too
* without `cnkPragmaStmt`, `mnkPNode` is also obsolete; it's removed
  too
* remove the now-unused report kinds associated with the
  `.computedGoto` pragma


saem (approved):
Wow, that was a lot of stuff to support that little pragma. Glad to see it go.

PR #1213, opened 2024-02-28
fix(mirgen): respect `--panics:on` for overflow checks

zerbina:
## Summary

Fix overflow checks for binary integer arithmetic being treated as
potentially raising exceptions when panics are enabled (`--panics:on`).
This didn't affect correctness, but some optimizations were inhibited.

## Details

Use `buildDefectMagicCall` for emitting the binary arithmetic operation
in `mirgen`. If panics are enabled, this makes sure a `mnkCall` is
emitted rather than a `mnkCheckedCall`. 

Checked calls (i.e., calls that can raise an exception) introduce
unstructured control-flow, which, depending on where the call is
located, can result in the compiler having to place additional
`wasMoved` and/or destructor calls.

An `--expandArc`-based test for making sure that all run-time checks
are correctly treated as not raising an exception is added.

### DIFF
--- a/tests/exception/truntime_check_panics.nim
+++ b/tests/exception/truntime_check_panics.nim
@@ -1,0 +1,58 @@
+discard """
+  description: '''
+    Ensure that the compiler-inserted run-time checks don't have exceptional
+    exits when panics are enabled.
+  '''
+  targets: native
+  matrix: "--panics:on --hints:off --expandArc:test"
+  action: compile
+  nimout: '''
+--expandArc: test
+scope:
+  def a: array[0..0, int]
+  chckIndex(arg a, arg i)
+  discard a[i]
+  chckBounds(arg a, arg 0, arg i)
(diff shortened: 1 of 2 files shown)


tests/exception/truntime_check_panics.nim
```diff
@@ -0,0 +1,58 @@
...
+  action: compile
+  nimout: '''
+--expandArc: test
+scope:
+  def a: array[0..0, int]
+  chckIndex(arg a, arg i)
+  discard a[i]
+  chckBounds(arg a, arg 0, arg i)
+  def _0: openArray[int] = toOpenArray a, 0, i
+  def _1: int = addI(arg i, arg i)
```
zerbina, line 17: Prior to the fix, this was:
```
  def _1: int = addI(arg i, arg i) (raises)
```
(notice the `raises`)

saem (approved):

saem:
/merge

Looks like we got SDK'd again?

PR #1214, opened 2024-02-29
internal: implement `--profiler:on` support with MIR pass

zerbina:
## Summary

Make routine instrumentation with `nimProfile` calls a MIR pass, moving
more logic out of the C code generator and working towards
`--profiler:on` support with all backends. `nimProfile` calls are also
no longer inserted into loops within `.asmNoStackFrame` routines.

## Details

* require a `ModuleGraph` instance for `mirpasses.applyPasses`, so that
  compilerprocs can be looked up
* add the `injectProfilerCalls` MIR pass. It injects `nimProfile` in
  the same way that `cgen` does
* remove injection of `nimProfile` calls from `cgen`
* don't instrument loops within `.asmNoStackFrame` procedures;
  injection of `nimProfile` at procedure entry was already skipped for
  these procedures
* temporarily remove the `optProfiler` option from symbols when in JIT
  mode, that code generated for compile-time execution is not
  instrumented
* add a test for the `--profiler:on` feature, to make sure it works and
  keeps working

### DIFF
--- a/compiler/mir/mirpasses.nim
+++ b/compiler/mir/mirpasses.nim
@@ -8,7 +8,11 @@ import
   compiler/ast/[
     ast_query,
     ast_types,
+    lineinfos,
     types
+  ],
+  compiler/front/[
+    in_options
   ],
   compiler/mir/[
     analysis,
@@ -19,6 +23,10 @@ import
     mirconstr,
     mirtrees,
     sourcemaps
+  ],
+  compiler/modules/[
+    modulegraphs,
+    magicsys
   ],
   compiler/sem/[
     aliasanalysis,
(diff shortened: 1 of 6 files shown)


compiler/mir/mirpasses.nim
```diff
@@ -455,6 +488,16 @@ proc applyPasses*(body: var MirBody, prc: PSym, env: var MirEnv,
       # requires the extraction for cstring literals
       extractStringLiterals(body.code, env, c)
 
+  # instrument the body with profiler calls after all lowerings, but before
+  # optimization
+  if (sfPure notin prc.flags) and (optProfiler in prc.options) and
+     (target in {targetC, targetJs}):
+    # XXX: not enabled for the VM because compile-time and run-time cannot
```
saem: Could have the compiler pass system temporarily remove the flag from options and then restore it prior to the code generation pass.

tests/compilerfeatures/tprofiler.nim
```diff
@@ -0,0 +1,67 @@
...
+  traces[numTraces] = st
+  inc numTraces
+  enabled = true # re-enable
+
+{.pop.}
+
+# nothing will happen before the hook is set
+profilerHook = profileCallback
+# the "profiling requested" callback guards whether to invoke the profile
+# callback is invoked
```
saem:
```suggestion
# callback
```

saem (approved):
Minor typo you may want to fix, otherwise the suggestion for limiting the profile during compile time is very soft. You might have already thought of it yourself.

compiler/mir/mirpasses.nim
zerbina: Hm, no, I haven't considered doing this, but it would work, and the implementation cost is small. I'll implement it.

zerbina:
/merge

Aaand another try. The first failure was due to the flaky `tfragment_alloc.nim` test, the second one due to an unusual MacOS failure where xcodebuild could not be located.

PR #1215, opened 2024-03-01
make the CGIR goto-based and fix exception-handling bugs

zerbina:
## Summary

* extend the MIR's design with goto-based control-flow primitives
* implement the new control-flow primitives for the CGIR
* make the C code generator use the new control-flow primitives
* (C backend) fix multiple bugs with `getCurrentException` returning
  the wrong exception
* (C backend) fix exceptions being leaked when aborting a `finally`
  section

### Details

The main goals of replacing the higher-level control-flow primitives in
the MIR with lower-level, goto-based ones are to:
* be able to express more complex control-flow
* make control-flow-related transformations easier and/or possible
* reduce the cost of computing the data-flow graph
* make evolving the language and MIR easier

A guiding principle was that the design should allow for enough freedom
to generate efficient code with all three code generators (which differ
significantly in control-flow-related capabilities).

To be able to implement this incrementally, as a first step, the new
control-flow primitives are only added to the CGIR, with only `cgen`
actually supporting them. So that `jsgen` and `vmgen` continue to work
without change, the legacy primitives are kept, with `cgirgen_legacy`
(a full copy of the original `cgirgen`) continuing to produce the
legacy CGIR.

### MIR

An in-depth description of syntax and behaviour is added to
`docs/mir.rst`. The gist of the design is that:
* terminators of basic-blocks become explicit (an exception being `If`)
* jump targets are directly specified on jump/fork-like operations
  (e.g., `Goto`, `CheckedCall`, etc.)
* intercepted control-flow (via `finally` sections) is also specified
  on jump/fork-like operations
* `finally` sections are *not* required to be duplicated for each
  unique control-flow passing through them
* jump targets and join points are identified by *labels*, which are
  integers IDs unique within a body

As a consequence of this design, there's no more statement-in-statement
nesting.

### CGIR & Translation

The CGIR implements all the new control-flow primitives from the MIR,
with the only difference being that the CGIR's `ContinueStmt` doesn't
track the possible exits of a `Finally` section (it's information
irrelevant to the code generators).

`cgirgen` handles translation of the old-style primitives still used by
the MIR to the new-style ones. While not strictly needed at the moment,
translation is disabled of intra-procedure unreachable code is
disabled, meaning that unreachable code is effectively dropped. This is
so that `Finally` sections that never exit normally (e.g.,
`finally: return`) can be ensured to work.

### C code generation

The CGIR is first translated into a small, low-level, C-specific IR
(all implemented in `ccgflow`). Here, the focus is on figuring out what
C code to produce for `Finally` sections; unnecessary gotos are also
eliminated.

#### Finally Handling

The previous strategy for `finally` was to:
* emit an error-state considering version of the `finally` at the end
  of the `try`
* duplicate the body of all enclosing `finally`s at `break`s and
  `return`s

Now, only a single version of the `finally` section is ever emitted. If
the `finally` section has more than one possible exit, a run-time
dispatcher is used, like so:
```c
{
  NI32 Target0_;
  L1_1_:
  Target0_ = 0;
  goto L1_;
  L1_2_:
  Target0_ = 1;
  L1_: // never jumped to from the outside
  NI32 oldNimErr0_ = *nimErr_; *nimErr_ = NIM_FALSE; // temporarily disable error mode
  // ... body of the finally ...
  *nimErr_ = oldNimErr0_;
  switch (Target0_) {
  case 0: goto L2_;
  case 1: goto L3_;
  }
}
```

Depending on the context and complexity of the code in-between, an
optimizing C compiler is able to eliminate the dispatcher and
`Target0_` assignment by directly inlining the body at a `goto L1_1_;`.

For the common case of a `Finally` section having two exits - one for
exceptional control-flow cases, and one for the -, a
`if (NIM_UNLIKELY(*nimErr_)) goto <error_label>; goto <non_error>` is
used instead of a full dispatcher. This mirrors how the exit of a
`finally` section previously looked like.

The idea behind not duplicating `Finally` sections is that it reduces
the amount of C code the compiler has to output. Nonetheless, the IR
from `ccgflow` is flexible enough to use the duplication strategy,
should this be needed again.

#### Exception Handling

Whether control-flow crosses the boundary of a (i.e., leaves it) is
encoded in the MIR/CGIR via the `Leave` item in target lists. `ccgflow`
uses this information to make sure that leaving a `Finally` via
unstructured control-flow properly aborts the active exception (if any)
and that leaving an `Except` properly pops the handler.   

#### Other

The IR from `ccgflow` then drives the rest of C code generation. Since
there's no more statement-in-statement nesting, recursion reduces
significantly.

A side-effect of the flatter representation is that less C scopes are
used (only `if`, loops, `finally`, and `except` use C scopes), which
seems to harm C compiler optimization.

Support for the legacy control-flow primitives is fully removed.

### Exception Handling Runtime

The runtime part of exception handling for the C target is partially
overhauled. For compatibility with the `csources` compiler, the new
version is guarded behind the `nimskullNewExceptionRt` condsym.

A distinction is now made between an *in-flight* exception and the
*current* exception. Previously, both were treated as being one and the
same.

An exception becomes *in-flight* once it is raised (`raiseException2`),
and stops being in-flight when it is caught (`nimCatchException`).
While in-flight, an exception can be aborted (`nimAbortException`), by
breaking out of an intercepting `finally`. On being caught by a
handler, the exception is associated with the handler and the handler
pushed to the handler stack. When control-flow leaves the handler
(`nimLeaveExcept`), the handler is popped from the stack.

The new runtime fixes the `currentException` not being updated when
breaking out of a `finally` or raising from within an `except` handler.
In addition, an exception caught by a handler can now be raised again
within the handler (e.g., `except CatchableError as e: raise e`)
without reference cycles being introduced.

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -15,7 +15,8 @@ import
   compiler/backend/[
     cgmeth,
     cgir,
-    cgirgen
+    cgirgen,
+    cgirgen_legacy
   ],
   compiler/front/[
     msgs,
@@ -368,6 +369,13 @@ proc generateIR*(graph: ModuleGraph, idgen: IdGenerator, env: MirEnv,
   ## Translates the MIR code provided by `code` into ``CgNode`` IR and,
   ## if enabled, echoes the result.
   result = cgirgen.generateIR(graph, idgen, env, owner, body)
+  echoOutput(graph.config, owner, result)
+
+proc generateIRLegacy*(graph: ModuleGraph, idgen: IdGenerator, env: MirEnv,
+                 owner: PSym, body: sink MirBody): Body =
+  ## Translates the MIR code provided by `code` into legacy ``CgNode`` IR and,
+  ## if enabled, echoes the result.
+  result = cgirgen_legacy.generateIR(graph, idgen, env, owner, body)
   echoOutput(graph.config, owner, result)
 
 # ------- handling of lifted globals ---------
--- a/compiler/backend/ccgcalls.nim
+++ b/compiler/backend/ccgcalls.nim
@@ -48,20 +48,20 @@ proc isHarmlessStore(p: BProc; canRaise: bool; d: TLoc): bool =
   else:
     result = false
 
-proc exitCall(p: BProc, callee: CgNode, canRaise: bool) =
+proc exitCall(p: BProc, call: CgNode) =
   ## Emits the exceptional control-flow related post-call logic.
-  if p.config.exc == excGoto:
+  if call.kind == cnkCheckedCall:
     if nimErrorFlagDisabled in p.flags:
-      if callee.kind == cnkProc and sfNoReturn in p.env[callee.prc].flags and
-         canRaiseConservative(p.env, callee):
+      if call[0].kind == cnkProc and sfNoReturn in p.env[call[0].prc].flags and
+         canRaiseConservative(p.env, call[0]):
         # when using goto-exceptions, noreturn doesn't map to "doesn't return"
         # at the C-level. In order to still support dispatching to wrapper
         # procedures around ``raise`` from inside ``.compilerprocs``, we emit
         # an exit after the call
         p.flags.incl beforeRetNeeded
         lineF(p, cpsStmts, "goto BeforeRet_;$n", [])
-    elif canRaise:
-      raiseExit(p)
+    else:
+      raiseExit(p, call[^1])
 
 proc fixupCall(p: BProc, le, ri: CgNode, d: var TLoc,
                callee, params: Rope) =
@@ -86,7 +86,7 @@ proc fixupCall(p: BProc, le, ri: CgNode, d: var TLoc,
         pl.add(addrLoc(p.config, d))
         pl.add(~");$n")
         line(p, cpsStmts, pl)
-        exitCall(p, ri[0], canRaise)
+        exitCall(p, ri)
     else:
       pl.add(~")")
       if isHarmlessStore(p, canRaise, d):
@@ -96,7 +96,7 @@ proc fixupCall(p: BProc, le, ri: CgNode, d: var TLoc,
         initLoc(list, locCall, d.lode, OnUnknown)
         list.r = pl
         genAssignment(p, d, list)
-        exitCall(p, ri[0], canRaise)
+        exitCall(p, ri)
       else:
         var tmp: TLoc
         getTemp(p, typ[0], tmp)
@@ -104,12 +104,12 @@ proc fixupCall(p: BProc, le, ri: CgNode, d: var TLoc,
         initLoc(list, locCall, d.lode, OnUnknown)
         list.r = pl
         genAssignment(p, tmp, list)
-        exitCall(p, ri[0], canRaise)
+        exitCall(p, ri)
         genAssignment(p, d, tmp)
   else:
     pl.add(~");$n")
     line(p, cpsStmts, pl)
-    exitCall(p, ri[0], canRaise)
+    exitCall(p, ri)
 
 proc reifiedOpenArray(p: BProc, n: CgNode): bool {.inline.} =
   # all non-parameter openArrays are reified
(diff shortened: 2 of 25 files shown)


zerbina:
There's a pre-existing bug with exception handling that still persists:
```nim
try:
  try:
    raise CatchableError.newException("a")
  finally:
    var e = CatchableError.newException("b")
    try:
      try:
        raise e
      finally:
        try:
          raise e 
          # `activeException.up` now points to `activeException`; the stack
          # of in-flight exceptions is corrupted
        except:
          discard
    except:
      discard
    # both the current and active exception are now nil, but error mode is still
    # active (correctly so). The next exception handler is going to crash

except CatchableError as e:
  doAssert e.msg == "a" 
```
This snippet crashes (previously the wrong exception reached the final handler).

Personally, I think the best solution here is to panic when an exception that's already in-flight is raised again, although that's something better implemented in a separate PR.

saem (approved):
Did a quick skim and then went through the commits. Looks good, really excited to see the remaining backends come over.

disruptek:
This branch exhibits some new compiler crashes in the cps tests when run without `panics:on`, though I haven't looked too closely and it's possible that cps shares some blame here.  Nonetheless, the improvements are not fully bug-compatible with tip. :shrug:

zerbina:
@disruptek Thanks for testing the branch with CPS. I was able to isolate and fix the problem.

---

The problem was elimination of unreachable being too aggressive giving the current. Take the following MIR:
```
scope:
  try:
    block L0:
      if cond:
        scope:
          return
      scope:
        return
    def x = construct (...)
    if cond:
      scope:
        return
  finally:
    =destroy(name x)
```
which is currently produced for:
```nim
if cond:
  return
else:
  return

var x = SomeType(...)
if cond:
  return
```

The first two returns are part of a *scoped* context (they're part of a `scope` block and there's no `if`, `repeat`, `try`, etc. in-between), and thus to-CGIR translation of MIR statements following them is disabled. Translation is thus inactive at the end of the `L0` block, meaning that everything up to the `finally` was discarded.

In theory, this is correct, but since the `def x` wasn't translated, information about the local is missing when `=destroy(name x)` (which is reachable) is translated. For the time being, translation is forcefully enabled again at every non-`except`, non-`finally` join point. Once the MIR itself uses the goto-based control-flow representation, this workaround becomes obsolete.

alaviss:
> Personally, I think the best solution here is to panic when an exception that's already in-flight is raised again, although that's something better implemented in a separate PR.

We could use an alive count for this, if the current exception is the same as in-flight head, then just increase the alive count so we could pop it once we leave the handler that raised it.

zerbina:
> We could use an alive count for this, if the current exception is the same as in-flight head, then just increase the alive count so we could pop it once we leave the handler that raised it.

Hm, I'm not sure I understand of how that would help. Since the in-flight stack is singly-linked list, pushing an exception to the stack that's already part of the stack corrupts the stack.

alaviss:
What I meant is something like this

```nim
proc pushException(e: Exception) =
  if inflight == e:
    e.extraRef.inc
  else:
    e.up = inflight
    inflight = e

proc popException() =
  if e.extraRef > 0:
    dec e.extraRef
  else:
    inflight = inflight.up
```

zerbina:
Ah, I see what you mean, thanks for the example.

Unfortunately, this would only work for the narrow case where the in-flight exception at the head of the stack is raised again. Consider:
```nim
var e1, e2: CatchableError = ...
try:
  raise e1
finally:
  try:
    raise e2
  finally:
    try:
      raise e1
    finally:
      try:
        raise e2
```
This would still fail.

Even incrementing the `extraRef` counter when the exception is not the in-flight head wouldn't help: the meaning of the stack would be ambiguous. `e2 -> e1 -> e2 -> e1` and `e2 -> e1 -> e1 -> e2`, and `e2 -> e2 -> e1 -> e1` couldn't be distinguished from each other.

alaviss:
True that, maybe we should just give up on linked list and use a seq :p

Or maybe a linked list of arrays so we can have a guaranteed stack count then allocate extra as needed

```nim
type
  ExceptionStack = ptr object
    data: array[<a sensible number>, ref Exception]
    len: int
    prev: ExceptionStack
```

zerbina:
/merge

I think the issue with corrupting the in-flight exception stack should be fixed separately. It's pre-existing, so no regression is introduced.

PR #1216, opened 2024-03-02
Add EMSCRIPTEN_KEEPALIVE to N_LIB_EXPORT if emscripten

theSherwood:
## Summary
* `N_LIB_EXPORT` now expands to `__attribute__((used, visibility("default")))` when `__EMSCRIPTEN__` is defined.
* Compiling to wasm with emscripten won't export symbols without `__attribute__((used))`.

---

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/lib/nimbase.h
+++ b/lib/nimbase.h
@@ -188,8 +188,6 @@ __EMSCRIPTEN__
 
 #  ifdef __cplusplus
 #    define N_LIB_EXPORT  NIM_EXTERNC __declspec(dllexport)
-#  elif __EMSCRIPTEN__
-#    define N_LIB_EXPORT  EMSCRIPTEN_KEEPALIVE __declspec(dllexport)
 #  else
 #    define N_LIB_EXPORT  NIM_EXTERNC __declspec(dllexport)
 #  endif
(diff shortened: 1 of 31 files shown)


lib/nimbase.h
```diff
@@ -188,6 +188,8 @@ __AVR__
 
 #  ifdef __cplusplus
 #    define N_LIB_EXPORT  NIM_EXTERNC __declspec(dllexport)
+#  elif __EMSCRIPTEN__
+#    define N_LIB_EXPORT  EMSCRIPTEN_KEEPALIVE __declspec(dllexport)
```
zerbina:
```suggestion
```
The defines here are only meant for when targeting native Windows.

lib/nimbase.h
```diff
@@ -220,8 +222,13 @@ __AVR__
 #    define N_FASTCALL_PTR(rettype, name) rettype (*name)
 #    define N_SAFECALL_PTR(rettype, name) rettype (*name)
 #  endif
-#  define N_LIB_EXPORT NIM_EXTERNC __attribute__((visibility("default")))
-#  define N_LIB_EXPORT_VAR  __attribute__((visibility("default")))
+#  ifdef __EMSCRIPTEN__
+#    define N_LIB_EXPORT  EMSCRIPTEN_KEEPALIVE __attribute__((visibility("default")))
+#    define N_LIB_EXPORT_VAR  EMSCRIPTEN_KEEPALIVE __attribute__((visibility("default")))
```
zerbina: Since the macro comes from the `emscripten.h` header, the header would need to included here, but I'm not sure whether that's a good idea.

Personally, I'd suggest replacing `EMSCRIPTEN_KEEPALIVE` with what the macro currently expands to:

```suggestion
#    define N_LIB_EXPORT __attribute__((used, visibility("default")))
#    define N_LIB_EXPORT_VAR __attribute__((used, visibility("default")))
```

Of course, this would stop working should emscripten exported C functions start to work differently.

zerbina (changes requested):
Thank you for the PR!

The define within the Windows-related section doesn't need to change, and something needs to be done about the `emscripten.h` header dependency required for the `EMSCRIPTEN_KEEPALIVE` macro (I left a suggestion).

lib/nimbase.h
theSherwood: Okay. Sounds good

lib/nimbase.h
theSherwood: Sounds good to me

theSherwood:
Closing in favor of https://github.com/nim-works/nimskull/pull/1218 because I merged from the wrong branch.

PR #1221, opened 2024-03-05
Remove nkStmtListType and nkBlockType nodes

Clyybber:
## Summary
* Remove  `nkStmtListType`  and  `nkBlockType`  node kinds and replace
the remaining usages of  `nkStmtListType`  with  `nkStmtListExpr` 

## Details
*  `not nil`  is implemented as a  `TypeTrait`  magic and no longer
reaches mirgen
* A workaround for untyped AST in sempass2 has been added to the check
introduced in cef64c469f167d4eeb3edda85c78040ccc95bfe9 and accordingly a 
`knownIssue`  testcase for the underlying issue


lib/system.nim
```diff
@@ -3043,3 +3043,7 @@ when defined(nimDebugUtils):
     {.define(nimCompilerDebug).}
     n
     {.undef(nimCompilerDebug).}
+
+# Overloads for not nil
+proc `not`*(a: typedesc[ref], b: typeof(nil)): typedesc {.magic: "Not".}
```
saem: Should we further constrain the parameter `b` with an AST constraint of `{nnkNilLit}`?

compiler/front/condsyms.nim
```diff
@@ -75,3 +75,4 @@ proc initDefines*(symbols: StringTableRef) =
   defineSymbol("nimskullNoMagicNewAssign")
   defineSymbol("nimskullNoFloat128")
   defineSymbol("nimskullNewExceptionRt")
+  defineSymbol("nimskullHasNkStmtListTypeAndNkBlockTypeRemoved")
```
saem:
```suggestion
  defineSymbol("nimskullRemovedNkStmtListTypeAndNkBlockType")
```

Slightly shorter of you prefer, still reads well, IMO

compiler/sem/sempass2.nim
```diff
@@ -1313,8 +1313,11 @@ proc track(tracked: PEffects, n: PNode) =
       if iterCall[1].typ != nil and
          iterCall[1].typ.skipTypes(abstractVar).kind notin {tyVarargs, tyOpenArray}:
         createTypeBoundOps(tracked, iterCall[1].typ, iterCall[1].info)
-    
+
     if tracked.owner.kind != skMacro and iterCall.kind in nkCallKinds and
+       iterCall[0].typ != nil and # XXX: untyped nkForStmt can reach here
```
saem: Maybe adding a known issue test for the bug surfaced by `typetraits.elementType` will help us get to the bottom of things. As we suspect this issue will show up again in different ways.

saem (commented):
A few things to consider in comments, but otherwise looks good to me.

zerbina (commented):
I like the solution with the magic routine, ~it also fixes the `not nil` suffix not being usable in generic bodies, I think~ (**edit**: that already works. I misremembered, apologies).

However, the type expression being evaluated in `semfold` is too late: the type of the expression should already be computed in semantic analysis. My suggestion would be to use the `TypeTrait` (in line with the existing `not` type operators) and then compute the type in `semTypeTrait`. `semfold` would only need to turn the expression into an `nkType` node then.

compiler/front/condsyms.nim
Clyybber: Changed it to `nimskullNoNkStmtListTypeAndNkBlockType`

lib/system.nim
Clyybber: Hmm, that would be possible I think, but then system.nim would depend on an experimental feature, so I'd rather keep the signature as simple as possible.

saem:
FYI: PR body needs an update given the revised implementation.

compiler/sem/semmagic.nim
```diff
@@ -175,7 +175,18 @@ proc evalTypeTrait(c: PContext; traitCall: PNode, operand: PType, context: PSym)
...
-    return typeWithSonsResult(tyNot, @[operand])
+    if traitCall.len == 3:
+      c.config.internalAssert traitCall[2].kind == nkNilLit
+
+      var notNilTyp = copyType(operand, nextTypeId c.idgen, operand.owner)
+      copyTypeProps(c.graph, c.idgen.module, notNilTyp, operand)
+      notNilTyp.flags.incl(tfNotNil)
+
+      return makeTypeDesc(c, notNilTyp).toNode(traitCall.info)
+
```
zerbina:
```suggestion
      # the operand is not generic anymore, let ``semTypeNode`` produce a
      # type with the not-nil modifier applied
      return makeTypeDesc(c, semTypeNode(c, traitCall, nil)).toNode(traitCall.info)
```
A soft suggestion based on our discussion on Matrix.

lib/system.nim
```diff
@@ -3043,3 +3043,6 @@ when defined(nimDebugUtils):
     {.define(nimCompilerDebug).}
     n
     {.undef(nimCompilerDebug).}
+
+# Overload for not nil
+proc `not`*(a: typedesc[ref], b: typeof(nil)): typedesc {.magic: "TypeTrait".}
```
zerbina:
```suggestion
proc `not`*(a: typedesc[ptr or ref], b: typeof(nil)): typedesc {.magic: "TypeTrait".}
```
So that `T not nil` also works with pointer types. Also, perhaps it'd make sense to move the procedure together with the other type constructor procedures (`not`, `and`, and `or`)?

compiler/sem/semtypes.nim
```diff
@@ -2124,7 +2112,9 @@ proc semTypeNode(c: PContext, n: PNode, prev: PType): PType =
             # I suggest revisiting this once the language decides on whether
             # `not nil` should be the default. We can then map nilable refs
             # to other types such as `Option[T]`.
-            result = makeTypeFromExpr(c, newTree(nkStmtListType, n.copyTree))
+            var copyN = n.copyTree
+            copyN[1].typ = makeTypeDesc(c, result)
```
zerbina, line 2118:
```suggestion
            # the `not` procedure expects a typedesc, so make sure the
            # operand is a `tyTypeDesc`  
            copyN[1].typ = makeTypeDesc(c, result)
```
A soft suggestion that would improve clarity a bit, I think. Feel free to ignore.

compiler/sem/sempass2.nim
Clyybber: Done :)

lib/system.nim
Clyybber: The other type constructor procedures appear before the definition of `typeof` which this one uses so I'd have to move those down or `typeof` up.

saem:
Looks like a timeout, so just giving CI a wee kick

lib/system.nim
zerbina: Oh, I hadn't considered that. The `system` module really needs to be cleaned up and reordered at some point, but I think it's fine if `not` stays where it is for now.

zerbina: Did you encounter problems with using `ptr or ref`?

lib/system.nim
```diff
@@ -3043,3 +3043,6 @@ when defined(nimDebugUtils):
     {.define(nimCompilerDebug).}
     n
     {.undef(nimCompilerDebug).}
+
+proc `not`*(a: typedesc[ref], b: typeof(nil)): typedesc {.magic: "TypeTrait", noSideEffect.}
```
zerbina:
```suggestion
proc `not`*[T: ref or ptr](a: typedesc[T], b: typeof(nil)): typedesc {.magic: "TypeTrait", noSideEffect.}
```
This should work. `liftParamType` doesn't recurse into `tyTypeDesc` (although it should), so my original `typedesc[ref or ptr]` suggestion won't work, sorry.

Clyybber: Thanks!

zerbina (approved):
The changes look good to me. Thank you for working through all the review comment!

PR #1224, opened 2024-03-07
std/asyncjs: value passed to `catch` is always `Error`

zerbina:
## Summary

* NimSkull exceptions leaving `.async` procedure are automatically
  turned into JavaScript `Error`s
* errors retrieved by `catch` that were originally NimSkull exceptions
  always use only the original exception's message (no "uncaught
  exception" prefix)

## Details

* the `.async` macro wraps the body in a `try`/`except` that turns
  catchable errors into `Error`s
* depending on whether a procedure is called within a `try` (at
  run-time), `raise` either raises a NimSkull exception or JavaScript
  error
* the `tasyncjs_pragma` test is updated to not rely on the transformed
  body being a statement list

### DIFF
--- a/lib/js/asyncjs.nim
+++ b/lib/js/asyncjs.nim
@@ -73,6 +73,8 @@ type
   PromiseJs* {.importjs: "Promise".} = ref object
   ## A JavaScript Promise.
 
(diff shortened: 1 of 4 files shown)


saem (approved):

saem:
/merge

The failure looks to be the intermitent one, going to queue this for mering.

PR #1225, opened 2024-03-07
vmgen: use the new-style CGIR

zerbina:
## Summary

Use the goto-using version of the CGIR for VM code generation,
removing another usage of the legacy variant. This also fixes the
current exception (`getCurrentException`) being incorrect after
raising from within an `except`, on the VM backend.

### VM changes

To accomodate the code generator, two small changes are made to the
instruction interface:
* instead of the instruction position from where to extract the
  register, the register index is directly encoded in `ehoLeave` EH
  instructions
* `opcBranch` must now be followed by a `opcTJmp` (jump if true)
  instead of a `opcFJmp` (jump if false)

### Code generator changes

The CGIR's control-flow representation translates more or less directly
to the VM's control-flow instructions, so code generation for it is
straightforward.

* the new-style CGIR is enabled in both `vmbackend` and `vmjit`
* no pre-pass over the CGIR - like is used by the C code generator - is
  needed
* registers for locals are freed at the end of structured blocks
  (`cnkIf`, `cnkFinally`, `cnkExcept`, and loops)
* EH instructions are emitted as the jump actions are processed
* jump actions for exceptional exits (e.g., `cnkCheckedCall`,
  `cnkRaise`, etc.) are usually the same for regions of code, so a
  caching mechanism is used to prevent repeated EH instructions
  sequences in that case. (not required for correctness)

### DIFF
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -3239,7 +3239,7 @@ proc genExpr*(c: var TCtx; body: sink Body): Result[int, VmGenDiag] =
   var d: TDest = -1
   try:
     let eh = genSetEh(c, n.info)
-    if n.kind == cnkStmtListExpr:
+    if n.kind == cnkStmtList:
       # special case the expression here so that ``gen`` doesn't have to
       for i in 0..<n.len-1:
         c.gen(n[i])
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -1032,6 +1032,15 @@ proc genEhCode(c: var TCtx, n: CgNode) =
 
 proc genExcept(c: var TCtx, n: CgNode) =
   ## Emits the EH code for a ``cnkExcept``.
+
+  # simple but high-impact optimization: if the last EH exit we need to patch
+  # is the preceding EH instruction, eliminate the instruction (it'd just be
+  # a single instruction jump)
+  if c.prc.ehExits.len > 0 and
+     c.prc.ehExits[^1] == (n[0].label, c.ehCode.high.uint32):
+    c.ehCode.setLen(c.ehCode.len - 1)
+    c.prc.ehExits.setLen(c.prc.ehExits.len - 1)
+
   # patch all EH instructions targeting the handler:
   for it in take(c.prc.ehExits, n[0].label):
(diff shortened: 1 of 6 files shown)


compiler/vm/vmgen.nim
```diff
@@ -127,15 +127,25 @@ type
...
     bkBlock   ## labeled block
-    bkTry     ## the ``try`` and ``except`` clause of a ``finally``-having
-              ## try statement
     bkExcept  ## ``except`` clause
     bkFinally ## ``finally`` clause
 
+  BlockInfo = object
+    label: BlockId
+    oldRegisterCount: int
+      ## upper bound of allocated registers at the beginning of the block
```
Clyybber:
```suggestion
    oldRegisterCount: int
      ## upper bound of allocated registers at the beginning of the block
    label: BlockId
```
Really inconsequential, but due to alignment one could save 8 bytes (on 64bit archs) by swapping these two fields.

compiler/vm/vmgen.nim
```diff
@@ -438,14 +448,66 @@ proc patchSetEh(c: var TCtx, p: TPosition) =
...
+
+    c.ehTable.add (pos, uint32(c.ehCode.len - 1))
+  of cnkTargetList:
+    if n.len == 1 and n[0].kind == cnkResume:
+      # if there's nothing responding to the exception within the current
+      # procedure, no EH code needs to be associated with the instruction
+      return
+
+    if c.prc.lastPath == nil or (let shared = comparePaths(n, c.prc.lastPath);
+       shared < n.len):
```
Clyybber:
```suggestion
    if c.prc.lastPath == nil or comparePaths(n, c.prc.lastPath) < n.len:
```
since `shared` isn't used later. Disregard this if you wrote it this way for clarity :D

zerbina: Huh, looks like I missed removing `shared` (it was used at one point), thanks.

compiler/vm/vmgen.nim
zerbina: It doesn't hurt to swap them. But yeah, one would need structured block-like nesting in the millions for this to be noticeable :D

saem (approved):

Issue #720, opened 2023-05-24
Crash when compiling a-b.nim


iacore:
The compiler crashes with trying to compile a file with `-` in it.

```
❯ touch sol-nimskull.nim

❯ nimskull c sol-nimskull.nim
.......................................................
Error: invalid module name: '<nil tree>' [rsemInvalidModuleName]
modules.nim(114, 16) compiler msg instantiated here [MsgOrigin]
No stack traceback available
To create a stacktrace, rerun compilation with './koch temp $1 <file>'
```

disruptek:
```
Error: invalid module name: '<nil tree>' [rsemInvalidModuleName]
modules.nim(107, 16) compiler msg instantiated here [MsgOrigin]
```
```
Nimskull Compiler Version 0.1.0-dev.21252 [linux: amd64]

Source hash: 1854db10bf8b35214222d69dbf90c21d7e8caf5f
Source date: 2024-03-07

active boot switches: -d:release -d:danger⏎
```

zevv:
So, what's the required behavior here. Just removing the `isNimIdentifier()` call makes things compile and work, although there's no way to explicitly refer to the module name space. Do we want to fail with a sensible error message instead of a crash, or should just be made to work, possibly by dropping the `-` or` replacing the `-` with `_`?

saem:
At the very least a sensible error message, instead of a crash.

---

In the case of getting it to work, I'm not really sure how to balance:

The funky ways underscores can/could be inserted/omitted in identifies, and a potentially mandatory unskippable underscore.

iacore:
there is also the Rust way of turning a-b into a_b

zerbina:
Fixed by #1354.

PR #1228, opened 2024-03-08
Fix error message `<nil tree>` when importing module with illegal name

zevv:
## Summary

This fixes the error message:

``` 
  Error: invalid module name: '<nil tree>'
```

when trying to import a module with invalid name (for example, a module containing a hyphen in the name).

This fixes issue #720.

## Details

Alternatively, when just removing the `isNimIdentifier()` check for the module name alltogther, we could just allow module names with hyphens to work. I'm not sure if this will give unintended interactions with specific platform file systems though.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1633,7 +1633,7 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       result = "using '.' instead of '/' in import paths is deprecated"
 
     of rsemInvalidModuleName:
-      result = "invalid module name: '$1'" % r.ast.render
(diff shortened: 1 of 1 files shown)


saem (changes requested):
Could you add a regression test?

saem:
In `compiler/modules/modules.nim`, the `localReport` can be switched to the variant that adds line info, something like `localReport(graph.config, result.info, reportSym(rsemInvalidModuleName, result))`. It's not ideal, but at least we'll get some line info.

Alternatively, it might work to set the testament spec to `file: ""`

zerbina:
Superseded by #1354.

PR #1229, opened 2024-03-08
Remove nkNone and replace it with nkError

Clyybber:
## Summary
* Remove `nkNone` and replace it with `nkError`
* Move some  `nkError`  branches to the start of the their case
statements

## Details
`nkNone`  could be replaced by either  `nkEmpty`  or  `nkError` . I've
decided for  `nkError`  here, as  `nkNone`  nodes already signified
errors, albeit compiler bugs.
That means that an uninitialized node that would previously be of kind 
`nkNone`  is now of kind  `nkError`  and will cause a crash if it's
uninitialized  `diag`  field is accessed.
This is IMO preferable to  silently bubbling up uninitialized nodes
(that signify a compiler bug), but should not actually matter in
practice, as I don't think there are any recent or open known cases
where  `nkNone`  nodes were actually produced.

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -483,7 +483,7 @@ template copyNodeImpl(dst, src, processSonsStmt) =
   of nkSym: dst.sym = src.sym
   of nkIdent: dst.ident = src.ident
   of nkStrLiterals: dst.strVal = src.strVal
-  of nkEmpty, nkNone, nkNilLit, nkType, nkCommentStmt: discard "no children"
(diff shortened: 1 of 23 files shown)


zerbina (approved):
Looks good to me, and I agree that using `nkError` in place of `nkNone` is better than using `nkEmpty`.

A minor nit-pick, but I think it'd be a good idea to mention that some `nkError` branches were moved to the start of the their case statements.

saem (approved):
LGTM

PR #1230, opened 2024-03-08
jsgen: use the new-style CGIR

zerbina:
## Summary

* use the goto-using CGIR for JavaScript code generation
* as a side-effect, multiple bugs with tracking of the current
  exception are fixed 
* exceptions thrown from `asm` or `.emit` code cannot be expected to be
  caught anymore (this was already undefined behaviour)

## Details

The CGIR for a body is first passed to the new `jsflow` module, where
the shape of the necessary JavaScript is computed, with the focus on
the control-flow constructs (`if`, `switch`, `try`, etc.). Constructs
are translated as follows:
* `cnkGotoStmt` and `cnkJoinStmt` are translated to `break` and labeled
  blocks (JavaScript doesn't support goto directly)
* `cnkCaseStmt` dispatchers are translated to `switch`/`case`
  statements
* `cnkFinally` and `cnkExcept` are translated to `try`/`finally` and
  `try`/`catch`, respectively

It's first computed which statements `try`s and labeled blocks *must*
enclose. Afterwards, their starts are reordered such that valid
JavaScript is formed.

Since `finally` and `catch` clauses can then intercept `break`s or
`throw`s that they shouldn't, `finally` and `catch` clauses are guarded
by a boolean flag. Immediately before an explicit (`break` and `throw`)
or implicit (procedure call) jump that would enter a clause it
shouldn't, the boolean locals for the problematic clauses are set to
false. Upon exit of a disabled clause, the local is set back to true.

#### Case Statements

Direct translation of `cnkCaseStmt`s into `switch`/`case` statement
where each branch is a `break` would result in very deep block nesting
for `case` statements with many branches. The resulting JavaScript code
can cause the V8 JavaScript engine to fail/crash when compiling the
code!

Therefore, the jump targets need to be inlined whenever possible.
`jsflow` computes the code regions eligible for inlining, and `jsgen`
handles the actual inlining.

#### Current Exception Tracking

When control-flow leaves an `except` handler, the `prevJSError` global
(whose content is returned by `getCurrentException`) needs to be reset
to the exception caught by the enclosing handler, something that was
previously missing for the most part.

The exception is restored:
* at the end of a block when the block is broken out of from within a
  `catch`
* at the start of a `finally` clause intercepting a `break` from within
  a `catch`
* at the start of a `finally` clause when it's entered due to an
  exception

If there's no enclosing `catch` within a procedure when the current
exception has to be restored, the value of `lastJSError` is captured at
procedure entry. Tracking of the current exception still doesn't
include in-flight exceptions, so `tfinally6.nim` still fails.

#### Unhandled Exception Reporting

Instead of incrementing, at run-time, a global counter (`excHandler`)
before entering a `try` and decrementing it again on exit, and then
reporting unhandled exceptions when throwing, the module-init
procedures (i.e., top-level code) are wrapped in a `try`/`catch` that
reports the escaping exceptions.

This is simpler to implement and should also have less run-time
overhead.

#### Throwing with `asm`

`asm` and `emit` statements are considered not to raise, and thus they
don't prevent `except` and `finally` sections from being eliminated by
unreachable code elimination.

The `tnativeexc` test is updated to not rely on undefined behaviour.

### Unreachable Code Elimination

Labels that mark the exit of case statements are now eliminated too, if
structured control-flow doesn't leave the case statement (i.e., each
branch ends in a noreturn-like statement).


zerbina:
`jsgen` now tries to inline the jump target of case dispatcher statements into the switch-case statement. This fixes node.js failing to compile the megatest (`thtmlparser.nim` is the offender here). As a side note, I've also tried to run the code in FireFox (SpiderMonkey engine), and there it works.

The heuristic that's currently in use is not correct however, and some case dispatcher shapes (none of which are possible at the moment) would cause incorrect code generation.

zerbina:
Okay, that should be it. `catch` and `finally` clauses for which it is necessary can now be disabled at run-time, resolving problems a mismatch between expected and actual control-flow. For example, the following MIR (CGIR at the moment):
```
if ...
  goto [L1, L2]
end
goto L2
finally L1:
  ...
  continue
L2:
...
```
translates to:
```js
L2: {
  try {
    if (...) {
      break L2;    
    }
    break L2;
  } finally {
    ...
  }
}
```
The second `break` would also visit the `finally`, even though the original goto explicitly states that it should not.

While CGIR such as the above is not possible at the moment, except handlers are not merged into one at the moment, so:
```nim
try:
  raise ...
except A as e:
  ...
except:
  ...
```
is currently translated into:
```js
try {
  try {
    throw ...
  } catch (Exception1_) {
    // rethrow if no type match
  }
} catch (Exception1_) {
  ...
}
```
where it has to be made ensure that exceptions raised from within the first exception handler are not handled by the outer `try`.

zerbina:
The changes do increase the complexity of the code generator quite a bit. I think at least part of it is due to the ad-hoc nature of the change: `jsgen` was not designed with this kind of translation in mind -- rewriting `jsgen` from scratch against the MIR *should* reduce some of the complexity. However, I think that the larger part of the reason is that going from low-level, goto-using code to code with breaks, blocks, and try/catch/finally has just some inherent complexity.

I experimented with different approaches, such as having `jsflow` emit a more elaborate IR (akin to `ccgflow`), or performing the switch-case body inlining in `jsflow`, but the solution I converged on ultimately felt like the least complex.

compiler/backend/jsgen.nim
```diff
@@ -1227,7 +1272,7 @@ proc genArgs(p: PProc, n: CgNode, r: var TCompRes; start=1) =
   assert(typ.len == typ.n.len)
   var emitted = start-1
 
-  for i in start..<n.len:
+  for i in start..<(1 + numArgs(n)):
```
```suggestion
  for i in start..numArgs(n):
```
since `<` is basically a `- 1`, I'm assuming this simplification works

compiler/backend/jsgen.nim
```diff
@@ -1247,11 +1292,11 @@ proc genArgs(p: PProc, n: CgNode, r: var TCompRes; start=1) =
 
 proc genOtherArg(p: PProc; n: CgNode; i: int; typ: PType;
                  generated: var int; r: var TCompRes) =
-  if i >= n.len:
+  if i >= numArgs(n) + 1:
```
```suggestion
  if i > numArgs(n):
```

I believe this simplification is still equivalent.

compiler/backend/jsgen.nim
```diff
@@ -1274,7 +1319,7 @@ proc genPatternCall(p: PProc; n: CgNode; pat: string; typ: PType;
     case pat[i]
     of '@':
       var generated = 0
-      for k in j..<n.len:
+      for k in j..<(1 + numArgs(n)):
```
```suggestion
      for k in j..numArgs(n):
```
same as before

compiler/backend/jsgen.nim
```diff
@@ -1341,7 +1386,7 @@ proc genEcho(p: PProc, n: CgNode, r: var TCompRes) =
   useMagic(p, "rawEcho")
   r.res.add("rawEcho(")
   # the first argument is a literal type that we don't need
-  for i in 2..<n.len:
+  for i in 2..<(1 + numArgs(n)):
```
```suggestion
  for i in 2..<numArgs(n):
```

compiler/backend/jsgen.nim
```diff
@@ -2087,15 +2126,23 @@ proc genProcBody(p: PProc, prc: PSym): Rope =
...
+    result.add(p.indentLine("var Exception0_ = lastJSError;$n" % []))
+
+  result.add(p.body)
   if prc.typ.callConv == ccSysCall:
     result = ("try {$n$1} catch (e) {$n" &
       " alert(\"Unhandled exception:\\n\" + e.message + \"\\n\"$n}") % [result]
+  elif sfModuleInit in prc.flags:
+    # report an unhandled exception when a |NimSkull| exception escapes
+    # module-level code
+    # XXX: this is common logic across all backends; it needs to be handled at
```
saem, line 2140: Ah fun! This also provides some clues as to how we might handle `defect` -> `exception` conversion for crossing package bounardies.

compiler/backend/jsflow.nim
```diff
@@ -0,0 +1,360 @@
...
+    finallys: PackedSet[BlockId]
+      ## labels that denote finally sections
+    inline: Table[BlockId, int]
+      ## maps all blocks that can be inlined into swith-case statements to
+      ## the blocks' 'end' item
+
+const
+  Terminators = {stkReturn, stkTerminator}
+
+func finalTarget*(n: CgNode): CgNode =
```
saem, line 62:
```suggestion
func finalTarget*(n: CgNode): CgNode =
  ## given a label or target list, retrieves the target
```

saem (approved):
I'm under the weather, so I don't think I was all that thorough in my review, but the conversion looks good.

compiler/backend/jsflow.nim
zerbina: Thanks, I missed that one.

Varriount:
Why does translation of switch/case cause nesting? Wouldn't something like 
```nim
case direction
of North:
  ...
of South:
  ...
of East:
  ...
of West:
  ...
```
be translated to (roughly)
```javascript
switch (direction) {
  case North:
    ...
    break;
  case South:
    ...
    break;
  case East
    ...
    break;
  case West
    ...
    break;
}
```

Or does the goto-based IR make that impossible?

compiler/backend/jsflow.nim
```diff
@@ -0,0 +1,360 @@
...
+      assert it[0].label in marker
+      struct(stkCatch, it[0].label)
+      if it.len > 1:
+        # not a catch-all handler; raising might continue
+        exit(it[^1], isError=true)
+    else:
+      discard "not relevant"
+
+  # the list of openings and closing produced by the first pass will in most
+  # cases not be valid JavaScript code. We have to "solve" therepresentation
```
Varriount:
```suggestion
  # cases not be valid JavaScript code. We have to "solve" the representation
```

compiler/backend/jsgen.nim
zerbina: I've replaced the `(1 + numArgs(n))` expressions with a call to the new `callLen` procedure, which does the same thing but should convey the intention better.

zerbina:
> Why does translation of switch/case cause nesting? Wouldn't something like
> 
> be translated to (roughly)

That's how the translation works now, yes. In the MIR (CGIR, at the moment), a `case` statement is an exhaustive dispatcher that works like this:
```nim
case value
of 1: goto L1
of 2: goto L2
of 3: goto L3
L1:
# branch 1
goto CaseExit
L2:
# branch 2
goto CaseExit
L3:
# branch 3
goto CaseExit
CaseExit:
```

Gotos are translated to `break` statement, so the naive translation (the one used prior to [this](https://github.com/nim-works/nimskull/pull/1230/commits/180bf94da2c4136ccfaa3062acfa9edf49768931) commit) results in the following JavaScript output:
```js
CaseExit: {
  L3: {
    L2: {
      L1: {
        switch (value) {
        case 0: break L1;
        case 1: break L2;
        case 2: break L3;
        }
      }
      ...
      break CaseExit;
    }
    ...
  }
  ...
}
```

compiler/backend/jsflow.nim
```diff
@@ -0,0 +1,361 @@
...
+type
+  StructKind* = enum
+    stkTry   ## start of a 'try' statement
+    stkBlock ## start of a labeled block
+
+    stkStructStart ## start of an 'if' or 'while'
+    stkCatch   ## a 'catch' clause of a 'try' statement
+    stkFinally ## a 'finally' clause of a 'try' statement
+
+    stkEnd     ## end of a catch, finally, or block
```
Clyybber: AFAICT this also marks the end of a `stkStructStart`?

Clyybber (approved):

compiler/backend/jsflow.nim
zerbina: Good catch, it does. I'll update the comment.

zerbina:
/merge

Thank you for the reviews, everyone!

PR #1233, opened 2024-03-12
Unify tree equivalence checking procedures

Clyybber:
## Summary

Unify tree equivalence checking procedures in the compiler by way of a
template that generates these procedures. This fixes differing comments
between forward declaration and implementation breaking the comparison.
This also fixes a bug where VM constant deduplication ignored node
literal differences due to comments. 

## Details

* The various tree equivalence checking procedures across the compiler
now all use a common template  `makeTreeEquivalenceProc` .
*  `nkCommentStmt`  nodes' equality is now **ignored** in 
`exprStructuralEquivalent`  and  `exprStructuralEquivalentStrictSym` ,
fixing an issue where differing comments could prevent matching forward
declarations, see the test added in 
`tests/lang_callable/generics/tforwardgeneric.nim` 
*  `nkCommentStmt`  nodes' equality is now **checked for** in 
`cmpNodeCnst`  fixing an issue where constants containing NimNodes were
incorrectly deduplicated, see the test added in 
`tests/lang_callable/macros/tmacros_various.nim` 
* Removed `trees.sameTree` due to it being dead-code
* Add comments explaining preexisting issues with  `nkFloatLiterals` 
and  `nkNimNodeLit`  equality checking


zerbina:
It's not a full review yet, but regarding `nkNimNodeLit`, it is already possible for it to store trees besides `nkSym`. This the case when a `NimNode` is part of aggregate types that cross the VM-to-compiler boundary (e.g., via `const C = (1, newLit("string"))`).

For the dead-code removal, could you split it out into a separate PR? That would make it easier to restore (parts of) the removed implementation, should they become needed/useful again in the future.

Clyybber:
> For the dead-code removal, could you split it out into a separate PR? That would make it easier to restore (parts of) the removed implementation, should they become needed/useful again in the future.

Undid the dead-code removal

> It's not a full review yet, but regarding nkNimNodeLit, it is already possible for it to store trees besides nkSym. This the case when a NimNode is part of aggregate types that cross the VM-to-compiler boundary (e.g., via const C = (1, newLit("string"))).

Ah nice, then I'll try to come up with a test case demonstrating the change in semantics, thanks!

saem:
A quick note about the PR body, referencing internal commits should be for reviewer notes, the body itself should describe what's going on/why.

Reviewing the change now in more detail.

compiler/ast/trees.nim
```diff
@@ -44,61 +44,64 @@ proc sameFloatIgnoreNan(a, b: BiggestFloat): bool {.inline.} =
...
-        for i in 0..<a.len:
-          if not exprStructuralEquivalent(a[i], b[i],
-                                          strictSymEquality): return
-        result = true
+template makeTreeEquivalenceProc*(
+  name, relaxedKindCheck, symCheck, floatCheck, typeCheck, commentCheck) {.dirty.} =
+  # Defines a tree equivalence checking procedure.
+  # This skeleton is shared between all recursive
+  # PNode equivalence checks in the compiler code base
+  # It might be possible to unify more of them with each other.
```
saem:
```suggestion
  ## Defines a tree equivalence checking procedure.
  ## This skeleton is shared between all recursive
  ## `PNode` equivalence checks in the compiler code base
  ## It might be possible to unify more of them with each other.
```

compiler/ast/trees.nim
```diff
@@ -44,61 +44,64 @@ proc sameFloatIgnoreNan(a, b: BiggestFloat): bool {.inline.} =
...
+      of nkIntLiterals:     result = a.intVal == b.intVal
+      of nkFloatLiterals:   result = floatCheck
+        # XXX: Using float equality, even if partially tamed through
+        # sameFloatIgnoreNan, causes inconsistencies due to it lacking
+        # the substition and reflexivity property.
+      of nkStrLiterals:     result = a.strVal == b.strVal
+      of nkType:            result = typeCheck
+      of nkCommentStmt:     result = commentCheck
+      # XXX: nkNimNodeLit should probably always be checked strictly.
+      # Currently this doesn't matter as only nkSym NimNode literals are created
```
saem:
```suggestion
      #      Currently this doesn't matter as only nkSym NimNode literals are created
```

Align with the above so it's clear that it's part of the same comment.

Also, if that's the case this should be enforced via an assertion.

compiler/ast/trees.nim
```diff
@@ -44,61 +44,64 @@ proc sameFloatIgnoreNan(a, b: BiggestFloat): bool {.inline.} =
...
+      case a.kind
+      of nkError:           unreachable()
+      of nkEmpty, nkNilLit: result = true
+      of nkSym:             result = symCheck
+      of nkIdent:           result = a.ident.id == b.ident.id
+      of nkIntLiterals:     result = a.intVal == b.intVal
+      of nkFloatLiterals:   result = floatCheck
+        # XXX: Using float equality, even if partially tamed through
+        # sameFloatIgnoreNan, causes inconsistencies due to it lacking
+        # the substition and reflexivity property.
```
saem:
```suggestion
        #      sameFloatIgnoreNan, causes inconsistencies due to it lacking
        #      the substition and reflexivity property.
```

align so it's clear that it's part of the same thing

compiler/ast/trees.nim
```diff
@@ -44,61 +44,64 @@ proc sameFloatIgnoreNan(a, b: BiggestFloat): bool {.inline.} =
...
-    of nkError:
-      unreachable()
-    of nkWithSons:
-      if a.len == b.len:
-        for i in 0..<a.len:
-          if not sameTree(a[i], b[i]): return
-        result = true
+makeTreeEquivalenceProc(exprStructuralEquivalent,
+  relaxedKindCheck = false,
+  symCheck =     a.sym.name.id == b.sym.name.id, # same symbol as string is enough
```
saem: If you're going to align the params, which I recommend, I'd align the `=`, as well. It's far easier to skim I find.

compiler/ast/types.nim
```diff
@@ -652,6 +652,7 @@ proc sameTypeOrNil*(a, b: PType, flags: TTypeCmpFlags = {}): bool =
     else: result = sameType(a, b, flags)
 
 proc equalParam(a, b: PSym): TParamsEquality =
+  # This is not commutative!
```
```suggestion
  ## Compares parameters `a` and `b` to indicated whether they're equal.
  ## Note that this operation is not commutative, so when comparing a
  ## forward declaration to an implementation, ensure `b` is the param from
  ## the forward declaration. 
```

The previous phrasing made it seem like a thing to be fixed rather than be mindful of. I made it part of the docs as it should be available as part of referencing it. Also, added further docs to help explain the overall purpose of this proc.

compiler/sem/procfind.nim
```diff
@@ -58,6 +58,7 @@ proc searchForProcAux(c: PContext, scope: PScope, fn: PSym): PSym =
         return
       of paramsIncompatible:
         localReport(c.config, fn.info, reportSym(rsemAmbiguousCall, fn))
+          # Should be rsemDefaultParamIsIncompatible
```
saem:
```suggestion
          # TODO: Should be rsemDefaultParamIsIncompatible
```

It shows up in searches and/or some plugins allow for browsing by todos.

compiler/ast/trees.nim
```diff
@@ -44,61 +44,64 @@ proc sameFloatIgnoreNan(a, b: BiggestFloat): bool {.inline.} =
...
+  name, relaxedKindCheck, symCheck, floatCheck, typeCheck, commentCheck) {.dirty.} =
+  # Defines a tree equivalence checking procedure.
+  # This skeleton is shared between all recursive
+  # PNode equivalence checks in the compiler code base
+  # It might be possible to unify more of them with each other.
+  proc name(a, b: PNode): bool =
+    result = false
+    if a == b:
+      result = true
+    elif a != nil and b != nil and (a.kind == b.kind or relaxedKindCheck):
```
saem, line 57:
```suggestion
    elif a.isNil or b.isNil:
      # only one is `nil`, so automatically fail
      result = false
    elif a.kind == b.kind or relaxedKindCheck:
```

I think it's a bit easier to read, between the removal of the negation checks and shorter conditionals. Since it's now also exhaustive for all cases, the initial `result = false` could be dropped.

saem (approved):
Few stylistic comments that I think would help readability.

I really like the unification of tree equivalence, I think it'll likely help eliminate a number of bugs.

compiler/ast/trees.nim
Clyybber: Removed in ea33c54054580cd75880b41b5cfa215e6e170ba6 as it's not actually the case (thanks to @zerbina for pointing that out).

compiler/ast/types.nim
Clyybber: Thanks, I rephrased it slightly to
```
  ## Returns wether parameters `a` and `b` are considered equal.
  ## Note that this operation is not commutative, so when comparing
  ## a forward declaration to an implementation, ensure `a` is the
  ## param from the forward declaration.
  ```
  in 6a6c0a37b51be0e87437e0329d8450da2468c0ce

File: tests/lang_callable/macros/tmacros_various.nim [hunk 8]
```diff
@@ -328,3 +330,16 @@ block: # bug #15118
 
   block:
     flop("b")
+
+block: # Ensure nkCommentStmt equality is not ignored when checking for const duplicates:
```
saem: Sorry for being daft, is there a duplicate constant value check, and here we're ensuring that it's not triggered?

Also is that an error or a warning?

compiler/ast/trees.nim
Clyybber: It's not actually exhaustive since there's no else branch after the kind check branch.
I slightly prefer less spread out conditionals since single terms are easier to apply boolean transformations/simplifactions to, although in this case there is nothing to be simplified anymore of course.
I don't feel strongly either way.

tests/lang_callable/macros/tmacros_various.nim
Clyybber: There's a duplicate constant value check in `vmgen.nim` for literals, `cmpNodeCnst` for NimNode literals, that deduplicates constants as vm code is generated. It used to not account for differing `nkCommentStmt`s so it counted the second "comment 2" as a duplicate of "comment 1" and incorrectly optimized it away to point at the "comment 1" node

saem: Ah, got it, some of the above explanation making it into the above comment might help, it's a bit arcane -- soft suggestion.

compiler/ast/types.nim
```diff
@@ -652,6 +652,10 @@ proc sameTypeOrNil*(a, b: PType, flags: TTypeCmpFlags = {}): bool =
     else: result = sameType(a, b, flags)
 
 proc equalParam(a, b: PSym): TParamsEquality =
+  ## Returns wether parameters `a` and `b` are considered equal.
```
zerbina:
```suggestion
  ## Returns whether parameters `a` and `b` are considered equal.
```

tests/lang_callable/macros/tmacros_various.nim
Clyybber: Added the explanation to the test in ae58d49dd051593a2155edc350923a5fe412f99c

compiler/vm/vm.nim
```diff
@@ -1550,8 +1550,7 @@ proc rawExecute(c: var TCtx, t: var VmThread, pc: var int): YieldReason =
     of opcEqNimNode:
       decodeBC(rkInt)
       regs[ra].intVal =
-        ord(exprStructuralEquivalent(regs[rb].nimNode, regs[rc].nimNode,
-                                     strictSymEquality=true))
+        ord(exprStructuralEquivalentStrictSymAndComm(regs[rb].nimNode, regs[rc].nimNode))
```
zerbina, line 1553: Could you add a test for this bugfix? (`NimNode` equality now considers comments)

Alternatively, you also use the version ignoring comments and add a `TODO`.

compiler/vm/vm.nim
```diff
@@ -2834,7 +2833,7 @@ proc rawExecute(c: var TCtx, t: var VmThread, pc: var int): YieldReason =
       else:
         block search:
           for existing in g.cacheSeqs[destKey]:
-            if exprStructuralEquivalent(existing, val, strictSymEquality=true):
+            if exprStructuralEquivalentStrictSymAndComm(existing, val):
```
zerbina, line 2836: Same comment as above applies. (comments are now considered for trees included in a `CacheTable`)

tests/lang_callable/macros/tmacros_various.nim
```diff
@@ -328,3 +330,16 @@ block: # bug #15118
...
+    result = nnkCommentStmt.newNimNode()
+    result.strVal = s
+
+  macro testComments =
+    const C1 = (1, createComment("comment 1"))
+    const C2 = (1, createComment("comment 2"))
+    echo treeRepr(C1[1])
+    echo treeRepr(C2[1])
+
+  testComments
```
zerbina:
```suggestion
  const C1 = (1, createComment("comment 1"))
  const C2 = (1, createComment("comment 2"))
  static:
    echo treeRepr(C1[1])
    echo treeRepr(C2[1])
```
A soft suggestion for reducing the number of used language features a bit (macro expansion + compile-time evaluations vs. only compile-time evaluation).

zerbina (approved):
Looks good, not having largely the same logic duplicated multiple times throughout the codebase is much better. I've left a few small suggestions, and there are also two fixes that I think are currently not mentioned/accounted for.

Regarding the PR description, could you mention in the details section that `trees.sameTree` is removed due to it being dead-code? I was initially quite confused where it went :D

compiler/vm/vm.nim
Clyybber: Comment equality was already considered before this PR, because `exprStructuralEquivalent` checked `nkCommentStmt` nodes for equality, so there's no change in behaviour here.

compiler/vm/vm.nim
Clyybber: Comment equality was already considered before this PR, because `exprStructuralEquivalent` checked `nkCommentStmt` nodes for equality, so there's no change in behaviour here.

compiler/vm/vm.nim
zerbina: Oh, you're right, my mistake.

Clyybber:
/merge

Thanks for the reviews :)

PR #1234, opened 2024-03-12
cgir: remove the legacy CGIR variant

zerbina:
## Summary

All code generators use the new goto-using variant of the CGIR now,
rendering the legacy variant obsolete. This is an internal-only
refactoring.

## Details

* remove the `cgirgen_legacy` module
* remove the node kinds belonging to the legacy variant
* remove the sets introduced for the transition
* remove some leftover usages of `cnkStmtListExpr` (they were dead code
  already)


saem (approved):

zerbina:
Hm, strange. The compiler crashed during the leak test (but only during the second attempt at merging). I don't see how the changes here could have caused that, so I'll give merging another go.

PR #1235, opened 2024-03-12
macros: disallow creating special nodes manually

zerbina:
## Summary

Attempting to manually create a node (via `newNimNode`) of kind
`nnkError`, `nnkIdent`, `nnkSym`, or `nnkType` now results in a
compiler error. Creating them was previously allowed, but the
compiler could crash when such incomplete nodes appear in macro
output.

## Details

* all node kinds where the initial state is not valid (`nkError`,
  `nkIdent`, `nkSym`, and `nkType`) are forbidden
* the `opcNNewNimNode` implementation guards against the node kinds
* a VM event, diagnostic, and report is added for the error
* to not introduce a new variant, the `msg` variant is re-used for the
  new VM event
* the untyped `nnkType` test case is removed from `tempty_type_nodes`;
  macros cannot create untyped `nnkType` nodes anymore

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -987,6 +987,7 @@ type
     adVmFieldNotFound
     adVmNotAField
     adVmFieldUnavailable
+    adVmCannotCreateNode
     adVmCannotSetChild
     adVmCannotAddChild
     adVmCannotGetChild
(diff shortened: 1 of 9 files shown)


saem (approved):
The more specific variant might be nice, but this is still a marked improvement and it wouldn't necessarily buy much.

PR #1236, opened 2024-03-13
forbid hook routines raising exceptions

zerbina:
## Summary

This is enforced as follows:
* if a hook routine potentially raises an exception, a compile-time
  error is reported
* if a hook routine raises a defect at run-time, the program panics

Exception effects of hooks side-steps effect tracking for
procedures, and if a hook does raise at run-time, behaviour was
previously undefined. Disallowing hooks to raise exceptions resolves
both issues.

Performance of the produced executables also improves significantly
(depending on the code), as destroy hooks not being able to raise
results in better code generation.

## Details

The implementation is made up of three parts:
* statically enforcing that no exceptions are raised by hooks (in
  `sempass2`)
* preventing exceptions from exiting hooks at run-time (in `mirgen`/
  `liftdestructors`)
* support in the runtime for panicking on unhandled exceptions

### Static Detection

* for hooks, identified by the presence of the `sfOverriden` flag,
  `sempass2` tests against an empty `.raises` specification, ensuring
  that no (tracked) exceptions can be raised
* the existing exception specification is always replaced
* no error is reported when `.raises: []` was explicitly specified, to
  give precedence to the `can raise unlisted exception` error.
* the symbol of hooks is marked with `sfNeverRaises`, to enforce at
  run-time that no exceptions (defects) leave the routine

The meaning of `sfNeverRaises` is changed from being a hint/guarantee
to being a request.

### Enforcement

* `mirgen` wraps the body of `sfNeverRaises` procedure in a
  `try: ... except: nimUnhandledException()`
* elimination of unreachable code in `cgirgen` removes the `except` if
  it's not used in practice
* all synthesized hook procedure (`liftdestructors`) are flagged with
  `sfNeverRaises`; the `canRaise` tracking is removed

### Runtimes

* C runtime: `nimUnhandledException` displays the exception and quits
  the process
* node.js JS runtime: `nimUnhandledException` displays the exception
  and quits the process
* non-node.js JS runtime: `nimUnhandledException` re-raises the
  exception (there's currently no way to terminate the program)
* VM: `nimUnhandledException` is a  `vmop` that raises a
  `vmEvtUnhandledException` event, which aborts execution

### Standard Library and Tests

* fix the `Task` `=destroy` hook being inferred to raise exceptions
* fix `GC_fullCollect` being inferred to raise exceptions

### Tests

Three individual tests need to be adjusted to the language change:
* `tnew` and `gctest` use `debugEcho` instead of `write` (the former
  has no effects)
* `tarcmisc` has to cast away the raise effects of `Stream.close` for
  now

### Specification

The beginnings of a specification test category for hook routines is
added. At the moment, it only covers the exception behaviour.

### DIFF
--- a/compiler/ast/report_enums.nim
+++ b/compiler/ast/report_enums.nim
@@ -422,6 +422,7 @@ type
     rsemCallingConventionMismatch
     rsemHasSideEffects
     rsemCantPassProcvar
+    rsemHookCannotRaise
     rsemUnlistedRaises
     rsemUnlistedEffects
     rsemOverrideSafetyMismatch
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1303,6 +1303,9 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemXCannotRaiseY:
       result = "'$1' cannot raise '$2'" % [r.ast.render, r.raisesList.render]
 
+    of rsemHookCannotRaise:
+      result = "a hook routine is not allowed to raise. ($1)" % r.typ.render
+
     of rsemUnlistedRaises, rsemWarnUnlistedRaises:
       result.add("$1 can raise an unlisted exception: " % r.ast.render,
                  r.typ.render)
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -2058,6 +2058,10 @@ proc generateCode*(graph: ModuleGraph, env: var MirEnv, owner: PSym,
 
   c.scopeDepth = 1
   c.add MirNode(kind: mnkScope)
+  if sfNeverRaises in owner.flags:
+    c.add MirNode(kind: mnkTry, len: 1)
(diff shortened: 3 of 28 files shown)


zerbina:
For user-defined effects (tags) the same reasoning applies, meaning that `.tags: []` should too be enforced for hooks, I think. That's what I had implemented originally, but leaving it to a separate PR (and possible discussion) seemed better, so I removed it again.

zerbina:
In line with what @saem started for the specification test cleanup, I've used a `s99` prefix for the hook category, to signal that its placement needs some further thought / likely isn't final.

saem:
The quoted bullet point below is incomplete, want write sure what you meant to write:

> the symbol of hooks is marked with sfNeverRaises, to enforce the

zerbina:
> The quoted bullet point below is incomplete, want write sure what you meant to write:

Oh, indeed. I've completed the sentence now.

saem (approved):
Very cool, there is the bit in the PR body, but this is good to merge afterwards.

PR #1237, opened 2024-03-13
std/dynlib: fix `loadLibPatterns` having unknown effects

zerbina:
## Summary

The `loadLibPatterns` procedure was erroneously inferred to have
unknown effects (`RootEffect`), making it unusable in `.tags: []`
procedures. This is now fixed.

## Details

`loadLibPatterns` calls the at-that-point forwarded procedure
`loadLib`, so effect inference falls back to assume that
`loadLibPatterns` can have effects.

Since `loadLib`, as well as the other forward-declared procedures,
don't have effects in practice, this is now explicitly specified
on the forward declarations.

### DIFF
--- a/lib/pure/dynlib.nim
+++ b/lib/pure/dynlib.nim
@@ -48,22 +48,24 @@ import std/strutils
 type
   LibHandle* = pointer ## A handle to a dynamically loaded library.
 
-proc loadLib*(path: string, globalSymbols = false): LibHandle {.gcsafe.}
+{.pragma: noeffects, gcsafe, raises: [], tags: [].}
+
+proc loadLib*(path: string, globalSymbols = false): LibHandle {.noeffects.}
   ## Loads a library from `path`. Returns nil if the library could not
   ## be loaded.
 
-proc loadLib*(): LibHandle {.gcsafe.}
+proc loadLib*(): LibHandle {.noeffects.}
   ## Gets the handle from the current executable. Returns nil if the
   ## library could not be loaded.
 
-proc unloadLib*(lib: LibHandle) {.gcsafe.}
+proc unloadLib*(lib: LibHandle) {.noeffects.}
   ## Unloads the library `lib`.
 
 proc raiseInvalidLibrary*(name: cstring) {.noinline, noreturn.} =
   ## Raises a `LibraryError` exception.
   raise newException(LibraryError, "could not find symbol: " & $name)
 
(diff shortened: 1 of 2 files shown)


zerbina:
I've labeled it as such, but I'm not sure if this PR can be considered a bugfix. I'm also not sure whether it warrants a test (there's currently none).

saem (commented):
I think in terms of tests ensuring the effects and tags are empty, and it's GC safe, is sufficient. It would further memorialize the reasoning behind the change. Not strictly necessary.

zerbina:
I suspect that there are going to be similar fixes in the future, and having a separate file for each seems excessive. Therefore, I think placing them all in a single test file would make sense. @saem: Do you have a suggestion for where to place such test and what to name it?

saem:
I would put it in `tests/stdlib/tsystem.nim`. As for it being too big a test files, I think splitting the tests up should follow splitting up `system`.

zerbina:
> I would put it in `tests/stdlib/tsystem.nim`. As for it being too big a test files, I think splitting the tests up should follow splitting up `system`.

Hm, my concern was/is that having a test file like:
```nim
proc test() {.gcsafe, raises: [], tags: [].} =
  # procedure calls go here
```
for each instance of some fix in the standard library similar to the one in this PR would introduce a lot of test overhead. My thinking was that it'd make sense to have a single test file with the above structure where all future fixes would be added to.

Anyways, deciding this at a later point in time is probably fine too.

saem:
> > I would put it in `tests/stdlib/tsystem.nim`. As for it being too big a test files, I think splitting the tests up should follow splitting up `system`.
> 
> Hm, my concern was/is that having a test file like:
> 
> ```nim
> proc test() {.gcsafe, raises: [], tags: [].} =
>   # procedure calls go here
> ```
> 
> for each instance of some fix in the standard library similar to the one in this PR would introduce a lot of test overhead. My thinking was that it'd make sense to have a single test file with the above structure where all future fixes would be added to.

This doesn't have to be done in this PR, but I think this is a type of check/test assertion, which can be part of common test code somewhere. My sense is this is tug of war between moving tests to where we have the testing facility vs having the testing facility where the rest of the test code lives. I suspect this is not a unifying property of these procedures so grouping by the test facility/property under test doesn't make sense, but who knows.

> Anyways, deciding this at a later point in time is probably fine too.

Yup, we can rearrange things later as we learn more.

saem (approved):
Test looks good.

PR #1239, opened 2024-03-13
std/macros: const for nodes requring special init

saem:
## Summary

Introduce a constant  `nnkRequireInitKinds`  for all nodes that require
special initialization, and therefore cannot be constructed via 
`newNimNode` . This allows imports of  `std/macros`  that construct
arbitrary nodes to create a filter.

## Details

Added the constant to  `std/macros` , with export, as well as a test to
ensure we can detect any node kinds not in the list in  `tmacros` .

### DIFF
--- a/tests/stdlib/metaprogramming/tmacros.nim
+++ b/tests/stdlib/metaprogramming/tmacros.nim
@@ -144,3 +144,13 @@ block: # extractDocCommentsAndRunnables
     
   proc c() {.checkComments("Hello world").} =
     ## Hello world
+
+block some_node_kinds_require_specialize_initialization:
+  ## ensure we can init all the rest as a "negative" test
+  const kindsNotRequiringInit = {low(NimNodeKind)..high(NimNodeKind)} -
+                                  nnkRequireInitKinds
+
+  doAssert kindsNotRequiringInit.len > 0
+
+  for k in kindsNotRequiringInit.items:
+    discard newNimNode(k)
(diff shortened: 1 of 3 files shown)


tests/stdlib/metaprogramming/tmacros.nim
```diff
@@ -144,3 +144,13 @@ block: # extractDocCommentsAndRunnables
...
+
+block some_node_kinds_require_specialize_initialization:
+  ## ensure we can init all the rest as a "negative" test
+  const kindsNotRequiringInit = {low(NimNodeKind)..high(NimNodeKind)} -
+                                  nnkRequireInitKinds
+
+  doAssert kindsNotRequiringInit.len > 0
+
+  for k in kindsNotRequiringInit.items:
+    discard newNimNode(k)
```
zerbina:
```suggestion
  static:
    doAssert kindsNotRequiringInit.len > 0

    for k in kindsNotRequiringInit.items:
      discard newNimNode(k)
```
Of course, the compile shouldn't crash when using `newNimNode` outside of a compile-time context. There is a (misplaced) check in `cgen` for this, but it doesn't seem to trigger at the right time.

saem: Thanks, the `cgen` crash is what had me confused.

saem:
/merge

Gonna merge this, it's pretty beneign and we can always break/change it.

saem:
I thought it might be intermittent, it's not. I'll take a peak tomorrow, but it's really weird that this change would result in a leak.

zerbina (approved):

PR #1238, opened 2024-03-13
Don't use float equality for AST comparisons

Clyybber:
## Summary
Float equality lacks the substitution nor reflexivity properties usually
expected from an equality operator, so it's not correct to use float
equality in AST comparisons. This PR changes it so that they are
compared for bit equality.

## Details
*   `0.0`  and  `-0.0`  are not being considered equal by the compiler
anymore, this affects:
    -  static parameters in generic types and procedures, see `tests/statictypes/tstatictypes.nim`
    - default arguments for forward declarations, see `tests/errmsgs/tforwarddecl_defaultparam.nim`
    - ```macros.`==`(a, b: NimNode)```, see `tests/lang_callable/macros/tmacros_various.nim`
    - term-rewriting macros, see `tests/lang_experimental/trmacros/trmacros_various2.nim`

*  `trees.exprStructuralEquivalentStrictSym`  and it's only usage in 
`sem/semfoldnim`  have been removed

PR #1240, opened 2024-03-14
sem: remove float analysis from `sem/guards`

Clyybber:
## Summary

Remove float analysis from `sem/guards` since it's currently unsound.

## Details

`sem/guards`  is only used for static bounds checking (
`--staticBoundChecks` ),  `not nil`  checking, warning for checked field
accesses ( `--warning:ProveField` ) and field accesses of fields
annotated with  `{.guard: ...}` .
Since object variants with a  `float`  discriminator are invalid and
bound checks only concern integers, only conversions to float range
types would be affected under  `--staticBoundChecks` .

It's not clear whether the analysis method here can work with floats
without special treatment in a lot of cases, due to float equality
lacking the subsitution and reflexivity property or the comparison
relations not being total.

### DIFF
--- a/compiler/sem/guards.nim
+++ b/compiler/sem/guards.nim
@@ -39,14 +39,22 @@ from compiler/ast/reports_sem import reportAst,
 from compiler/ast/report_enums import ReportKind
 
 const
-  someEq = {mEqI, mEqF64, mEqEnum, mEqCh, mEqB, mEqRef, mEqProc,
+  # Float operations are not analysed as
+  # it is currently unclear wether they
+  # can be analysed in a sound manner
+  # with the approach used here
+  someEq = {mEqI, mEqEnum, mEqCh, mEqB, mEqRef, mEqProc,
     mEqStr, mEqSet, mEqCString}
+    # `mEqF64` excluded here as it lacks the
+    # substition and reflexivity property
 
   # set excluded here as the semantics are vastly different:
-  someLe = {mLeI, mLeF64, mLeU, mLeEnum,
+  someLe = {mLeI, mLeU, mLeEnum,
             mLeCh, mLeB, mLePtr, mLeStr}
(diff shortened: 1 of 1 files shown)


compiler/sem/guards.nim
```diff
@@ -39,14 +39,22 @@ from compiler/ast/reports_sem import reportAst,
 from compiler/ast/report_enums import ReportKind
 
 const
-  someEq = {mEqI, mEqF64, mEqEnum, mEqCh, mEqB, mEqRef, mEqProc,
+  # Float operations are not analysed as
+  # it is currently unclear wether they
```
saem:
```suggestion
  # it is currently unclear whether they
```

saem (approved):
One minor typo.

So now we're using very simple symbol equality during tree equivalence checks in guards. Are we relying on guards outside of things like checked arithmetic which might be impacted?

saem:
For the summary, the impact here would be not removing floating point operation checks in certain cases?

Clyybber:
> So now we're using very simple symbol equality during tree equivalence checks in guards. Are we relying on guards outside of things like checked arithmetic which might be impacted?
> For the summary, the impact here would be not removing floating point operation checks in certain cases?

`sem/guards` is only used for static bounds checking (`--staticBoundChecks`), not nil checking and warning for checked field accesses (`--warning:ProveField`) and field accesses of fields annotated with `{.guard: ...}`, so it isn't actually used for optimization/removing checks.
Since object variants with a float kind are invalid and bound checks concern integers, only float range checks would be affected under `--staticBoundChecks`.
This PR doesn't actually change to use bit equality for floats (leaving that for #1238), but it does make it so that it shouldn't matter. (Although there are AFAICT other soundness issues in `sem/guards` so there might be cases where it does still matter)

saem:
I moved the extra info into details, guards stuff seems so broken.

PR #1241, opened 2024-03-14
internal: fix an incorrect assertion

zerbina:
## Summary

An assertion in the `injectdestructors` module was incorrect, which led
to the compiler aborting on `x.a = f()` assignments, where `a` is a
discriminator field.

## Details

Only raw lvalue expressions must not appear as the source operand of a
branch-switch assignment, rvalue expression (which aren't included in
`ModifierNodes`) are okay.


saem (approved):
Fixed a very minor typo in the description.

Clyybber (approved):

PR #1242, opened 2024-03-14
CI: Workaround leak sanitizer test failures due to high `vm.mmap_rnd_bits` 

Clyybber:
## Summary
Works around a leak sanitizer crash when  `vm.mmap_rnd_bits`  is 32 bits
by setting it to 28. This currently blocks all CI runs.

See https://github.com/actions/runner-images/issues/9491


saem (approved):
Can you set a reminder to check back in this in a week or so, so we can drop the workaround?

PR #1244, opened 2024-03-15
ci: pin Xcode and macOS SDK version

### LINKED ISSUE #1231: CI failures due to inconsistent binaries buildt on M1 runners
Sometimes the [Binaries from git and source archive should be the same](https://github.com/nim-works/nimskull/blob/devel/.github/workflows/ci.yml#L554) check fails, presumably because of SDK updates making it to one runner before the other.

Most of the times it's [Build release artifacts from source archive (macOS (M1))](https://github.com/nim-works/nimskull/blob/devel/.github/workflows/ci.yml#L322) that needs to be restarted to fix the failure,
other times it's [Build release binaries (macOS (M1))](https://github.com/nim-works/nimskull/blob/devel/.github/workflows/ci.yml#L122) that needs restarting.

The most recent example at the time of writing can be seen at https://github.com/nim-works/nimskull/actions/runs/8237550847/job/22531163678#step:9:30

alaviss:
## Summary
Recently we have seen an excessive amount of reproducibility failures
due to differences in SDK version between M1 jobs.

This PR pins both the Xcode version and the macOS SDK version to avoid
fluctuations caused by runner updates.

## Details
* M1 jobs appears to be flip-flopping between two macOS SDK versions:
14.0 and 14.2. This is found by doing manual analysis via  `llvm-otool` 
on the resulting binaries.
* GitHub's documentation suggests that a new 20240219 version of the M1
runner is still being rolled out. While the changelog does not show any
differences in Xcode configuration, we do observe these changes via our
reproducibility tests.
* GitHub's documentation also outlined that several SDK versions are
available, including 14.0 and 14.2. My speculation is that 14.2 was made
the default (either by accident or intentional) during the runner
updates. As such, pin the SDK back to 14.0 to remove this variable.

Fixes https://github.com/nim-works/nimskull/issues/1231

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,6 +155,18 @@ jobs:
           Invoke-WebRequest https://curl.se/ca/cacert.pem -OutFile (Join-Path $binPath "cacert.pem")
         shell: pwsh
 
+      - name: Set Xcode version (macOS M1)
+        if: runner.os == 'macOS' && runner.arch == 'ARM64'
+        uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: "15.0.1"
+
+      - name: Set macOS SDK version (macOS M1)
+        if: runner.os == 'macOS' && runner.arch == 'ARM64'
+        run: |
+          sdkpath=$(xcrun --sdk macosx14.0 --show-sdk-path)
+          echo "SDKROOT=$sdkpath" >> "$GITHUB_ENV"
+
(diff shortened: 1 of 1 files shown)


saem (approved):
Woof, that's kinda ugly, but I guess it's necessary to include that conditional set each time.

alaviss:
/merge

With two consecutive runs passed, I think this probably fixed the bug. Now onto the third and final run.

PR #1245, opened 2024-03-15
sem: `NimNode` equality respects `nkType`'s type

Clyybber:
## Summary
`exprStructuralEquivalentStrictSymAndComm`  no longer ignores type
equality for  `nkType`  nodes.
This affects  ``macros.`==` ``  for NimNodes and  `macrocache.incl` 
(and  `ic` ).

## Details
A test using ``macros.`==` `` has been added.

### DIFF
--- a/compiler/ast/trees.nim
+++ b/compiler/ast/trees.nim
@@ -97,7 +97,6 @@ makeTreeEquivalenceProc(exprStructuralEquivalentStrictSymAndComm,
   relaxedKindCheck = false,
   symCheck     = a.sym == b.sym,
   floatCheck   = sameFloatIgnoreNan(a.floatVal, b.floatVal),
-  typeCheck    = true,
   typeCheck    = a.typ == b.typ,
   commentCheck = a.comment == b.comment
 )
--- a/compiler/ast/trees.nim
+++ b/compiler/ast/trees.nim
@@ -98,6 +98,7 @@ makeTreeEquivalenceProc(exprStructuralEquivalentStrictSymAndComm,
   symCheck     = a.sym == b.sym,
   floatCheck   = sameFloatIgnoreNan(a.floatVal, b.floatVal),
   typeCheck    = true,
+  typeCheck    = a.typ == b.typ,
   commentCheck = a.comment == b.comment
 )
 export exprStructuralEquivalentStrictSymAndComm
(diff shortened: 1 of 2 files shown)


tests/lang_callable/macros/tmacros_various.nim
```diff
@@ -346,3 +347,22 @@ block:
...
+  # Ensure nkType equality is not ignored by `==` for NimNode
+  macro checkEq(a, b: typed) =
+    echo a == b
+
+  type Exception1 = object of Exception
+  type Exception2 = object of Exception
+  checkEq (;
+    try:
+      discard
+    except Exception1:
```
saem, line 362: Shouldn't identifier/symbol equality check cause these to be unequal, without needing to consider type?

I think the only way to test this would be macro cache and having two object types with otherwise the same name.

compiler/ast/trees.nim
```diff
@@ -97,7 +97,7 @@ makeTreeEquivalenceProc(exprStructuralEquivalentStrictSymAndComm,
   relaxedKindCheck = false,
   symCheck     = a.sym == b.sym,
   floatCheck   = sameFloatIgnoreNan(a.floatVal, b.floatVal),
-  typeCheck    = true,
+  typeCheck    = a.typ == b.typ,
```
saem, line 100: Doesn't this make type comparison apply for all nodes?

saem (commented):
I'm not sure the test is right for demonstrating the issue, but it does show other issues in the first place.

Also, we might need to rename a parameter in your equality procedure construction, if it's only for `nkType`.

The pr title should be stated in current behavior, rather than what doesn't happen, but we can figure that out after.

tests/lang_callable/macros/tmacros_various.nim
Clyybber: These get turned into nkType during sem so there isn't any identifier or sym left in the non-type ast

compiler/ast/trees.nim
Clyybber: Perhaps, I tend to think so but am not entirely sure yet.Though for nkType it only makes sense  to check it's type either way.

saem: I think we should ensure that it only checks types for nodes where the type stands in for what would otherwise be literal AST. An example of which would be `nlType`

saem:
@Clyybber I updated the PR title and body, take a look, and kick-off the merge if you're satisfied.

PR #1247, opened 2024-03-16
make locals' lifetimes bounded by lexical scope for `if`

zerbina:
## Summary

Temporaries and locals that start their storage duration within
`if`/`elif` condition expressions are now destroyed, if necessary,
at the end of the `if`/`elif` branch, not at the end of the `if`'s
enclosing scope. This makes the end of their storage duration
consistent with the end of their lexical scope.

## Details

Making the storage duration end at the end of the `if`/`elif` is
achieved by `mirgen` wrapping the code for each `nkElifBranch` in a
`scope`.

### `while` Lowering

Each `nkElifBranch` being wrapped in a (lifetime) scope, means that
lowering `while cond` into `while true: (if not cond: break; body)`
no longer works, since locals and temporaries starting their
storage duration during evaluation of the `cond` expression would
be destroyed at the end of the `if`, instead of at the end of the
`while`'s body.

To ensure correct and expected lifetimes, `cond` is first assigned
to a temporary, which the `if` then uses for branching. For
example, `while (let x = y; x > 1): body` is lowered by `transf`
into:
```nim
while true:
  let tmp = (let x = y; x > 1)
  if not tmp:
    break
  body
```

Turning `while cond` into `while true: (if cond: body else: break)`
would work too, but it's decide against, because:
* it, currently, results in worse cursor inference results (meaning
  more copies)
* it expands to more MIR code (more work for data-flow analysis)

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1459,12 +1459,13 @@ proc genIf(c: var TCtx, n: PNode, dest: Destination) =
 
   template genElifBranch(branch: PNode, extra: untyped) =
     ## Generates the code for a single ``nkElif(Branch|Expr)``
-    let v = genUse(c, branch[0])
-    c.subTree mnkIf:
-      c.use v
-      c.scope:
-        genBranch(c, branch.lastSon, dest)
-        extra
+    c.scope:
+      let v = genUse(c, branch[0])
+      c.subTree mnkIf:
+        c.use v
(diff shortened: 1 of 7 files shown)


zerbina:
I think the PR description was a bit too abstract/unclear, so I've reworded parts of it and made it a bit more concrete.

saem (approved):

PR #1254, opened 2024-03-23
fix: exception handler not entered for indirect calls

### LINKED ISSUE #1253: incorrect control-flow when indirect call raises with `--panics:on`
### Specification

Indirect calls with procedural values that don't have an explicit exception specification are treated as not raising an exception, when `--panics:on`, resulting in improper control-flow.

### Example

```nim
# build with `--panics:on`

type Proc = proc () {.nimcall.}

proc test(x: Proc) =
  try:
    x()
  except CatchableError:
    echo "caught"

test(proc() =
)
```

### Actual Output

```
Error: unhandled exception:  [CatchableError]
```

### Expected Output

```
caught
```

### Additional Information

* reduced from this [insideout branch's test](https://github.com/disruptek/insideout/blob/c324ceaaa78d6de6eca4c9d92bf59947010dd83e/tests/t33_links.nim)
* with the following [workaround](https://github.com/disruptek/insideout/blob/c324ceaaa78d6de6eca4c9d92bf59947010dd83e/insideout/runtimes.nim#L370), the aforementioned insideout test succeeded

zerbina:
## Summary

Fix exception handlers sometimes being ignored for indirect calls when
building with `--panics:on`.

Fixes https://github.com/nim-works/nimskull/issues/1253

## Details

The `canRaise` procedure treated callables with a `tyProc` without an
exception effect list as not being able to raise, which is incorrect.
No exception effect list means that the exception effects are unknown,
not that there are none.

When panics are enabled, the result of `canRaise` decides whether a
routine call is a checked call (one that can potentially raise).
Procedural values with a type without an explicit `.raises`
specification have no exception effect list, so calls of them were
treated as not raising, leading to the enclosing exception handlers
being elided, if there are no other statements potentially entering the
handler.

`canRaise` is changed to conservatively consider a missing exception
effect list to mean "can raise". In addition, the procedure is
partially refactored to use a case statement, making it not ignore
invalid `tyProc` types anymore.

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -641,14 +641,22 @@ proc canRaise(fn: PNode): bool =
     result = false
   elif fn.kind == nkSym and fn.sym.magic == mEcho:
     result = true
-  else:
-    # TODO check for n having sons? or just return false for now if not
-    if fn.typ != nil and fn.typ.n != nil and fn.typ.n[0].kind == nkSym:
-      result = false
+  elif fn.typ != nil and fn.typ.kind == tyProc:
+    let effects {.cursor.} = fn.typ.n[0]
+    case effects.kind
+    of nkSym:
+      result = false # callable has no effects
+    of nkEffectList:
+      # if the effects were either not computed yet or there's no explicit
+      # specification, assume that the procedure can raise
+      result = effects.len < effectListLen or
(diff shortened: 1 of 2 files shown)


saem (approved):
Hilariously, I think `nkNone` in order to mean effects not computed and `nkEmpty` to mean unknown effects would be pretty nice, but oh well.

PR #1257, opened 2024-03-28
make MCS work when left-hand operand is erroneous

### LINKED ISSUE #1255: UFCS issue iterator into macro
This is kind of difficult for me to explain but it works differently from what I expected and I don't really quite know why

### Example

This example which is what I usually used in nim does not work when trying to read lines from a file into a seq

```nim
import std/[sequtils]
let lines = "filename".lines.to_seq
```

This just brings a type mismatch:

```
day19.nim(16, 25) Error: undeclared field: 'lines' for type system.string [type declared in /home/sotolf/bin/nim-0.1.0-dev.21274/lib/system.nim(34, 3)]
```

I managed to get it working by doing it as follows:

```nim
import std/[sequtils]
let lines = to_seq "filename".lines
```

I'm not quite sure if this is something that is expected, at least in my mind the two examples should generate the same code, unless I'm misunderstanding something. That's the only snag that I hit in my files even though I can be rather sloppy with how I use UFCS, with that line change everything else works like a charm.

zerbina:
## Summary

When using the method-call syntax, templates/macros with `untyped`
parameters are now properly resolved to for `a.f`, `a.f(b)`, and
`a.f = b`, when `a` is itself not semantically valid. The manual is
updated accordingly.

Fixes https://github.com/nim-works/nimskull/issues/1255. 

## Details

In `builtinFieldAccess`, the typed left-hand expression was directly
written back to the input AST, meaning that in case of an error, the
resulting `nkDotCall` already contained an error, preventing recovery
via an `untyped` template/macro.

The input AST is now only modified if the left-hand expression is not
an error, so that the dot-call rewrite gets the original AST. A
regression test covering the three basic dot expression rewrites is
added.

Since the first operand in a `.(a, b)` (originating from `a.b`) can
now be untyped, `resolveOverloads` now needs to ensure that said
operand is typed before attempting to access the expression's type.

---

## Notes for Reviewers
* `builtinFieldAccess` should not modify the input AST at all, but that requires a larger refactor

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/tests/lang_exprs/tmcs_with_erroneous_operand.nim
+++ b/tests/lang_exprs/tmcs_with_erroneous_operand.nim
@@ -1,0 +1,22 @@
+discard """
+  description: '''
+    Ensure that using the method-call syntax doesn't result in an error when
+    the LHS cannot be typed, but an eligible template/macro with an untyped
+    parameter exists.
+  '''
+  action: compile
+"""
+
+template f(x: untyped) =
+  discard
+
+template f(x, y: untyped) =
+  discard
+
+template `f=`(x, y: untyped) =
+  discard
+
+# must not result in an "undeclared identifier" error
+nonexistentName.f
+nonexistentName.f(1)
+nonexistentName.f = 1
--- a/compiler/sem/semcall.nim
+++ b/compiler/sem/semcall.nim
@@ -358,6 +358,8 @@ proc resolveOverloads(c: PContext, n, nOrig: PNode,
     if overloadsState == csEmpty and result.state == csEmpty:
       if efNoUndeclared notin flags: # for tests/pragmas/tcustom_pragma.nim
         if n[0] != nil and n[0].kind == nkIdent and n[0].ident.s in [".", ".="] and n[2].kind == nkIdent:
+          # the first operand might have not been analyzed yet; make sure it is
+          n[1] = semExprWithType(c, n[1])
           let sym = n[1].typ.typSym
           if sym == nil:
             let msg = getMsgDiagnostic(c, flags, n, f)
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -1740,13 +1740,21 @@ proc builtinFieldAccess(c: PContext, n: PNode, flags: TExprFlags): PNode =
       result = semSym(c, n, s, flags)
     return
 
-  n[0] = semExprWithType(c, n[0], flags)
+  let
+    lhs = semExprWithType(c, n[0], flags)
+    rhs = legacyConsiderQuotedIdent(c, n[1], n)
   var
-    i = legacyConsiderQuotedIdent(c, n[1], n)
-    ty = n[0].typ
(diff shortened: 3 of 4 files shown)


tests/lang_exprs/tmcs_with_erroneous_operand.nim
```diff
@@ -0,0 +1,22 @@
...
+  discard
+
+template f(x, y: untyped) =
+  discard
+
+template `f=`(x, y: untyped) =
+  discard
+
+# must not result in an "undeclared identifier" error
+nonexistentName.f
```
saem, line 20: This is susceptible to double evaluation of compile time effects? `(static: echo "hi"; nonexistentName).f`

Longer term with both `untyped` overloads being analysed first and perhaps restricting compile time effects/making them transactional maybe that's not an issue? There might be precedence issues with the former. 🤔

zerbina: Yep, it is susceptible to the double evaluation, as well as leakage of symbol table changes (`(var x = 0; nonexistentName).f`).

I figured that, while not ideal, this is an improvement over now, where `(static: ...; nonexistentName).f` doesn't compile at all.

saem: I'm trying to think of whether this will allow something to work that never should have. If it doesn't then that's an improvement, otherwise we're accepting known invalid programs, which would be a regression.

---

**Possible rewrite**: I thought about this possible rewrite, but I think it leads to a bunch more issues than necessarily all that much better a solution.

An alternative/complementary approach I was thinking of was what if we did some untyped AST rewriting to solve the precedence issues, which I believe this is?

For example, `(var x = 0; nonexistentName).f`, can be thought of as `(var x = 0; nonexistentName.f)`. Pretty sure we can come up with a rewrite rule(s) over untyped AST that has the desired effect. I guess the only "issue" is that an `un/typed` leading parameter used in an MCS won't receive the entire parenthetical syntax, but that wouldn't happen now anyways.

The rewrite rules might be:
1. If there is only one constituent unwrap
2. If there is more than one, wrap the last one in parenthesis, and move the dot operator and RHS in, then apply the rules again

After this change we're saying that an untyped dot operator (pre-rewriting rule) has higher priority than the untyped parenthetical expression.

Once evaluating untyped parameter overloads first rule goes in, we can defer the rewrite until the `untyped` are handled, meaning we'll be able to see the entire LHS in those cases.

zerbina:
@saem and I continued the discussion on [Matrix](https://matrix.to/#/!MAXxvfCYnfYgZqSizw:matrix.org/$QiiMjRA_3AhDKkvFcjpdTFRqIcntcg10F13dLf7oILk), and the conclusion was that a template like `template f(x: untyped)` should generally not match for `x.f`, at least for now.

Therefore, the method-call syntax limitation as described in the manual is going to stay.

PR #1261, opened 2024-03-29
fix(sem): wrong error reported for `raise`

### LINKED ISSUE #112: raise might eat errors when used with strformat
### Example

```nim
import std/strformat

proc exec(cmd: string, args: varargs[string]) =
  let x = @[cmd] & args
  raise newException(CatchableError):
    fmt"{x}"
  # Only referring to a "broken" variable would this issue happen
```

### Actual Output

```
test.nim(5, 3) Error: only a 'ref object' can be raised
```

### Expected Output

The real cause of the error (invalid expression for `x`) should show up instead

```
test.nim(4, 18) Error: type mismatch: got <seq[string], varargs[string]>
but expected one of:
proc `&`[T](x, y: seq[T]): seq[T]
  first type mismatch at position: 2
  required type for y: seq[T]
  but expression 'args' is of type: varargs[string]
proc `&`[T](x: T; y: seq[T]): seq[T]
  first type mismatch at position: 2
  required type for y: seq[T]
  but expression 'args' is of type: varargs[string]
proc `&`[T](x: seq[T]; y: T): seq[T]
  first type mismatch at position: 2
  required type for y: T
  but expression 'args' is of type: varargs[string]
5 other mismatching symbols have been suppressed; compile with --showAllMismatches:on to see them

expression: @[cmd] & args
```

### Additional information

* Mainline does not have this issue

zerbina:
## Summary

Fix errors in expressions within `raise` statements being hidden
behind an "only a 'ref object' can be raised" error.

Fixes https://github.com/nim-works/nimskull/issues/112 

## Details

* make `semRaise` `nkError`-aware (proper error propagation)
* don't modify input AST in `semRaise`
* replace `localReport` in `semRaise` with `nkError`; the necessary
  diagnostics corresponding to the legacy reports are introduced

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -1730,20 +1730,26 @@ proc semCase(c: PContext, n: PNode; flags: TExprFlags): PNode =
     result.typ = typ
 
 proc semRaise(c: PContext, n: PNode): PNode =
-  result = n
   checkSonsLen(n, 1, c.config)
-  if n[0].kind != nkEmpty:
-    n[0] = semExprWithType(c, n[0])
-    var typ = n[0].typ
-    if not isImportedException(typ, c.config):
-      typ = typ.skipTypes({tyAlias, tyGenericInst})
-      if typ.kind != tyRef:
-        localReport(c.config, n.info, reportTyp(
-          rsemCannotBeRaised, typ))
-
-      if typ.len > 0 and not isException(typ.lastSon):
-        localReport(c.config, n.info, reportTyp(
-          rsemCannotRaiseNonException, typ))
+  result = shallowCopy(n)
+  case n[0].kind
+  of nkEmpty:
+    # make sure to copy, the nfSem flag needs to be included
+    result[0] = copyNode(n[0])
+    result[0].flags.incl nfSem
+  else:
+    result[0] = semExprWithType(c, n[0])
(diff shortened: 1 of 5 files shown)


compiler/sem/semstmts.nim
```diff
@@ -1730,20 +1730,26 @@ proc semCase(c: PContext, n: PNode; flags: TExprFlags): PNode =
...
+  else:
+    result[0] = semExprWithType(c, n[0])
+    let typ = result[0].typ
+    if result[0].kind == nkError:
+      result = c.config.wrapError(result)
+    elif not isImportedException(typ, c.config):
+      let refTyp = typ.skipTypes({tyAlias, tyGenericInst})
+      if refTyp.kind != tyRef:
+        result = c.config.newError(result,
+          PAstDiag(kind: adSemCannotBeRaised, excType: typ))
```
saem: Did you already consider making `excType` a query, basically fetching it from `wrongNode`?

saem (approved):
New behavior itself is all good. One minor consideration of using a query instead of saving state for the exception type, but that's about it.

compiler/sem/semstmts.nim
zerbina: No, I had forgotten about the move to a more query-focused diagnostic-to-report tranlsation. Thank you for the reminder -- I'll change it.

PR #1265, opened 2024-04-02
std/socket: Use 0 as default protocol number

iacore:
## Summary

* The default protocol number is changed to 0, as specified by POSIX.
* We are changing it to align the standard library to POSIX standards.

---

## Notes for Reviewers

Prior as https://github.com/nim-lang/Nim/pull/19674.

I don't know what I am doing. I only know that the default protocol number should be 0 (default protocol for AF & socket type, to be determined by the OS).

`nativesocket.nim` has a lot of duplicated code that could be simplified with meta-programming techniques. We should clean that up.

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/lib/posix/posix_freertos_consts.nim
+++ b/lib/posix/posix_freertos_consts.nim
@@ -199,6 +199,7 @@ var LWIP_NSC_IPV6_ADDR_STATE_CHANGED* {.importc: "LWIP_NSC_IPV6_ADDR_STATE_CHANG
 var IF_NAMESIZE* {.importc: "IF_NAMESIZE", header: "<net/if.h>".}: cint
 
 # <sys/socket.h>
+const IPPROTO_DEFAULT* = cint(0)
 var IPPROTO_IP* {.importc: "IPPROTO_IP", header: "<sys/socket.h>".}: cint
 var IPPROTO_IPV6* {.importc: "IPPROTO_IPV6", header: "<sys/socket.h>".}: cint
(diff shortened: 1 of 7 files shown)


iacore:
I forgot to mark this branch as ready for review.

saem:
it doesn't pass ci and the pr message isn't even attempted at, this isn't ready for review.

iacore:
Is it now ready for review?

zerbina:
Please elaborate on why you want to make this change. `nativesockets` doesn't claim to adhere to the POSIX standard, and I also don't see why it should.

iacore:
If I want to use UDP, the default protocol number is wrong. Berkeley socket API says default protocol should be 0.

sol-nimskull.nim
saem, line 1: what is this file?

saem (changes requested):
Can you describe not just the change, but what's the reasoning for the change. The `net` module is meant for cross-platform socket interface, posix alone isn't that, so simply saying that's what it recommends doesn't make sense across all platforms, at least at first glance.

I don't need fancy prose, feel free to use bullet points or whatever makes it easier. But at this point I'd have to do a deep dive across platforms and the various APIs, refresh my understanding the module design as it is today, then reason through the changes made and guess as to what you're thinking. If that's what you're expecting I don't see most people having time to do that any time soon.

sol-nimskull.nim
iacore: I don't know why it ended up there

PR #1266, opened 2024-04-02
mir: store string data separately from nodes

zerbina:
## Summary

Instead of storing string values via `PNode`s in `MirNode`, the
`string` data is stored in a separate table, indexed by `StringId`.
This removes a `PNode` dependency from the MIR and eliminates the
string copy at the MIR/CGIR edge.

### MIR

* the `mnkStrLit` node kind and variant are introduced. The variant
  stores a `StringId`
* the actual string data is stored in a `BiTable[string]` in `MirEnv`
* internally, `StringId` is a `LitId`, which is used for `BiTable`
  lookup
* `mnkLiteral` no longer covers string literals
* the `LiteralDataNodes` set is introduced, currently covering
  `mnkLiteral` and `mnkStrLit`
* `mirgen` translates string literals to `mnkStrLit` nodes, adding the
  string values to the environment in the process

### CGIR

* `CgNode` also uses the indirection via a `StringId` instead of
  storing the raw string directly

### Code generation

* `cgen` and `jsgen` use the `getString` routine added to both code
  generators, for querying a node's string value from the
  environment
* as an interim solution, for generating the C constant for a raw
  string value, a new `genStringLiteral` overload is added to
  `ccgliterals`
* `compat.newStrNode` is obsolete and thus removed

### VM back-end

* `vmserialize.initFromExpr` needs access to a `MirEnv` to look up the
  string data
* `vmjit` provides read-only access to a `JitState`'s `MirEnv` through
  the new public `env` accessor (the `compilerbridge` module need it
  for the `initFromExpr` calls)
* the `MirEnv`'s string table is re-used for the created `PackedEnv`

### Other

* `DataTable` lookup becomes slightly faster, since only the ID for
  strings needs to be compared and hashed
* the `extractStringLiterals` MIR pass can search for string literals
  (`mnkStrLit`) without an extra conditional

### DIFF
--- a/compiler/backend/cgen.nim
+++ b/compiler/backend/cgen.nim
@@ -101,6 +101,9 @@ const
     ## the procedure contains top-level code, which currently affects how
     ## emit, asm, and error handling works
 
+template getString(p: BProc, n: CgNode): untyped =
+  p.env[n.strVal]
+
 proc findPendingModule(m: BModule, s: PSym): BModule =
   let ms = s.itemId.module  #getModule(s)
   result = m.g.modules[ms]
--- a/compiler/backend/cgen.nim
+++ b/compiler/backend/cgen.nim
@@ -101,7 +101,7 @@ const
     ## the procedure contains top-level code, which currently affects how
     ## emit, asm, and error handling works
 
-template getString(p: BProc, n: CgNode): untyped =
+template getString(p: BProc, n: CgNode): string =
   p.env[n.strVal]
 
 proc findPendingModule(m: BModule, s: PSym): BModule =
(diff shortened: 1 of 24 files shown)


compiler/backend/cgen.nim
```diff
@@ -101,6 +101,9 @@ const
     ## the procedure contains top-level code, which currently affects how
     ## emit, asm, and error handling works
 
+template getString(p: BProc, n: CgNode): untyped =
```
saem: Couldn't this be a more precise type than `untyped`?

saem (approved):

compiler/backend/cgen.nim
zerbina: Yep, `string` would work.

Me using `untyped` is habit developed from `typed` templates having some issues when used in conjunction with `lent`/`var` return values (fixed in the `devel` compiler, but not in the csources compiler), though in this case, using `string` *should* be no problem.

I'll change it.

saem: Got it, we might be getting close to updating csources again.

zerbina:
The CI failure is likely caused by an update of the zstd package for the macOS runner. Until all runners use the same image, the only workaround is re-running CI until the reproducibility test passes.

PR #1268, opened 2024-04-04
mir: reference fields by their position

zerbina:
## Summary

Instead of storing the fields' `PSym` in `MirNode`, the fields'
position is stored instead. This is progress towards:
* making `MirNode` a plain data-type
* decoupling the MIR from `sem`'s data types

## Details

* `mnkField` stores the field position as an `int32`, intended to allow
  for negative field positions in the future
* a field position uniquely identifies a field within a record
  hierarchy; allowing for unambiguous symbol look- up
* the new `types.lookupInType` overload accepting an `int` parameter is
  used for looking up a field's symbol
* `types.lookupInType` uses `astalgo.lookupInRecord` internally, which
  is based on `vmtypegen.findInType`
* `findInType` is removed and its usage replaced by `lookupInRecord`
* to ease the transition, the CGIR still references fields via their
  `PSym` directly
* the `tdatatables` test is adjusted to the `MirNode` change


saem (approved):

zerbina:
It looks like the Linux runner is now also affected by the `zstd` version difference.

saem:
I've kicked the Linux build again, I'm hoping it works, or I'll rerun the whole thing.

zerbina:
Okay, that was a waste of resources and time, but the Linux runner image versions finally matched. Let's hope it works again for the merge queue CI run...

PR #1270, opened 2024-04-06
mir: split up `mnkLiteral`

zerbina:
## Summary

Split up `mnkLiteral` into its constituent kinds and replace the
`PNode`-based data representation for literals with a data-oriented
one. This removes the last `ref` type usage from within the `MirNode`
record-case. In addition, comparing numeric values becomes a single
integer comparison.

## Details

`mnkLiteral` is split into `mnkNilLit`, `mnkIntLit`, `mnkUIntLit`,
`mnkFloatLit`, and `mnkAstLit`. Same as with the CGIR, numeric type
width is not part of the syntax (it's not needed).

The `MirNode` variant for the numeric literal nodes stores a
`NumberId`, which is a `LitId` uniquely identifying a bit-pattern
(stored as `BiggestInt` in `MirEnv.numbers`).

A shared storage is used because:
* it's easier to implement (one table instead of three; no special
  handling for floats is needed)
* less memory usage (value storage is shared between the three numeric
  literal kinds)
* only one ID type is needed, instead of three
* reinterpreting bit-patterns is made slightly easier

Literal AST (identified by the new `AstId` type) doesn't use a table,
only a `Store`. Their sparse usage doesn't warrant the implementation
complexity / run-time overhead of hashing/comparing the trees (though
`treetabs` could be used).

### Translation

* for integer and float literals, `mirgen` performs the same
  translation that `cgirgen` previously did in `translateLit`
* so that the code generators don't need to be adjusted, the CGIR still
  stores the numeric and AST values directly in `CgNode`
* to prevent excessive parameter passing in `cgirgen`, a pointer to the
  `MirEnv` is stored in `TranslateCl`. Access to the environment is
  needed for looking up the values for `NumberId` and `AstId`
* `compat.translate` now requires a `MirEnv` argument, for looking up
  the numerical values
* `cgirgen.translateLit` is obsolete and thus removed
* raw field symbols are transported to the code generator via a
  `mnkAstLit`/`cnkAstLit` (a `mnkLiteral` node was previously used)

### VM backend

* the `MirEnv.numbers` table is re-used for the `PackedEnv`. This
  allows for also re-using the `NumberId`s (they can be treated
  directly as a `LitId`)
* the serialization logic is adjusted to the new MIR syntax

### Tests

* the `mnkLiteral` usage in `tdatatables.nim` is replaced
* `tcompilerapi.nim` passed untyped float values to
  `nimeval.callRoutine`, which no longer works, since `constDataToMir`
  always expects typed literal nodes now. The test is changed to pass
  typed `nkFloatLit` nodes

### DIFF
--- a/compiler/mir/mirenv.nim
+++ b/compiler/mir/mirenv.nim
@@ -43,8 +43,13 @@ type
       ## includes both normal globals and threadvars
     procedures*: SymbolTable[ProcedureId, PSym]
 
+    numbers*: BiTable[BiggestInt]
+      ## all numerical values referenced by the MIR, stored as bit patterns
     strings*: BiTable[string]
       ## all string data referenced by the MIR
+    asts*: Store[AstId, PNode]
+      ## all AST fragments referenced by the MIR. No unification is
+      ## performed
 
     bodies*: OrdinalSeq[ConstId, DataId]
       ## associates each user-defined constant with its content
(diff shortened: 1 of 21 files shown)


compiler/mir/mirenv.nim
```diff
@@ -43,8 +43,13 @@ type
       ## includes both normal globals and threadvars
     procedures*: SymbolTable[ProcedureId, PSym]
 
+    numbers*: BiTable[BiggestInt]
```
saem, line 46: Would it be too much to have an `mnkEnumField` kind and a store for the field values?

saem (approved):
Bit of a bummer to lose the enum field names, but I also understand it might cause a lot of issues putting that mapping in.

compiler/mir/mirenv.nim
zerbina: I think it would be too much, yeah. The existence of enum fields is irrelevant to the mid-end and code generators -- they only care about the numeric values.

Nonetheless, if you want, I can update the pretty-printer to render enum values as their field names again.

saem: > Nonetheless, if you want, I can update the pretty-printer to render enum values as their field names again.

That would be nice if not too much trouble, just makes tracing and debugging easier.

zerbina: I've updated the MIR pretty-printing of literals to produce results that match the previous ones (the PR description is updated already).

saem: Thanks

PR #1271, opened 2024-04-06
cgen: simpler `openArray` parameter handling

zerbina:
## Summary

Handle the special treatment of `openArray` types when emitting
parameter declarations, and remove the `kind` parameter from
`cgen.getTypeDesc` and friends.

## Details

When requesting the type name/description for an `openArray[T]` type in
a parameter position, the `openArray[T]` type was effectively treated
as a `ptr T`. The relevant logic was in `mapType` and
`getOpenArrayDesc`, with the `kind` parameter providing the context
necessary for deciding how to interpret `openArray`.

Not only is this error-prone (i.e., when passing the wrong `TSymKind`),
it's also unnecessarily complex. Requesting the element type in
`genProcParams` and then appending the pointer specification (`*`)
achieves the same.

The `kind` parameter is removed from `getTypeDesc` and friends, and
`tyOpenArray` types are always committed to the type cache (which was
previously not possible). In addition, emitting an extra `Len`
parameter per nested `openArray` type (e.g.: `openArray[openArray[T]]`)
is removed -- these are not accounted for anywhere else.

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -146,7 +146,7 @@ proc genOpenArrayConv(p: BProc; d: TLoc; a: TLoc) =
 proc genAssignment(p: BProc, dest, src: TLoc) =
   # This function replaces all other methods for generating
   # the assignment operation in C.
-  case mapType(p.config, dest.t, skVar)
+  case mapType(p.config, dest.t)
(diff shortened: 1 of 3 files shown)


saem (approved):

saem:
/merge

Wow, that was a lot of complexity.

PR #1273, opened 2024-04-08
mir: add dedicated 'mutable' version for view creation

zerbina:
## Summary

Encode in syntax whether a `mnkView`/`mnkToSlice` creates a mutable
views, by introducing two new node kinds. This allows creation of a
data-flow graph without access to type information.

## Details

* add `mnkMutView` and `mnkToMutSlice` to the MIR
* `mirgen` picks which operation to use based on the expressions
  return type
* `mirexec.computeDfg` now decides the data-flow operation(s) for a
  view creation operation based purely on syntax
* the mutable versions are currently pretty-printed in the same way as
  their non-mutable counterpart
* no new CGIR node kinds are introduced, as they're not needed

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1835,13 +1835,13 @@ proc genx(c: var TCtx, e: PMirExpr, i: int) =
     c.buildOp mnkStdConv, n.typ:
       recurse()
   of pirToSlice:
-    c.buildOp pick(n.typ, mnkMutToSlice, mnkToSlice), n.typ:
+    c.buildOp pick(n.typ, mnkToMutSlice, mnkToSlice), n.typ:
       recurse()
   of pirToSubSlice:
     # the array operand is a PMIR expression already, but the operands
     # specifying the bounds are not
     let
-      op = pick(n.typ, mnkMutToSlice, mnkToSlice)
+      op = pick(n.typ, mnkToMutSlice, mnkToSlice)
       a = n.orig[2]
       b = n.orig[3]
     if optBoundsCheck in c.userOptions and needsBoundCheck(n.orig[1], a, b):
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1728,10 +1728,15 @@ proc genx(c: var TCtx, e: PMirExpr, i: int) =
   template recurse() =
     genx(c, e, i - 1)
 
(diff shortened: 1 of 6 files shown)


compiler/mir/mirgen.nim
```diff
@@ -1728,6 +1728,11 @@ proc genx(c: var TCtx, e: PMirExpr, i: int) =
   template recurse() =
     genx(c, e, i - 1)
 
+  proc pick(typ: PType, mut, nonMut: MirNodeKind): MirNodeKind =
+    case typ.skipTypes(abstractInst).kind
+    of tyVar: mut
+    else:     nonMut
```
```suggestion
  proc viewKind(typ: PType, kind: MirNodeKind): MirNodeKind =
    proc pick(typ: PType, mut, nonMut: MirNodeKind): MirNodeKind =
      case typ.skipTypes(abstractInst).kind
      of tyVar: mut
      else:     nonMut

    case kind
    of mnkView: pick(typ, mnkMutView, mnkView)
    of mnkToSlice: pick(typ, mnkMutToSlice, mnkToSlice)
    else: unreachable(kind)
```

That might be cleaned up a tad more perhaps, but this way the mut vs non-mut decision is in one place (harder to accidentally swap param positions). Might not be worth it, just something to consider.

compiler/mir/mirtrees.nim
```diff
@@ -148,6 +149,7 @@ type
                   ##   upper bound
     # XXX: consider using a separate operator for the slice-from-sub-sequence
     #      operation
+    mnkMutToSlice ## version of ``mnkMutToSlice`` for creating a mutable slice
```
saem:
```suggestion
    mnkMutToSlice ## version of ``mnkToSlice`` for creating a mutable slice
```

I think it might also be better to rename this to `mnkToMutSlice`, it's perhaps more clear we're going to end up with a mutable slice, rather than drop that property.

saem (approved):
Left a few small soft suggestions, I believe they'd be improvements.

compiler/mir/mirgen.nim
zerbina: Yeah, problems arising from accidentally swapping the parameters is something I've also thought about. I'll look into how much effort it'd take to integrate this into the PMIR instead -- if too much, I'd use your suggestion.

compiler/mir/mirtrees.nim
zerbina: That seems like an improvement -- I'll change it.

PR #1274, opened 2024-04-08
sem: reduce cascading errors in call expressions

zerbina:
## Summary

* reduce cascading errors in call expressions when there's one or more
  erroneous argument expressions (especially noticeable with
  `nim check` and `nimsuggest`)
* reduce cascading errors with generic type instantiation expressions
* a macro/template with an `untyped` parameter is no longer ambiguous
  with another macro/template that has a `typed` parameter, when the
  argument has an error

## Details

There was already support for "faux matches", but it stopped working
with the proliferation of `nkError` usage. A "faux match" is a match
during overload resolution that could only work because unknown types
where treated as wildcards.

Upon encountering an erroneous argument, instead of failing the match,
`matchesAux` now continues, with `paramTypesMatchAux` also letting the
error node through. This restores basic faux match support in the case
of errors.

Multiple other changes/fixes are required to make faux matches work
properly for errors:
* the `fauxMatch` field of `TCandidate` wasn't reset by
  `initCandidate`, leading to faux matches carrying over to other
  candidates
* for proper error propagation, `semResolvedCall` needs to wrap error
  matches in a wrapper error
* `semGeneric` needs to account for faux matches, by reporting the
  errors, if any

Special-casing `typed` parameters is no longer needed -- the same
parameter type matching as with other non-`untyped` parameters apply.

### Recovery through `untyped`

So that error-recovery through `untyped` templates/macros continues to
work, `cmpCandidates` always gives a higher weight to non-erroneous
candidates. If both candidates are erroneous, the normal comparison
applies.

This also fixes the overload tie behaviour between `typed` and
`untyped` parameters in the face of errors (refer to
`tuntyped_overload_tie_with_typed_with_error.nim`).

### Ambiguous faux matches

If overload resolution finds two equally well matching candidates, but
both only matched because of errors, `resolveOverloads` doesn't report
an ambiguity error, since it'd only be follow-up error.

### Other tests

`tcon1.nim` had to be adjusted to produce the same results. Due to the
missing closing parentheses, and faux matches now working again,
the `p` overload three parameter was marked as being used *twice*,
leading to it being shown before the other overload (which was
considered to only be used *once*).

Finally, two new tests are added for preventing future regressions with
faux matches.

### DIFF
--- a/compiler/sem/semcall.nim
+++ b/compiler/sem/semcall.nim
@@ -530,7 +530,10 @@ proc semResolvedCall(c: PContext, x: TCandidate,
   if x.hasFauxMatch:
     result = x.call
     result[0] = newSymNode(finalCallee, getCallLineInfo(result[0]))
-    if containsGenericType(result.typ) or x.fauxMatch == tyUnknown:
+    if x.fauxMatch == tyError:
+      # at least one argument expression was erroneous
+      result = c.config.wrapError(result)
+    elif containsGenericType(result.typ) or x.fauxMatch == tyUnknown:
       result.typ = newTypeS(x.fauxMatch, c)
       if result.typ.kind == tyError: incl result.typ.flags, tfCheckedForDestructor
     return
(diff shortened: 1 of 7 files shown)


saem (approved):

saem:
I believe this is the right direction, the only issue, which is preexisting, is that "faux match" is kinda a terrible name.

PR #1275, opened 2024-04-11
mir: distinguish between proc values and references

zerbina:
## Summary

Introduce `mnkProcVal`, for taking over the "procedural value" meaning
of `mnkProc`, the latter which only means "reference to procedure" now.
This clarifies the syntax a bit, but, more importantly, allows
`mnkProc` nodes to not require a type.

## Details

* add the `mnkProcVal` kind and update the grammar
* where the value interpretation was used, replace usages of `mnkProc`
  with `mnkProcVal`
* for the sake of keeping the code generators as is, `mnkProc` is still
  translated to a typed `cnkProc`
* `mirconstr.buildCall` no longer requires a procedure type
* all static calls use `mnkProc` instead of `mnkProcVal`

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -281,7 +281,7 @@ iterator deps*(tree: MirTree): lent MirNode =
       # skip over the name slot:
       i = NodePosition tree.operand(i, 1)
       continue
-    of mnkProc:
+    of mnkProc, mnkProcLit:
       yield tree[i]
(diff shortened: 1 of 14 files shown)


saem (approved):

saem:
I somewhat prefer `mnkProcVal` over `mnkProcLit`

zerbina:
That sounds much better, thanks -- I'll change it.

PR #1276, opened 2024-04-12
fix: internal compiler errors with closure iterators

zerbina:
## Summary

Fix two bugs with the closure iterator transformation that resulted in
internal compiler errors. One affected yields in `while` loops, the
other yields in object down-conversion expressions.

## Details

1. whether a split is required wasn't propagated upwards for
   `nkWhileStmt` statements, leading to their enclosing statement-list
   expressions to not be lifted, if there was no other source for
   splits
2. the `nkObjDownConv` node wasn't properly handled by
   `lowerStmtListExpr`, meaning that the statement-list expression
   unpacking wasn't performed

Two regression tests are added to `titer_issues.nim`.


saem (approved):

saem:
/merge

Nice find.

PR #1277, opened 2024-04-15
mir: use `TypeId`

zerbina:
## Summary

Store references to types as `TypeId` instead of `PType`, making
`MirNode` a plain data-type. Besides speeding up the compiler, this
also paves the way to a data-oriented type IR for the MIR.

### Overview

* all MIR-level types use `TypeId` instead of `PType` (currently only
  `MirNode` and `Local`)
* the new `TypeEnv` type from the `mirtypes` module manages the type
  storage and mappings
* a `TypeEnv` instance is embedded in `MirEnv`
* registering a `PType` with `TypeEnv` yields a `TypeId`
* querying type information requires access to a `TypeEnv` instance

### Mapping Types

`TypeEnv` stores the `PType`-to-`TypeId` mappings in a `TypeTable`.
`TypeTable`, provided by the new `typemaps` module, is a custom
`Table` type that de-duplicates the `PType` keys by using structural
or symbolic comparison/hashing for the keys, thereby significantly
reducing the number of mappings and registered types.

At the moment, `PType` is still used as the type IR, so `TypeEnv` only
stores a `PType` for each `TypeId`.

The built-in types are mapped to statically known IDs, making some
translation and MIR construction easier.

### Constant expressions

To not require passing a `TypeEnv` to `datatables.getOrPut`, only the
type IDs are compared. This doesn't produce the same results as
`sameBackendType`, which ignores, for example, the distinct type
modifiers and aliases. As a result, `IntAlias(1)` and `int(1)` now
result in two `DataTable` entries instead of one.

Only using `TypeId` comparison/hashing does speed up `getOrPut`,
however.

### CGIR

* the CGIR still uses `PType`
* `cgirgen` turns the input `TypeId`s back into `PType`s
* since `cgirgen` creates new `PType` and assigns them to `Local`s,
  access to a *mutable* `MirEnv` is now required (in order to register
  the new `PType`s)

### VM backend

Packing constant data requires querying type information, and access to
a `TypeEnv` is thus required. A pointer to the `TypeEnv` instance is
passed along in `DataEncoder`.

### VM Evaluation Pass

Initializing the JIT environment requires a basic type environment,
which is not present when using the VM evaluation pass, as the `system`
module isn't yet processed at the time the `PEvalContext` is
initialized.

To delay setting up the JIT environment as much as necessary, the
evaluation pass is restructured:
* only a VM instance is assigned to the `ModuleGraph` at the start
* `setupGlobalCtx` creates a proper evaluation context from the
  instance
* evaluating statements, which now calls `setupGlobalCtx`, is prevented
  for declarative statements

This ensures that `initJit` is only called after the processing of the
`system` module has set up the basic type environment.

### Compiler Performance

`MirNode` being a plain data-type (no destructor) significantly speeds
up all MIR processing that creates new nodes or copies/moves them,
especially `TreeChangeset` application (~ -70%).

In addition, no copy/sink/destroy needing to be injected for `MirNode`s
also speeds up bootstrapping, since it means less work for the
destructor injection pass.

### DIFF
--- a/compiler/backend/compat.nim
+++ b/compiler/backend/compat.nim
@@ -24,7 +24,8 @@ import
   ],
   compiler/mir/[
     mirenv,
-    mirtrees
+    mirtrees,
+    mirtypes
   ],
   compiler/utils/[
     bitsets,
@@ -146,7 +147,7 @@ proc translate*(t: MirTree, env: MirEnv): CgNode =
     template tree(k: CgNodeKind, body: untyped): CgNode =
       ## Convenience template for setting up the tree node and iterating the
       ## input node's child nodes.
-      let res {.inject.} = newExpr(k, unknownLineInfo, n.typ)
+      let res {.inject.} = newExpr(k, unknownLineInfo, typ)
       res.kids.newSeq(t[i - 1].len)
       for j in 0..<res.len:
         res.kids[j] = body
@@ -154,12 +155,14 @@ proc translate*(t: MirTree, env: MirEnv): CgNode =
       inc i # consume the end node
       res
 
-    let n {.cursor.} = t[i]
+    let
+      n {.cursor.} = t[i]
+      typ {.cursor.} = env[n.typ]
     inc i # advance to the first child node
     case n.kind
     of mnkObjConstr:
       tree cnkObjConstr:
-        let field = lookupInType(n.typ, t[i].field.int)
+        let field = lookupInType(typ, t[i].field.int)
         inc i # advance to the arg node
         CgNode(kind: cnkBinding, info: unknownLineInfo,
                kids: @[CgNode(kind: cnkField, field: field),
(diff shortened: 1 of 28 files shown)


compiler/backend/compat.nim
```diff
@@ -146,20 +147,22 @@ proc translate*(t: MirTree, env: MirEnv): CgNode =
     template tree(k: CgNodeKind, body: untyped): CgNode =
       ## Convenience template for setting up the tree node and iterating the
       ## input node's child nodes.
-      let res {.inject.} = newExpr(k, unknownLineInfo, n.typ)
+      let res {.inject.} = newExpr(k, unknownLineInfo, typ)
```
saem, line 154: No change required, just a bit weird that `typ` is used before it's definition below.

compiler/mir/mirgen.nim
```diff
@@ -1150,42 +1148,33 @@ proc genSetConstr(c: var TCtx, n: PNode) =
...
     # need
 
-    c.subTree (if isConsume: mnkConsume else: mnkArg): # the environment
-      if n[1].kind == nkNilLit:
-        # it can happen that a ``nkNilLit`` has no type (i.e. its typ is nil) -
-        # we ensure that the nil literal has the correct type
-        # TODO: prevent a ``nkNilLit`` with no type information from being
-        #       created instead
-        c.add MirNode(kind: mnkNilLit,
-                      typ: c.graph.getSysType(n[1].info, tyNil))
```
saem, line 1181: Sorry, might be a silly question, but when did the change to ensure `nkNilLit` had a `typ` field happen? IIUC, this should be sem fix but I can't recall it. Maybe not quite which specific commit or anything, but more was this a series of fixes over time and became vestigial, or was there something specific?

compiler/mir/typemaps.nim
```diff
@@ -0,0 +1,187 @@
...
+proc hash*(x: Type): Hash {.inline.} =
+  ## Leaked implementation detail -- do not use.
+  hash(PType x)
+
+proc `[]`*[T](t: TypeTable[T], key: PType): lent T {.inline.} =
+  ## Looks up the item for `key`.
+  t.inner[Type key]
+
+proc `[]=`*[T](t: var TypeTable[T], key: PType, val: sink T) {.inline.} =
+  t.inner[Type key] = val
```
saem, line 184:
```suggestion
  ## Assign a `val` for the `key`.
  t.inner[Type key] = val
```

compiler/mir/typemaps.nim
```diff
@@ -0,0 +1,187 @@
...
+    # a distinct type so that a new hash and equality operator can be
+    # attached
+  TypeTable*[T] = object
+    inner: Table[Type, T]
+
+proc safeId(n: PNode): NodeId {.inline.} =
+  if n != nil: n.id
+  else:        NodeId -1
+
+proc cmp(a, b: PType): bool {.noSideEffect.}
```
saem: Why not `func`?

saem (approved):
Looks good.

All comments are minor/non-blocking.

compiler/mir/mirgen.nim
zerbina: No, not a silly question. This removal was originally intended to be a separate PR, but it felt to minuscule to warrant the PR overhead.

This wasn't an issue with sem, rather with `transf` (as only there are `nkClosure` nodes created), but I don't remember which change exactly fixed this bug. If I remember correctly, the problem was related to the JavaScript backend being selected, so *believe* it was #586 that fixed it.

compiler/mir/typemaps.nim
zerbina: No specific reason. The compiler complained (rightfully so) about that calling the forward-declared `cmp` had side effects, so I mindlessly added the pragma. I'll change it to use `func`.

compiler/backend/compat.nim
zerbina: It is. I do prefer using either early binding or dedicated parameters for templates, and there's nothing speaking against using early binding here -- I'll change it.

zerbina:

PR #1278, opened 2024-04-17
fix behaviour of closed-over locals in `and`/`or` expressions

zerbina:
## Summary

Fix closed-over locals defined as part of `and`/`or` expressions not
having their default value when their defined-in sub-expression was
short-circuited.

## Details

Bindings part of `and`/`or` expressions were so far hoisted to the
start of their scope by the "unscoped def" handling in `cgirgen`, which
doesn't apply to lifted locals, since these don't have a `def` (they're
part of the environment object).

Performing the hoisting in `transf` makes sure the behaviour of closed-
over locals matches that of not closed-over ones, and it also removes
the reliance on `cgirgen`.

The hoisting works by scanning transformed `and`/`or` expressions for
bindings that are within the *same* scope as the `and`/`or` expression,
and moving them to the start of the `and`/`or` expression.

For example, `(let a = 0; x) or (let b = 0; y)` is transformed into
`(let a; let b; (a = 0; x) or (b = 0; y))`.

Lifted globals (`.global` and `.thread`) have to be accounted for by
the hoisting pass, since they weren't lifted at this point.

### DIFF
--- a/compiler/sem/transf.nim
+++ b/compiler/sem/transf.nim
@@ -1032,6 +1032,81 @@ proc flattenTree(root: PNode): PNode =
   else:
     result = root
 
+proc transformAndOr(c: PTransf, n: PNode): PNode =
+  ## Transforms both operands and hoists all locals within them that are in
+  ## the outermost scope to the start of the and/or expression:
+  ##
+  ##   (let x = 0; a) or (var y; y = 1; b)
+  ##   # ->
+  ##   (let x; var y; (x = 0; a) or (y = 1; b))
+  ##
+  ## This makes sure that the bindings' lifetime is delimited by the `and`/`or`
+  ## expression's enclosing scope, even after lowering the expression.
+  proc hoist(c: PTransf, n: sink PNode, target: PNode): PNode {.nimcall.} =
+    # traverse all statements/expressions within the *same* scope
+    case n.kind
+    of nkVarSection, nkLetSection:
+      result = newTreeI(nkStmtList, n.info)
+
+      for it in n.items:
+        case it.kind
+        of nkIdentDefs:
+          if (it[0].kind == nkSym and (sfGlobal notin it[0].sym.flags or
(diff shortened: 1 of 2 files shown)


compiler/sem/transf.nim
```diff
@@ -1032,6 +1032,81 @@ proc flattenTree(root: PNode): PNode =
...
+        of nkIdentDefs:
+          if (it[0].kind == nkSym and (sfGlobal notin it[0].sym.flags or
+              it[0].sym.owner.kind notin routineKinds)) or
+             it[0].kind == nkDotExpr:
+            # local or module-level global. The initializer expression must be
+            # processed first, since locations defined therein start their
+            # lifetime earlier. That is, for ``var x = (var y = 0; y)`` the
+            # ``var y`` must be hoisted first
+            let src = hoist(c, move it[2], target)
+            target.add newTreeI(n.kind, it.info,
```
saem, line 1062: When `n.kind` is an `nkLetSection` aren't we going to cause potential issues downstream when assigning via such a handle?

Not really expecting a change, seeing as it does work, given the test; at most a comment, but not sure it requires that.

saem (approved):
Just a question/discussion point, it's good to merge regardless.

compiler/sem/transf.nim
zerbina: That's a good question, thank you for asking. Including and beyond `transf`, `let` and `var` sections are effectively interchangeable (only the symbol kind matters). A `let` binding not being assigned a value immediately is also fine.

What would be a problem is if a `let` location is either read from *before* it's assigned to, or assigned to multiple times (`mirgen` relies on both to never happen), but the hoisting cannot cause either.

saem: Thanks for clarifying, that was my guess. I also think it's good in terms of long term direction, because making `let`, or some variant, to be an initialize once before use, rather than once at the time of definition, would be nice.

PR #1279, opened 2024-04-17
mir: shrink size of `MirNode`

zerbina:
## Summary

Shrink the size of a `MirNode` from 24 byte (on a 64bit target) to 16
byte, reducing memory consumption and copying cost, as well as
increasing the number of nodes fitting in a cache line.

## Details

The `len` field was an int, resulting in the case having an
alignment requirement and size of 8 byte (for a 64bit target).
Since the `len` field is never negative, it's changed into a `uint32`
field. This also makes the size of a `MirNode` target-independent.


saem (approved):
Yup, this should pay off nicely over time.

PR #1281, opened 2024-04-19
mir: implement the new control-flow primitives

### LINKED ISSUE #1303: value is implicitly moved even though it's used afterwards
### Example

```nim
type Object = object
  has: bool

var wasDestroyed = false

  if x.has:
    wasDestroyed = true

proc f_sink(x: sink Object) =
  # disarm the destructor:
  x.has = false

proc test() =
  var o = Object(has: true)
  try:
    try:
      f_sink(o) # must not be sunken
    except IOError:
      # omitting the inner try makes the code work as expected
      discard "not reached"
  except CatchableError:
    doAssert not wasDestroyed
    discard o # use `o`, preventing it from being moved earlier

test()
doAssert wasDestroyed
```

### Actual Output

The second assertion fails, as `o` was erroneously sunken into the call argument, where the destructor was then disarmed, and `wasDestructor`  thus never set to true.

### Expected Output

The program compiles and runs without error. `o` is used beyond the procedure call, so it should not be moved.

### Additional Information

The culprit is here: https://github.com/nim-works/nimskull/blob/f401e73639c2d15f2aefbad04c96199c8308b3dc/compiler/sem/mirexec.nim#L524-L533

There's no connection between the `raise` and the outer `except` in the control-flow graph, due to a missing edge between the inner and outer `except`.

zerbina:
## Summary

Replace the legacy control-flow primitives with the new ones outlined
in the documentation. This is a refactor for the most part, but - as a
side effect - a bug is fixed too, and destructors are no longer called
on not-yet-alive locations.

Fixes #1303.

## Details

The new control-flow primitives are those documented in the `mir.rst`
file -- the node kinds for the legacy ones are all removed. They're the
same as those used by the CGIR, removing the largest remaining
difference between the two IRs, preparing for the CGIR's removal.

In broad:
* translating the `try`/`block`/`break` control-flow to the goto-based
  control-flow now happens during AST -> MIR translation (`mirgen`)
  instead of during MIR -> CGIR translation (`cgirgen`)
* elimination of unreachable code now happens during AST -> MIR
  translation already
* all necessary destroy operations are injected during AST -> MIR
  translation
* the "destructor injection" pass now only collapses sink operations
  and eliminates destroy operations that turn out to be redundant

Adding new finally sections after the initial MIR production would
require patching all intercepted gotos and checked calls, which is
complex and hurts composition of the pass, hence this approach being
decided against.

### Multiple `finally`s Per Scope

Instead of using a single `finally` per scope (or none at all),
multiple ones are used, depending on the placement of unstructured
scope exits. As a consequence:
```nim
block:
  var x = Object()
  if cond:
    break # no longer invokes the destructor for `y`
  var y = Object() # destruction doesn't require a finally
```

This generally enables better code generation, since locals destroyed
in `finally` sections don't have to conservatively be zero'ed on scope
entry anymore.

### Other Changes

* the MIR pretty-printer (`mir/utils.nim`) is updated to render the new
  node kinds
* `transf` no longer wraps `while` loop bodies in a `block`, which was
  previously required for proper destruction of locals defined as part
  of the body
* `mirtrees.numArgs` now only returns the number of arguments in a call
  tree, not the number of child nodes
* repeated are `pirMat`/`pirMatCursor` nodes are guarded against during
  PMIR production
* the `mnkDefUnpack` node kind is obsolete now that destroy operations
  are placed by `mirgen` -- it's removed

### Test Changes

* tests using now-removed MIR node kinds are changed to use different
  ones
* some assertions in now-working tests are updated
* the `--expandArc`-using tests are updated to the changes in MIR
  rendering


compiler/mir/mirgen_blocks.nim
```diff
@@ -0,0 +1,293 @@
...
+    of bkBlock:
+      label*: PSym
+        ## the symbol of the block's label. nil if it's an internal block
+    of bkScope:
+      numRegistered: int
+        ## number of entities registered for the scope in the to-destroy list
+      scopeExits: seq[LabelId]
+        ## unordered set of follow-up targets
+    of bkTryFinally:
+      doesntExit*: bool
```
zerbina, line 46: This field is currently never set to 'true'. There's no technical reason for not implementing the "doesn't exit" tracking as part of this PR, but it's only an optimization, and I'd prefer implementing it as a follow-up in order to not grow the PR any further.

compiler/sem/injectdestructors.nim
```diff
@@ -92,15 +94,15 @@ import
     mirconstr,
     mirenv,
     mirtrees,
-    sourcemaps
+    sourcemaps,
+    utils # unused import, but keeping it significantly speeds up the compiler
+    # XXX: this is possibly caused by some hooks changing which module they're
+    #      part of
```
zerbina, line 100: I spent multiple days trying to figure out why exactly this happens. While the simple experiment of using the `.inline` calling convention for all synthesized `=copy` and `=sink` hooks showed to indeed fix this performance issue (i.e., leaving the `utils` import out no longer had an impact on compiler performance), I didn't manage to prove *why* that's the case, and eventually gave up for the time being.

tests/arc/topt_wasmoved_destroy_pairs.nim
```diff
@@ -6,93 +6,103 @@ discard """
...
+        add(name a, consume _5)
+        goto [L2]
+  scope:
+    def _6: seq[int] = move x
+    add(name b, consume _6)
+  goto [L0, L3]
+  finally (L0):
+    =destroy(name b)
+    =destroy(name a)
+    continue {L3}
```
zerbina, line 23: This a minor regression compared to the previous destructor injection. Since the previous injection used data-flow analysis, it could prove that there was no unstructured scope exit (the `f()` call in line 9, in this case) where `a` or `b` were alive, so no `finally` was necessary.

At the moment, `mirgen` can only use a control-flow analysis, so a `finally` has to be used now. The performance impact is generally minimal, however, and future improvements to `mirgen` could allow the `finally` to omitted again.

zerbina:
Hm, there looks to be some issue with destructor injection / move analysis somewhere. I'll investigate.

zerbina:
Okay, I was completely off with my initial guess on why building the release binaries on Windows failed. The problem is simply due to `build.bat` not using a response file. Adding a new module brought the length of the link command above the command-line length limit, thus cutting off part of the command and causing linking to fail.

compiler/mir/mirgen.nim
```diff
@@ -149,22 +146,24 @@ type
...
 
     sp: SourceProvider
 
-    numLabels: int ## provides the ID to use for the next label
     scopeDepth: int ## the current amount of scope nesting
     inLoop: int
       ## > 0 if the current statement/expression is part of a loop
+    injectDestructors: bool
+      ## whether injection of destroy operations is enabled
+    isDisabled: bool
```
```suggestion
    unreachable: bool
```

Possibly a better name, nbd either way.

compiler/mir/mirgen.nim
```diff
@@ -447,6 +460,24 @@ template buildDefectMagicCall(c: var TCtx, m: TMagic, t: TypeId,
...
+  let label = c.builder.allocLabel()
+  c.buildStmt mnkIf:
+    cond
+    c.add labelNode(label)
+  body
+  c.buildStmt mnkEndStruct:
+    c.add labelNode(label)
+
+proc register(c: var TCtx, loc: Value) =
+  ## If `val` has a destructor and destroy injection is enabled for the
```
saem: `val`? Should this be something more like "if the value in the `loc` has a destructor ..."?

compiler/mir/mirgen.nim
```diff
@@ -1921,7 +2096,20 @@ proc genx(c: var TCtx, e: PMirExpr, i: int) =
...
-    recurse()
+    if c.isDisabled:
+      # don't translate the expression if it's unreachable. The callsite still
+      # expects some expression, and thus a default-intialized temporary is
+      # used. Do note that the code is unreachable, and the assignment is thus
+      # never evaluated -- it's just there to uphold the callsite's syntactic
+      # expectations
+      # XXX: ideally, non-terminating statement list expressions should
+      #      have the trailing expression cut off and be turned into
+      #      ``nkStmtList`` nodes at an earlier stage
```
saem, line 2107: yeah, at the very latest `transf`

saem (approved):
Left a few very minor comments, overall this is a really exciting step towards dropping CGIR.

compiler/mir/mirgen.nim
zerbina: Describing the context rather than the consequence seems like the better choice, yeah.

compiler/mir/mirgen.nim
zerbina: Yep, good catch.

compiler/mir/mirgen_blocks.nim
zerbina: Implemented as part of #1329.

PR #1283, opened 2024-04-23
sem: proper error propagation for set construction

zerbina:
## Summary

Refactor/fix set construction analysis such that:
* `nkRange` nodes are properly analyzed after macro expansion
* errors within set constructions are propagated properly

## Details

Only `semSetConstr` is modified:
* use `shallowCopy` instead of creating a new node (no sequence
  resizing, keeps the persistent node flags)
* always re-analyze `nkRange` nodes (they were previously never typed)
* guard against ill-formed `nkRange` AST (could previously crash the
  compiler)
* don't modify the input AST
* replace `localReport` with using `nkError`
* propagate `nkError`


saem (approved):
This is a nice improvement.

PR #1284, opened 2024-04-23
sem: clean up `case` analysis and fix edge cases

zerbina:
## Summary

Clean up, improve, and fix `case` analysis. Errors are properly
propagated, input AST is no longer modified, and multiple bugs are
fixed:
* `nkRange` nodes in `nkOfBranch` AST in macro output is properly
  analyzed
* converters are correctly taken into account for branch labels
  (previously resulted in a compiler crash)
* empty set, array, and seq values in otherwise non-empty `of` branches
  no longer result in spurious "not all cases are covered" errors

## Details

Both the `case` statements and `case`-object declarations are affected.
For the range syntax analysis (`semBranchRange`):
* evaluating the expression happens *after* fitting, so that converter
  calls are accounted for (they previously stayed, crashing the
  compiler later on)
* the `localReport`s are replaced with diagnostics; two new diagnostics
  corresponding to the previously used reports are added
* errors are wrapped properly
* "cannot be string range" errors now take precedence over
  "empty range" errors
* the operand type is passed directly as a parameter, instead of
  indirectly via the case statement AST

For the `of` branch analysis (`semCaseBranch`):
* `semConstExpr` (which uses `localReport`) is replaced with
  `semExprWithType` + `evalConstExpr`
* the input AST is no longer modified
* empty set, array, and seq values no longer, erroneously, count
  towards the number of values that need to be covered
* `nkRange` nodes are always (re-)analyzed
* only the selector's type is passed as the parameter, checking for
  duplicates moves to the callsite

The `case` statement analysis (`semCase`) is changed to not modify
input AST. For wrapping the `nkCaseStmt` in an error, the AST is
scanned for errors afterwards.

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -1675,7 +1675,6 @@ proc semCase(c: PContext, n: PNode; flags: TExprFlags): PNode =
         suggestEnum(c, x, caseTyp)
     case x.kind
     of nkOfBranch:
-      checkMinSonsLen(x, 2, c.config)
       let branch = semCaseBranch(c, n[0].typ, x, covered)
       # XXX: errors need to be propagated
       for e in walkErrors(c.config, branch):
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -1641,16 +1641,16 @@ proc semFor(c: PContext, n: PNode; flags: TExprFlags): PNode =
   closeScope(c)
 
 proc semCase(c: PContext, n: PNode; flags: TExprFlags): PNode =
-  result = n
   checkMinSonsLen(n, 2, c.config)
+  result = copyNodeWithKids(n)
   openScope(c)
-  pushCaseContext(c, n)
-  n[0] = semExprWithType(c, n[0])
+  result[0] = semExprWithType(c, n[0]) # selector operand
+  let selector = result[0].typ
   var chckCovered = false
   var covered: Int128 = toInt128(0)
   var typ = commonTypeBegin
   var hasElse = false
-  let caseTyp = skipTypes(n[0].typ, abstractVar-{tyTypeDesc})
+  let caseTyp = skipTypes(selector, abstractInst-{tyTypeDesc})
   const shouldChckCovered = {tyInt..tyInt64, tyChar, tyEnum, tyUInt..tyUInt64, tyBool}
   case caseTyp.kind
   of shouldChckCovered:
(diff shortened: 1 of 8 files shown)


compiler/sem/semstmts.nim
```diff
@@ -1659,76 +1659,90 @@ proc semCase(c: PContext, n: PNode; flags: TExprFlags): PNode =
...
-      typ = commonType(c, typ, x[0])
-      if (chckCovered and covered == toCover(c, n[0].typ)) or hasElse:
+      let branch = shallowCopy(x)
+      branch[0] = semExprBranchScope(c, x[0])
+      result[i] = branch
+      if (chckCovered and covered == toCover(c, selector)) or hasElse:
         localReport(c.config, x.info, SemReport(kind: rsemUnreachableElse))
       hasElse = true
       chckCovered = false
     else:
```
saem, line 1704:
```suggestion
    of nkError:
      result[i] = x
    else:
```

Do we want to do this instead so `nkError` is nested and then wrapped and reported correctly later on?

compiler/sem/semtypes.nim
```diff
@@ -747,80 +749,97 @@ proc checkForOverlap(c: PContext, t: PNode, currentEx, branchIndex: int) =
...
+          x = evalConstExpr(c, x)
+        to.add x
+        inc covered
+
+  checkMinSonsLen(branch, 2, c.config)
+  # not a one-to-one mapping between the input and output AST
+  result = copyNode(branch)
+
+  for i in 0..<branch.len-1:
+    let b = branch[i]
```
saem:
```suggestion
  for i, b in branch.pairs:
```

That should be the same.

tests/lang_stmts/casestmt/tbranch_with_empty_set.nim
```diff
@@ -0,0 +1,22 @@
...
+    counted towards the number of required labels, leading to errors being
+    reported for valid case statements.
+  '''
+  action: compile
+"""
+
+var x: bool
+
+case x
+of true, []: # would cause "not all cases are covered"
```
saem, line 13: Huh, I didn't know an array of values was legal.

tests/lang_stmts/casestmt/tlabel_considers_converter.nim
```diff
@@ -0,0 +1,30 @@
...
+"""
+
+converter f2i(x: float): int8 =
+  int8(x)
+
+proc test(x: int8): int =
+  case x
+  of 1.0:
+    1
+  of 2.2 .. 3.3: # with range syntax: 2 .. 3
```
saem, line 14: I know this is an artifact of allowing for converters, it's still weird. The new behaviour is better, of course.

saem (approved):
Found a potential issue where an error input will turn into illegal syntax rather than reporting an error, otherwise, everything else is minor/stylistic.

The change looks good to me. 😃

compiler/sem/semstmts.nim
saem: Thought about this again, and I think if we have an error then it should be illegal syntax unless it's already an illegal syntax error (avoid repeated wrapping). As for better reporting that should be handled in the illegal syntax error message generation code.

Ultimately, I think this means my suggestion is also not correct as it needs to check for the error kind in the node and then let it through if illegal syntax otherwise wrap it.

Thoughts?

zerbina: The only two reasons an `nkError` node can appear here are:
1. the error wrapping in `semCase` is faulty
2. a macro transplanted an error it somehow got hold of into the AST

Number 1 would be a compiler bug, and number 2 could also be considered one (macros shouldn't be able to get hold of `nkError` nodes, at least at the moment), so I think reporting an illegal syntax error (which raises an exception) is a reasonable thing to happen.

Alternatively, `internalError` could be used, though it'd be inconsistent with all other analysis (e.g., `semIf`, `semTry`, etc.), where `nkError`s in the input are treated as illegal syntax.

compiler/sem/semtypes.nim
zerbina: Not quite, since the last slot needs to be skipped. I've now replaced the manual arithmetic with usage of the more robust `branchLabels` iterator.

tests/lang_stmts/casestmt/tbranch_with_empty_set.nim
zerbina: Yeah, it's not really documented anywhere, I think. `semCaseBranch` uses type-based dispatching now, so it's a least made somewhat more obvious in the code.

compiler/sem/semstmts.nim
saem: Makes sense, I suspect the second option is quite possible, but the current approach is reasonable.

compiler/sem/semtypes.nim
saem: Ah, I totally zoned on the `<` and the `-1`, good call.

PR #1286, opened 2024-04-27
mirgen: better code generation for `high`

zerbina:
## Summary

Omit unnecessary to-slice conversions for `high(x)` calls where `x` is
a `seq` value, resulting in a small compiler speed up. All backends are
affected.

## Details

There's no dedicated `high` overload for `seq`, so the `openArray`
overload is picked for `seq` arguments. The implicit argument
conversion was previously translated as is, resulting in an extra
temporary and conversion operation, both which are unnecessary, since
all three code generators support `high` for `tySequence` operands.

While an optimizing C compiler is able to eliminate the conversion,
doing it on the NimSkull side still benefits debug builds and the
JavaScript and VM target, as well as reduce the workload for the MIR
passes and code generators.

### Tests

* add a new test category for optimization-related tests
* add a test to ensure there's no conversion for `high(seq)` calls

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -940,6 +940,16 @@ proc genMagic(c: var TCtx, n: PNode; m: TMagic) =
         c.userOptions = {}
         genx(c, n[1])
         c.userOptions = orig
+  of mHigh:
+    # custom translation in order to skip both explicit and implicit to-slice
+    # conversions; those are unnecessary
+    c.buildMagicCall mHigh, rtyp:
+      var arg = n[1]
+      while arg.kind in {nkConv, nkHiddenStdConv, nkHiddenSubConv} and
+            classifyBackendView(arg.typ) == bvcSequence:
+        arg = arg[^1]
+
+      c.emitOperandTree arg, sink=false
 
   # arithmetic operations:
   of mAddI, mSubI, mMulI, mDivI, mModI, mPred, mSucc:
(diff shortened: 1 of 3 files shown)


zerbina:
For the following snippet:
```nim
proc test(x: seq[int]): int =
  result = high(x)
```
the previous `--expandArc` output was:
```
scope:
  def _2: openArray[int] = toOpenArray x
  result = mHigh(arg _2)
```
now it's:
```
scope:
  result = mHigh(arg x)
```

---

Regarding the statement about optimizing C compilers, the following C code, which is roughly equivalent to how the C code produced by the compiler looks like:
```c
struct Seq {
    int len;
    struct {int cap; int data[];} *p;
};

struct OpenArray {
    int* p;
    int len;
};

int high(struct Seq s) {
    struct OpenArray arr;
    arr.p = (s.p != 0) ? &s.p->data[0] : 0;
    arr.len = s.len;
    return arr.len - 1;
}
```
[compiles down to just the subtraction](https://godbolt.org/z/azbE3Pxjz).

saem (commented):
Something to consider, not a blocker: did you want to copy the example into an `--expandArc` test in order to ensure we preserve the optimization?

zerbina:
> Something to consider, not a blocker: did you want to copy the example into an `--expandArc` test in order to ensure we preserve the optimization?

That's a good call, thanks -- I'll add a test.

PR #1288, opened 2024-04-28
cgen: better code generation for `.noreturn` calls

zerbina:
## Summary

Improve C code generation for `.noreturn` calls, by omitting the
unnecessary error flag check for them.

## Details

If a procedure only returns due to exceptional unwinding, the
`if(NIM_UNLIKELY(*nimErr_))` guard guarding the unwinding is
unnecessary, since it's guaranteed that the error flag is set (all
exits of the procedure are due to exceptional unwinding).

Unconditionally jumping to the error handling target makes it clear
to the C compiler that control-flow doesn't continue normally after
the call

### Other improvements

In addition, the `canRaiseConservative` usage is unnecessary too, since
a `cnkCheckedCall` already signals that the call can raise. The
procedure has no more users and is thus removed.

Finally, for compilers that support it, a call to a `.noreturn`
procedure that doesn't raise is now followed by an `__assume(0)`
statement, communicating to the C compiler that control-flow never
returns to the caller.

### DIFF
--- a/compiler/backend/ccgcalls.nim
+++ b/compiler/backend/ccgcalls.nim
@@ -76,18 +76,30 @@ proc isHarmlessStore(p: BProc; ri: CgNode, d: TLoc): bool =
 
 proc exitCall(p: BProc, call: CgNode) =
   ## Emits the exceptional control-flow related post-call logic.
+  let isNoReturn = call[0].kind == cnkProc and
+                   sfNoReturn in p.env[call[0].prc].flags
   if call.kind == cnkCheckedCall:
-    if nimErrorFlagDisabled in p.flags:
-      if call[0].kind == cnkProc and sfNoReturn in p.env[call[0].prc].flags and
-         canRaiseConservative(p.env, call[0]):
-        # when using goto-exceptions, noreturn doesn't map to "doesn't return"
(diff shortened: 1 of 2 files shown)


saem (approved):

zerbina:
For an example of how code generation improves, consider the following:
```nim
proc test(num: int) =
  let val =
    if num == 2: 1
    elif num == 4: 2
    else: unreachable() # call to a `.noreturn` procedure

  if val >= 1:
    echo "here"
```
The produced C code was roughly equivalent to the following:
```c
int* errorFlag;
void unreachable();

void test(int num) {
    int val = 0;
    if (num == 2) {
        val = 1;
    } else if(num == 4) {
        val = 2;
    } else {
        unreachable();
        if (*errorFlag) goto BeforeRet_;
    }
    if (num >= 1) {
        printf("here");
    }
    BeforeRet_:
    return;
}
```
([optimized assembler](https://godbolt.org/z/fhzPa7n6z))

After the PR, the `if(*errorFlag)` guard is gone. As can be seen [here](https://godbolt.org/z/xvsdG7a6b), an optimizing C compiler (`gcc`, in this case) is then able omit the code path for the (impossible) `*errorFlag == 0` case.

PR #1289, opened 2024-04-29
system: mark `raiseX` runtime procedures as noreturn

zerbina:
## Summary

Mark the `raiseX` runtime procedures used by the C backend as
`.noreturn`, to allow for generation of more efficient code.

## Details

For less code duplication, the `.noinline`, `.noreturn`, and
`.compilerproc` pragmas are combined into the custom `.errorPrc`
pragma.

At the moment, code generation is not affected, since `cgen` itself
emits calls to the `raiseX` procedures, but this is going to change
once MIR passes perform lowering of the runtime checks.

### DIFF
--- a/lib/system/chcks.nim
+++ b/lib/system/chcks.nim
@@ -10,41 +10,45 @@
 # Implementation of some runtime checks.
 include system/indexerrors
 
-proc raiseRangeError(val: BiggestInt) {.compilerproc, noinline.} =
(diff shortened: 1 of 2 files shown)


saem (approved):
Looks good.

I added the `refactor` label, as there is barely any behaviour change (would have to do crazy things to observe any once MIR performs the lowering). Feel free to remove the label if it doesn't make sense to you.

PR #1290, opened 2024-04-30
lower `mChckRange` magic with MIR pass

zerbina:
## Summary

Lower the `mChckRange` magic with a MIR pass instead of as part of C
code generation, shrinking down `cgen`. A bug where no ranger errors
were reported for some `uint`/`uint64`-to-signed-integer conversions is
fixed too.

## Details

The lowering pass is implemented in the new `rtchecks` module, which is
also going to host the lowerings for the other run-time check magics.

The pass turns `chckRange(x, a, b)` into the MIR equivalent of
`(if x < a or b < x: raiseRangeError(...); Target(x))`.

### Float checks

For to-`float` range checks, the to-be-converted value is first
converted to the output type, and then it's compared against the
specified lower and upper bound.

### Integer checks

For to-integer range checks, the lower and upper bound integer values
are first converted to the source type, but only of they're within the
source type's range. For example, for
`chckRange(x, 0'i16, 32_768'i16)`, where `x` is an `int8`, the
`32_768 < x` is unnecessary, since it's guaranteed to be false. 

### Bugfix

For `uint`-to-`int` conversion range checks, the lower bound was
previously not considered, resulting in conversions such as
`range[1..5](x)` not triggering a range error when `x` < 1. The MIR
pass addresses this.

A regression test is added to `trangechecks.nim`.

### Other

`mirgen` incorrectly assigned the `range`/`enum` type to the
intermediate temporary when lowering `dec`/`inc` into a `-`/`+` + range
check + assignment. Since the input and output type for the range check
are then the same, the `mChckRange` lowering pass would erroneously
optimize the call away.

To address this, `mirgen` now assigns the `range`/`enum` type's
underlying integer type to the temporary.


saem (approved):
`numchecks` might be slightly better, as it's not going to cover things like ref, pointers, etc.

zerbina:
For context, the module is going to cover the lowering of `mChckNaN`, `mChckObj`, `mChckObj` `mChckField`, `mChckIndex`, and `mChckBounds`, as well as integer/float overflow checks. `numchecks` would cover some of them, but I think it doesn't apply well to field, index, and object checks. 

The reason I don't like the current name `checks` is that the name's not very descriptive. Going just by the name, it could also contain some checks/verification for MIR code.

saem:
> For context, the module is going to cover the lowering of `mChckNaN`, `mChckObj`, `mChckObj` `mChckField`, `mChckIndex`, and `mChckBounds`, as well as integer/float overflow checks. `numchecks` would cover some of them, but I think it doesn't apply well to field, index, and object checks.

Ah, I forgot to consider field, index, and object.

> The reason I don't like the current name `checks` is that the name's not very descriptive. Going just by the name, it could also contain some checks/verification for MIR code.

That's helpful context, the other thing I thought of was `runtime_checks`, it's more indicative of the context (when) within which they're run. Related alternatives would be `rtchecks` or `rt_checks`.

zerbina:
I think `rtchecks` sounds good -- I'll update the name. Thank you!

PR #1293, opened 2024-04-30
fix: `move` not resetting source location

zerbina:
## Summary

Use a MIR pass to lower the `move` and `wasMoved` magic calls into
assignments, fixing `move` not always resetting the source location for
both the VM and JS backends. The C backend was not affected.

## Details

A `x = move(y)` call is now lowered into the following MIR:
```
x = move y
y = default()
```
and a `wasMoved(x)` call into:
```
x = default()
```
which is equivalent in behaviour to the code previously produced by
`cgen` -- using a shared MIR pass for the lowering ensures that the
behaviour is consistent across all backends.

The implementation of the two magics is removed from each code
generator, and a test for making sure `move` works correctly is added.

### Previous behaviour

For the VM backend, the source location was never reset on `move`,
while for the JS backend, the source location was reset with
`genericReset`, which left record-like locations empty (reading the
fields then yield `undefined`) and ignores primitive types like int or
float.

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -1502,23 +1502,6 @@ proc skipAddr(n: CgNode): CgNode =
   if n.kind == cnkHiddenAddr: n.operand
   else:                       n
 
-proc genWasMoved(p: BProc; n: CgNode) =
-  var a: TLoc
-  let n1 = n[1].skipAddr
-  if true:
-    initLocExpr(p, n1, a, {lfWantLvalue})
-    resetLoc(p, a)
-    #linefmt(p, cpsStmts, "#nimZeroMem((void*)$1, sizeof($2));$n",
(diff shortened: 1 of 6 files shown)


zerbina:
Scanning the body of every procedure for magics has overhead, and it'd be better to only run the pass if the relevant magics are actually used within the body.

My plan is store a `set[TMagic]` in `MirBody`, for keeping track of all magics currently present in the body. A pass for lowering, say, `mWasMoved` then only has to run when the magic is present in the set.

---

As for `mMove`, it ideally should be turned into `def _1 = move x; wasMoved(name x)` *before* the move analyzer runs. At the moment, the move analyzer doesn't know that a `move(x)` call actually moves the value out of the location, preventing some optimizations.

For example, for:
```nim
var x = @[...]
var y = move(x)
```
the compiler currently injects a destructor for both `x` and `y`, even though `x` is guaranteed to not require destruction (the value was moved on all paths).

PR #1294, opened 2024-04-30
refactor(pragmas): improve `processCompile`

zerbina:
## Summary

Refactor the `pragmas.processCompile` procedure to not mutate input
AST, and add a specification test for the undocumented pattern-matching
version of `.compile`.

## Details

For `pragmas.processCompile`:
* don't mutate input AST
* use proper error propagation (i.e., keep the shape of the AST intact)
* test that the specified path is relative before trying to interpret
  it as a relative path (less work)

The `.compile` pragma has a pattern-matching version, but this was
not documented anywhere. A specification test derived from the current
behaviour is added for it.

### DIFF
--- a/compiler/sem/pragmas.nim
+++ b/compiler/sem/pragmas.nim
@@ -709,9 +709,9 @@ proc relativeFile(c: PContext; name: string, info: TLineInfo;
 
 proc processCompile(c: PContext, n: PNode): PNode =
   ## compile pragma
-  ## produces (mutates) `n`, which must be a callable, analysing its arg, or returning
-  ## `n` wrapped in an error.
-  result = n
+  ## Produces the pragma with all arguments evaluated, or returns an error.
+  ## If the pragma is well-formed, the external files to compile are added to
+  ## the build.
   proc docompile(c: PContext; it: PNode; src, dest: AbsoluteFile; customArgs: string) =
     var cf = Cfile(nimname: splitFile(src).name,
                    cname: src, obj: dest, flags: {CfileFlag.External},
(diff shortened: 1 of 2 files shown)


zerbina:
There's an idea that occurred to me while working on this refactor. Instead of having `pragmas` dispatch to the correct handling for built-in pragmas based on the identifier, most built-in pragmas could be turned into proper symbols, by introducing a new `pragma` pragma.

It would work as follows:
```nim
# in system.nim
proc compile*(name: static[string]; args: static[string] = "") {.pragma, magic: "Compile".}
proc compile*(pair: static[tuple[src, dest: string]]) {.pragma, magic: "Compile".}
```

In essence, a procedure marked with `pragma` is added special pragma-only namespace (that is, they're only visible in pragma contexts). Analysing a pragma then amounts to a call to `semOverloadedCall`, plus subsequent evaluation of the magic procedure.

This would have a number of benefits:
* the built-in pragmas become "real" entities with proper symbols (easier inspection by both the compiler and macros)
* they can be documented directly within the code
* auto-completion and hover-over-symbol becomes available for built-in pragmas
* the interface (operand types, arity, etc.) becomes explicit
* built-in pragmas become can be overloaded with user pragmas of the same name
* the implementation could leverage `sigmatch`/`semcall`, replacing all the manual operand analysis in `pragmas.nim` (making it a lot less error-prone)

Outside of the `system` module, the `pragma` pragma would only be usable on `template`s and `macro`s -- templates/macros without the `pragma` pragma would no longer be considered in a pragma context.

saem:
> Outside of the `system` module, the `pragma` pragma would only be usable on `template`s and `macro`s -- templates/macros without the `pragma` pragma would no longer be considered in a pragma context.

Would this mean that a pragma macro, like those in say the CPS library, would no longer transform procedures unless they were updated to include the `pragma` pragma?

zerbina:
> Would this mean that a pragma macro, like those in say the CPS library, would no longer transform procedures unless they were updated to include the `pragma` pragma?

That'd be the case, yeah. In my opinion, using a separate namespace for pragma routines makes sense, and I think macros/templates meant to be used as pragma are generally not meant for other use (and vice versa), but it's not something I'm set on.

saem:
My instincts suggest that pure metadata style annotations should be signaled differently than the transformational variety. So the double dipping with `pragma` isn't ideal.

Even with that concern in mind, I still think the idea is very close to ideal. It's got all the right properties, it composes well, reduces special casing, incrementally adds a tiny bit of capability but gets a lot of mileage out of it.

Maybe if we used `pure` in conjunction with `pragma` to signal metadata only, or introduced a dedicated one like `annotation` or `metadata`, then treating `pragma` as the transformation variety would be cleaner.

saem (approved):

zerbina:
Hm, yeah, separating metadata/symbol-modification from transformation seems like a good idea. Statement and/or expression pragmas (such as `compile` or `passc`) fit neither, so that'd require some additional thought.

PR #1295, opened 2024-04-30
support multiple external C files having the same name

### LINKED ISSUE #1291: linking error when using `{.compile.}` pragma on .c files with same name but located on different paths
### Example

```c
// bar.c
int bar_simple(int a, int b) {
  return a + b;
}
```
```c
// foo/bar.c
int bar_foo(int a, int b) {
  return a * b;
}
```
```nim
{.compile: "bar.c".}
{.compile: "foo/bar.c".}

proc bar_simple(a, b: cint): cint {.importc.}
proc bar_foo(a, b: cint): cint {.importc.}

proc main() =
  echo bar_simple(8, 16)
  echo bar_foo(32, 64)

when isMainModule:
  main()
```

### Actual Output

```
/usr/bin/ld: ~/.cache/nimskull/main_d/bar.c.o: in function `bar_foo':
bar.c:(.text+0x0): multiple definition of `bar_foo'; ~/.cache/nimskull/main_d/bar.c.o:bar.c:(.text+0x0): first defined here
/usr/bin/ld: ~/.cache/nimskull/main_d/@mmain.nim.c.o: in function `main__main_7':
@mmain.nim.c:(.text+0x12b): undefined reference to `bar_simple'
collect2: error: ld returned 1 exit status
```

### Expected Output

```
24
2048
```

### Possible Solution

Nimskull doesn't concatenate complete path into compiled `.o` files from `{.compile.}` pragma as does with `.nim.c.o` compiled from .nim files
- current behavior: `bar.c` and `foo/bar.c` compile to `bar.c.o` so there is a conflict
- expected behavior: `bar.c` and `foo/bar.c` compile to `@mbar.c.o` and `@mfoo@sbar.c.o` respectively

### Additional Information

compiles on Nim

zerbina:
## Summary

Adding multiple C files sharing the same name to the build via the
`compile` pragma is now supported.

Closes https://github.com/nim-works/nimskull/issues/1291.

## Details

For the non-pattern `compile` pragma, the C file name is turned into a
unique object file name via `withPackageName`, mirroring what the C
backend does for the C files it generates.

To prevent the object files for the external C files from colliding
with the NimSkull module object files, they're placed under the
`external` sub-directory within the cache directory.

### DIFF
--- a/compiler/backend/extccomp.nim
+++ b/compiler/backend/extccomp.nim
@@ -393,6 +393,11 @@ proc toObjFile*(conf: ConfigRef; filename: AbsoluteFile): AbsoluteFile =
   # Object file for compilation
   result = AbsoluteFile(filename.string & "." & CC[conf.cCompiler].objExt)
 
+proc externalObjFile*(conf: ConfigRef; cfile: AbsoluteFile): AbsoluteFile =
+  ## Returns the path of the object file to output for the external C file
+  ## `cfile`.
+  result = toObjFile(conf, completeGeneratedExtFilePath(conf, cfile))
+
 proc addFileToCompile*(conf: ConfigRef; cf: Cfile) =
   conf.toCompile.add(cf)
 
--- a/compiler/front/options.nim
+++ b/compiler/front/options.nim
@@ -1300,6 +1300,19 @@ proc completeGeneratedFilePath*(conf: ConfigRef; f: AbsoluteFile,
   result = subdir / RelativeFile f.string.splitPath.tail
   #echo "completeGeneratedFilePath(", f, ") = ", result
 
+proc completeGeneratedExtFilePath*(conf: ConfigRef, f: AbsoluteFile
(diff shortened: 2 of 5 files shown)


zerbina:
Hm, there's a problem I didn't account for: an external C file with the same name as a `.nim` using a `.nim.c` extension leads to a name collision of the object files.

One solution would be to simply dismiss this being a problem, a second one to mangle the name in a way that no collisions are possible. I think 'll give the second option a try.

**Edit:** There's an even simpler, yet more robust solution. Instead of putting the object files for external C files in the same directory as the normal object files, they can be placed in a sub-directory.

saem:
> **Edit:** There's an even simpler, yet more robust solution. Instead of putting the object files for external C files in the same directory as the normal object files, they can be placed in a sub-directory.

Having it separate makes a lot of sense, they're like second class compilation units. ;)

PR #1298, opened 2024-05-04
fix(sem): issues with method-call syntax in templates

### LINKED ISSUE #1292: dot function call syntax inside a template can't lookup symbols of an imported module when the template is called from another module
### Example

```nim
# bar.nim
type
  BarNumber = ref object
    a, b: cint

proc newBar*(a, b: cint): BarNumber =
  new result
  # Create New Foo
  result.a = a
  result.b = b

proc sum*(foo: BarNumber): cint =
  foo.a + foo.b

proc mul*(foo: BarNumber): cint =
  foo.a * foo.b

```
```nim
# foo.nim
import bar

template foobar*(a, b: cint): cint =
  let bar = newBar(a, b)
  # Dot Function Call Syntax
  # can't lookup bar.nim symbols when this
  # template is called on another module
  var result: cint
  result += bar.sum()
  result += bar.mul()
  result

```
```nim
# main.nim
import foo

proc main() =
  echo foobar(10, 12)

when isMainModule:
  main()
```
### Actual Output

```
src/foo.nim(7, 17) Error: undeclared field: 'sum' for type bar.BarNumber [type declared in src/bar.nim(2, 3)]
```

### Expected Output

```
142
```

### Additional Information
this error only occurs when using dot function call syntax, it compiles when using `sum(bar)` instead `bar.sum()` for example
also compiles when `bar.nim` is also imported on `main.nim`
```nim
# foo.nim
import bar

template foobar*(a, b: cint): cint =
  let bar = newBar(a, b)
  # Using Simple Call Syntax
  # can lookup symbols of bar.nim
  # using this template on another module
  var result: cint
  result += sum(bar)
  result += mul(bar)
  result
```

### References
this compiles on Nim

zerbina:
## Summary

Fix two issues with the method-call syntax in non-dirty templates:
* non-overloaded symbols weren't bound early, falling back to late
  binding (https://github.com/nim-works/nimskull/issues/1292)
* gensym'ed symbols couldn't be used as the callee

### `semTemplBody` restructuring

Restructure `semTemplBody` so that the `TemplCtx.noGenSym` field
becomes unnecessary:
* the name part of `nkExprColonExpr`/`nkExprEqExpr` can only bind to
  template parameters, so `qualifiedLookUp` is used directly
* resolving an identifier to a symbol or symbol choice is moved to
  `templBindSym`
* for dot expressions (`a.b`), handling symbol binding for the `b` part
  is done manually via symbol lookup + `templBindSym`

### Missing early binding

* if `b` in `a.b` is a non-overloaded symbol, wrap it in a symbol-
  choice, to prevent `dotTransformation` from discarding it again
* for backwards-compatibility an *open* symbol choice is used in this
  case, not a *closed* one (non-overloaded symbols are closed by
  default)
* a `knownIssue` test is added for the incorrect symbol binding in dot
  expressions (`tclosed_symbol_with_method_call`)

### Bind `gensym` symbols

* if `b` in `a.b` resolves to a gensym, the symbols is now bound
* only routine, type, and generic parameter symbols are bound for `b`
  in `a.b`. This prevents, incorrect symbol binding in cases such as
  `var x = 1; a.x = 2`
* to still allow `a.b` resolving to a field access even if a gensym was
  bound to `b` (and later turned into `b'gensym` identifier),
  `builtInFieldAccess` and `propertyWriteAccess` strip the `'gensym`
  suffix from the identifier before trying to use it
* so that rendering of custom string literals (which reach `sem` as
  `(DotExpr (...) (Ident 'suffix))` stays the same,
  `renderer.isCustomLit` has to account for symbol-choices in the
  second slot

A bug with `symChoice` uncovered by the other changes is fixed: gensyms
were added to the symbol choice, even if `isField` (now rename to the
more accurately named `noGenSyms`) was true.

### Misc

* various comments documenting issues with how the symbol binding works
  are added

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -147,12 +147,12 @@ const
     ## not be performed for arguments to these magics.
 
 proc getPIdent*(a: PNode): PIdent {.inline.} =
-  ## Returns underlying `PIdent` for `{nkSym, nkIdent}`, or `nil`.
-  # xxx consider whether also returning the 1st ident for {nkOpenSymChoice, nkClosedSymChoice}
-  # which may simplify code.
+  ## Returns underlying `PIdent` for `{nkSym, nkIdent, nkOpenSymChoice,
+  ## nkClosedSymChoice}`, or `nil`.
   case a.kind
   of nkSym: a.sym.name
   of nkIdent: a.ident
+  of nkSymChoices: a[0].sym.name
   else: nil
 
 proc getIdentLineInfo*(n: PNode): TLineInfo =
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -147,12 +147,12 @@ const
     ## not be performed for arguments to these magics.
 
 proc getPIdent*(a: PNode): PIdent {.inline.} =
-  ## Returns underlying `PIdent` for `{nkSym, nkIdent, nkOpenSymChoice,
-  ## nkClosedSymChoice}`, or `nil`.
+  ## Returns underlying `PIdent` for `{nkSym, nkIdent}`, or `nil`.
+  # xxx consider whether also returning the 1st ident for {nkOpenSymChoice, nkClosedSymChoice}
+  # which may simplify code.
   case a.kind
   of nkSym: a.sym.name
   of nkIdent: a.ident
-  of nkSymChoices: a[0].sym.name
   else: nil
 
 proc getIdentLineInfo*(n: PNode): TLineInfo =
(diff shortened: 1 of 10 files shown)


saem (approved):
Sorry it took me a bit. I was approaching the review by looking at the changes, rather than reasoning about it from first principles, I ended going around in circles for a while. 

The change look good including the notes regarding future fixes, which should further consolidate half baked analysis into a few purpose-built procedures.

saem:
/merge

BTW, I made some very minor updates to the PR description for typos.

PR #1299, opened 2024-05-05
fix(testament): batching being ignored for megatest

### LINKED ISSUE #1037: testament: `cat megatest` doesn't consider the selected batch
The `megatest` category of testament ignores enabled batching, leading to all test suite runs
executing all joinable tests rather than only those from the active batch.

### Possible Solution

The problem is with: https://github.com/nim-works/nimskull/blob/de8025ba317d737c14d5c14789ddc0ca53f007f7/testament/categories.nim#L364

Here, `computeEarly` is always called with `inCurrentBatch = true`, meaning that the batch
feature is effectively ignored. A straightforward solution would be to set up a `TTest` from the
`spec` first, and then pass the value of the test's `inCurrentBatch` property to `computeEarly`. 

Alternatively, batching could be implemented as a filter over the files, removing the need
for `inCurrentBatch`.

### Additional Information

This bug is a regression introduced by #770.

zerbina:
## Summary

Fix batching being ignored for the megatest. Only tests part of the
current batch are now joined into the megatest.

Fixes https://github.com/nim-works/nimskull/issues/1037

## Details

In order to not duplicate the computation of whether the test is in the
current batch, a pseudo `TTest` instance is created that's then passed
to `computeEarly`.

### DIFF
--- a/testament/categories.nim
+++ b/testament/categories.nim
@@ -360,9 +360,12 @@ proc runJoinedTest(r: var TResults, targets: set[TTarget], testsDir, options: st
       for file in walkDirRec(testsDir / cat):
         if isTestFile(file):
           try:
-            let spec = parseSpec(file, cat.Category.defaultTargets, nativeTarget())
-            if isJoinableSpec(spec, targets, computeEarly(spec, true)):
-              specs.add spec
+            # setup a pseudo test instance for the purpose of computing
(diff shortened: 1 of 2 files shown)


saem (approved):

saem:
/merge

Yup, it seems faster.

PR #1302, opened 2024-05-08
threads: improve interface and fix bugs

zerbina:
## Summary

Make `Thread` a proper thread handle, fixing multiple bugs and making
the interface more robust:
* a thread is now officially allowed to outlive the `Thread` instance
  (which would previously lead to crashes)
* a `Thread` handle for a running thread can now be safely moved around
* there's now a `createThread` overload that returns a `Thread`
  instance
* `Thread.finished` having a race condition

## Details

The management-data for a thread was previously part of the `Thread`
type itself, with a pointer to the `Thread` instance passed to the
internal thread procedure, requiring the `Thread` instance for a
running thread to always stay at the same memory location.

As a consequence of this design, moving the `Thread` instance, or the
`Thread` instance going out of scope before the thread is finished,
resulted in undefined behaviour.

To address this problem, the management-data is made part of a heap-
allocated structure (`ThreadCore`), which is then shared between the
handle and spawned thread. The type uses manual reference counting in
order to support detached threads (threads that outlive their handle).

Since a `Thread` instance can now be moved around in memory, the base
`createThread` procedure can now return a `Thread`, which is the better
interface, as it prevents leaks caused by forgetting to destroy the
`Thread` instance first. For backwards compatibility, an in-place
`createThread` overload is added.

### Tests

Two new tests are added:
* a test for ensuring that the thread's internal data is freed
  correctly
* a test for ensuring that the thread parameter is destroyed

The pre-existing `tthread_destroy_before_finish.nim` test had a few
issues, which are now fixed:
* it had no proper description
* it wasn't guaranteed to actually fail (in the edge-case that the
  spawned thread finishes before the `reset` call)
* access to the `ok` global was not atomic (resulting in a race
  condition)
* it unnecessarily imported `std/os`

### DIFF
--- a/lib/system/threads.nim
+++ b/lib/system/threads.nim
@@ -61,16 +61,42 @@ const
 # use ``stdcall`` since it is mapped to ``noconv`` on UNIX anyway.
 
 type
-  Thread*[TArg] = object
-    core: PGcThread
-    sys: SysThread
+  ThreadCore[TArg] = object
+    ## The internal managnement data associated with a thread. Allocated
+    ## by the spawning thread, and - initially - owned by both the spawning
+    ## and spawned thread (shared ownership). If the spawned thread finishes
+    ## before the `Thread <#Thread>` handle owned by the spawning thread goes
+    ## out of scope, the spawning thread frees the instance, otherwise the
+    ## spawned thread does.
+    when emulatedThreadVars:
+      tls: ThreadLocalStorage
+
+    rc: int
+      ## ref-counter. Has a maximum value of 2. All operations on it must be
+      ## atomic
+
     when TArg is void:
       dataFn: proc () {.nimcall, gcsafe.}
     else:
       dataFn: proc (m: TArg) {.nimcall, gcsafe.}
       data: TArg
 
+  Thread*[TArg] = object
+    core: ptr ThreadCore[TArg]
+    sys: SysThread
+
 proc `=copy`*[TArg](x: var Thread[TArg], y: Thread[TArg]) {.error.}
+proc `=destroy`[TArg](x: var Thread[TArg])
+
+proc release[TArg](core: ptr ThreadCore[TArg]) =
+  if atomicDec(core.rc, 1, ATOMIC_ACQ_REL) == 0:
+    deallocShared(core)
+
+proc `=destroy`[TArg](x: var Thread[TArg]) =
+  if x.core != nil:
+    # the spawning thread doesn't own the data passed along to the thread,
+    # so don't touch it
+    release(x.core)
 
 var
   threadDestructionHandlers {.rtlThreadVar.}: seq[proc () {.closure, gcsafe, raises: [].}]
@@ -94,7 +120,7 @@ proc threadTrouble() {.raises: [], gcsafe.}
   ## defined in system/excpt.nim
 
 when true:
-  proc threadProcWrapDispatch[TArg](thrd: ptr Thread[TArg]) {.raises: [].} =
+  proc threadProcWrapDispatch[TArg](thrd: ptr ThreadCore[TArg]) {.raises: [].} =
     try:
       when TArg is void:
         thrd.dataFn()
@@ -106,25 +132,26 @@ when true:
       afterThreadRuns()
 
 template threadProcWrapperBody(closure: untyped): untyped =
-  var thrd = cast[ptr Thread[TArg]](closure)
-  var core = thrd.core
-  when declared(globalsSlot): threadVarSetValue(globalsSlot, thrd.core)
-  threadProcWrapDispatch(thrd)
+  let core = cast[ptr ThreadCore[TArg]](closure)
+  when declared(globalsSlot):
+    threadVarSetValue(globalsSlot, addr(core.tls))
+  threadProcWrapDispatch(core)
   # Since an unhandled exception terminates the whole process (!), there is
   # no need for a ``try finally`` here, nor would it be correct: The current
   # exception is tried to be re-raised by the code-gen after the ``finally``!
   # However this is doomed to fail, because we already unmapped every heap
   # page!
 
+  when TArg isnot void:
+    # the spawned thread has ownership of the extra data, destroy it:
+    reset(core.data)
+
   when compileOption("gc", "orc"):
     # run a full garbage collection pass in order to free all cells
     # kept alive only through reference cycles
     GC_fullCollect()
 
-  # mark as not running anymore:
-  thrd.core = nil
-  thrd.dataFn = nil
-  deallocShared(cast[pointer](core))
+  release(core)
 
 {.push stack_trace:off.}
 # NOTE: the `threadProcWrapper` is currently special-cased by the compiler to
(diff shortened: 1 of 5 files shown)


zerbina:
As a follow-up, I think it'd make sense to have the thread parameter be a `sink` parameter. This would allow *moving* values into the created thread (including `ref` and closure values), something that's not possibly at the moment (a copy is always required).

While some higher-level facility (thread pool, coroutines, etc.) is the better choice in most cases, I still consider moving values into the thread something that `createThread` should support.

lib/system/threads.nim
```diff
@@ -61,16 +61,42 @@ const
...
+    ## The internal managnement data associated with a thread. Allocated
+    ## by the spawning thread, and - initially - owned by both the spawning
+    ## and spawned thread (shared ownership). If the spawned thread finishes
+    ## before the `Thread <#Thread>` handle owned by the spawning thread goes
+    ## out of scope, the spawning thread frees the instance, otherwise the
+    ## spawned thread does.
+    when emulatedThreadVars:
+      tls: ThreadLocalStorage
+
+    rc: int
```
saem, line 74: Preface: I like pretty much everything about this PR. The only thing that's got me a little nervous is the reduction(?) of structure, although we didn't really have it because of crashes/undefined behaviour. Anyhow, this and likely all other comments in this PR are to see if we can have structured concurrency, or the infrastructure for it, or at least set it up so it's not too painful a transition in the future. With that in mind, my comment:

---

So, I'm wondering, if it might be better to have this:

```nim
  owner: ptr Thread
  alive: bool  ## whether the child is still alive
```

Together `owner` and `alive` would be equivalent to a ref count while also incorporating a single owner concept.

Future extension and usage: we could consider ownership change and ensuring a child doesn't exist longer than it should. This would also open up the possibility of setting a different intial `owner` in the future via an extended `createThread` api, that allowed spawning a thread by a child thread for a common parent thread. All the check related to this, would be runtime checks initially, but I bet we could statically check this stuff in the future.

saem (commented):
First and foremost, I really like the changes here, so many issues got fixed all in one fell swoop, really strong endorsement for the design.

I left one comment to discussing how we might retain some structured concurrency aspects, or at least how we might afford it in the future. You might have already though of it, a better alternative, or what have you. Would love to hear you thoughts, it might just be a case of leave it as a future endeavour and this is a big leap forward regardless. I'd like to discuss it a bit before commiting.

lib/system/threads.nim
zerbina: My view of `Thread` is that it's a low-level primitive that other facilities (like a thread pool) built upon.

My model of the current implementation:
* `Thread` is a handle to a thread. It has unique ownership of the thread. Moving the handle therefore moves ownership over it (to whoever owns the destination location)
* `ThreadCore` is the internal-only management data, invisible to the programmer. Its two possible owners are: the running thread and the corresponding `Thread` handle
* a *detached thread* is a thread that outlives the `Thread`

Structured concurrency facilities are then built upon `Thread` (at least that's how I think it could work). For example, a simple handle type that prevents the thread from outliving the handle could look like this:
```nim
type JThread[Arg] = object
  # the name is borrowed from C++'s `std::jthread`
  has: bool
  thread: Thread[Arg]

proc `=copy`[T](x: var JThread[T]) {.error.} # unique ownership

proc `=destroy`[T](x: var JThread[T]) =
  if x.has:
    x.thread.joinThread()
    `=destroy`(x.thread)

# when a `JThread` goes out of scope, the thread is awaited first 
```

---

Regarding the idea you brought up, I'm not sure I fully understand it:
* unique ownership for `ThreadCore` would mean that detaching threads is impossible
* I'm not sure what the knowledge of the exact owner is useful for. Re-parenting the internal management data of a thread to another `Thread` instance doesn't seem useful nor necessary to me
* `ThreadCore` keeping a reference to a `Thread` cannot work, because `Thread` is not guaranteed to be stationary in memory. If `Thread` moves, the pointer would become stale

> we could consider ownership change and ensuring a child doesn't exist longer than it should.

If I understand correctly, something like the `JThread` implementation from above could also achieve this?

> This would also open up the possibility of setting a different intial `owner` in the future via an extended `createThread` api, that allowed spawning a thread by a child thread for a common parent thread.

Hm, what would be the effective difference compared to:
```nim
var
  child: Thread[void]  

proc run() {.thread.} =
  # ---- child thread 1
  var t = (createThread[void])(proc () =
    # ---- child thread 2
    # do something
  )
  # ...
  # now move ownership of the child thread to the parent thread:
  {.cast(gcSafe).}:
    child = move t

# ---- main thread
var t = (createThread[void])(run)
t.joinThread() # wait for child thread 1 to finish
# the main thread now owns child thread 2
```

saem: Thank you for the thoughtful response, in hindsight I realized I made a pretty big error: I misread the diff thinking it was `Thread`, the `rc` field should have tipped me off, sorry.

> Structured concurrency facilities are then built upon `Thread` (at least that's how I think it could work). For example, a simple handle type that prevents the thread from outliving the handle could look like this:
> 
> ```nim
> type JThread[Arg] = object
>   # the name is borrowed from C++'s `std::jthread`
>   has: bool
>   thread: Thread[Arg]
> 
> proc `=copy`[T](x: var JThread[T]) {.error.} # unique ownership
> 
> proc `=destroy`[T](x: var JThread[T]) =
>   if x.has:
>     x.thread.joinThread()
>     `=destroy`(x.thread)
> 
> # when a `JThread` goes out of scope, the thread is awaited first 
> ```

Completely agree, the ownership concept should be in `Thread` or a derivative (e.g.: `jThread` as above). Yup, in a structured concurrency approach there would always have to be some "owner" or the thread is the root thread, akin to a "pid 1".

> > we could consider ownership change and ensuring a child doesn't exist longer than it should.
> 
> If I understand correctly, something like the `JThread` implementation from above could also achieve this?

Yes, although ideally we'd want to figure out/prevent threads escaping their "scope", my thinking was the thread resource itself was scoped, rather than solely focusing on the `Thread` handle. It would largely prevent accidentally moving a thread, via its handle `Thread`, to a location with a longer lifetime.

> > This would also open up the possibility of setting a different intial `owner` in the future via an extended `createThread` api, that allowed spawning a thread by a child thread for a common parent thread.
> 
> Hm, what would be the effective difference compared to:
> 
> ```nim
> var
>   child: Thread[void]  
> 
> proc run() {.thread.} =
>   # ---- child thread 1
>   var t = (createThread[void])(proc () =
>     # ---- child thread 2
>     # do something
>   )
>   # ...
>   # now move ownership of the child thread to the parent thread:
>   {.cast(gcSafe).}:
>     child = move t
> 
> # ---- main thread
> var t = (createThread[void])(run)
> t.joinThread() # wait for child thread 1 to finish
> # the main thread now owns child thread 2
> ```

It's similiar, but by having some sort of owner parameter we decouple location assignment/handling from the instantiator (the owner might want to change how/where things are stored).

This is firmly in the category of "potential future ideas": perhaps having `run` take a context parameter, which is then used in conjunction with `createThread` to create a `Thread` under `main`. At which point `createThread` would have to return a view, `lent Thread`, that could be used within `run`, but ultimately ownership of the viewed `Thread` lies with `main`. But I digress; and all this at least seems like a possibility we can **meaningfully** consider.

----

OK, having had a chance to read your comments and reply, this is my current thinking:

1. the current approach: at a minimum we a first pass runtime oriented structured concurrency approach
    1. this is categorically better than _none_, which is what we currently have
2. if we want more static enforcement, we probably need some facility, or convention, to pin a handle to some lifetime, such that intra-lifetime moves are ok, but not inter-ilfetime; but that's way out of scope (hah!) for this discussion

zerbina: Thanks a lot for elaborating, I think I understand better now the overall direction you want to take with structured concurrency.

PR #1306, opened 2024-05-12
always clear slots when growing `seq` with `setLen`

zerbina:
## Summary

Growing a `seq[T]` with `setLen` after shrinking it now doesn't retain
the slots' previous state, default-initializing them instead. This was
already the behaviour for destructor-having element types, but not for
other types.

## Details

For efficiency, two optimizations are used:
* the slots memory is only default-initialized when *growing* with
  `setLen`, not when shrinking
* if the element type supports it, a single `zeroMem` call is used for
  clearing the new slots after growing

For types that don't support `zeroMem`, `default(T)` is assigned to
every new slot, which fixes type header for non-pure inheritable
objects not being initialized (and `of` thus crashing).

The `seq` implementation procedures have to assume that after resizing,
the memory of the new slots is in an undefined state, which means that
no hook calls (`=copy` or `=sink`) must be used. This is guaranteed by
performing the assignments within `.nodestroy` procedures.

Since slot assignments never expand to hook calls now, the payload
memory doesn't have to be zero'ed on allocation anymore.


lib/system/seqs_v2.nim
```diff
@@ -93,7 +104,13 @@ proc grow*[T](x: var seq[T]; newLen: Natural; value: T) =
...
   xu.len = newLen
   for i in oldLen .. newLen-1:
-    xu.p.data[i] = value
+    when supportsCopyMem(T):
+      # no copy hook exists, a direct assignment can be used
+      xu.p.data[i] = value
+    else:
+      # the slot is in an unknown state, so a direct assignment (which would
+      # invoke the copy hook) cannot be used
+      rawAssign(xu.p.data[i], value)
```
zerbina, line 113: I think introducing a magic like the following would make sense:
```nim
proc unknown[T](x: var T): var T {.magic: "Unknown".}
```

What it'd do is mark the argument location as being in an unknown/uninitialized state, preventing the compiler from expanding an assignment into a hook call. It's usage would look like this: `unknown(xu.p.data[i]) = value`.

The idea is to provide a more precise tool than `.nodestroy` for safely and efficiently implementing low-level facilities where destructor-having types need to be assigned to some uninitialized memory location.

In this specific case, the `rawAssign` workaround would become obsolete.

saem: Good idea, for name, perhaps `asUninitialized` is a tad more descriptive and since it likely won't be used often, the extra length seems acceptable.

lib/system/seqs_v2.nim
```diff
@@ -93,7 +104,13 @@ proc grow*[T](x: var seq[T]; newLen: Natural; value: T) =
     xu.p = cast[typeof(xu.p)](prepareSeqAdd(oldLen, xu.p, newLen - oldLen, sizeof(T), alignof(T)))
   xu.len = newLen
   for i in oldLen .. newLen-1:
-    xu.p.data[i] = value
+    when supportsCopyMem(T):
```
saem, line 107: If you're particularly worried about performance, we could guard this with a `when not defined(danger):` so it runs faster in `danger` builds?

saem (approved):
Looks good.

I had one idea by turning off this safer behaviour in `danger` builds, which might restore lost performance; it's not a blocker, though.

The other option is to later introduce `setLenNoClear` and `growNoClear`, or something like that, which avoid clearing/initializing memory, and using that in a few key places.

lib/system/seqs_v2.nim
zerbina: My personal preference is that `setLen` always does the same thing, regardless of the current build mode. While the regression does frustrate me a bit, I don't think it's too much of a problem, and the planned improvements to the backend should offset the regression again.

PR #1307, opened 2024-05-12
add the `supportsZeroMem` type trait

zerbina:
## Summary

The type is intended for libraries to figure out whether `zeroMem`
leaves a location in its "default-initialized" state, something that's
useful to know for optimization or correctness purposes in generic
code.

## Details

Without the type trait, either approximation via `is`, or a type-
inspection macro would have to be used. A built-in type trait is
better than a macro, since it's both more efficient and usable in
the `system` module.

The only types not supporting `zeroMem` for initialization are:
* types that don't have a zero-default (such as `range` types, enums
  not starting at zero, `.requiresInit` types, not-nil pointers, etc.)
* object types that have a type header field (non-pure object types
  that can be inherited from)
* incomplete package-level object types
* aggregate types containing the former three

The `nimskullHasSupportsZeroMem` conditional symbol is added so that
support for type-trait can be detected at compile-time.

### DIFF
--- a/compiler/sem/semmagic.nim
+++ b/compiler/sem/semmagic.nim
@@ -223,12 +223,14 @@ proc evalTypeTrait(c: PContext; traitCall: PNode, operand: PType, context: PSym)
                      hasDestructor(t)
     result = newIntNodeT(toInt128(ord(not complexObj)), traitCall, c.idgen, c.graph)
   of "supportsZeroMem":
-    # * if the type requires initialization, zero-initializing it is not valid
-    # * zero-initialization is not valid for object types with type fields
-    #   either
+    # Zero initialization is not valid for:
+    # * types requiring explicit initialization
+    # * partial types (package-level objects)
+    # * object types with a type header
     proc pred(t: PType): bool =
-      # object with type header?
-      t.kind == tyObject and not isObjLackingTypeField(t)
+      # object with type header? or package-level object?
+      t.kind == tyObject and (not isObjLackingTypeField(t) or
+        sfForward in t.sym.flags)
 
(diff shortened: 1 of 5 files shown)


compiler/sem/semmagic.nim
```diff
@@ -222,6 +222,16 @@ proc evalTypeTrait(c: PContext; traitCall: PNode, operand: PType, context: PSym)
     let complexObj = containsGarbageCollectedRef(t) or
                      hasDestructor(t)
     result = newIntNodeT(toInt128(ord(not complexObj)), traitCall, c.idgen, c.graph)
+  of "supportsZeroMem":
+    # * if the type requires initialization, zero-initializing it is not valid
+    # * zero-initialization is not valid for object types with type fields
```
saem: I couldn't find it in a search, maybe it got removed, but I thought there were partial/unknown size objects (basically c structs where we know the initial fields, but not the later ones). So anything with an unknown size would also not be zero-initializeable?

Perhaps, that's already covered in these conditions.

saem (approved):
Looks good.

I didn't quite recall, nor able to come up with it in searches, but for objects with unknown components, and therefore size, I don't believe should be considered zero initialized?

compiler/sem/semmagic.nim
zerbina: Oh, I believe you mean package-level objects. Yeah, those have to conservatively be treated as not zero-initializable, for the reasons you outlined, but aren't at the moment. Thank you, I'd have missed that!

saem: Ah yeah, package-level objects, thanks for piecing together my clues and reminding me of the name. 😄

PR #1308, opened 2024-05-12
run destructors for `.threadvar`s on thread exit

### LINKED ISSUE #1301: ` .threadvar`s aren't destroyed on thread exit
Thread-local variables (`.threadvar`) aren't destroyed when a thread exits, possibly resulting in memory leaks.

### Example

```nim
type Object = object
  val: int

  if x.val > 0:
    echo "destroyed: ", x.val

var tv {.threadvar.}: Object
tv = Object(val: 1) # initialize on the main thread

proc test() =
  proc run() {.thread.} =
    tv = Object(val: 2) # initialize the thread variable on another thread
  
  var t: Thread[void]
  t.createThread(run)
  t.joinThread()

test()
```

### Actual Output

```
```

### Expected Output

```
destroyed: 2
destroyed: 1
```

### Possible Solution

The compiler needs to emit a destructor for all thread-local variables in the NimSkull program, which is then called on thread exit.

zerbina:
## Summary

All `.threadvars`s that have destructors are now destroyed on thread
exit (including the main thread). Module-level `.threadvar`s are
destroyed *before* procedure-scoped `.threadvar`s, mirroring the
destruction order of non-threadlocal globals.

Fixes https://github.com/nim-works/nimskull/issues/1301.

## Details

Two new module operators are introduced:
* `ThreadDeinit` (destroys all module-level `.threadvar`s)
* `ThreadPostDeinit` (destroys all procedure-scoped `.threadvar`s)

They have the same role for `.threadvar`s as `Deinit` and `PostDeinit`
do for normal globals. In order to make implementation slightly
simpler, destruction of procedure-scoped `.threadvar`s is also done as
part `PostDeinit`. This means that `ThreadPostDeinit` doesn't have to
be called on program teardown.

Finally, `cbackend` (the other backends don't support threads at the
moment) generates and emits the `nimTeardownThreadVars` procedure,
which is then importc'ed by `threads` and run after the destruction
callbacks (which might use threadvars themselves).

Since all backends use `backends.generateMain` (either directly or
indirectly), destruction of `.threadvar`s also automatically works with
the JS and VM backends.


saem:
> * the destruction order for threadvars (and globals in general) likely needs to be redone from first principles at a later stage

I was just about to ask about this, because the order seems backwards -- sounds good, it can be discussed elsewhere.

saem (approved):

PR #1311, opened 2024-05-16
fix: move analyzer regression with variants

zerbina:
## Summary

Separate the `switch` lowering and destructor injection into two proper
passes and apply them concurrently, fixing a move analyzer regression.

## Details

* move the `switch` lowering out of `injectDestructorCalls`
* pass a `Changeset` instead of a mutable `MirBody` to the two passes
  in `injectdestructors`
* destructor injection and `switch` lowering are applied in a single
  batch by `backends.process` 
* the `switch` lowering is still disabled for the VM backend
* update the doc comment of `injectDestructorCalls` to reflect reality

### Move Analyzer

Applying both passes concurrently - instead of the destructor-
injection/move-analysis after `switch` lowering - fixes a regression
from ae543125c222693c915e3cfaa3cc3edb0110c569

The move analyzer now sees a discriminator assignment as:
```nim
x.discr = y
```
instead of:
```nim
if x.discr != y:
  =destroy(x) # variant destructor
x.discr = y
```

allowing it to perform automatic moves from non-variant fields again
(refer to `tmove_from_non_variant_field.nim`).

### Misc

* fix MIR pretty-printing for object/ref constructors -- the colon was
  missing between the field and value

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -26,6 +26,7 @@ import
     injecthooks,
     mirbodies,
     mirbridge,
+    mirchangesets,
     mirconstr,
     mirenv,
     mirgen,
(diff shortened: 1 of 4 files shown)


saem:
> Separate the `switch` lowering and destructor injection into two proper passes and apply them concurrently, fixing a move analyzer regression.

Aren't they applied more sequentially, and not concurrently?

saem (approved):
One minor question about the PR message body, but otherwise it's good to go.

zerbina:
> > Separate the `switch` lowering and destructor injection into two proper passes and apply them concurrently, fixing a move analyzer regression.
> 
> Aren't they applied more sequentially, and not concurrently?

Previously, both passes were implemented in the `injectDestructoCalls` procedure, with application working as follows:
```
input ----(`switch` lowering)----> X ----(destructor injection)----> output
```
now, application change to:
```
input ----(`switch` lowering)----> output
    \                              /
      ---(destructor injection)---
```
that is, both passes are applied to the same tree and the changes are then (effectively) merged. While it would have been possibly to keep `injectDestructorCalls` applying both passes, I separated them so that the decision-making of whether to actually apply the passes moves can move to the callsite.

saem:
/merge

Thanks for the detailed explanation, the diagrams are super handy.

PR #1313, opened 2024-05-19
fix(js): slow code generation

zerbina:
## Summary

Replace an algorithm - used by the JavaScript code generator -
potentially taking an exponential amount of time with one taking
linear time. In some edge cases, the previous algorithm was so slow
that the compiler would effectively get stuck.

## Details

`jsflow` is responsible for translating the goto-based control-flow of
the CGIR into JavaScript control-flow (`break`, `try`, etc.). This
requires reordering the initial list of block starts (openings) and
ends such that all openings are paired with their corresponding end.

### Previous Algorithm

The algorithm used for reordering the list works by:
1. start with the last slot in the list
2. shift the block/try opening in the slot backwards (towards the
   start) until it is on the same nesting depth as its corresponding
   end
3. repeat 2 until a fixpoint is reached (the item in the slot no longer
   requires shifting)
4. repeat the previous two steps with the previous slot

Consider the following (where `n` is a value >= 2):
```
block { // 0
  block { // n
    ... // (n - 2) .. 1
    block { // n - 1
    } // 0
  } // 1
  ...
} // n
```

In the situation show above, reaching a fixpoint for the slot
originally containing the `n - 1` block requires `3 * (2 ^ (n - 2))`
iterations/shifts (exponential time).

For large `n` (> 30), the time spent on sorting grew so large that the
compiler would appear to be effectively stuck.

### New Algorithm

The new algorithm requires `n` number of shifts in the worst-case,
where `n` is the number of labels in the input CGIR.

It works by first computing an order value for every label, based on
the position of the label. The relevant openings (block and try) are
then reordered such that they're only enclosed by openings with a
higher order value.

### DIFF
--- a/compiler/backend/jsflow.nim
+++ b/compiler/backend/jsflow.nim
@@ -173,15 +173,13 @@ proc sort(structs: var seq[Structure]) =
       order[idx] = orderVal
       inc orderVal
 
-  var
-    stack: seq[tuple[pos: int, order: int]]
-    i = 0
+  var stack: seq[tuple[pos: int, order: int]]
 
   # now we want to establish the following: for each parent/child opening, the
   # parent has a *greater* order value than the child. Structure, catch, and
   # finally openings must stay associated with their original statement -- they
   # cannot move
-  while i < structs.len:
+  for i in 0..<structs.len:
     template invariant(): bool =
       # the loop invariant
       stack.len < 2 or stack[^2].order > stack[^1].order
@@ -226,8 +224,6 @@ proc sort(structs: var seq[Structure]) =
       discard
 
     assert invariant()
-    # we're always making progress, the loop is guaranteed to terminate
-    inc i
 
 proc toStructureList*(stmts: openArray[CgNode]): StructDesc =
   ## Creates and returns the JavaScript control-flow-construct-focused
--- a/compiler/backend/jsflow.nim
+++ b/compiler/backend/jsflow.nim
@@ -59,6 +59,15 @@ type
 const
   Terminators = {stkReturn, stkTerminator}
 
+proc rotateRight[T](s: var seq[T], a, b: int) =
+  ## Rotates the items in slice a..b to the right by one element.
+  let backup = s[b]
+  var i = b
+  while i > a:
+    s[i] = s[i - 1]
+    dec i
+  s[a] = backup
+
 func finalTarget*(n: CgNode): CgNode =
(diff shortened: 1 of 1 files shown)


zerbina:
Here's a code snippet that triggers the scenario described in the commit message. I've originally derived it from the code generated for a closure iterator.
```nim
proc doRaise() =

proc test(a: int) =
  block:
    block L0:
      try:
        case a
        of 0:
          doRaise()
          break L0
        of 1: return
        of 2: break L0
        of 3: break L0
        of 4: break L0
        of 5: break L0
        of 6: break L0
        of 7: break L0
        of 8: break L0
        of 9: break L0
        of 10: break L0
        of 11: break L0
        of 12: break L0
        of 13: break L0
        of 14: break L0
        of 15: break L0
        of 16: break L0
        of 17: break L0
        of 18: break L0
        of 19: break L0
        of 20: break L0
        of 21: break L0
        of 22: break L0
        of 23: break L0
        of 24: break L0
        of 25: break L0
        of 26: break L0
        of 27: break L0
        of 28: break L0
        of 29: break L0
        of 30: break L0
        else:  return
        echo "a"
      except:
        discard
      echo "b"
    echo "c"
  echo "d"

test(1)
```
With the current `-d:release` devel compiler, compiling the above for the JavaScript target takes around 77 seconds (!) for me. Adding more `of ...: break L0` branches continues to increase the compile time.

With the new algorithm, compilation takes ~0.15 seconds.

compiler/backend/jsflow.nim
```diff
@@ -143,6 +152,83 @@ func endsInTerminator(structs: seq[Structure], start: int): bool =
...
+    of stkEnd:
+      # if the pre-conditions hold, the end corresponds to opening at the stack
+      # head -- pop the block
+      let e = stack.pop()
+      assert e.order == order[it.label.int]
+    of stkTerminator, stkReturn:
+      discard
+
+    assert invariant()
+    # we're always making progress, the loop is guaranteed to terminate
```
saem: If this is guaranteed, then should this be a `for` loop instead of a `while` loop?

saem (approved):
One minor stylistic comment, otherwise good to merge, the new algorithm is a nice improvement.

compiler/backend/jsflow.nim
zerbina: Yep, that'd be better -- good call.

zerbina:
/merge

PR #1319, opened 2024-05-22
fix(sem): crash with `void` fields in instantiated tuple types

### LINKED ISSUE #1316: compiler segfaults with `void` and `distinct`
### Specification

The compiler is segfaulting with some uses of `void`. In some cases, `void` can be used as a type argument to some generic type. In other cases it causes the compiler to segfault. I originally thought that `default` was in part to blame but I have no idea now.

### Example

```nim
type
  EntryBox*[K, V] = object
    entry*: tuple[key: K, value: V]

  Node*[K, V] = object
    entries*: array[2, EntryBox[K, V]]

  Wrapper1*[K, V] = object
    node*: Node[K, V]

  Wrapper2*[K] {.borrow: `.`.} = distinct Wrapper1[K, void]
  WrapperRef2*[K] = ref Wrapper2[K]

proc init_2*[K](): WrapperRef2[K]  =
  var res = default(WrapperRef2[K])
  res.node = Node[K, void]()
  return res

var s = init_2[int]()
```

### Actual Output

Fails to compile due to segfault:

```
./scripts/build.sh: line 312: 58542 Segmentation fault: 11  ( ${NIM} -c --cc: ${CC} --gc: arc --opt: speed --threads: off --profiler: off --stackTrace: off --d: release --nimcache: ${PATH_TO_C_ASSETS} c ${FILE} )
```

### Expected Output

Should compile

### Additional Information

Similar to the snippet above, this snippet uses `void` and causes the compiler to segfault, despite the lack of use of `distinct`.

```nim
type
  EntryBox*[K, V] = object
    entry*: tuple[key: K, value: V]

  Node*[K, V] = object
    entries*: array[2, EntryBox[K, V]]

  Wrapper1*[K, V] = object
    node*: Node[K, V]
  WrapperRef1*[K, V] = ref Wrapper1[K, V]

  WrapperRef2*[K] = object
    re: WrapperRef1[K, void]

proc init_2*[K](): WrapperRef2[K]  =
  result.re = WrapperRef1[K, void]()

var s = init_2[int]()
```

The next snippet compiles without problem despite the use of `void` and prints `1`, which is what we would expect.

```nim
type
  Obj*[K, V] = object
    k*: K
    v*: V
  KeyOnly*[K] = Obj[K, void]

var ko = KeyOnly[uint8]()
echo sizeof(ko)
```

zerbina:
## Summary

Handle removal of `void` fields from tuple types properly when
instantiating generic types, fixing various compiler crashes.

Fixes https://github.com/nim-works/nimskull/issues/1316.

## Details

The problem was that `void` fields weren't removed from the type, like
what happens for `tyObject` and `tyProc` types. Therefore:
* `tyVoid` types unexpectedly reached the code generators as part of
  records
* `computeSizeAlign` crashed because it expects the `nkRecList` of a
  `tyTuple` to only contain `nkSym` nodes, but `replaceTypeVarsN`
  replaced the `nkSym` node for `void` fields with empty `nkRecList`
  nodes earlier

In order to remove the `void` fields properly:
* custom logic is used for the tuple's record AST, which - compared to
  `replaceTypeVarsN` - also updates the fields' position
* the type and record slots of void are removed with a generalized
  version of `eraseVoidParams` (`eraseVoidTypes`)

Finally, the `propagateFieldFlags` call for tuple types is removed. It
is redundant,  since the type flags are already propagated when the
type variables are replaced.

### DIFF
--- a/compiler/sem/semtypinst.nim
+++ b/compiler/sem/semtypinst.nim
@@ -658,24 +658,31 @@ proc handleGenericInvocation(cl: var TReplTypeVars, t: PType): PType =
           attachedAsgn, col)
       excl mm.flags, tfFromGeneric
 
-proc eraseVoidParams*(t: PType) =
-  # transform '(): void' into '()' because old parts of the compiler really
-  # don't deal with '(): void':
-  if t[0] != nil and t[0].kind == tyVoid:
-    t[0] = nil
-
-  for i in 1..<t.len:
+proc eraseVoidTypes(t: PType; start = 0) =
+  ## Removes all ``tyVoid`` items from `t`. If `t` has attached AST, the slots
+  ## corresponding to the type items are removed too.
+  for i in start..<t.len:
     # don't touch any memory unless necessary
     if t[i].kind == tyVoid:
       var pos = i
       for j in i+1..<t.len:
         if t[j].kind != tyVoid:
           t[pos] = t[j]
-          t.n[pos] = t.n[j]
+          if t.n != nil:
+            t.n[pos] = t.n[j]
           inc pos
       setLen t.sons, pos
-      setLen t.n.sons, pos
+      if t.n != nil:
+        setLen t.n.sons, pos
       break
+
+proc eraseVoidParams*(t: PType) =
+  # transform '(): void' into '()' because old parts of the compiler really
+  # don't deal with '(): void':
+  if t[0] != nil and t[0].kind == tyVoid:
+    t[0] = nil
+
+  eraseVoidTypes(t, start=1)
 
 proc skipIntLiteralParams*(t: PType; idgen: IdGenerator) =
   for i in 0..<t.len:
@@ -690,20 +697,6 @@ proc skipIntLiteralParams*(t: PType; idgen: IdGenerator) =
   # param, the results gets infected with static as well:
   if t[0] != nil and t[0].kind == tyStatic:
     t[0] = t[0].base
-
-proc propagateFieldFlags(t: PType, n: PNode) =
-  # This is meant for objects and tuples
-  # The type must be fully instantiated!
-  if n.isNil:
-    return
-  #internalAssert n.kind != nkRecWhen
-  case n.kind
-  of nkSym:
-    propagateToOwner(t, n.sym.typ)
-  of nkRecList, nkRecCase, nkOfBranch, nkElse:
-    for son in n:
-      propagateFieldFlags(t, son)
-  else: discard
 
 proc replaceTypeVarsTAux(cl: var TReplTypeVars, t: PType): PType =
   template bailout =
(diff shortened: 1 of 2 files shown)


saem (commented):
These might be "big conceptual" things that sorta extend beyond the scope of "making things more consistent and not crash" of this PR. I'll lay them out here for at least some preliminary discussion, I think what I'm struggling most with is reconciling long term needs with short term (at the very least the crash, and inconsistency, aren't good). Points that popped up in my head:

Should removal of `void` fields, and parameters, happen after sem (including compile time), but before runtime-codegen?

Also, there is `void` vs `unit`, where the former is the absence of anything, while the latter is unitary values, both wouldn't make it to runtime, but at least the latter would be there for comptime. I'm not sure whether we should take the `unit` or true `void` interpretation of `void` in this case, thoughts?

If we're dropping fields, and to a degree parameters, then sub-type relationships get broken where fields that exist in one interface don't in another. I don't really have a question, just an alarm bell in my head.

---

With all that said, I guess we could go for consistency for now, but say "we will almost undoubtedly break this later", because the first and third points are especially concerning.

zerbina:
> Should removal of `void` fields, and parameters, happen after sem (including compile time), but before runtime-codegen?

When first thinking about the topic, I'd have said yes, but now I tend towards no, mainly because I think there's value in:
* having `void` behave in line with what it's name implies
* being able to represent "absence of anything"

> Also, there is `void` vs `unit`, where the former is the absence of anything, while the former is the unitary values, both wouldn't make it to runtime, but at least the latter would be there for comptime. I'm not sure whether we should take the `unit` or true `void` interpretation of `void` in this case, thoughts?

Maybe it makes sense to have both `void` and `unit`? That would allow existing code to stay working for the most part. On the implementation side, not changing the meaning of `void` would also be a lot less error prone (all direct and indirect usages of `tyVoid` would have to be audited).

`void` would continue to drop fields/parameters during generic parameter substitution, while `unit` would keep them around until code generation.

If dropping parameters is something that should not be supported, I think it'd be better to not allow binding `void` to generic parameters that are substituted for field/parameter types (only `unit` could be used for this). That is:
```nim
type
  A[T] = object
    when T isnot void: # `T` is never substituted for a field/parameter type
      x: int
    else:
      x: float
  B[T] = object
    x: T # `T` would be inferred as `T: not void`

A[void]() # allowed
B[void]() # rejected before instantiating
B[unit]() # allowed
```

---

Going the route of consistency is the easiest solution for the issue (i.e.: substituting `void` into tuples always leads to internal compiler errors in almost all cases) I can think of. At the same time, it doesn't complicate changing the meaning of `void` (or introducing the `unit` type).

saem:
> Going the route of consistency is the easiest solution for the issue (i.e.: substituting `void` into tuples always leads to internal compiler errors in almost all cases) I can think of. At the same time, it doesn't complicate changing the meaning of `void` (or introducing the `unit` type).

Yeah, I forgot to consider cost of change after the fact and that either stays the same, or more likely gets a bit easier (consistency in implementation and testing).

---

### Long-term

> Maybe it makes sense to have both `void` and `unit`?

We might end up there, but `void` having the power to drop fields even from sem, might have some serious consequences for the type system. Presently, I'm leaning towards disallowing it in the future. Consider the following:

```nim
type
  A[T] = object
    foo: int
    bar: T
    baz: float

discard A[int]()
discard A[void]()
```

The `int` variant, or any other non-`void` type, now has a different field count, the field positions are also different, and the interface doesn't match up either. This could start having funny effects with field-access/call-resolution as `A[void]().bar` would skip built-in field access and go to call resolution. This might have some serious consequences for implementing deeper semantic analysis of generic routines (than what we do today) and I wouldn't be surprised if this had some seriously confusing interactions with `concept`s.

PR #1321, opened 2024-05-26
cgen: better code-gen for `seq.len`

zerbina:
## Summary

Don't store the result of a `seq.len` in an additional temporary. An
optimizing C compiler can easily eliminate such temporaries, but doing
it on the NimSkull side reduces the amount of work for the C compiler.

## Details

* the temporary is a leftover from when argument expression purity
  wasn't guaranteed
* the `cgen.getIntTemp` procedure is now obsolete and thus removed


saem (approved):
Nice addition to delete ratio. :D

PR #1322, opened 2024-05-26
lower float overflow checks with MIR pass

zerbina:
## Summary

Lower checked float arithmetic operations with an MIR pass, instead
of as part of C code generation. This is a pure refactoring, with the
goal of shrinking down the C code generator.

## Details

* the checked float arithmetic operations are turned into an unchecked
  operation + infinity check (replicating what `cgen` did)
* the lowering in `cgen` for float arithmetic is removed

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -1386,20 +1386,6 @@ proc genStrEquals(p: BProc, e: CgNode, d: var TLoc) =
   else:
     binaryExpr(p, e, d, "#eqStrings($1, $2)")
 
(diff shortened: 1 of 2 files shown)


zerbina:
Another benefit of moving the lowering of run-time checks into the MIR phase, but one that's not immediately obvious at the moment, is that both the future WASM and `asm.js` code generators can also use the MIR passes, meaning that they don't have to implement code generation for the checks themselves.

So while the amount of code in the compiler currently grows, this will eventually be offset by the smaller (and simpler and less error-prone) `asm.js` and WASM code generators.

saem (approved):

PR #1323, opened 2024-05-27
improve code generation for `@[...]`

zerbina:
## Summary

Don't use an intermediate temporary for the array construction but
instead move the items into the constructed seq directly. All backends
are affected.

## Details

`mirgen` now translates a `@[...]` expression into a `mnkSeqConstr`
tree rather than:
```
def _1 = [...] # array construction
def _2 = arrToSeq(consume _1)
```
This gets rid of the unnecessary temporary and shallow array copy.

Handling of `mArrToSeq` in the code generators is also improved /
cleaned up:
* the handling of `cnkArrayConstr` operands is removed (it's a leftover
  from before the MIR rework)
* `jsgen`: the operand to `mArrToSeq` call is guaranteed to be owned,
  so no extra copy is required
* `cgen`: the `TLoc` passed to `genArrToSeq` is always a non-empty
  loc, so the `if d.k == locNone` check is unnecessary and thus removed

### DIFF
--- a/compiler/backend/ccgexprs.nim
+++ b/compiler/backend/ccgexprs.nim
@@ -991,13 +991,6 @@ proc genSeqConstr(p: BProc, n: CgNode, d: var TLoc) =
 
 proc genArrToSeq(p: BProc, n: CgNode, d: var TLoc) =
   var elem, a, arr: TLoc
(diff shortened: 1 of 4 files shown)


saem (approved):
Yay for optimizations (even if it's a restoration).

PR #1325, opened 2024-05-28
lower integer overflow checks with MIR pass

zerbina:
## Summary

Lower the integer overflow checks with an MIR pass, instead of as part
of C code generation. This is mainly a refactoring, with the goal of
shrinking down the C code generator. As a side-effect, overflow checks
are very slightly faster.

## Details

The MIR produced by the lowering is similar to what the C code
generator previously output, but with one major difference: integer
operands smaller than the target's integer size now use dedicated
compilerprocs instead of first converting to `int`, performing the
checked arithmetic, and then comparing against the integer bounds.

The main reason for this change is that it makes the lowering pass
simpler and requires less MIR code, but it also has the benefit of
producing more efficient code. Run-time behaviour is not affected.

No dedicated gcc built-ins for overflow-checked arithmetic with
`short int` (int16) and `signed char` (int8) exists, so all checked
arithmetic compilerprocs now map to the generic gcc built-ins.

### DIFF
--- a/compiler/mir/rtchecks.nim
+++ b/compiler/mir/rtchecks.nim
@@ -397,6 +397,114 @@ proc emitObjectCheck(tree; call; graph; env; bu) =
     bu.emitCall(tree, call, env.addCompilerProc(graph, "raiseObjectConversionError")):
       discard
 
+# XXX: currently cannot be moved to within ``emitCheckedBinaryIntOp`` due to a
+#      csource compiler bug
+const ArithChcks: array[mAddI..mModI, array[tyInt..tyInt64, string]] = block:
+  # compile the names of the procedure at compile-time
+  var res: array[mAddI..mModI, array[tyInt..tyInt64, string]]
+  for op, it in res.mpairs:
+    let base = case op
+      of mAddI: "nimAddInt"
+      of mSubI: "nimSubInt"
+      of mMulI: "nimMulInt"
+      of mDivI: "nimDivInt"
+      of mModI: "nimModInt"
+
+    # the inner array stores the per-width names, which is the base name
+    # suffixed with the width
+    for i in 0..3:
+      it[[tyInt8, tyInt16, tyInt32, tyInt64][i]] = base & $(8 shl i)
+
+    # for the moment, there's still a generic-width version
+    it[tyInt] = base
+
+  res
+
+proc emitCheckedBinaryIntOp(tree; call; graph; env; bu): Value =
+  ## Emits the lowered version of a checked binary arithmetic operation.
+  ## Usually, it looks like this:
+  ##   def _1
+  ##   def _2 = nimAddInt(arg a, arg b, name _1)
+  ##   if _2:
+  ##     raiseOverflow()
+  ##   result = _1
+  let
+    magic = tree[call + 1].magic
+    t = env[tree[call].typ].skipTypes(abstractRange)
+    x = NodePosition tree.argument(call, 0)
+    y = NodePosition tree.argument(call, 1)
+
+  # divison-by-zero is checked for separately, so that a dedicated error can
+  # be reported:
+  if magic in {mDivI, mModI}:
+    let cond = bu.wrapTemp BoolType:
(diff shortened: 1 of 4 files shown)


zerbina:
The C code produced for:
```nim
proc test(a, b: int16): int16 =
  a + b
```
looked roughly like this (cleaned up for clarity):
```c
typedef short int NI16;

void raiseOverflow();

NI16 test(NI16 a, NI16 b) {
    NI16 result;
    int tmp;
    if (__builtin_sadd_overflow(a, b, &tmp)) {
        raiseOverflow();
        goto Error;
    }
    if (tmp < -32768 || tmp > 32767) {
        raiseOverflow();
        goto Error;
    }
    result = (NI16)tmp;
    Error:
    return result;
}
```
Now it looks like this:
```c
typedef short int NI16;

void raiseOverflow();

NI16 test(NI16 a, NI16 b) {
    NI16 result;
    NI16 tmp;
    if (__builtin_add_overflow(a, b, &tmp)) {
        raiseOverflow();
        goto Error;
    }
    result = tmp;
    Error:
    return result;
}
```
Comparing both snippets with the [compiler explorer](https://godbolt.org) shows that both the current gcc (14.1) and clang (18.1.0) produce less instructions.

compiler/mir/rtchecks.nim
```diff
@@ -397,6 +397,131 @@ proc emitObjectCheck(tree; call; graph; env; bu) =
...
+      of mAddI: "nimAddInt"
+      of mSubI: "nimSubInt"
+      of mMulI: "nimMulInt"
+      of mDivI: "nimDivInt"
+      of mModI: "nimModInt"
+
+    # the inner array stores the per-width names, which is the base name
+    # suffixed with the width
+    for i in 0..3:
+      it[[tyInt8, tyInt16, tyInt32, tyInt64][i]] = base & $(8 shl i)
```
saem:
```suggestion
    const intKinds = [tyInt8, tyInt16, tyInt32, tyInt64]
    for i, k in intKinds.pairs:
      it[k] = base & $(8 shl i)
```

I believe this is equivalent and a tad clearer, but not a required change.

compiler/mir/rtchecks.nim
```diff
@@ -397,6 +397,131 @@ proc emitObjectCheck(tree; call; graph; env; bu) =
...
+
+  result = bu.allocTemp(tree[call].typ)
+  bu.subTree mnkDef:
+    bu.use result
+    bu.add MirNode(kind: mnkNone)
+
+  let kind = t.skipTypes({tyEnum}).kind
+  if kind in {tyUInt..tyUInt64, tyBool}:
+    # enums using an unsigned integer as the underlying type can reach here
+    # (due to lowering ``succ`` and ``pred`` lowering), and no integer overflow
```
saem:
```suggestion
    # (due to ``succ`` and ``pred`` lowering), and no integer overflow
```

I think this is what you meant.

compiler/mir/rtchecks.nim
```diff
@@ -397,6 +397,131 @@ proc emitObjectCheck(tree; call; graph; env; bu) =
...
+          bu.emitByVal env.makeLiteral(litKind, lastOrd(graph.config, t),
+                                       result.typ)
+          bu.emitByVal result
+
+    bu.buildIf cond:
+      bu.emitCall(tree, call, env.addCompilerProc(graph, "raiseOverflow")):
+        discard
+
+proc emitUnaryOverflowCheck(tree; call; graph; env; bu) =
+  ## Emits the overflow check for a integer negation operation:
```
saem:
```suggestion
  ## Emits the overflow check for an integer negation operation:
```

saem (approved):
A few minor suggestions, nothing blocking; looks good to me.

compiler/mir/rtchecks.nim
zerbina: Yep, it's much easier to read.

zerbina:

PR #1326, opened 2024-05-28
fix `result` not being cleaned up when procedure raises

### LINKED ISSUE #1305: procedure result is not cleaned up if the procedure raises
### Example

```nim

type Object = object
  has: bool

var wasDestroyed: bool

  if x.has:
    wasDestroyed = true

proc create(doRaise: bool): Object =
  result = Object(has: true)
  if doRaise:

proc test() =
  try:
    var x = create(true)
  except:
    # the value created in `create` must have been destroyed at this point
    doAssert wasDestroyed

test()
```

### Actual Output

The assertion fails.

### Expected Output

The program compiles and runs successfully.

### Possible Solution

Either the caller or callee needs to destroy the result in case the callee raises. The simplest solution would be wrapping the body of a procedure that can raise in a `try`/`except` that destroys the `result` and re-raises the exception.

### Additional Information

This is a problem at the MIR-level, meaning that all backends are affected.

zerbina:
## Summary

The `result` variable is now destroyed when the procedure exits due to
a raised exception, fixing a memory/resource leak. All backends were
affected.

Fixes https://github.com/nim-works/nimskull/issues/1305.

## Details

* the `result` is destroyed within a special try-finally that only
  intercepts exceptional control-flow
* the new `errorOnly` field designates a `bkTryFinally` as only
  intercepting exceptional control-flow
* `mirgen` handles emitting the destroy operation and injecting the
  `try`

When the `result` variable is not initialized on any control-flow paths
leading to the destroy operation, the destroy is elided by the "destroy
elimination" pass; a test (`tno_destroy_for_empty_result.nim`) is added
to ensure this works.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -2400,10 +2400,21 @@ proc generateCode*(graph: ModuleGraph, env: var MirEnv, owner: PSym,
   var c = initCtx(graph, config, owner, move env)
   c.sp.active = (body, c.sp.map.add(body))
 
+  let
+    needsTerminate = sfNeverRaises in owner.flags
+    needsCleanup = (c.injectDestructors and
+                    owner.kind in routineKinds and
+                    owner.typ[0] != nil and
+                    hasDestructor(owner.typ[0]))
+
   c.withBlock bkBlock: # the target for return statements
-    if sfNeverRaises in owner.flags:
+    if needsTerminate:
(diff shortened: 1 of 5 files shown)


saem (approved):

zerbina:
/merge

It looks like the merge command previously failed due to GitHub spuriously being unable to return the PR body (refer to https://github.com/nim-works/nimskull/actions/runs/9278374707/job/25529280808).

PR #1328, opened 2024-05-29
sem: fix crash when `sizeof` is called on generic type

### LINKED ISSUE #1317: compiler fatal error on lack of type arguments
### Specification

Instead of returning information to the user about needing to specify type arguments for `Wrapper` (as in `Wrapper[int, int]`). Instead the compiler hits a fatal error.

### Example

```nim
type
  Wrapper*[K, V] = object
    node*: (K, V)

echo sizeof(Wrapper)
```

### Actual Output

```
Fatal: Internal unreachable code executed - ccgtypes.nim(708, 18) (getTypeDescAux(tyGenericParam)) should never be called.
```

### Expected Output

I would expect the compiler to fail to compile the code and print a message to the user about the lack of type arguments.

theSherwood:
## Summary
* Changes  `semSizeof`  to check that the argument to  `sizeof`  is a
concrete type and report an error if that is not the case
* Fixes issue in which  `sizeof`  is called with a generic type, causing
the compiler to crash

## Details

Fixes: https://github.com/nim-works/nimskull/issues/1317

* The following code was causing the compiler to crash. Now it should
cause the compiler to print an error message about concrete types.
```nim
type
  GenericType[K, V] = object
    field: (K, V)

let size = sizeof(GenericType)
```
* adds  `adSemTIsNotAConcreteType`  and maps it to 
`rsemTIsNotAConcreteType` 
* adds minimal test featuring the snippet above


mrgaturus:
what about name report enum entires as `adSemTypeNotConcrete` and `rsemTypeNotConcrete`?

compiler/sem/semexprs.nim
```diff
@@ -2833,8 +2833,13 @@ proc semSizeof(c: PContext, n: PNode): PNode =
   of 2:
     #restoreOldStyleType(n[1])
     n[1] = semExprWithType(c, n[1])
-    n.typ = getSysType(c.graph, n.info, tyInt)
-    result = foldSizeOf(c.config, n, n)
+    if containsGenericType(n[1].typ):
+      n[1] = c.config.newError(n[1], PAstDiag(kind: adSemTIsNotAConcreteType,
+                                              wrongType: n[1].typ.sons[0]))
```
```suggestion
      # report the type, not the typedesc
      n[1] = c.config.newError(n[1], PAstDiag(kind: adSemTIsNotAConcreteType,
                                              wrongType: n[1].typ[0]))
```
* it's easy to overlook, so I think an explicit comment makes sense
* `typ[...]` instead of `typ.sons[...]` is generally preferred

tests/lang_callable/generics/tsizeof_generic_error.nim
```diff
@@ -0,0 +1,9 @@
+discard """
+  errormsg: "'GenericType' is not a concrete type"
+"""
```
zerbina, line 4:
```suggestion
  line: 10
"""
```
To make sure the error is reported where expected.

zerbina (commented):
Thank you for the PR, the fix itself looks good, but it doesn't yet handle the case where `system.sizeof` is overloaded.

Contrary to what the manual says, `sizeof` *can* be overloaded. Given the following snippet:
```nim
proc sizeof(a, b: int) = discard
# ^^ not actually picked, but results in `sizeof` being an overloaded symbol

sizeof(int)
```
the `mSizeOf` magic is not handled by `semMagic`, but by `magicsAfterOverloadResolution`, which doesn't call `semSizeOf`, and thus generic arguments aren't rejected.

---

While not something to be done in this PR, `alignof` is affected by the same problem.

zerbina:
@theSherwood: In `tsizeof3`, `sem` relied on `sizeof` to silently work when supplied with generic parameters. Now it doesn't, so the test fails.

I've thought about it for a moment, and concluded that there's no easy workaround for it. There's a proper fix I have in mind (not analysing arguments to generic invocation when any of them reference type variables), but that requires a separate PR. It's also significantly more involved than the `sizeof` fix, but if you're interested, I could try to walk you through it.

theSherwood:
@zerbina: Dang. That's unfortunate. If it were just finishing up this PR, I could do it, but I probably don't have much time I can reasonably devote to nimskull in the near future. I'd like to chip in now and then as time allows, but I don't think I can commit to anything. Thanks for your patience through all this and sorry to leave you in the lurch.

zerbina:
> If it were just finishing up this PR, I could do it, but I probably don't have much time I can reasonably devote to nimskull in the near future. I'd like to chip in now and then as time allows, but I don't think I can commit to anything.

That's totally fine, no worries. I'll look into making a fix for the problem with generic instantiations, and will notify you once the the problem is fixed in devel, so that you can finish this PR.

zerbina:
@theSherwood: The problem of `sizeof(T)` being analyzed while `T` is still unresolved should be fixed now. Could you merge the `devel` branch into your branch and check whether the failing test succeeds?

zerbina (changes requested):
Nice, that fixed it.

What remains is fixing generic types reaching the code generator when `sizeof` is overloaded:
```nim
proc sizeof(a, b: int) = discard

discard sizeof(GenericType)
# ^^ this still causes the internal assertion to fail
```

zerbina (approved):
Nice work, thank you for seeing this through!

PR #1329, opened 2024-05-31
transf: remove unreachable code

zerbina:
## Summary

Eliminate unreachable statements and expression during `transf`,
instead of in `mirgen`. This:
* removes some workarounds from `mirgen`
* improves the quality of unreachable code elimination
* fixes a compiler crash related to non-returning expressions

## Details

Statement lists can have trailing unreachable code, and expressions can
therefore be non-returning (e.g., `(;break; 1)`).

`mirgen` made an attempt at eliminating both, but the approach used
was indirect and didn't handle expressions well, leading to crashes for
code such as `f((;return; var x = 0; 1), x)`, because the `var x`
definition wasn't translated, but the subsequent usage of `x` was.

Removing unreachable statements/expressions is now done by an AST-based
pass that's run after early `transf` (so that iterator inlining took
place and the AST is less complex) but before the `closureiters`
transformation. The pass is implemented by the `unreachable_elim`
module.

Beyond fixing the crashes, this has the additional benefits that:
* the `closureiters` transformation doesn't take unreachable code into
  account
* `mirgen` can detect `finally` clauses that have no structured exit
  and emit better code for them. The handling for the no-structured-
  exit case already existed (`doesntExit`), just the detection was
  missing
* more straightforward MIR translation, using less conditionals

A test for the various possible scenarios of unreachable expressions/
statements is added.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -188,9 +188,10 @@ proc isCursor(n: PNode): bool =
   else:
     false
 
-func doesReturn(n: PNode): bool =
-  # the void type indicates that the expression/statement doesn't return
-  n.typ == nil or n.typ.kind != tyVoid
+template doesReturn(n: PNode): bool =
+  ## Returns whether `n` is a statement with a structured exit.
+  mixin c
+  n.typ != c.graph.noreturnType
 
 func initDestination(v: sink Value, isFirst, sink: bool): Destination =
   var flags: set[DestFlag]
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -1802,7 +1802,8 @@ proc genTry(c: var TCtx, n: PNode, dest: Destination) =
   if hasFinally:
     # the finally clause also applies to the except clauses, so it's
     # pushed first
-    c.blocks.add Block(kind: bkTryFinally)
+    c.blocks.add Block(kind: bkTryFinally,
+                       doesntExit: not doesReturn(n[^1][0]))
 
   if hasExcept:
     c.blocks.add Block(kind: bkTryExcept)
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -157,11 +157,6 @@ type
       ## > 0 if the current statement/expression is part of a loop
     injectDestructors: bool
       ## whether injection of destroy operations is enabled
-    unreachable: bool
-      ## set to true when entering unreachable code (e.g., statements
-      ## immediately following a `break`). Disables:
-      ## * translation of AST
-      ## * emission of control-flow statements at block/scope end
 
     # input:
     userOptions: set[TOption]
@@ -193,21 +188,9 @@ proc isCursor(n: PNode): bool =
   else:
     false
 
-func endsInNoReturn(n: PNode): bool =
-  ## Tests and returns whether the simple or compound statement `n` ends in a
-  ## no-return statement
-
-  # TODO: this is a patched version of ``sem.endsInNoReturn`` that also
-  #       considers ``nkPragmaBlock``. Move this procedure somewhere common and
-  #       replace ``sem.endsInNoReturn`` with it
-  const SkipSet = {nkStmtList, nkStmtListExpr, nkPragmaBlock}
-  var it {.cursor.} = n
-  while it.kind in SkipSet and it.len > 0:
-    it = it.lastSon
-
-  result = it.kind in nkLastBlockStmts or
-    (it.kind in nkCallKinds and it[0].kind == nkSym and
-     sfNoReturn in it[0].sym.flags)
+func doesReturn(n: PNode): bool =
+  # the void type indicates that the expression/statement doesn't return
+  n.typ == nil or n.typ.kind != tyVoid
 
 func initDestination(v: sink Value, isFirst, sink: bool): Destination =
   var flags: set[DestFlag]
@@ -275,12 +258,15 @@ template subTree(c: var TCtx, n: MirNode, body: untyped) =
   c.builder.subTree n:
     body
 
-template scope(c: var TCtx, body: untyped) =
+template scope(c: var TCtx, exits: bool, body: untyped) =
+  ## `exits` signals whether the body of the scope has a structured control-
+  ## flow exit.
+  let e = exits
   inc c.scopeDepth
   c.builder.subTree mnkScope:
     let prev = c.blocks.startScope()
     body
-    c.blocks.closeScope(c.builder, prev, not c.unreachable)
+    c.blocks.closeScope(c.builder, prev, e)
   dec c.scopeDepth
 
 template use(c: var TCtx, val: Value) =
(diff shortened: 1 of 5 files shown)


zerbina:
@saem already brought it up in the past, and since it's heavily related, I'm mentioning it here again. It would make sense to:
* make a distinction between absence of value/type (`void`) and unitary value (`unit`), the latter also being a new built-in type
* type non-terminal statements (such as assignments) as `unit`, and terminal statements (such as `break`) as `void`

This would a have few significant benefits to the language and compiler:
* the not-well integrated `noreturn` pragma is subsumed by the type system. That is, `void` as the return type means "doesn't return (normally)"
* whether a procedure doesn't return is simply inferred by normal type inference
* detection of unreachable code becomes much simpler and exhaustive (everything following a `void` expression is unreachable)
* `unit` is usable as the type of parameters/variables (compared to `void`) and stays around until code generation, which can be useful for generic code

If it's enforced that a `void` expression is the last expression in a block (i.e., that `(;return; 1)` is illegal), then the whole `unreachable_elim` module becomes obsolete (faster compiler, lower complexity, one less source of bugs).

saem:
> @saem already brought it up in the past, and since it's heavily related, I'm mentioning it here again. It would make sense to:
> 
> * make a distinction between absence of value/type (`void`) and unitary value (`unit`), the latter also being a new built-in type
> * type non-terminal statements (such as assignments) as `unit`, and terminal statements (such as `break`) as `void`

Yup, this would be a huge improvement.

> This would a have few significant benefits to the language and compiler:
> 
> * the not-well integrated `noreturn` pragma is subsumed by the type system. That is, `void` as the return type means "doesn't return (normally)"
> * whether a procedure doesn't return is simply inferred by normal type inference
> * detection of unreachable code becomes much simpler and exhaustive (everything following a `void` expression is unreachable)
> * `unit` is usable as the type of parameters/variables (compared to `void`) and stays around until code generation, which can be useful for generic code

Yup, I think treating `void` as outlined above makes sense, where it serves both as an absence of a value ("a logically uninhabited data type") and abnormal return. We can later consider whether separating it from "never returns", `never` makes sense, but that's a separate step.

> If it's enforced that a `void` expression is the last expression in a block (i.e., that `(;return; 1)` is illegal), then the whole `unreachable_elim` module becomes obsolete (faster compiler, lower complexity, one less source of bugs).

This is probabl better as a warning, people can promote it to an error, in `sem`/`transf` we can simply drop everything after a `void` expression in a statement list of any sort.

termermc:
> @saem already brought it up in the past, and since it's heavily related, I'm mentioning it here again. It would make sense to:
> 
>     * make a distinction between absence of value/type (`void`) and unitary value (`unit`), the latter also being a new built-in type
> 
>     * type non-terminal statements (such as assignments) as `unit`, and terminal statements (such as `break`) as `void`
> 
> 
> This would a have few significant benefits to the language and compiler:
> 
>     * the not-well integrated `noreturn` pragma is subsumed by the type system. That is, `void` as the return type means "doesn't return (normally)"
> 
>     * whether a procedure doesn't return is simply inferred by normal type inference
> 
>     * detection of unreachable code becomes much simpler and exhaustive (everything following a `void` expression is unreachable)
> 
>     * `unit` is usable as the type of parameters/variables (compared to `void`) and stays around until code generation, which can be useful for generic code
> 
> 
> If it's enforced that a `void` expression is the last expression in a block (i.e., that `(;return; 1)` is illegal), then the whole `unreachable_elim` module becomes obsolete (faster compiler, lower complexity, one less source of bugs).

Compiler logic aside, is the idea here that `unit` can be reasoned about similar to `null` can in JS (value to represent nothing) and `void` like `undefined` (there is no value)? How does this factor in with generics?

Take this code for example:

```nim
type MyObj[T] = object
  field: T

var obj1: MyObj[int]
var obj2: MyObj[void]

echo obj1.field.type
# int

echo obj2.field.type
# Error: undeclared field: 'prop' for type MyObj
```

Using `void` as the generic type for `MyObj` results in `field` disappearing completely, meaning any access is a compiler error. Would using `unit` mean still it exists? If those are the semantics we want to go with, I agree that it could be useful to add. With that said, would we still allow `void` as a generic value?

My experience with `unit` comes from Kotlin, where it effectively replaces the `void` type, and has almost the same semantics except for making it usable as a value type instead of just "nothing". You can't use `void` in Kotlin at all if I remember correctly, and it's only an internal type. If we were to go down the `unit` road, it'd probably be wise to constrain the use of `void` by users to avoid confusion. Like I don't think using `void` in generics is a good idea most of the time if you have `unit` you could use instead.

Also, what would the C representation of `unit` be? Can I `exportc` a proc whose signature contains `unit`? What does that look like? 

Just some miscellaneous thoughts.

zerbina:
> This is probabl better as a warning, people can promote it to an error, in `sem`/`transf` we can simply drop everything after a `void` expression in a statement list of any sort.

Yeah, especially when quickly iterating and wanting to seal off some code fragment, a hard error would be annoying.

The idea with the error was to prevent code like `float((;return; 1))` or `call(0, (;return; 1), 2)` from compiling, because "correcting" such statements/expressions is a bit more complex (see  `unreachable_elim`).

However, if `(;return; 1)` is typed as `void` (and not as `int` as it is now), then `float((;return; 1))` wouldn't compile anyway, as `void` and `float` are incompatible, so I don't think reporting a hard error for trailing statements is really necessary.

zerbina:
> Compiler logic aside, is the idea here that `unit` can be reasoned about similar to `null` can in JS (value to represent nothing) and `void` like `undefined` (there is no value)? How does this factor in with generics?

`unit` and `void` would be similar to JS' `null` and `undefined`, though the former two are *types* while the latter are *values*.

> Would using `unit` mean still it exists?

Yep, `unit` stays around until code generation. It's a concrete type that can be use like any concrete type (for parameters, locals, globals, fields, etc.). For example, the following would work:
```nim
type Tup = tuple[a, b: unit]

proc f(x: unit) =
  discard

var t: Tup
let a = t.a
let b = a
f(a)
```

> If we were to go down the `unit` road, it'd probably be wise to constrain the use of `void` by users to avoid confusion. Like I don't think using `void` in generics is a good idea most of the time if you have `unit` you could use instead.

There was a recent [discussion](https://github.com/nim-works/nimskull/pull/1319#issuecomment-2130496516) about `void` vs. `unit` in the context of generics. To summarize, it'll likely become error to attempt substituting a field/parameter with `void`, though the details aren't fully worked out yet.

The only place where the `void` type is allowed is for the return type specification, where it designates the procedure as not returning (subsuming the `.noreturn` pragma, as I mentioned earlier). If no type is specified, the return type is inferred to be `unit`.   

> Also, what would the C representation of `unit` be? Can I `exportc` a proc whose signature contains `unit`? What does that look like?

There would be no C representation for the `unit` type, as it doesn't reach the code generator. Much like `typedesc` and `static` parameters, `unit` parameters are also gone by the time code generation is hit.

Consider:
```nim
var a: (int, unit, int)
var b: (int, int)

proc f(a: int, b: unit, c: int) = discard
let x = f(1, other(), 2)
```
For the code generators, the code would look like this:
```nim
var a: (int, int)
var b: (int, int)

proc f(a, c: int) = discard
other()
f(1, 2)
```

> Just some miscellaneous thoughts.

Thanks, I appreciate the input.

zerbina:
The `unreachable_elim` module name looks a bit out of place, but I think that's fine, given its workaround-like nature. Having said that, if anyone has a better name, I'm not opposed to changing it.

saem (approved):
I wasn't able to come up with a better name for `unreachable_elim`, even so, it's descriptive.

PR #1333, opened 2024-06-05
WIP: rewrite the C code generator

zerbina:
This PR is a from-scratch rewrite of the C code generator. It's the
successor to #424, incorporating many ideas and lessons-learned from
the latter. The concrete goals are to:
* significantly speed up the C code generation phase
* having a significantly simpler and more modular implementation
* getting rid of the architectural concessions that had to be made for
  the legacy C code generator
* improving the output C code (smaller amount, more efficient, easier
  to inspect)

A big focus is on using a data-oriented design. The planned
architecture is as follows:
* the code generator gathers the MIR bodies of all alive entities
* each MIR body is translated to an intermediate representation (the
  CIR)
* the intermediate representation is an AST-like IR closely resembling
  C's syntax
* the code generator itself doesn't handle tracking of dependencies
  (types, procs, etc.) -- the orchestrator does
* once the CIR for all bodies in the program is generated, they're
  rendered into textual C code and assembled into complete C files,
  which are then written to disk

Using an intermediate IR has multiple benefits over directly
translating to text:
* smaller memory footprint, especially for larger expressions with many
  duplicate identifiers
* less temporary string allocations; one single character buffer can be
  used at the end
* the IR it can store semantic information (e.g., referenced procedure,
  types, etc.)
* rendering is decoupled from translation (more modular, and the code
  generator doesn't have to worry about formatting)
* small peephole syntax optimizations are possible (e.g., collapsing
  `(*a).x` to `a->x`)

The new code generator operates directly on the MIR, no full in-between
IR (like the CGIR) is used. Prior to code generation, the MIR is
lowered to a degree where:
* translating it to C code is straightforward
* the code generator doesn't have to discover or synthesize new alive
  entities

**The PR is a work in progress. While the broad design and direction is
likely final already, many details are most likely going to change.**

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


zerbina:
For helping me in the development, I've added a small profiling utility: the `measure` module, which provides the `measure` template. Once the PR is finished, I'll remove it again.

Various key procedure are instrumented with `measure` template, counting the number of runs, time taken, and - optionally - the number of allocations/deallocations. When compilation is done, the counters are both echoed and written to an SQLite database.

zerbina:
I've implemented the basic code orchestration flow. It's rather simple, especially compared to before: the orchestrator runs common backend processing (`backends`), which produces the events driving code generation (`processEvent`). Once all events were processed (and thus all code generated), the TU's dependencies (types, procedure declarations, inline procedures, etc.) are gathered, and everything is rendered as C code.

Except for registering some new identifiers, the new C code generator itself doesn't modify any global state -- it simply takes a MIR body and outputs the CIR for it. No new entities are registered with the MIR environment. This makes it possible to handle `inline` procedures fully within the orchestrator, which in turn renders the complicated "first seen in module" tracking in `backends` obsolete.

Finally, the new orchestrator also addresses the issue of ostensibly small changes in one module causing many modules (sometimes the whole project) to be recompiled, something which got exacerbated when the code generation orchestrator architecture was first introduced.

zerbina:
The next big blocker is the missing type IR for the MIR. It's the basis for lowering `seq`, `string`, `openArray`, and `set`, and there are also some type-related architectural unknowns (e.g., how type dependencies of a procedure are tracked exactly) that can only be figured out with the MIR type IR in place.

With the MIR's type IR in place, further work on both `cgen` and the various type-related lowerings that the new `cgen` needs can happen concurrently. Lowering `seq`, `string`, etc. doesn't depend on the new code generator, meaning that it can implemented in separate PRs.

zerbina:
The list of things that need to be moved outside the code generator (via separate PRs) before the rewrite can be completed:
* [ ] creation of RTTI objects
* [ ] RTTI header initialization for `object`s and `ref`s
* [x] lowering complex `finally`s into `goto`s (#1468)
* [ ] turning non-integer `cast`s into `memcpy`s (#1411)
* [ ] lowering of `set` operations
* [ ] lowering of `seq` and `string` operations
* [ ] lowering of `openArray` operations
* [ ] stack-trace handling (i.e., injecting `nimln`, `nimFrame`, and `popFrame` calls)

zerbina:
Superseded by #1625.

PR #1334, opened 2024-06-06
mir: remove the tree delimiter nodes

zerbina:
## Summary

Instead of marking the end of a subtree with a dedicated node, the
number of child nodes in a subtree is now stored in the subtree's root
node. The idea is to:
* reduce the memory footprint of MIR trees
* make the storage format more streamlined
* simplify some traversal

## Details

The idea of using dedicated end nodes originated at a time where:
* fast back-/up-wards MIR traversal was required
* statements could be nested

Neither is the case anymore, and replacing the delimiter `mnkEnd` nodes
with storing the length has some benefits:
* significantly less nodes per `MirTree` (reduced memory usage, faster
  scanning, faster copying, etc.)
* `O(1)` length lookup
* lower cost of changing a subtree root node, or replacing it with its
  single child node
* simpler recursive traversal, as the end nodes don't have to be
  ignored/considered anymore

The downside of removing the delimiters nodes are that:
* slightly higher tree construction cost: the number of child nodes
  needs to be kept track of
* walking "upwards" (i.e., towards the root) in the tree has a higher
  cost

Ultimately, the upsides outweigh the downsides, especially since
walking the tree upwards is seldomly done.

### Adjustments

For memory efficiency, some subtree root nodes (like `mnkPathPos`)
stored extra information in the root node. Since all subtree root nodes
now need to store the length, this is no longer possible, and the
information has to be moved into separate nodes:
* for call trees, whether the call has side-effects is stored as an
  immediate value (using the new `mnkImmediate` node) in the tree's
  first slot
* for field/positional access, the field/position is stored in the
  tree's second slot

For easier processing of `mnkObjConstr`/`mnkRefConstr` trees, the
`mnkBinding` is introduced, which groups the `mnkField` and associated
expression together in a subtree.

The rest of the changes are about:
* updating the `MirBuilder` routines to track the child node count and
  patch the root nodes
* updating the construction of MIR trees in various places
* updating the various tree traversal logic that expected the old
  layout


zerbina:
Using a `-d:release` compiler and compiling the current `devel` (96cefc4fd6d78ff5fd26555db3bf30d8257f8d5e) compiler, I get the following benchmarks results:
```
hyperfine --warmpup 1 nim c --compileOnly --verbosity:0 --hints:off --warnings:off compiler/nim.nim

before  15.218 s ±  0.181 s
now     14.989 s ±  0.169 s
```

saem:
FYI: minor updates to PR description (typo/grammar)

compiler/backend/cgirgen.nim
```diff
@@ -96,8 +96,7 @@ func enter(t: MirBody, cr: var TreeCursor): lent MirNode {.inline.} =
   result = get(t, cr)
 
 func leave(t: MirBody, cr: var TreeCursor) =
-  assert t.code[cr.pos].kind == mnkEnd, "not at the end of sub-tree"
-  inc cr.pos
+  discard "obsolete; a no-op for backwards compatibility"
```
saem: Is the intention to deprecate and remove the function in follow-ups?

compiler/backend/cgirgen.nim
```diff
@@ -383,13 +385,12 @@ proc argToIr(tree: MirBody, cl: var TranslateCl,
...
-  # the inner node may be a tag node
   n = tree.get(cr)
   case n.kind
-  of mnkTag:
-    # it is one, the expression must be an lvalue
-    result = (true, lvalueToIr(tree, cl, cr))
-    leave(tree, cr)
+  of mnkImmediate:
+    # the argument must be a 'name' one, ignore the tag and expect
+    # a lvalue expression
```
saem:
```suggestion
    # an lvalue expression
```

compiler/mir/mirconstr.nim
```diff
@@ -287,21 +300,33 @@ func add*(bu: var MirBuilder, id: SourceId, n: sink MirNode) =
...
 func emitFrom*(bu: var MirBuilder, tree: MirTree, n: NodePosition) =
   ## Emits the sub-tree at `n` within `tree` into `bu`'s front buffer.
   bu.front.apply(bu.currentSourceId)
   bu.front.nodes.add toOpenArray(tree, int n, int tree.sibling(n)-1)
   bu.front.cursor = bu.front.len
+  inc bu.front.num # register the new node
+
+proc start*(bu: var MirBuilder, n: sink MirNode): Context {.inline.} =
+  ## Starts a subtree with a root node of `kind`. Must be paired with a call
+  ## to `stop <#stop,MirBuilder,NodePosition>`_.
```
saem: Did you mean `finish`?

compiler/mir/mirconstr.nim
```diff
@@ -287,21 +300,33 @@ func add*(bu: var MirBuilder, id: SourceId, n: sink MirNode) =
...
+  bu.add(n)
+  result = Context(num: bu.front.num, pos: NodeIndex(bu.front.len - 1))
+  bu.front.num = 0
+
+proc finish*(bu: var MirBuilder, saved: Context) {.inline.} =
+  ## Finishe the manual subtree and restores the `saved` context previously
```
saem:
```suggestion
  ## Finish the manual subtree and restores the `saved` context previously
```

saem (approved):
A few minor comments, but I think it's good to go.

compiler/backend/cgirgen.nim
zerbina: Yep. If you prefer the procedure being removed now, I could do so as part of this PR, though I think it staying is fine, given `cgirgen`'s soon-to-be-obsolete nature.

compiler/mir/mirconstr.nim
zerbina: Yes -- `finish` was originally named `stop`. The documentation generator should really start to report errors in the case of missing link targets.

zerbina:

compiler/backend/cgirgen.nim
saem: I was wondering if you wanted to mark it `deprecated`, but your broader point about the entire layer being removed makes this less relevant overall. Good to merge either way in my opinion.

zerbina: I've removed it now.

PR #1335, opened 2024-06-07
fix(sem): generic distinct types getting the same hooks

zerbina:
## Summary

Fix instantiating a generic distinct type with generic parameters not
used in the body not resulting in a unique type internally, resulting
in all instances thereof sharing the very same generic hook
instantiations.

## Details

There were two issues in `semtypinst`:
1. no new type instance was created for `tyDistinct` types when there
   are no generic parameters used in the body
2. no new symbol being created for the `tyDistinct` when instantiating
   it

As a consequence of point 1, `sameType` and `hashType` treated the
instantiated `tyDistinct` types as being the same. `hashType` is used
by hook lifting for fetching the canonical type, so all `tyDistinct`
types resulting from a generic type with phantom type information got
the same set of hooks.

For point 2, the consequence was that the symbol for instantiated
generic distinct types always pointed to the original `tyGenericBody`,
not the instantiated type.

Same as with `tyObject` types (which are also nominal types), a new
`tyDistinct` instance is always created during instantiation, and a new
symbol is created for it.

### Tests

* the `tphantom_distinct_hooks` test is added, which is largely a copy
  of `tphantom_object_hooks`
* the `tphantom_distinct_types` typerel test is added for preventing
  future regressions. It already succeeded previously

### DIFF
--- a/compiler/sem/semtypinst.nim
+++ b/compiler/sem/semtypinst.nim
@@ -543,7 +543,7 @@ proc instantiate(cl: var TReplTypeVars, t: PType): PType =
     result = instCopyType(cl, t)
     result[0] = replaceTypeVarsT(cl, t[0])
 
-    # always create a new symbol, even if
+    # the type also needs a proper symbol
     result.sym = copySym(t.sym, nextSymId cl.c.idgen)
     result.sym.flags.incl sfFromGeneric
     result.sym.owner = t.sym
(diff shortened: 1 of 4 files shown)


compiler/sem/semtypinst.nim
```diff
@@ -537,6 +537,18 @@ proc instantiate(cl: var TReplTypeVars, t: PType): PType =
 
     if cl.c.computeRequiresInit(cl.c, result):
       result.flags.incl tfRequiresInit
+  of tyDistinct:
+    # same as with object types, create a new type instance even if the body
+    # doesn't change during instantiation
+    result = instCopyType(cl, t)
+    result[0] = replaceTypeVarsT(cl, t[0])
+
+    # always create a new symbol, even if
```
saem: sentence fragment, but also perhaps a duplicate of the comment above?

saem (approved):
minor comment issue, but otherwise it's more consistent now.

PR #1336, opened 2024-06-07
openssl: new wrapper targeting OpenSSL 3+

### LINKED ISSUE #1160: add OpenSSL 3.0+ support
Now that OpenSSL 1.1.1 is scarce in the wild, we should start updating the wrappers and stdlib to OpenSSL 3.0 at minimum to make sure tooling using `std/net` runs without additional packages on most *nix.

It should be safe to drop OpenSSL <1.1.1 support, as those are no longer widely used and does not support the latest TLSv1.3 standard.

We should drop LibreSSL support for now as well, as we don't have the resource to test against that library, and platforms that use LibreSSL also supports OpenSSL.

This is more or less a short term goal to get existing tools running well.

## Useful References

- [PEP 644](https://peps.python.org/pep-0644/): Python decision to only support 1.1.1 and above.

alaviss:
## Summary
* what changed and how?
* why are we changing it?

## Details
* anything else

Fixes https://github.com/nim-works/nimskull/issues/1160

---

## Notes for Reviewers
Details to be done after everything is working

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


lib/pure/net.nim
```diff
@@ -826,35 +807,45 @@ when defineSsl:
...
+      raiseSSLError("X.509 certificate encoding failed")
+    result.setLen(encLen)
+
+  proc getPeerCertificates*(sslHandle: ptr SSL): seq[Certificate] {.since: (1, 1).} =
     ## Returns the certificate chain received by the peer we are connected to
     ## through the OpenSSL connection represented by `sslHandle`.
     ## The handshake must have been completed and the certificate chain must
     ## have been verified successfully or else an empty sequence is returned.
-    ## The chain is ordered from leaf certificate to root certificate.
+    ## The chain is ordered from leaf certificat to root certificate.
```
saem:
```suggestion
    ## The chain is ordered from leaf certificate to root certificate.
```

minor typo

PR #1338, opened 2024-06-07
fix: wrong code generation for synthesized object hooks

### LINKED ISSUE #1337: The pointer passed to super type hook is not converted to the super type
GCC 14 enforces a new language change that prohibit implicit conversion between incompatible pointer types in C. Some codegen that used to work no longer does.

### Example

```nim
type X = object of RootObj
proc `=destroy`(x: var X) = discard

type Y = object of X
let y = Y()
```

### Actual Output

```
@mtest.nim.c: In function ‘eqdestroy___test_6’:
@mtest.nim.c:127:28: error: passing argument 1 of ‘eqdestroy___test_2’ from incompatible pointer type [-Wincompatible-pointer-types]
  127 |         eqdestroy___test_2(dest);
      |                            ^~~~
      |                            |
      |                            tyObject_Y__mnLzXY1lPCpgpET0c8sT3w *
@mtest.nim.c:62:87: note: expected ‘tyObject_X__oKPdmARtpzUKjBTlF51xpw *’ but argument is of type ‘tyObject_Y__mnLzXY1lPCpgpET0c8sT3w *’
   62 | N_LIB_PRIVATE N_NIMCALL(void, eqdestroy___test_2)(tyObject_X__oKPdmARtpzUKjBTlF51xpw* x);
      |                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
@mtest.nim.c: In function ‘eqtrace___test_15’:
@mtest.nim.c:130:27: error: passing argument 1 of ‘eqtrace___test_27’ from incompatible pointer type [-Wincompatible-pointer-types]
  130 |         eqtrace___test_27(dest, env);
      |                           ^~~~
      |                           |
      |                           tyObject_Y__mnLzXY1lPCpgpET0c8sT3w *
@mtest.nim.c:63:86: note: expected ‘tyObject_X__oKPdmARtpzUKjBTlF51xpw *’ but argument is of type ‘tyObject_Y__mnLzXY1lPCpgpET0c8sT3w *’
   63 | N_LIB_PRIVATE N_NIMCALL(void, eqtrace___test_27)(tyObject_X__oKPdmARtpzUKjBTlF51xpw* dest, void* env);
      |                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
```

### Possible Solution

- Cast from Y to X before calling the super type hook
- Disable the new warnings in `nimbase.h`

### References
- [Porting to GCC 14#Type checking on pointer types](https://gcc.gnu.org/gcc-14/porting_to.html#incompatible-pointer-types)

zerbina:
## Summary

Fix a bug with synthesis of lifetime-tracking hooks for object types
using inheritance, which showed up as a C compiler error when using
recent versions of clang or gcc.

Fixes https://github.com/nim-works/nimskull/issues/1337.

## Details

When synthesizing the hooks for object types using inheritance, the
destination accessor expression wasn't wrapped in an object conversion
prior to emitting the parent object handling.

Without the conversion, the synthesized hook passed along its own
parameter to the call of the super type's hook as-is. This is wrong
in general (the argument's type doesn't match the parameter's), but
didn't cause concrete issues until gcc and clang started to disallow
implicit conversions between incompatible pointer types.

### DIFF
--- a/tests/lang_objects/destructor/tsuper_type_destructor_bug.nim
+++ b/tests/lang_objects/destructor/tsuper_type_destructor_bug.nim
@@ -1,0 +1,23 @@
+discard """
+  description: '''
+    Regression test for an internal typing issue with synthesized hooks for
+    objects using inheritance
+  '''
+  targets: c js vm
+"""
+
+type
+  Parent = object of RootObj
+    x: int
+  Sub = object of Parent
+
+proc `=destroy`(x: var Parent) =
+  # ensure that the parameter is at least accessible at run-time
+  doAssert x.x == 1
+
+# the destroy hook for `Sub` is synthesized by the compiler
+
+proc test() =
+  var x = Sub(x: 1)
+
+test()
--- a/compiler/sem/liftdestructors.nim
+++ b/compiler/sem/liftdestructors.nim
@@ -196,7 +196,12 @@ proc fillBodyObj(c: var TLiftCtx; n, body, x, y: PNode; enforceDefaultOp: bool) 
 
 proc fillBodyObjTImpl(c: var TLiftCtx; t: PType, body, x, y: PNode) =
   if t.len > 0 and t[0] != nil:
-    fillBody(c, skipTypes(t[0], abstractPtrs), body, x, y)
+    # also apply the operation to the super type. An up-conversion is required
+    # for proper typing
+    let
+      base = skipTypes(t[0], abstractPtrs)
+      obj = newTreeIT(nkObjUpConv, x.info, base, x)
+    fillBody(c, base, body, obj, y)
   fillBodyObj(c, t.n, body, x, y, enforceDefaultOp = false)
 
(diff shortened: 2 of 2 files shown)


tests/lang_objects/destructor/tsuper_type_destructor_bug.nim
alaviss, line 1: I'm thinking that we should either add a `{.passC.}` or a config file to turn on `-Werror=incompatible-pointer-types` for this test

zerbina: The config file approach is already used by another test, but it's brittle and I'd rather not embrace it further. 

I'll make a separate PR that adds debug functionality for pretty-printing the initial MIR, and then this test can make sure the MIR for the destroy hook is sensible.

PR #1340, opened 2024-06-08
sem: keep shape of array constructors in typed AST

zerbina:
## Summary

Explicit indices in array constructors now stay in the AST after
typing. Since macros are able to observe this, this change is a
**breaking change**.

## Details

Keeping the explicit indices is also necessary for being able to
retype the AST after it was typed once (without the index, array
constructors for arrays not starting at zero would have a
different type afterwards).

Instead of dropping the explicit index together with the parent
`nkExprColonExpr` node, both are kept by `semArrayConstr`, and then
later discarded by `semfold` (during constant evaluation) and `transf`,
where they are no longer needed.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -772,32 +772,19 @@ proc semArrayConstr(c: PContext, n: PNode, flags: TExprFlags): PNode =
     result.sons.setLen(n.len)
     for i, it in n.pairs:
       # first, analyse the index expression (if one exist)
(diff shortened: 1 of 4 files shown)


zerbina:
While #1192 is not my focus at the moment, the change was easy enough that I figured I implement it right away.

saem (approved):
looks good to me. outside of normalization, which this wasn't, i think preserving the shape is the way to go.

PR #1343, opened 2024-06-08
compiler: add the `--showir` switch

zerbina:
## Summary

* replace the debug defines for enabling IR printing with the proper
  `--showir` switch
* the switch has two forms: `--showir:<ir>` (enable globally) and
  `--showir:<ir>:<name>` (enable for procedures with given name)
* the supported IR names are: `transf`, `mir_in`, `mir_out`, `cgir`,
  and `vm`
* the old debug defines are removed (they now do nothing)

Using a proper compiler switch brings some usability improvements
(validation, style-insensitive names), and will make extending the
functionality in the future a bit easier.

## Details

* the enabled names are stored `ConfigRef` using in a `StringTableRef`,
  to allow for style-insensitive name lookup
* a `StringTableRef` only supports string-to-string mappings so
  only the showing of one IR per name is supported at the moment
* the `mir_out` option is new: it shows the MIR as it looks like right
  before `cgirgen`
* `mir_in` (corresponding to the previous `nimShowMir` define), now
  shows pretty-printed MIR, instead of the tree representation
* the `debug.rst` documentation mentioning the obsolete defines is
  updated 

### Misc

* a bug with the MIR pretty-printer, where rendering locals with no
  name crashed the compiler, is fixed

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -390,7 +390,7 @@ proc translate*(id: ProcedureId, body: PNode, graph: ModuleGraph,
 
   echoInput(graph.config, prc, body)
   result = generateCode(graph, env, prc, config.tconfig, body)
-  echoMir(graph.config, prc, result)
+  echoMir(graph.config, prc, result, env)
 
   # now apply the passes:
   process(result, prc, graph, idgen, env)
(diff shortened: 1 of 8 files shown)


saem (approved):
This is a really nice debug facility, cool.

PR #1345, opened 2024-06-13
sem: don't analyze dependent operands to `static`

zerbina:
## Summary

Makes instantiations such as `Generic[someProc(T)]` work, where
`T` is some generic parameter (and receiving type parameter is a
`static` one) -- the `someProc(T)` expression is typed once `T` is
substituted, making the behaviour consistent with `array` and `range`.

## Details

Fully analyzing (i.e., with `semExpr`) arbitrary expressions that might
reference unresolved type variables generally doesn't work. Operands to
generic invocations were eagerly analyzed (in `matchesAux`), usually
resulting in errors when the operand needs to be a `static` value.

Disallowing expressions dependent on unresolved type variables in this
context would be a regression, since some expressions are special-cased
to work (such as `sizeof(T)`).

To address the problem, before typing the operand, it's first checked
whether it depends on unresolved type variables. If it does, the
expression is treated as a generic expression, with a `tyFromExpr`
assigned as its type -- otherwise it's analyzed as usual.

If the operand has an unknown type (`tyFromExpr`), it's wrapped in a
conversion to a type derived from the generic parameter's constraint,
to make sure the operand types later. If no concrete type can be
derived from the constraint, a type mismatch is reported.

To not convolute `matchesAux` further, the argument matching where the
callee is a `tyGenericBody` is moved into the new `matchesType`. It
does largely the same as `matchesAux`, but with everything not
applicable to types removed (such as the varargs handling).

### DIFF
--- a/compiler/sem/sigmatch.nim
+++ b/compiler/sem/sigmatch.nim
@@ -3120,6 +3120,127 @@ proc matchesAux(c: PContext, n, nOrig: PNode, m: var TCandidate, marker: var Int
   m.error.firstMismatch.pos = a
   m.error.firstMismatch.formal = formal
 
+proc matchesType(c: PContext, n: PNode, m: var TCandidate,
+                 marker: var IntSet) =
+  ## Matches the arguments taken from invocation expression `n` against the
+  ## ``tyGenericBody`` callee and fills `m` with the results. `marker` is
+  ## updated with the matched-against formal positions.
+  m.state = csMatch # until proven otherwise
+  m.error.firstMismatch = MismatchInfo()
+
+  # pre-pass: make sure the AST is valid. `n` is production AST, so it can be
+  # modified in-place
+  var hasError = false
+  for i in 1..<n.len:
+    if n[i].kind == nkExprEqExpr:
+      let (ident, err) = considerQuotedIdent(c, n[i][0])
+      if err != nil:
+        n[i][0] = err
+        hasError = true
+      else:
+        n[i][0] = newIdentNode(ident, n[i][0].info)
+
+  if hasError:
+    # abort early
+    m.state = csNoMatch
+    m.call = c.config.wrapError(n)
+    return
+
+  m.call = newNodeI(n.kind, n.info, m.callee.len)
+  m.call[0] = n[0]
+
+  var f = 0
+  var i = 1
+  var formal: PSym
+
+  while i < n.len:
+    # select the formal parameter:
+    var operand: PNode
+    case n[i].kind
+    of nkExprEqExpr:
+      # explicit parameter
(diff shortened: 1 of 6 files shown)


zerbina:
The `genericParams` typetrait macro is extremely brittle, currently relying on input AST modifications. I haven't tracked down where and why exactly some input AST is now no longer modified, but something seems to have changed there.

I'm working on a separate PR for making `genericParams` work such that it relies less on implementation details.

tests/lang_callable/generics/tdependent_operands_to_static_2.nim
```diff
@@ -0,0 +1,21 @@
...
+  errormsg: "cannot instantiate Type"
+  line: 21
+  knownIssue: '''
+    `Type`s generic parameter is not detected as containing a `static`, thus
+    full analysis is not disabled (which subsequently fails)
+  '''
+"""
+
+type
+  Type[T: string | static float] = object ## must be an int or static float
```
saem:
```suggestion
  Type[T: string | static float] = object ## must be a string or static float
```

compiler/sem/sigmatch.nim
```diff
@@ -3120,6 +3120,127 @@ proc matchesAux(c: PContext, n, nOrig: PNode, m: var TCandidate, marker: var Int
...
+          #      wholly incompatible, later analysis of the conversion will
+          #      yield an error. Non-exact matches where the types have a
+          #      "convertible" relationship will not result in an error
+          arg = newTreeI(nkConv, operand.info,
+                         newNodeIT(nkType, operand.info, formal.typ.base),
+                         operand)
+          arg.typ = makeTypeFromExpr(c, copyNodeWithKids(arg))
+      else:
+        # we don't know the argument's type, nor can we enforce that it'll
+        # matche the formal type later -> type mismatch
```
```suggestion
        # match the formal type later -> type mismatch
```

saem (approved):
The previous usage of `semExpr`, just led to a big layering violation, using the generic pre-pass is the way to go.

tests/lang_callable/generics/tdependent_operands_to_static_2.nim
saem: Unless you meant `static int or float`, because of the convertable relationship?

zerbina: Your suggested change is correct. I've originally used `int | static float`, but `int` and `float` can be converted to each other, which could make the test work accidentally.

zerbina: I made a minor change to your suggestion (replacing `an` with `a`) but otherwise applied it as is.

PR #1346, opened 2024-06-14
macros: keep instantiation parameters with `getType`

### LINKED ISSUE #1223: getType doesn't reveal complete generic arguments
### Specification

`getType` doesn't reveal complete generic arguments even for concrete types.

### Example

```nim
import std/macros

type
  A[B, C] = object
    x: B
    y: C

  P[B, C] = proc(bar: A[B, C])

proc foo(bar: A[int, float]) =
  discard

macro bug[B, C](a: P[B, C]) =
  echo treeRepr(getType(a))

bug(P[int, float] foo)
bug(foo)
```

### Actual Output

```
BracketExpr
  Sym "proc"
  Sym "void"
  Sym "A"
BracketExpr
  Sym "proc"
  Sym "void"
  Sym "A"
```

### Expected Output

```
BracketExpr
  Sym "proc"
  Sym "void"
  BracketExpr
    Sym "A"
    Sym "int"
    Sym "float"
BracketExpr
  Sym "proc"
  Sym "void"
  BracketExpr
    Sym "A"
    Sym "int"
    Sym "float"
```

### Possible Solution

A workaround is to use `getTypeInst` but I don't see why `getType` shouldn't work.

### Additional Information

```
Nimskull Compiler Version 0.1.0-dev.21251 [linux: amd64]

Source hash: 5aca007cd7ef41a1128fbe6d6eb667ce612f847c
Source date: 2024-03-06

active boot switches: -d:release -d:danger⏎
```

zerbina:
## Summary

`getType` now keeps instantiation parameters, which were previously
irrecoverably lost. This also allows retrieving the instantiation
parameters when only having access to an alias of the instantiated
type.

Fixes #1223.

## Details

Expanding `tyGenericInst` to `(BracketExpr <Callee> ...)` is the only
way of keeping the type information intact, since `tyGenericInst` do
not have symbols attached to them (in which case translating to an
`nnkSym` would be preferable, as it'd keep the AST flat).

Existing tests are adjusted.

### Aliases

A second goal of changing `getType` behaviour for `tyGenericInst` is to
allow retrieving the instantiation parameters in the presence of
type aliases. Given:
```nim
type
  Generic[A, B] = object
  Alias = Generic[int, string]
```

it was previously not possible to retrieve the instantiation
parameters for `Generic[int, string]` when given the expression
`Alias`, but now it is (with `getType`).

### Other

The internal `enumutils.enumFullRange` macro relied on the old
`getType` behaviour. To restore its functionality, it now uses
`getTypeImpl`.

### DIFF
--- a/compiler/vm/vmdeps.nim
+++ b/compiler/vm/vmdeps.nim
@@ -159,17 +159,13 @@ proc mapTypeToAstX(cache: IdentCache; t: PType; info: TLineInfo;
     for i in 0..<t.len:
       result.add mapTypeToAst(t[i], info)
   of tyGenericInst:
-    if inst:
-      if allowRecursion:
-        result = mapTypeToAstR(t.lastSon, info)
-      else:
-        result = newNodeX(nkBracketExpr)
-        #result.add mapTypeToAst(t.lastSon, info)
-        result.add mapTypeToAst(t[0], info)
-        for i in 1..<t.len-1:
-          result.add mapTypeToAst(t[i], info)
+    if inst and allowRecursion:
+      result = mapTypeToAstR(t.lastSon, info)
     else:
-      result = mapTypeToAstX(cache, t.lastSon, info, idgen, inst, allowRecursion)
+      result = newNodeX(nkBracketExpr)
+      result.add mapTypeToAst(t[0], info)
+      for i in 1..<t.len-1:
+        result.add mapTypeToAst(t[i], info)
   of tyGenericBody:
     if inst:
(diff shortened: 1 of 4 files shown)


zerbina:
For testing whether absolute URLs are the culprit for issues not being auto-closed, I've referenced the issues via only the number. Let's see what happens.

saem:
I'm going to apply the `bug` tag as the non-preserving behaviour seems wrong from the get go, and the issuing being fixed is also marked `bug`

saem (approved):

PR #1349, opened 2024-06-17
openssl: support OpenSSL 3.0 and above

### LINKED ISSUE #1160: add OpenSSL 3.0+ support

alaviss:
## Summary
The only ABI change between version 1.1.1 and 3.0 for our usage is 
`SSL_get_peer_certificate`  being splitted into 
`SSL_get1_peer_certificate`  (which is compatible with the prior symbol)
and  `SSL_get0_peer_certificate` .

This PR modifies  `SSL_get_peer_certificate`  in the wrapper to use the
new symbol if available. No changes in other programs are required.

## Details
*  `SSL_get_peer_certificate`  will now select either 
`SSL_get1_peer_certificate`  or  `SSL_get_peer_certificate`  depending
on which symbols are available.
* DLL names for OpenSSL 3.x has been added for macOS, Windows and
Linux.
* The symbols used for certificate verification are no longer
unconditionally hidden on Windows. They were hidden previously as Nim
ships old OpenSSL 1.0 which did not have these symbols.

Fixes https://github.com/nim-works/nimskull/issues/1160

### DIFF
--- a/lib/wrappers/openssl.nim
+++ b/lib/wrappers/openssl.nim
@@ -63,19 +63,19 @@ elif useWinVersion:
         DLLUtilName* = "libeay32.dll"
   elif defined(cpu64):
     const
-      DLLSSLName* = "(libssl-1_1-x64|ssleay64|libssl64).dll"
-      DLLUtilName* = "(libcrypto-1_1-x64|libeay64).dll"
+      DLLSSLName* = "(libssl-3-x64|libssl-1_1-x64|ssleay64|libssl64).dll"
+      DLLUtilName* = "(libcrypto-3-x64|libcrypto-1_1-x64|libeay64).dll"
   else:
     const
-      DLLSSLName* = "(libssl-1_1|ssleay32|libssl32).dll"
-      DLLUtilName* = "(libcrypto-1_1|libeay32).dll"
+      DLLSSLName* = "(libssl-3|libssl-1_1|ssleay32|libssl32).dll"
+      DLLUtilName* = "(libssl-3|libcrypto-1_1|libeay32).dll"
 
   from std/winlean import SocketHandle
 else:
   when defined(osx):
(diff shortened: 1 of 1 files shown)


lib/wrappers/openssl.nim
```diff
@@ -796,10 +796,14 @@ when defined(nimHasStyleChecks):
 
 # Certificate validation
 # On old openSSL version some of these symbols are not available
-when not defined(nimDisableCertificateValidation) and not defined(windows):
-
-  proc SSL_get_peer_certificate*(ssl: SslCtx): PX509{.cdecl, dynlib: DLLSSLName,
-      importc.}
+when not defined(nimDisableCertificateValidation):
```
zerbina, line 799: Is the procedures below this point becoming available on Windows intentional? If so, could you elaborate on the reason behind it?

zerbina (commented):
I left a question regarding the removal of the `not defined(windows)` condition, but otherwise this looks good to me.

lib/wrappers/openssl.nim
alaviss: The reason is that there is no reason for them not to be available.

For mainline Nim, they are hidden away due to how Nim used to ship old OpenSSL 1.0 DLLs, which doesn't have these symbols, and was never enabled even after they upgraded to 1.1.1.

I'll add this to the PR description.

zerbina (approved):

PR #1351, opened 2024-06-18
make the REPL (`nim secret`) work again

zerbina:
## Summary

Properly configure the compiler when running the `nim secret`
command, making the compiler built-in REPL usable again.

## Details

* the compiler is configured for the VM backend when the
  `cmdInteractive` command is active
* `echo` output is no longer prefixed by `>>>`
* a test is added to ensure the basic REPL functionality works
* the standard output stream is flushed after writing `>>>` or `...`,
  so that the test is able to read the output

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -2617,8 +2617,7 @@ To create a stacktrace, rerun compilation with './koch temp $1 <file>'
       )
 
     of rintEchoMessage:
-      result = if conf.cmd == cmdInteractive: ">>> " & r.msg
-               else:                          r.msg
+      result = r.msg
 
     of rintCannotOpenFile, rintWarnCannotOpenFile:
       result = "cannot open file: $1" % r.file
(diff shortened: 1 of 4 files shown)


compiler/front/cli_reporter.nim
```diff
@@ -2617,8 +2617,7 @@ To create a stacktrace, rerun compilation with './koch temp $1 <file>'
       )
 
     of rintEchoMessage:
-      result = if conf.cmd == cmdInteractive: ">>> " & r.msg
-               else:                          r.msg
+      result = r.msg
```
saem, line 2620: Note to self remove reporting from this code path.

tests/compilerfeatures/trepl.nim
```diff
@@ -0,0 +1,54 @@
+discard """
+  description: "Ensure that the basic REPL functionality works"
+  targets: native
```
saem, line 3: Should a timeout be added because this test is susceptible to bugs causing it to get stuck on reads?

saem (approved):
Looks good, minor suggestion about adding a timeout to a test.

tests/compilerfeatures/trepl.nim
zerbina: Good call, I'll add a timeout. It currently won't protect against the test getting stuck, however, since testaement's `timeout` feature is only an "after the fact" check.

PR #1352, opened 2024-06-19
ci: replace vcpkg-action with simpler setup-vcpkg

alaviss:
## Summary
johnwason/vcpkg-action depends on various other actions that are now
out-of-date and is using deprecated NodeJS 16. Given that vcpkg
configuration and bootstrap is very simple, replace our usage with a
simpler homemade setup-vcpkg instead.

## Details
* Added a new  `setup-vcpkg`  action. This action is rather simple and
consist only of exporting environment variables to configure vcpkg.
* Replace usage of `johnwason/vcpkg-action` with `setup-vcpkg`.

### DIFF
--- a/.github/actions/setup-vcpkg/action.yml
+++ b/.github/actions/setup-vcpkg/action.yml
@@ -1,0 +1,62 @@
+name: Setup vcpkg
+description: Clone and setup vcpkg
+inputs:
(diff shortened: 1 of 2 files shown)


saem (approved):
It's a tad arcane but not a problem

PR #1354, opened 2024-06-19
Produce a meaningful error message for invalid module name

### LINKED ISSUE #720: Crash when compiling a-b.nim

alaviss:
## Summary
While the AST form of the error generated by  `modules/modulepaths`  is
handled in  `cli_reporter` , the symbol form was not, creating terrible
error messages such as:

    Error: invalid module name '<nil tree>'

This PR makes cli_reporter aware of the symbol form generated by 
`modules/modules` , fixing the bad error message.

## Details
*  `cli_reporter`  now handles the symbol form of 
`rsemInvalidModuleName`  correctly.
*  `modules/modules`  now raises  `rsemInvalidModuleName`  with line
information, allowing user and tools to know which file the error
originated from.

Fixes https://github.com/nim-works/nimskull/issues/720

### DIFF
--- a/tests/errmsgs/tinvalid-name.nim
+++ b/tests/errmsgs/tinvalid-name.nim
@@ -1,4 +1,6 @@
 discard """
-  description: Test that a proper error message is printed for a module with invalid name
+  description: '''
+    Test that a proper error message is printed for a module with invalid name
+  '''
   errormsg: "invalid module name: 'tinvalid-name'"
 """
--- a/tests/errmsgs/tinvalid-name.nim
+++ b/tests/errmsgs/tinvalid-name.nim
@@ -1,0 +1,4 @@
+discard """
+  description: Test that a proper error message is printed for a module with invalid name
+  errormsg: "invalid module name: 'tinvalid-name'"
+"""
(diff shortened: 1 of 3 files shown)


tests/errmsgs/tinvalid-name.nim
```diff
@@ -0,0 +1,4 @@
+discard """
+  description: Test that a proper error message is printed for a module with invalid name
```
zerbina:
```suggestion
  description: '''
    Test that a proper error message is printed for a module with invalid name
  '''
```
Nit-pick: line length.

zerbina (approved):
Looks good to me, thank you for reviving the fix!

I think the current description is okay as is, but for future PRs, especially those with a more outside-facing change (like this fix), could you focus the summary more on the user-facing changes? Along the lines of "this was the visible previous behaviour, this is how it behaves now", with the implementation details going into the "Details" section.

PR #1355, opened 2024-06-19
fix incorrect typing in closure-iterator C code-gen

### LINKED ISSUE #1353: closure iterator that captures the outer proc environment generates incompatible pointer assignments
This is an another pointer type compatibility issue, found while compiling testament.

### Example

```nim
proc main() =
  let x = 0
  iterator c(): int {.closure.} =
    yield x

  var iter = c

main()
```

### Actual Output

```
@mtest.nim.c: In function ‘c__test_3’:
@mtest.nim.c:231:20: error: assignment to ‘tyObject_Env_testdotnim_c___QRjx23R1ttwg6yg9a8vWR2w *’ from incompatible pointer type ‘tyObject_Env_testdotnim_main___lcRWa04p7qwyxS5SaoH5UA *’ [-Wincompatible-pointer-types]
  231 |         colonenvP_ = (tyObject_Env_testdotnim_main___lcRWa04p7qwyxS5SaoH5UA*) ClE_0;
      |                    ^
```

### Expected Output

Compilation should be successful

### References

- [Porting to GCC 14#Type checking on pointer types](https://gcc.gnu.org/gcc-14/porting_to.html#incompatible-pointer-types)

zerbina:
## Summary

Fix the C code generator emitting incorrectly typed code for closure
closure-iterators, which recent, stricter C compilers report errors
for.

Fixes https://github.com/nim-works/nimskull/issues/1353.

## Details

When computing the environment for a closure iterator where the
iterator closes over some outer locals, there's already a symbol in
the hidden parameter slot, using the environment type of the enclosing
routine.

The symbol is retyped during closure iterator environment computation,
but the *node* referencing it was not updated. `cgen` used the MIR
local's type for the definition, but the *node's* type (which was still
the enclosing procedure's environment type) for the cast, resulting in
a pointer type mismatch.

The node referencing the symbol is now also updated, and an assertion
to make sure the types match is added in `cgen`.

### DIFF
--- a/compiler/backend/cgen.nim
+++ b/compiler/backend/cgen.nim
@@ -683,6 +683,7 @@ proc closureSetup(p: BProc, prc: PSym) =
   # prc.ast[paramsPos].last contains the type we're after:
   var ls = lastSon(prc.ast[paramsPos])
   p.config.internalAssert(ls.kind == nkSym, prc.info, "closure generation failed")
+  p.config.internalAssert(ls.typ == ls.sym.typ) # sanity check
   var env = ls.sym.position + 1 # parameters start at ID 1
 
   let n = newLocalRef(LocalId(env), ls.info, ls.typ)
--- a/compiler/sem/lambdalifting.nim
+++ b/compiler/sem/lambdalifting.nim
@@ -769,6 +769,8 @@ proc liftLambdas*(g: ModuleGraph; fn: PSym, body: PNode;
       param = getHiddenParam(g, fn)
     else:
       param.typ = t # replace with the correct type
+      # also update the symbol *node's* type
+      fn.ast[paramsPos][^1].typ = t
 
     prepareInnerRoutines(d, idgen, t, fn.info)
     # the environment instance is not setup here; that's done at the iterator's


saem (approved):
Sorry, if I'm missing something, but wouldn't adding the previously broken code (per the bug), with some modification, as a test case be sufficient? It's not perfect, but if we had that scenario tested then we would have caught this, is my thinking.

zerbina:
> Sorry, if I'm missing something, but wouldn't adding the previously broken code (per the bug), with some modification, as a test case be sufficient?

No worries, it's somewhat non-obvious. The problem affects all closure closure-iterators (regardless of their complexity), so there's already test coverage for the bug (e.g., in `tests/closure/tclosure.nim`). Incompatible pointer assignments being reported as errors was only introduced in GCC 14 (previously those resulted in just warnings), which CI is seemingly not using yet, so no C compiler errors are reported by CI at the moment.

Now, adding a test where the GCC or Clang warning is promoted to an error is *possible*, but GCC, at least, has the quirk that promoting a warning to an error only works if C compiler warnings are not disabled across the board (which the NimSkull compiler does by default, by passing `-w`), so the test has to *override* the NimSkull C compiler options.

This is exactly what I did in [this](https://github.com/nim-works/nimskull/pull/1095) PR, but said approach has the significant downside that as soon as the NimSkull compiler starts requiring some mission-critical flags to be passed along to the C compiler, the tests will cease to work, so I'd rather not embrace this approach further.

saem:
Oh wow, that's a much bigger problem -- thanks for the detailed explanation.

PR #1358, opened 2024-06-26
deprecate the `.codegenDecl` pragma

zerbina:
## Summary

The `.codegenDecl` pragma is going to be removed eventually. For a
transition period, a deprecation warning is now reported when using it.


zerbina:
The earliest the built-in `.codegenDecl` pragma can be removed is when the `asyncjs` module is reworked to not depend on it. At most, the pragma can stay around until the new JavaScript code generator is implemented, which I'd currently estimate as being around 3 to 4 months away.

saem (approved):

PR #1359, opened 2024-06-26
asyncjs: don't use JavaScript `async`

zerbina:
## Summary

* `asyncjs` no longer uses JavaScript's `async`
* `result` in `.async` procedures now works like it does everywhere else
* `method`s can no longer use `.async`

## Details

* the goal is removing the usage of the deprecated `.codegenDecl` pragma
* routines marked with the `.async` pragma internally use a closure
  iterator that is `then` chained to a promise object when `await`ing

---

## To-Do
- [ ] make imported exceptions properly catchable within closure iterators (separate PR)
- [ ] re-add support for returning `Future`s from `.async` procedures

## Notes for Reviewers
* prior discussion about this can be found [here](https://gist.github.com/zerbina/d09b55530a596bc9e2cb12a0460ca074?permalink_comment_id=5012850#gistcomment-5012850)
* an important prerequisite for the new JavaScript code generator 

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/lib/js/asyncjs.nim
+++ b/lib/js/asyncjs.nim
@@ -8,25 +8,14 @@
 
 ## This module implements types and macros for writing asynchronous code
 ## for the JS backend. It provides tools for interaction with JavaScript async API-s
-## and libraries, writing async procedures in Nim and converting callback-based code
-## to promises.
+## and libraries, writing async procedures in |NimSkull| and converting
+## callback-based code to promises.
 ##
-## A Nim procedure is asynchronous when it includes the `{.async.}` pragma. It
-## should always have a `Future[T]` return type or not have a return type at all.
-## A `Future[void]` return type is assumed by default.
+## A |NimSkull| procedure is asynchronous when it includes the `{.async.}`
+## pragma. If the return type is not `Future[T]`, it is opaquely turned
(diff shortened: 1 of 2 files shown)


lib/js/asyncjs.nim
```diff
@@ -52,8 +41,10 @@
...
 ## ========================
 ##
-## Nim currently generates `async/await` JavaScript code which is supported in modern
-## EcmaScript and most modern versions of browsers, Node.js and Electron.
+## |NimSkull| generates JavaScript code that uses the ``Promise`` and ``Error``
+## APIs, both which are supported by most modern versions of browsers, Node.js
+## and Electron.
+##
 ## If you need to use this module with older versions of JavaScript, you can
 ## use a tool that backports the resulting JavaScript code, as babel.
```
saem, line 49:
```suggestion
## use a tool that backports the resulting JavaScript code, such as babel.
```

saem: I know it's not strictly part of the change, but it was right there; not a requirement to update it.

saem (approved):
Minor doc suggestion

PR #1360, opened 2024-06-26
lower `string`/`float` case statements with MIR pass

zerbina:
## Summary

Use an MIR pass to lower `string` and `float` case statements into `if`
chains, replacing the C code generator logic.

## Details

* add the `lowerCase` pass to `mirpasses`
* it's currently only enabled for the C target
* the lowering is a straightforward MIR port of the code generation
  previously implemented by `cgen` for `string`/`float` case statements
* the `cgen` handling for `string` and `float` case statements is
  removed

### DIFF
--- a/compiler/mir/mirpasses.nim
+++ b/compiler/mir/mirpasses.nim
@@ -664,7 +664,7 @@ proc lowerCase(tree: MirTree, graph: ModuleGraph, env: var MirEnv,
   ## Lowers case statements with string or float selectors. For large string-
   ## case statements, a hash-table optimization is used.
   const stringCaseThreshold = 8
-    # above X strings a hash-switch for strings is generated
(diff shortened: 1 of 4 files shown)


compiler/mir/mirpasses.nim
```diff
@@ -652,6 +659,137 @@ proc lowerMove(tree: MirTree, changes: var Changeset) =
     else:
       discard "not relevant"
 
+proc lowerCase(tree: MirTree, graph: ModuleGraph, env: var MirEnv,
+               changes: var Changeset) =
+  ## Lowers case statements with string or float selectors. For large string-
+  ## case statements, a hash-table optimization is used.
+  const stringCaseThreshold = 8
+    # above X strings a hash-switch for strings is generated
```
```suggestion
    ## above X strings a hash-switch for strings is generated
```

saem (approved):

PR #1361, opened 2024-06-28
sem: correctly type empty containers in macro calls

zerbina:
## Summary

Internal-only issue. Empty container values passed to `static` macro/
template parameters used empty container types. Now they're converted
to the formal parameter's type.

## Details

* wrap the arguments to `static` macro/template parameters in
  implicit conversions, if needed for the match
* use post-match argument fitting for macro calls
* for a smaller impact, template calls still don't use post-match
  argument fitting
* remove the `vmserialize` workaround for empty set types (which no
  longer reach there now)

### DIFF
--- a/compiler/sem/sigmatch.nim
+++ b/compiler/sem/sigmatch.nim
@@ -2317,8 +2317,9 @@ proc paramTypesMatchAux(m: var TCandidate, f, a: PType,
           else:                # value on the type
             arg.typ.n
 
-        # XXX: this is super hacky; it's the non-template/macro logic from
-        #      below but in worse
+        # XXX: the implicit conversion handling is duplicated from the non-
+        #      template/non-macro path. Template and macro arguments shouldn't
+        #      be special-cased like this
         case r
         of isEqual: n
         of isGeneric:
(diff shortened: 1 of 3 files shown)


compiler/sem/sigmatch.nim
```diff
@@ -2311,10 +2311,25 @@ proc paramTypesMatchAux(m: var TCandidate, f, a: PType,
...
-        else:                # value on the type
-          arg.typ.n
+        let n =
+          if arg.typ.n.isNil:  # no value on the type
+            argSemantized
+          else:                # value on the type
+            arg.typ.n
+
+        # XXX: this is super hacky; it's the non-template/macro logic from
+        #      below but in worse
```
saem: Sentence fragment?

Not sure what was intended with it, so I can't provide a suggestion.

saem (approved):
Minor typo(?) in a comment, but otherwise good to go.

compiler/sem/sigmatch.nim
zerbina: No worries! What I wanted to communicate here is that the handling of implicit conversions is duplicated from the non-template/non-macro path below, but without all the intricacies. I'll reword the comment.

saem: Thanks

PR #1362, opened 2024-06-29
nversion: read version from text file

alaviss:
## Summary
Instead of statically encoding the version as a structure, read it from
a text file. This allows tools to modify the version string without
creating invalid NimSkull, and allow for easier rearrangement of the
code in the future.

## Details
* Added validation for version string suffix according to SemVer rules.
* Compiler (base) version is now stored in text form in 
`compiler/version.txt` .


compiler/utils/nversion.nim
```diff
@@ -32,19 +43,24 @@ func parse*(s: string): Version =
...
     result.major, result.minor, result.patch, result.suffix
-  ):
+  ) or not result.suffix.isValidSuffix:
     raise newException(ValueError):
-      "Invalid version string: " & s
+      "Invalid version string: " & s.escape()
+
+const CompilerVersionSuffix* {.strdefine.} = ""
+  ## The suffix to attach to the compiler version. This is meant to be
+  ## declared by build tools to signify development version for example.
```
zerbina, line 52:
```suggestion
const
  CompilerVersionSuffix* {.strdefine.} = ""
    ## The suffix to attach to the compiler version. This is meant to be
    ## declared by build tools to signify development version for example.
```
Very minor thing that doesn't necessarily have to change. Using a block from the start makes adding more constants to it easier in the future.

zerbina (approved):
Looks good to me. I left one small, optional suggestion to use a `const` block.

compiler/utils/nversion.nim
alaviss: I'd rather that new constants be put at the end of the file like its currently is :P

PR #1363, opened 2024-06-29
remove the `.goto` pragma/feature

zerbina:
## Summary

The `.goto` pragma is neither documented nor tested, and most of the
compiler is unaware of it. Everything related to the feature is removed
from the compiler.

## Details

* the `.goto` pragma could be used on variables of enum type,
  turning all assignments to the variable into jumps to branches of an
  associated `case` statement
* remove the `sfGoto` symbol flag
* remove the reports associated with the `.goto` pragma
* keep the `wGoto` enum; it's part of the C keyword set

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -433,7 +433,6 @@ const
 
   sfNoForward*     = sfRegister       ## forward declarations are not required (per module)
   sfExperimental*  = sfOverriden      ## module uses the .experimental switch
-  sfGoto*          = sfOverriden      ## var is used for 'goto' code generation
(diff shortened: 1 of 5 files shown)


saem (approved):
I'd perhaps add a comment to `wGoto` saying it:

` ## was previously used but isn't currently, keeping it because it'll likely be used in the future`

Or something like that.

zerbina:
Apologies, I forgot about the double-duty of some `TSpecialWord`. `wGoto` is part of the C keywords, and therefore *must* stay around for the C name mangling to continue mangling locals named `goto`.

I've updated the PR message accordingly.

PR #1364, opened 2024-06-29
publisher: use actions/download-artifact directly

alaviss:
## Summary
Instead of depending on a third-party actions to download built
artifacts, use GitHub's CLI to search for the wanted ID and first-party
actions to download the wanted artifacts.

## Details
* Replaced usage of  `dawidd6/action-download-artifact`  with 
`actions/download-artifact` .

### DIFF
--- a/.github/workflows/publisher.yml
+++ b/.github/workflows/publisher.yml
@@ -28,47 +28,64 @@ jobs:
       # Publish action needs a checkout
       - uses: actions/checkout@v4
 
+      - name: Obtain latest successful run id
+        id: finder
+        run: |
+          run_id=$(gh run list \
+            -c "$COMMIT" \
+            -w "$WORKFLOW" \
+            -s "$CONCLUSION" \
+            --limit 1 \
+            --json databaseId \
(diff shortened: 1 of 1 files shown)


.github/workflows/publisher.yml
```diff
@@ -28,47 +28,64 @@ jobs:
...
+        run: |
+          run_id=$(gh run list \
+            -c "$COMMIT" \
+            -w "$WORKFLOW" \
+            -s "$CONCLUSION" \
+            --limit 1 \
+            --json databaseId \
+            --jq '.[].databaseId')
+          if [[ -z $run_id ]]; then
+            echo "::error::Could not found any CI run for commit $COMMIT"
```
zerbina:
```suggestion
            echo "::error::Could not find any CI run for commit $COMMIT"
```

zerbina (approved):
Looks good to me, though CI is not my main area of expertise, so it's better if @saem still takes a look.

saem (approved):
Not enough to drop the dependency, but still good.

PR #1365, opened 2024-06-30
mirbodies: track next label in `MirBody`

zerbina:
## Summary

Track the ID to use for new labels in `MirBody`. This greatly reduces
the overhead of `Changeset` creation, since the source body doesn't
have to be scanned.

## Details

* track the next label ID with `MirBody.nextLabel`
* on creation, the `Changeset` next label ID is initialized from the ID
  stored with the `MirBody`
* add the `createBody` routine for creating a proper `MirBody` from a
  `MirBuilder` instance
* use `createBody` to replace manual `MirBody` construction

### Performance

Compile times are effectively unaffected. A small improvement of a few
milliseconds is measurable, but it's insignificant. However, using more
batches in the future will make the improvement more pronounced.

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -420,13 +420,13 @@ proc produceFragmentsForGlobals(
       bu.setSource(m.add(n))
       bu.subTree mnkScope: discard
 
-  func finish(bu: sink MirBuilder, m: var SourceMap, n: PNode
-             ): auto {.nimcall.} =
+  func finish(bu: sink MirBuilder, n: PNode, body: var MirBody) {.nimcall.} =
+    var map = move body.source
     if bu.front.len > 0:
-      bu.setSource(m.add(n))
+      bu.setSource(map.add(n))
       bu.subTree mnkEndScope: discard
-    # we're creating a body here, so there is no list of locals yet
-    result = finish(bu, default(Store[LocalId, Local]))
+
+    body = createBody(bu, map)
 
   var init, deinit, threadDeinit: MirBuilder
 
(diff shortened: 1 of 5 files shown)


saem (approved):
(just checking my understanding, change looks good)

IIUC, we can rely on this being safe because this is only happening on body creation (after we've got a finalized builder), so we can't run into a situation where multiple active builders would lead to non-unique id assignment.

zerbina:
> IIUC, we can rely on this being safe because this is only happening on body creation (after we've got a finalized builder), so we can't run into a situation where multiple active builders would lead to non-unique id assignment.

Yep, your understanding is correct for the most part.

So what's happening is the following:
1. a `MirBuilder` is created for generating a body (e.g., in `mirgen`)
2. labels are allocate with the builder and it's internal `nextLabel` value is thus increased
3. once done, `createBody` is called, which initializes the `MirBody` with the builder's `nextLabel` value

Later, when a `Changeset` is created for modifying a `MirBody`:
1. the `Changeset` is initialized with the body's `nextLabel` value
2. when creating a temporary `MirBuilder`, said builder is initialized with the changeset's ` nextLabel` value
3. new labels are allocated with the temporary builder
4. upon closing the temporary builder, it's updated `nextLabel` value is propagated back to the changeset
5. upon applying the changeset to the body, the body's `nextLabel` field is updated from the changeset

While the `Changeset` API doesn't guard against it, having multiple temporary `MirBuilder` active at the same time is disallowed, meaning that non-unique ID assignment cannot (legally) happen.

PR #1367, opened 2024-07-01
sem: proper error when alignof is called on generic type

theSherwood:
## Summary
*  `alignof`  calls on generic arguments resulted in 
`Fatal: Internal unreachable code executed`  (C backend)
* This fix causes the compiler to report a proper error in such cases: 
`<type> is not a concrete type` 

## Details
* To do this we have to check that the argument to  `alignof`  is a
concrete type before it reaches the backend.
* We add `semAlignOf` to do this.
* Handles overloaded `alignof`.

### DIFF
--- a/compiler/sem/semmagic.nim
+++ b/compiler/sem/semmagic.nim
@@ -62,20 +62,13 @@ proc semSizeOf(c: PContext, n: PNode): PNode =
                                             magic: mSizeOf))
 
 proc semAlignOf(c: PContext, n: PNode): PNode =
-  case n.len
-  of 2:
-    n[1] = semExprWithType(c, n[1])
-    if containsGenericType(n[1].typ):
-      # report the type, not the typedesc
-      n[1] = c.config.newError(n[1], PAstDiag(kind: adSemTIsNotAConcreteType,
-                                              wrongType: n[1].typ[0]))
-      result = n
-    else:
-      n.typ = getSysType(c.graph, n.info, tyInt)
-      result = foldAlignOf(c.config, n, n)
+  if containsGenericType(n[1].typ):
+    # report the type, not the typedesc
+    n[1] = c.config.newError(n[1], PAstDiag(kind: adSemTIsNotAConcreteType,
+                                            wrongType: n[1].typ[0]))
+    result = c.config.wrapError(n)
   else:
-    result = c.config.newError(n, PAstDiag(kind: adSemMagicExpectTypeOrValue,
-                                            magic: mAlignOf))
+    result = foldAlignOf(c.config, n, n)
 
 type
   SemAsgnMode = enum asgnNormal, noOverloadedSubscript, noOverloadedAsgn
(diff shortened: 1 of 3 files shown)


compiler/sem/semmagic.nim
```diff
@@ -422,7 +438,7 @@ proc magicsAfterOverloadResolution(c: PContext, n: PNode,
   of mSizeOf:
     result = semSizeOf(c, n)
   of mAlignOf:
-    result = foldAlignOf(c.config, n, n)
+    result = semAlignOf(c, n)
```
zerbina:
```suggestion
    result = semAlignOf(c, n)
```
Trailing whitespace.

compiler/sem/semmagic.nim
```diff
@@ -61,6 +61,22 @@ proc semSizeOf(c: PContext, n: PNode): PNode =
...
+      # report the type, not the typedesc
+      n[1] = c.config.newError(n[1], PAstDiag(kind: adSemTIsNotAConcreteType,
+                                              wrongType: n[1].typ[0]))
+      result = n
+    else:
+      n.typ = getSysType(c.graph, n.info, tyInt)
+      result = foldAlignOf(c.config, n, n)
+  else:
+    result = c.config.newError(n, PAstDiag(kind: adSemMagicExpectTypeOrValue,
+                                            magic: mAlignOf))
```
zerbina:
```suggestion
  if containsGenericType(n[1].typ):
    # report the type, not the typedesc
    n[1] = c.config.newError(n[1], PAstDiag(kind: adSemTIsNotAConcreteType,
                                            wrongType: n[1].typ[0]))
    result = c.config.wrapError(n)
  else:
    result = foldAlignOf(c.config, n, n)
```
The `alignof` case is slightly different from `sizeof`. `semSizeOf` is called from `semMagic`, where `n` is something that hasn't been typed or checked in any way yet (hence the `semExprWithType` and `n.len` check).

`semAlignOf` is only called from `magicsAfterOverloadResolution`, which is itself called after overload resolution was successful, meaning that:
* the call is fully typed already
* the number of arguments is correct

Something I missed during review of the `sizeof` PR is the error wrapping. If a `sem` procedure returns AST that contains an error, the AST must be wrapped in an error node itself (this is what the `wrapError` call is for).

zerbina (changes requested):
Thank you for the PR!

Checking the `alignof` call doesn't need the analysis `semSizeOf` does, since the former magic is never processed before overload resolution. I left a suggestion for addressing this.

As for the PR message, could you restructure it a bit? For a bugfix such as this PR, the general direction is that:
* the summary describes what happened before (i.e., generic `alignof` arguments resulted in miscompilation or compiler crashes) and what happens now (a proper compiler error is reported)
* the details sections describes how the bug is fixed (i.e., by checking that the argument is not a generic type, to prevent the call from reaching the backend)
* statements about relations to other PRs go into the "Notes For Reviewers" part (which is not included in the final commit message)

zerbina (approved):

PR #1370, opened 2024-07-04
ci: split docs deployment from publisher and deploy directly to pages

alaviss:
## Summary
Documentation will now be built using a separated workflow that will be
run whenever publisher finishes. This allows us to use the latest page
deployment code regardless of the commit that triggered publisher.

## Details
* Added  `get`  sub-command to  `release_manifest` , which allows one to
extract binary archive name from a release manifest.
* "Generated docs" artifacts are now removed. Instead the generated
documentation will be obtained from the release tarball for Linux amd64.
This allows docs publishing to run regardless of whether the artifact
for the latest release has expired or not.
* Documentation deployment will now skip  `gh-pages`  branch and deploy
directly using  `actions/deploy-pages` . This allows us to drop the
heavyweight  `gh-pages`  branch and remove the latency between push and
page being deployed.
* Documentation deployment can now be triggered on-demand in case of a
prior failure.
* Documentation deployment will now always be done using latest code in 
`devel` .
* Development documentation is also pushed to  `/devel`  sub-folder.
This is preliminary setup for versioned docs in the future.


doc/ci.rst
```diff
@@ -245,6 +245,21 @@ Only one instance of this pipeline might be run at any given time, due to its
...
+"Deploy documentation"
+----------------------
+
+As the name suggests, this pipeline constructs and deploys the documentation
+website. This pipeline is dependent on the
+`Release manifest tool <#tools-release-manifest-tool>`_.
+
+Unlike other pipelines, the code ran will always be that of the latest commit
+in ``devel`` branch. This meant that the code should always be aware of
+differences between versions of |NimSkull| being deployed.
```
saem:
```suggestion
Unlike other pipelines, the code run will always be that of the latest commit
in ``devel`` branch. This means that the code should always be aware of
differences between versions of |NimSkull| being deployed.
```

I believe this is what you meant.

saem (approved):
Couple typos in the docs you might want to look at, otherwise good.

alaviss:
/merge

Here we go

PR #1373, opened 2024-07-07
clean up symbol error reporting

zerbina:
## Summary

* don't report "cannot call" error when using symbol of routines
  where the definition has an error
* don't report "only identifier found is an error" when using `let`/
  `var`/`const` symbol where the initializer expression has an error
* these errors don't provide any additional information and only
  clutter compiler and nimsuggest output

## Details

* use a common erroneous symbol handling path in `semSym`
* don't turn `skLet`, `skVar`, and `skConst` into `skError`s when the
  definition AST has an error
* use the silent `adWrappedSymError` diagnostic for erroneous callees
* remove the `CalleeHasAnError` diagnostic and report
* add tests to make sure the new error reporting behaviour works
* adjust the `tpointerprocs.nim`, which is affected by the different
  error reporting

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1154,7 +1154,6 @@ type
     adSemUndeclaredField
     adSemCannotInstantiate
     adSemWrongNumberOfGenericParams
-    adSemCalleeHasAnError
     # sem
     adSemExpressionHasNoType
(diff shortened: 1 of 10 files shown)


saem (approved):
`adWrappedSymError` is so much better, really wish this approach occurred to me when I was working on this.

PR #1380, opened 2024-07-12
changelog: move "Other" into categories

alaviss:
While this will no longer allow the category to be collapsible, it will
allow the category to be omitted when nothing goes there.

This addresses the empty "Other" section as seen in
https://github.com/nim-works/nimskull/releases/tag/0.1.0-dev.21377

### DIFF
--- a/.github/changelog.json
+++ b/.github/changelog.json
@@ -80,9 +80,13 @@
           "exhaustive": true
         }
       ]
+    },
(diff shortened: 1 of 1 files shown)


alaviss:
/generate-changelog

Perfect time to test this out

saem (approved):

PR #1382, opened 2024-07-12
mir: add a dedicated IR for types

zerbina:
## Summary

* implement a dedicated type IR for the MIR and back-end phase
* use the new IR in parts of the C code generator
* use a new name mangling scheme for types that doesn't rely on
  signature hashes

## Details

The idea behind introducing a dedicated type IR for the MIR and back-
end phase is to:
* have a simpler and more regular IR during lowering and code
  generation
* decouple type lowering from code generation 
* make `PType` solely used by `sem` and `transf`

The key ideas for the type IR are that:
* same ID (`HeaderId`) means same type; there are no duplicates like
  with `PType`
* transformations and code generation have easy access to the various
  stages of a type's representation (including the `PType`)
* lineage to `PType`s is kept track of

The current design likely covers more ground than it ultimately needs
to, but this was a deliberate choice in order to ease the transition to
the new IR.

### Translation

The lowering and or fixes performed by `mirtypes` during the
translation of types intends to exactly match what was previously
implemented in `ccgtypes`.

So that the IR is well-tested, all `PType`s entering `mirgen` are
translated to the IR representation.

### Code generation

Most of the existing is still oblivious to the new type IR, with only
the type emission in the C code generation. Some analysis that was
previously performed on `PType`s is ported to operate on MIR type
descriptions.

The `mangling` module implements a semi-stable, reversible name
mangling for types. It's necessary since not all MIR types have an
originating-from `PType` (such as the internal payload type for `seq`s
and `string`s), meaning that the `sighash`-based mangling cannot be
used there.

### AST

Associating fields with names in the C code generator is now done via
the new type IR, meaning that the `locId` field on `TSym` is obsolete;
it's removed.

### Tests

* multiple `ccodecheck` test are adjusted to `int` now translating to a
  sized integer instead of `NI`
* due to usage of the to-MIR-type-translation,
  `tempty_typed_expressions_issues.nim` now also fails for JS

### DIFF
--- a/compiler/backend/mangling.nim
+++ b/compiler/backend/mangling.nim
@@ -63,7 +63,7 @@ proc mangle(g: ModuleGraph, typ: PType): string =
       # use the local ID as the name. "H" stands for "hidden"
       result = "H" & $s.itemId.item & "_M" & withLen(m)
     else:
-      # non-exported objec type's don't necessarily have unique names within a
+      # non-exported object types don't necessarily have unique names within a
       # module; the ID is included to produced a unique name
       result = "L" & mangleWithLen(s.name.s) & "_" & $s.itemId.item & "_M" &
                withLen(m)
--- a/compiler/backend/mangling.nim
+++ b/compiler/backend/mangling.nim
@@ -1,0 +1,199 @@
+## Implements the "name mangling" used by the code generators. Name mangling
+## needs to make sure that:
+## * the resulting identifier is valid according to the target's rules
+## * different entities get different names
+
+import
+  compiler/ast/[
+    ast_types,
+    ast_query
+  ],
+  compiler/mir/[
+    mirtypes,
+    mirtrees
+  ],
+  compiler/modules/[
+    modulegraphs
+  ],
+  compiler/utils/[
+    idioms
+  ]
+
+from compiler/backend/ccgutils import mangle
+
+const
+  CallConvToShort: array[TCallingConvention, string] = [
+    "ni", "st", "cd", "sa", "sy", "in", "ni", "fa", "cl", "nc"
+  ] ## every name must be unique and composed out of the same number of
+    ## letters
+
+proc mangle(g: ModuleGraph, env: TypeEnv, desc: TypeHeader): string
+proc mangle(g: ModuleGraph, env: TypeEnv, id: TypeId): string
+
+proc mangle(g: ModuleGraph, typ: PType): string =
+  ## For object types, returns a unique name based on the original symbol.
+  ## An empty string otherwise.
+  if typ.isNil:
+    return ""
+
+  proc withLen(s: string): string {.inline.} =
+    result = $s.len
+    result.add s
+
+  template mangleWithLen(s: string): string =
+    withLen(mangle(s))
+
+  if typ.kind == tyObject and typ.typeInst != nil:
+    # XXX: not very stable, but guaranteed to be unique and it also works
+    #      with the current IC mechanism
+    result = "I" & mangleWithLen(typ.typeInst[0].sym.name.s) & "_" & $typ.id
+  elif typ.kind == tyObject or (typ.sym != nil and sfImportc in typ.sym.flags):
+    # imported types are nominal types from the perspective of name mangling
+    let
+      s = typ.sym
+      m {.cursor.} = g.ifaces[typ.itemId.module].uniqueName
+
+    # the underscores are not strictly necessary; they're only added for
+    # better readability
+    if sfExported in s.flags:
+      # exported object types have unique name within their defining module.
+      # "G" stands for "global"
+      result = "G" & mangleWithLen(s.name.s) & "_M" & withLen(m)
+    elif sfAnon in s.flags:
+      # use the local ID as the name. "H" stands for "hidden"
+      result = "H" & $s.itemId.item & "_M" & withLen(m)
+    else:
+      # non-exported objec type's don't necessarily have unique names within a
+      # module; the ID is included to produced a unique name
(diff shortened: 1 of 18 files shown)


zerbina:
It's easier to not mess with tuple equality for the moment, so I went with using `sighashes` to unify generic object instantiations. This restores the original behaviour of `Generic[(int,)]` and `Generic[tuple[x: int]]` translating to the same nominal C type.

zerbina:
Okay, that fixed it. The `sighashes`-based approach merged types that the MIR considered to be different, since `sighashes` is unable to produce unique hashes for `tyObject` types in some cases (e.g., two object types with the same content in adjacent scopes). 

Merging instantiations of generic object types is now based on the MIR representation, meaning that there can be no disagreements.

compiler/backend/mangling.nim
```diff
@@ -0,0 +1,199 @@
...
+    # better readability
+    if sfExported in s.flags:
+      # exported object types have unique name within their defining module.
+      # "G" stands for "global"
+      result = "G" & mangleWithLen(s.name.s) & "_M" & withLen(m)
+    elif sfAnon in s.flags:
+      # use the local ID as the name. "H" stands for "hidden"
+      result = "H" & $s.itemId.item & "_M" & withLen(m)
+    else:
+      # non-exported objec type's don't necessarily have unique names within a
```
saem:
```suggestion
      # non-exported object types don't necessarily have unique names within a
```

compiler/mir/mirtypes.nim
```diff
@@ -24,26 +36,128 @@ import
...
+      ## items are indices (biased by 1) into the `headers` sequence
+    numStructs: int
+      ## the number of occupied slots in `structs`
+
+    canon: Table[HeaderId, TypeId]
+      ## maps headers of canonical type descriptions to their type symbol
+
+    instances: Table[(int, HeaderId), TypeId]
+      ## associates a generic type ID + instance body with a type symbol. This
+      ## is used for eliminating  same-shaped instantiations of a generic
```
saem:
```suggestion
      ## is used for eliminating same-shaped instantiations of a generic
```

compiler/mir/mirtypes.nim
```diff
@@ -62,58 +176,948 @@ const
...
+
+func `==`*(a, b: FieldId): bool {.borrow, inline.}
+func `==`(a, b: IntVal): bool {.borrow, inline.}
+
+func hash(env: TypeEnv, t: TypeHeader): Hash =
+  ## Computes the structural hash for the type `t`.
+  result = hash(t.kind)
+  case t.kind
+  of tkVoid, tkBool, tkChar, tkPointer, tkString, tkCstring:
+    discard "no content to hash"
```
saem:
```suggestion
    discard "no additional content to hash"
```

might be too pedantic, but it's not that bool, char, etc fields don't contribute to the hash, they do via `kind` in the initialization of `result`, it's just there is no more after that.

compiler/mir/mirtypes.nim
```diff
@@ -62,58 +176,948 @@ const
...
+  case a.kind
+  of tkVoid, tkBool, tkChar, tkPointer, tkString, tkCstring:
+    true
+  of tkInt, tkUInt, tkFloat, tkPtr, tkRef, tkVar, tkLent, tkSeq,
+     tkUncheckedArray, tkOpenArray, tkSet, tkIndirect, tkImported:
+    a.a == b.a
+  of tkArray:
+    a.a == b.a and a.b == b.b
+  of tkRecord, tkUnion, tkTaggedUnion:
+    if a.b - a.a == b.b - b.a: # same number of fields?
```
saem: might be worth introducing a `fieldCount` routine?

```nim
func fieldCount(t: TypeHeader): int {.inline.} =
  ## counts the number of fields, or params if `t` is a routine
  assert t.kind in {tkRecord, tkUnion, tkTaggedUnion, tkProc, tkClosure}
  t.b - t.a
```

compiler/mir/mirtypes.nim
```diff
@@ -62,58 +176,948 @@ const
...
+  result.header = TypeHeader(kind: tkRecord, size: size, align: align, b: 1)
+  result.fields.add RecField(typ: base, align: offset.int16)
+
+proc open(kind: TypeKind; size: IntVal, align: int16): RecordBuilder =
+  result.header = TypeHeader(kind: kind, size: size, align: align)
+
+proc openProc(env: TypeEnv, kind: TypeKind, conv: TCallingConvention,
+              ret: TypeId, isVarargs: bool): ProcBuilder =
+  ## Opens a builder for a procedure-like type.
+  result.header = TypeHeader(kind: kind, b: 1,
```
saem, line 522:
```suggestion
  assert kind in {tkProc, tkClosure}
  result.header = TypeHeader(kind: kind, b: 1,
```

Unless you don't want asserts in builders? I can suggest the rest if you'd like.

saem (approved):
A few minor suggestions no blockers.

I scanned the logic for any differences but didn't notice any, FWIW.

compiler/mir/mirtypes.nim
zerbina: Using assertions here wouldn't hurt.

compiler/mir/mirtypes.nim
zerbina: Yeah, as is, the intent is not entirely clear.

compiler/mir/mirtypes.nim
zerbina: I don't/didn't consider the type kind as content, hence the current wording.

zerbina:
Many thanks for the review, @saem.

PR #1383, opened 2024-07-14
IC: fix type-bound operator lookup

zerbina:
## Summary

Fix the compiler crashing in some contrived situations when using
`--ic:on`.

## Details

* populate the `canonTypes` table in `ic/cbackend` before starting the
  mid-/back-end processing
* type instances that weren't processed by `createTypeBoundOps` (e.g.,
  type instances for parameters) need the `canonTypes` lookup table for
  looking up the type-bound hooks
* the added test provides more information about the circumstances of
  the crash

### DIFF
--- a/tests/ic/tcanonical_types.nim
+++ b/tests/ic/tcanonical_types.nim
@@ -1,6 +1,6 @@
 discard """
   description: '''
-    Regression test for compiler crash due to missing hook attachements. The
+    Regression test for compiler crash due to missing hook attachments. The
     test is contrived and depends on compiler implementation details.
   '''
 """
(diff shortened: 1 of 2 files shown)


tests/ic/tcanonical_types.nim
```diff
@@ -0,0 +1,23 @@
+discard """
+  description: '''
+    Regression test for compiler crash due to missing hook attachements. The
```
saem:
```suggestion
    Regression test for compiler crash due to missing hook attachments. The
```

saem (approved):
Small typo, not a blocker

PR #1386, opened 2024-07-21
sem, pragmas: fix fatal pragma

starsiderfox:
## Summary

Using  `{.fatal:"...".}`  did not show any message to the user and it
didn't stop compilation.
This commit fixes both issues.

## Details

-  `adSemFatalError`  now carries a message just like 
`adSemCustomUserError` .
- All reports of `fatal` severity abort compilation.
- Testament now also consider fatal errors to be errors.
- A simple test case has been included.
-  `isCompilerFatal`  has been removed as it seems it lost its purpose.
The check for  `rextCmdRequiresFile`  seems to be unnecessary. A fatal
report is produced elsewhere (the only place where 
`rextCmdRequiresFile`  is used) before this function is ever called, so
it was redundant and it doesn't break existing code.

### DIFF
--- a/compiler/ast/reports.nim
+++ b/compiler/ast/reports.nim
@@ -224,6 +224,13 @@ func severity*(
       of repDebug:    report.debugReport.severity()
       of repExternal: report.externalReport.severity()
 
+func isCompilerFatal*(report: Report): bool =
+  ## Check if report stores fatal compilation error
+  case report.category:
+    of repSem:      report.semReport.severity() == rsevFatal
+    of repInternal: report.internalReport.severity() == rsevFatal
+    else: false # only sem and internal can be fatal
+
 func toReportLineInfo*(iinfo: InstantiationInfo): ReportLineInfo =
   ReportLineInfo(file: iinfo[0], line: uint16(iinfo[1]), col: int16(iinfo[2]))
 
--- a/compiler/ast/reports.nim
+++ b/compiler/ast/reports.nim
@@ -224,13 +224,6 @@ func severity*(
       of repDebug:    report.debugReport.severity()
       of repExternal: report.externalReport.severity()
 
-func isFatalSeverity*(report: Report): bool =
-  ## Returns whether the report severity is rsevFatal
-  case report.category:
-    of repSem:      report.semReport.severity() == rsevFatal
-    of repInternal: report.internalReport.severity() == rsevFatal
-    else: false # only sem and internal can be fatal
-
 func toReportLineInfo*(iinfo: InstantiationInfo): ReportLineInfo =
   ReportLineInfo(file: iinfo[0], line: uint16(iinfo[1]), col: int16(iinfo[2]))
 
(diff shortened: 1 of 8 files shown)


compiler/front/cli_reporter.nim
```diff
@@ -3428,6 +3427,13 @@ func astDiagToLegacyReport(conf: ConfigRef, diag: PAstDiag): Report {.inline.} =
         kind: rsemPragmaRecursiveDependency,
         sym: diag.userPragma,
         ast: diag.wrongNode)
+  of adSemFatalError:
```
zerbina: Please combine this with the `adSemCustomUserError` branch, by initializing `kind` with `kind` (i.e., `kind: kind`).

compiler/front/msgs.nim
```diff
@@ -294,7 +294,8 @@ proc errorActions(
     eh: TErrorHandling
   ): tuple[action: TErrorHandling, withTrace: bool] =
   result = (doNothing, false)
-  if conf.isCompilerFatal(report):
+  if conf.isCompilerFatal(report) or
+      (report.category == repSem and report.semReport.severity == rsevFatal):
```
zerbina: Please move this condition into `isCompilerFatal`.

zerbina (changes requested):
This is a very clean and concise fix. Thank you!

There are two small things that need adjustment, but otherwise this looks good.

zerbina:
Also, please make sure the PR message follows the guidelines. The first line works well as the summary; the details section should describe what changed internally (i.e., `adSemFatalError` now carries an error message, and sem reports with fatal severity are considered fatal errors).

compiler/ast/reports.nim
```diff
@@ -224,6 +224,13 @@ func severity*(
       of repDebug:    report.debugReport.severity()
       of repExternal: report.externalReport.severity()
 
+func isFatalSeverity*(report: Report): bool =
```
saem: This procedure is not a general query as to whether something is fatal, it's specifically the compiler's viewpoint to treat only internal and semantic errors as fatal, regardless of present and future code.

The general query is handled via the `severity` routine defined above.

starsiderfox: I consider this function to be of a different nature, unlike `isCompilerFatal`, it's not meant for deciding what the compiler does, as it merely reports whether the severity is fatal, nothing else. I added it because it may be slightly  more efficient than using `severity == rsevFatal`. I added this to the notes of the PR.

zerbina: Please remove `isFatalSeverity` and replace its usage with `severity == rsevFatal`. The (very minor) efficiency improvement is outweighed by the problematic introduction of another source of truth. If a new fatal non-sem/non-internal report is ever added, `isFatalSeverity` will silently stop doing what it should.

zerbina (changes requested):
Apart from `isFatalSeverity` (refer to the comment I left regarding it), this PR is good to be merged.

compiler/ast/reports.nim
zerbina: Thanks.

zerbina (approved):
Looks good from my side, thank you.

saem (approved):
LGTM

PR #1397, opened 2024-08-04
fix C compilation errors with closures and nested types

zerbina:
## Summary

Fix a bug with hook synthesis that led to C compiler errors when using
some closure procedures or `ref`s of non-top-level `object` type.

## Details

* `sighashes` now uses the symbol ID as the object representation for
  anonymous object/enum types and object/enum types not defined at the
  top-level
* this means that `hashType(x) != hashType(y)` when
  `sameType(x, y) == false`, for object and enum types
* as a consequence, unique hook procedures are used for `ref T` types
  that used the aforementioned types for `T`
* RTTI, which is assigned to types based on type hashes, also uses
  unique instances per object type now

Multiple same-shaped anonymous/non-top-level object types are *not*
merged into a single one by the C code generator, which due to all
sharing the same set of hook procedures, resulted in C compiler errors
due to implicit conversions between incompatible pointer types.

### DIFF
--- a/compiler/sem/sighashes.nim
+++ b/compiler/sem/sighashes.nim
@@ -164,28 +164,18 @@ proc hashType(c: var MD5Context, t: PType; flags: set[ConsiderFlag]) =
       else:
         c.hashSym(t.sym)
 
-      var symWithFlags: PSym
-      template hasFlag(sym): bool =
-        let ret = {sfAnon, sfGenSym} * sym.flags != {}
-        if ret: symWithFlags = sym
-        ret
-      if hasFlag(t.sym) or (t.kind == tyObject and t.owner.kind == skType and t.owner.typ.kind == tyRef and hasFlag(t.owner)):
-        # for `PFoo:ObjectType`, arising from `type PFoo = ref object`
-        # Generated object names can be identical, so we need to
-        # disambiguate furthermore by hashing the field types and names.
-        if t.n.len > 0:
-          let oldFlags = symWithFlags.flags
-          # Hack to prevent endless recursion
-          # xxx instead, use a hash table to indicate we've already visited a type, which
-          # would also be more efficient.
-          symWithFlags.flags.excl {sfAnon, sfGenSym}
-          hashTree(c, t.n, flags + {CoHashTypeInsideNode})
-          symWithFlags.flags = oldFlags
-        else:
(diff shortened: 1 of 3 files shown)


compiler/sem/sighashes.nim
```diff
@@ -164,28 +164,18 @@ proc hashType(c: var MD5Context, t: PType; flags: set[ConsiderFlag]) =
...
-          symWithFlags.flags = oldFlags
-        else:
-          # The object has no fields: we _must_ add something here in order to
-          # make the hash different from the one we produce by hashing only the
-          # type name.
-          c &= ".empty"
+      if t.sym.flags * {sfAnon, sfGenSym} != {} or
+         (t.kind == tyObject and t.owner.kind == skType and
+          tfRefsAnonObj in t.owner.typ.flags):
+        # one ore more of the following are true for the type:
```
saem:
```suggestion
        # one or more of the following are true for the type:
```

compiler/sem/sighashes.nim
```diff
@@ -164,28 +164,18 @@ proc hashType(c: var MD5Context, t: PType; flags: set[ConsiderFlag]) =
...
-          hashTree(c, t.n, flags + {CoHashTypeInsideNode})
-          symWithFlags.flags = oldFlags
-        else:
-          # The object has no fields: we _must_ add something here in order to
-          # make the hash different from the one we produce by hashing only the
-          # type name.
-          c &= ".empty"
+      if t.sym.flags * {sfAnon, sfGenSym} != {} or
+         (t.kind == tyObject and t.owner.kind == skType and
+          tfRefsAnonObj in t.owner.typ.flags):
```
saem, line 169: For the top level check wouldn't it make sense to check if the owning symbol is not an `skModule`, or would that be a problem for generic instantiations?

saem (approved):
Looks good.

compiler/sem/sighashes.nim
zerbina: Testing whether the owner is an `skModule` symbol wouldn't be able to detect the `Nested` type from `tseparate_hooks1.nim` (where the owner is a module symbol).

`sem` marks such types' symbols with the `sfGenSym` at the moment, so that they can be detected here. The only proper solution to this mess is getting rid of `sighashes`, really.

PR #1400, opened 2024-08-06
ci: build and test the compiler with GCC 14

zerbina:
## Summary

Add a CI workflow for making sure the most recent version of common C
compilers work. Only GCC 14 is tested at the moment.

## Details

* the in-beta 24.04 Ubuntu runner image is used, since it comes with
  GCC 14 installed
* the compatibility job builds the compiler, tools, and tests with GCC
  14
* two test cases that currently fail to compile (due to
  `-Werror=incompatible-pointer-types`) are split into standalone test
  files and marked as known issues
* the workflow is configured to run for push and PR events (except for
  in-draft PRs)

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -93,3 +93,12 @@ jobs:
       - name: Print all test errors
         if: failure()
         run: bin/nim r tools/ci_testresults
+
+  passed:
+    name: All C compatibility tests passed
+    needs: [test]
+    if: failure() || cancelled()
+    runs-on: ubuntu-latest
+
+    steps:
+      - run: exit 1
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -1,0 +1,95 @@
+name: Test C compiler compatibility
+on:
+  push:
+    branches-ignore:
+      - devel
+      # Github Merge Queue temporary branches
+      - gh-readonly-queue/**
+
+  pull_request:
+    # Only consider PRs to devel
+    branches:
+      - devel
+
+    # Type of events to run CI on
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+
+  merge_group:
+    # Test all additions to the merge queue
+
+# Run every script action in bash
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  binaries:
+    name: Build binaries with most recent GCC version
+    runs-on: ubuntu-24.04
+    # Don't run for draft PRs
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == 'false'
+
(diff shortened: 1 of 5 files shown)


saem (commented):
I did a quick once over, looks good thus far.

saem (approved):

.github/workflows/compatibility.yml
```diff
@@ -0,0 +1,95 @@
...
+          sudo update-alternatives --set gcc /usr/bin/gcc-14
+
+      - name: Run tester
+        run: ./koch.py test --batch:"$TEST_BATCH" --tryFailing --targets:c all
+        env:
+          TEST_BATCH: ${{ matrix.batch }}_${{ matrix.total_batch }}
+
+      - name: Print all test errors
+        if: failure()
+        run: bin/nim r tools/ci_testresults
```
alaviss, line 95:
```suggestion
        run: bin/nim r tools/ci_testresults

  passed:
    name: All C compatibility tests passed
    needs: [test]
    if: failure() || cancelled()
    runs-on: ubuntu-latest

    steps:
      - run: exit 1
```

You need to add an extra "summary" job so that we can make this a required check. The reason is that GitHub does not have required workflows but required checks, of which is generated per job per workflow.

If we don't have a summary job, then we would need to mark every matrix-spawned job as required, which is suboptimal.

The summary job used here is a bit different from the other ones within the repo. For these, I took advantage of the fact that *skipped* jobs are considered passing to GitHub to avoid allocating a (very short lived) runner when everything passes.

alaviss (changes requested):
Only one change required so that we can make this new CI merge blocking.

Everything else LGTM.

.github/workflows/compatibility.yml
zerbina: Oh, yeah, I forgot about that, thanks.

.github/workflows/compatibility.yml
```diff
@@ -0,0 +1,104 @@
...
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  binaries:
+    name: Build binaries with most recent GCC version
+    runs-on: ubuntu-24.04
+    # Don't run for draft PRs
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
```
alaviss:
```suggestion
    if: ${{ !github.event.pull_request.draft }}
```

This one can be simplified a bit like this. Accessing operators yields `null` if a property is not found and accessing properties of `null` gives `null`, which is falsy.

We need the `${{}}` wrapping because `!text` is yaml tag syntax. Alternatively you could just use double quotes to wrap this.

alaviss (approved):
LGTM. There's only one small nit, but everything else looks perfect.

PR #1402, opened 2024-08-07
fix wrong hooks being picked for tuple types

zerbina:
## Summary

* fix tuples such as `((A, B),)` and `((A,), B)` being assigned the
  the same set of hooks
* fix the aforementioned tuples sharing the same RTTI object

## Details

* add the tuple length to the type representation during type hashing
  (`sighash`), so that `((A,), B)` and `((A, B),)` result in two
  different `SigHash`es
* `liftdestructors` uses `sighashes` for canonicalizing structural
  types, thus previously treating said tuples as the same type
* RTTI creation, which also uses `SigHash`, was affected in the same
  way

### DIFF
--- a/compiler/sem/sighashes.nim
+++ b/compiler/sem/sighashes.nim
@@ -188,6 +188,9 @@ proc hashType(c: var MD5Context, t: PType; flags: set[ConsiderFlag]) =
     c.hashTree(t.n, {})
   of tyTuple:
     c &= char(t.kind)
+    # add the length so that (int, int) and ((int,), int) have different
+    # representations
+    c &= t.len
     if t.n != nil and CoType notin flags:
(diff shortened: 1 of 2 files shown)


zerbina:
Prior to #1382, `sighashes.hashType` was also used by the C code generator for merging types, which meant that:
```nim
var x: (int, (int, int))
var y: (int, (int,), int)
y[2] = 1
```
resulted in a C compiler error, due to `y` using the same C type as `x`, which is:
```c
struct tyTuple__xxx1 {
  NI Field0;
  NI Field1;
};
struct tyTuple__xxx2 {
  NI Field0;
  tyTuple__xxx1 Field1;
};
```

saem (approved):

PR #1403, opened 2024-08-07
sem: fix error propagation for tuple construction

zerbina:
## Summary

* fix errors in anonymous tuple constructions not being propagated
* fix the error for colon expressions in anonymous tuple constructions
  being "invalid expression" instead of "named expression not allowed
  here"

## Details

* properly propagate errors from element expressions
* don't analyze `nkExprColonExpr` as a normal expression
* don't modify input AST
* skip tuple type constructor analysis if any of the element
  expressions is erroneous

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -3148,39 +3148,49 @@ proc semTuplePositionsConstr(c: PContext, n: PNode, flags: TExprFlags): PNode =
                           "expected nkTupleConstr, got: " & $n.kind)
   
   let
-    tupExp = n                  # we don't modify n, but compute the type:
+    tupExp = shallowCopy(n)
     typ = newTypeS(tyTuple, c)  # leave typ.n nil!
-  for i in 0..<tupExp.len:
-    tupExp[i] = semExprWithType(c, tupExp[i], {}) # xxx: claim of not modifying
-                                                  #      n is dubious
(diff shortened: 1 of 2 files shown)


compiler/sem/semexprs.nim
```diff
@@ -3148,39 +3148,49 @@ proc semTuplePositionsConstr(c: PContext, n: PNode, flags: TExprFlags): PNode =
...
+    let isTupleType = tupExp[0].typ.kind == tyTypeDesc
     # check if either everything or nothing is tyTypeDesc
     for i in 1..<tupExp.len:
-      if tupExp[i].kind == nkExprColonExpr:
-        hasError = true
-        # xxx: not sure if this modification is safe
-        tupExp[i] = c.config.newError(tupExp[i],
-                                      PAstDiag(kind: adSemNamedExprNotAllowed))
-      elif isTupleType != (tupExp[i].typ.kind == tyTypeDesc):
+      if isTupleType != (tupExp[i].typ.kind == tyTypeDesc):
```
saem, line 3184: Do we want to allow this in a compile time context?

saem (approved):

compiler/sem/semexprs.nim
zerbina: At some point, yes, but only very little of the compiler is aware of `tyTypeDesc`'s additional meaning of "fist-class type value", so I don't think it's realistic right now.

PR #1412, opened 2024-08-12
packedsets: reimplement using roaring bitmap

alaviss:
## Summary
* what changed and how?
* why are we changing it?

## Details
* anything else

Fixes full_issue_url

---

## Notes for Reviewers

Todo list:

* [ ] Debug the current performance deficit when used by the compiler
* [ ] Write a proper PR description
  * [ ] Benchmark between `HashSet`, old `PackedSet` and Roaring against the compiler + some test dataset
* [ ] Remove `packedsets0`, which is the old `PackedSet`.
* [ ] Write more tests for containers and roaring
<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/lib/std/packedsets.nim
+++ b/lib/std/packedsets.nim
@@ -1,165 +1,37 @@
 #
 #
-#            Nim's Runtime Library
-#        (c) Copyright 2012 Andreas Rumpf
+#                 NimSkull's runtime library
+#              (c) Copyright 2012 Andreas Rumpf
+#     (c) Copyright 2024 Leorize <leorize+oss@disroot.org>
 #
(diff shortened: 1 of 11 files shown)


alaviss:
While Roaring Bitmap provides unrivaled speed in set math, it unfortunately has too much overhead for get/set operations, which is frequently used within the compiler. This stems from the focus of the structure to optimize for analytic purposes (via set math).

The compiler does not make use of these strengths, so Roaring will end up slowing down the compiler by 10-15% for a 2% reduction in memory usage.

This code will return in the form of an external package in the future.

PR #1422, opened 2024-08-18
fix missing destruction when discarding complex expression

### LINKED ISSUE #1421: Missing `=destroy` call for `discard`ed complex expressions
### Example

```nim
proc foo(): ref int {.discardable.} =
  new(int)

proc main() =
  try:
    foo()
  except ValueError:
    return

main()
```

### Actual Output

```
$ nim r --showIR:mir_out:main -d:traceArc test.nim

-- MIR: main
scope:
  def _1: ref int
  scope:
    def _2: ref int = foo() -> [L0]
    _1 := move _2
    goto [L1]
  except (L0) type(ValueError) else [Resume]:
    scope:
      goto [Leave(L0), L2]
  L1:
L2:

-- end

[Allocated] 0x7fe16ebb9040 result: 0x7fe16ebb9050
```

### Expected Output

```
-- MIR: main
scope:
  scope:
    def _1: ref int = foo() -> [L0]
    =destroy(name _1)
    goto [L1]
  except (L0) type(ValueError) else [Resume]:
    scope:
      goto [Leave(L0), L2]
  L1:
L2:

-- end

[Allocated] 0x7f53a87df040 result: 0x7f53a87df050
[ABOUT TO DESTROY] 0x7f53a87df040
[Freed] 0x7f53a87df040
```

### Additional Information

* Explicitly discarding the value will cause `=destroy` to be generated.
* This was found while debugging [insideout](https://github.com/disruptek/insideout) memory leaks.

zerbina:
## Summary

Fix the result of `if`, `case`, `block`, and `try` expression not being
destroyed when `discard`ing the expression.

Fixes https://github.com/nim-works/nimskull/issues/1421.

## Details

* in `mirgen`, a value-like PMIR expression is now requested for the
  discarded expression
* `toValue` expects a properly value-like PMIR expression, which a
  complex expression is not, hence the missing destructor
* the expression is now properly materialized, which means that the a
  destructor call is registered for the temporary

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -2185,15 +2185,14 @@ proc gen(c: var TCtx, n: PNode) =
           c.builder.pop(f)
   of nkDiscardStmt:
     if n[0].kind != nkEmpty:
-      let e = exprToPmir(c, n[0], false, false)
+      var e = exprToPmir(c, n[0], false, false)
       case classify(e)
-      of Rvalue:
-        discard toValue(c, e, e.high, mnkDefCursor)
-      of OwnedRvalue:
+      of Rvalue, OwnedRvalue:
         # extend the lifetime of the value
         # XXX: while not possible at the moment, in the future, the
         #      discard statement could destroy the temporary right away
-        discard toValue(c, e, e.high, mnkDef)
+        wantValue(e) # forces materialization
+        discard toValue(c, e, e.high)
       of Lvalue:
         c.buildStmt mnkVoid:
           genx(c, e, e.high)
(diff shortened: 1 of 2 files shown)


saem:
> * `toValue` expects a properly value-like PMIR expression, which a
>   complex expression, hence the missing destructor

I think the bullet point is incomplete, did you mean: 

"which a complex expression didn't produce,"?

saem (approved):

zerbina:
> I think the bullet point is incomplete, did you mean:
> 
> "which a complex expression didn't produce,"?

Ah, thank you, good catch. I meant that a complex expression is not a value-like expression (in this context) -- it's corrected now.

PR #1423, opened 2024-08-19
strutils: fix buffer overflow in formatBiggestFloat

alaviss:
## Summary
Fixed a potential buffer overflow in `struts.formatBiggestFloat`.

## Details
Previously,  `strutils.formatBiggestFloat`  was implemented using 
`sprintf`  and as such came with a risk of buffer overflow. This was
observed via tests in disruptek/insideout.

* Replaced `sprintf` with `snprintf`.
* Remove the 2 KiB temporary buffer in favor of directly outputting to
the target buffer.
* Added extra error checking code for errors from `snprintf`.

### DIFF
--- a/lib/pure/strutils.nim
+++ b/lib/pure/strutils.nim
@@ -2306,8 +2306,11 @@ func validIdentifier*(s: string): bool {.rtl, extern: "nsuValidIdentifier".} =
 
 # floating point formatting:
 when not defined(js):
-  func c_sprintf(buf, frmt: cstring): cint {.header: "<stdio.h>",
-                                     importc: "sprintf", varargs}
+  func c_snprintf(buf: cstring, size: csize_t, frmt: cstring): cint
+                 {.importc: "snprintf", varargs, sideEffect.}
+  func c_strerror(errnum: cint): cstring
+                 {.importc: "strerror", sideEffect.}
+  var c_errno {.importc: "errno", header: "<errno.h>".}: cint
 
 type
   FloatFormatMode* = enum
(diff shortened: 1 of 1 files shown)


lib/pure/strutils.nim
```diff
@@ -2359,26 +2362,64 @@ func formatBiggestFloat*(f: BiggestFloat, format: FloatFormatMode = ffDefault,
...
+      result = newString(L + 1)
+      L = formatTo(result, result.len.csize_t)
     else:
       frmtstr[1] = floatFormatToChar[format]
       frmtstr[2] = '\0'
-      L = c_sprintf(addr buf, addr frmtstr, f)
-    result = newString(L)
-    for i in 0 ..< L:
+      template formatTo(buf: cstring, size: csize_t): cint =
+        # This has the side-effect of invalidating c_errno, but we also do
```
saem, line 2402: Is this not duplicated, can we not move it to a common scope?

saem (approved):
Don't suppose there is an easy enough reproduction for a test, or do you think it's too implementation specific?

lib/pure/strutils.nim
alaviss: The two calls are different, one has precision and one hasn't.

alaviss:
> Don't suppose there is an easy enough reproduction for a test, or do you think it's too implementation specific?

It's too implementation specific and won't be of any help if we move to a native impl at one point.

lib/pure/strutils.nim
saem: Thanks, was on the phone and I missed that while scrolling up and down.

saem:
/merge

(updated the PR body slightly)

PR #1426, opened 2024-08-21
macros: add `stamp` for unquoting a template body

alaviss:
## Summary

Add  `stamp`  to  `std/macros` , which takes a template body and applies
it immediately. This is the same behaviour as the previous  `quote` 
that was changed after https://github.com/nim-works/nimskull/pull/1393.

## Details

After https://github.com/nim-works/nimskull/pull/1393 we have a hole in
the stdlib for  `quote` -like interpolation but with automatic binding
of symbols to reduce boilerplate. Various libraries within the ecosystem
such as  `criterion`  and  `npeg`  have found the feature useful and
extensively employ it.

Reintroduces old  `quote`  behavior as  `stamp` , without support for
custom operators and explicitly document the template-like nature of 
`stamp` 's body as well as its pitfalls.

This macro was originally written and shared to Matrix by @zerbina.

### DIFF
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -1722,3 +1722,126 @@ proc extractDocCommentsAndRunnables*(n: NimNode): NimNode =
         result.add ni
       else: break
     else: break
+
+macro stamp*(body: untyped): NimNode =
+  ## Accepts a template body and returns the AST that represents it.
+  ##
+  ## Identifiers within `body` are bound to symbols from the caller's scope in
+  ## the same fashion as a template. As a special case, `result` is excluded
+  ## from automatic binding.
+  ##
+  ## The template body is hygienic, as such identifiers declared within might
+  ## turn into `gensym` symbols. This behavior can be overridden using
+  ## `{.gensym.}` or `{.inject.}` pragmas at declaration sites. Consult the
+  ## language manual for details on template hygiene.
+  ##
+  ## Within `body`, placeholders, which are references to values in the caller's
+  ## scope delimited by backticks, are substituted with the referenced values.
+  runnableExamples:
+    import std/strutils
+    import std/times
+
+    macro logQuote(msg: string) =
+      ## Log the given message with timestamp
+      # Using `quote` requires binding many symbols explicitly so that users
+      # don't have to import the providers themselves
+      let
+        # Make sure that `$` is selected from this scope to get `$` for `DateTime`
+        stringify = bindSym"$"
+        # Bind to `now` so that users don't have to import times
+        now = bindSym"now"
+        # Bind to `strutils.%` so users don't have to import strutils
+        format = bindSym"%"
+
+      quote:
+        echo `format`("[$1]\t$2", [stringify(`now`()), `msg`])
+
+    macro log(msg: string) =
+      ## Log the given message with timestamp
+      # Using `stamp`, `echo`, `%`, `$` and `now` are bound automatically to
+      # macro scope and users won't have to import times or strutils manually.
+      stamp:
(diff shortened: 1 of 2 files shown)


lib/core/macros.nim
```diff
@@ -1722,3 +1722,127 @@ proc extractDocCommentsAndRunnables*(n: NimNode): NimNode =
         result.add ni
       else: break
     else: break
+
+macro stamp*(body: untyped): NimNode =
+  ## Accepts a template body and returns the AST that represents it.
```
saem:
```suggestion
  ## Accepts a template body, immediately applies it, and returns the resulting AST.
```

saem (approved):
Minor correction in the doc block to make it explicit what the behaviour of `stamp` is, also I'll update the PR body a bit, just lead with the introduction of the new macro and its behaviour.

lib/core/macros.nim
```diff
@@ -1722,3 +1722,127 @@ proc extractDocCommentsAndRunnables*(n: NimNode): NimNode =
...
+  ## from automatic binding.
+  ##
+  ## The template body is hygienic, as such identifiers declared within might
+  ## turn into `gensym` symbols. This behavior can be overridden using
+  ## `{.gensym.}` or `{.inject.}` pragmas at declaration sites. Consult the
+  ## language manual for details on template hygiene.
+  ##
+  ## Within `body`, placeholders, which are references to `NimNode` values in
+  ## the caller's scope delimited by backticks, are substituted with copies of
+  ## the referenced values.
```
zerbina: This part needs to be adjusted:
* the unquoted expressions can be of any type, not just `NimNode`
* only the `ref` is copied for `NimNode`s, not the node itself (the comment doesn't make this clear at the moment)

lib/core/macros.nim
```diff
@@ -1722,3 +1722,127 @@ proc extractDocCommentsAndRunnables*(n: NimNode): NimNode =
...
+      macro log(msg: string) =
+        stamp:
+          let time = $now() # implicitly gensym-ed
+          echo time, "\t", `msg`
+
+      log("first")
+      log("second") # All OK!
+
+  runnableExamples:
+    # `stamp` automatically bind to symbols within the caller scope, which can
```
zerbina:
```suggestion
    # `stamp` automatically binds to symbols within the caller scope, which can
```

lib/core/macros.nim
```diff
@@ -1722,3 +1722,127 @@ proc extractDocCommentsAndRunnables*(n: NimNode): NimNode =
...
+    case n.kind
+    of nnkAccQuoted:
+      result = ident("_" & $args.len)
+      args.add n[0]
+    else:
+      for i in 0..<n.len:
+        n[i] = extract(n[i])
+      result = n
+
+  let body = extract(body)
```
zerbina:
```suggestion
  proc extract(n: NimNode, args: var seq[NimNode]): NimNode =
    ## Extract backticks-delimited expressions.
    case n.kind
    of nnkAccQuoted:
      result = ident("_" & $args.len)
      args.add n[0]
    else:
      for i in 0..<n.len:
        n[i] = extract(n[i], args)
      result = n

  let body = extract(body, args)
```
Not using a closure reduces macro compile- and run-time a bit.

zerbina (changes requested):
Thank you for the elaborate documentation and examples! I'm not super happy of having this behaviour be in the standard library, but at least it's properly documented now, with all it's quirks.

There's one problem with the documentation that needs fixing, and some few other things that should be changed, but this looks good to me otherwise.

saem:
Updated the PR body so it leads with the change.

zerbina (approved):

PR #1432, opened 2024-08-24
fix: import current dir relative modules

### LINKED ISSUE #1430: multiple files import from relative parent folder won't compile
### Example
there are a `foo.nim` and a `bar.nim` in the parent folder of current module

```nim
import ../[foo, bar]
```

### Actual Output

```
Error: cannot open file: ../[foo,bar]
```

### Expected Output
modules are imported properly

### Possible Solution
importing one by one each line compiles
```nim
import ../foo
import ../bar
```

if modules are inside a folder from relative parent folder it compiles
```nim
import ../folder/[foo, bar]
```

### References
works on nim

saem:
## Summary

Attempts to import modules using relative import with multiple module
syntax like  `import ./[foo, bar]`  no longer result in an error.

Fixes: https://github.com/nim-works/nimskull/issues/1430

## Details

`./some_dir/[foo]`  worked because the AST resulted in an  `infix`  node
for the second directory separator  `/` , but a  `prefix`  node when
using  `./[foo]` ,  `importEval`  now takes this into account. A test
has been included along with the fix.

Along with this change some preexisting import statement tests were
adjusted away from using  `when false`  to flag known issues to the
builtin testament functionality of the  `knownissue`  field.

### DIFF
--- a/compiler/modules/importer.nim
+++ b/compiler/modules/importer.nim
@@ -377,7 +377,29 @@ proc evalImport*(c: PContext, n: PNode): PNode =
   var hasError = false
 
   for it in n.sons:
-    if it.kind == nkInfix and it.len == 3 and it[2].kind == nkBracket:
+    if it.kind == nkPrefix and it.len == 2 and it[1].kind == nkBracket:
+      let
+        sep = it[0]
+        impTmpl =
+          block:
+            let t = newNodeI(nkPrefix, it.info, 2)
+            t[0] = sep
+            t
+          ## `impTmpl` is copied/instanced in a loop below including setting
+          ## the second child
+
+      for x in it[1]:
+        let imp = copyTree(impTmpl)
+
+        # transform `./[a as b] to `./a as b`
+        imp[1] =
+          if x.kind == nkInfix and x[0].ident.s == "as":
+            x[1]
+          else:
+            x
+
+        hasError = impMod(c, imp, x.info, result).kind == nkError or hasError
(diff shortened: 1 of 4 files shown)


compiler/modules/importer.nim
```diff
@@ -377,7 +377,29 @@ proc evalImport*(c: PContext, n: PNode): PNode =
...
+
+      for x in it[1]:
+        let imp = copyTree(impTmpl)
+
+        # transform `./[a as b] to `./a as b`
+        imp[1] =
+          if x.kind == nkInfix and x[0].ident.s == "as":
+            x[1]
+          else:
+            x
```
zerbina:
```suggestion
      for x in it[1].items:
        let imp = newTreeI(nkPrefix, it.info,
          it[0],
          # transform `./[a as b] to `./a as b`
          if x.kind == nkInfix and x[0].ident.s == "as":
            x[1]
          else:
            x
        )
```

I know this code was copied from below, but the suggestion should improve clarity a bit. Creating an incomplete temporary tree just to repeatedly copy it afterwards is confusing (besides being inefficient).

Aside: I don't think the `transform x as y` comment is correct. If I'm reading the code correctly, what happens is that `./[x as a]` is turned into `./x` (the `nkInfix` handling has the same problem).

zerbina (commented):
I left a soft suggestion for improving code clarity, but the PR looks good to me otherwise.

saem:
Yup, that code seems simpler

PR #1433, opened 2024-08-25
sem: fix type inference for `static` parameters

zerbina:
## Summary

* fix compiler crash when passing empty aggregate values to `static`
  parameters
* fix `static T` parameter values not being of type `T`, when the
  argument is not directly of type `T` but requires a conversion
* fix converters not being considered for arguments to `static`
  parameters

## Details

When matching an expression against a formal parameter that:
* is a `static` parameter
* contains a `static` type somewhere (e.g., `int or static[float]`)

the expression was compile-time evaluated first, and on success,
assigned a `tyStatic` type. If the `tyStatic` argument type matched the
formal `tyStatic` type, the argument type was bound to the formal
`tyStatic` as-is.

In case an implicit conversion is necessary, the conversion was applied
only *after* the evaluated `tyStatic` was bound to the parameter type,
leading to:
* the actual value inlined at usages of the `static` parameter having
  the wrong type
* type inference for empty aggregate types not working

### Solution

If the formal type is a `static T`, full argument matching (including
implicit conversion and converter handling) for between the argument
and `T` is performed first, and only in case of success is the
expression compile-time evaluated, and the resulting `tyStatic` type
bound to the parameter type.

Post-match fitting has to take place prior to const evaluation (in
order to correctly type empty aggregates), so a new procedure
`tryEvalStaticArgument` is introduced that handles the static
arguments.

The new `static` handling renders the macro/template `static` special-
casing obsolete; replacing the `static` argument with the evaluation
result is now done in `evalTemplateArgs`.

### Typed AST

For calls to routines with `static` arguments, the typed argument
expression now stays in the AST as-is, instead of being replaced with
the evaluated value.

### DIFF
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -710,12 +710,14 @@ proc tryEvalStaticArgument(c: PContext, n: PNode): PNode =
   # types again:
   n.flags.incl nfSem
 
-  result = tryConstExpr(c, n)
-  if result != nil:
+  let e = tryConstExpr(c, n)
+  if e != nil:
     let typ = newTypeS(tyStatic, c)
-    typ.sons = @[result.typ]
-    typ.n = result
-    result = copyNodeWithKids(n)
+    typ.sons = @[e.typ]
+    typ.n = e
+    result =
+      if e == n: copyNodeWithKids(n)
+      else:      n
     result.typ = typ
 
 when not defined(nimHasSinkInference):
--- a/compiler/sem/sem.nim
+++ b/compiler/sem/sem.nim
@@ -694,6 +694,30 @@ proc semRealConstExpr(c: PContext, n: PNode): PNode =
   if result.kind != nkError:
     result = evalConstExpr(c, result)
 
+proc tryEvalStaticArgument(c: PContext, n: PNode): PNode =
+  ## Tries to evaluate an expression passed to a static parameter, while
+  ## overload resolution is still in progress. Returns nil if not successful.
+  var n = n
+  # `n` comes from sigmatch after a match and thus needs post-match fitting
+  if n.kind in {nkHiddenStdConv, nkHiddenSubConv, nkHiddenCallConv}:
+    # post-match fitting expects AST it can freely modify, which is guaranteed
+    # to be true for `n`, so copy the tree first
+    n = fitNodePostMatch(c, copyTree(n))
+  else:
+    n = copyNodeWithKids(n)
+
+  # prevent re-semming of the expression, which would throw away the
+  # types again:
+  n.flags.incl nfSem
+
+  result = tryConstExpr(c, n)
+  if result != nil:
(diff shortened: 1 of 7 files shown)


zerbina:
The failure of `tgeneric_various.nim` is due to proc symbols passed to `static[proc()]` parameters now being turned into closures (as they should). The evaluated result is an `nkClosure`, however, which can currently not pass through `sem` again.

compiler/sem/sem.nim
```diff
@@ -694,6 +694,30 @@ proc semRealConstExpr(c: PContext, n: PNode): PNode =
...
+  # prevent re-semming of the expression, which would throw away the
+  # types again:
+  n.flags.incl nfSem
+
+  result = tryConstExpr(c, n)
+  if result != nil:
+    let typ = newTypeS(tyStatic, c)
+    typ.sons = @[result.typ]
+    typ.n = result
+    result = copyNodeWithKids(n)
```
saem: since we already made a copy earlier, do we need to make a copy again?

saem (approved):

compiler/sem/sem.nim
zerbina: Yep, we do, as `prepareNode` would run into infinite recursion otherwise. Consider the following situation:
1. `n` is passed to `tryConstExpr`
2. if `n` is already a constant expression and there's nothing to fold or evaluate, `tryConstExpr` returns `n` as-is
3. `result == n` is true in that case
4. a `tyStatic` with `PType.n` set to `result` (i.e., `n`) is created

If `result` would be set to `n` (without a copy), the `result.typ = typ` assignment below would result in a type where `typ.n.typ == typ`.

The shallow copy makes sure that `typ.n.typ != typ`, though I could change `result = copyNodeWithKids(n)` to:
```nim
result =
  if n != result: n
  else:           copyNodeWithKids(n)
```
in order to get around an unnecessary shallow copy, and to also emphasize why the copy is needed.

saem: Thanks for the details explanation, I forgot that `tryConstExpr` can act like an identity function in certain cases. I agree, the suggested code that avoids the copy is clearer in that regard.

zerbina: I've adjusted the code a bit -- it should be clearer now.

zerbina:
For making `static[proc()]` parameters work again, I've added handling for `nkClosure` to `semExpr` -- `nkClosure` AST is now turned back into an `nkHiddenStdConv`. I think it's an okay solution for now; the alternatives would have been to either:
* implement full analysis for `nkClosure`
* OR turn closures without an environment into just an `nkSym` node in `vmcompilerserdes`

saem (approved):
I think the this solution, where `semexprs` handles `nkClosure` is better as it's less action at a distance (compared to `vmserdes`), and full `nkClosure` analysis would be built into `sem`, likely involving `semexprs` anyhow.

PR #1435, opened 2024-08-28
openssl, net: remove openssl pre-1.1.1 and libressl support

alaviss:
## Summary
Remove support for OpenSSL older than 3.0, as it has reached EoL and
should no longer be used in production.

Since version 1.1.1 is still widely used in LTS distributions such as
RHEL 8 or Ubuntu 20.04, support for it is retained, however.

## Details
* LibreSSL support is also removed from the wrapper, as it has not been
tested and likely no longer works with the latest versions of LibreSSL.
* Removed shims for older OpenSSL from the wrapper. This is a breaking
change although the impacts should be light to non-existent.
* Removed unnecessary initialization functions from  `net` , since
OpenSSL now automatically initializes the necessary components.
*  `protTLSv1`  is now deprecated as TLSv1.0 is no longer safe, and
OpenSSL has deprecated the related procedures.
*  `openssl111`  define has been added to handle statically linking with
OpenSSL 1.1.1, since the symbol  `SSL_get_peer_certificate`  depended on
it.
* CI for M1 macOS has been configured to always use Homebrew-supplied
OpenSSL.

### DIFF
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -518,10 +518,6 @@ proc fromSockAddr*(sa: Sockaddr_storage | SockAddr | Sockaddr_in | Sockaddr_in6,
 
 when defineSsl:
   CRYPTO_malloc_init()
-  doAssert SslLibraryInit() == 1
-  SSL_load_error_strings()
-  ERR_load_BIO_strings()
-  OpenSSL_add_all_algorithms()
 
   proc sslHandle*(self: Socket): SslPtr =
     ## Retrieve the ssl pointer of `socket`.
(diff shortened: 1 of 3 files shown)


lib/pure/net.nim
```diff
@@ -625,29 +621,18 @@ when defineSsl:
     var newCTX: SslCtx
     case protVersion
     of protSSLv23:
-      newCTX = SSL_CTX_new(SSLv23_method()) # SSlv2,3 and TLS1 support.
+      newCTX = SSL_CTX_new(TLS_method()) # SSLv3, TLS 1.0 and above
     of protSSLv2:
       raiseSSLError("SSLv2 is no longer secure and has been deprecated, use protSSLv23")
     of protSSLv3:
       raiseSSLError("SSLv3 is no longer secure and has been deprecated, use protSSLv23")
     of protTLSv1:
```
saem, line 629: might be worth marking the enum field definition above, with a comment at least, saying that it's deprecated/unsupported.

saem (approved):
Minor suggestion, otherwise looks good.

N.B.: I updated the PR body slightly to lead with the removal

lib/pure/net.nim
alaviss: I don't think we should bother about this one too much, the API itself is broken and should be replaced in the future.

alaviss:
Unfortunately the SSL hacks must stay to shim 1.1 and 3.0, so I brought those back.

lib/pure/net.nim
```diff
@@ -625,30 +621,25 @@ when defineSsl:
...
-    #
-    # From OpenSSL >= 1.1.0, this setting is set by default and can't be
-    # overriden.
-    if newCTX.SSL_CTX_set_ecdh_auto(1) != 1:
-      raiseSSLError()
+    when false:
+      # TODO: this used to "work" when we only perform the config for
+      # version >= 1.1.1. However, it seems that the previous gating condition
+      # was wrong and this code was never ran, giving the impression that
+      # it worked.
```
alaviss, line 638: This is a fun one, I'll have a follow up PR later.

saem (approved):

PR #1437, opened 2024-08-28
fix(sem): errors in lambdas no longer crash

### LINKED ISSUE #1381: Crash in transf due to malformed proc passed as argument
Example 1:

    var x: seq[proc(i: int)]
    x.add proc(e: auto) = echo i

Example 2:

    proc x(f: proc(i: int)) = discard
    x proc(e: auto) = echo i

The first example produces:

    nim.nim                  nim
    cbackend.nim             generateCode
    cbackend.nim             generateCode
    backends.nim             preprocess
    transf.nim               transformBodyWithCache
    transf.nim               transformBody
    transf.nim               processTransf
    transf.nim               transform nkError 389603 test.nim(2, 23)
    idioms.nim               unreachableImpl
    assertions.nim           raiseAssert
    Error: unhandled exception: transf.nim(1223, 16) unreachable: errors can't reach here [AssertionDefect]

and the second example does the same except with `nkError 389709`

saem:
## Summary

Errors in lambdas requiring inference ( `proc(e: auto) = echo e` ), no
longer crash the compiler. Instead errors within the lambda body are
reported and lambda inference fails as it should.

## Details

`semInferredLambda`  failed to wrap its output when there were errors.
This would reach  `transf`  and result in a compiler crash by hitting an 
`unreachable`  assertion.

Now errors in the body are reported immediately (to provide context for
inference failure wrt matching the body) and AST output from the
inference attempt is correctly wrapped such that inference appropriately
fails.

Fixes https://github.com/nim-works/nimskull/issues/1381

In addition, some minor refactoring was done to avoid shadowing the
input parameter  `n`  and making it look like we were mutating the
input.

### DIFF
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -2213,11 +2213,6 @@ proc semInferredLambda(c: PContext, pt: TIdTable, n: PNode): PNode {.nosinks.} =
   pushProcCon(c, s)
   addResult(c, result, result.typ[0])
   result[bodyPos] = semProcBody(c, result[bodyPos])
-  if result[bodyPos].kind == nkError:
-    # output the error now otherwise we'll just get an inferred lambda failure
-    # without an explanation, ideally this should be explained/added context of
-    # the inferred lambda error itself.
-    c.config.localReport(result[bodyPos])
   s.ast[bodyPos] = hloBody(c, result[bodyPos])
   s.ast[bodyPos] = foldInAst(c.module, s.ast[bodyPos], c.idgen, c.graph)
   trackProc(c, s, s.ast[bodyPos])
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -2228,6 +2228,7 @@ proc semInferredLambda(c: PContext, pt: TIdTable, n: PNode): PNode {.nosinks.} =
   for k in result.items:
     if k.kind == nkError:
       result = c.config.wrapError(result)
+      break
   # alternative variant (not quite working):
   # var prc = arg[0].sym
   # let inferred = c.semGenerateInstance(c, prc, m.bindings, arg.info)
--- a/compiler/sem/semstmts.nim
+++ b/compiler/sem/semstmts.nim
@@ -2182,45 +2182,52 @@ proc semProcAnnotation(c: PContext, prc: PNode): PNode =
 proc semInferredLambda(c: PContext, pt: TIdTable, n: PNode): PNode {.nosinks.} =
   ## used for resolving 'auto' in lambdas based on their callsite
   addInNimDebugUtils(c.config, "semInferredLambda", n, result)
-  var n = n
   let original = n[namePos].sym
   let s = original #copySym(original, false)
   #incl(s.flags, sfFromGeneric)
   #s.owner = original
 
-  n = instantiateTypesInBody(c, pt, n, original)
-  result = n
+  result = instantiateTypesInBody(c, pt, n, original)
   s.ast = result
-  n[namePos].sym = s
-  n[genericParamsPos] = c.graph.emptyNode
+  result[namePos].sym = s
+  result[genericParamsPos] = c.graph.emptyNode
   # for LL we need to avoid wrong aliasing
-  n[paramsPos] = newNodeI(nkFormalParams, n[paramsPos].info, n.typ.n.len)
-  for i, p in n.typ.n.pairs:
-    n[paramsPos][i] =
+  result[paramsPos] = newNodeI(nkFormalParams, result[paramsPos].info,
+                               result.typ.n.len)
+  for i, p in result.typ.n.pairs:
+    result[paramsPos][i] =
       case i
       of 0: # return type
-        newNodeIT(nkType, n.info, n.typ[0])
+        newNodeIT(nkType, n.info, result.typ[0])
       else: # copy instantiated parameters
-        n.typ.n[i]
-  s.typ = n.typ
-  let params = n.typ.n
-  for i in 1..<params.len:
-    if params[i].typ.kind in {tyTypeDesc, tyGenericParam,
-                              tyFromExpr}+tyTypeClasses:
-      localReport(c.config, params[i].info, reportSym(
-        rsemCannotInferTypeOfParameter, params[i].sym))
-    #params[i].sym.owner = s
+        if p.typ.kind in {tyTypeDesc, tyGenericParam,
+                          tyFromExpr}+tyTypeClasses:
+          localReport(c.config, p.info, reportSym(
+            rsemCannotInferTypeOfParameter, p.sym))
+        #params[i].sym.owner = s
+        result.typ.n[i]
+  s.typ = result.typ
   openScope(c)
   pushOwner(c, s)
-  addParams(c, params)
+  addParams(c, result.typ.n)
   pushProcCon(c, s)
-  addResult(c, n, n.typ[0])
-  s.ast[bodyPos] = hloBody(c, semProcBody(c, n[bodyPos]))
+  addResult(c, result, result.typ[0])
+  result[bodyPos] = semProcBody(c, result[bodyPos])
+  if result[bodyPos].kind == nkError:
+    # output the error now otherwise we'll just get an inferred lambda failure
+    # without an explanation, ideally this should be explained/added context of
+    # the inferred lambda error itself.
+    c.config.localReport(result[bodyPos])
+  s.ast[bodyPos] = hloBody(c, result[bodyPos])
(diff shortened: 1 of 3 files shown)


compiler/sem/semstmts.nim
```diff
@@ -2182,45 +2182,52 @@ proc semProcAnnotation(c: PContext, prc: PNode): PNode =
...
   pushProcCon(c, s)
-  addResult(c, n, n.typ[0])
-  s.ast[bodyPos] = hloBody(c, semProcBody(c, n[bodyPos]))
+  addResult(c, result, result.typ[0])
+  result[bodyPos] = semProcBody(c, result[bodyPos])
+  if result[bodyPos].kind == nkError:
+    # output the error now otherwise we'll just get an inferred lambda failure
+    # without an explanation, ideally this should be explained/added context of
+    # the inferred lambda error itself.
+    c.config.localReport(result[bodyPos])
```
zerbina: Have you considered moving the `localReport` to `sigmatch.instantiateRoutineExpr`? The shortcoming/issue is with `sigmatch`, so I think the workaround should be located there.

When keeping the `localReport` here, it should move to after the `foldInAst` call, since that can introduce new errors.

tests/lang_callable/generics/tauto_inference_failure_due_to_body_error.nim
```diff
@@ -0,0 +1,11 @@
+discard """
+  description: "Fail lambda inference due to error without the body."
```
zerbina:
```suggestion
  description: "Fail lambda inference due to error within the body."
```
I believe this is what you meant?

tests/lang_callable/generics/tauto_inference_failure_due_to_body_error.nim
```diff
@@ -0,0 +1,11 @@
...
+  errormsg: "undeclared identifier: 'i'"
+  line: 11
+"""
+
+# This is a regression test, ensure we report errors inside lambdas during
+# inference
+
+var x: seq[proc(i: int)]
+x.add proc(e: auto) = echo i
```
zerbina:
```suggestion
  line: 13
"""

# This is a regression test, ensure we report errors inside lambdas during
# inference

proc test(p: proc(x: int)) =
  discard

test proc(e: auto) = i
```
This makes the regression test a bit more focused (no `echo`, no `seq`, simpler lambda inference).

compiler/sem/semstmts.nim
```diff
@@ -2182,45 +2182,52 @@ proc semProcAnnotation(c: PContext, prc: PNode): PNode =
...
+  s.ast[bodyPos] = hloBody(c, result[bodyPos])
   s.ast[bodyPos] = foldInAst(c.module, s.ast[bodyPos], c.idgen, c.graph)
   trackProc(c, s, s.ast[bodyPos])
   popProcCon(c)
   popOwner(c)
   closeScope(c)
+  # wrap in an error if there were issues along the way
+  for k in result.items:
+    if k.kind == nkError:
+      result = c.config.wrapError(result)
```
zerbina, line 2225:
```suggestion
      result = c.config.wrapError(result)
      break
```
Otherwise the errors would be unnecessarily stacked when more than one child node of `result` is an error.

zerbina (commented):
I left a few suggestions regarding corrections/improvements.

One problem there is with the change as it is, is that errors in the body of generic lambda expressions will now be reported *twice*: once in `semInferredLambda`, and once when the wrapper `nkError` is reported.

My inability to come up with a satisfying solution is why I didn't make a PR myself. What *could* work is only wrapping the `s.ast` in an error (but not `result`). `sigmatch` would then detect that case, `localReport` the error, and fail the overload.

It is of course also possible to just output the errors twice, though there should be an `XXX`/`TODO`/`FIXME` mentioning that then, I think.

---

Two remarks regarding the PR description:
* inferring the `auto` types in the lambda expression is already done by the time `semInferredLambda` is called, so an error happening there is - strictly speaking - not an inference failure
* the tree returned by `instantiateTypesInBody` is detached from the input AST, so there was already no input AST mutation (besides altering the symbol) going on

compiler/sem/semstmts.nim
saem: I hadn't considered placing it in `sigmatch.instantiateRoutineExpr`, but since `semInferredLambda` is only ever called there it makes sense. I'll move it over.

saem:
> I left a few suggestions regarding corrections/improvements.
> 
> One problem there is with the change as it is, is that errors in the body of generic lambda expressions will now be reported _twice_: once in `semInferredLambda`, and once when the wrapper `nkError` is reported.

Doubled error reporting compared to a crash is an improvement in my mind. 

> My inability to come up with a satisfying solution is why I didn't make a PR myself. What _could_ work is only wrapping the `s.ast` in an error (but not `result`). `sigmatch` would then detect that case, `localReport` the error, and fail the overload.

I think not having `result` be an error is the wrong intention for the procedure, if anything `s.ast` shouldn't be wrapped. As far as the signature goes the procedure initially matches enough, but the body has to sem in order for the inference to be considered successful, the body is sort of a contract on the `auto` parameter.

> It is of course also possible to just output the errors twice, though there should be an `XXX`/`TODO`/`FIXME` mentioning that then, I think.
> 
> Two remarks regarding the PR description:
> 
> * inferring the `auto` types in the lambda expression is already done by the time `semInferredLambda` is called, so an error happening there is - strictly speaking - not an inference failure

In my mind I consider the body of the lambda to be a part of the contract on the `auto` parameter.

> * the tree returned by `instantiateTypesInBody` is detached from the input AST, so there was already no input AST mutation (besides altering the symbol) going on

Good point, will update.

saem:
So I've given this some more thought and I think the fundamental issue is that there is no way to differentiate between a failure of the contract on `auto` parameter via the body vs a programmer error.

If I keep following this reasoning then it seems to me that anything with `auto` can't exactly participate in overload resolution, or rather it needs to be "terminal" in overload resolution, because of the ambiguity above. We can't have it silently match something else, just because there is an error in the body as we can't tell if that's a contract failure or a programmer error.

So what that tells me we need to produce an error in order to halt further analysis of overloads.

Does that sound about right?

zerbina:
> Doubled error reporting compared to a crash is an improvement in my mind.

Of course. It wasn't mentioned anywhere, so I figured I'd at least bring it up, and like I said, double-reporting works as a solution for now.

> I think not having `result` be an error is the wrong intention for the procedure

I agree. I was just describing a temporary compromise that could be made.

zerbina:
> If I keep following this reasoning then it seems to me that anything with `auto` can't exactly participate in overload resolution

This is more of a remark than anything else, but as far as `typeRel` and `proc` type inference are concerned, a lambda with `auto` parameters is no different from the symbol of a generic routine. That is:
```nim
proc generic[T](x: T) = discard

proc p(x: proc(x: int)) = discard

p generic
# is equivalent to (for typeRel):
p proc(x: auto) = ...
```

My opinion is that within the scope of overload resolution / parameter type inference, both should behave the same way.

Also, to be clear, this is not something that I think is blocking this PR.

zerbina:
> Does that sound about right?

Sorry for not answering earlier. I thought about this for a bit, but wasn't able to come up with a satisfactory answer.

Building upon what you said, there are currently three different cases for lambda body errors:
1. an error occurred during the generic pre-pass (e.g., a symbol couldn't be bound)
2. analysis of the body fails due to the inferred type of a parameter
3. analysis of the body fails due to a programmer error (e.g., illegal `break`)

Number 1 *could* be detected early on (but currently isn't), while 2 and 3 are - like you said - not really distinguishable (depending on what falls under number 2).

Matching something else does seem fine to me when `semInferredLambda` failed for one overload since, because:
* if there's a non-inference error, the other overloads would fail too
* an overload with `untyped` might exist and could recover from the error

If `semInferredLambda` fails due to an inference-related error (e.g., `when arg is Type: {.error.}`), another non-`untyped` overload might still succeed.

compiler/sem/sigmatch.nim
```diff
@@ -2214,7 +2219,7 @@ proc instantiateRoutineExpr(c: PContext, bindings: TIdTable, n: PNode): PNode =
       else:
         newSymNode(inferred, n.info)
   of nkProcTy, nkIteratorTy:
-    # possible in a concept context. There's nothing to instantiate
+    # possibly in a concept context. There's nothing to instantiate
```
zerbina: What I meant to convey here is that `nkProcTy` and `nkIteratorTy` are possible *only* in a concept context; "possibly" - to me - reads like both nodes can also appear outside of a concept context (which they can't).

(Not a blocker, I just wanted to provide some background on what the intended meaning is/was.)

saem: Ah, got it. Thanks for the explanation.

PR #1438, opened 2024-08-29
net: add support for Windows certificate store

alaviss:
## Summary
When applicable, load the Windows Root Certificate store into OpenSSL
verification store. This allows  `cacert.pem`  requirement to become
optional and provide support for custom certificates within the root
store (ie. in corporate environments).

## Details
* On OpenSSL >= 3.2.0, load the Windows Root Certificate store and don't
require  `cacert.pem`  to be present.
* Removed `cacert.pem` from CI as we use a recent OpenSSL there.


saem (approved):
Nice improvement

PR #1443, opened 2024-09-02
fix(sem): crash with procedure with error used as argument

### LINKED ISSUE #1384: compiler crashes when using erroneous generic routine symbol as call argument
### Example

```nim
proc generic[T](x: T) =
  missing # provoke the generic pre-pass to fail (via a symbol lookup error) 

proc f(p: proc(x: int)) =
  discard

f(generic) # <- instantiation is attempted even though the symbol is erroneous
```

### Actual Output

```
Error: unhandled exception: field 'sons' is not accessible for type 'TNode' using 'kind = nkError' [FieldDefect]
```

### Expected Output

```
Error: undeclared identifier: 'missing'
```

### Possible Solution

Analyzing the expression goes through `semExpr`, but not `semSym` (which would have wrapped the symbol usage in an error), due to the symbol being a `skProc`.

If the `symChoice` call part of the `nkIdent` handling of `semExpr` results in a single symbol, the symbol can be checked (and wrapped in an error) early. For the case where a symbol-choice is returned, the error wrapping can only happen where the choice is ultimately resolved.

### LINKED ISSUE #1442: Crash when trying to use a closure with error inside another closure
### Specification

When a closure has an error, the compiler crash when you try to pass it to a function while inside another closure.

### Example

```nim
proc c*(x: proc()) =
    discard
    
proc a() =
    proc b() =
        foo
    proc x() =
        c(b)
```

### Actual Output

```
nim.nim                  nim
sem.nim                  semStmtAndGenerateGenerics nkProcDef 386682 crash.nim(4, 0)
semexprs.nim             semExpr nkProcDef 386682 crash.nim(4, 0)
semstmts.nim             semRoutineDef nkProcDef 386682 crash.nim(4, 0)
semexprs.nim             semProcBody
semexprs.nim             semExpr nkStmtList 386690 crash.nim(5, 4)
semstmts.nim             semStmtList nkStmtList 386690 crash.nim(5, 4)
semexprs.nim             semExpr nkProcDef 386701 crash.nim(7, 4)
semstmts.nim             semRoutineDef nkProcDef 386701 crash.nim(7, 4)
sempass2.nim             trackProc
sempass2.nim             track
sempass2.nim             trackCall
sempass2.nim             trackOperandForIndirectCall
sempass2.nim             propagateEffects
Error: unhandled exception: field 'sons' is not accessible for type 'TNode' using 'kind = nkError' [FieldDefect]
```

### Expected Output

```
crash.nim(6, 9) Error: undeclared identifier: 'foo'
```

zerbina:
## Summary

Fix a compiler crash when passing the name of a procedure with an error
as an argument to another routine.

## Details

Overloadable and overloaded procedure symbols forewent the `nkError`
wrapping done by `semSym`, thus not preventing instantiation or later
inspection by `sempass2`, crashing the compiler with an `IndexDefect`
on attempting to access the symbol's AST.

Non-overloaded procedure symbols can be passed to `semSym` directly,
which takes care of the wrapping. To handle overloaded symbols,
`sigmatch.paramTypesMatch` passes the picked symbol to `semExpr` first,
which subsequently calls `semSym`, also marking the symbols as used.

The problem with using `semExpr` in `paramTypesMatch` is that the
returned `nkError` currently dismisses the overload, resulting in an
unnecessary "type mismatch" compiler error (with `nimsuggest`, `check`,
etc.).

Fixes https://github.com/nim-works/nimskull/issues/1384.
Fixes https://github.com/nim-works/nimskull/issues/1442.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -3561,6 +3561,7 @@ proc semExpr(c: PContext, n: PNode, flags: TExprFlags = {}): PNode =
       result = symChoice(c, n, s, scClosed)
       if result.kind == nkSym:
         markIndirect(c, result.sym)
+        result = semSym(c, n, result.sym, flags)
     of skEnumField:
       if overloadableEnums in c.features:
         result = enumFieldSymChoice(c, n, s)
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -3561,7 +3561,10 @@ proc semExpr(c: PContext, n: PNode, flags: TExprFlags = {}): PNode =
       result = symChoice(c, n, s, scClosed)
       if result.kind == nkSym:
         markIndirect(c, result.sym)
-        result = semSym(c, n, result.sym, flags)
+        # the symbol was alrady marked as used, don't mark it as such again
+        # (via ``semSym``)
+        if result.sym.ast.isError:
+          result = c.config.newError(result, PAstDiag(kind: adWrappedSymError))
     of skEnumField:
       if overloadableEnums in c.features:
         result = enumFieldSymChoice(c, n, s)
--- a/compiler/sem/sigmatch.nim
+++ b/compiler/sem/sigmatch.nim
@@ -2602,9 +2602,9 @@ proc paramTypesMatch*(
 
     else:
       # only one valid interpretation found, executing argument match
-      markUsed(candidate.c, arg.info, arg[bestArg].sym)
+      let realArg = candidate.c.semExpr(c, arg[bestArg], {})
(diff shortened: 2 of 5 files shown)


saem (approved):

zerbina:
The problem with calling `semSym` in after `symChoice` is that `markUsed` is called on the symbols *twice* (once by `symChoice`, and once by `semSym`), which is what currently results in the `nimsuggest` test failures.

Either the error wrapping needs to be split out of `semSym`, or `symChoice` needs to stop marking a single symbol as used. The former is likely the better solution, at least for the moment.

PR #1445, opened 2024-09-03
fix(sem): fix crash routine pragma redefinition

### LINKED ISSUE #1444: Crash with mismatching pragmas in forward declaration
### Specification

When a proc has a forward declaration with mismatching pragmas, the compiler doesn't properly report the error and crashes instead.

### Example

```nim
proc a()
proc a() {.gcsafe.} =
    discard
```

### Actual Output

```
nim.nim                  nim
sem.nim                  semStmtAndGenerateGenerics nkProcDef 386658 example.nim(3, 0)
semexprs.nim             semExpr nkProcDef 386658 example.nim(3, 0)
semstmts.nim             semRoutineDef nkProcDef 386658 example.nim(3, 0)
reports_sem.nim          reportSymbols
assertions.nim           failedAssertImpl
assertions.nim           raiseAssert
Error: unhandled exception: reports_sem.nim(224, 31) `symbols.len == 1`  [AssertionDefect]
```

### Expected Output

Some error. For reference, nim says "pragmas are only allowed in the header of a proc" but they're allowed to be in the definition as well (as long as they match). Something like "pragmas should be in the proc declaration" would be better.

saem:
## Summary

The compiler no longer crashes when encountering a pragma redefinition
error on a routine. This was due to unnecessary asserts in error message
code

## Details
* dropped the asserts that added no value
* dropped the redundancy in the error message itself, no need to repeat
the symbol for the definition, as they will be the same
* error message generation now properly uses string format ( `%` )
instead of concatenation

Fixes https://github.com/nim-works/nimskull/issues/1444

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1581,8 +1581,8 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       let
         proto = r.symbols[0]
         s = r.symbols[1]
-      result = "pragmas are only allowed in the header of a routine; '$1' " &
-               "header at $2 redefined at $3" %
+      result = ("pragmas are only allowed in the header of a routine; '$1' " &
+                "header at $2 redefined at $3") %
                [proto.name.s, conf $ proto.info, conf $ s.info]
 
     of rsemDisjointFields:
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1578,11 +1578,12 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       result = "'method' needs a parameter that has an object type"
 
     of rsemUnexpectedPragmaInDefinitionOf:
-      let proto = r.symbols[0]
-      let s = r.symbols[1]
-      result = "pragmas are only allowed in the header of a proc; redefinition of $1" %
-        ("'" & proto.name.s & "' from " & conf $ proto.info &
-        " '" & s.name.s & "' from " & conf $ s.info)
+      let
+        proto = r.symbols[0]
+        s = r.symbols[1]
+      result = "pragmas are only allowed in the header of a routine; '$1' " &
+               "header at $2 redefined at $3" %
+               [proto.name.s, conf $ proto.info, conf $ s.info]
 
     of rsemDisjointFields:
       result = ("The fields '$1' and '$2' cannot be initialized together, " &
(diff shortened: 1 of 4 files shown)


tests/lang_callable/tpragma_redefinition_on_routine_def_err.nim
zerbina, line 1: Given that the test is for an error message, I think it would make sense to move it to the `errmsgs` category, but that's only a soft suggestion.

compiler/front/cli_reporter.nim
```diff
@@ -1578,11 +1578,12 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
...
-      let proto = r.symbols[0]
-      let s = r.symbols[1]
-      result = "pragmas are only allowed in the header of a proc; redefinition of $1" %
-        ("'" & proto.name.s & "' from " & conf $ proto.info &
-        " '" & s.name.s & "' from " & conf $ s.info)
+      let
+        proto = r.symbols[0]
+        s = r.symbols[1]
+      result = "pragmas are only allowed in the header of a routine; '$1' " &
+               "header at $2 redefined at $3" %
```
```suggestion
      result = ("pragmas are only allowed in the header of a routine; '$1' " &
                "header at $2 redefined at $3") %
```
That should fix the test failure.

zerbina (approved):
Looks good to me. I left one minor suggestion regarding the test file placement.

(I believe you also want to update the PR title?)

tests/lang_callable/tpragma_redefinition_on_routine_def_err.nim
saem: I debated that, I think error messages should be part of the feature/functionality they're defining in the error case, where `errmsgs` is for stuff we can't otherwise categorize, but maybe that's not as wise an idea as I think it is.

Thoughts?

saem:
> Looks good to me. I left one minor suggestion regarding the test file placement.

Thanks

> (I believe you also want to update the PR title?)

Done

tests/lang_callable/tpragma_redefinition_on_routine_def_err.nim
zerbina: Hm, from the top of my head, I'd consider there *being an error* part of the feature, while a test for the *exact message(s)* feels like something separate from it.

Following that line of though, in this case, completing a forward-declaration with a header using different pragmas being an error would belong under `lang_callable`, while a test for the message itself would belong under `errmsgs`.

Given that there's only a single message format here, there's no point in having two separate tests, so I think the current test location is correct. For ease of later test navigation/lookup, what you could do is add a `error_message` label to the test, but that's just another (very) soft suggestion.

PR #1449, opened 2024-09-05
fix: crash when passing NimNode to static parameter

### LINKED ISSUE #1448: Compiler crash on compile-time NimNode variable as a `static NimNode` param to macros.
### Example

```nim
import std/macros

var myNode {.compileTime.} = newStmtList()

macro test(n: static NimNode) = discard

myNode.test()
```

### Actual Output

```
fatal.nim(50)            sysFatal
Error: unhandled exception: mirgen.nim(2501, 18) unreachable: nkNimNodeLit [AssertionDefect]
```

### Expected Output

The compiler should allow this as valid code.

zerbina:
## Summary

* fix a compiler crash when passing a NimNode to a `static` parameter
* fix a compiler crash when evaluating a `NimNode`-returning constant
  expression

## Details

There were two problems:
* `nkNimNodeLit` wasn't handled in constant data expression by `mirgen`
* `NimNode` values returned directly from a VM invocation weren't
  wrapped in `nkNimNodeLit` trees

For handling `NimNode` values in `vm.regToNode`, the existing
deserialization logic for `NimNode` values in `vmcompilerserdes` is
moved to a separate procedure, so that it can be used by `regToNode`.

The `opcRepr` implementation relied on `regToNode` always returning
an unwrapped `PNode` -- it is adjusted to manually handle `NimNode`
values.

A test covering both issues is added.

Fixes https://github.com/nim-works/nimskull/issues/1448.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -2490,6 +2490,8 @@ proc constDataToMir*(env: var MirEnv, n: PNode): MirTree =
       bu.use toFloatLiteral(env, n)
     of nkStrLiterals:
       bu.use strLiteral(env, n.strVal, typ)
+    of nkNimNodeLit:
+      bu.use astLiteral(env, n[0], n.typ)
     of nkHiddenStdConv, nkHiddenSubConv:
       # doesn't translate to a MIR node itself, but the type overrides
       # that of the sub-expression
--- a/compiler/vm/vmcompilerserdes.nim
+++ b/compiler/vm/vmcompilerserdes.nim
@@ -212,6 +212,16 @@ proc deserializeArray*(
   if hasError:
     result = c.config.wrapError(result)
 
+proc deserializeNimNode*(c: TCtx, n: PNode, formal: PType,
+                         info: TLineInfo): PNode =
+  ## Makes sure `n` (which represent a NimNode value) is valid and turns it
+  ## into a proper NimNode literal.
+  if cyclicTree(n):
+    result = c.config.newError(
+      newNodeIT(nkEmpty, info, formal),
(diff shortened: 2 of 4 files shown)


compiler/mir/mirgen.nim
```diff
@@ -2490,6 +2490,8 @@ proc constDataToMir*(env: var MirEnv, n: PNode): MirTree =
       bu.use toFloatLiteral(env, n)
     of nkStrLiterals:
       bu.use strLiteral(env, n.strVal, typ)
+    of nkNimNodeLit:
```
saem, line 2493: Yay, `nkNimNodeLit` is becoming more pervasive, this means fewer places where AST is ambiguously passed around.

compiler/vm/vmcompilerserdes.nim
```diff
@@ -212,6 +212,16 @@ proc deserializeArray*(
...
+proc deserializeNimNode*(c: TCtx, n: PNode, formal: PType,
+                         info: TLineInfo): PNode =
+  ## Makes sure `n` (which represent a NimNode value) is valid and turns it
+  ## into a proper NimNode literal.
+  if cyclicTree(n):
+    result = c.config.newError(
+      newNodeIT(nkEmpty, info, formal),
+      PAstDiag(kind: adCyclicTree, cyclic: n))
+  else:
+    result = newTreeIT(nkNimNodeLit, info, formal): n
```
saem, line 224: Don't see this as an issue now, and probably ever, but it looks like we're going to treat `nkNimNodeLit` to mean either a literal/static AST value or dynamic AST value. Where the latter one is generated on the fly, at compile time.

saem (approved):

PR #1450, opened 2024-09-06
fix(sem): analysis considers nested noreturn

### LINKED ISSUE #1440: raise within a nested `if` invalidate the outer if-expression
### Example

```nim
let s = [0]

let x =
  if s[0] == 0:
    0
  else:
    if true:
      raise newException(CatchableError, "error")
    else:
      raise newException(CatchableError, "another error")
```

### Actual Output

```
test.nim(5, 5) Error: expression '0' is of type 'int literal(0)' and has to be used (or discarded)
```

### Expected Output

This should not error, as the alternative branch is a `noreturn`.

### Additional Information

* If the `raise` is not nested in an another `if`, no error will occur

saem:
## Summary
No-return analysis ensures that nested final expressions (
`block/case/if/try` ) are accounted for no-return itself.

## Details
Prior to this change if the last expression within a greater expression
was a no-return expression it would not be deemed as such if it was a 
`block/case/if/try` , instead errors would be raised related to discard
checks or that path not resulting in the correct type.

For example, the following would result in an error where  `0`  must be
used or discarded:
```nim
proc foo() =
  let x =
    if false: 10
    else:
      block:
        return
```

Fixes https://github.com/nim-works/nimskull/issues/1440

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -690,6 +690,16 @@ iterator genericParamsInMacroCall*(macroSym: PSym, call: PNode): (PSym, PNode) =
     if posInCall < call.len:
       yield (genericParam, call[posInCall])
 
+proc endsInNoReturn*(n: PNode): bool =
+  ## checks if expression `n` ends in an unstructured exit (raise, return,
+  ## etc), calls of noreturn proc, or
+  var it = n
+  while it.kind in {nkStmtList, nkStmtListExpr} and it.len > 0:
+    it = it.lastSon
+  result = it.kind in nkLastBlockStmts or
+    it.kind in {nkIfStmt, nkIfExpr, nkTryStmt, nkCaseStmt, nkTryStmt, nkBlockExpr} and it.typ.isNil or
+    it.kind in nkCallKinds and it[0].kind == nkSym and sfNoReturn in it[0].sym.flags
+
 type
   NodePosName* = enum
     ## Named node position accessor
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -692,7 +692,7 @@ iterator genericParamsInMacroCall*(macroSym: PSym, call: PNode): (PSym, PNode) =
 
 proc endsInNoReturn*(n: PNode): bool =
   ## checks if expression `n` ends in an unstructured exit (raise, return,
-  ## etc), calls of noreturn proc, or
+  ## etc) or calls of noreturn proc. This is meant to be called on an semmed `n`.
   var it = n
   while it.kind in {nkStmtList, nkStmtListExpr} and it.len > 0:
     it = it.lastSon
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -691,13 +691,15 @@ iterator genericParamsInMacroCall*(macroSym: PSym, call: PNode): (PSym, PNode) =
       yield (genericParam, call[posInCall])
 
 proc endsInNoReturn*(n: PNode): bool =
-  ## checks if expression `n` ends in an unstructured exit (raise, return,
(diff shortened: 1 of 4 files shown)


compiler/ast/ast_query.nim
```diff
@@ -690,6 +690,16 @@ iterator genericParamsInMacroCall*(macroSym: PSym, call: PNode): (PSym, PNode) =
     if posInCall < call.len:
       yield (genericParam, call[posInCall])
 
+proc endsInNoReturn*(n: PNode): bool =
+  ## checks if expression `n` ends in an unstructured exit (raise, return,
+  ## etc), calls of noreturn proc, or
```
```suggestion
  ## etc) or calls of noreturn proc. This is meant to be called on an semmed `n`.
```

compiler/ast/ast_query.nim
```diff
@@ -690,6 +690,16 @@ iterator genericParamsInMacroCall*(macroSym: PSym, call: PNode): (PSym, PNode) =
     if posInCall < call.len:
       yield (genericParam, call[posInCall])
 
+proc endsInNoReturn*(n: PNode): bool =
+  ## checks if expression `n` ends in an unstructured exit (raise, return,
+  ## etc) or calls of noreturn proc. This is meant to be called on an semmed `n`.
```
```suggestion
  ## Checks if expression `n` ends in an unstructured exit (raise, return,
  ## etc.) or a call of a noreturn proc. This is meant to be called on a
  ## semmed `n`.
```
Fixes a few small issues (`an` instead of `a`, plural for call, etc.).

tests/lang_exprs/tnoreturn_nested.nim
```diff
@@ -0,0 +1,53 @@
+discard """
+  description: "
+    Ensure nested noreturn statements are considererd in noreturn analysis
+  "
```
zerbina:
```suggestion
  description: """
    Ensure nested noreturn statements are considererd in noreturn analysis
  """
```

compiler/ast/ast_query.nim
```diff
@@ -690,6 +690,16 @@ iterator genericParamsInMacroCall*(macroSym: PSym, call: PNode): (PSym, PNode) =
...
+  var it = n
+  while it.kind in {nkStmtList, nkStmtListExpr} and it.len > 0:
+    it = it.lastSon
+  result = it.kind in nkLastBlockStmts or
+    it.kind in {nkIfStmt, nkIfExpr, nkTryStmt, nkCaseStmt, nkTryStmt, nkBlockExpr} and it.typ.isNil or
```
```suggestion
    (it.kind in {nkIfStmt, nkTryStmt, nkCaseStmt, nkBlockStmt} and
     it.typ.isEmptyType()) or
```
`nkIfExpr` and `nkBlockExpr` don't need to be handled here, since they're guaranteed to have a non-`void` type.

Some noreturn compound statements can also have a non-nil `typ` (when assigned the `enforceVoidContext` type), so `isEmptyType` (which returns true for both `typ == nil` and `typ.kind == tyVoid`) has to be used.

tests/lang_exprs/tnoreturn_nested.nim
```diff
@@ -0,0 +1,53 @@
...
+
+  doAssert x is int
+
+block nested_in_case:
+  let x =
+    if true:
+      0
+    else:
+      let s = false
+      case
```
saem:
```suggestion
      case s
```

tests/lang_exprs/tnoreturn_nested.nim
```diff
@@ -0,0 +1,53 @@
+discard """
+  description: """
+    Ensure nested noreturn statements are considererd in noreturn analysis
+  """
```
zerbina:
```suggestion
  description: '''
    Ensure nested noreturn statements are considererd in noreturn analysis
  '''
```
My bad, it needs to be triple single-quotes.

PR #1451, opened 2024-09-07
testament: fix missing error reporting

zerbina:
## Summary

Fix tests with invalid specs being silently ignored (they were neither
compiled nor run) when using `all` or `cat megatest`.

## Details

The `isJoinableSpec` procedure didn't prevent tests with an invalid
spec (`reInvalidSpec`) from being joined, which overrode the failure
with `reJoined`. Since the specs then had no target enabled, they were
also not included in the megatest.

A few invalid test specifications went by unnoticed due to this bug --
they're now fixed.

### DIFF
--- a/tests/lang_callable/macros/tmodify_result_assignment.nim
+++ b/tests/lang_callable/macros/tmodify_result_assignment.nim
@@ -3,7 +3,7 @@ discard """
     Partially typed ``nkReturnStmt`` AST is properly re-analyzed after macro
     invocation
   '''
-  errmsg: "got <string> but expected 'int'"
+  errormsg: "got <string> but expected 'int'"
   line: 26
 """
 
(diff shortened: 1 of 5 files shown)


saem (approved):

tests/lang_callable/macros/tmodify_result_assignment.nim
```diff
@@ -3,7 +3,7 @@ discard """
     Partially typed ``nkReturnStmt`` AST is properly re-analyzed after macro
     invocation
   '''
-  errmsg: "got <string> but expected 'int'"
+  errormsg: "got <string> but expected 'int'"
   line: 26
```
saem, line 7:
```suggestion
  line: 25
```

This will appease the test, but I'm not sure if that's right given the `copyLineInfo` on line 22.

zerbina: The error should be on line 26, but the `copyLineInfo` copies to the wrong target node (`result.body` instead of `result.body[0][1]`).

PR #1452, opened 2024-09-07
fix(sem): avoid no-return false positives

saem:
## Summary
no-return analysis now no longer confuses  `void`  branches for true
no-returns.

## Details

No-return analysis now dives into the trees of  `if/try/block/case` 
statements in order to determine if they are in fact no-return
scenarios. This is only done for the statement variety.

The updated analysis besides being more correct, in that it avoids false
positive no-return identification, it still allows for a number of false
negatives. Chiefly that unstructured exits ( `return/break/raise` ) are
only considered in the trailing position of statement lists and not part
way through. This is considered acceptable as there is an overall
improvement in accuracy and it should cover most code.

This includes a fix for an NPE crash in discard-or-use error reporting.

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -695,11 +695,25 @@ proc endsInNoReturn*(n: PNode): bool =
   ## etc.) or a call of a noreturn proc. This is meant to be called on a
   ## semmed `n`.
   var it = n
-  while it.kind in {nkStmtList, nkStmtListExpr} and it.len > 0:
-    it = it.lastSon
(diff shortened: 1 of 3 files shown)


zerbina (approved):
Looks good to me. Using recursion might make the analysis a tad easier to understand/read, but `endsInNoReturn` will be obsoleted by a proper `void` type anyway.

saem:
Thanks for the review.

> Looks good to me. Using recursion might make the analysis a tad easier to understand/read, but `endsInNoReturn` will be obsoleted by a proper `void` type anyway.

I did think about it, but already felt it's a bit pricey (cost to benefit) as is and didn't want to push it any further than I already have.

PR #1454, opened 2024-09-10
fix(sem): noreturn analysis for `void` exprs

### LINKED ISSUE #1453: false positive in nested noreturn algorithm
### Example

```nim
let x =
  if true:
    1
  else:
    if true:
      echo "fun"
    else:
      raise newException(Defect, "gone")
```

### Actual Output

Complies successfully

### Expected Output

Should not compile at all

### Possible Solution

This should be changed to verify that either it has a type (a real expression), or all branches must ends with noreturn.

https://github.com/nim-works/nimskull/blob/677964ce59a40112b7d21000db65c5a6f8dd8ffe/compiler/ast/ast_query.nim#L703-L704

saem:
## Summary

`void`  branches preceeding with no-return terminal branches are checked
with no-return analysis to avoid false positives.

## Details

The analysis is updated to recursively check  `if/elif/of/except` 
non-terminal branches as well as the terminal branch.

Fixes https://github.com/nim-works/nimskull/issues/1453

### DIFF
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -704,12 +704,26 @@ proc endsInNoReturn*(n: PNode): bool =
       result = endsInNoReturn(it[^1])
       if not result:
         break
+    result = result and n[^1].kind == nkElse
   of nkCaseStmt:
     # skip the selector expression
     for i in 1..<n.len:
       result = endsInNoReturn(n[i])
       if not result:
         break
+    # xxx: this duplicates logic in `semstmts.semCase`, this should eventually
+    #      be combined
+    let
+      caseType = n[0].typ.skipTypes(abstractInst - {tyTypeDesc})
+      requiresElse =
+        case caseType.kind
+        of tyCaseExhaustive:
+          false
+        of tyRange:
+          not (caseType[0].skipTypes(abstractInst).kind in tyCaseExhaustive)
+        else:
+          true
+    result = result and (not requiresElse or n[^1].kind == nkElse)
   of nkTryStmt:
     # ignore the 'finally' -- it doesn't contribute to the type
     for i in 0..<(n.len - ord(n[^1].kind == nkFinally)):
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -694,11 +694,16 @@ proc endsInNoReturn*(n: PNode): bool =
   ## Checks if expression `n` ends in an unstructured exit (raise, return,
   ## etc.) or a call of a noreturn proc. This is meant to be called on a
   ## semmed `n`.
+  const
+    stmtListLike = {nkStmtList, nkStmtListExpr, nkElifBranch, nkElse,
+                    nkOfBranch, nkExceptBranch}
+    cntrlFlowStmts = {nkIfStmt, nkCaseStmt, nkBlockStmt, nkTryStmt}
   var it = n
-  while it.kind in {nkStmtList, nkStmtListExpr, nkElifBranch, nkElse, nkOfBranch, nkExceptBranch} and it.len > 0 or
-        it.kind in {nkIfStmt, nkCaseStmt, nkBlockStmt, nkTryStmt} and it.typ.isEmptyType:
+  while it.kind in stmtListLike and it.len > 0 or
+        it.kind in cntrlFlowStmts and it.typ.isEmptyType:
     case it.kind
-    of nkStmtList, nkStmtListExpr, nkBlockStmt, nkOfBranch, nkElifBranch, nkElse, nkExceptBranch:
+    of nkStmtList, nkStmtListExpr, nkBlockStmt, nkOfBranch, nkElifBranch,
+       nkElse, nkExceptBranch:
       it = it.lastSon
     of nkIfStmt, nkCaseStmt:
       # look through all but the last branch, which is covered after the loop
@@ -715,7 +720,8 @@ proc endsInNoReturn*(n: PNode): bool =
     else:
       unreachable()
   result = it.kind in nkLastBlockStmts or
-    it.kind in nkCallKinds and it[0].kind == nkSym and sfNoReturn in it[0].sym.flags
+           it.kind in nkCallKinds and it[0].kind == nkSym and
+              sfNoReturn in it[0].sym.flags
 
 type
   NodePosName* = enum
--- a/compiler/ast/ast_query.nim
+++ b/compiler/ast/ast_query.nim
@@ -711,19 +711,15 @@ proc endsInNoReturn*(n: PNode): bool =
       result = endsInNoReturn(n[i])
       if not result:
         break
-    # xxx: this duplicates logic in `semstmts.semCase`, this should eventually
-    #      be combined
     let
-      caseType = n[0].typ.skipTypes(abstractInst - {tyTypeDesc})
+      caseType = n[0].typ.skipTypes(abstractRange)
       requiresElse =
         case caseType.kind
-        of tyCaseExhaustive:
-          false
-        of tyRange:
-          not (caseType[0].skipTypes(abstractInst).kind in tyCaseExhaustive)
-        else:
+        of tyFloat..tyFloat64, tyString:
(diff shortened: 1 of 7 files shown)


zerbina:
Ah, dang -- I thought about this case when reviewing the previous PR, but somehow convinced myself it isn't a problem.

I do think it makes sense to fully use recursion now -- in my opinion, it's easier to reason about:
```nim
proc endsInNoReturn*(n: PNode): bool =
  ## Checks if expression `n` ends in an unstructured exit (raise, return,
  ## etc.) or a call of a noreturn proc. This is meant to be called on a
  ## semmed `n`.
  case n.kind
  of nkStmtList, nkStmtListExpr:
    result = n.len > 0 and endsInNoReturn(n[^1])
  of nkBlockStmt, nkExceptBranch, nkElifBranch, nkElse, nkOfBranch:
    result = endsInNoReturn(n[^1])
  of nkIfStmt:
    for it in n.items:
      result = endsInNoReturn(it[^1])
      if not result:
        break
  of nkCaseStmt:
    # skip the selector expression
    for i in 1..<n.len:
      result = endsInNoReturn(n[i])
      if not result:
        break
  of nkTryStmt:
    # ignore the 'finally' -- it doesn't contribute to the type
    for i in 0..<(n.len - ord(n[^1].kind == nkFinally)):
      result = endsInNoReturn(n[i])
      if not result:
        break
  of nkLastBlockStmts:
    result = true
  of nkCallKinds:
    result = n[0].kind == nkSym and sfNoReturn in n[0].sym.flags
  else:
    result = false
```
(this code also fixes the `tnoreturn_nested.nim` test failure.)

saem:
> Ah, dang -- I thought about this case when reviewing the previous PR, but somehow convinced myself it isn't a problem.

You're not the only one. 😆 

> I do think it makes sense to fully use recursion now -- in my opinion, it's easier to reason about:
> 
> ```nim
> proc endsInNoReturn*(n: PNode): bool =
>   ## Checks if expression `n` ends in an unstructured exit (raise, return,
>   ## etc.) or a call of a noreturn proc. This is meant to be called on a
>   ## semmed `n`.
>   case n.kind
>   of nkStmtList, nkStmtListExpr:
>     result = n.len > 0 and endsInNoReturn(n[^1])
>   of nkBlockStmt, nkExceptBranch, nkElifBranch, nkElse, nkOfBranch:
>     result = endsInNoReturn(n[^1])
>   of nkIfStmt:
>     for it in n.items:
>       result = endsInNoReturn(it[^1])
>       if not result:
>         break
>   of nkCaseStmt:
>     # skip the selector expression
>     for i in 1..<n.len:
>       result = endsInNoReturn(n[i])
>       if not result:
>         break
>   of nkTryStmt:
>     # ignore the 'finally' -- it doesn't contribute to the type
>     for i in 0..<(n.len - ord(n[^1].kind == nkFinally)):
>       result = endsInNoReturn(n[i])
>       if not result:
>         break
>   of nkLastBlockStmts:
>     result = true
>   of nkCallKinds:
>     result = n[0].kind == nkSym and sfNoReturn in n[0].sym.flags
>   else:
>     result = false
> ```
> 
> (this code also fixes the `tnoreturn_nested.nim` test failure.)

Yeah, I didn't want to go that route, but I don't think anything else will be pleasant to read.

compiler/ast/ast_query.nim
```diff
@@ -694,27 +694,34 @@ proc endsInNoReturn*(n: PNode): bool =
...
+  case n.kind
+  of nkStmtList, nkStmtListExpr:
+    result = n.len > 0 and endsInNoReturn(n[^1])
+  of nkBlockStmt, nkExceptBranch, nkElifBranch, nkElse, nkOfBranch:
+    result = endsInNoReturn(n[^1])
+  of nkIfStmt:
+    for it in n.items:
+      result = endsInNoReturn(it[^1])
+      if not result:
+        break
```
zerbina, line 706:
```suggestion
        break
    # only noreturn if the 'if' is exhaustive
    result = result and n[^1].kind == nkElse
```

compiler/ast/ast_query.nim
```diff
@@ -694,27 +694,34 @@ proc endsInNoReturn*(n: PNode): bool =
...
+  of nkCaseStmt:
+    # skip the selector expression
+    for i in 1..<n.len:
+      result = endsInNoReturn(n[i])
+      if not result:
+        break
```
zerbina, line 713:
```suggestion
        break
    # only noreturn if the 'case' is exhaustive
    result = result and n[^1].kind == nkElse
```

zerbina (commented):
Ugh, the handling for non-exhaustive `if`/`case` statements is still missing.
```nim
if true:
  1
else:
  if true:   # <- `endsInNoReturn` would return true
    return
```

If added two suggestions for the fix. You might also want to add an additional test.

zerbina:
@saem: Sorry for aborting the merge, but I just noticed another issue that's probably better fixed as part of this PR.

saem:
> @saem: Sorry for aborting the merge, but I just noticed another issue that's probably better fixed as part of this PR.

No problem, I'll add some more tests. I haven't been able to think about the testing for this change all that well, so I'll take a day or two with it.

compiler/ast/ast_query.nim
saem: Is this true, because if it's an enum and you list all possible values, then an `else` isn't required, i.e.:

```nim
var x: bool
case x
of true: discard
of false: discard
```

zerbina: You're right, those would lead to false negatives. An updated suggestion:
```suggestion
        break
    if result:
      result = n[0].typ.skipTypes(abstractRange).kind notin {tyString, tyFloat..tyFloat64} or
               n[^1].kind == nkElse
```
As per `semCase`, only `string` and `float` values can be non-exhaustive while omitting the `else` branch.

saem: I'll see, but I suspect I'm going to have to use/introduce a node flag to pass the information along.

saem: > You're right, those would lead to false negatives. An updated suggestion:
> 
> As per `semCase`, only `string` and `float` values can be non-exhaustive while omitting the `else` branch.

D'oh, I saw this after my change, this is simpler as I can rely on the initial exhaustiveness check instead of redoing the whole thing. I'll update my latest change with a simplified version.

saem:
I expect the tests to pass, and we could consider this PR done, although see the caveat below.

---

Thinking about it more an area where this might not work is something nested inside a loop:

```nim
proc foo() =
  let x =
  while true:
    return
```

The chance of writing that code is very small, and thankfully it's only a false-negative, but it looks like I need to expand the traversal to include `while` and `for` loop bodies. If I'm being really pedantic, then I should also check for `while (return; true): ...`, but that applies to all predicate positions.

compiler/ast/ast_query.nim
```diff
@@ -694,27 +694,44 @@ proc endsInNoReturn*(n: PNode): bool =
...
+    let
+      caseType = n[0].typ.skipTypes(abstractRange)
+      requiresElse =
+        case caseType.kind
+        of tyFloat..tyFloat64, tyString:
+          true
+        else:
+          false
```
```suggestion
      requiresElse = caseType.kind in {tyFloat..tyFloat64, tyString}
```
A soft suggestion. I don't think the `case` helps readability.

I'd have also suggested inlining `caseType`, but that would have made the line longer than 80 characters, so I let it be.

saem: I dropped `caseType` and made it one expression, I much prefer the expression form, thanks.

zerbina (approved):
Looks good to me.

> The chance of writing that code is very small, and thankfully it's only a false-negative, but it looks like I need to expand the traversal to include while and for loop bodies

Given that neither `while` nor `for` can be used as expressions at the moment (outside of `typeof` expressions, but those are super special anyway), I don't think this is something to worry about.

It'd become relevant once introducing a `unit` type and making `while` and `for` real expressions, but at that point `endsInNoReturn` would be obsolete anyway.

PR #1455, opened 2024-09-10
varpartitions: add support for if/try/case paths

alaviss:
## Summary
This commit adds basic support for paths in conditional cases. Conditionals are considered valid paths when all branches of execution that does not terminate execution yields the same path.

## Details
* anything else

Fixes full_issue_url

---

## Notes for Reviewers

* Right now this is for my own experimentation, so progress is gonna be slow as I play around with this.

Todo list:

- [ ] Write tests
- [ ] Update experimental docs

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/sem/varpartitions.nim
+++ b/compiler/sem/varpartitions.nim
@@ -362,6 +362,40 @@ proc pathExpr(node: PNode; owner: PSym): PNode =
           break
       else:
         break
+    of nkElifExpr, nkElifBranch, nkElseExpr, nkElse, nkExceptBranch, nkOfBranch:
+      n = n.lastSon
+    of nkIfExpr, nkIfStmt, nkTryStmt, nkCaseStmt:
+      block samePathCheck:
+        let firstExpr =
+          case n.kind
+          of nkTryStmt:
+            n[0] # First element is try body
+          of nkCaseStmt:
+            n[1].lastSon # First branch is the second element
+          else:
+            n[0].lastSon
+        let firstValidPath = pathExpr(firstExpr, owner)
+        # The first expression should be a valid path
+        if firstValidPath == nil:
+          break samePathCheck
+
+        let remainderStart = if n.kind == nkCaseStmt: 2 else: 1
+        for idx in remainderStart ..< n.len:
+          # Only check branches with a type, as no types implies noreturn for
+          # expressions
+          if not n[idx].lastSon.typ.isEmptyType():
+            # If the first path is not a symbol, then we currently can't check
+            # if other paths points to the same resource, so bail.
+            if firstValidPath.kind != nkSym:
+              break samePathCheck
+
+            let path = pathExpr(n[idx], owner)
+            if path.kind != nkSym or path.sym.itemId != firstValidPath.sym.itemId:
+              break samePathCheck
+
+        result = firstValidPath
+
+      break
     else:
       break
   # borrowFromConstExpr(n) is correct here because we need 'node'


compiler/sem/varpartitions.nim
```diff
@@ -362,6 +362,40 @@ proc pathExpr(node: PNode; owner: PSym): PNode =
...
+            n[1].lastSon # First branch is the second element
+          else:
+            n[0].lastSon
+        let firstValidPath = pathExpr(firstExpr, owner)
+        # The first expression should be a valid path
+        if firstValidPath == nil:
+          break samePathCheck
+
+        let remainderStart = if n.kind == nkCaseStmt: 2 else: 1
+        for idx in remainderStart ..< n.len:
```
saem, line 383: Any trailing `finally` should be removed from the iteration since it's not part of the expression, as in it will always be a statement, right?

compiler/sem/varpartitions.nim
```diff
@@ -362,6 +362,40 @@ proc pathExpr(node: PNode; owner: PSym): PNode =
...
+          # Only check branches with a type, as no types implies noreturn for
+          # expressions
```
saem, line 385: They're either noreturn or unit, the difference being that the latter doesn't "abort" the current expression like the former does. You might need to rely on the `endsInNoReturn` analysis.

compiler/sem/varpartitions.nim
```diff
@@ -362,6 +362,40 @@ proc pathExpr(node: PNode; owner: PSym): PNode =
           break
       else:
         break
+    of nkElifExpr, nkElifBranch, nkElseExpr, nkElse, nkExceptBranch, nkOfBranch:
```
saem, line 365: Should this dig into `nkBlock` expression and statement varieties?

saem (commented):
Just some initial thoughts

zerbina:
In my opinion, extending the path expression specification to have `if cond: x else: x` be considered a path expression is unnecessarily (in terms of implementation effort) and also not very useful).

A simpler yet more useful solution would be to allow `block`/`if`/`case`/`try` expressions in a view creation context. Implementation-wise, for `varpartitions`, this would amount to just adding the following procedure to `varpartitions`:
```nim
proc borrowFromComplex(c: var Partitions; dest: PSym, src: PNode) =
  ## Borrows from all relevant sub-expressions of `src`.
  case src.kind
  of nkStmtListExpr, nkBlockExpr, nkExceptBranch:
    borrowFromComplex(c, dest, src[^1])
  of nkIfExpr:
    for it in src.items:
      borrowFromComplex(c, dest, it[^1])
  of nkCaseStmt:
    for (_, it) in branches(src):
      borrowFromComplex(c, dest, it[^1])
  of nkTryStmt:
    for i in 0..<(src.len - ord(src[^1].kind == nkFinally)):
      borrowFromComplex(c, dest, src[i])
  of nkHiddenAddr:
    # currently necessary to skip the outermost hidden-addr
    borrowFromComplex(c, dest, src[0])
  elif endsInNoReturn(src):
    discard "nothing to do"
  else:
    borrowFrom(c, dest, src)
```

and then changing https://github.com/nim-works/nimskull/blob/677964ce59a40112b7d21000db65c5a6f8dd8ffe/compiler/sem/varpartitions.nim#L594 to

```nim
borrowFromComplex(c, dest.sym, src)
```

`borrowFrom` is additive, so to speak, so calling it multiple times means that multiple locations are borrowed from. This nets you support for the `let x: lent T = if cond: a else: b` case.

You'll still get a run-time or compiler crash when doing so, since `if`, `block`, etc. are npt expected to appear as an `nkHiddenAddr` operand. I can quickly make a PR with the necessary preparations, if you want.

PR #1460, opened 2024-09-11
fix: incorrect code generation for parameter borrowing

### LINKED ISSUE #1457: borrows of parameters below `byref` threshold ends up leasing the copied parameter
### Example

```nim
{.experimental: "views".}
type
  Mint = object
    data: int

  Option = object
    value: lent Mint

proc `=copy`(dst: var Mint, src: Mint) {.error.}

proc some(x: Mint): Option =
  Option(value: x)

proc main() =
  var x = Mint(data: 10)
  let optlent = some(x)

main()
```

### Actual Output

C Code for `some(x)`:

```c
N_LIB_PRIVATE N_NIMCALL(_L6Option_2_M4test, some__test_8)(_L4Mint_1_M4test x) {
	_L6Option_2_M4test result;
	_L4Mint_1_M4test* _2;
	nimfr_("some", "/tmp/test.nim");
	nimln_(12, "/tmp/test.nim");
	_2 = (&x);
	nimZeroMem((void*)(&result), sizeof(_L6Option_2_M4test));
	result.value = _2;
	popFrame();
	return result;
}
```

### Expected Output

The `x` parameter should be a pointer, not a stack copy.

### Additional information

- If `Mint` is tagged with `{.byref.}`, the generated code is correct.

zerbina:
## Summary

Fix parameters below the pass-by-reference size threshold not being
passed by reference when the procedure returns a non-direct view,
resulting in access violations at run-time when trying to access the
returned view.

## Details

* consider immutable non-direct views (e.g., `object` types with `lent`
  fields) when deciding whether pass-by-reference is used for the first
  parameter of a procedure
* move the `classifyViewType` procedures from `typeallowed` to `types`,
  as they're wholly unrelated to the "type allowed" checks

Fixes https://github.com/nim-works/nimskull/issues/1457.


compiler/ast/types.nim
```diff
@@ -1558,6 +1561,74 @@ proc classifyBackendView*(t: PType): BackendViewKind =
...
+    # it doesn't matter what these types contain, 'ptr openArray' is not a
+    # view type!
+    result = noView
+
+proc classifyViewType*(t: PType): ViewTypeKind =
+  var marker = initIntSet()
+  result = classifyViewTypeAux(marker, t)
+
+proc directViewType*(t: PType): ViewTypeKind =
+  # does classify 't' without looking recursively into 't'.
```
saem, line 1621:
```suggestion
  ## classifies 't' without looking recursively into 't'.
```

A tad clearer and shorter, also made it a doc comment as it's relevant to what to expect from the procedure.

saem (approved):
One minor comment related suggestion, but not a blocker.

compiler/ast/types.nim
zerbina: I agree, but I figured to not modify the moved-over code, so as to allow git tooling to better detect the moved code.

PR #1465, opened 2024-10-18
sem: replace `addr` calls with `nkAddr`

zerbina:
## Summary

In typed AST, address-of operations are now *always* represented by
`nkAddr` trees. This simplifies some compiler logic, makes processing
for typed macros easier, and fixes an effect tracking bug with `addr`.

## Details

This is effectively a revert of
https://github.com/nim-lang/nim/pull/10814.
Not turning calls to `mAddr` into `nkAddr` was done to prevent the
unsafe address semantics from being lost, but those no longer exist.

Lowering the call into an `nkAddr` tree has the benefit that it
simplifies AST analysis and processing, as address-of operation can now
always be detected by `PNode.kind == nkAddr`. Old code for detecting
`mAddr` magic calls is removed.

The effect tracking in `sempass2` had no special case for `mAddr`
calls, thus treating them as normal calls, which led to `addr(x)` being
treated as an indirect invocation of `x`, when `x` is of procedural
type. With the `mAddr` call now being lowered earlier, this is no
longer the case.

### DIFF
--- a/compiler/ast/astalgo.nim
+++ b/compiler/ast/astalgo.nim
@@ -599,13 +599,8 @@ proc iiTablePut(t: var TIITable, key, val: int) =
     iiTableRawInsert(t.data, key, val)
     inc(t.counter)
 
-proc isAddrNode*(n: PNode): bool =
-  case n.kind
-    of nkAddr, nkHiddenAddr: true
-    of nkCallKinds:
-      if n[0].kind == nkSym and n[0].sym.magic == mAddr: true
-      else: false
-    else: false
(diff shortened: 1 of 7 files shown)


saem (approved):

zerbina:
/merge

The `tslow_tables.nim` failure is definitely spurious (the timeout probably needs to be increased at some point), and the `tinternal_thread_cleanup.nim` failure too looks like it is spurious (though I haven't seen the test fail in the past).

I'm giving merging another try.

PR #1472, opened 2024-12-05
fix compiler crash with `.global` initializers

### LINKED ISSUE #1470: internal error: 'index out of bounds' in Nimskull with some vtable implementation that uses closures
This is a _mostly_ minimal example. I have no idea what's going on. It works on mainline Nim.

```nim
type Impl = object
proc call(this: var Impl) = discard

type Vtable = object
  call: proc (this: pointer) #{.nimcall.} # <------ works if declared as nimcall

static: debugEcho typeof(Vtable.call)

type Object = object
  obj: pointer
  vtable: ptr Vtable
proc call(this: Object) = this.vtable.call(this.obj)

proc getVtable[T]: ptr Vtable =
  var vtable {.global.} = Vtable(call: proc (this: pointer) = call(cast[var T](this)))
  addr vtable

let vtable = getVtable[Impl]()
```

## Expected Output

Nothing, or a clear error message saying you shouldn't do this thing or it makes the compiler hurt.

## Actual Output

```
nim.nim(154)             nim
main.nim(539)            mainCommand
main.nim(217)            commandCompileToC
cbackend.nim(406)        generateCode
backends.nim(834)        generateCode
backends.nim(445)        produceFragmentsForGlobals
mirgen.nim(2351)         generateAssignment
mirgen.nim(1465)         genLocInit
mirgen.nim(1421)         genLocDef
mirgen.nim(1406)         genAsgn
mirgen.nim(1336)         genAsgnSource
mirgen.nim(2084)         genx
mirgen.nim(1252)         genObjConstr
mirgen.nim(697)          emitOperandTree
mirgen.nim(692)          genArgExpression
mirgen.nim(2154)         genx
mirgen.nim(480)          register
mirgen_blocks.nim(211)   register
fatal.nim(50)            sysFatal
Error: unhandled exception: index out of bounds, the container is empty [IndexDefect]
```

## Additional Info

afcd3c0aba         <----- bug introduced here

zerbina:
## Summary

Fix the compiler crashing for initializers of `.global` variables where
the expression contains locals or temporaries requiring destruction.

Fixes https://github.com/nim-works/nimskull/issues/1470.

## Details

MIR generation for `.global` init fragments is done via
`mirgen.generateAssignment`, which didn't push a scope prior to
translation. This caused any call to `mirgen_blocks.register` to crash
the compiler, as the scope/block list was empty.

A scope is now pushed prior to translating the expression, fixing the
crash. In addition to fixing the crash, this slightly changes
behaviour: unscoped temporaries/locals created as part of a `.global`
initializer are now destroyed immediately after the initialization,
instead of at the end of the module's pre-init procedure. Destruction
order w.r.t. `.global` initializers is unspecified, so this change
shouldn't be a problem.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -2388,16 +2388,18 @@ proc generateAssignment*(graph: ModuleGraph, env: var MirEnv,
   ## `builder`'s currently selected buffer.
   assert n.kind == nkIdentDefs and n.len == 3
   var c = initCtx(graph, config, nil, move env)
-  # treat the code as top-level code so that no 'def' is generated for
-  # assignments to globals
-  c.scopeDepth = 1
 
   template swapState() =
     swap(c.sp.map, source)
     swap(c.builder, builder)
 
   swapState()
-  genLocInit(c, n[0], n[2])
+  # treat the code as top-level code so that no 'def' is generated for
+  # assignments to globals
+  c.scope(true):
+    # a scope is required, otherwise locals/temporaries cannot be registered
+    # for destruction
+    genLocInit(c, n[0], n[2])
   swapState()
   env = move c.env # move back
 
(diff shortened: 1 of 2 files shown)


saem (approved):

alaviss:
@zerbina can you rebase this?

PR #1482, opened 2024-12-25
cgirgen: remove the `var` expression fixup

zerbina:
## Summary

Remove a fixup in `cgirgen` related to `var` parameters in `cgirgen`
and correct the AST construction necessitating it. There's no change in
language semantics.

## Details

In typed `PNode` AST, `var` parameters *always* have to be dereferenced
first, including when being passed to other `var` parameters (resulting
in `(HiddenAddr (HiddenDeref x))` trees). Multiple transformations /
places where AST is constructed in the compiler didn't produce AST
adhering to this.

The result was incorrect MIR code being generated (because the location
*storing* the `var` parameter was being passed to another parameter,
not the location the `var` parameter *references*), though this didn't
cause any trouble so far. `cgirgen` detected this situation and
adjusted its output accordingly, producing well-formed CGIR AST.

The sources of the problematic `PNode` AST were:
* parameter hoisting in `semexprs`
* hook lifting in `liftdestructors`
* method dispatcher generation in `cgmeth`

They're changed to produce AST adhering to the aforementioned
requirements, and the fixup in `cgirgen` is removed.

### DIFF
--- a/compiler/sem/liftdestructors.nim
+++ b/compiler/sem/liftdestructors.nim
@@ -106,7 +106,7 @@ proc defaultOp(c: var TLiftCtx; t: PType; body, x, y: PNode) =
     body.add newAsgnStmt(x, call)
 
 proc genAddr(c: var TLiftCtx; x: PNode): PNode =
-  newTreeIT(nkHiddenAddr, x.info, makeVarType(x.typ.owner, x.typ, c.idgen)): x
+  newTreeIT(nkHiddenAddr, x.info, makeVarType(x.typ.owner, x.typ, c.idgen), x)
 
 proc genWhileLoop(c: var TLiftCtx; i, dest: PNode): PNode =
   let cmp = genBuiltin(c, mLtI, "<", i)
--- a/compiler/sem/liftdestructors.nim
+++ b/compiler/sem/liftdestructors.nim
@@ -264,12 +264,7 @@ proc getCycleParam(c: TLiftCtx): PNode =
 proc newHookCall(c: var TLiftCtx; op: PSym; x, y: PNode): PNode =
   #if sfError in op.flags:
   #  localReport(c.config, x.info, "usage of '$1' is a user-defined error" % op.name.s)
-  result = newNodeI(nkCall, x.info)
-  result.add newSymNode(op)
-  if op.typ.sons[1].kind == tyVar:
-    result.add genAddr(c, x)
-  else:
-    result.add x
+  result = newTreeI(nkCall, x.info, newSymNode(op), genAddr(c, x))
   if y != nil:
     result.add y
   if op.typ.len == 4:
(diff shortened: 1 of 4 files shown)


compiler/sem/semexprs.nim
```diff
@@ -3491,10 +3491,22 @@ proc hoistParamsUsedInDefault(c: PContext, call, letSection, defExpr: var PNode)
...
+        call[paramPos] =
+          newTreeIT(nkHiddenAddr, letSection.info, hoistedVarSym.typ,
+            newDeref(newSymNode(hoistedVarSym)))
+      else:
+        # Refer the original arg to its hoisted sym
+        call[paramPos] = newSymNode(hoistedVarSym)
 
-    # arg we refer to is a sym, wether introduced by hoisting or not doesn't matter, we simply reuse it
-    defExpr = call[paramPos]
+    # arg is either a sym, wether introduced by hoisting or not doesn't
```
```suggestion
    # arg is either a sym, whether introduced by hoisting or not doesn't
```

minor typo

compiler/sem/semexprs.nim
```diff
@@ -3491,10 +3491,22 @@ proc hoistParamsUsedInDefault(c: PContext, call, letSection, defExpr: var PNode)
         newNodeI(nkEmpty, letSection.info),
         call[paramPos])
 
-      call[paramPos] = newSymNode(hoistedVarSym) # Refer the original arg to its hoisted sym
+      if hoistedVarSym.typ.kind == tyVar:
```
saem, line 3494: we don't need to consider any skipTypes/unwrapping because this is after generic resolution etc, yes?

compiler/sem/liftdestructors.nim
```diff
@@ -106,11 +106,7 @@ proc defaultOp(c: var TLiftCtx; t: PType; body, x, y: PNode) =
     body.add newAsgnStmt(x, call)
 
 proc genAddr(c: var TLiftCtx; x: PNode): PNode =
-  if x.kind == nkHiddenDeref:
-    checkSonsLen(x, 1, c.g.config)
-    result = x[0]
-  else:
-    result = newTreeIT(nkHiddenAddr, x.info, makeVarType(x.typ.owner, x.typ, c.idgen)): x
+  newTreeIT(nkHiddenAddr, x.info, makeVarType(x.typ.owner, x.typ, c.idgen)): x
```
saem:
```suggestion
  newTreeIT(nkHiddenAddr, x.info, makeVarType(x.typ.owner, x.typ, c.idgen), x)
```

nbd, just read weird with the `:`

saem (approved):
Some minor suggestions, nothing blocking, and one question.

compiler/sem/semexprs.nim
zerbina: In part, yes. There's also the assumption throughout parts of the compiler that `tyVar` cannot appear be nested in a `tyAlias` or the like.

PR #1493, opened 2025-02-22
allow multiple parameters using an underscore as name

zerbina:
## Summary

Using an underscore as a parameter's name now makes the symbol
anonymous, allowing multiple parameters to use a single underscore.
This affects all routine definitions as well procedure type.

## Details

The motivation is to not require one to give a name to unused
parameters, which is often the case for procedure type parameters and
sometimes for lambda expression parameters. It also makes the meaning
of a single underscore used in a definition context more consistent. 

In `semProcTypeNode` (which is called for all parameter lists),
`isDiscardUnderscore` is called on the parameter symbol, marking it as
a gensym in case it's an underscore, thus preventing it from later
being added to the symbol table. The manual name check ignores
underscore parameters.

So that underscore parameters cannot be referenced in named parameter
passing, `getNamedParamFromList` ignores underscore parameters.

---

## Notes for Reviewers
* longer term, I think underscores should be handled the same in all
  definition positions (i.e., creating an anonymous entity)

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/tests/lang/s02_core/s02_procedures/t99_underscore_parameter_cannot_reference.nim
+++ b/tests/lang/s02_core/s02_procedures/t99_underscore_parameter_cannot_reference.nim
@@ -1,0 +1,9 @@
+discard """
+  description: '''
+    An underscore parameter cannot be reference in named parameter passing.
+  '''
+  action: reject
+"""
+
+proc test(_: int) = discard
+test(_ = 1)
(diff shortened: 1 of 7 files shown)


zerbina:
I'm not exactly happy with the implementation yet. Forward declarations and re-definition checks also don't consider anonymous parameters at the moment.

An alternative implementation could use `sfAnon` to mark the parameters as anonymous, which might be better.

tests/lang/s02_core/s02_procedures/t99_underscore_parameter_cannot_reference.nim
```diff
@@ -0,0 +1,9 @@
+discard """
+  description: '''
+    An underscore parameter cannot be reference in named parameter passing.
```
saem, line 3:
```suggestion
    An underscore parameter cannot be referenced in named parameter passing.
```

PR #1494, opened 2025-02-25
lexopt: lexer for command line

alaviss:
## Summary
Introduce an experimental command line lexer  `lexopt` , meant to
replace  `parseopt`  in all use cases.

## Details

`lexopt`  is an user-driven lexer, where the next token interpretation
is decided by the user at runtime. This was originally designed to
supplement a command line parser implementation, but I've found that
this design is easier to work with for almost all use cases compared to 
`parseopt` .

#### Support for  `--input in.txt`  and  `-ooutput.txt`  syntax without
explicit declaration

Unlike  `parseopt` , where  `shortNoVal`  and  `longNoVal`  must be
declared ahead of time to support such syntax,  `lexopt`  can handle
these cases dynamically through  `value()` . This removes the burden of
maintaining multiple sources of truth, and allow command line parsers to
dictate the parsing strategy dynamically.

Example:

```nim
import std/parseopt
import std/options
import experimental/lexopt

const TestCmd = @["-j4", "-b", "not_an_arg"]

type Opts = object
  j: string
  arg: string
  b: bool

proc popt(): Opts =
  var parser = initOptParser(
    TestCmd,
    shortNoVal = {'b'} # -b has to be declared as "not having a value"
  )
  while (parser.next(); parser.kind != parseopt.cmdEnd):
    if (parseopt.cmdShortOption, "j") == (parser.kind, parser.key):
      result.j = parser.val
    elif (parseopt.cmdShortOption, "b") == (parser.kind, parser.key):
      result.b = true
    elif parser.kind == cmdArgument:
      result.arg = parser.key

proc lopt(): Opts =
  var lex = initCmdLexer(TestCmd)
  while (let (kind, option) = lex.next(); kind != lexopt.cmdEnd):
    if (lexopt.cmdShort, "j") == (kind, option):
      result.j = lex.value().get() # .value() returns Option[string]
    elif (lexopt.cmdShort, "b") == (kind, option):
      result.b = true
    elif kind == lexopt.cmdValue:
      result.arg = option

doAssert popt() == lopt()
```

#### Option values can be tested for "empty" vs "unspecified"

`lexopt` 's  `value()`  function returns  `Option[string]` , allowing
programs to distinguish between having a value and not having one. 
`parseopt`  has no equivalence.

Example:

```nim
import std/parseopt
import std/options
import experimental/lexopt

const TestCmdNoVal = @["--separator"]
const TestCmdVal = @["--separator="]

type Opts = object
  separator: Option[string]

proc popt(args: openArray[string]): Opts =
  var parser = initOptParser(args)
  while (parser.next(); parser.kind != parseopt.cmdEnd):
    if (parseopt.cmdLongOption, "separator") == (parser.kind, parser.key):
      result.separator = some parser.val # not possible to yield None

proc lopt(args: openArray[string]): Opts =
  var lex = initCmdLexer(args)
  while (let (kind, option) = lex.next(); kind != lexopt.cmdEnd):
    if (lexopt.cmdLong, "separator") == (kind, option):
      result.separator = lex.value()

doAssert popt(TestCmdVal) == lopt(TestCmdVal)

doAssert lopt(TestCmdNoVal).separator == none(string)
# doAssert popt(TestCmdNoVal).separator == none(string) # This will fail
```

#### Forced handling of unexpected option values

Unlike  `parseopt` , where all flags are "assumed" to take a value and
require explicit code to check for unexpected values,  `lexopt` 
requires values to be handled should it exists.  `UnexpectedValueError` 
will be raised for any option with values that were not handled.

This mainly benefits handwritten command line parsers, where the
alternative is to add value existence checking code to all options not
expecting one.

Example:

```nim
import experimental/lexopt

const TestCmd = @["--dont-take-val="]

proc main() =
  var lex = initCmdLexer(TestCmd)
  while (let (kind, _) = lex.next(); kind != lexopt.cmdEnd):
    discard

doAssertRaises(UnexpectedValueError):
  main()
```

#### Miscellaneous improvements

- Support for zero-copy ownership takeover:  `initCmdLexer()`  can
consume a  `seq[string]`  directly.
-  `-`  and  `--`  are treated as  `cmdValue` s directly and not as
special cases (ie.  `(parseopt.cmdLongOption, "")`  or 
`(parseopt.cmdShortOption, "")` ).
-  `remaining()`  consuming iterator allows for iterating through
not-yet-parsed arguments without allocating any  `seq` s. The iterator
will prioritize moving values where applicable.

### DIFF
--- a/lib/experimental/lexopt.nim
+++ b/lib/experimental/lexopt.nim
@@ -6,9 +6,7 @@
 # See the file "copying.txt", included in this distribution, for
 # details about copyright.
 
-import std/[strutils, options]
-
-## This module provides an user-driven command line lexer.
+## This module provides a user-driven command line lexer.
 ##
 ## Unlike `std/parseopt`, this lexer parses command line tokens as requested
 ## by the caller.
@@ -41,42 +39,44 @@ runnableExamples:
   doAssert opts.color == true
   doAssert opts.output == "outfile.txt"
 
-const ValueDelims* = {':', '='}
-  ## Characters that delimit between the option and its value
+import std/[strutils, options]
+
+const ValueSeparators* = {':', '='}
+  ## Characters that separate an option's name from its value
 
 type
   CmdlineKind* = enum
     cmdEnd ## End of command line
-    cmdLong ## A long option (ie. `--opt`)
-    cmdShort ## A short option (ie. `-o`)
+    cmdLong ## A long option (i.e. `--opt`)
+    cmdShort ## A short option (i.e. `-o`)
     cmdValue ## A value that is not an option
 
   CmdLexer* = object
     cmdline: seq[string] ## The command line to parse
     index: int ## The index of `cmdline` to be processed
-    valueIdx: int ## \
-      ## For long options, the index of `ValueDelims`.
+    valueIdx: int
+      ## For long options, the index of `ValueSeparators`.
       ##
       ## For short options, the index of the next flag to be processed.
       ##
       ## Otherwise, it is unused.
 
   LexError* = object of CatchableError
-    ## Errors occurred while lexing
+    ## Errors occurring during lexing.
 
   UnexpectedValueError* = object of LexError
-    ## A value was provided but was not consumed
+    ## A value was provided but not consumed.
     opt*: string ## The option that a value was provided to
     value*: string ## The provided value
 
 proc initCmdLexer*(args: sink seq[string]): CmdLexer =
   ## Creates a new `CmdLexer`. `args` should be the command line parameters as
-  ## returned by `os.commandLineArgs()`.
+  ## returned by `commandLineParams <os.html#commandLineParams>`_.
   CmdLexer(cmdline: args)
 
 proc initCmdLexer*(args: openArray[string]): CmdLexer =
   ## Creates a new `CmdLexer`. `args` should be the command line parameters as
-  ## returned by `os.commandLineArgs()`.
+  ## returned by `commandLineParams <os.html#commandLineParams>`_.
   CmdLexer(cmdline: @args)
 
 template current(l: CmdLexer): string =
@@ -90,7 +90,7 @@ proc isDash(s: string): bool = s == "-"
 proc isDashDash(s: string): bool = s == "--"
 
 proc prefix*(kind: CmdlineKind): string =
-  ## Return the prefix string for a given command line parameter kind
+  ## Returns the prefix string for a given command line parameter kind.
   case kind
   of cmdLong:
     "--"
@@ -99,7 +99,7 @@ proc prefix*(kind: CmdlineKind): string =
   else:
     ""
 
-func newUnexpectedValueError(opt, value: string): ref UnexpectedValueError =
+func newUnexpectedValueError(opt, value: sink string): ref UnexpectedValueError =
   (ref UnexpectedValueError)(
     msg: "unexpected value '" & value & "' for '" & opt & "'",
     opt: opt,
@@ -118,19 +118,22 @@ proc next*(l: var CmdLexer): (CmdlineKind, string) =
   ##    If this value is not consumed before the next call to `next()`, an error
   ##    will be raised. See below for more details.
   ## 3. `--foo` is treated as a long option with key `foo`.
-  ## 4. `-abco` is treated as into `a`, `b`, `c`, `o` short options, returned
-  ##    one per `next()` call.
-  ## 5. `-abco:foo` and `-abco=foo` are treated as `a`, `b`, `c`, `o` short
-  ##    options, with `foo` being the value of option `o`.
+  ## 4. `-a:foo` and `-a=foo` are treated as short options with key `a` and
+  ##    value `foo`.
+  ## 5. `-a` is treated as a short option with key `a`.
+  ## 6. `-abco` is equivalent to `-a`, `-b`, `-c`, `-o` appearing in sequence.
+  ##    Four calls to `next()` are required to consume `-abco`.
+  ## 7. `-abco:foo` and `-abco=foo` are equivalent to `-a`, `-b`, `-c`,
+  ##    `-o:foo` appearing in sequence.
   ##
   ##    If this value is not consumed before the next call to `next()`, an error
   ##    will be raised. See below for more details
-  ## 6. Everything else is considered values.
+  ## 8. Everything else is considered values.
   ##
-  ## If the previous option have an unconsumed value (eg. `value()` was not
+  ## If the previous option have an unconsumed value (e.g. `value()` was not
   ## called for `--opt:foo`), `UnexpectedValueError` will be raised.
   type
-    LexState {.pure.} = enum
+    LexState = enum
       ## Current lexer state
       Value ## A value that is not an option
       ShortOpt ## A short option
@@ -150,7 +153,7 @@ proc next*(l: var CmdLexer): (CmdlineKind, string) =
       else:
         LexState.Value
     elif l.current.isShortOpt:
-      if l.valueIdx > 0 and l.current[l.valueIdx] in ValueDelims:
+      if l.valueIdx > 0 and l.current[l.valueIdx] in ValueSeparators:
         LexState.ShortValue
       elif not l.current.isDash:
         LexState.ShortOpt
@@ -161,10 +164,13 @@ proc next*(l: var CmdLexer): (CmdlineKind, string) =
 
   case currentState
   of LongOpt:
-    let delimIdx = l.current.find(ValueDelims)
-    if delimIdx > 2:
-      l.valueIdx = delimIdx
-      result = (cmdLong, l.current[2..<delimIdx])
+    let sepIdx = l.current.find(ValueSeparators)
+    if sepIdx > 2:
+      # XXX: It is intentional that `--:` and `--=` is not rejected
+      # and falls into the alternative branch to avoid introducing syntax
+      # errors as a concept. However, this is open to change.
+      l.valueIdx = sepIdx
+      result = (cmdLong, l.current[2..<sepIdx])
     else:
       result = (cmdLong, l.current[2..^1])
       inc l.index
@@ -204,24 +210,21 @@ proc value*(l: var CmdLexer, delimitedOnly = false): Option[string] =
   if l.index < l.cmdline.len:
     # The current index is parsed and contains a value
     if l.valueIdx > 0:
-      # valueIdx is ValueDelims for either `--foo:bar` or `-o:bar`
-      if l.current[l.valueIdx] in ValueDelims:
-        result = some(l.current[l.valueIdx + 1..^1])
-
-      # This is a short option being parsed, return the remainder
-      # as a value if it is valid
+      # valueIdx is ValueSeparators for either `--foo:bar` or `-o:bar`
+      if l.current[l.valueIdx] in ValueSeparators:
+        result = some(l.current[(l.valueIdx + 1)..^1])
       elif not delimitedOnly:
+         # This is a short option being parsed, return the remainder
+         # as a value if it is valid
         result = some(l.current[l.valueIdx..^1])
-
-      # Do nothing otherwise
       else:
+        # Do nothing otherwise
         return
 
       l.valueIdx = 0
       inc l.index
-
-    # The current index is not parsed, treat it as a value if valid
     elif not delimitedOnly:
+      # The current index is not parsed, treat it as a value if valid
       result = some(l.current)
       inc l.index
 
@@ -236,10 +239,11 @@ iterator remaining*(l: var CmdLexer): string =
   ## short option returned and will raise `UnexpectedValueError`.
   if l.valueIdx > 0:
     if l.current.isLongOpt:
-      raise newUnexpectedValueError(l.current[0..<l.valueIdx], l.current[l.valueIdx + 1..^1])
+      raise newUnexpectedValueError(l.current[0..<l.valueIdx],
+                                    l.current[l.valueIdx + 1..^1])
     else:
       let value =
-        if l.current[l.valueIdx] in ValueDelims:
+        if l.current[l.valueIdx] in ValueSeparators:
           l.current[l.valueIdx + 1..^1]
         else:
           l.current[l.valueIdx..^1]
--- a/lib/experimental/lexopt.nim
+++ b/lib/experimental/lexopt.nim
@@ -1,0 +1,251 @@
+#
+#
+#                 NimSkull's runtime library
+#     (c) Copyright 2024 Leorize <leorize+oss@disroot.org>
+#
+# See the file "copying.txt", included in this distribution, for
+# details about copyright.
+
+import std/[strutils, options]
+
+## This module provides an user-driven command line lexer.
+##
+## Unlike `std/parseopt`, this lexer parses command line tokens as requested
+## by the caller.
+runnableExamples:
+  import std/options
+
+  type
+    Opts = object
+      paths: seq[string]
+      color: bool
+      output: string
+
+  var lexer = initCmdLexer(["--color", "/", "-o", "outfile.txt"])
+  var (kind, option) = lexer.next()
+  var opts = Opts()
+  while kind != cmdEnd:
+    if (kind == cmdLong and option == "color") or (kind == cmdShort and option == "c"):
+      opts.color = true
+    elif (kind == cmdLong and option == "output") or (kind == cmdShort and option == "o"):
+      let value = lexer.value()
+      if value.isNone:
+        quit "error: '--output' must be followed by a value"
+      opts.output = value.get()
+    elif kind == cmdValue:
+      opts.paths.add option
+
+    (kind, option) = lexer.next()
+
+  doAssert opts.paths == ["/"]
+  doAssert opts.color == true
+  doAssert opts.output == "outfile.txt"
+
+const ValueDelims* = {':', '='}
+  ## Characters that delimit between the option and its value
+
+type
+  CmdlineKind* = enum
+    cmdEnd ## End of command line
+    cmdLong ## A long option (ie. `--opt`)
+    cmdShort ## A short option (ie. `-o`)
+    cmdValue ## A value that is not an option
+
+  CmdLexer* = object
+    cmdline: seq[string] ## The command line to parse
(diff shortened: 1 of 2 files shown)


lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
+#
+#
+#                 NimSkull's runtime library
+#     (c) Copyright 2024 Leorize <leorize+oss@disroot.org>
+#
+# See the file "copying.txt", included in this distribution, for
+# details about copyright.
+
+import std/[strutils, options]
```
zerbina: Please move the imports below the module-level documentation.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+
+## This module provides an user-driven command line lexer.
```
zerbina:
```suggestion
## This module provides a user-driven command line lexer.
```

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+##
+## Unlike `std/parseopt`, this lexer parses command line tokens as requested
+## by the caller.
```
zerbina, line 12:
```suggestion
```
A very soft suggestion, given that `lexopt` is meant to replace `parseopt` at some point. It's totally fine to keep this reference around until `parseopt` is removed.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+      opts.paths.add option
+
+    (kind, option) = lexer.next()
+
+  doAssert opts.paths == ["/"]
+  doAssert opts.color == true
+  doAssert opts.output == "outfile.txt"
+
+const ValueDelims* = {':', '='}
+  ## Characters that delimit between the option and its value
```
zerbina:
```suggestion
  ## Characters that separate an option's name from its value
```
I think this phrasing is a bit clearer. Given that `ValueDelims` doesn't delimit the value (i.e., marks its end), maybe it should be named `ValueSep`?

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  CmdlineKind* = enum
+    cmdEnd ## End of command line
+    cmdLong ## A long option (ie. `--opt`)
+    cmdShort ## A short option (ie. `-o`)
+    cmdValue ## A value that is not an option
+
+  CmdLexer* = object
+    cmdline: seq[string] ## The command line to parse
+    index: int ## The index of `cmdline` to be processed
+    valueIdx: int ## \
```
zerbina:
```suggestion
    valueIdx: int
```
Minor suggestion. Most of the compiler's codebase doesn't use these.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  doAssert opts.output == "outfile.txt"
+
+const ValueDelims* = {':', '='}
+  ## Characters that delimit between the option and its value
+
+type
+  CmdlineKind* = enum
+    cmdEnd ## End of command line
+    cmdLong ## A long option (ie. `--opt`)
+    cmdShort ## A short option (ie. `-o`)
```
zerbina:
```suggestion
    cmdLong ## A long option (i.e. `--opt`)
    cmdShort ## A short option (i.e. `-o`)
```
Small nitpick.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+    index: int ## The index of `cmdline` to be processed
+    valueIdx: int ## \
+      ## For long options, the index of `ValueDelims`.
+      ##
+      ## For short options, the index of the next flag to be processed.
+      ##
+      ## Otherwise, it is unused.
+
+  LexError* = object of CatchableError
+    ## Errors occurred while lexing
```
zerbina:
```suggestion
    ## Errors occurring during lexing.
```

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+
+  UnexpectedValueError* = object of LexError
+    ## A value was provided but was not consumed
```
zerbina:
```suggestion
    ## A value was provided but not consumed.
```

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+    opt*: string ## The option that a value was provided to
+    value*: string ## The provided value
+
+proc initCmdLexer*(args: sink seq[string]): CmdLexer =
+  ## Creates a new `CmdLexer`. `args` should be the command line parameters as
+  ## returned by `os.commandLineArgs()`.
```
zerbina:
```suggestion
  ## returned by `commandLineParams <os.html#commandLineParams>`_.
```
Longer, but provides a proper link in the HTML documentation.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  CmdLexer(cmdline: args)
+
+proc initCmdLexer*(args: openArray[string]): CmdLexer =
+  ## Creates a new `CmdLexer`. `args` should be the command line parameters as
+  ## returned by `os.commandLineArgs()`.
```
zerbina:
```suggestion
  ## returned by `commandLineParams <os.html#commandLineParams>`_.
```

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+
+proc isLongOpt*(s: string): bool = s.startsWith("--")
+  ## Returns whether `s` can be parsed as a long option.
+proc isShortOpt*(s: string): bool = s.startsWith('-')
+  ## Returns whether `s` can be parsed as a short option.
+proc isDash(s: string): bool = s == "-"
+proc isDashDash(s: string): bool = s == "--"
+
+proc prefix*(kind: CmdlineKind): string =
+  ## Return the prefix string for a given command line parameter kind
```
zerbina:
```suggestion
  ## Returns the prefix string for a given command line parameter kind.
```

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  case kind
+  of cmdLong:
+    "--"
+  of cmdShort:
+    "-"
+  else:
+    ""
+
+func newUnexpectedValueError(opt, value: string): ref UnexpectedValueError =
```
zerbina:
```suggestion
func newUnexpectedValueError(opt, value: sink string): ref UnexpectedValueError =
```
Not critical, given the function's exceptional nature, but it does get rid of some copies.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  ##
+  ## 1. `--` and `-` are considered values.
+  ## 2. `--foo:bar` and `--foo=bar` are treated as long options with key `foo`
+  ##    and `bar` is considered the value of `foo`.
+  ##
+  ##    If this value is not consumed before the next call to `next()`, an error
+  ##    will be raised. See below for more details.
+  ## 3. `--foo` is treated as a long option with key `foo`.
+  ## 4. `-abco` is treated as into `a`, `b`, `c`, `o` short options, returned
+  ##    one per `next()` call.
```
zerbina:
```suggestion
  ## 4. `-abco` is equivalent to `-a`, `-b`, `-c`, and `-o` appearing in a
  #     sequence. Four calls to `next` are required to consume `-abco`. 
```
I think this is clearer (soft suggestion).

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  ## 5. `-abco:foo` and `-abco=foo` are treated as `a`, `b`, `c`, `o` short
+  ##    options, with `foo` being the value of option `o`.
```
zerbina:
```suggestion
  ## 5. `-abco:foo` and `-abco=foo` are equivalent to `-a`, `-b`, `-c`, and
  ##    `-o:foo` short options appearing in a sequence.
```
Similar to my above suggestion.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  ##
+  ##    If this value is not consumed before the next call to `next()`, an error
+  ##    will be raised. See below for more details
+  ## 6. Everything else is considered values.
```
zerbina: For completeness' sake, I think single short options (i.e., `-a`) should also be covered.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  ##
+  ## If the previous option have an unconsumed value (eg. `value()` was not
```
zerbina:
```suggestion
  ## If the previous option have an unconsumed value (e.g. `value()` was not
```

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  ## called for `--opt:foo`), `UnexpectedValueError` will be raised.
+  type
+    LexState {.pure.} = enum
```
zerbina: Why did you use `pure`? (it doesn't have to change, I'm just curious).

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+
+  case currentState
+  of LongOpt:
+    let delimIdx = l.current.find(ValueDelims)
+    if delimIdx > 2:
+      l.valueIdx = delimIdx
+      result = (cmdLong, l.current[2..<delimIdx])
+    else:
+      result = (cmdLong, l.current[2..^1])
+      inc l.index
```
zerbina, line 176: Not a strong opinion, but I personally think disallowing `--:` and `--=` would be better. If they stay allowed, I think there should be comment here mentioning that this is intentional.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  of ShortOpt:
+    l.valueIdx = if l.valueIdx == 0: 1 else: l.valueIdx
+    result = (cmdShort, $l.current[l.valueIdx])
```
zerbina, line 179: Same here.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  ## 1. `--foo:bar` or `--foo=bar` yields `bar` if called after `next()`
+  ##    returns `foo`.
+  ## 2. `-abco:foo` or `-abco=foo` yields foo if called after `next()` returns
+  ##    `o`.
+  ## 3. If `delimitedOnly` is true, syntaxes below will not be considered.
+  ## 4. `-abcooutput.txt` yields `output.txt` if called after `next()` returns
+  ##    `o`.
+  ## 5. The current parameter is treated as a value, regardless of whether it
+  ##    can be considered as an option (ie. `--foo --bar` will yield `--bar` if
+  ##    this procedure is called after `--foo`).
```
zerbina, line 209:
```suggestion
  ## 
  ## If `delimitedOnly` is false, the following syntaxes are considered:
  ##
  ## 1. `-abcooutput.txt` yields `output.txt` if called after `next()` returns
  ##    `o`.
  ## 2. The current parameter is treated as a value, regardless of whether it
  ##    can be considered as option (i.e. `--foo --bar` will yield `--bar` if
  ##    this procedure is called after `--foo`).
```
Soft suggestion. I think having each bullet point describe an interpretation makes the comment easier to skim.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  if l.index < l.cmdline.len:
+    # The current index is parsed and contains a value
+    if l.valueIdx > 0:
+      # valueIdx is ValueDelims for either `--foo:bar` or `-o:bar`
+      if l.current[l.valueIdx] in ValueDelims:
+        result = some(l.current[l.valueIdx + 1..^1])
```
zerbina:
```suggestion
        result = some(l.current[(l.valueIdx + 1)..^1])
```
So that one doesn't have to know the operator precedence (soft suggestion).

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+      else:
+        return
+
+      l.valueIdx = 0
+      inc l.index
+
+    # The current index is not parsed, treat it as a value if valid
+    elif not delimitedOnly:
+      result = some(l.current)
+      inc l.index
```
zerbina:
```suggestion
      elif not delimitedOnly:
        # This is a short option being parsed, return the remainder
        # as a value if it is valid
        result = some(l.current[l.valueIdx..^1])
      else:
        # Do nothing otherwise
        return

      l.valueIdx = 0
      inc l.index
    elif not delimitedOnly:
      # The current index is not parsed, treat it as a value if valid
      result = some(l.current)
      inc l.index
```
A purely stylistic suggestion. The spacing between the branches and the comments on the same indentation level are not common in this codebase.

lib/experimental/lexopt.nim
```diff
@@ -0,0 +1,251 @@
...
+  ##
+  ## If the previous option returned from `next()` have a value that was not
+  ## consumed by `value()`, `UnexpectedValueError` will be raised.
+  ##
+  ## As a special case, any short options within a bundle (ie. `-abcde`) that
+  ## have not been processed by `next()` is considered the value of the last
+  ## short option returned and will raise `UnexpectedValueError`.
+  if l.valueIdx > 0:
+    if l.current.isLongOpt:
+      raise newUnexpectedValueError(l.current[0..<l.valueIdx], l.current[l.valueIdx + 1..^1])
```
zerbina:
```suggestion
      raise newUnexpectedValueError(l.current[0..<l.valueIdx],
                                    l.current[(l.valueIdx + 1)..^1])
```
To keep the line length below 80 character.

tests/stdlib/experimental/tlexopt.nim
```diff
@@ -0,0 +1,505 @@
+discard """
+description: "Tests for the lexopt module"
```
zerbina:
```suggestion
  description: "Tests for the lexopt module"
  targets: "c js vm"
```
To increase test coverage for the JS and VM backends.

zerbina (commented):
Overall, this looks good to me. I like the flexibility in parsing command line options `lexopt` provides through a single-yet-simple interface. Also, thank you for providing an exhaustive test suite.

My suggestions are mostly regarding grammar and code style. One addition to maybe consider is something like `dropValue` (or similar), which efficiently skips the current value without returning it.

With `lexopt` there's now three separate modules (`lexopt`, `parseopt`, and `parseopt2`) providing command-line parsing/lexing, which I think is two too many. For backwards compatibility, both `parseopt` and `parseopt2` could be changed to use `lexopt` internally, so that there's no duplication of lexer code, but that's something for a future PR.

lib/experimental/lexopt.nim
alaviss: I think this can be kept for now.

lib/experimental/lexopt.nim
alaviss: IIRC not having the continuation mark causes the docs to become top-level, but I'll have to test that.

lib/experimental/lexopt.nim
alaviss: It's a habit whenever I decide not to have an enum prefix.

I think mainline Nim now have module-like resolution for them and you don't need `{.pure.}` anymore, but I'm keeping it here to be safe.

lib/experimental/lexopt.nim
alaviss: I'm not big in introducing the concept of syntax errors for command line, where no rigid grammar exists, which is why these cases are in a bit of a limbo.

I decided to have them parse into something that can be caught by simple code, for example:

```nim
let (kind, opt) = lexer.next()
case kind
of cmdLong:
  case opt
  of "foo":
  of "bar":
  else:
    # error handling would handle `--=something` automatically
```

I can't say that this is the best way to handle it, but I think keeping this as an "implementation-defined" case is not a bad idea, hence the lack of any comments confirming whether this is intended or not.

lib/experimental/lexopt.nim
alaviss: I think I still prefer the numbered form more.

alaviss:
> One addition to maybe consider is something like dropValue (or similar), which efficiently skips the current value without returning it.

That was actually removed in favor of

```nim
discard lexer.value()
```

since this captures the intent a lot better.

> For backwards compatibility, both parseopt and parseopt2 could be changed to use lexopt internally, so that there's no duplication of lexer code, but that's something for a future PR.

I agree, but my focus will be on `experimental/cmdline`, which would be the main user of this module.

lib/experimental/lexopt.nim
zerbina: The absence of `## \` does not cause the doc comment to become top-level. I've double-checked before suggestion the removal.

lib/experimental/lexopt.nim
zerbina: For the record, you don't need `.pure` to be able to access the enum fields through a dot access, nor does `.pure` enforce that you have to use a dot access.

Looking at the implementation, I'd recommend not using `.pure` for enums defined in procedures at the moment, as they ignore scoping rules:
```nim
proc p() =
  type Test {.pure.} = enum
    A

let x = A # compiles fine
```

lib/experimental/lexopt.nim
zerbina: The case being implementation-defined and subject to change is fine, in my opinion, but due the lack of comment, I initially thought that this was a case you hadn't considered. Only after looking at the tests did it became clear that the behaviour is intended.

With comment, I didn't mean that it needs to be mentioned in the doc comment, just that there should be a comment above the marked code saying something along the lines of "`--:` not being rejected is deliberate", so that future readers don't trip on it like I did.

zerbina:
> That was actually removed in favor of
> 
> ```nim
> discard lexer.value()
> ```
> 
> since this captures the intent a lot better.

Ah, okay. I thought providing a tool to eliminate the unnecessary copy/destruction might be a good idea (given the PRs overall focus on no unnecessary copies), but having the smaller surface area is probably worth more.

lib/experimental/lexopt.nim
alaviss: A comment was added to the other branch instead.

zerbina (approved):
Looks good to me.

PR #1495, opened 2025-03-01
compiler: add the `--timetrace` switch

zerbina:
## Summary

Add the `--timetrace` (`off` by default) switch for enabling or
disabling time tracing in the compiler. When enabled, a `.json`
trace (using the Chrome Trace Event format) of what the compiler did
and when is dumped next to the compiled module.

## Details

The feature is intended to aid compiler development and help users
analyze compile times. It is built upon the existing `tracer` module.

Instrumented are:
* top-level statement parsing
* semantic analysis of top-level statements and routines
* whole module processing
* `transf`
* compile-time code generation and execution
* cursor inference, MIR generation, and MIR processing
* final C, JS, and VM code generation
* backend C compiler and linker invocation

If tracing is disabled, the events are still recorded, but no dump of
them is created. The JSON file name is made up of the project's name
plus the current time and date, so that traces aren't overwritten.

Due to a csources compiler bug, enums used at compile-time as an
array's index type enum can only have 128 elements, and thus the
(unused) `seqsv2` switch had to be removed in order to add the
`timetrace` switch (`CmdSwitchTextKind` is the problematic enum).

### Tracer Adjustments

* the obsolete `tikNimScript` event group is removed
* the `tikMirgen` and `tikPasses` event groups are added
* a `traceLoc` overload accepting a statement to trace is added
* `trace_dump` uses `escapeJson` instead of `escape` to escape strings

### DIFF
--- a/compiler/ast/parser.nim
+++ b/compiler/ast/parser.nim
@@ -48,6 +48,7 @@ import
     pathutils,
     astrepr,
     idioms,
+    tracer
   ]
 
 type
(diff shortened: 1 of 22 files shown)


saem (approved):
Nice, some extra timing information should help track improvements, or regressions, over time.

PR #1496, opened 2025-03-16
cmdline: new module for building CLIs

alaviss:
## Summary

Introducing  `cmdline` , a new experimental module aiming to simplify
the process of building command line interfaces.

## Details

The standard library currently lacks an option for quickly spinning up
command line tools. While  `parseopt`  and  `lexopt`  can handle the
process of reading command line parameters, they are not equipped to:

- Generate help messages for a command
- Parse the command line corresponding to a declared interface
- Automatically handle errors in user input and print helpful
diagnostic

`cmdline`  is designed to fulfill those requirements, and to become the
standard method to construct command line interfaces.

A key part in the design of this module is to facilitate support for
other frontends. Many applications now feature an interface that
provides:

- Colored errors
- Colored help messages
- Producing auto correct suggestions on errors
- etc.

While  `cmdline`  can implement all of these features on its own, it
would balloon the complexity of the module beyond what is reasonable for
a standard library component (factoring in the current maintainer
bandwidth). As such a few design constraints was applied:

- The core structures ( `Cli`  and  `parse()` ) are tightly coupled for
both simplicity and to enforce a standardized interpretation.
- Core structures avoid the use of meta programming: a  `Cli`  must be
able to be built using only  `proc` s. This provide a stable basis for
sugar like DSLs and automatic generators to work with.
- Auxiliary functions like  `help()` ,  `run()`  are implemented using
**only** public interfaces, making sure that user can bring their own
implementation. Additionally, the components that generate portions of
help messages/diagnostics are also made public, allowing a custom
composite to be made.


lib/experimental/cmdline/parsers.nim
```diff
@@ -0,0 +1,52 @@
...
+#
+# See the file "copying.txt", included in this distribution, for
+# details about copyright.
+
+## This module defines default command line parsers for primitive types
+
+import std/strutils
+import std/typetraits
+
+proc parseCli*(T: typedesc[SomeInteger], value: string): T =
```
saem, line 65: `tryParseCli` variants that don't raise but instead return a default value (taken as a parameter) would be handy, as people will be doing a lot of that.

saem:
Not sure what your plan for docs is, but I recommend only covering the public elements and making it largely runnable examples. I've found the lack of the latter has been the biggest gap in docs. Not to mention it helps with setting us up for LLMs, because they're so example driven.

lib/experimental/cmdline/parsers.nim
alaviss: I'm not sure about the value of that for CLI purposes. More often than not you want to error out because wrong inputs from user meant that they make a mistake and shouldn't be ignored. For general parsing `parseutils` is what you want to use.

Other than that, a `try` variant never make much sense to me, as it'd require you to always construct a (potentially expensive) object regardless of parsing results. I could entertain the idea of using `Result[T]`, which would come with free support for defaults, but the ecosystem at the moment deals in exceptions.

saem: The idea with try variants is that you use in band error values to avoid the crappiness of `try...except` wrapping.

`Result[T]` is better, I keep forgetting we have one in the stdlib/experimental (I remember it as for the compiler). 😆

alaviss: I elected to not use `Result[T]` for now.

Most of the APIs an user would call to aid with parsing raises `ValueError` themselves, and at present this makes it a lot simpler to write a custom `parseCli` implementation, which I expect to happen more than explicitly invoking these.

We could add a capturing feature to `Result[T]` which would automatically turn exceptions into an error `Result`, which then an user can just do something like:

```nim
let val = capture(parseCli(int, "not int")).valueOr(-1)
```

and would work for all existing APIs, not just this one.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,1945 @@
...
+    alias: Table[Parameter, seq[string]] ## Mapping of Parameter to aliases.
+    parent: Store[Parameter, Parameter] ## Mapping of Parameter to their parent.
+
+    # Documentation storage
+    #
+    # Might be useful to support not having these for space-constrained
+    # targets using a define.
+    usage: Store[Parameter, string] ## Canonical usage for Parameters.
+    placeholder: Store[Parameter, string] ## Canonical placeholder for
+                                          ## Parameters. Only used for flags.
```
zerbina: Not a blocker, but why not put the doc comment on the next line? Re-formating the text in case of name/type changes becomes simpler that way.

alaviss: Force of habit

doc/lib.rst
```diff
@@ -386,6 +383,18 @@ Parsers
   to be somewhat error-correcting, so that even some "wild HTML" found on the
   web can be parsed with it.
 
+Command line parsers
```
saem:
```suggestion
Command Line Parsers
```

Please use title case for titles

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2205 @@
...
+##
+## Options:
+##   --help           display help message
+##   -f, --force      force removal
+##   -r, --recursive  recursively remove files
+## ```
+
+type
+  Cli*[T] {.requiresInit.} = object
+    ## A command line interface.
```
saem:
```suggestion
    ## A command line interface description.
```

It's a bit more accurate

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2205 @@
...
+    paramName*: string ## Name of the parameter triggering help, as specified by
+                       ## input.
+    param*: Parameter ## Handle to the parameter triggering help.
+
+  ParameterKind* {.pure.} = enum
+    Command
+    Flag
+    Positional
+
+  Parameter* = distinct uint32
```
saem:
```suggestion
  ParameterId* = distinct uint32
```

I would suffix these with `Id`, so it's clear that they need to be passed to something in order to be useful/get more information, etc.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2205 @@
...
+  ## This command can then be matched using any of the provided `names` in
+  ## addition to its canonical name.
+  ##
+  ## See also:
+  ## - `name func <#name,sinkCommandBuilder[T],string>`_
+  result = b
+  result.aliases.setLen(0)
+
+  # Not the fastest method, but it's expected that users will
+  # specify at most 4 of these.
```
saem:
```suggestion
  # Not the fastest method, but it's assumed that users will
  # specify a handful (approximately 1-4) of these.
```

`assumed`, unlike `expected`, doesn't imply we might, or could in the future, enforce this

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2205 @@
...
+type
+  Cli*[T] {.requiresInit.} = object
+    ## A command line interface.
+    ##
+    ## See also:
+    ## - `initCli proc <#initCli,sinkCommandBuilder[T]>`_
+    command: Table[Command, CliCommand] ## Lookup mapping of command to lookup tables.
+    parser: Store[Parameter, ParserAny[T]] ## Parser to process input for Parameter.
+    name: Store[Parameter, string] ## Canonical names for all Parameters.
+    alias: Table[Parameter, seq[string]] ## Mapping of Parameter to aliases.
```
saem: Just so it's more self documenting and avoids some potential gotchas, we should probably use a set like structure instead of a `seq` to store the list of aliases, to avoid duplicates.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2805 @@
...
+    "unexpected error parsing command line: " & error.msg
+
+proc run*[T](
+  cli: Cli[T],
+  accumulator: var T,
+  args: sink seq[string] = commandLineParams(),
+  messageOutput: File = stdmsg,
+) =
+  ## Parses the command line, collecting changes made by added parsers
+  ## to `accumulator`.
```
saem:
```suggestion
  ## Parses the command line `args` based on the description in `cli`, with configured
  ## parsers updating values in the `accumulator`.
```

Makes it a little bit more clear as to what's going on.

saem (commented):
Did another pass and came up with some suggestions

lib/experimental/cmdline.nim
alaviss: We don't have any such thing at the moment :disappointed: 

Right now the places that modify the table actually perform the required checks, so I'd say that it should be fine.

saem: Bummer, we could really use a set like that. Although, sets push the need for interfaces pretty hard, because there are so many different implementations. 😬

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
+#
+#
+#                 NimSkull's runtime library
+#     (c) Copyright 2025 Leorize <leorize+oss@disroot.org>
+#
+# See the file "copying.txt", included in this distribution, for
+# details about copyright.
+
+from os import commandLineParams
```
zerbina: Could you remove the usage of `commandLineParams`?

It would allow using `cmdline` on a broader set of targets, such as NimScript, standalone VM execution (`nim vm`), and JavaScript code running in the browser. Ideally, no such concessions should be necessary to support other backends, but at the moment, they are.

Separate from the above, I also think not a good idea to hide where the command line comes from (by using `commandLineParams()` as the default value for `run.args`).

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+import std/sequtils
+import std/strutils
+import std/tables
+
+import std/private/containers
+
+import cmdline/parsers
+export parsers
+
+import lexopt
```
zerbina:
```suggestion
import
  std/[
    algorithm,
    hashes,
    options,
    sequtils,
    strutils,
    tables
  ],
  std/private/[
    containers
  ],
  cmdline/[
    parsers
  ],
  experimental/[
    lexopt
  ]

export parser
```
It's a tad longer, but new code generally uses import sections and blocks.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+
+## This module implements a fast and lightweight declarative command line
```
zerbina, line 9: Please move the module-level documentation above the imports.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+## ```
+
+type
+  Cli*[T] {.requiresInit.} = object
+    ## A command line interface description.
+    ##
+    ## See also:
+    ## - `initCli proc <#initCli,sinkCommandBuilder[T]>`_
+    command: Table[CommandId, CliCommand]
+      ## Lookup mapping of command to lookup tables.
```
zerbina, line 311: These comments feel a bit redundant -- the type already shows what the field is meant for. I also think the fields here should use plural (e.g., `commands` instead of `command`).

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    ShowHelp ## Abort and show help message.
+    DisableFlagProcessing ## ParameterIds following this will be considered
+                          ## to be values.
+
+  MaybeAction* = Action | void
+    ## Typeclass to support parsers returning either `Action` or nothing.
+
+  FlagParser*[T; R: MaybeAction] = proc (option, value: string, accumulator: var T): R
+    ## A parser for flag with name `option` and value `value`. The accumulator
+    ## passed to `run()`_ or `parse()`_ can be accessed and modified via
```
zerbina:
```suggestion
    ## passed to `run`_ or `parse`_ can be accessed and modified via
```
Could you remove the parenthesis? I don't think provide any benefit to clarity, especially given that the link is a local one.

(The same goes for other run()_ and parse()_ occurrences below)

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    ## `accumulator`.
+    ##
+    ## A parser may choose to return an `Action` or not return anything,
```
zerbina: I known that inlining is not possible, but the doc comment duplication is a bit much, and I'd prefer if the expectations of a parser are put under a "Writing Parsers" (or something along that line) section in the top-level documentation, to which every parser type then links.

Autocomplete user won't have immediate access to the information then (unless using an IDE that supports following links in documentation boxes), but I don't think it's super important for the types.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    ## which will be interpreted as `Action.Continue`.
+    ##
+    ## A parser may raise `ValueError` for invalid inputs, which will be
+    ## processed accordingly by the library. Any other exceptions signify
+    ## an internal error and will not be handled automatically.
+    ##
+    ## .. _run(): #run,Cli[T],T,sinkseq[string],File
+    ## .. _parse(): #parse,Cli[T],T,sinkseq[string]
```
zerbina:
```suggestion
    ## .. _run: #run,Cli[T],T,sinkseq[string],File
    ## .. _parse: #parse,Cli[T],T,sinkseq[string]
```

doc/lib.rst
```diff
@@ -386,6 +383,18 @@ Parsers
...
 
+Command Line Parsers
+++++++++++++++++++++
+
+* `experimental/cmdline <cmdline.html>`_
+  The `cmdline` module implements a lightweight declarative command-line parser,
+  providing an easy to use solution for command line interfaces.
+
+* `experimental/lexopt <lexopt.html>`_
+  The `lexopt` module implements an user-driven command-line option lexer.
```
zerbina:
```suggestion
  The `lexopt` module implements a user-driven command-line option lexer.
```

zerbina (commented):
I've gathered some first suggestions/remarks, mostly regarding style, some typos, and documentation, but I haven't looked at all code yet.

lib/experimental/cmdline.nim
alaviss: I think having the default makes usage a lot simpler. `run()` is meant to take the wheel and handle all the boring stuff for you, including process termination, and is relatively simple on it's own. For a more controlled experience, `parse()` is what you want.

I could use a `when` to remove `commandLineParams` as default for targets without command line parameters, or simply remove `run` for them altogether, if that works.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    # targets using a define.
+    usage: Store[ParameterId, string] ## Canonical usage for ParameterIds.
+    placeholder: Store[ParameterId, string]
+      ## Canonical placeholder for ParameterIds. Only used for flags.
+
+  CliCommand = object
+    ## Lookup table for flags and positionals.
+    flag: Table[string, ParameterId]
+      ## Lookup mapping of flag names to ParameterId.
+    command: Table[string, CommandId]
```
zerbina: I think it'd be clearer if the field is called `subcommands`. At least the documentation should mention that it stores the subcommands.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    aliases: seq[string]
+    usage: string
+    placeholder: string
+
+  PositionalBuilder*[T] = object
+    ## Builder for command line positional parameters.
+    ##
+    ## See also:
+    ## - `positionalBuilder proc <#positionalBuilder,Cli[T]>`_
+    posParser: ParserAny[T]
```
zerbina: Could you use `PositionalParser` directly, with boolean fields for catch-all and optional? The implementation would become clearer and simpler, in my opinion.

lib/experimental/cmdline.nim
alaviss: > These comments feel a bit redundant -- the type already shows what the field is meant for.

The comments are mostly seeds for additional details (if needed) in the future. I could drop them if wanted, of course.

> I also think the fields here should use plural (e.g., `commands` instead of `command`).

I think they are better in singular form, the reason being while they technically store multiples of X, the usage of them is usually a direct binding with an ID, so I prefer to name them as if I'm dealing with a single object (and if we ever have a SoA feature this comes in handy).

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    posName: string
+    usage: string
+
+  CommandBuilder*[T] = object
+    ## Builder for command line subcommands.
+    ##
+    ## See also:
+    ## - `commandBuilder proc <#commandBuilder,Cli[T]>`_
+    cmdParser: ParserAny[T]
```
zerbina: The comment above also applies here, just with `CommandParser`.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    aliases: seq[string]
+    usage: string
+    isDefault: bool
+
+  Action* {.pure.} = enum
+    ## Action to be taken after parsing.
+    Continue ## Continue parameter parsing.
+    ShowHelp ## Abort and show help message.
+    DisableFlagProcessing ## ParameterIds following this will be considered
+                          ## to be values.
```
zerbina: The comment is in conflict with the name. The name says that only flag processing is disabled, whereas the documentation says that commands, positionals, and flags will be considered values (presumably for flags).

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+
+  MaybeAction* = Action | void
+    ## Typeclass to support parsers returning either `Action` or nothing.
+
+  FlagParser*[T; R: MaybeAction] = proc (option, value: string, accumulator: var T): R
+    ## A parser for flag with name `option` and value `value`. The accumulator
```
zerbina: I don't get the point of having the `option` parameter. Wouldn't a separate procedure be used instead of dispatching on the `option` within the procedure?

From what I've seen, none of the tests and runnable examples actually make use of `option`, too.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    ##
+    ## The parser may raise `ValueError` for invalid inputs, which will be
+    ## processed accordingly by the library. Any other exceptions signify
+    ## an internal error and will not be handled automatically.
+    ##
+    ## .. _run(): #run,Cli[T],T,sinkseq[string],File
+    ## .. _parse(): #parse,Cli[T],T,sinkseq[string]
+
+  CommandParser*[T; R: MaybeAction] = proc (command: CommandId, accumulator: var T): R
+    ## A parser for command parameter with command `command`. The accumulator
```
zerbina, line 414: From what I can tell, a `CommandParser` doesn't actually parse anything in the common sense of the word. Rather, it's meant to *process* a command appearing.

Personally, I think using a different name would be better. "processor" could work, although "command processing" usually has a slightly different meaning.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    ## which will be interpreted as `Action.Continue`.
+    ##
+    ## The parser may raise `ValueError` for invalid inputs, which will be
+    ## processed accordingly by the library. Any other exceptions signify
+    ## an internal error and will not be handled automatically.
+    ##
+    ## .. _run(): #run,Cli[T],T,sinkseq[string],File
+    ## .. _parse(): #parse,Cli[T],T,sinkseq[string]
+
+  FlagTypedParser*[T; U; R: MaybeAction] = proc (option: string, value: U, accumulator: var T): R
```
zerbina: Using "processor" for the name would fit better, I think, as the actual parsing is delegated to `parseCli`.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ParserKind {.pure.} = enum
+    Command
+    Flag
+    FlagOptionalValue
+    Positional
+    OptionalPositional
+    CatchAll
+    OptionalCatchAll
+
+  ParserAny[T] = object
```
zerbina, line 442: Have you considered only using `FlatOptionalParser` internally? `FlagBuilder.parser` would create a wrapper, meaning there would be no change in public interface.

The benefit is that you then don't need to use `ParserAny` in `FlagBuilder`, which would improve clarity of the code by a good bit, in my opinion.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    case kind: ParserKind
+    of FlagOptionalValue: optParser: FlagOptionalParser[T, Action]
+    of ParserKind.Flag: parser: FlagParser[T, Action]
+    of ParserKind.Positional, OptionalPositional, CatchAll,
+       OptionalCatchAll: posParser: PositionalParser[T, Action]
+    of ParserKind.Command: cmdParser: CommandParser[T, Action]
```
zerbina:
```suggestion
    of FlagOptionalValue:
      optParser: FlagOptionalParser[T, Action]
    of ParserKind.Flag:
      parser: FlagParser[T, Action]
    of ParserKind.Positional, OptionalPositional, CatchAll,
       OptionalCatchAll:
      posParser: PositionalParser[T, Action]
    of ParserKind.Command:
      cmdParser: CommandParser[T, Action]
```
Longer, but arguably also easier to read.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  result = b
+  result.cmdName = name
+  result.aliases.keepItIf: it != result.cmdName
+
+func name*[T](b: sink FlagBuilder[T], name: string): FlagBuilder[T] =
+  ## Sets the canonical name of this flag, which is used to identify
+  ## this flag on the command line.
+  ##
+  ## If `name` is one-character long, it can also be recognized on the command
+  ## line using the short form syntax (e.g. `-n`).
```
zerbina:
```suggestion
  ## If `name` is only one ASCII character long, it may use the short form
  ## syntax (e.g. `-n`).
```
A soft suggestion. It's a bit more concise and also makes it clear that non-ASCII Unicode character cannot use the short form.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  for name in names.items:
+    if name == result.cmdName or name in result.aliases:
+      continue
+    result.aliases.add names
+
+func alias*[T](b: sink FlagBuilder[T], names: varargs[string]): FlagBuilder[T] =
+  ## Sets aliases for this flag.
+  ##
+  ## The command can be matched using any of the provided `names` in addition to
+  ## its canonical name.
```
zerbina:
```suggestion
  ## The flag can be matched using any of the provided `names` in addition to
  ## its canonical name.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  # Not the fastest method, but it's assumed that users will
+  # specify a handful (approximately 1-4) of these.
+  for name in names.items:
+    if name == result.flagName or name in result.aliases:
+      continue
+    result.aliases.add names
+
+func default*[T](b: sink CommandBuilder[T]): CommandBuilder[T] =
+  ## Marks this command as the default subcommand of the current dispatcher.
+  ## When no subcommand are specified on the command line, this command
```
zerbina:
```suggestion
  ## When no subcommand is present on the command line, this command
```
A soft suggestion. I think it's slightly easier to understand this way.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ## will be selected.
+  ##
+  ## Only one default subcommand is permitted per command. Attempting to
+  ## add more than one is considered an error.
```
zerbina:
```suggestion
  ## Only one default subcommand is permitted per command. Attempting to
  ## add another default subcommand to a command will cause an error.
```
A soft suggestion. The new wording is more definitive.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+      .addTo(cli, RootCommand)
+
+    doAssert cli.parse(@[]) == "acted"
+
+  result = b
+  result.isDefault = true
+
+func catchAll*[T](b: sink PositionalBuilder[T]): PositionalBuilder[T] =
+  ## Marks this positional as "catch all". All positional parameters encountered
+  ## starting at this positional will be handled by the associated parser.
```
zerbina, line 727:
```suggestion
  ## Marks this positional as "catch all". All positional parameters
  ## appearing beyond this positional will be handled by the associated
  ## parser.
```
Soft suggestion.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+      .parser((v, var s) => s.add v)
+      .addTo(cli)
+
+    doAssert cli.parse(@["a", "b", "c"]) == ["a", "b", "c"]
+
+  result = b
+  let parser =
+    case result.posParser.kind
+    of ParserKind.Command..FlagOptionalValue:
+      nil
```
zerbina: If I'm reading the code correctly, this case is can never happen, no? If you don't opt for removing the `ParserAny` usage for `PositionalBuilder`, I think the case expression here should be removed.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+      .addTo(cli)
+
+    doAssert cli.parse(@[]) == ""
+    doAssert cli.parse(@["a"]) == "a"
+
+  result = b
+  let parser =
+    case result.posParser.kind
+    of ParserKind.Command..FlagOptionalValue:
+      nil
```
zerbina: The above comment also applies here.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  b.parser(
+    proc (command: CommandId, accumulator: var T): Action =
+      p(command, accumulator)
+  )
+
+func optionalParser*[T](
+  b: sink FlagBuilder[T],
+  p: sink FlagOptionalParser[T, Action],
+): FlagBuilder[T] =
+  ## Sets the parser for this flag, and mark the flag as not requiring any value.
```
zerbina:
```suggestion
  ## Sets the parser for this flag, and marks the flag as not requiring any
  ## value.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## Flags with optional value will only receive their value when it is
+  ## specified inline, for example: `--flag=value` or `--flag:value`.
```
zerbina:
```suggestion
  ## For flags with an optional value, their parser will only be called when
  ## the value is specified inline, for example: `--flag=value` or
  ## `--flag:value`.
```
Soft suggestion. I think the meaning of "receive" is not entirely clear in this context.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+
+  result = b
+  result.flagParser = ParserAny[T](kind: FlagOptionalValue, optParser: p)
+
+func optionalParser*[T](
+  b: sink FlagBuilder[T],
+  p: sink FlagOptionalParser[T, void],
+): FlagBuilder[T] =
+  ## Sets the parser for this flag, and mark the flag as not requiring any value.
+  ## On successful parse, `Action.Continue` is taken as the default action.
```
zerbina:
```suggestion
  ## Sets the parser for this flag with `Action.Command` as the default value,
  ## and marks the flag as not requiring any value.
```
For symmetry with `CommandBuilder.parser`.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## Flags with optional value will only receive their value when it is
+  ## specified inline, for example: `--flag=value` or `--flag:value`.
```
zerbina, line 871:
```suggestion
  ## For flags with an optional value, their parser will only be called when
  ## the value is specified inline, for example: `--flag=value` or
  ## `--flag:value`.
```
Soft suggestion. I think the meaning of "receive" is not entirely clear in this context.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+
+    doAssertRaises(HelpError):
+      discard cli.parse(@["--string", "help"])
+
+  result = b
+  result.flagParser = ParserAny[T](kind: ParserKind.Flag, parser: p)
+
+func parser*[T](b: sink FlagBuilder[T], p: sink FlagParser[T, void]): FlagBuilder[T] =
+  ## Sets the parser for this flag, and mark the flag as requiring values.
+  ## On successful parse, `Action.Continue` is taken as the default action.
```
zerbina:
```suggestion
  ## Sets the parser for this flag with `Action.Continue` as the default
  ## action, and marks the flag as requiring a value.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    proc (option: string, value: string, accumulator: var T): Action =
+      p(option, value, accumulator)
+  )
+
+func optionalParser*[T, U; R: MaybeAction](
+  b: sink FlagBuilder[T],
+  _: typedesc[U],
+  parser: sink FlagOptionalTypedParser[T, U, R],
+): FlagBuilder[T] =
+  ## Sets the parser for this flag, and mark the flag as not requiring any value.
```
zerbina:
```suggestion
  ## Sets the parser for this flag, and marks the flag as not requiring any
  ## value.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## Input string values from the command line will first be parsed using
+  ## `parseCli` before handing off to the parser. See `parsers module`_ for
+  ## more information.
+  ##
+  ## Flags with optional value will only receive their value when it is
+  ## specified inline, for example: `--flag=value` or `--flag:value`.
```
zerbina, line 978:
```suggestion
  ## For flags with an optional value, the parser will only be called when
  ## the value is specified inline, for example: `--flag=value` or
  ## `--flag:value`.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+        let value = value.map(proc (x: string): U = parseCli(U, x))
+        parser(option, value, r)
+    )
+
+proc parser*[T, U; R: MaybeAction](
+  b: sink FlagBuilder[T],
+  _: typedesc[U],
+  parser: sink FlagTypedParser[T, U, R],
+): FlagBuilder[T] =
+  ## Set the parser for this flag, and mark the flag as requiring values.
```
zerbina:
```suggestion
  ## Sets the parser for this flag, and marks the flag as requiring values.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## Input string values from the command line will first be parsed using
+  ## `parseCli` before handing off to the parser. See `parsers module`_ for
+  ## more information.
+  ##
+  ## When `U` is `bool`, this flag behaves like a switch and does not
+  ## require a value to be passed. If values are to be passed, it must be
+  ## inlined (i.e. `--flag=false`).
+  ##
+  ## Flags with required value can receive any of the following forms:
```
zerbina:
```suggestion
  ## Flags with a required value can receive any of the following forms:
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##   satisfactory results with multi-line `usage`.
+  result = b
+  result.usage = usage
+  result.placeholder = placeholder
+
+func describe*[T](
+  b: sink PositionalBuilder[T],
+  usage: sink string,
+): PositionalBuilder[T] =
+  ## Sets a short `usage` description and a value `placeholder` for this flag.
```
zerbina:
```suggestion
  ## Sets a short `usage` description for this positional.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## .. important::
+  ##   Input parameters are assumed to meet operational constraints.
+  result = cli.name.add(name)
+  discard cli.usage.add(usage)
+  discard cli.placeholder.add(placeholder)
+  discard cli.parser.add(parser)
+  discard cli.parent.add(parent)
+
+func initCli*[T](b: sink CommandBuilder[T]): Cli[T] =
```
zerbina:
```suggestion
func initCli*[T: not void](b: sink CommandBuilder[T]): Cli[T] =
```
You'll get a rather cryptic error when passing a `CommandBuilder[void]` instance to `initCli`, otherwise.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## The root command must:
+  ##
+  ## - Have no `parser` set.
+  ## - Have no `aliases`.
+  ##
+  ## Unlike subcommands, the root command may:
+  ##
+  ## - Have no `name` set. The `name` of a root command is used for
+  ##   documentation generation and does not participate in parsing.
```
zerbina: I don't understand this comment. What happens if a name *is* set?

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## The root command can be referred to using `RootCommand` constant.
```
zerbina:
```suggestion
  ## The root command can be referred to using the `RootCommand` constant.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  command: CommandId,
+): CommandId {.discardable.} =
+  ## Adds the command specified by `b` as a subcommand of `command`.
+  ##
+  ## A subcommand can only be added to `command` if:
+  ##
+  ## - `command` does not contain any positional parameters.
+  ## - A non-empty `name` must be set.
+  ## - None of the names specified by either `name` or `alias` are added to
+  ##   `command` previously.
```
zerbina:
```suggestion
  ## - `command` has no positional parameters.
  ## - The subcommand has a non-empty name.
  ## - Neither the name nor aliases are used by previously added subcommands.
```
So that the items all read as predicates.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## `ValueError` will be raised if any of the specified constraints are
+  ## violated.
+  ##
+  ## See also:
+  ## - `commandBuilder proc <#commandBuilder,Cli[T]>`_
+  assert b.cmdName.len > 0, "Command name must not be empty"
```
zerbina: According to the doc comment this should be an error. I think it's fine if an assertion is used, but then the documentation needs to be adjusted.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  command: CommandId = RootCommand,
+): FlagId {.discardable.} =
+  ## Adds the flag specified by `b` to `command`.
+  ##
+  ## A flag can only be added to `command` if:
+  ##
+  ## - A non-empty `name` must be set.
+  ## - None of the names specified by either `name` or `alias` are added to
+  ##   `command` previously.
+  ## - A `parser` or `optionalParser` must be set.
```
zerbina:
```suggestion
  ## - The flag has a non-empty name.
  ## - Neither the name nor aliases are used by previously added flags.
  ## - A parser or optional parser was set for the flag.
```
So that the items all read as predicates.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## `ValueError` will be raised if any of the specified constraints are
+  ## violated.
+  ##
+  ## See also:
+  ## - `flagBuilder proc <#flagBuilder,Cli[T]>`_
+  assert b.flagName.len > 0, "Flag name must not be empty"
```
zerbina: Should raise a `ValueError` according to the documentation.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  if b.flagName in cli.command[command].flag:
+    raise newException(ValueError, "Flag '" & b.flagName & "' already exists")
+  for alias in b.aliases.items:
+    assert alias != "", "Flag alias cannot be empty"
+    if alias in cli.command[command].flag:
+      raise newException(ValueError, "Flag '" & alias & "' already exists")
+
+  assert not b.flagParser.isNil(), "Parser must be non-nil"
```
zerbina: Same here.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ## Adds the positional parameter specified by `b` to `command`.
+  ##
+  ## A positional parameter can only be added to `command` if:
+  ##
+  ## - `command` does not contain any subcommands.
+  ## - A non-empty `name` must be set and must not have been added to `command`
+  ##   previously.
+  ## - A `parser` must be set.
+  ## - If non-optional, must be added prior to any optional parameters.
+  ## - If catch all, must be added after all other parameters.
```
zerbina:
```suggestion
  ## - `command` does not contain any subcommands.
  ## - The parameter has a non-empty name.
  ## - The name is not used by any previously added positional parameters.
  ## - A parser was set for the parameter.
  ## - (only when non-optional) No optional parameters have been added yet.
  ## - (only when not catch-all) No catch-all parameter have been added yet.
```
So that all items read as predicates.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## `ValueError` will be raised if any of the specified constraints are
+  ## violated.
+  ##
+  ## See also:
+  ## - `positionalBuilder proc <#positionalBuilder,Cli[T]>`_
+  assert b.posName.len > 0, "Positional name should not be empty"
```
zerbina: Should raise a `ValueError` according to the documentation.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  for param in cli.command[command].positional.items:
+    if b.posName == cli.name[param]:
+      raise newException(ValueError):
+        "Positional with name '" & cli.name[param] & "' already exists"
+
+  assert not b.posParser.isNil(), "Parser must be non-nil"
```
zerbina: Should raise a `ValueError` according to the documentation.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+      .name("act")
+      .addTo(cli, RootCommand)
+    doAssert cli.nameOf(actCmd) == "act"
+
+  cli.name[ParameterId command]
+
+iterator namesOf*(cli: Cli, flag: FlagId): lent string =
+  ## Yields names that can be used to refer to `flag`.
+  ##
+  ## The first name yielded is always the canonical name of `flag`.
```
zerbina:
```suggestion
  ## Returns all names that can be used to refer to `flag`. The canonical name
  ## is always returned first.
```
A soft suggestion.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    yield cli.name[ParameterId flag]
+    for name in cli.alias[ParameterId flag].items:
+      yield name
+  except KeyError:
+    discard "Flag has no aliases"
+
+iterator namesOf*(cli: Cli, command: CommandId): lent string =
+  ## Yields names that can be used to refer to `command`.
+  ##
+  ## The first name yielded is always the canonical name of `command`.
```
zerbina:
```suggestion
  ## Returns all names that can be used to refer to `command`. The canonical
  ## name is always returned first.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+      .addTo(cli)
+
+    doAssert cli.classify(ParameterId RootCommand) == ParameterKind.Command
+    doAssert cli.classify(ParameterId strFlag) == ParameterKind.Flag
+    doAssert cli.classify(ParameterId strPos) == ParameterKind.Positional
+
+  case cli.parser[param].kind
+  of ParserKind.Command: ParameterKind.Command
+  of ParserKind.Flag, FlagOptionalValue: ParameterKind.Flag
+  of ParserKind.Positional..OptionalCatchAll: ParameterKind.Positional
```
zerbina:
```suggestion
  of ParserKind.Command:
    ParameterKind.Command
  of ParserKind.Flag, FlagOptionalValue:
    ParameterKind.Flag
  of ParserKind.Positional..OptionalCatchAll:
    ParameterKind.Positional
```
Alternatively, you could also align the values horizontally.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+
+  cli.parentOf(command)
+    .flatMap(
+      proc (parent: CommandId): Option[bool] =
+        cli.defaultCommandOf(parent)
+          .map(
+            proc(x: CommandId): bool = x == command
+          )
+    )
+    .get(otherwise = false)
```
zerbina:
```suggestion
  let parent = cli.parentOf(command)
  if parent.isSome:
    let default = cli.defaultCommandOf(parent.get)
    default.isSome and default.get == command
  else:
    false
```
It's shorter, easier to read (in my opinion), and a lot more efficient, as there's no copy of `cli`.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+
+func isValueOptional*(cli: Cli, flag: FlagId): bool =
+  ## Returns whether a value must be specified on the command line for `flag`.
```
zerbina:
```suggestion
  ## Returns whether a value for `flag` is optional on the command line.
```
The comment is inverted.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+
+    doAssert cli.isValueOptional(strFlag)
+    doAssert not cli.isValueOptional(strReqFlag)
+    doAssert cli.isValueOptional(switchFlag)
+
+  cli.parser[ParameterId flag].kind == FlagOptionalValue
+
+func isOptional*(cli: Cli, positional: PositionalId): bool =
+  ## Returns whether a value must be specified on the command line for
+  ## `positional`.
```
zerbina:
```suggestion
  ## Returns whether a value for `positional` is optional on the command line.
```
The comment is inverted.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+      .parser((v, var s) => Action.Continue)
+      .addTo(cli)
+
+    doAssert not cli.isCatchAll(reqPos)
+    doAssert cli.isCatchAll(anyPos)
+
+  cli.parser[ParameterId positional].kind in {CatchAll, OptionalCatchAll}
+
+proc helpFlagBuilder*[T](
+  cli: var Cli[T],
```
zerbina:
```suggestion
  cli: Cli[T],
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ## - `addHelpFlag proc <#addHelpFlag,Cli[T],CommandId,sinkstring,varargs[string]>`_
+  cli.flagBuilder
+    .name(name)
+    .optionalParser(
+      proc (_: string, v: Option[string], r: var T): Action = ShowHelp
+    )
+    .describe("display help message")
+
+proc addHelpFlag*[T](
+  cli: var Cli[T],
```
zerbina, line 1928:
```suggestion
  cli: Cli[T],
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  let (kind, option) =
+    if not ctx.isValueOnly:
+      ctx.lexer.next()
+    else:
+      ctx.lexer
+        .value()
+        .map(
+          proc (value: string): (CmdlineKind, string) =
+            (cmdValue, value)
+        )
```
zerbina:
```suggestion
        .map(proc (value: auto): auto = (cmdValue, value))
```
A soft suggestion.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ## An initial value for the internal accumulator can be specified using
+  ## `defaults`.
+  ##
+  ## See also:
+  ## - `run proc <#run,Cli[T],sinkseq[string],File,sinkT>`_
+  result = defaults
+  parse(cli, result, args)
+
+iterator flags*(cli: Cli, command: CommandId): FlagId =
+  ## Iterates through all added flags for `command`.
```
zerbina:
```suggestion
  ## Returns all flags for `command`.
```
A soft suggestion.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  # Not the most efficient, but CLIs shouldn't be big enough for this to be an
+  # issue
+  for i in 0 ..< cli.parser.nextId.int:
+    if cli.parent[ParameterId i] == ParameterId(command) and
+      cli.parser[ParameterId i].kind in {ParserKind.Flag, FlagOptionalValue}:
+      yield FlagId(i)
+
+iterator positionals*(cli: Cli, command: CommandId): PositionalId =
+  ## Iterates through all added positional parameters for `command`.
```
zerbina:
```suggestion
  ## Returns all positional parameters for `command`.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  if not cli.command[command].isDispatcher:
+    for i in cli.command[command].positional.items:
+      yield PositionalId(i)
+
+iterator subcommands*(cli: Cli, command: CommandId): CommandId =
+  ## Iterates through all added subcommands for `command`.
```
zerbina:
```suggestion
  ## Returns all subcommands for `command`.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  for (param, parent) in cli.parent.pairs:
+    if parent == ParameterId(command) and
+      cli.parser[param].kind == ParserKind.Command:
+      yield CommandId(param)
+
+func flagsUsage*(cli: Cli, command: CommandId): string =
+  ## Produces an usage message for all flags registered for `command`.
```
zerbina:
```suggestion
  ## Produces a usage message for all flags registered for `command`.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  ##
+  ## For each flag, the rendered message contains the first short and long
+  ## name, a placeholder and the described `usage`.
+  ##
+  ## For flags with required values, the placeholder is rendered as `VALUE`
+  ## if not set.
+  ##
+  ## A placeholder is not shown for flags with optional values if not set.
```
zerbina:
```suggestion
```
I'd leave it out, as `describe` already describes these cases.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  else:
+    result.add '<'
+    result.add cli.nameOf(positional)
+    result.add '>'
+
+  if cli.isCatchAll(positional):
+    result.add "..."
+
+func positionalsUsage*(cli: Cli, command: CommandId): string =
+  ## Produces an usage message for all positionals registered for `command`.
```
zerbina:
```suggestion
  ## Produces a usage message for all positionals registered for `command`.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+      display.alignLeft:
+        if usage.len > 0: posPad else: 0
+    result.add usage
+
+func commandUsage*(
+  cli: Cli,
+  command: CommandId,
+  rootName: string = cli.nameOf(RootCommand)
+): string =
+  ## Produces an usage message for the given command. The message is meant to
```
zerbina:
```suggestion
  ## Produces a usage message for the given command. The message is meant to
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+  elif cli.isDispatcher(command):
+    result.add " <COMMAND>"
+
+  for positional in cli.positionals(command):
+    if result.len > 0:
+      result.add ' '
+    result.add cli.displayOf(positional)
+
+func subcommandsUsage*(cli: Cli, command: CommandId): string =
+  ## Produces an usage message for all subcommands registered for `command`.
```
zerbina:
```suggestion
  ## Produces a usage message for all subcommands registered for `command`.
```

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2819 @@
...
+    result.add "\n\nArguments:\n"
+    result.add args
+
+  if flags.len > 0:
+    result.add "\n\nOptions:\n"
+    result.add flags
+
+func prettifyError*[T](cli: Cli[T], error: ref ParseError): string =
+  ## Produce a pretty error message for `error`. The provided `error` must be
+  ## produced by `parse(cli)`.
```
zerbina:
```suggestion
  ## Produces a pretty error message for `error`. `error` must have been
  ## raised by `parse(cli)`.
```

zerbina (commented):
It's not a full review, but covers the most important parts, I think.

My main focus was on documentation and builder internals. I've only skimmed over the runner implementation, the tests, and the runnable examples.

I like the public API, though the longer I looked at it, the more I thought about whether an API that statically enforces constraints (e.g., via the builder creation procedure having required parameters for names, etc.) wouldn't be better (not implying that anything needs to change in this regard).

The code I've found fairly hard to read, largely due to the runnable example, but also because of the code style and the duplicated documentation. The workarounds for making tailcalls work also don't help.

There's also a bug that I've found:
```nim
var cli = commandBuilder(int).initCli()
cli.flagBuilder
  .name("b")
  .parser(
    proc(option: auto, value: auto, x: var int): auto =
      discard)
  .addTo(cli)

var i = 0
cli.run(i, @["-ab"])
# crashes due to an `unreachable` in `parse` being reached
```

My comments and suggestions are largely regarding comments, but I've also left some regarding naming and the implementation of builders.

lib/experimental/cmdline.nim
alaviss: > Wouldn't a separate procedure be used instead of dispatching on the `option` within the procedure?

It allows for implementing things like `--no-foo` vs `--foo` by aliasing on `--foo`.

It's generally useful to expose input values whenever possible IMO, there's always an usage for someone and it's not hidden internal information.

lib/experimental/cmdline.nim
alaviss: While `parseCli` would be doing parsing here, it's not unlikely that the user might perform additional verification based on the collected state so far, which is a form of parsing to me :grin:

lib/experimental/cmdline.nim
alaviss: It's not possible since we need reflection information for `isValueOptional`. And also the parsing routine is ill-equipped to handle `ParseError` generated by user-provided parsers.

The other motivation was to avoid closures, but that ship sailed the moment I allow parsers to not return `Action`.

lib/experimental/cmdline.nim
zerbina: The information isn't really hidden, as one explicitly provides the names/aliases the parsers acts on, but I do get your point.

lib/experimental/cmdline.nim
alaviss: The accumulator also acts as an input that could inform whether a command is legal at a given position. At least that's the excuse I would give :laughing:

The real reason is that I centered a lot of the vocab used around `parser()`, which makes for a consistent-ish API.

lib/experimental/cmdline.nim
alaviss: Yea that's outdated docs back when only flags are supported. And apparently it also got caught by my substitution filter for the `Id` suffix the other day.

lib/experimental/cmdline.nim
zerbina: > It's not possible since we need reflection information for `isValueOptional`.

Hm, why? `FlagOptionalValue` wouldn't go away -- there would just be no `FlagParser[T, Action]` field in `ParserAny`.

lib/experimental/cmdline.nim
alaviss: The default value of `parser` would use `ParserKind.Command`. Handling them here removes the ordering constraint between modifiers and parsers.

lib/experimental/cmdline.nim
alaviss: I suppose I should rephrase this, I'd welcome suggestions if you have any.

What is meant here is that the root command doesn't have to be named (since the name is used for docgen only), while subcommands must have a name.

EDIT: to answer the question directly: A name on the root command doesn't participate in parsing, so nothing would happen if one is set, hence its optionality.

lib/experimental/cmdline.nim
alaviss: Oh, right, my bad. But the other reason stands, since error handling for no value case would have to be done outside of the actual parser, as the facilities for handling parser-raised `ParseError` doesn't exist and it would force exposing certain internal states.

I could invent a new "internal" exception for this purpose, but that wouldn't change anything other than where the handling code is, and the current method prevents unnecessary code duplication.

An alternative to just unwrap unconditionally after outside checks, but I'd prefer a different prototype to avoid internal bugs.

lib/experimental/cmdline.nim
alaviss: I'm pretty sure this one has to be `var` :grin:

zerbina: Oh, right, my bad.

alaviss:
> I like the public API, though the longer I looked at it, the more I thought about whether an API that statically enforces constraints (e.g., via the builder creation procedure having required parameters for names, etc.) wouldn't be better (not implying that anything needs to change in this regard).

Thanks.

The idea about builder creation statically enforcing the requirements did cross my mind, but there were a couple issues leading to them not being adopted:

- It prevents certain composing patterns: `helpFlagBuilder` couldn't just yield a builder with only the parser (it doesn't do this, but other people might).
- Some requirements can only be ergonomically set via an external procedure: a key example is a flag's parser, with its variants. It's certainly possible to have the parser be inline (i.e. `flagBuilder` vs `optionalFlagBuilder`), but it'd make overriding a parser looks weird (imagine `optionalFlagBuilder(...).parser()`).

There's a small runtime cost to offloading these checks there, but in general I think the API is a lot easier to use and compose like this.

lib/experimental/cmdline.nim
alaviss: This phrasing wouldn't work well, since the parsers are also called without a value (i.e. `--opt-int` calls `opt-int`'s parser with a `none(int)`).

I think "receive" works here, given that there are supplementary examples below.

lib/experimental/cmdline.nim
saem: Dropping the dependency upon `commandLIneParams` is a must imo, and `when`, even though its a compromise, at least puts us in the right direction.

This also let's us make sure we can include this in stress testing the js backend.

lib/experimental/cmdline.nim
saem: > > I also think the fields here should use plural (e.g., `commands` instead of `command`).
> 
> I think they are better in singular form, the reason being while they technically store multiples of X, the usage of them is usually a direct binding with an ID, so I prefer to name them as if I'm dealing with a single object (and if we ever have a SoA feature this comes in handy).

This isn't in keeping in the style with how we've been writing things and I'm not seeing a good reason to deviate, tbh. If it's many, it should be plural. Reading through the code it is weird that we have to continually provide a `CommandId` to `cli.command`.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2856 @@
...
+    Continue ## Continue parameter parsing.
+    ShowHelp ## Abort and show help message.
+    DisableFlagProcessing ## Parameters following this will no longer be
+                          ## interpreted as flags.
+
+  MaybeAction* = Action | void
+    ## Typeclass to support parsers returning either `Action` or nothing.
+
+  FlagParser*[T; R: MaybeAction] = proc (option, value: string, accumulator: var T): R
+    ## A parser for flag with name `option` and value `value`. The accumulator
```
saem: I think `name` is a much better name for the parameter currently called `option`. 😆

lib/experimental/cmdline.nim
saem: Reading through it some more, I think `extractor` is a better term, in that it's more precise, in describing what's going on based on the API + examples/tests I've read thus far, it's extracting a value.

N.B.: for the builder api, `extract` reads better, e.g.: `.extract(noop)`.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2856 @@
...
+
+proc hash(x: ParameterId): Hash {.borrow.}
+
+proc `==`*(a, b: ParameterId): bool {.borrow.}
+proc `==`*(a, b: CommandId): bool {.borrow.}
+proc `==`*(a, b: FlagId): bool {.borrow.}
+proc `==`*(a, b: PositionalId): bool {.borrow.}
+
+# FIXME: Put this in system.nim
+func drop[T](_: sink T) = discard
```
saem, line 556: Any reason to not break this out into a PR now?

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2856 @@
...
+      .name("string2")
+      .describe("a string")
+      .parser((_, v, var s) => (s = v))
+      .addTo(cli)
+    doAssert cli.placeholderOf(strFlag) == "STR"
+    doAssert cli.placeholderOf(str2Flag) == ""
+
+  cli.placeholder[ParameterId flag]
+
+func isDispatcher*(cli: Cli, command: CommandId): bool =
```
saem: An alias of `hasSubcommands` seems appropriate, the dispatcher terminology is mentioned elsewhere, but I don't think it'd come to mind.

lib/experimental/cmdline.nim
zerbina: A perhaps cleaner solution compared to using `when` is moving `run` into its own module, where it can freely depend upon `os`, with a possible candidate for the name being `cmdlinerunner`.

In my opinion, `run` is part of the "porcelain layer", and thus doesn't really belong with the rest of `cmdline`. Having it separate also makes sure that the implementation doesn't start accidentally rely on private `cmdline` APIs.

lib/experimental/cmdline.nim
zerbina: I'm not sure whether the regularity is something desired here, as the operation is conceptually (a little) different from the others, but the name is something that can still be easily adjusted later (if at all) -- the module is experimental, after all.

lib/experimental/cmdline.nim
zerbina: I'm not entirely following, sorry. As far as I can tell, `optParser` and `parser` are only called in `parseFlag`, and therefore having only `optParser` would only require changing changing the `parser.parser(option, ...)` call to take the option directly (plus creating a wrapper in `FlagBuilder.parser`).

In any case, looking at the code again, I don't think this would result in a noticeable improvement.

zerbina:
> There's a small runtime cost to offloading these checks there, but in general I think the API is a lot easier to use and compose like this.

Yea, not being as composable as it is now is also something that came to my mind. My main issue with the dynamicness is less the runtime cost, but more that errors are only reported at run-time. There's also the increase in surface for implementation errors (due to the require error checking), but that's something the tests help with.

However, like I said, I think the builder-based API is fine as is.

lib/experimental/cmdline.nim
alaviss: Without some magic in the compiler the `sink` might turn into a hidden copy if not careful.

I don't want to do the compiler part atm.

saem: hmm, sounds like we'll need a `consume` like feature, rather than `sink`

lib/experimental/cmdline.nim
alaviss: I think I might just phase the dispatcher terminology out.

lib/experimental/cmdline.nim
alaviss: I think for API consistency I'll stick with `Parser` here. It's easier to convey these as "parser" and "typed parser", as they both put data into the accumulator. As a bonus, APIs that can take both will clearly signal that they are exclusive and will override the other (i.e. `flagBuilder().parser().parser(bool)` vs. `flagBuilder().parser().extractor()`).

tests/stdlib/experimental/tcmdline.nim
```diff
@@ -0,0 +1,2191 @@
+discard """
+description: "Tests for the cmdline module"
+targets: "c js vm"
+knownIssue.vm: "tables crashes with 'trying to access location outside of VM's memory'"
```
zerbina:
```suggestion
  description: "Tests for the cmdline module"
  targets: "c js vm"
  knownIssue.vm: '''
    `.tailcall` in combination with `sink` is not supported due to `vmgen` bug,
    causing an error at run-time 
  '''
```
This make the known-issue description more helpful to someone wanting to look into the issue.

lib/experimental/cmdline.nim
```diff
@@ -0,0 +1,2859 @@
...
+    let error = (ref CommandError)(error)
+    "could not parse subcommand '" & error.commandName & "': " & error.msg
+  elif error of HelpError:
+    "help requested"
+  else:
+    "unexpected error parsing command line: " & error.msg
+
+# FIXME: Change the detection to stdmsg once
+# https://github.com/nim-works/nimskull/pull/1547 is merged
+when declared(os.commandLineParams) and declared(stderr):
```
zerbina: Why not move these procedures into a separate module?

alaviss: The UX is horrible, since users would have to go import an another module to actually use cmdline (for typical usage).

If I use `export`, then the exported module is "invisible" in the generated docs, not to mention that it wouldn't simplify any of these conditionals.

alaviss:
Rebased to import #1547

lib/experimental/cmdline.nim
zerbina: > If I use `export`, then the exported module is "invisible" in the generated docs

What do you mean by invisible?

> not to mention that it wouldn't simplify any of these conditionals.

Why? There'd be no need for any conditional when `run` is in a separate module. If one imports the separate module with the VM/JS backend, compilation would simply fail.

alaviss: > What do you mean be invisible?

The docs won't show up. It's a tradeoff I'd take for something like `cmdline/parsers`, since they're mostly helpers that you wouldn't normally interact with, not for something crucial to the typical usage flow.

> Why? There'd be no need for any conditional when `run` is in a separate module. If one imports the separate module with the VM/JS backend, compilation would simply fail.

I consider requiring to import two modules a deal breaker for cmdline, where the typical flow ends with user calling `run`.

lib/experimental/cmdline.nim
saem: I personally don't see putting data into an accumulator as a key characteristic of a parser, the first, and predominant, thing that comes to mind is screwing around with character/token streams.

alaviss: The point is that I'd prefer to use only one term to refer to all of these. I think `parser` fits this role best, but I could be convinced otherwise.

lib/experimental/cmdline.nim
zerbina: > I consider requiring to import two modules a deal breaker for cmdline, where the typical flow ends with user calling `run`.

You wouldn't need to import two modules, just the one providing `run`, as that also exports `cmdline`. The routines not being listed directly on the separate module's documentation page can be easily helped with by linking to `cmdline` at the top, in my opinion.

I'd prefer there being no `when` workarounds here, given the current state of multi-platform/backend support, but in the end, I don't consider the presence of `when` a blocker for this PR.

zerbina (approved):
From my side, the changes look good. Nice work!

The test coverage (from both the tests and runnable examples) looks great, though I've had trouble following the code due to repetitive shape and sheer volume, thus making it likely that I've missed something in this area.

In the future, please use merge instead of rebase, at least when there are review comments already -- the rebase makes it hard to keep track of what was reviewed already.

I'm not merging the PR yet, so that @saem can take another look.

lib/experimental/cmdline.nim
saem: I'm ok to give it a shot and see what it feels like when folks, including myself, use it and then come back to this, it's not like we guarantee stability yet, doubly so for anything experimental.

tests/stdlib/experimental/tcmdline.nim
```diff
@@ -0,0 +1,2194 @@
+discard """
+  description: "Tests for the cmdline module"
+  targets: "c js vm"
```
saem, line 3: noice!

saem (approved):

saem:
merge

nice job, @alaviss!

saem:
/merge

nice job, @alaviss!

PR #1498, opened 2025-03-19
lang: add the `.tailcall` calling convention

zerbina:
## Summary

Add the `.tailcall` calling convention to the language. It designates
routines as being guaranteed to take over their caller's stack frame,
but only when the caller also uses the `.tailcall` calling convention.

### Design

Guaranteed tail-call elimination is an important requirement for
programs written in functional style, continuation-passing styles, or
any other style that requires deep, potentially unbounded recursion.

Implementing it efficiently via a macro is hardly possible, if at all,
making it a good candidate for being a built-in language feature.

Only enforcing tail calls for `.tailcall` calls within `.tailcall`
routines, removes the need for adapter routines at the non-`.tailcall`
to `.tailcall` edge, but it also makes the static semantics of
`.tailcall` calls callsite-dependent.

### Implementation

A description of how the portable tail-call elimination works is
provided in the `tailcallelim.rst`.

The implementation is scattered throughout multiple stages:
* `sem` and `sempass2` take care of the analysis part and synthesizing/
  instantiating the needed types and routines
* `transf` and `mirgen` perform the portable tail call elimination
* `injecthooks` handles the "trailing destructor call" detection and
  related reporting
* the code generators implement storing the parameter tuple into the
  type-erased storage (the new `mStoreParams` magic)

Whether `mirgen` performs the portable tail call elimination is
configured through the `goTailCallElim` option. When not present,
must-become-sibling-call calls are translated to the new `mnkTailCall`
MIR operation.

The VM backend uses true sibling calls, as the `mStoreParams` magic
cannot be implemented by the VM due to its strict typing requirements.


alaviss:
Is it feasible to add a helper/pragma to allow calling `musttail` procs outside of tail positions (i.e. by not forcing a transform of the caller)?

It's trivial enough to write a quick proc wrapper around such calls, but it'd be better to just tell the compiler not to transform the caller.

zerbina:
> Is it feasible to add a helper/pragma to allow calling `musttail` procs outside of tail positions (i.e. by not forcing a transform of the caller)?

An expression pragma forcing the `.musttail` call to be trampolined in-place is doable with a reasonable amount of effort, yes.

compiler/sem/semtypes.nim
```diff
@@ -1928,6 +1945,10 @@ proc semProcTypeWithScope(c: PContext, n: PNode,
     # we're still interested in implicit tags and raises pragmas
     n[1] = implicitPragmas(c, s, n[1], {wTags, wRaises})
 
+  # instantiat the type of the continuation for .musttail procedures
```
saem:
```suggestion
  # instantiate the type of the continuation for .musttail procedures
```

doc/manual.rst
```diff
@@ -4101,6 +4107,59 @@ Overloading of the subscript operator
...
+Tail Call
+---------
+
+A tail call is a routine call that is the very last operation taking place in
+a routine's body.
+
+A call is considered a tail call iff:
+* it is the *tailing expression* of a `return`, and the `return` is a
+  *tailing return*
+* it is the *tailing expression* of a rotine's body
```
```suggestion
* it is the *tailing expression* of a routine's body
```

doc/manual.rst
```diff
@@ -4101,6 +4107,59 @@ Overloading of the subscript operator
...
+* in a `try` body
+* in a statement list with implicit destructor calls at its end
+* after a `defer` statement in a statement list
+* in an `except` clause of a `try` with a `finally` clause
+
+`.musttail` Call
+----------------
+
+Since `.musttail` routine calls are always they take over their caller's stack
+frame, some restriction are placed on their arguments:
```
```suggestion
Since `.musttail` routine calls always take over their caller's stack frame,
some restrictions are placed on their arguments:
```

doc/manual.rst
```diff
@@ -4101,6 +4107,59 @@ Overloading of the subscript operator
...
+* arguments to `sink` parameters can be arbitrary expressions, as long as
+  ownership of the value can be transferred
+* arguments to `openArray` parameters must be views into caller parameters or
+  globals
+* arguments to `var`, pass-by-reference, or pass-by-value-with-overridden-copy-
+  operator type parameters must be lvalue expressions derived from caller
+  parameters or globals
+* pass-by-value parameters without for whose type `supportsCopyMem` returns
+  `true`` act like `sink` parameters
```
```suggestion
* pass-by-value parameters with a type for which `supportsCopyMem` returns
  `true` act like `sink` parameters
```

saem (commented):
partial review, I still have to go through the tests properly.

---

Current Thoughts

I like the change, my only suggestion is that the calling convention should perhaps be called `tailcall`, my reasoning is as follows:
* the calling conventions are pretty much all things one must do/adhere to anyways, so that's redundant
* `tailcall` is easier to search both in the doc (fewer ambiguities) and on the internet, for the general concept
* `musttail` is a bit overloaded term with people who have a gcc/llvm background making all sorts of misplaced assumptions

compiler/sem/tailcall_elim.nim
```diff
@@ -0,0 +1,525 @@
...
+    emLast
+
+proc error(config: ConfigRef, info: TLineInfo, msg: string) =
+  # TODO: use dedicated reports
+  config.localReport(info, reportStr(rsemUserError, msg))
+
+# -------------- analysis --------------
+
+proc checkArg(g: ModuleGraph, owner: PSym, n: PNode, i: int, formal: PType) =
+  ## Analyses the `i`-th argument `n` for an call to a procedure with type
```
saem:
```suggestion
  ## Analyses the `i`-th argument `n` for a call to a procedure with type
```

compiler/sem/tailcall_elim.nim
```diff
@@ -0,0 +1,525 @@
...
+    else:
+      discard "acts like a sink parameter"
+
+proc supportsSiblingCalls(s: PSym): bool =
+  # sibling-calling (necessary for tail-call elimination) requires modifying
+  # a caller's signature, therefore, as a rule of thumb: every routine whose
+  # signature cannot easily and transparently be changed to something else
+  # cannot use tail calls
+  if s.typ == nil:
+    echo s
```
saem: Leftover debug code?

compiler/sem/tailcall_elim.nim
```diff
@@ -0,0 +1,525 @@
...
+
+  # note: the `emExpr` mode only needs to be activated only on the
+  # statement -> expression edge
+  case n.kind
+  of nkReturnStmt:
+    if n[0].kind == nkAsgn:
+      recurse(n[0][1], mode + {emLast} - {emExpr})
+  of nkCallKinds:
+    if n[0].kind == nkSym and n[0].sym.magic == mRunnableExamples:
+      # the body was not type properly, nor is it relevant for the
```
saem:
```suggestion
      # the body was not typed properly, nor is it relevant for the
```

saem (approved):
Went through the rest of the code as is, I didn't go quite as a deep as I know there is a direction change to remove the caller transform requirement, along with associated restrictions (around iterator, exportc, methods, etc), coming.

I can do another pass after that, if you'd like.

compiler/sem/tailcall_elim.nim
zerbina: Yep, thanks, I'd have likely missed that.

saem (approved):

compiler/backend/ccgtypes.nim
```diff
@@ -342,8 +342,8 @@ proc prepareParameters(m: BModule, t: PType): seq[TLoc] =
 proc prepareParameters(m: BModule, prc: PSym): seq[TLoc] =
   result = prepareParameters(m, prc.typ)
   if prc.typ.callConv == ccTailcall:
-    # add a valid enough loc that the header can be generated properly. The
-    # real loc is filled in later
+    # add a param loc that's valid enough for the header can be generated
```
saem:
```suggestion
    # add a param loc that's valid enough for the header to be generated
```

Clyybber (approved):

compiler/backend/cgirgen.nim
```diff
@@ -388,7 +388,11 @@ proc callToIr(tree: MirBody, cl: var TranslateCl, n: MirNode,
               cr: var TreeCursor): CgNode =
   ## Translate a valid call-like tree to the CG IR.
   let info = cr.info
-  result = newExpr((if n.kind == mnkCall: cnkCall else: cnkCheckedCall),
+  result = newExpr((case n.kind
+                    of mnkCall:        cnkCall
+                    of mnkCheckedCall: cnkCheckedCall
+                    of mnkTailCall:    cnkTailCall
+                    else: unreachable()),
                    info, cl.map(n.typ))
```
alaviss: Purely cosmetic, but I think it's worth moving this `case` into it's own `let callKind`.

compiler/backend/jsbackend.nim
```diff
@@ -122,7 +123,8 @@ proc generateCode*(graph: ModuleGraph, mlist: sink ModuleList) =
   ## writes it to the output file.
   let
     globals = newGlobals(graph)
-    bconf = BackendConfig(tconfig: TranslationConfig(magicsToKeep: NonMagics))
+    bconf = BackendConfig(tconfig: TranslationConfig(magicsToKeep: NonMagics,
+                                                     options: {goTailCallElim}))
```
alaviss: Once the inner expression requires breaking lines, I believe the outer should be broken down as well. It keeps the door open for adding things in the future.

```suggestion
    bconf = BackendConfig(
      tconfig: TranslationConfig(
        magicsToKeep: NonMagics,
        options: {goTailCallElim}
      )
    )
```

compiler/front/cli_reporter.nim
```diff
@@ -496,6 +496,14 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       result = ("accessed location '$1' doesn't exist in the current " &
                "compile-time context") % [r.ast.render]
 
+    of rsemNoTailingExpression:
+      result = "call to .tailcall routine is not the tailing expression of " &
+               "a return"
+
+    of rsemArgumentMustBorrowFromParameter:
+      result = "argument doesn't borrow from parameter, global, or ptr " &
```
alaviss:
```suggestion
      result = "argument doesn't borrow from parameter, global, or pointer " &
```

I'd say to keep short forms out of diagnostics.

compiler/front/cli_reporter.nim
```diff
@@ -1364,6 +1372,30 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
       result = "an exception can only be re-raised within the scope of an" &
                " except, with no finally in-between"
 
+    of rsemCleanupPreventsTailCall:
+      result = "'$1' requires cleanup that prevents a tail call" %
+               [r.ast.render]
```
alaviss, line 1377: Nothing actionable now, but this is kinda hard to grasp for people without a firm understanding of the language.

Maybe we should create diagnostic explainer manual pages like Rust in the future.

lib/system.nim
```diff
@@ -3064,3 +3064,17 @@ when defined(nimDebugUtils):
...
+  ParamBlob = object
+    ## The type to use for the storage of parameters.
+    data: array[256, byte]
+
+  Continuation[T] = object
+    ## Internal type only meant to be used by the compiling compiler. Needed by
+    ## tail-call elimination.
+    case done: bool
+    of false:
+      next: proc(env: pointer): Continuation[T] {.nimcall.}
```
alaviss, line 3078: Can we enforce that `next` cannot be `nil` and use it as the discriminator instead of `done`? That'd prevent us from wasting 7 bytes of alignment for `done`.

This doesn't have to be solved now fwiw.

tests/lang/s99_tailcall/t02_params_by_ref.nim
alaviss, line 1: Quick question: should there be tests for `{.byref.}` and `{.bycopy.}` pragmas as well?

alaviss (approved):
Only minor styling points and some questions, none of which are blocking.

Looks great otherwise to me.

alaviss:
Minor but non-blocking issue with diagnostics:

```nim
import macros

proc fib(c: int, a = 0, b = 1): int {.tailcall.} =
  case c
  of 0: a
  of 1: b
  else: fib(c - 1, b, a + b)

const BigFib = fib(50)

proc main() =
  echo fib()

main()
```

yields:

```
test.nim(12, 11) Error: type mismatch: got <>
but expected one of:
proc fib[type node](c: int; a = 0; b = 1; :env): int

expression: fib()
```

The internal implementation leaked into errors.

tests/lang/s99_tailcall/t02_params_by_ref.nim
saem: `byref` and `bycopy` are questionable features, so not bothering with proliferating testing for them seems reasonable.

compiler/front/cli_reporter.nim
zerbina: I agree, but the usage of `ptr` here was to express that this only applies to `ptr T` parameters, and not `pointer` parameters.

Having said that, borrowing from a `pointer` projection is not possible anyway (you'd have to cast it to `ptr` first), so I think using `pointer` here is fine.

lib/system.nim
zerbina: Sure, that'd also shorten the construction logic.

zerbina: Hm, on second thought, this wouldn't provide much of a benefit. Only `Continuation[void]` is reduced in size, and for other cases (such as when `T` is a type that cannot be zero-initialized), it would actually hurt efficiency instead, since `result` is then always default-initialized, regardless of whether `next` is `nil` or not.

zerbina:
> Minor but non-blocking issue with diagnostics:

This is hard to fix properly, as `transf` currently needs to modify symbols in-place (`.closure`s are affected by the same problem), but it can be worked/hacked around with by modifying `renderer`.

**Edit:** I've implemented the workaround.

lib/system.nim
alaviss: > since `result` is then always default-initialized, regardless of whether `next` is `nil` or not.

Why is a `bool` discriminator different for this case?

zerbina: Because it's a real discriminator. If you use:
```nim
type Continuation[T] = object
  next: ...
  result: T
```
then `next` and `result` are just two adjacent fields with no further relationship between them.

A `proc` value cannot be used as a discriminator, not even internally.

zerbina:
/merge

Thanks for the reviews, everyone!

PR #1501, opened 2025-03-26
fix(ci): always raise error after printing error logs

alaviss:
## Summary
Fixes a regression from #1475 where failing test runs are still marked
as passing.

Issue was reported by @zerbina on Matrix.

## Details
Due to some bash shenanigans,

    if ! ./koch tests; then
      exitCode=$? # yields 0!
      ...
      exit $exitCode # ends up being "successful"
    fi

while there are ways to workaround this, it doesn't really matter since
GitHub Actions is only looking for a non-zero result, so hard-code 
`exit 1`  instead to force failing the step.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -171,12 +171,11 @@ jobs:
       - name: Run tester
         run: |
           if ! ./koch.py test --batch:"$TEST_BATCH" --tryFailing all; then
-            oldExitCode=$?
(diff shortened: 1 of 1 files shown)


zerbina:
The failing test on Windows is known to be flaky. Retrying.

alaviss:
One of the few times where flaky tests verify that your fix works

zerbina:
It looks like you're hitting all the flaky tests today.

PR #1503, opened 2025-03-28
sem: wrap expression routine bodies in `return`

zerbina:
## Summary

Wrap expression routine bodies in `return` statements instead of
`result` assignments. This preserves the information that the
expression is returned directly.

## Details

The information that the expression is returned directly gets lost when
lowering it to a `result = body` statement, as those assignments cannot
be distinguished from the ones written directly by the programmer.

In addition, compared to the previous transform, the `return` is
applied to the last expression in the list (if the body is an
expression list), not the whole list.

Both `docgen` and `suggest` previously had to look for doc comments and
examples within trailing top-level assignment due to the whole body
expression being used as the assignment RHS -- these workarounds are no
longer necessary and thus removed.

### DIFF
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -2370,7 +2370,17 @@ proc semProcBody(c: PContext, n: PNode): PNode =
       # are not expressions:
       fixNilType(c, result)
     else:
-      result = semReturn(c, newTreeI(nkReturnStmt, n.info, result))
+      if result.kind == nkStmtListExpr:
+        # in order to preserve doc comments, apply the return to the last
+        # statement in the list, not to the whole list
+        result.transitionSonsKind(nkStmtList)
+        result.typ = nil
+        let last = semReturn(c, newTreeI(nkReturnStmt, n.info, result[^1]))
+        result[^1] = last
+        if last.kind == nkError:
+          result = c.config.wrapError(result)
+      else:
+        result = semReturn(c, newTreeI(nkReturnStmt, n.info, result))
   else:
     result = discardCheck(c, result, {})
 
--- a/compiler/sem/semexprs.nim
+++ b/compiler/sem/semexprs.nim
@@ -2370,10 +2370,7 @@ proc semProcBody(c: PContext, n: PNode): PNode =
       # are not expressions:
       fixNilType(c, result)
     else:
-      var a = newNodeI(nkAsgn, n.info, 2)
-      a[0] = newSymNode(c.p.resultSym)
-      a[1] = result
-      result = semAsgn(c, a)
+      result = semReturn(c, newTreeI(nkReturnStmt, n.info, result))
   else:
     result = discardCheck(c, result, {})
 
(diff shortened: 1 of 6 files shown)


compiler/sem/semexprs.nim
```diff
@@ -2370,10 +2370,17 @@ proc semProcBody(c: PContext, n: PNode): PNode =
...
+        # in order to preserve doc comments, apply the return to the last
+        # statement in the list, not to the whole list
+        result.transitionSonsKind(nkStmtList)
+        result.typ = nil
+        let last = semReturn(c, newTreeI(nkReturnStmt, n.info, result[^1]))
+        result[^1] = last
+        if last.kind == nkError:
+          result = c.config.wrapError(result)
+      else:
+        result = semReturn(c, newTreeI(nkReturnStmt, n.info, result))
```
alaviss: Shouldn't this turn into `elif` and `else` instead of nested?

compiler/sem/semexprs.nim
```diff
@@ -2370,10 +2370,17 @@ proc semProcBody(c: PContext, n: PNode): PNode =
...
-      var a = newNodeI(nkAsgn, n.info, 2)
-      a[0] = newSymNode(c.p.resultSym)
-      a[1] = result
-      result = semAsgn(c, a)
+      if result.kind == nkStmtListExpr:
+        # in order to preserve doc comments, apply the return to the last
+        # statement in the list, not to the whole list
+        result.transitionSonsKind(nkStmtList)
+        result.typ = nil
+        let last = semReturn(c, newTreeI(nkReturnStmt, n.info, result[^1]))
```
saem: I can't recall, but is it possible to end a statement list expression with a doc comment (and an expression being the last non-doc comment node)?

If that's the case then we'll need to search for the last expression, otherwise this code should be fine.

saem (approved):
I have one possible concern, but it might turn out to be bogus in which case this is good to merge, IMO

@alaviss' stylistic comment isn't bad, but not a hard change from my end

compiler/sem/semexprs.nim
alaviss: Doesn't seem to be the case with this example:

```nim
import macros

macro dumpTyped(n: typed) =
  echo treeRepr n

dumpTyped:
  proc foo: bool =
    true
    ## doc
```

yields:

```
test.nim(8, 5) Error: expression 'true' is of type 'bool' and has to be used (or discarded)
```

for both this branch and `devel`. Seems to me that `nkCommentStmt` participate in expression determination.

zerbina: > I can't recall, but is it possible to end a statement list expression with a doc comment (and an expression being the last non-doc comment node)?

It's allowed by the syntax, but not the semantics. `semStmtList` treats `nkCommentStmt` like real statements (except when reporting `unreachable code` warnings), and thus they can never be the last item in a valid `nkStmtListExpr`.

The `return` wrapping takes place *after* the body expression was typed, meaning that only valid `nkStmtListExpr` reach here.

PR #1504, opened 2025-03-31
fix memory corruption with indirect `.tailcall` calls  

zerbina:
## Summary

Fix a C/JS code generator bug leading to some `.tailcall` routines
receiving garbage values as their parameters, or failing to compile,
when called indirectly.

## Details

Both `cgen` and `jsgen` use the original `PType` to decide how
parameter passing works. For `.tailcall` proc types representing
procedure pointers, the proc type doesn't match the actual signature
(which is always `proc(env: pointer): Continuation[T]`), leading to the
environment pointer being passed incorrectly (e.g., by address).

This is worked around by having both `cgen` and `jsgen` manually emit
the argument setup for indirect `.tailcall` calls.

### DIFF
--- a/compiler/backend/ccgcalls.nim
+++ b/compiler/backend/ccgcalls.nim
@@ -242,6 +242,7 @@ proc genAsgnCall(p: BProc, le, ri: CgNode, d: var TLoc) =
        ri[0].kind != cnkProc:
     # an indirect tailcall call. The real signature doesn't match the `tyProc`,
     # so we manually emit the call
+    # XXX: this is fundamentally a hack see https://github.com/nim-works/nimskull/pull/1504
     assert numArgs(ri) == 1
     var callee, arg: TLoc
     initLocExpr(p, ri[0], callee)
--- a/compiler/backend/ccgcalls.nim
+++ b/compiler/backend/ccgcalls.nim
@@ -238,6 +238,15 @@ proc genClosureCall(p: BProc, le, ri: CgNode, d: var TLoc) =
 proc genAsgnCall(p: BProc, le, ri: CgNode, d: var TLoc) =
   if ri[0].typ.skipTypes({tyGenericInst, tyAlias, tySink}).callConv == ccClosure:
     genClosureCall(p, le, ri, d)
+  elif ri[0].typ.skipTypes(abstractInst).callConv == ccTailcall and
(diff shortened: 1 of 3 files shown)


zerbina:
The proper fix would be to:
* distinguish between signature and procedure pointer types at the MIR level
* only lower `.tailcall` procedure pointer types (not signature types)
* use MIR types instead of `PType` for the argument handling in the code generators

After all, being able to selectively lower or modify types (without the changes being visible to sem) was what the MIR type IR was intended for, among other things.

compiler/backend/jsgen.nim
```diff
@@ -1317,7 +1317,16 @@ proc genInfixCall(p: PProc, n: CgNode, r: var TCompRes) =
 
 proc genCall(p: PProc, n: CgNode, r: var TCompRes) =
   gen(p, n[0], r)
-  genArgs(p, n, r)
+  if n[0].kind != cnkProc and
+     n[0].typ.skipTypes(abstractInst).callConv == ccTailcall:
+    # an indirect tailcall call. The signature doesn't match reality, so we
+    # manually handle the argument
+    # XXX: this is fundamentally a hack
```
```suggestion
    # XXX: this is fundamentally a hack see https://github.com/nim-works/nimskull/pull/1504
```

compiler/backend/ccgcalls.nim
```diff
@@ -238,6 +238,15 @@ proc genClosureCall(p: BProc, le, ri: CgNode, d: var TLoc) =
 proc genAsgnCall(p: BProc, le, ri: CgNode, d: var TLoc) =
   if ri[0].typ.skipTypes({tyGenericInst, tyAlias, tySink}).callConv == ccClosure:
     genClosureCall(p, le, ri, d)
+  elif ri[0].typ.skipTypes(abstractInst).callConv == ccTailcall and
+       ri[0].kind != cnkProc:
+    # an indirect tailcall call. The real signature doesn't match the `tyProc`,
+    # so we manually emit the call
```
saem, line 244:
```suggestion
    # so we manually emit the call
    # XXX: this is fundamentally a hack see https://github.com/nim-works/nimskull/pull/1504
```

saem (approved):
Two minor suggestions so a future person knows where to look for fix ideas

PR #1505, opened 2025-03-31
improve the "cleanup prevents tail call" error message

zerbina:
## Summary

Improve the error message reported when a cleanup of some local
prevents a tail call from taking place, making it easier for the
user to fix the problem.

## Details

* the error now points to the call, not the local/temporary that
  prevents the tail call 
* when a named local prevents a tail call, the message provides the
  name
* when a temporary prevents a tail call, the message points to where
  the temporary comes from 

`mirgen` is updated to use more precise origin information, so that
the errors point to where one would expect.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1373,8 +1373,12 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
                " except, with no finally in-between"
 
     of rsemCleanupPreventsTailCall:
-      result = "'$1' requires cleanup that prevents a tail call" %
-               [r.ast.render]
+      if r.sym != nil:
+        result = "cannot tail call; local '$1' requires cleanup" %
+                 [r.symstr]
+      else:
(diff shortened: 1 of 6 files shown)


alaviss (approved):

saem (approved):
Yup, that's much better

zerbina:
@saem: It looks like there's some issue with installing `pcre` on Windows, affecting all new workflow runs.

saem:
> @saem: It looks like there's some issue with installing `pcre` on Windows, affecting all new workflow runs.

I assumed it was flakey and just ran it again, thanks for looking at it further. I'm hoping it's transient.

zerbina:
> I assumed it was flakey and just ran it again, thanks for looking at it further. I'm hoping it's transient.

Unfortunately, it doesn't seem to be. To summarize, the `pcre` build script requires CMake 2.8.5 compatibility, which was dropped from CMake 4.0.0 -- a CMake version that became the default for all new Windows 2022 runners two days ago.

PR #1521, opened 2025-04-14
fix `or`, `and`, and `|` type operators

### LINKED ISSUE #1516: `|` outside of type context doesn't return proper type
### Example

```nim
import macros

const Skippable = {ntyAlias, ntyTypeDesc}

proc getTypeSkip(n: NimNode, skip = Skippable): NimNode =
  result = getType(n)
  if result.typeKind in skip:
    case result.typeKind
    of ntyAlias:
      result = getTypeSkip(result, skip)
    of ntyTypeDesc:
      result = getTypeSkip(result[1], skip)
    else:
      discard "return as is"

proc getTypeImplSkip(n: NimNode, skip = Skippable): NimNode =
  result = getTypeImpl:
    getTypeSkip(n, skip)

macro dumpT(T: typedesc): untyped =
  echo "== T"
  echo treeRepr T
  echo "== getTypeImplSkip T"
  echo treeRepr getTypeImplSkip T
  echo "=="

dumpT(int | float | string) # Gets none
dumpT(SomeInteger) # But this would give the full type
```

### Actual Output

```
== T
Infix
  Sym "|"
  Infix
    Sym "|"
    Sym "int"
    Sym "float"
  Sym "string"
== getTypeImplSkip T
Sym "none"
==
== T
Sym "SomeInteger"
== getTypeImplSkip T
BracketExpr
  Sym "or"
  Sym "int"
  Sym "int8"
  Sym "int16"
  Sym "int32"
  Sym "int64"
  Sym "uint"
  Sym "uint8"
  Sym "uint16"
  Sym "uint32"
  Sym "uint64"
==
```

### Expected Output

```
== T
Infix
  Sym "|"
  Infix
    Sym "|"
    Sym "int"
    Sym "float"
  Sym "string"
== getTypeImplSkip T
BracketExpr
  Sym "or"
  Sym "int"
  Sym "float"
  Sym "string"
==
== T
Sym "SomeInteger"
== getTypeImplSkip T
BracketExpr
  Sym "or"
  Sym "int"
  Sym "int8"
  Sym "int16"
  Sym "int32"
  Sym "int64"
  Sym "uint"
  Sym "uint8"
  Sym "uint16"
  Sym "uint32"
  Sym "uint64"
==
```

### Additional Information

* I'm not sure if `or[int, int8, ...]` is actually something we want to expose

### LINKED ISSUE #1517: chained type `or` expression doesn't work when passed as parameter to typedesc
### Example

```nim
template foo(T: typedesc) = discard

foo(int or string or float)
```

### Actual Output

```
test.nim(3, 19) Error: type mismatch: got <int or string, typedesc[float]>
but expected one of:
proc `or`(a, b: typedesc): typedesc
  required type for a: typedesc
  but expression '[type node]' is of type: int or string
proc `or`(x, y: bool): bool
  required type for x: bool
  but expression '[type node]' is of type: int or string
proc `or`(x, y: int): int
  required type for x: int
  but expression '[type node]' is of type: int or string
proc `or`(x, y: int16): int16
  required type for x: int16
  but expression '[type node]' is of type: int or string
proc `or`(x, y: int32): int32
  required type for x: int32
  but expression '[type node]' is of type: int or string
proc `or`(x, y: int64): int64
  required type for x: int64
  but expression '[type node]' is of type: int or string
proc `or`(x, y: int8): int8
  required type for x: int8
  but expression '[type node]' is of type: int or string
proc `or`(x, y: uint): uint
  required type for x: uint
  but expression '[type node]' is of type: int or string
proc `or`(x, y: uint16): uint16
  required type for x: uint16
  but expression '[type node]' is of type: int or string
proc `or`(x, y: uint32): uint32
  required type for x: uint32
  but expression '[type node]' is of type: int or string
proc `or`(x, y: uint64): uint64
  required type for x: uint64
  but expression '[type node]' is of type: int or string
proc `or`(x, y: uint8): uint8
  required type for x: uint8
  but expression '[type node]' is of type: int or string

expression: [type node] or float
```

### Expected Output

No error

### Additional Information

* Replace `or` with `|` and this will compile

zerbina:
## Summary

* fix `and` and `or` not returning `typedesc` in expression context
* fix `|` not creating an `or` meta type

Fixes https://github.com/nim-works/nimskull/issues/1516
Fixes https://github.com/nim-works/nimskull/issues/1517

## Details

* turn the `and` and `or` system routines into `typedesc` templates
  that use the language built-in `and` and `or` type operators, which
  requires no special support from the compiler, and it also yields a
  proper `typedesc`
* remove the `and`, `or`, and `|` typetraits, as they're no longer used
* turn into `|` system procedure into a `typedesc` template that
  works the same as the `or` template. Previously, the `|` procedure
  returned a `typedesc[None]`

The templates have to use `mixin or` and `mixin and`, as otherwise
the identifiers would bind the already defined `and` and `or` boolean
operators.

### DIFF
--- a/compiler/sem/semmagic.nim
+++ b/compiler/sem/semmagic.nim
@@ -205,10 +205,6 @@ proc evalTypeTrait(c: PContext; traitCall: PNode, operand: PType, context: PSym)
 
   let s = trait.sym.name.s
   case s
-  of "or", "|":
-    return typeWithSonsResult(tyOr, @[operand, operand2])
-  of "and":
-    return typeWithSonsResult(tyAnd, @[operand, operand2])
   of "not":
     if traitCall.len == 3:
       c.config.internalAssert traitCall[2].kind == nkNilLit
--- a/lib/system.nim
+++ b/lib/system.nim
@@ -123,11 +123,20 @@ proc compileOption*(option, arg: string): bool {.
 {.push warning[GcMem]: off, warning[Uninit]: off.}
 # {.push hints: off.}
 
-proc `or`*(a, b: typedesc): typedesc {.magic: "TypeTrait", noSideEffect.}
+template `or`*(a, b: typedesc): typedesc =
   ## Constructs an `or` meta class.
+  mixin `or`
+  a or b
 
-proc `and`*(a, b: typedesc): typedesc {.magic: "TypeTrait", noSideEffect.}
+template `|`*(a, b: typedesc): typedesc =
+  ## An alias for `or <#or,typedesc,typdesc>`_. Constructs an `or` meta class.
+  mixin `or`
+  a or b
+
+template `and`*(a, b: typedesc): typedesc =
   ## Constructs an `and` meta class.
+  mixin `and`
+  a and b
 
 proc `not`*(a: typedesc): typedesc {.magic: "TypeTrait", noSideEffect.}
   ## Constructs an `not` meta class.
@@ -1478,10 +1487,6 @@ const
 
 
 include "system/memalloc"
-
-
-proc `|`*(a, b: typedesc): typedesc = discard
-
 include "system/iterators_1"
 
 
--- a/tests/lang_objects/metatype/tmetatype_issues.nim
+++ b/tests/lang_objects/metatype/tmetatype_issues.nim
@@ -157,3 +157,24 @@ block t3338:
   var t2 = Bar[int32]()
   t2.add()
   doAssert t2.x == 5
+
+block or_result_not_usable:
+  # regression test for the result of `or` (in expression context) not being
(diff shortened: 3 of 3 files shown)


saem (approved):

alaviss (approved):
Thanks for the quick fix.

It's certainly weird to know that we have a set of super magical operators that isn't declared in `system`. But nothing would explain `T not nil` otherwise :stuck_out_tongue: 

While I'd prefer that we don't have those, I don't think it's anything to fret about until we figure out how to properly handle `typedesc` and whether it should be a kind of CT-specific value rather than the current state of being a Schrodinger type.

PR #1528, opened 2025-04-17
mirgen: fix internal type error

zerbina:
## Summary

Fix a type error in the `mirgen`-produced MIR code. Observable program
and compiler behaviour doesn't change.

## Details

For `var x; x in {1, 2, 3}` where `x` has a type other than `int`,
`mirgen` produced less-equal operations where the operands had a
different types, which is disallowed according to the MIR's semantics.

Nothing enforces these semantics, nor does anything rely on the
operand's types being equal, so this went by unnoticed.

To fix the issue, the upper or lower bound check is omitted if
unnecessary, and the integer constant operand is changed to always use
the same type as the to-be-range-checked value (`x` in the above
example).

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -874,37 +874,47 @@ proc genInSetOp(c: var TCtx, n: PNode) =
     let
       se = n[1]
       x  = n[2]
-      elemTyp = x.typ.skipTypes(abstractRange)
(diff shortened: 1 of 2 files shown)


saem (approved):

zerbina:
Two different known-to-be-flaky test failures in a row. Third time's the charm.

PR #1529, opened 2025-04-18
system: add the `unreachable` routine

zerbina:
## Summary

Add the `unreachable` routine, plus some convenience overloads, to the
`assertions`  module.  `unreachable`  behaves the same as 
`doAssert false` ,
but it also marks the control-flow path as `.noreturn`.

## Details

Most if not all projects using `case` expressions a lot tend to require
some sort of  `unreachable` , and giving its foundational nature, it
makes
sense to have `unreachable` be part of `system`.

The implementation from `compiler/utils/idioms` is moved into the
`system/assertions` module, with some improvements made to the doc
comments. All now-unused imports of the `idioms` module in the
compiler's codebase are removed.

### DIFF
--- a/lib/system/assertions.nim
+++ b/lib/system/assertions.nim
@@ -141,6 +141,15 @@ template unreachable*() =
   ##
   ## but the intention is clearer and the call marks the execution path as not
   ## returning anything.
+  runnableExamples:
+    for i in 1..5:
+      if i mod 2 == 0:
+        echo:
+          case i
+          of 2: "two"
+          of 4: "four"
+          else: unreachable()
+
   unreachableImpl("", instantiationInfo(-1))
 
 template unreachable*(msg: string) =
--- a/lib/system/assertions.nim
+++ b/lib/system/assertions.nim
@@ -113,3 +113,42 @@ template doAssertRaises*(exception: typedesc, code: untyped) =
     except: raisedForeign()
   if wrong:
     raiseAssert(begin & " nothing was raised" & msgEnd)
+
+# ------------- unreachable -------------
+
+func unreachableImpl(str: string, loc: InstantiationInfo) {.
+    noinline, noreturn.} =
+  var msg: string
+  msg.toLocation(loc.filename, loc.line, loc.column + 1)
+  msg.add:
+    if str.len > 0: " unreachable: "
+    else:           " unreachable"
+  msg.add str
+  raiseAssert(msg)
+
+func unreachableImpl(e: enum, loc: InstantiationInfo) {.noinline, noreturn.} =
+  ## A bit more efficient than ``unreachable($e)``, as the stringification
+  ## code is not part of the callsite (less I-cache pressure).
+  unreachableImpl($e, loc)
+
+template unreachable*() =
+  ## Marks a point in the program as being unreachable. At run-time, behaves
+  ## the same as:
+  ##
+  ## .. code-block:: nim
+  ##
+  ##   doAssert false, "unreachable"
+  ##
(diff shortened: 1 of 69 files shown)


lib/system/assertions.nim
```diff
@@ -113,3 +113,42 @@ template doAssertRaises*(exception: typedesc, code: untyped) =
     except: raisedForeign()
   if wrong:
     raiseAssert(begin & " nothing was raised" & msgEnd)
+
+# ------------- unreachable -------------
+
+func unreachableImpl(str: string, loc: InstantiationInfo) {.
+    noinline, noreturn.} =
+  var msg: string
+  msg.toLocation(loc.filename, loc.line, loc.column + 1)
```
alaviss, line 122: I suppose this is to allow a location to pops out even when `--stacktrace:off` is used?

I think the answer is yes (since we won't pass repro builds otherwise) but I'll just confirm it again: if `--excessiveStackTrace`/`--listFullPaths` is off, the paths will not be absolute, right?

lib/system/assertions.nim
```diff
@@ -113,3 +113,42 @@ template doAssertRaises*(exception: typedesc, code: untyped) =
...
+template unreachable*() =
+  ## Marks a point in the program as being unreachable. At run-time, behaves
+  ## the same as:
+  ##
+  ## .. code-block:: nim
+  ##
+  ##   doAssert false, "unreachable"
+  ##
+  ## but the intention is clearer and the call marks the execution path as not
+  ## returning anything.
```
alaviss, line 143: I think it's worth highlighting that this is a good way to handle impossible cases in a conditional expression like case statement.

The example I wrote here is not very good, but the idea here is that some examples would be helpful.

```suggestion
  ## but the intention is clearer and the call marks the execution path as not
  ## returning anything and can be used within conditional expressions.
  runnableExamples:
    type Direction = enum Up, Down, Left, Right
    proc oppositeX(dir: Direction): Direction =
      case dir
      of Left: Right
      of Right: Left
      else: unreachable()
```

alaviss (approved):
LGTM.

I only have a small query and a documentation suggestion for this.

lib/system/assertions.nim
zerbina: > I suppose this is to allow a location to pops out even when `--stacktrace:off` is used?

Yep, that's correct. It's the same thing `assert` does, too.

> I think the answer is yes (since we won't pass repro builds otherwise) but I'll just confirm it again: if `--excessiveStackTrace`/`--listFullPaths` is off, the paths will not be absolute, right?

From the perspective of `unreachableImpl`, the message will contain whatever `loc` provides. The default for `instantiationInfo` (which is what `unreachable` uses) is for the "short name" of the module to be returned, which is just the filename (without the path).

lib/system/assertions.nim
zerbina: I'll add an example.

saem: This example might work better.

```nim
runnableExamples:
  let arr = [1, 2, 3, 4, 5]
  for i in arr.items:
    if i mod 2 == 0:
      case i
      of 2: echo "two"
      of 4: echo "four"
      else: unreachable()
```

saem (approved):
I like it, I left a suggested example; otherwise I think it's good to merge.

lib/system/assertions.nim
zerbina: I've used the `mod` example with two small adjustments:
* `1..5` instead of an extra array
* a case expression instead of a case statement, as the case statement would also compile with `unreachable` replaced by `discard`

PR #1530, opened 2025-04-19
add the `suspend` language primitive

zerbina:
## Summary

Add the `suspend` form to the language, which is used for reifying
local, delimited continuations. It provides a basic building-block for
writing sub-programs in continuation-passing style.

## Details

The feature is designed such that it provides an efficient, low-level
way for reifying continuations, to allow for building abstractions on
top of it. The `suspend` interface is a compromise between ergonomics,
ease-of-implementation, and compiler robustness.

For implementing `suspend`, the lower-level `fork` and `resume`
operators are added to the MIR, which `suspend` is lowered to as part
of `mirgen`.

As much as possible of the static semantics is implemented in the sem
phase, but some errors can only be detected once the contents of the
context object are known, requiring their reporting to be part of the
relevant MIR passes. These errors are therefore only reported for live
routines.

Due to shortcomings of the MIR, using `suspend` in during compile-time
evaluation is currently not possible.

### Concrete Changes

* add the `nkSuspend` node kind
* add the `mSuspend` magic
* implement `mSuspend` analysis in `semmagic` and `semexprs`
* add `nkSuspend` handling in `dfa`, `sempass2`, `semfold`,
  and `transf`
* add translation logic for `nkSuspend` to `mirgen` and `proto_mir`
* add the `mnkFork` and `mnkResume` MIR forms
* add the `createForwardedOps` and `completeForwardedOps` procs, used
  for creating type-bound ops whose header is created separately from
  the body; needed for the type-bound ops of continuation
  context objects
* add the `mirtypes.complete` procedure
* add support to the `backends.process` iterator for completing
  forwarded procedures during the MIR phase, by keeping track of
  registered forwarded procedures and polling whether they're completed
* make the `tailcall_analysis.genApply` procedure not require access to
  a `PContext` instance
* add support for creating new procedures that have a MIR body
  already (via `MirEnv.pbodies`)
* add the the `continuations` module, providing the 'fork' and 'resume'
  lowering passes

### Internal Fixes

* fix `mirtrees.subNode` iterating further than it should,
  when `start > 0`

---

## To-Do
* [x] implement the `fork`/`land` lowering pass
* [x] ~make the `.tailcall` lowering a MIR pass (so that it can happen after `fork`/`land` lowering)~
* [x] move the fix for generic `.tailcall` routines to a separate PR
* [x] rename `land` to `resume` (the existing `mnkResume` needs to be renamed to `mnkUnwind`)
* [x] update the manual regarding `suspend` with the recent changes
* [x] implement the static semantics for `suspend` as described in the manual
* [x] add tests for `suspend`
* [x] cleanup the tests some more
* [x] write a proper commit message


compiler/sem/semmagic.nim
```diff
@@ -410,6 +410,100 @@ proc semPrivateAccess(c: PContext, n: PNode): PNode =
   c.currentScope.allowPrivateAccess.add t.sym
   result = newNodeIT(nkEmpty, n.info, getSysType(c.graph, n.info, tyVoid))
 
+proc semSuspend(c: PContext, n: PNode, s: PSym, flags: TExprFlags): PNode =
+  ## Analyzes a 'suspend' magic call, producing a typed AST or an error. If
+  ## the call doesn't have the right shape, analysis fall back to overload
```
saem:
```suggestion
  ## the call doesn't have the right shape, analysis falls back to overload
```

Possible typo.

compiler/sem/semmagic.nim
```diff
@@ -410,6 +410,100 @@ proc semPrivateAccess(c: PContext, n: PNode): PNode =
...
+  prc.addParam("c", newTypeWithSons(c, tySink, @[obj]), n.info, c)
+
+  # set up the type to use for the local:
+  let tup = newTypeS(tyTuple, c)
+  tup.rawAddSon(obj)
+  tup.rawAddSon(prc)
+
+  c.openScope()
+  # create a let section and type that. This makes sure the symbol is properly
+  # registered everywhere, and retyping is also taken care. The initializer
```
saem:
```suggestion
  # registered everywhere, and retyping is also taken care of. The initializer
```

Minor typo.

doc/manual.rst
```diff
@@ -4166,6 +4166,85 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+
+A continuation is created via the `system.suspend` procedure, which
+accepts three arguments.
+
+Syntactically, the second argument must be an identifier.
+
+1. let `T` be the type of the first argument
+2. let `R` be the return type of the caller
+3. let `Ctx` be a unique type such that:
+  * `Ctx is object` evaluates to true
```
saem, line 4187:
```suggestion
  * `Ctx is object` evaluates to true
```

Should this be tuple?

doc/manual.rst
```diff
@@ -4166,6 +4166,85 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+  * an instance of `Ctx` is always copyable
+4. let `P` be a type:
+  * if `T is void`, then let `P` be `proc(e: sink Ctx): R`
+  * if `T isnot void`, then let `P` be `proc(p: sink T, e: sink Ctx): R`
+5. let `cont` refer to the *continuation* (a procedure of type `P`)
+6. let `local` refer to the identifier appearing as the second argument
+7. let `ctx` refer to the saved context (a value of type `Ctx`)
+8. let `call` refer to the third argument expression
+
+`call` is typed as if would the expression were the following:
```
```suggestion
`call` is typed as if the expression were the following:
```

doc/manual.rst
saem: Nevermind, I'm thinking for the closure

doc/manual.rst
```diff
@@ -4166,6 +4166,85 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+
+The value returned by a resumed caller is returned by the continuation
+invocation -- the same goes for raised exceptions.
+
+Cleanup
+~~~~~~~
+
+If `ctx` (or a copy thereof) goes out scope without having been consumed by a
+call to the continuation, cleanup happens as if unwinding would take place
+right the `suspend` call, but without `finally`:idx: clauses being visited.
```
```suggestion
right after the `suspend` call, but without `finally`:idx: clauses being visited.
```

I think that's what was meant?

lib/system.nim
```diff
@@ -3070,6 +3070,12 @@ when defined(nimDebugUtils):
 proc `not`*[T: ref or ptr](a: typedesc[T], b: typeof(nil)): typedesc {.magic: "TypeTrait", noSideEffect.}
   ## Constructs a `not nil` type.
 
+when defined(nimskullHasSuspend):
+  proc suspend*[T](with: typedesc[T], name, call: untyped): T {.magic: "Suspend", noSideEffect.} =
```
saem: As of this moment, reading the PR, reminding myself or redex vs continuation definitions (Lexi's talk), and seeing as `suspend` creates a reified continuation, I'm inclined to say `continuation` is a better name than `suspend`.

Beyond, continuation is largely what would follow in a `continuation(...): ...` call, I have a feeling that `suspend` is a bit too evocative of something like async/await or coroutines.

---

The above is my current thing, and undoubtedly as I reread this PR as it evolves, and especially the docs, I'll end up with a different suggestion(s) and preference. I would just treat this as a data point.

alaviss: Here's an alternative interface I made to look less magical:

`proc suspend[T; Prc: proc](with: typedesc[T], name, prc: Prc, args: tupleFromArgs(Prc)): T`

where `tupleFromArgs()` is a "magic" that convert the argument list into a tuple type

---

Re: the name `suspend`, I think it's a good one, as when viewed from the writer's PoV, you "stop" the procedure at the specified point, and then resume it elsewhere. The fact that it creates a continuation is merely an implementation detail so you can do the resuming part.

doc/manual.rst
```diff
@@ -4166,6 +4166,85 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+* no local with a disabled copy operator must *potentially* store a value
+* no `var`:idx: or `openArray`:idx: parameter (or a borrow thereof) must be live
+* the `suspend` call is not statically located within an `except`:idx: or
+  `finally`:idx: clause
+* the `suspend` call does not appear in top-level code
+
+Evaluation of `suspend`:idx: works as follows:
+1. the current local context is saved
+2. `local` is intialized with the tuple `(ctx, prc)`
+3. all local variables (except `local`) or sink parameters of the caller used
```
```suggestion
3. all local variables (except `local`) and sink parameters of the caller used
```

I believe this was the intention

lib/system.nim
saem: I like the generality of `tupleFromArgs`, it'd be handy for template/macro writing. I do think writing `suspend` directly will be less pleasant. Perhaps overloading can help there, with something like a typed macro extracting the caller from a call.

zerbina: I agree that having `tupleFromArgs` (though I'd name it differently) would be useful. It's possible to implement via a macro, but this is error prone and effectively requires using the unstable macro type API.

---

> Here's an alternative interface I made to look less magical:
> 
> `proc suspend[T; Prc: proc](with: typedesc[T], name, prc: Prc, args: tupleFromArgs(Prc)): T`

As is, this won't work, for multiple reasons:
1. `name` being `untyped` was deliberate, it's not a procedure: the parameter accepts an untyped identifier. However, changing `name` to `untyped` in your signature also won't work
2. proper parameter passing cannot be emulated with a tuple. Of course, one could require the argument being an immediate tuple construction and then construct a call from it, but this a lot more "magical" - for both the compiler and user - than the current interface
3. using overloaded routines as the receiver (i.e. `prc`) is not possible

There's also another internally-facing problem: there's no call of `prc` in the typed AST, which means that multiple places in the compiler would need special handling for `suspend` (which the current interface gets around).

Finally, keep in mind that `suspend` is a *non-overloadable magic* at the moment, and with its current shape, it cannot be anything else. In other words, the signature you see in system is largely irrelevant to the compiler -- it's only there for documentation purposes.

Ultimately, I believe that dedicated abstract (and concrete) syntax would be a better fit for this feature, though I'm rather reluctant to take that route.

zerbina:
Hm, there's one thing I didn't consider: address stability. The address of a local doesn't stay the same across a `suspend`, which would mean that code such as:
```nim
proc test() =
  var a = 0
  var b = addr a
  suspend ...
  echo b[]
```

is going to invoked undefined behaviour upon evaluation `b[]`, as `b` would point into the memory of a dead frame at that point.

The only solution to this I can think of is disallowing taking the address of a location where the address value *might* be live beyond a `suspend` call. That is:
```nim
proc test() =
  var a = 0
  block:
    var b = 0
    discard addr(b) # allowed
  discard addr(a)   # disallowed
  suspend ...
  discard addr(a) # allowed, because there's no following `suspend`
```

Of course, this rule won't be able to reject code such as:
```nim
proc hiddenAddr(x: var int): ptr int =
  addr x

proc test() =
  var a = 0
  var b = hiddenAddr(a)
  suspend ...
  echo b[]
```

Finally, implementing this rule is going to be a nuisance: either `sempass2` needs to be extended with a dedicated pass (prone to mistakes and hard to maintain), or a MIR pass has to be used, which would mean that neither `nim check` or `nimsuggest` will report an error.

zerbina:
Some more fundamental changes to the MIR are in order for this PR to progress, I think. While not *impossible*, it's currently quite tricky to add types/procedures not based on pre-existing `PType`s/`PSym`s; the same goes for *modifying* existing MIR types and procedures.

With how I'm planning to implement `fork` and `land`, modifying types is necessary for producing the final continuation context `object`, while modifying procedures is necessary for filling the continuation context's hooks with life.

Using enough hacks and maiming the compiler even more than the MIR-based implementation likely will *could* maybe allow for using a `PNode`-based AST transformation, but I'd rather not do that, even though it's likely (but not guaranteed) to be quicker to implement.

zerbina:
There is another issue that I haven't put much thought into so far. At the moment (in a language without `suspend`), `finally` (or `defer`) may be used for implementing, among other things, clean up that *must* happen whenever control-flow leaves a scope.

`break`, `continue`, `raise`, etc. all enter `finally` sections, but `suspend` itself does not (nor can and should it). If the continuation is resumed exactly once, the aforementioned poses no problem, as the `finally` section enclosing the `suspend` is guaranteed to be run.

However, if the continuation is resumed more than once, the `finally` section would be entered more than once. Similarly, if the continuation is dismissed (i.e., never resumed), the `finally` section is never entered. Both cases would effectively break all `finally`-based cleanup/destruction logic.

Using automatic cleanup via destructors addresses this problem, as copying the continuation (necessary for resuming it multiple times) duplicates all locals part of the environment, whereas not using it / dismissing it (which calls the continuation context's destructor) automatically destroys all locals part of the environment.

Not all cleanup logic is able to use destructors, however, such as manually managed resources or ones whose cleanup requires additional state. This is problematic for composition. Consider the following:
```nim
template with(path: string, name, body: untyped) =
  block:
    var name = open(name, fmRead) # open file
    defer: name.close()
    body

```
without `suspend`, the template's author can rely on `name` always being closed (unless the code calls `quit`, of course), but with `suspend`, they no longer can, as the code passed to the template might suspend the running procedure and then dismiss the continuation.

There are multiple solutions that come to mind:
1. ignore the problem and do nothing (a non-solution)
2. disallow `suspend` within a local `try`s that have a `finally` clause (including the implicit `try`/`finally` created by `defer`)
3. run the `finally` clauses that apply to the `suspend` call when the continuation (and its context) are dismissed

Number 3 is only a partial solution, however, as while it prevent the finallys from being not run at all, it doesn't protect against them being evaluated multiple times. It's also tricky (maybe even impossible given the current compiler) to implement.

Number 2 is the only solution that fully solves the problem, but it would significantly limit the usefulness of `suspend`, making it compose much worse.

If anyone has another idea for a solution, please let me know.

zerbina:
Okay, after some more thinking, I've come to the conclusion that solution 2 (i.e., disallow `suspend` being within a `try` with a `finally` clause) is the only viable solution at the moment.

While less broadly applicable, `suspend` will still be useful, and further development of the language + compiler improvements might make it possible to lift the restriction again.

saem:
> While less broadly applicable, `suspend` will still be useful, and further development of the language + compiler improvements might make it possible to lift the restriction again.

Something to work and build with is better than nothing, and better than an architectural/design deadend

zerbina:
I gave up on the AST transformation approach. While doable, it would just be a maintenance nightmare, as the compiler lacks the tools for doing AST transformations in a maintainable way. Still, I made the transformation work to the point of `showcase.nim` at least compiling and running.

Large parts of the static semantics are currently not enforced, dynamic semantics in `suspend`-using routines change in subtle ways, and it's easy to crash the compiler, but it's nonetheless possible to try out the feature now, should anyone be interested.

While I'm aware that the implementation is very shaky, I'd still be interested in bug/crash reports, as they would provide some additional guidance to what needs test coverage.

zerbina:
Hm, it's obvious now that I've thought about it, but `suspend` cannot be used in `for` loop bodies, given how `yield` is currently implemented. Duplicating the `for` loop body into every additional `yield` statement means that `suspend` statements/expressions have to be duplicated too, but this would require also creating new continuation types, something that's not possible anymore after `sem`.

I'll adjust the specification so that `suspend` in a `for` loop is disallowed. Changing the `for` loop transformation such that the `for` body is not duplicated anymore (which is a good idea in general) will allow for `suspend` being usable in `for` loop bodies.

**---- Edit start ----**

On second thought, there's a fairly straightforward solution to the aforementioned problem, and it's one that'd also make way for some future improvements. The idea is to support multiple `land` operations sharing the same context type. When this is the case, the context object is turned into a `case` object, using a branch per static `land`.

In the future, this MIR feature could be made available at the NimSkull-level, where it'd be useful for implementing something like `.closure` iterators using `suspend`, something that's not possible to do efficiently with the current `suspend`.

**---- Edit end ----**

---

Separately, I'm also going to change the `suspend` specification such that not only a (limited subset) of call expressions are allowed in the "suspension context" position, but all statements. `suspend` will gets its own syntax form in the AST anyway, and the current restrictions are just too limiting in practice.

zerbina:
In order to get something that works in a shorter amount of times, I've cut some corners:
* **continuation procedures don't use `.tailcall` anymore.** This is because of a phase-ordering issue: `fork`/`land` lowering happens after portable tail-call elimination, but it needs to happen the other way around. Tail-call elimination needs to become a MIR pass, and while initially planned and not extraordinarily hard, it's out of scope for this PR, so I've postponed it.
* **`suspend` cannot be used in `.tailcall` routines anymore.** Doesn't work due to the aforementioned phase-ordering issue.
* **immutable parameters cannot be used beyond a `suspend`.** Previously, the idea was to copy such parameters into the continuation context object, but this is fairly tricky to implement. The compiler simply reports an error now.
* **taking the address of a to-be-saved local is undefined behaviour (instead of causing an error).** It's fairly easy to implement, but it would mean adding yet more misplaced error reporting.

Beyond the corner cutting, the `fork`/`land` lowering is now complete. It being a MIR pass (as originally intended) makes it much more robust, when compared to the AST-based pass. Since move optimization took place by the time routines are split up, the saved continuation context is usually also a lot smaller.

---

`suspend` has rather awkward and non-intuitive operational semantics, which also shows up in the implementation. Locals used within the `suspend` body need to be captured, as - from the perspective of the `suspend` body - the values have been saved into the continuation already. `transf` implements the relevant logic, as it's best equipped for such search-and-replace tasks. 

I (still) think replacing the `suspend` interface with something along the lines of "call-with- current-continuation" (calling a routine and passing the current continuation as the x-th parameter), but that's something for future work.

zerbina:
The PR is almost ready for review, but it's not fully there yet -- I've moved it out of the draft state so that test results can be inspected already, making it easier to track progress.

---

Overall, the implementation is okay-ish. The continuation-related splitting happens at a fundamentally incorrect phase, but that's due to destructor injection and move analysis happening too late (i.e., during the MIR phase instead of as part of semantic analysis).

Using the MIR to host the continuation transformation exacerbates the problem already present for move analysis and destructor injection: namely, that the MIR's detached-from-source nature makes it easy change semantics w.r.t., destruction and/or where moves/copies happen by applying some seemingly minor and/or innocuous improvement to, e.g., `mirgen`.

Anyhow, this is something that'll, most likely, only be addressed in the far future. At the moment, an AST-based transformation would be much worse (as made evident - at least in parts - by [this now-reverted commit](https://github.com/nim-works/nimskull/pull/1530/commits/772f75369048abd90971fde9792f3b4812e76f9f)).

doc/manual.rst
```diff
@@ -4166,6 +4166,112 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+
+The value returned by a resumed callee is returned by the continuation
+invocation -- the same goes for raised exceptions.
+
+Copying
+~~~~~~~
+
+When copying a continuation instance, all saved values in the context object
+are copied, via a full copy, in the reverse order the locals they come from
+were declared.
```
alaviss, line 4264: I'm not sure if defining a copy ordering is useful. This sounds like something implementations should be free to decide on.

doc/manual.rst
```diff
@@ -4166,6 +4166,112 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+concrete first-class value (a *continuation instance*), the `suspend`:idx: form
+is used.
+
+A continuation instance is the reified continuation (a procedure) plus the
+saved state needed to run it.
+
+Every available call syntax may be used for `suspend`, including the command
+and dot-call syntax. `suspend` is not a real procedure and does not participate
+in overload resolution, but the name in the callee position may be qualified
+as `system.suspend`.
```
alaviss, line 4182: This will have to be clarified in terms of overloading resolution.

What would "not participate in overload resolution" mean? Does that mean this is basically a new keyword? Or is it more that "when `system.suspend` is in scope, it always win"? What happens if `import system except suspend` is used?

doc/manual.rst
```diff
@@ -4166,6 +4166,112 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+4. let `P` be a type such that:
+  * if `T is void`, then `P` is `proc(e: sink Ctx): R {.nimcall.}`
+  * if `T isnot void`, then `P` is `proc(p: sink T, e: sink Ctx): R {.nimcall.}`
+5. let `cont` refer to the *continuation* (a procedure of type `P`)
+6. let `ctx` refer to the saved context (a value of type `Ctx`)
+7. let `L` refer to the location which is named by the
+   second argument
+8. let `blk` refer to the third argument
+
+Both `T` and `R` must be concrete types.
```
alaviss, line 4200: It might be better to qualify this as "concrete types or `void`". I personally don't want `void` to be considered a "concrete" type, but compiler side might have a different take on this.

At the very least, `void` is special enough to warrant an explicit mention.

doc/manual.rst
```diff
@@ -4166,6 +4166,112 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+  * an instance of `Ctx` is always copyable
+4. let `P` be a type such that:
+  * if `T is void`, then `P` is `proc(e: sink Ctx): R {.nimcall.}`
+  * if `T isnot void`, then `P` is `proc(p: sink T, e: sink Ctx): R {.nimcall.}`
+5. let `cont` refer to the *continuation* (a procedure of type `P`)
+6. let `ctx` refer to the saved context (a value of type `Ctx`)
+7. let `L` refer to the location which is named by the
+   second argument
+8. let `blk` refer to the third argument
+
```
alaviss, line 4199: I think it might be worth it to write down the `proc` prototype `suspend` roughly corresponds to here, with a disclaimer that it's for visualization purposes only. That would make the docs a bit easier to read and cross-reference.

doc/manual.rst
```diff
@@ -4166,6 +4166,112 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+Both `T` and `R` must be concrete types.
+
+If `R` is `void`, `blk` must be a statement. If `R` is not `void`, `blk` may
+either be an expression fitting type `R` or a statement.
```
alaviss, line 4203: I don't think "fitting type" is a terminology that we have precedents for in the manual, but I might be wrong.

Would something like "expression of type `R`" be better?

doc/manual.rst
```diff
@@ -4166,6 +4166,112 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+* inside a `try` block that has a `finally` clauses
+* in a statement list or expression after a `defer` statement
+* inside a `finally` or `except` clause
+* inside a `defer` block
+* inside a `static` or `const` context
+* as or within the third argument of a `suspend`
+* outside of a routine
+* as part of an `iterator` body
+* as part of a `.tailcall` routine body
+
```
alaviss, line 4218: I think right now we have restrictions for `method`, `converter` and `macro` too, right?

doc/manual.rst
```diff
@@ -4166,6 +4166,112 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+Saving the context refers to saving the value of locals (which includes owning
+parameters and the `result` variable, if any) in a context object. Views (both
+implicit and explicit) and non-owning parameters cannot be saved: if the
+continuation would needs them, a static error is reported. When a local that
+needs to be saved has a type that is non-copyable, a static error is reported.
```
alaviss, line 4223: Opinion: we should not specify the mechanism of action ("a static error is reported"), but instead something along the lines of "is a semantic error" or "is invalid".

doc/manual.rst
```diff
@@ -4166,6 +4166,112 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+
+The value of the following locals need to be saved:
+* locals requiring cleanup and that store a value at the time the
```
alaviss, line 4226: I think it's worth removing this point from the spec or downgrade it into details about the current implementation.

The reason is that we should leave the door open to allow for immediate cleanup of unused locals. As it stands, suspending already meant the useful lifetime of such locals are very limited, as old locations are more or less gone.

alaviss (commented):
Some of my initial thoughts looking at the manual.

doc/manual.rst
zerbina: For better or worse, copying usually implies destruction and is also allowed to have high-level effects (I/O, modification of globals, etc.). Given this circumstance, I'm of the opinion that the copy order should be well-defined.

doc/manual.rst
zerbina: It works the same as `addr`, `sizeof`, `quote`, and other pseudo-keywords do.

doc/manual.rst
zerbina: `void` is a concrete type, as it carries all information necessary to make type judgements about and regarding it.

I'll make it explicit that this covers `void`.

doc/manual.rst
alaviss: I'm of the opinion that developers should not rely on such details and that copying should be unordered. We already limit what can be done in hooks and I don't think it's wise to enable questionable (ab)use of implementation details.

doc/manual.rst
zerbina: Hm, what made you conclude this? There are no such restrictions, except for maybe `macro`, which is theoretically supported, but having `suspend` inside it currently leads to an "cannot evaluate at compile time" error.

alaviss: It was just a wild guess, apologies.

Though allowing it on `converter` is very questionable, as no one ever explicitly invokes one. But then again I think converters are weird to begin with.

In general this not an issue.

doc/manual.rst
zerbina: I strongly disagree. `suspend` should have as little effect on the local behaviour of the surrounding code as possible, and this includes lifetimes and cleanup.

doc/manual.rst
zerbina: > Would something like "expression of type `R`" be better?

Maybe, I'm not entirely sure.

For what it's worth, "expression of type `R`" is what I had used originally, but I changed it because it's ambiguous whether "of type T" includes subtyping or not. In this case, subtyping is allowed.

doc/manual.rst
alaviss: What would be the actual benefit of this? Right now all I can see is that users will have to perform early `drop` to keep their context small, and then we have the issue where temporaries introduced by expressions would bloat the context silently.

This creates a situation similar to the problem with `tailcall` (requiring explicit `drop` for unused resources), except that the compiler would happily keep such copies silently.

doc/manual.rst
alaviss: We can use "can be assigned to a location of type `R`", which would reuse the "Assignment compatibility" definition in the manual.

doc/manual.rst
zerbina: The benefit is that programmer expectations aren't invalidated. It allows reasoning about how the body of a procedure behaves at run-time as if there was no `suspend`.

If the lifetime of temporaries is a problem (in that it requires explicit dropping, makes the saved context too large, etc.), then that should be addressed separately instead of working around it by adding some exceptions to `suspend`.

alaviss: Thinking about it more, I think my concerns regarding cleanup is separate from the actual implementation here.

The main practical issue actually comes from our cleanup rules not allowing early temporaries termination. Solving that should be done elsewhere, but I don't mind if it's done preemptively for continuations.

The other issue about unused locals still stands for me, while technically it would be "the same as" the code if it was not suspended, context bloat is a real practical issue that should be solved. I believe there's a case to tweak the rules to avoid this headache on the user side.

zerbina:
With the standalone tail-call elimination pass from `devel` (which I've merged into the PR), continuation procedures can use the `.tailcall` calling convention and the "cannot suspend in `.tailcall` routine" requirement can be removed (both are implemented/changed already).

One thing to think about is whether to only use the `.tailcall` calling convention when the routine enclosing the `suspend` uses it, the `.nimcall` convention otherwise. This would likely improve performance a bit for routines that don't use tail calls (as `.tailcall` does have some overhead), with the downside that continuation procedures don't all have the same shape anymore, which would make abstraction for erasing the type harder to write.

zerbina:
Update on the progress of this PR, summarizing the insights gathered while discussing the feature on Matrix:

@alaviss made an attempt at using `suspend` for replacing the usage of [Nim-CPS](https://github.com/nim-works/cps) in [nim-sys](https://github.com/alaviss/nim-sys). The main issue alaviss reported was the delimiter being at the current procedure exit making `suspend` hard to apply for use in `nim-sys`, and that implementing a cross-procedure `suspend` on top of the built-in `suspend` is hard if not impossible to implement efficiently and in a way that's mostly transparent to the user.

I've tried using `suspend` as a tool for reifying the remainder of the current procedure, for use with continuation-passing style programs, but the inability to capture `var` and non-owning immutable parameters renders it useless in all real-world applications I could come up with that don't rely heavily on macros and/or only use sink parameters.

Overall, `supend` as it is now seems to be too limited for any practical usage, even when using it purely as a building block.

Without some way to statically or dynamically ensure that captured parameter don't outlive the continuations, so that at least all non-`var` parameters can be used beyond suspension points, I don't see a path towards `suspend` becoming useful.

For the moment, `suspend` is a dead end.

PR #1531, opened 2025-04-19
docs: mark tutorial docs as legacy

saem:
## Summary

Move all three tutorials into a legacy section on the documentation home
page

## Details

Move all three tutorials further down the docs page in under a legacy
title. This is to indicate that they're no longer really maintained, and
they shouldn't supersede the spec, reference, standard library docs,
etc.

### DIFF
--- a/doc/docs.rst
+++ b/doc/docs.rst
@@ -58,6 +58,11 @@ For Users
 For Users (Legacy)
 ==================
 
+.. warning::
+   These documents are not maintained and might not correspond to the
+   current language semantics. Consult the `manual <manual.html>`_ and
+   `standard library <lib.html>`_ for the most up-to-date information.
+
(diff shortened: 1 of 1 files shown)


doc/docs.rst
```diff
@@ -63,3 +54,15 @@ For users
 
 - | `Index <theindex.html>`_
   | The generated index.
+
+For Users (Legacy)
+==================
+
```
alaviss, line 60: We should add a small warning block here (and potentially on top of the documents themselves)

```rst
.. warning::
   These documents are not maintained and might not correspond to the
   current language semantics. Consult the `manual <manual.html>`_ and
   `standard library <lib.html>`_ for the most up-to-date information.
```

alaviss (approved):

PR #1536, opened 2025-04-24
varpartitions: fix memory leak with tuple unpacking

zerbina:
## Summary

Fix a compiler bug that led to destructors not being called for some
local tuple-unpacking definitions.

## Details

Locals introduced via tuple-unpacking `let`/`var` statements were
allowed to be inferred as cursors. Since `mirgen` assumes that
unpacked-into locals are always owning, values are moved into inferred-
as-cursor unpack locals, causing memory leaks due to missing
destruction.

To fix the issue, locals introduced via tuple-unpacking are prevented
from ever being cursors. Alternatively, `mirgen` could be changed to
not assume unpacked-into locals always being owning, but this is more
complex to implement.

### DIFF
--- a/compiler/sem/varpartitions.nim
+++ b/compiler/sem/varpartitions.nim
@@ -807,6 +807,10 @@ proc computeLiveRanges(c: var Partitions; n: PNode) =
       else:
         for i in 0..<child.len-2:
           registerVariable(c, child[i])
+          if child.kind == nkVarTuple:
+            # an unpacking definition where the rhs is not a literal tuple
+            # construction. The vars are always owning
+            pretendOwnsData(c, child[i].sym)
           #deps(c, child[i], last)
 
   of nkAsgn, nkFastAsgn:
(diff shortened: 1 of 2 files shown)


compiler/sem/varpartitions.nim
```diff
@@ -807,6 +807,10 @@ proc computeLiveRanges(c: var Partitions; n: PNode) =
       else:
         for i in 0..<child.len-2:
           registerVariable(c, child[i])
+          if child.kind == nkVarTuple:
+            # an unpacking definition where the rhs is not a literal tuple
+            # construction. The vars are always owning
```
alaviss, line 812: From the PR description, it seems to me that we need this because `mirgen` assumes ownership. I think it's worth adding a note about that, since that would open the door to undoing this in the future (assuming that it's not gonna be done via MIR instead).

alaviss (approved):
Maybe add `varpartitions:` to the title, but this looks good to me.

I do have a small comment about the comment in code.

compiler/sem/varpartitions.nim
zerbina: I actually considered mentioning `mirgen` when writing the comment, and in the past, I employed such mentions/links quite often, but I no longer think that this a good idea, because:
* mentioning expectations of other not-directly-related processing to justify/explain some local behaviour is a workaround to no well-defined IR existing. Continued use of such mentions only exacerbates the ad-hoc semantics of AST
* these manual links are very susceptible to becoming stale, and a lot of them already did

PR #1538, opened 2025-05-01
cli_reporter: print friendlier hook mismatch error message

Luyten-Orion:
## Summary
The correct hook signature is now reported, and is explained in a clear
and concise way.

## Details
Originally an unhelpful error message, 
`signature for '=hook' must be proc[T: object](x: var T)` , was
displayed to the user if they attempted to override the default
implementation for a type.

The error message for incorrect hook signatures have changed to:
* Omit the generic parameters
* Render the  `noSideEffect`  pragma in the type signature, as hooks do
not currently taint the calling code, which could lead to the
expectations of what would have been side-effect free code, to contain
side-effects.
* State the intended meaning of `T` in plain English.
* Use the correct signature for each hook.

`semOverride`  now passes the hook's magic signature rather than the
definition, and we now render the magic's signature instead of
hardcoding the hook values.

### DIFF
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1600,7 +1600,16 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
         "` can be defined only in the same module with its type (" & r.typ.render & ")"
 
     of rsemUnexpectedTypeBoundOpSignature:
-      result = "signature for '" & r.symstr & "' must be proc[T: object](x: var T)"
+      let value =
+        case r.symstr
+        of "=copy":    "proc(x: var T, y: T)"
+        of "=destroy": "proc(x: var T)"
+        of "=sink":    "proc(x: var T, y: T)"
+        of "=trace":   "proc(x: var T, env: pointer)"
+        else:          unreachable()
+
+      result = "'$1' must satisfy the signature '$2' where T is 'distinct' or 'object'" % [
+               r.symstr, value]
 
     of rsemRebidingDeepCopy:
       result = "cannot bind another 'deepCopy' to: " & r.typ.render
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1600,16 +1600,18 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
         "` can be defined only in the same module with its type (" & r.typ.render & ")"
 
     of rsemUnexpectedTypeBoundOpSignature:
-      let value =
-        case r.symstr
-        of "=copy":    "proc(x: var T, y: T)"
-        of "=destroy": "proc(x: var T)"
-        of "=sink":    "proc(x: var T, y: T)"
-        of "=trace":   "proc(x: var T, env: pointer)"
-        else:          unreachable()
+      let typ = copyType(r.typ, r.typ.itemId, r.typ.owner)
 
-      result = "'$1' must satisfy the signature '$2' where T is 'distinct' or 'object'" % [
-               r.symstr, value]
+      # set the calling convention to 'nimcall' and remove 'explicitCallConv' so it's not rendered
+      typ.callConv = ccNimCall
+      typ.flags.excl tfExplicitCallConv
+
+      let msg =
+        if r.symstr.normalize == "=deepcopy": "where T is 'ptr' or 'ref' of either 'distinct' or 'object'"
+        else: "where T is 'distinct' or 'object'"
+    
+      result = "'$1' must satisfy the signature '$2' $3" % [
+              r.symstr, typ.render, msg]
 
     of rsemRebidingDeepCopy:
       result = "cannot bind another 'deepCopy' to: " & r.typ.render
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1602,16 +1602,19 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
     of rsemUnexpectedTypeBoundOpSignature:
       let typ = copyType(r.sym.typ, r.sym.typ.itemId, r.sym.typ.owner)
 
-      # set the calling convention to 'nimcall' and remove 'explicitCallConv' so it's not rendered
+      # set the calling convention to 'nimcall' and remove 'explicitCallConv'
+      # so it's not rendered
       typ.callConv = ccNimCall
       typ.flags.excl tfExplicitCallConv
 
       let msg =
-        if r.sym.magic == mDeepCopy: "where T is 'ptr' or 'ref' of either 'distinct' or 'object'"
-        else: "where T is 'distinct' or 'object'"
+        if r.sym.magic == mDeepCopy:
+          "where T is 'ptr' or 'ref' of either 'distinct' or 'object'"
+        else:
+          "where T is 'distinct' or 'object'"
     
-      result = "'$1' must satisfy the signature '$2' $3" % [
-              r.str, typ.render, msg]
+      result = "'$1' must satisfy the signature '$2' $3" %
+               [r.str, typ.render, msg]
 
     of rsemRebidingDeepCopy:
       result = "cannot bind another 'deepCopy' to: " & r.typ.render
--- a/compiler/front/cli_reporter.nim
+++ b/compiler/front/cli_reporter.nim
@@ -1600,18 +1600,18 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
         "` can be defined only in the same module with its type (" & r.typ.render & ")"
 
     of rsemUnexpectedTypeBoundOpSignature:
-      let typ = copyType(r.typ, r.typ.itemId, r.typ.owner)
+      let typ = copyType(r.sym.typ, r.sym.typ.itemId, r.sym.typ.owner)
 
       # set the calling convention to 'nimcall' and remove 'explicitCallConv' so it's not rendered
       typ.callConv = ccNimCall
       typ.flags.excl tfExplicitCallConv
 
       let msg =
-        if r.symstr.normalize == "=deepcopy": "where T is 'ptr' or 'ref' of either 'distinct' or 'object'"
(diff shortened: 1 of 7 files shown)


compiler/front/cli_reporter.nim
```diff
@@ -1600,7 +1600,16 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
...
 
     of rsemUnexpectedTypeBoundOpSignature:
-      result = "signature for '" & r.symstr & "' must be proc[T: object](x: var T)"
+      let value =
+        case r.symstr
+        of "=copy":    "proc(x: var T, y: T)"
+        of "=destroy": "proc(x: var T)"
+        of "=sink":    "proc(x: var T, y: T)"
+        of "=trace":   "proc(x: var T, env: pointer)"
+        else:          unreachable()
```
alaviss: We ~~don't align of branches~~**EDIT** turns out we do in skull (or nim, for that matter).

You should add a message to `unreachable()`, like `unreachable("unexpected hook: " & r.symstr)`. This would allow a future developer to quickly figure out what case to add.

I'm not sure myself if `unreachable()` is the tool to use here, or if we should go with an `assert` instead to only error in development builds. That would prevent a missed case from crashing the compiler. A fallback string like `FIXME: unknown hook signature` should do.

compiler/front/cli_reporter.nim
```diff
@@ -1600,7 +1600,16 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
         "` can be defined only in the same module with its type (" & r.typ.render & ")"
 
     of rsemUnexpectedTypeBoundOpSignature:
-      result = "signature for '" & r.symstr & "' must be proc[T: object](x: var T)"
+      let value =
+        case r.symstr
```
alaviss:
```suggestion
        case r.symstr.normalize
```

Do normalization to catch `=cOpY` and/or `=c_o_p_y`. You should probably add a test case for that as well.

compiler/front/cli_reporter.nim
alaviss: You should add a case for [`deepcopy`](https://github.com/nim-works/nimskull/blob/fb920984a449b72d45b78e227048412f8ca97a66/doc/manual_experimental.rst?plain=1#L1063-L1081) as well.

alaviss (changes requested):

compiler/front/cli_reporter.nim
saem: > We don't align of branches in skull (or nim, for that matter).

I'm personally ok with it in this case as it does make it easier to read

Luyten-Orion:
My only reason for aligning the branches was just by scrolling up and
looking at the existing style, though it *is* all over the place...

tests/errmsgs/thook_sigmatch_1.nim
```diff
@@ -0,0 +1,8 @@
+discard """
+  errormsg: "'=copy' must satisfy the signature 'proc(x: var T, y: T)' where T is 'distinct' or 'object'"
+  line: 8
+"""
+
+type typ = ref object
+
+proc `=copy`(dest: var typ, src: typ) = discard
```
zerbina:
```suggestion
type Typ = ref object

proc `=copy`(dest: var Typ, src: Typ) = discard
```

It's just a test, but in the repository, types generally start with an uppercase letter. You could also shorten the test by using `int` instead of a ref type.

The same goes for the other tests.

zerbina (changes requested):
Thank you for the PR, Robyn.

Branching over the name of the symbol works somewhat, but it has a few problems:
* it doesn't consider aliases. For example, `=` is also allowed as the copy hook name (whether that's a good thing is a separate issue)
* it's a non-exhaustive `case` expression, meaning that you get no static error when adding a new hook and not updating the case expression
* it duplicates the shape of the signatures, meaning that changing the signature of a hook requires one additional place that manually needs to be updated

Additions or changes to hooks are fairly uncommon, but I still think that the code should be more robust in the face of changes.

My suggestions is the following: if the signature doesn't match, fetch the actual type of the hook from the hooks generic definition in `system.nim`, pass that to the report, and replace the `case` expression with just rendering the type.

This eliminates the decision making in the reporter/renderer and there won't be any additional manual labor required when making some changes to hooks in the future.

The report is produced in the `semOverride` procedure in `compiler/sem/semstmts.nim`. How to fetch the type is not exactly obvious, so if you get stuck, feel free to ping me for assistance/guidance.

tests/errmsgs/thook_sigmatch_1.nim
Luyten-Orion: Resolved by using `int` as the testing type

compiler/front/cli_reporter.nim
```diff
@@ -1600,7 +1600,18 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
         "` can be defined only in the same module with its type (" & r.typ.render & ")"
 
     of rsemUnexpectedTypeBoundOpSignature:
-      result = "signature for '" & r.symstr & "' must be proc[T: object](x: var T)"
+      let typ = copyType(r.sym.typ, r.sym.typ.itemId, r.sym.typ.owner)
+
+      # set the calling convention to 'nimcall' and remove 'explicitCallConv' so it's not rendered
```
```suggestion
      # set the calling convention to 'nimcall' and remove 'explicitCallConv'
      # so it's not rendered
```
Small nit-pick. New code should try to not exceed a line width of 80 characters.

compiler/front/cli_reporter.nim
```diff
@@ -1600,7 +1600,18 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
...
+      typ.callConv = ccNimCall
+      typ.flags.excl tfExplicitCallConv
+
+      let msg =
+        if r.sym.magic == mDeepCopy: "where T is 'ptr' or 'ref' of either 'distinct' or 'object'"
+        else: "where T is 'distinct' or 'object'"
```
```suggestion
        if r.sym.magic == mDeepCopy:
          "where T is 'ptr' or 'ref' of either 'distinct' or 'object'"
        else:
          "where T is 'distinct' or 'object'"
```
To get the line width below 80 character.

compiler/front/cli_reporter.nim
```diff
@@ -1600,7 +1600,18 @@ proc reportBody*(conf: ConfigRef, r: SemReport): string =
...
+    
+      result = "'$1' must satisfy the signature '$2' $3" % [
+              r.str, typ.render, msg]
```
```suggestion
      result = "'$1' must satisfy the signature '$2' $3" %
               [r.str, typ.render, msg]
```
Super minor, but the other code in this module generally puts the `[` on the next line when the remainder doesn't fit.

compiler/sem/semstmts.nim
```diff
@@ -2367,8 +2369,13 @@ proc semOverride(c: PContext, s: PSym, info: TLineInfo) =
             rsemInseparableTypeBoundOp, obj, sym = s))
         return
     if sfSystemModule notin s.owner.flags:
+      let op =
+        if name == "=": attachedAsgn
+        else: AttachedOpToStr.find(name).TTypeAttachedOp
```
zerbina: This duplicates what `let k = ...` does above. I'd suggest moving `k` to after the `if s.magic == mAsgn: return` statement, so that you can re-use its value here.

compiler/sem/semstmts.nim
```diff
@@ -2298,7 +2298,8 @@ proc semOverride(c: PContext, s: PSym, info: TLineInfo) =
             rsemInseparableTypeBoundOp, obj, sym = s))
     if not noError and sfSystemModule notin s.owner.flags:
       localReport(c.config, info, reportSym(
-        rsemUnexpectedTypeBoundOpSignature, s))
+        rsemUnexpectedTypeBoundOpSignature, getSysMagic(c.graph, info,
+        s.name.s.normalize, AttachedOpToMagic[op]), str = s.name.s))
```
```suggestion
        rsemUnexpectedTypeBoundOpSignature,
        getSysMagic(c.graph, info, AttachedOpToStr[op], AttachedOpToMagic[op]),
        str = s.name.s))
```

compiler/sem/semstmts.nim
```diff
@@ -2328,7 +2329,8 @@ proc semOverride(c: PContext, s: PSym, info: TLineInfo) =
           rsemInseparableTypeBoundOp, t, sym = s))
     else:
       localReport(c.config, info, reportSym(
-        rsemUnexpectedTypeBoundOpSignature, s))
+        rsemUnexpectedTypeBoundOpSignature, getSysMagic(c.graph, info,
+        "deepcopy", mDeepCopy), str = s.name.s))
```
```suggestion
        rsemUnexpectedTypeBoundOpSignature,
        getSysMagic(c.graph, info, "deepcopy", mDeepCopy),
        str = s.name.s))
```
The line break in the middle of the call and the missing extra indentation makes it read like `"deepcopy"`, `mDeepCopy`, etc. are part of the `reportSym` call.

compiler/sem/semstmts.nim
```diff
@@ -2367,8 +2369,13 @@ proc semOverride(c: PContext, s: PSym, info: TLineInfo) =
...
         return
     if sfSystemModule notin s.owner.flags:
+      let op =
+        if name == "=": attachedAsgn
+        else: AttachedOpToStr.find(name).TTypeAttachedOp
+
       localReport(c.config, info, reportSym(
-        rsemUnexpectedTypeBoundOpSignature, s))
+        rsemUnexpectedTypeBoundOpSignature, getSysMagic(
+        c.graph, info, name, AttachedOpToMagic[op]), str = s.name.s))
```
```suggestion
        rsemUnexpectedTypeBoundOpSignature,
        getSysMagic(c.graph, info, AttachedOpToStr[op], AttachedOpToMagic[op]),
        str = s.name.s))
```
Better use the name from the table (it's more future-proof).

zerbina (changes requested):
Nice work! Conceptually, the changes look good to me.

I left a few comments regarding code style, and there's also a duplicated computation. Once those are addressed the PR is good-to-be-merged from my side.

compiler/sem/semstmts.nim
Luyten-Orion: This has been done, and `k` is now `hookOp` for clarity.

zerbina (approved):

alaviss (approved):
Thanks for the good work, @Luyten-Orion

PR #1543, opened 2025-05-11
add static interfaces in the form of signature types

zerbina:
Signature types are meant to provide static interfaces, which would be a first step towards type-checked generic routines and (maybe) eventually removing `.borrow`.

**The name `signature` is a working title.** It's not yet clear whether the feature will take over `concept` or stay as its own entity, and thus the `concept` syntax is simply reused.

**The current specification is not meant to represent the final form.** It's a version of the feature as discussed so far that's narrowed-down as much as possible in order to make a robust implementation feasible/possible with the current compiler. Further restrictions due to issues discovered during implementation are likely.

The feature was inspired by the `signature` feature of Standard ML.

---

## To-Do
- [x] ~add a signature pragma~
- [x] add dedicated `TType` kinds for signature types and their applied form 
- [ ] implement signature body analysis (partially done)
- [x] implement argument-dependent lookup (for signature types)
- [x] implement signature application (will be based on the existing borrow mechanism, most likely)
- [ ] add tests


doc/manual.rst
```diff
@@ -4166,6 +4183,85 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+The `concept` keyword must be followed by an identifier, which is the name of
+the placeholder used within the body to refer to the type that is described.
+From this point onwards, this identifier is referred to as the `placeholder`.
+
+The body must be a non-empty list of routine *declarations*. Only `func`:idx:,
+`proc`:idx:, `iterator`:idx:, and `method`:idx: are allowed, `template`:idx:
+and `macro`:idx: are not.
+
+For every routine declaration in the body, at least one parameter must be of
+type `placholder`, `var placeholder`, or `sink placeholder`. Declarations must
```
beef331: `placholder` -> `placeholder`

doc/manual.rst
```diff
@@ -4166,6 +4183,85 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+* (only if `tags` is provided for `x`) the tags of `x` must be a superset of
+  those of `y`
+* (only if `raises` is provided for `x`) the raised exceptions must be a
+  superset of the one of `y`
+
+If `y` is generic, it is instantiated first. If there are either none, or two
+or more candidates that satisfy the requirements listed above, application
+fails. If sucessful, the resulting application is said to *bind* to `y`.
+
+Two applied signatures are equal are equal iff both are applications of the
```
beef331: `iff` -> `if`

doc/manual.rst
```diff
@@ -4166,6 +4183,85 @@ For the above `openArray` and `var` rules, whether the expressions refers to
...
+  `A` is expected
+* `A` is l-value convertible to `T`
+
+Application of signature type `S` happens when a value of type `T` is
+implicitly or explicitly converted to `S`.
+
+On application of `S` to `T`, for every routine declaration `x` of `S`, a
+routine `y` must be *visible* in the current context such that:
+* `x` and `y` have the same name
+* after all `placholder`s in the parameter list of `x` are replaced with `T`,
```
beef331: `placholder` -> `placeholder`

doc/manual.rst
zerbina: The spelling is deliberate. "iff" is a short form of "if and only if" ([source](https://en.wikipedia.org/wiki/If_and_only_if)).

beef331: Perhaps it's not shorter if you have to provide this message along with all reading material :stuck_out_tongue:

zerbina: Well, `iff` is already used in other places of the manual.

Nonetheless, you're right in that the usage does make the text less accessible; I'll expand it.

PR #1545, opened 2025-05-13
lexopt: remove bundle special case in `remaining()`

alaviss:
## Summary
The  `remaining()`  iterator no longer considers the remainder of a
partially processed bundle as a value that has to be handled, and
instead produces a set of short options bundle.

## Details
The idea I had when this "feature" was implemented is to avoid value
synthesis in  `remaining()` , making sure that only the original input
is yielded. That way, you can distinguish between  `-pstuff`  from 
`-p -stuff`  when  `remaining()`  is used.

However,  `lexopt`  was designed with the premise that an user would
never have to distinguish between different kinds of inline value
representation, since "delimited" is the only supported type of inline
parameter. As such this feature is just an error that you cannot avoid.

This commit removes the special case, and synthesize a short option
bundle from the unprocessed portions of the active bundle instead.

### DIFF
--- a/lib/experimental/lexopt.nim
+++ b/lib/experimental/lexopt.nim
@@ -233,22 +233,17 @@ iterator remaining*(l: var CmdLexer): string =
   ##
   ## If the previous option returned from `next()` have a value that was not
   ## consumed by `value()`, `UnexpectedValueError` will be raised.
-  ##
-  ## As a special case, any short options within a bundle (ie. `-abcde`) that
-  ## have not been processed by `next()` is considered the value of the last
-  ## short option returned and will raise `UnexpectedValueError`.
   if l.valueIdx > 0:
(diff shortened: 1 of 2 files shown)


saem (approved):
Change looks good, but the summary could use an update, it's currently:

> Remove the special case in the remaining() iterator where the remainder of an in-progress bundle is also considered a value that has to be handled

I think this would read better:

> The `remaining()` iterator no longer considers the remainder of a partially processed bundle as a value that has to be handled, and instead produces a set of short options bundle.

The above summary gives a better idea of how the behaviour has changed, rather than focusing on the fact that it has.

alaviss:
/merge

Applied your recs, lets go

saem:
I'm too tired to debug the ci failure, but just in case it's transient, I kicked off another merge.

PR #1546, opened 2025-05-15
parseutils: make `parseBiggestFloat` non-magic

beef331:
## Summary

Make  `parseutils.parseBiggestFloat`  not rely on compiler magic,
enabling changing  `std/parseutils`  to use  `openArray[char]`  in the
future. Behaviour and interface stay the same.

## Details

There's no need for  `parseBiggestFloat`  to use compiler magic, so it's
changed to be a normal procedure, making it possible to change its
implementation and interface without touching the compiler.

The implementation is part of  `parseutils`  and dependent on the
backend:
* for the C backend, the implementation is copied verbatim from 
`strmantle.nimParseBiggestFloat` 
* for the JS backend, the implementation is copied verbatim from 
`jssys.nimParseBiggestFloat` 
* the VM implements the procedure via a VM hook

The now obsolete  `mParseBiggestFloat`  magic and the  `opcParseFloat` 
VM operation are removed.

For csources compatibility, the old magic procedure is still used when
the new   `nimskullHasNoParseFloatMagic`  conditional symbol is absent.
The related  `.compilerprocs`  in  `strmantle`  and  `jssys`  also have
to be kept.

### DIFF
--- a/compiler/front/condsyms.nim
+++ b/compiler/front/condsyms.nim
@@ -78,3 +78,4 @@ proc initDefines*(symbols: StringTableRef) =
   defineSymbol("nimskullNoNkStmtListTypeAndNkBlockType")
   defineSymbol("nimskullNoNkNone")
   defineSymbol("nimskullHasSupportsZeroMem")
+  defineSymbol("nimskullHasNoParseFloatMagic")
--- a/compiler/vm/vmops.nim
+++ b/compiler/vm/vmops.nim
@@ -65,6 +65,7 @@ from std/times import getTime
 from std/hashes import hash
 from std/osproc import nil
 from system/formatfloat import writeFloatToBufferSprintf
+from std/parseutils import parseBiggestFloat
 
 from compiler/modules/modulegraphs import `$`
 
@@ -352,6 +353,13 @@ iterator basicOps*(): Override =
     of 1: setResult(a, round(getFloat(a, 0)))
     of 2: setResult(a, round(getFloat(a, 0), getInt(a, 1).int))
     else: doAssert false, $n
+
+  override "stdlib.parseutils.parseBiggestFloat", proc(a: VmArgs) {.nimcall.} =
+    var num: BiggestFloat
+    copyMem(num.addr, a.getHandle(1).rawPointer, sizeof(num))
+    let parsed = a.getString(0).parseBiggestFloat(num, int a.getInt(2))
+    copyMem(a.getHandle(1).rawPointer, num.addr, sizeof(num))
+    a.setResult(parsed)
 
   wrap1s(getMD5, md5op)
 
--- a/lib/pure/parseutils.nim
+++ b/lib/pure/parseutils.nim
@@ -569,11 +569,233 @@ proc parseUInt*(s: string, number: var uint, start = 0): int {.
   if result != 0:
     number = uint(res)
 
-proc parseBiggestFloat*(s: string, number: var BiggestFloat, start = 0): int {.
-  magic: "ParseBiggestFloat", importc: "nimParseBiggestFloat", noSideEffect.}
-  ## Parses a float starting at `start` and stores the value into `number`.
-  ## Result is the number of processed chars or 0 if a parsing error
-  ## occurred.
+
+when not defined(nimskullHasNoParseFloatMagic):
+  proc parseBiggestFloat*(s: string, number: var BiggestFloat, start = 0): int {.
+    magic: "ParseBiggestFloat", importc: "nimParseBiggestFloat", noSideEffect.}
+elif defined(js):
+  proc parseFloatNative(a: string): float =
+    let a2 = a.cstring
+    asm """
+    `result` = Number(`a2`);
+    """
+
+  proc parseBiggestFloat*(s: string, number: var BiggestFloat, start: int): int {.noSideEffect.} =
+    var sign: bool
+    var i = start
+    if s[i] == '+': inc(i)
+    elif s[i] == '-':
+      sign = true
+      inc(i)
+    if s[i] == 'N' or s[i] == 'n':
+      if s[i+1] == 'A' or s[i+1] == 'a':
+        if s[i+2] == 'N' or s[i+2] == 'n':
+          if s[i+3] notin IdentChars:
+            number = NaN
+            return i+3 - start
+      return 0
+    if s[i] == 'I' or s[i] == 'i':
+      if s[i+1] == 'N' or s[i+1] == 'n':
+        if s[i+2] == 'F' or s[i+2] == 'f':
+          if s[i+3] notin IdentChars:
+            number = if sign: -Inf else: Inf
+            return i+3 - start
+      return 0
+
+    var buf: string
+      # we could also use an `array[char, N]` buffer to avoid reallocs, or
+      # use a 2-pass algorithm that first computes the length.
+    if sign: buf.add '-'
+    template addInc =
+      buf.add s[i]
+      inc(i)
+    template eatUnderscores =
+      while i in 0..s.high and s[i] == '_': inc(i)
+    while i in 0..s.high and s[i] in {'0'..'9'}: # Read integer part
+      buf.add s[i]
+      inc(i)
+      eatUnderscores()
+    if i in 0..s.high and s[i] == '.': # Decimal?
+      addInc()
+      while i in 0..s.high and s[i] in {'0'..'9'}: # Read fractional part
+        addInc()
+        eatUnderscores()
+    # Again, read integer and fractional part
+    if buf.len == ord(sign): return 0
+    if i in 0..s.high and s[i] in {'e', 'E'}: # Exponent?
+      addInc()
+      if s[i] == '+': inc(i)
+      elif s[i] == '-': addInc()
+      if s[i] notin {'0'..'9'}: return 0
+      while s[i] in {'0'..'9'}:
+        addInc()
+        eatUnderscores()
+    number = parseFloatNative(buf)
+    result = i - start
+else:
+  proc c_strtod(buf: cstring, endptr: ptr cstring): float64 {.importc: "strtod", header: "<stdlib.h>", noSideEffect.}
+  const powtens = [1e0, 1e1, 1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8, 1e9,
+              1e10, 1e11, 1e12, 1e13, 1e14, 1e15, 1e16, 1e17, 1e18, 1e19,
+              1e20, 1e21, 1e22]
+
+  proc parseBiggestFloat*(s: string, number: var BiggestFloat,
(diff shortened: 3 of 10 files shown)


compiler/vm/vmops.nim
```diff
@@ -353,6 +354,14 @@ iterator basicOps*(): Override =
     of 2: setResult(a, round(getFloat(a, 0), getInt(a, 1).int))
     else: doAssert false, $n
 
+  override "stdlib.parseutils.parseBiggestFloat", proc(a: VmArgs) {.nimcall.} =
+    var num: BiggestFloat
+    copyMem(num.addr, pointer(a.getHandle(1).p), sizeof(BiggestFloat))
+    let parsed = a.getString(0).parseBiggestFloat(num, int a.getInt(2))
+    a.setFloat(0, num)
+    a.setResult(parsed)
+
```
zerbina:
```suggestion
    var num: BiggestFloat
    copyMem(num.addr, a.getHandle(1).rawPointer, sizeof(BiggestFloat))
    let parsed = a.getString(0).parseBiggestFloat(num, int a.getInt(2))
    copyMem(a.getHandle(1).rawPointer, num.addr, sizeof(BiggestFloat))
    setResult(a, parsed)
```

It's almost correct, but instead of using `setFloat` (which just turns the argument register into a float register with the given value), you need to copy the float value to the location.

For making sure the handle has the correct type, you'd use `a.getHandle(1).typ.kind == akFloat and a.getHandle(1).typ.sizeInBytes == sizeof(BiggestFloat)`, but the general assumption in `vmops` - for better or worse - is that the inputs are all properly typed, so I'd personally leave it out.

zerbina:
Oh, my bad, I didn't think through the usage of `nimKochBootstrap` properly -- it's not applicable in this case.

Fortunately, the solution is simple: add a new conditional symbol (something like `nimskullHasNoParseFloatMagic`) in `condsyms.nim`, and then check for its absence in place of the `defined(nimKochBootstrap)`. That way, the magic will only be used when an old compiler is used.

compiler/vm/vmops.nim
```diff
@@ -353,6 +354,14 @@ iterator basicOps*(): Override =
     of 2: setResult(a, round(getFloat(a, 0), getInt(a, 1).int))
     else: doAssert false, $n
 
+  override "stdlib.parseutils.parseBiggestFloat", proc(a: VmArgs) {.nimcall.} =
+    var num: BiggestFloat
+    copyMem(num.addr, a.getHandle(1).rawPointer, sizeof(num))
+    let parsed = a.getString(0).parseBiggestFloat(num, int a.getInt(2))
+    copyMem(a.getHandle(1).rawPointer, num.addr, sizeof(num))
+    a.setResult(parsed)
+
```
zerbina, line 363:
```suggestion
```

lib/pure/parseutils.nim
```diff
@@ -569,11 +569,229 @@ proc parseUInt*(s: string, number: var uint, start = 0): int {.
   if result != 0:
     number = uint(res)
 
-proc parseBiggestFloat*(s: string, number: var BiggestFloat, start = 0): int {.
-  magic: "ParseBiggestFloat", importc: "nimParseBiggestFloat", noSideEffect.}
-  ## Parses a float starting at `start` and stores the value into `number`.
-  ## Result is the number of processed chars or 0 if a parsing error
-  ## occurred.
+
```
zerbina, line 572:
```suggestion
```

lib/pure/parseutils.nim
```diff
@@ -569,11 +569,229 @@ proc parseUInt*(s: string, number: var uint, start = 0): int {.
...
+when defined(nimKochBootStrap):
+  proc parseBiggestFloat*(s: string, number: var BiggestFloat, start = 0): int {.
+    magic: "ParseBiggestFloat", importc: "nimParseBiggestFloat", noSideEffect.}
+else:
+  when defined(js):
```
zerbina: Please use an `elif` here. It reduces the base indentation and also means that no de-indentation is necessary once the above guard is removed.

lib/pure/parseutils.nim
```diff
@@ -569,11 +569,229 @@ proc parseUInt*(s: string, number: var uint, start = 0): int {.
...
+    proc parseFloatNative(a: string): float =
+      let a2 = a.cstring
+      asm """
+      `result` = Number(`a2`);
+      """
+
+    #[
+    xxx how come code like this doesn't give IndexDefect ?
+    let z = s[10000] == 'a'
+    ]#
```
zerbina:
```suggestion
```
I know you the comment copied from `jssys`, but it belongs neither here nor there.

lib/pure/parseutils.nim
```diff
@@ -569,11 +569,229 @@ proc parseUInt*(s: string, number: var uint, start = 0): int {.
...
+    proc parseBiggestFloat*(s: string, number: var BiggestFloat, start: int): int {.noSideEffect.} =
```
zerbina:
```suggestion
    proc parseBiggestFloat*(s: string, number: var BiggestFloat,
                            start: int): int {.noSideEffect.} =
```

lib/pure/parseutils.nim
```diff
@@ -569,11 +569,229 @@ proc parseUInt*(s: string, number: var uint, start = 0): int {.
...
+  else:
+    proc c_strtod(buf: cstring, endptr: ptr cstring): float64 {.importc: "strtod", header: "<stdlib.h>", noSideEffect.}
+    const powtens = [1e0, 1e1, 1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8, 1e9,
+                1e10, 1e11, 1e12, 1e13, 1e14, 1e15, 1e16, 1e17, 1e18, 1e19,
+                1e20, 1e21, 1e22]
+    proc parseBiggestFloat*(s: string, number: var BiggestFloat, start = 0): int {.noSideEffect.} =
+      ## Parses a float starting at `start` and stores the value into `number`.
+      ## Result is the number of processed chars or 0 if a parsing error
+      ## occurred.
+      var
```
zerbina: Please also copy the comment about the implementation from `strmantle`.

zerbina (changes requested):
This is coming along nicely. I've left a few small requests regarding style and comments.

Please also remove the `opcParseFloat` VM opcode, together with its implementation.

lib/pure/parseutils.nim
```diff
@@ -569,11 +569,233 @@ proc parseUInt*(s: string, number: var uint, start = 0): int {.
...
+  proc parseBiggestFloat*(s: string, number: var BiggestFloat, start = 0): int {.
+    magic: "ParseBiggestFloat", importc: "nimParseBiggestFloat", noSideEffect.}
+elif defined(js):
+  proc parseFloatNative(a: string): float =
+    let a2 = a.cstring
+    asm """
+    `result` = Number(`a2`);
+    """
+
+  proc parseBiggestFloat*(s: string, number: var BiggestFloat, start: int): int {.noSideEffect.} =
```
zerbina:
```suggestion
  proc parseBiggestFloat*(s: string, number: var BiggestFloat,
                          start = 0): int {.noSideEffect.} =
```

lib/pure/parseutils.nim
```diff
@@ -569,11 +569,233 @@ proc parseUInt*(s: string, number: var uint, start = 0): int {.
...
+      if s[i] == '+': inc(i)
+      elif s[i] == '-': addInc()
+      if s[i] notin {'0'..'9'}: return 0
+      while s[i] in {'0'..'9'}:
+        addInc()
+        eatUnderscores()
+    number = parseFloatNative(buf)
+    result = i - start
+else:
+  proc c_strtod(buf: cstring, endptr: ptr cstring): float64 {.importc: "strtod", header: "<stdlib.h>", noSideEffect.}
```
zerbina:
```suggestion
  proc c_strtod(buf: cstring, endptr: ptr cstring): float64 {.
    importc: "strtod", header: "<stdlib.h>", noSideEffect.}
```

lib/pure/parseutils.nim
```diff
@@ -569,11 +569,233 @@ proc parseUInt*(s: string, number: var uint, start = 0): int {.
...
+  const powtens = [1e0, 1e1, 1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8, 1e9,
+              1e10, 1e11, 1e12, 1e13, 1e14, 1e15, 1e16, 1e17, 1e18, 1e19,
+              1e20, 1e21, 1e22]
```
zerbina:
```suggestion
  const powtens = [1e0, 1e1, 1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8, 1e9,
                   1e10, 1e11, 1e12, 1e13, 1e14, 1e15, 1e16, 1e17, 1e18, 1e19,
                   1e20, 1e21, 1e22]
```

compiler/front/condsyms.nim
```diff
@@ -78,3 +78,4 @@ proc initDefines*(symbols: StringTableRef) =
   defineSymbol("nimskullNoNkStmtListTypeAndNkBlockType")
   defineSymbol("nimskullNoNkNone")
   defineSymbol("nimskullHasSupportsZeroMem")
+  defineSymbol("nimskullHasNoParseFloatMagic"
```
zerbina:
```suggestion
  defineSymbol("nimskullHasNoParseFloatMagic")
```

zerbina (changes requested):
There are a few code style issues I missed, and the `parseBiggestFloat` implementation for the JS backend is missing the default value for `start`, but beyond that, the changes themselves look good to me.

lib/pure/parseutils.nim
```diff
@@ -580,7 +580,9 @@ elif defined(js):
     `result` = Number(`a2`);
     """
 
-  proc parseBiggestFloat*(s: string, number: var BiggestFloat, start: int): int {.noSideEffect.} =
+  {.checks: off.} # TODO: Make the parsing work with checks enabled
```
zerbina:
```suggestion
  {.push checks: off.} # TODO: Make the parsing work with checks enabled
```

lib/pure/parseutils.nim
```diff
@@ -632,11 +634,13 @@ elif defined(js):
         eatUnderscores()
     number = parseFloatNative(buf)
     result = i - start
+  {.checks: on.}
```
zerbina:
```suggestion
  {.pop.}
```

zerbina (changes requested):
There's one last issue with how checks are enabled/disabled. Right now, all run-time checks are unconditionally enabled after analysis of the `parseBiggestFloat` procedure.

zerbina (approved):
Good work.

Beyond allowing for `openArray` support, the procedure not being a magic makes changing to use a single implementation across all backends easier.

PR #1547, opened 2025-05-20
io: only define stdmsg when stdout/err are defined

alaviss:
## Summary
Makes sure that  `stdmsg`  is only defined when  `stdout`  and  `stderr` 
are defined.

## Details
For JS/VM backends,  `stderr`  and  `stdout`  don't exist, making 
`stdmsg`  unusable. However, feature detection via 
`when declared(stdmsg)`  would not be able to recognize this as 
`stdmsg`  remains defined even in those scenarios.

This PR moves  `stdmsg`  definition to be under the same conditionals
that defines  `stdout` / `stderr` , making sure that it is only defined
when it can be used.

### DIFF
--- a/lib/system/io.nim
+++ b/lib/system/io.nim
@@ -58,12 +58,12 @@ when not isNimVmTarget and not defined(js):
     stderr* {.importc: stderrName, header: "<stdio.h>".}: File
       ## The standard error stream.
 
-when defined(useStdoutAsStdmsg):
-  template stdmsg*: File = stdout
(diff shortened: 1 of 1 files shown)


zerbina (approved):
`stdin`, `stdout`, and `stderr` need to be available on the VM and JS backends too, so that the `when` workarounds are not needed anymore, but given that they aren't right now, it makes sense to also not define `stdmsg` for the JS and VM backends.

PR #1550, opened 2025-05-27
Add @include directive to nimconf

quantimnot:
## Summary
* Add @include directive to include other nimconf files into the current scope
* Cleanup and consolidate nimconf tests
---

In Nim, I use a modular nimscript setup for my projects. I really wish nimscript was still part of Nimskull. Not having it was a hurdle and turnoff to me, and probably others.

A while back, I improved the nimconf impl to at least give me some modularity. Here it is. Also, I found and fixed some crash[es], but I don't recall the details.


zerbina:
Thank you for the PR, @quantimnot. An include feature for configuration files seems useful, especially for projects structured as a monorepo.

What's the status on this? Is it ready for review? Or is there work still remaining?

> I really wish nimscript was still part of Nimskull. Not having it was a hurdle and turnoff to me, and probably others.

There were many problems with NimScript-based configuration, including slowness and general implementation issues, and since it also didn't fit in with the direction w.r.t. to the build process, it was removed.

It'll will most likely come back one day, though in a somewhat different form, where the script is driving the compiler directly, instead of just configuring some settings.

quantimnot:
Status? It was a lot of code ago when I did it. I think it is complete enough. I recall growing aggravated with finding more original design/impl bugs and having to stop using my time on it. It works, so I wanted to see if it would be useful to others. However, over that last couple of days, the thought occurred to me that maybe the best thing for Nimskull is to waste no time on old cruft, but implement views, cps -> implement new general parsing/... modules -> new config system.

If you think it is worth your precious time, then I think it is an improvement on the current state. Also, for some reason of which I can't recall, I added Nimscript support to your Testament in the branch. I can PR that if you think it's of value.

quantimnot:
That's a lot of red. I think it's failing because I left the nimscript test rewrites in there, but I removed the testament changes. I will address this only if say you want this :)

zerbina:
Thank you for the info, and sorry for the delayed response.

While I - personally - don't have an immediate use for includes in configuration files, I think others might, and from a quick skim, the changes also seem small and local enough to warrant the feature.

To be clear, I'd would merge the `@include` feature, though it's possible that there will be some change requests w.r.t. the implementation (I cannot say for certain as I haven't looked at it in-depth, yet).

I know that this is not a very strong expression of "want", and if it's not enough to motivate you, that's totally fine and I'd very much understand.

> Also, for some reason of which I can't recall, I added Nimscript support to your Testament in the branch. I can PR that if you think it's of value.

I don't think it's needed. NimScript has very little remaining uses, having been made obsolete by the dedicated VM backend for the most part, and thus doesn't warrant `testament` support.

saem:
To add a data point, I know a few folks, myself included, that would like to have an `@include` facility.

PR #1554, opened 2025-06-12
lambdalifting: omit unnecessary env creation

zerbina:
## Summary

* improve call performance for inner routines that call closure
  routines but don't have closed-over locals themselves
* improve performance in general for closures where some parent
  routine(s) don't have closed-over locals

## Details

Previously, a local closure environment was always allocated on entry
of a routine A that called routines closing over some locals outside of
A. If such routine doesn't have closed-over locals itself, the closure
environment only stored an up-pointer, thus being unnecessary.

An example of such a case:
```nim
proc top() =
  var x = 0
  proc inner() = # <- an closure environment was allocate here
    proc inner2() =
      x = 1
    inner2()
  ...
```

Lambda-lifting is changed to never create a local closure environment
for closure procedures that have no closed-over locals. Instead, the
environment parameter is passed to the inner closures directly.

To make the above change easier, the lambda-lifting pass is changed to
not return the env variable symbol anymore, instead creating a
preliminary `nkClosure` construction with the env symbol that the main
`transf` pass then completes.

### DIFF
--- a/compiler/sem/lambdalifting.nim
+++ b/compiler/sem/lambdalifting.nim
@@ -642,12 +642,7 @@ proc liftCapturedVars(n: PNode, graph: ModuleGraph, idgen: IdGenerator,
   ## environment field access, and usages of closure routines (except for
   ## iterators) are transformed into closure construction expressions.
   ##
-  ## In addition, the current enivornment (which can either be the root
-  ## routines local environment or some outer environment) is added as the
-  ## hidden parameter to each closure routine defined in the root
-  ## routine.
-  ##
-  ## `n` is assumed to be owned, and is thus modified in-place.
+  ## `n` is expected to be a production and is modified in-place.
   result = n
   case n.kind
   of nkSym:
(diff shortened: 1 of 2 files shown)


zerbina:
The motivation for this PR was to make code feasible where there are many deeply-nested inner routines in a procedure that all close over some top-level locals but (mostly) don't have captured locals themselves.

To provide a concrete example, consider:
```nim
proc module() =
  var x = 0 # top-level vars
  var y = 0

  proc a() =
    proc b() =
      proc c() =
        proc d() =
          x = 1
          y = 2
  ...
```
Previously, a new environment was created on every call to `a`, `b`, and `c`, incurring significant allocation / ref-counting overhead and also introducing many pointer indirections to the actual `x` and `y` access in `d` (i.e., the lowered access looked like this: `envP[].up[].up[].up[].x`).

Now, the above code is much closer w.r.t. performance to manually using an environment parameter, like so:
```nim
proc module() =
  type Env = object
    x: int
    y: int

  var env = Env(x: 0, y: 0)
  proc a(env: var Env) =
    proc b(env: var Env) =
      proc c(env: var Env) =
        proc d(env: var Env) =
          env.x = 1
          env.y = 1
  ...
```

saem (approved):

PR #1560, opened 2025-07-07
sempass2: fix bug with `.closure` inference

zerbina:
## Summary

Fix inference failure and the compiler subsequently crashing for
forwarded inner procedures and generic inner routines.

## Details

`.closure` inference for didn't consider forwarded routines properly
and generic owners were also not ignored, leading to some the affected
inner routines not being inferred as requiring the `.closure` calling
convention, which then led to an internal assertion failure during
lambda-lifting.

Still-forwarded routines don't have a body to analyze at the time of
usage, and thus have to be assumed to capture, unless an explicit
calling convention is specified that's not `.closure`.

### DIFF
--- a/compiler/sem/sempass2.nim
+++ b/compiler/sem/sempass2.nim
@@ -1642,23 +1642,28 @@ proc detectCapture(owner, top: PSym, n: PNode, marker: var IntSet): PNode =
     of skProc, skFunc, skIterator:
       # NOTE: a routine using the closure calling convention means that it
       # *may* captures something (it might not). A routine not using the
-      # closure calling convention means that it *can't* capture anything,
+      # closure calling convention means that it *cannot* capture anything,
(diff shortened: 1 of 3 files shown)


zerbina:
For anyone wanting to fix the problem with lambda lifting, I'd suggest the following approach:
1. add all free variables of a routine (i.e., locals whose definition is not part of the procedure) as hidden parameters in `sempass2`
2. replace the detection pass in `lambdalifting` (and `sempass2`) with just looking at the aforementioned hidden parameters of all used inner procedure
3. after the lambdalifting pass is done for a routine, remove the free variables added as hidden parameters

This would address the problem without using a type flag (like how it was done in the past), while simultaneously reducing the amount of scanning the compiler has to perform, as gathering the set of free variables (or detecting whether there are any) would only need to happen once.

saem (approved):

PR #1561, opened 2025-07-08
instantiate generics in the scope they're defined in

zerbina:
## Summary

Generic routines are now insantiated in their defined-in scope,
instead of always in the top-level scope, affecting mixin and other
late-binding symbol lookup.

## Details

All generic routines were previously instantiated in the top-level
scope, meaning that open identifiers couldn't bind to non-top-level
symbols, even if they're in the same scope the generic is part of.

`instantiateGeneric` now only unwinds the scope till the one the
generic was defined in. If the defined-in scope cannot be found (either
because the generic comes from a different module, or because it
escaped its scope via macro shenanigans), the current module's
top-level
scope is used for instantiation, same as before.

A test is added for the scoping behaviour and the existing `tbind.nim`
test is updated.


saem (approved):
Wow, this is one of those changes that after it's made, you kinda shake your head wondering why it didn't work that way in the first place.

PR #1562, opened 2025-07-10
stdlib: enumutils len calculates enum length

saem:
## Summary

Add a `len` routine to the `enumutils` module, which calculates the number of entries in an enum.

## Details

Besides adding the procedure, a corresponding test was also added to the existing `tenumutils`, and that test now also runs for the `vm` target instead of just `c` and `js`.

---

## Notes for Reviewers
* I'm tired so there might be a much better way to do this or preexisting functionality that I'm forgetting
* required for property based testing

### DIFF
--- a/lib/std/enumutils.nim
+++ b/lib/std/enumutils.nim
@@ -175,3 +175,7 @@ func symbolName*[T: enum](a: T): string =
     assert c1.symbolName == "c1"
   const names = enumNames(T)
   names[a.symbolRank]
+
+func len*[T: enum](E: typedesc[T]): int =
+  for e in E.items:
+    inc result
--- a/lib/std/enumutils.nim
+++ b/lib/std/enumutils.nim
@@ -176,6 +176,6 @@ func symbolName*[T: enum](a: T): string =
   const names = enumNames(T)
   names[a.symbolRank]
 
-func len*[T: enum](E: typedesc[T]): uint16 =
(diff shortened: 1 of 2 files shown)


lib/std/enumutils.nim
```diff
@@ -175,3 +175,7 @@ func symbolName*[T: enum](a: T): string =
     assert c1.symbolName == "c1"
   const names = enumNames(T)
   names[a.symbolRank]
+
+func len*[T: enum](E: typedesc[T]): uint16 =
```
saem: I might be remembering max `case` branches rather than enum size

saem: Checking `semtypes.semenum` we don't seem to have an upper limit for enums, if that's the case then we have a design bug with enums in that `case` has a branch limit, so it can't work exhaustively over all enums.

In the meantime, I think changing this to an `int` return type is probably the best bet.

alaviss (commented):
This already exists as [`typetraits.enumLen`](https://nim-works.github.io/nimskull/typetraits#enumLen.m%2Ctypedesc%5B%5D).

saem:
> This already exists as [`typetraits.enumLen`](https://nim-works.github.io/nimskull/typetraits#enumLen.m%2Ctypedesc%5B%5D).

Thanks.

Also lame that I didn't see that

PR #1563, opened 2025-07-17
mir: rename `mnkResume` to `mnkUnwind`

zerbina:
## Summary

Rename the MIR node kind `mnkResume` to `mnkUnwind`; behaviour doesn't
change.

## Details

The name "resume" is a good fit for a control-flow primitive needed by
delimited continuation support (for resuming after suspending/forking),
so `mnkResume` is renamed to `mnkUnwind` in order to make the name
"resume" available again.

Both the node kind as well as some routines related to it are renamed.
The pretty-printer also renders the unwind target as `Unwind`, and the
tests comparing textual MIR fragments are adjusted accordingly.

### DIFF
--- a/compiler/backend/cgirgen.nim
+++ b/compiler/backend/cgirgen.nim
@@ -348,7 +348,7 @@ proc targetToIr(tree: MirBody, cr: var TreeCursor): CgNode =
   case n.kind
   of mnkLabel:
     result = newLabelNode(n.label)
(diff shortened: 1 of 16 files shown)


saem (approved):
Change looks good, it'd be nice to have the reason for the rename (outside of continuation prep) in the message (but not a blocker).

PR #1565, opened 2025-07-20
ci(reproducible): add workaround for reprotest usage of sudo

alaviss:
## Summary
Configures sudo to workaround a bug in `reprotest`.

## Details
`sudo`  recent security fixes exposed a bug where reprotest was using
the program improperly, causing reproducible build tests to fail.

This commit configures  `sudo`  to allow  `reprotest`  to continue
working as before, as described in
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1108550

### DIFF
--- a/.github/workflows/reproducible.yml
+++ b/.github/workflows/reproducible.yml
@@ -56,6 +56,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


alaviss:
/merge

Let's hope this works

alaviss:
Doesn't work, unfortunately

PR #1566, opened 2025-07-20
ci(reproducible): disable sudo usage for reprotest

alaviss:
## Summary
Disables environment variations requiring `sudo` in `reprotest`.

## Details
`sudo` recent security fixes exposed a bug where reprotest was using
the program improperly, causing reproducible build tests to fail.

This commit disables all variations requiring `sudo` to workaround
the issue. This includes:

- `user_group`: Variations in build user/group. This can detect
  issues with non-normalized tar-archives, which we faced in the past.
  It's unfortunate, but for now we just have to be vigilant with
  changes to `niminst`.
- `domain_host`: Variations in build hostname. While it is possible to
  disable the use of `sudo`, this has always failed in CI. Since we
  don't use this information at all, it is rather safe to disable.

Ref: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1108550

### DIFF
--- a/.github/workflows/reproducible.yml
+++ b/.github/workflows/reproducible.yml
@@ -81,21 +81,22 @@ jobs:
       - id: reprotest
         name: Run reproducibility build
         run: |
+          set -euxo pipefail
+
           # Add a guest user for reprotest
           sudo useradd -m guest-builder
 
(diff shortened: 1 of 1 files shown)


alaviss:
/merge

Finally this worked.

PR #1567, opened 2025-07-21
net: do not reuse socket after failed connect

alaviss:
## Summary
Removes optimization in  `net.dial`  for reusing sockets for multiple 
`connect`  attempts.

## Details
The socket state after a failed connect attempt is unspecified in POSIX.
While on certain OS like Linux or Windows, the socket can be reused and
another  `connect`  attempt could be made, others will no longer accept
the socket for future operations. This was surfaced when macOS CI failed
sporadically when a test domain (example.com), failed to connect on
first try.

This commit removes the socket reuse mechanism entirely, given that this
is recommended by POSIX and Linux manuals for portable code, as well as
a lack of evidence that this optimization provides adequate performance
benefits to justify maintaining it for supported platforms.

Small improvements on resource cleanup and error messages are also
applied as part of this change.

### DIFF
--- a/lib/pure/net.nim
+++ b/lib/pure/net.nim
@@ -1925,52 +1925,39 @@ proc dial*(address: string, port: Port,
   let sockType = protocol.toSockType()
 
   let aiList = getAddrInfo(address, port, AF_UNSPEC, sockType, protocol)
+  defer: freeaddrinfo(aiList)
 
-  var fdPerDomain: array[low(Domain).ord..high(Domain).ord, SocketHandle]
-  for i in low(fdPerDomain)..high(fdPerDomain):
-    fdPerDomain[i] = osInvalidSocket
-  template closeUnusedFds(domainToKeep = -1) {.dirty.} =
-    for i, fd in fdPerDomain:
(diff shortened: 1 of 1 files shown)


lib/pure/net.nim
```diff
@@ -1925,52 +1925,39 @@ proc dial*(address: string, port: Port,
...
 
-  if success:
-    result = newSocket(lastFd, domain, sockType, protocol)
-  elif lastError != 0.OSErrorCode:
+  if lastError != 0.OSErrorCode:
+    # FIXME: This should be a collection of errors for each attempt, not the last one
     raiseOSError(lastError)
   else:
-    raise newException(IOError, "Couldn't resolve address: " & address)
+    raise newException(IOError, "No usable addresses found for: " & address)
```
alaviss, line 1960: Failure to resolve would have already been reported by `nativesockets.getAddrInfo`. POSIX also specifies that `getaddrinfo` would return at least one address on success, so we can only reach this place when all of them were addresses in unknown domains.

saem (approved):

PR #1568, opened 2025-07-26
mir: enable move analysis for `result`

zerbina:
## Summary

* support automatic moves from the high-level `result` variable
* omit implicit `result` variable initialization in some more cases
* add the MIR-level `return` statement, making all `result`
  usages explicit at that stage

## Details

The main motivation is to add a dedicated return statement to the
MIR -- the improved `result` variable analysis is a side-effect.

* add and document the `mnkReturn` MIR node kind (a statement)
* change `mirgen` to emit a MIR `return` statement at all
  procedure exits
* translate `mnkReturn` statements into CGIR gotos in `cgirgen`,
  instead of introducing a `cnkReturn` node kind. This allows CGIR body
  merging to continue working without change
* handle `mnkReturn` when computing an MIR data-flow graph

In procedures that return something, the `return` statement is required
to have the result local as the singular operand. The reason is that it
makes all result variable usage explicit in MIR code, which allows for
some follow-up simplifications/improvements:
* in `injectdestructors`, remove most of the special handling for the
  result. This enables move analysis for the result local

* remove the "leaves control-flow graph" condition for deciding whether
  the result local needs to be default initialized, which results in
  the initialization being omitted in procedures where the result local
  is only uninitialized when unwinding due to an exception

Tests inspecting `--showir` and `--expandArc` output are updated.

### DIFF
--- a/compiler/backend/cgirgen.nim
+++ b/compiler/backend/cgirgen.nim
@@ -65,6 +65,9 @@ type
       ## the list of all locals in the body, taken from the ``MirBody``.
       ## Only needed for updating the type for alias locals
 
+    returns: seq[CgNode]
+      ## goto statements to which the return label needs to be added
+
     inUnscoped: int
       ## whether the currently proceesed statement/expression is part of an
       ## unscoped control-flow context. Used to move definitions to the start
@@ -569,6 +572,12 @@ proc stmtToIr(tree: MirBody, env: MirEnv, cl: var TranslateCl,
     to cnkGotoStmt, targetToIr(tree, cr)
   of mnkLoop:
     to cnkLoopStmt, targetToIr(tree, cr)
+  of mnkReturn:
+    if n.len == 1:
+      skip(tree, cr) # skip the operand
+    let goto = newStmt(cnkGotoStmt, info)
+    cl.returns.add goto
+    stmts.add goto
   of mnkLoopJoin:
     to cnkLoopJoinStmt, targetToIr(tree, cr)
   of mnkJoin:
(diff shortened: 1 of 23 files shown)


zerbina:
Procedures such as the following benefit from the improved default-initialization of `result`:
```nim
type Large = object
  arr: array[128, int]

proc get() = discard
proc test(x: Large): Large =
  get()
  result = x
```
Previously `result` was default initialized in `test`, even though it's unnecessary. When `get` is complex enough or has a body not available in the C translation-unit, both gcc and clang were unable to optimize the `memset` away, at least in example shown above (see [here](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:1,endLineNumber:17,positionColumn:1,positionLineNumber:17,selectionStartColumn:1,selectionStartLineNumber:17,startColumn:1,startLineNumber:17),source:'%23include+%3Cstring.h%3E%0A%0Astruct+Large+%7B%0A++++int+arr%5B128%5D%3B%0A%7D%3B%0A%0Aint+get()%3B%0A%0Astruct+Large+square(struct+Large*+arr)+%7B%0A++++struct+Large+result%3B%0A++++memset(%26result,+0,+sizeof(result))%3B%0A++++if+(__builtin_expect(get()+%3D%3D+0,+0))+goto+BeforeRet_%3B%0A++++result+%3D+*arr%3B%0A++++BeforeRet_:%0A++++return+result%3B%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cclang_trunk,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-O3',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+clang+(trunk)+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)).

The pass for injecting default initialization for the result variable is not enabled for the JS and VM code generators, meaning that those backends are not impacted.

saem (approved):

PR #1569, opened 2025-07-28
mir: turn tail-call elimination into a standalone pass

zerbina:
## Summary

* turn the portable tail-call elimination into a standalone MIR pass
* fix a compiler crash when a tail-call is prevented from happening due
  to `result` variable cleanup

## Details

* remove the tail-call elimination from `mirgen`, including the
  `goTailCallElim` option. `mirgen` always emits `mnkTailCall`
  statements for tail calls now

* remove the now-obsolete `mEnsureNoDestroy` magic. Detecting tail-
  call-preventing cleanup is anchored on `mnkTailCall` instead

* change the MIR rules to allow the zero-form `return` following a
  tail-call. A MIR tail-call is always followed by a `return` now,
  making tail-call elimination a bit simpler

* add the `tailcall_elim` module implementing the portable tail-call
  elimination. The transformation produces the same code as the
  previous implementation, except for:

  * translate `.tailcall` proc types the same as other non-`.closure`
     proc types

  * register the 'result' variable in a hidden scope during MIR
     translation, so that emitting the cleanup logic is handled
     automatically

  * remove the indirect .tailcall call workarounds from `ccgcalls`
     and `jsgen`

  * use precise source/origin information for all references to
locations
     created with `mirgen.genLocation`, fixing a crash in `injecthooks`
     due to the origin of `destroy` statements not pointing to
     symbol nodes

* add a test for tail calls being prevented due to the `result`
  variable requiring cleanup

The tail-call elimination passes happens after the `injectdestructors`
passes. It's only enabled for the C and JS code generators, same as
before.

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -853,9 +853,6 @@ type
     mStoreParams
       ## storeParams(p, tup): savely stores the tuple in the storage pointed
       ## to by `p`
-    mEnsureNoCleanup
-      ## destructor calls following an ensureNoCleanup call result in a
-      ## compiler error
 
 # things that we can evaluate safely at compile time, even if not asked for it:
 const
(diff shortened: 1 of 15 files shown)


saem:
FYI: updated the PR message so it's easier to see which items relate to the changes in code generation

saem (approved):

PR #1572, opened 2025-08-03
transf: fix indirect tail calls not being detected as such

zerbina:
## Summary

Fix calls where the callee is a `.tailcall` procval not being
considered a tail call when the callee type is a generic type
instantiation.

## Details

Type skipping in `transf` was missing for the relevant "is callee a
`.tailcall` proc type" checks, leaving such calls as normal calls.

A MIR-output-based test is added to ensure the calls in question being
tail calls.

### DIFF
--- a/compiler/sem/transf.nim
+++ b/compiler/sem/transf.nim
@@ -1134,7 +1134,8 @@ proc transformCall(c: PTransf, n: PNode): PNode =
     else:
       result = s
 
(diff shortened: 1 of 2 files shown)


saem (approved):

saem:
/merge

FYI: I made a slight change to the PR description due to what I think was a typo, the meaning should be preserved

PR #1577, opened 2025-08-12
fix type inference from expressions involving `sink` params

zerbina:
## Summary

* fix numeric literal types and `sink`-modified types being returned
  by `typeof`
* fix tuple elements being inferred as `sink`-modified types
* fix parameters being inferred as `sink`-modified types
* fix conversions-to-self where the conversion source operand is a
  sink parameter not being reported as pointless conversions

## Details

`sink` is a modifier for parameters. It's only interaction with the
type system is that it distinguishes procedure parameters from each
other (`proc(x: sink int)` and `proc(x: int)` are different types).

It must not be part of the types computed for expressions, at least not
visibly, meaning that types introduced for expression must never
contain `sink` types, which wasn't always the case so far.

The `principalType` procedure is introduced, for computing the
principal type for an expression in a usage context in a single place.
It's used for:
* the input operand of conversion checking (didn't handle `sink`)
* the var/let/const type inference (nothing changes)
* array element type inference (nothing changes)
* tuple element type inference (didn't handle `sink`)
* set element type inference (nothing changes)
* the type returned by `typeof` (handled neither `sink` nor numeric
  literal types)
* default parameter type inference (handled neither sink nor
  numeric literal types)

Set type inference is changed to no longer skip the instance/alias
types, which should result in slightly better error messages.

A test is added covering the inference cases described above (except
conversion checking) -- the already-existing test for array type
inference is merged into it.


saem (approved):

saem:
Amazing what a little bit of consistent applications of rules does, maybe I should attempt putting `semTypeNode` in more places one of these days.

saem:
Currently blocked from merging this until this PR is merged: https://github.com/nim-works/nimskull/pull/1578

zerbina:
@saem Thank you for taking care of merging the upstream branch.

PR #1578, opened 2025-08-13
ci(windows): bump vcpkg revision to 2025.07.25

alaviss:
## Summary
Bump the vcpkg revision used to supply Windows libraries.

## Details
The current revision no longer compiles in CI due to external sources no
longer hosting the required binaries. Bump the revision to the latest to
fix the problem.


saem:
/merge

Thanks, this should unblock https://github.com/nim-works/nimskull/pull/1577

saem (approved):

PR #1579, opened 2025-08-13
ci: don't run on dependabot branches

alaviss:
## Summary
Turn `push` CI off for dependabot branches

## Details
As dependabot creates new PRs, it also pushes into a branch in our
repository. This causes both PR and push CI to run at the same time,
saturating our CI capacity.

This PR disables the redundant push CI for branches created by
dependabot.


saem (approved):
yay, less strain

PR #1582, opened 2025-08-14
lang: add "generated" symbol kind

### LINKED ISSUE #1539: compiler crash: `unittest.check` macro causes segfault in certain scenarios
### Specification

This bug causes the compiler to segfault in certain scenarios, I can't figure out why but a minimal repro has been created.

### Example

(Haven't minimised the code)
https://github.com/Luyten-Orion/Tomlys/tree/b47fc389ed6b5768eb7c42c36d6281cea314148e

Minimised code:
```nim
import std/[unittest]

type
  Test = object
    aF: int

proc a(t: Test): int = t.aF

check Test(a: 1).a == Test(a: 1).a
```

### Actual Output

I've compiled the code with `nim_dbg` and this is the result.
https://gist.github.com/Luyten-Orion/3f27be90040f108d172e1f78cf57e837

Minimised repro's nim_dbg output:
```nim
...................................................................................................Traceback (most recent call last)
nim.nim                  nim
sem.nim                  semStmtAndGenerateGenerics nkStmtList 1600385 /home/chronos/test/main.nim(9, 6)
semexprs.nim             semExpr nkStmtList 1600385 /home/chronos/test/main.nim(9, 6)
semstmts.nim             semStmtList nkStmtList 1600385 /home/chronos/test/main.nim(9, 6)
semexprs.nim             semExpr nkCommand 1600367 /home/chronos/test/main.nim(9, 6)
semexprs.nim             semDirectOp
semexprs.nim             afterCallActions
sem.nim                  semMacroExpr nkCommand 1600389 /home/chronos/test/main.nim(9, 6)
sem.nim                  semAfterMacroCall nkBlockStmt 1601006 /home/chronos/.nimskull/lib/pure/unittest.nim(752, 6)
semexprs.nim             semExpr nkBlockStmt 1601006 /home/chronos/.nimskull/lib/pure/unittest.nim(752, 6)
semexprs.nim             semBlock nkBlockStmt 1601006 /home/chronos/.nimskull/lib/pure/unittest.nim(752, 6)
semexprs.nim             semExpr nkStmtList 1601008 /home/chronos/.nimskull/lib/pure/unittest.nim(753, 8)
semstmts.nim             semStmtList nkStmtList 1601008 /home/chronos/.nimskull/lib/pure/unittest.nim(753, 8)
semexprs.nim             semExpr nkIfStmt 1601011 /home/chronos/.nimskull/lib/pure/unittest.nim(754, 8)
semstmts.nim             semIf
semexprs.nim             semExprWithType nkPrefix 1601013 /home/chronos/.nimskull/lib/pure/unittest.nim(754, 11)
semexprs.nim             semExprCheck nkPrefix 1601013 /home/chronos/.nimskull/lib/pure/unittest.nim(754, 11)
semexprs.nim             semExpr nkPrefix 1601013 /home/chronos/.nimskull/lib/pure/unittest.nim(754, 11)
semexprs.nim             semMagic
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkPrefix 1601013 /home/chronos/.nimskull/lib/pure/unittest.nim(754, 11)
semcall.nim              semOverloadedCall nkPrefix 1601013 /home/chronos/.nimskull/lib/pure/unittest.nim(754, 11)
semcall.nim              resolveOverloads nkPrefix 1601013 /home/chronos/.nimskull/lib/pure/unittest.nim(754, 11)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkPrefix 1601013 /home/chronos/.nimskull/lib/pure/unittest.nim(754, 11)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand nkInfix 1600982 /home/chronos/test/main.nim(9, 19)
semexprs.nim             semOperand
semexprs.nim             semExpr nkInfix 1600982 /home/chronos/test/main.nim(9, 19)
semexprs.nim             semMagic
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkInfix 1600982 /home/chronos/test/main.nim(9, 19)
semcall.nim              semOverloadedCall nkInfix 1600982 /home/chronos/test/main.nim(9, 19)
semcall.nim              resolveOverloads nkInfix 1600982 /home/chronos/test/main.nim(9, 19)
semcall.nim              pickBestCandidate
sigmatch.nim             matches nkInfix 1600982 /home/chronos/test/main.nim(9, 19)
sigmatch.nim             matchesAux
sigmatch.nim             prepareOperand nkCall 1600984 /home/chronos/test/main.nim(9, 16)
semexprs.nim             semOperand
semexprs.nim             semExpr nkCall 1600984 /home/chronos/test/main.nim(9, 16)
semexprs.nim             semDirectOp
semexprs.nim             semOverloadedCallAnalyseEffects nkCall 1600984 /home/chronos/test/main.nim(9, 16)
semcall.nim              semOverloadedCall nkCall 1600984 /home/chronos/test/main.nim(9, 16)
semcall.nim              resolveOverloads nkCall 1600984 /home/chronos/test/main.nim(9, 16)
semcall.nim              pickBestCandidate
sigmatch.nim             initCallCandidate
ast_query.nim            originatingModule
```

### Expected Output

For it to not crash the compiler.

### Additional Information

I've modified `main.nim` and it seems that the `unittest.check` macro is involved, as replacing checks with asserts work fine.
```nim
#suite "combinator":
#  test "accept-match@success":
block:
  let res = accept('a').run(State.init("abc"))
  assert res.isOk
  assert res.unsafeGet.value == 'a'
  assert res.unsafeGet.state.idx == 1

#  test "accept-match@failure":
block:
  let res = accept('a').run(State.init("b"))
  assert res.isErr
  assert res.error.state.idx == 0
```

It seems to be related to when a function/template/macro is called on an object (including ref objects), it also requires initialisation. However in my original code I did not do that either, so not sure what's the exact cause.

alaviss:
## Summary
* Adds  `skGenerated`  symbol kind and diagnostics for
used-but-undeclared symbols.
*  `macros.genSym`  no longer takes a  `kind`  parameter and instead
will convert to the correct type on declaration.

## Details
Previously, generated symbols are specialized directly into their target
kind at generation site. This however throws a wrench into the
compiler's assumption, where specialized symbols are considered to be
well-formed and can be used as is.

This PR introduces a new  `skGenerated`  symbol type, which is used by
all generated symbols until their specialization (i.e. a declaration
statement happened for the symbol). This allows symbol usage sites
within the compiler to catch use-before-declaration errors and emit
appropriate diagnostics.

The automatic specialization also provided a UX improvement for macros
users, where  `genSym`  no longer requires a  `kind`  to be specified
and can automatically morph to the correct type once appropriate ASTs
are emitted. The old form of  `genSym`  has been marked as deprecated.

Fixes https://github.com/nim-works/nimskull/issues/1539

### DIFF
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -2707,15 +2707,19 @@ proc rawExecute(c: var TCtx, t: var VmThread, pc: var int): YieldReason =
         delSon(regs[ra].nimNode, bb)
     of opcGenSym:
       decodeBC(rkNimNode)
-      let k = regs[rb].intVal
-      checkHandle(regs[rc])
-      assert regs[rc].handle.typ.kind == akString
+      let rh =
+        if regs[rb].kind == rkInt:
+          # Called from compatibility symbol
+          rc
+        else:
+          rb
+
+      checkHandle(regs[rh])
+      assert regs[rh].handle.typ.kind == akString
       # XXX: costly stringify of strVal... `getIdent` doesn't use openArray :(
-      let name = if regs[rc].strVal.len == 0: ":tmp"
-                 else: $regs[rc].strVal
-      guestValidate(k in 0..ord(high(TSymKind)),
-        "request to create symbol of invalid kind")
-      var sym = newSym(k.TSymKind, getIdent(c.cache, name), nextSymId c.idgen, c.module.owner, c.debug[pc])
+      let name = if regs[rh].strVal.len == 0: ":tmp"
+                 else: $regs[rh].strVal
+      var sym = newSym(skGenerated, getIdent(c.cache, name), nextSymId c.idgen, nil, c.debug[pc])
       incl(sym.flags, sfGenSym)
       regs[ra].nimNode = newSymNode(sym)
     of opcNccValue:
--- a/compiler/vm/vm.nim
+++ b/compiler/vm/vm.nim
@@ -2706,19 +2706,12 @@ proc rawExecute(c: var TCtx, t: var VmThread, pc: var int): YieldReason =
       for i in 0..<regs[rc].intVal.int:
         delSon(regs[ra].nimNode, bb)
     of opcGenSym:
-      decodeBC(rkNimNode)
-      let rh =
-        if regs[rb].kind == rkInt:
-          # Called from compatibility symbol
-          rc
-        else:
-          rb
-
-      checkHandle(regs[rh])
-      assert regs[rh].handle.typ.kind == akString
+      decodeB(rkNimNode)
+      checkHandle(regs[rb])
+      assert regs[rb].handle.typ.kind == akString
       # XXX: costly stringify of strVal... `getIdent` doesn't use openArray :(
-      let name = if regs[rh].strVal.len == 0: ":tmp"
-                 else: $regs[rh].strVal
+      let name = if regs[rb].strVal.len == 0: ":tmp"
+                 else: $regs[rb].strVal
       var sym = newSym(skGenerated, getIdent(c.cache, name), nextSymId c.idgen, nil, c.debug[pc])
       incl(sym.flags, sfGenSym)
       regs[ra].nimNode = newSymNode(sym)
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -2003,7 +2003,13 @@ proc genMagic(c: var TCtx; n: CgNode; dest: var TDest; m: TMagic) =
   of mNCallSite:
     if dest.isUnset: dest = c.getTemp(n.typ)
     c.gABC(n, opcCallSite, dest)
-  of mNGenSym: genBinaryABC(c, n, dest, opcGenSym)
+  of mNGenSym:
+    if n.len == 2:
+      # Unary genSym
+      genUnaryABC(c, n, dest, opcGenSym)
+    else:
+      # Deprecated binary form
+      genBinaryABC(c, n, dest, opcGenSym)
   of mMinI, mMaxI, mAbsI, mDotDot:
     c.genCall(n, dest)
   of mExpandToAst:
--- a/compiler/vm/vmgen.nim
+++ b/compiler/vm/vmgen.nim
@@ -2003,13 +2003,7 @@ proc genMagic(c: var TCtx; n: CgNode; dest: var TDest; m: TMagic) =
   of mNCallSite:
     if dest.isUnset: dest = c.getTemp(n.typ)
     c.gABC(n, opcCallSite, dest)
-  of mNGenSym:
-    if n.len == 2:
-      # Unary genSym
-      genUnaryABC(c, n, dest, opcGenSym)
-    else:
-      # Deprecated binary form
-      genBinaryABC(c, n, dest, opcGenSym)
+  of mNGenSym: genUnaryABC(c, n, dest, opcGenSym)
   of mMinI, mMaxI, mAbsI, mDotDot:
     c.genCall(n, dest)
   of mExpandToAst:
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -438,13 +438,17 @@ when defined(nimskullHasUnaryGenSym):
     ## Generates a fresh symbol that is guaranteed to be unique. The symbol
     ## needs to occur in a declaration context.
 
-proc genSym*(kind: NimSymKind = nskLet; ident = ""): NimNode {.
-  magic: "NGenSym", noSideEffect, deprecated: "genSym no longer takes a `kind` parameter".}
-  ## Generates a fresh symbol that is guaranteed to be unique. The symbol
-  ## needs to occur in a declaration context.
-  ##
-  ## This is a compatibility alias for `genSym`, the `kind` parameter is
-  ## ignored.
+  proc genSym*(kind: NimSymKind = nskGenerated; ident: string): NimNode {.
+    magic: "NGenSym", noSideEffect, deprecated: "genSym no longer takes a `kind` parameter".}
+    ## Generates a fresh symbol that is guaranteed to be unique. The symbol
+    ## needs to occur in a declaration context.
+    ##
+    ## This is a compatibility alias for `genSym`, the `kind` parameter is
+    ## ignored.
+else:
+  # Old prototype for bootstrapping
+  proc genSym*(kind: NimSymKind = nskLet; ident = ""): NimNode {.
+    magic: "NGenSym", noSideEffect.}
 
 proc callsite*(): NimNode {.magic: "NCallSite", benign, deprecated:
   "Deprecated since v0.18.1; use `varargs[untyped]` in the macro prototype instead".}
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -145,8 +145,7 @@ type
     nskProc, nskFunc, nskMethod, nskIterator,
     nskConverter, nskMacro, nskTemplate, nskField,
     nskEnumField, nskForVar, nskLabel,
-    nskStub,
-    nskGenerated = ord(nskStub) + 2
+    nskStub, nskGenerated
 
 const
   nnkLiterals* = {nnkCharLit..nnkNilLit}
--- a/lib/core/macros.nim
+++ b/lib/core/macros.nim
@@ -437,13 +437,24 @@ when defined(nimskullHasUnaryGenSym):
     ## Generates a fresh symbol that is guaranteed to be unique. The symbol
     ## needs to occur in a declaration context.
 
-  proc genSym*(kind: NimSymKind = nskGenerated; ident: string): NimNode {.
-    magic: "NGenSym", noSideEffect, deprecated: "genSym no longer takes a `kind` parameter".}
+  template genSym*(kind: NimSymKind = nskGenerated; ident: string): NimNode {.
+    deprecated: "genSym no longer takes a `kind` parameter".} =
     ## Generates a fresh symbol that is guaranteed to be unique. The symbol
     ## needs to occur in a declaration context.
     ##
-    ## This is a compatibility alias for `genSym`, the `kind` parameter is
-    ## ignored.
+    ## This is a compatibility alias for `genSym <#genSym,string>`_, the `kind`
+    ## parameter is ignored.
+    genSym(ident)
+
+  template genSym*(kind: NimSymKind): NimNode {.
+    deprecated: "genSym no longer takes a `kind` parameter".} =
+    ## Generates a fresh symbol that is guaranteed to be unique. The symbol
+    ## needs to occur in a declaration context.
+    ##
+    ## This is a compatibility alias for `genSym <#genSym,string>`_, the `kind`
+    ## parameter is ignored.
+    genSym()
+
 else:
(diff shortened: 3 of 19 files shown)


tests/errmsgs/tundeclared_symbol.nim
```diff
@@ -0,0 +1,14 @@
+discard """
+  cmd: '''nim check --msgFormat:sexp --filenames:canonical --hints:off $file'''
+  action: reject
+  nimoutFormat: sexp
```
saem: Please don't use `sexp`, that whole thing needs to be removed.

compiler/vm/vmgen.nim
```diff
@@ -2003,7 +2003,13 @@ proc genMagic(c: var TCtx; n: CgNode; dest: var TDest; m: TMagic) =
...
     if dest.isUnset: dest = c.getTemp(n.typ)
     c.gABC(n, opcCallSite, dest)
-  of mNGenSym: genBinaryABC(c, n, dest, opcGenSym)
+  of mNGenSym:
+    if n.len == 2:
+      # Unary genSym
+      genUnaryABC(c, n, dest, opcGenSym)
+    else:
+      # Deprecated binary form
+      genBinaryABC(c, n, dest, opcGenSym)
```
zerbina:
```suggestion
    prepare(c, dest, n.typ)
    let arg =
      if n.len == 2: # Unary genSym
        c.genx(n[1])
      else: # Deprecated binary form
        c.freeTemp(c.genx(n[1])) # eval the first argument but discard the result
        c.genx(n[2])
    c.gABC(n, opcGenSym, dest, arg)
    c.freeTemp(arg)
```
A little more code, but all legacy compatibility is implemented in the code generator this way.

compiler/vm/vm.nim
```diff
@@ -2707,15 +2707,19 @@ proc rawExecute(c: var TCtx, t: var VmThread, pc: var int): YieldReason =
...
-      assert regs[rc].handle.typ.kind == akString
+      let rh =
+        if regs[rb].kind == rkInt:
+          # Called from compatibility symbol
+          rc
+        else:
+          rb
+
+      checkHandle(regs[rh])
+      assert regs[rh].handle.typ.kind == akString
```
zerbina:
```suggestion
      checkHandle(regs[rb])
      assert regs[rb].handle.typ.kind == akString
```
A follow-up to the above comment.

compiler/vm/vm.nim
```diff
@@ -2707,15 +2707,19 @@ proc rawExecute(c: var TCtx, t: var VmThread, pc: var int): YieldReason =
...
       # XXX: costly stringify of strVal... `getIdent` doesn't use openArray :(
-      let name = if regs[rc].strVal.len == 0: ":tmp"
-                 else: $regs[rc].strVal
-      guestValidate(k in 0..ord(high(TSymKind)),
-        "request to create symbol of invalid kind")
-      var sym = newSym(k.TSymKind, getIdent(c.cache, name), nextSymId c.idgen, c.module.owner, c.debug[pc])
+      let name = if regs[rh].strVal.len == 0: ":tmp"
+                 else: $regs[rh].strVal
```
zerbina:
```suggestion
      let name = if regs[rb].strVal.len == 0: ":tmp"
                 else: $regs[rb].strVal
```

tests/errmsgs/tundeclared_symbol.nim
```diff
@@ -0,0 +1,15 @@
+discard """
+  cmd: '''nim check --hints:off $file'''
+  action: reject
+  nimout: '''
+tundeclared_symbol.nim(12, 17) Error: symbol used before declaration: 'c' [generated in tundeclared_symbol.nim(12, 17)]
+'''
```
zerbina:
```suggestion
  errormsg: "symbol used before declaration: 'c' [generated in tundeclared_symbol.nim(9, 17)]"
  line: 9
```
This is more in line with the other tests from `errmsgs`. I've also modified the line numbers, so applying the change should just work.

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -1,68 +1,120 @@
...
+  of nnkStmtList, nnkStmtListExpr, nnkForStmt:
+    nameExpr n[^1]
+  of nnkConstSection, nnkConstDef:
+    nameExpr n[0]
+  of nnkDiscardStmt:
+    nameExpr n[0]
+  of RoutineNodes:
+    n.name
+  of AtomicNodes:
+    n
```
zerbina:
```suggestion
  of nnkStmtList:
    nameExpr n[^1]
  of nkLetSection, nkVarSection, nkConstSection:
    n[0][0]
  of RoutineNodes:
    n.name
```
I would suggest just always using routines and let/var statements as the symbol suppliers. The only reason I sometimes used `nskForVar` was because it cost effectively nothing.

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -1,68 +1,120 @@
...
+  proc f() = discard
+do:
   func f() = #[tt.Error
       ^ (SemSymbolKindMismatch)]#
     discard
 
-replaceName(nskVar):
+replaceName:
+  for iter in [0]:
+    discard iter
```
zerbina:
```suggestion
  var iter = 0
```
A follow-up to the above suggestion.

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -1,68 +1,120 @@
...
+do:
   iterator iter() = #[tt.Error
           ^ (SemSymbolKindMismatch)]#
     discard
 
-replaceName(nskForVar):
+replaceName:
+  for conv in [0]:
+    discard conv
```
zerbina:
```suggestion
  proc conv() = discard
```
A follow-up to the above suggestion.

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -1,68 +1,120 @@
...
+do:
   converter conv() = #[tt.Error
            ^ (SemSymbolKindMismatch)]#
     discard
 
-replaceName(nskForVar):
-  method meth() = #[tt.Error
+replaceName:
+  for meth in [0]:
+    discard meth
```
zerbina:
```suggestion
  proc meth() = discard
```

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -1,68 +1,120 @@
...
 
-macro t1(): untyped =
-  # test that an error is correctly reported when inserting the wrong symbol
-  # into a non-routine definition
-  let b = genSym(nskLet, ident = "a")#[tt.Error
-               ^ (SemSymbolKindMismatch)]#
-  result = newVarStmt(b, newLit 1)
+replaceName:
+  let a = 0
+  a
```
zerbina:
```suggestion
  let a = 0
```
A follow-up to the above suggestion.

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -1,68 +1,120 @@
...
+do:
+  var a: int #[tt.Error
+     ^ (SemSymbolKindMismatch)]#
 
-t1()
+replaceName:
+  proc c = discard
+  c
```
zerbina:
```suggestion
  proc c() = discard
```
So that it's stylistically in line with the other test cases.

tests/lang_callable/macros/tsymbol_must_match.nim
```diff
@@ -1,68 +1,120 @@
...
+do:
+  const c = 10 #[tt.Error
+       ^ (SemSymbolKindMismatch)]#
+
+replaceName:
+  func T = discard
+  T
```
zerbina:
```suggestion
  func T() = discard
```
Same here.

zerbina (commented):
Looks great! I've left some minor suggestions regarding the tests, plus a suggestion for handling all `genSym` compatibility in `vmgen`.

For the PR description, could you add to the summary a mention regarding the `kind` parameter of `genSym` being ignored now and fully optional?

zerbina:
@alaviss: Regarding your question about whether the first-pass for templates should use `skGenerated`, my answer would be no. The name `newGenSym` is somewhat of a misnomer, as while the symbols are "generated", they're not in the same way as it happens for macros. `skGenerated` makes sense for user-generated symbols, but for templates, the symbols and their creation are controlled by the compiler.

compiler/vm/vmgen.nim
alaviss: Would it be better to just cut a new csources after this, then we can remove this compat generation code and switch to templates?

Actually I'm already `when`-ing the thing, so switching to templates might not be too big of a deal.

zerbina: Given that `genSym` is a magic, meaning that using it as a first-class procedure is not possible anyway, using a template + `.line` would be better, yea.

lib/core/macros.nim
```diff
@@ -432,10 +432,33 @@ proc bindSym*(ident: string | NimNode, rule: BindSymRule = brClosed): NimNode {.
...
+    ## needs to occur in a declaration context.
+
+  template genSym*(kind: NimSymKind = nskGenerated; ident: string): NimNode {.
+    deprecated: "genSym no longer takes a `kind` parameter".} =
+    ## Generates a fresh symbol that is guaranteed to be unique. The symbol
+    ## needs to occur in a declaration context.
+    ##
+    ## This is a compatibility alias for `genSym <#genSym,string>`_, the `kind`
+    ## parameter is ignored.
+    genSym(ident)
```
zerbina:
```suggestion
    {.line.}:
      discard kind 
      genSym(ident)
```
Even if the routine is deprecated, I still think it makes sense to keep its behaviour roughly as it was, especially if it as simple as it is here.

lib/core/macros.nim
```diff
@@ -432,10 +432,33 @@ proc bindSym*(ident: string | NimNode, rule: BindSymRule = brClosed): NimNode {.
...
+
+  template genSym*(kind: NimSymKind): NimNode {.
+    deprecated: "genSym no longer takes a `kind` parameter".} =
+    ## Generates a fresh symbol that is guaranteed to be unique. The symbol
+    ## needs to occur in a declaration context.
+    ##
+    ## This is a compatibility alias for `genSym <#genSym,string>`_, the `kind`
+    ## parameter is ignored.
+    genSym()
```
zerbina:
```suggestion
    {.line.}:
      discard kind
      genSym()
```

tests/lang_stmts/defer/tdefer_malformed_many_children.nim
```diff
@@ -2,7 +2,7 @@ discard """
   description: "`defer` must have exactly one child node (macro input)."
   errormsg: "illformed AST"
   file: "macros.nim"
-  line: 647
+  line: 670
```
zerbina:
```suggestion
  line: 674
```
Adjusts the line number to still be correct after the two previous changes.

zerbina (approved):
One minor suggestion for keeping backwards compatibility for `genSym`, as it's super easy to do in this case.

PR #1587, opened 2025-08-14
fix heap-cell cleanup calling the wrong destructor

zerbina:
## Summary

Fix `ref`s to objects calling the wrong destructor upon the refcount
reaching zero in some narrow circumstances involving generic types and
`sink` type modifiers.

## Details

Due to the sighash-based canonicalization used prior to type-bound op
lifting ignoring `sink` modifiers, the same hook was lifted for both
`ref Obj[proc(x: sink T)]` and `ref Obj[proc(x: T)]`, which led to the
wrong destructor being called on cleanup when the object type is final.

In a very narrow situation, where all of the following apply:
* the `ref` type itself is generic
* the object type is never used outside the `ref` type
* the object type has no user-defined destructor, only an implicit one
* the object type is non-final
* the types reach the MIR phase in the wrong order
the canonicalization led to the object not having its destructor called
at all, due to the destructor not being lifted, thus *leaking memory*
if the object type requires destruction for some of its fields.

Two types for which `sameType(a, b)` results in `false` should produce
different sighashes, and so hasing is changed to not ignore `sink`
types, fixing both issues.

RTTI creation in `ccgtypes` relied on `sink` not contributing to
sighashes; both `genTypeInfoV1` and `genTypeInfoV2` are changed to
manually skip `sink`, so that `sink T` and `T` continue to use the same
RTTI.

In addition, the scanning for types to lift the ops for in `sempass2`
is fixed to handle generic ref types used as the constructor in object
construction expressions, which makes the code work as intended.

### DIFF
--- a/compiler/backend/ccgtypes.nim
+++ b/compiler/backend/ccgtypes.nim
@@ -1116,7 +1116,7 @@ proc genTypeInfoV2Impl(m: BModule, t, origType: PType, name: Rope; info: TLineIn
     discard genTypeInfoV1(m, t, info)
 
 proc genTypeInfoV2(m: BModule, t: PType; info: TLineInfo): Rope =
-  let origType = t
+  let origType = skipTypes(t, {tySink})
   # distinct types can have their own destructors
   var t = skipTypes(origType, irrelevantForBackend + tyUserTypeClasses - {tyDistinct})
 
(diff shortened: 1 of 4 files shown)


zerbina:
This is a test case for the potential memory leak described in the PR description:
```nim
type Object = object

var numDestroyed = 0

  inc numDestroyed

type Ref[T] = ref object of RootObj
  # the object must be non-final (so that its destructor call requires RTTI)
  y: Object

proc test1() =
  discard Ref[proc(x: int)]()

proc test2() =
  discard Ref[proc(x: sink int)]()

# test1 being unused is deliberate. The `Ref` type that's first scanned by
# `sem` must either not reach the C code generator at all, or it must be
# visited by RTTI creation *after* the `Ref` type scanned after it.
test2()

doAssert numDestroyed == 1 # the assertion fails because `numDestroyed` is 0
```

Since it depends on too many not-directly related implementation details, the test doesn't have much value as a regression test, hence me not adding it to the test suite.

saem (approved):
Wow, how did you find this one?

PR #1591, opened 2025-08-17
use a `-d:release` compiler during CI

zerbina:
## Summary

* use a `-d:release` compiler during CI testing (instead of a
  `-d:danger` one) to be able to catch more issues 
* package a `-d:release` compiler into release bundles

## Details

A `-d:danger` compiler has both assertions and run-time checks
disabled, meaning that some compiler issues (e.g., access of inactive
field) cannot be detected with it.

To be able to catch these issues with CI, a `-d:release` compiler is
used now, which has both run-time checks and assertions enabled. Since
the compiler shipped in a release bundle should be the same as was
tested, release bundles now ship a `-d:release` compiler.

As a consequence, the compiler from release bundles becomes *slower*;
the same goes for CI test runs.

### DIFF
--- a/tools/koch/koch.nim
+++ b/tools/koch/koch.nim
@@ -366,7 +366,7 @@ type
 proc buildReleaseBinaries(args = "") =
   ## Build binaries needed for creating a release
   # Boot the compiler
-  boot("-d:danger " & args)
+  boot("-d:release " & args)
   # Build the tools
   buildTools(args)
 


zerbina:
I'll create separate PRs for fixing the bugs causing the test failures.

alaviss (approved):

zerbina:
For comparison, with this PR, the time the `CI` workflow takes to run is [~31.5 minutes](https://github.com/nim-works/nimskull/actions/runs/17106813759), whereas for the previous PR, the same workflow took [~29.5 minutes](https://github.com/zerbina/nimskull/tree/refs/heads/injectdestructors-fix-lifetime-lookup), meaning an increase by about 2 minutes (6%).

saem (approved):

PR #1592, opened 2025-08-20
injectdestructors: fix lifetime lookup

zerbina:
## Summary

Fix an internal issue with globals with multiple lifetimes during move
analysis, destroy elimination, and `wasMoved` injection.

## Details

Due to typed templates parameters and `for` loop bodies being inlined
into every `yield`, globals can currently have multiple lifetimes,
which `injectdestructors` has to handle.

So far, lifetime lookup always picked the last one starting before the
target operation. For lifetimes fully enclosing other ones, this
resulted in the wrong one being picked, subsequently triggering an
assertion during data-flow analysis.

Lookup is changed to return the lifetime most closely matching an
operation. This makes some code compile or work closer to what
one expects, but it doesn't fix the fundamental issue of there being
multiple lifetime in the first place.

In addition to the lookup change, `findScope` and `EntityInfo` are
renamed to `findLifetime` and `Lifetime`, respectively, making it
clearer what both do / represent.

### DIFF
--- a/compiler/mir/mirbridge.nim
+++ b/compiler/mir/mirbridge.nim
@@ -42,6 +42,7 @@ let reprConfig = block:
   rc.flags.excl trfShowFullSymTypes
   rc.flags.excl trfShowNodeTypes
   rc.flags.incl trfShowSymKind
+  rc.flags.incl trfShowSymId
   rc
 
 template isEnabled(config: ConfigRef, ir: IrName, name: string): bool =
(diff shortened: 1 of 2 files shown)


zerbina:
An example of a global with multiple lifetimes where one encloses the other:
```nim
template wrap(stmt: typed) =
  block:
    stmt
    if true:
     stmt

wrap:
  var x = @[1, 2, 3]
```
This previously triggered an assertion during data-flow analysis, as the lifetime lookup for ops on `x` always picked the outer one.

Nested lifetimes are of course nonsense. The template invocation should produce AST using two separate symbols, but `sem` cannot easily and cleanly be made to, hence the problem being worked around in `injectdestructors`.

saem (approved):

PR #1593, opened 2025-08-24
msgs: use effect-based diagnostic emission/handling

zerbina:
## Summary

* report errors/warnings/hints in the order they're created, fixing the
  out-of-order reporting
* make emission and handling of diagnostics effect-based

## Details

The current deferred approach to diagnostics, where most sem errors are
first placed into AST and then later emitted/reported in bulk has the
major problem of requiring the errors to first "bubble up", which
requires unnecessary work in some cases, also leading to out-of-order
reporting when mixed with what was previously considered
legacy reporting.

Emitting error diagnostics immediately allows handling them right away
(e.g., by terminating the program). Emission is modeled as an effect,
which is approximated via a callback.

### High-Level Description

* `emit` dispatches to the current handler, which is stored in
  `ConfigRef` for now, as that's already available everywhere
  diagnostics are emitted
* the high-level routines previously doing reporting now emit the
  diagnostic instead
* `newError` is treated as committing to the diagnostic, emitting
  it directly
* places that need to override diagnostic handling now override the new
  *diagnostic handler* not the *report hook*
* the default diagnostic handler sends the diagnostic to the report
  hook, same as before
* since `Report` is a superset of `PAstDiag`, `emit` uses `Report` to
  represent diagnostics

### Concrete Changes

* change internal error emission (`internalError` and `internalAssert`)
  to report the diagnostic directly and then quit
* change `localReport` to emit diagnostics
* change `globalReport` to always raise an exception
* remove the concept of diagnostic/report writability (it's obsolete)
* remove all after-the-fact emission of errors stored in ASTs
* remove all manual `compiles` short-circuiting logic
* remove the now-obsolete glue module `errorreporting`
* remove the now-unused/obsolete `doRaise` error action
* remove the now-obsolete `nimDebugUnreportedErrors` debug condsym
* adjust the `docgen` message handler to only use `globalReport` for
  locally-fatal errors, `localReport` otherwise

Places previously relying on `nkError` nodes for error recovery are
changed to use the diagnostic handler:
* overload resolution (diagnostic emitting during analysis are recorded
  in a separate container)
* `compiles` analysis (`tryExpr`)
* checking whether expressions are constant (`tryConstExpr`)
* call-operator-as-fallback handling; the relevant code in
  `semIndirectCall` is refactored
* `parseCode` macro API implementation (`vmdeps.parseCode`)

### Tests

Error messages and order changed in a few cases, and the expected
output of relevant tests are updated accordingly.


zerbina:
The change is not perfect. There are undoubtedly some bugs it introduces, or at least error messages that become worse.

Having to record diagnostics makes sigmatch a bit more complicated and reasoning about the matching somewhat harder, and I did experiment with introducing an `nkDiags` node (which wrapped a tree with the diagnostics recorded for it) to improve on this.

The `nkDiags` approach made recording and associating diagnostics with overload candidates a bit easier, but introducing a new meta node kind that has to be handled in a lot of case statement seemed like it was too much. In addition, installing an effect handler per-operand also has a significant performance overhead.

Despite the problems, I do think that this PR is an improvement.

compiler/front/msgs.nim
```diff
@@ -644,94 +635,78 @@ proc handleReport*(
...
+  ## Emits the diagnostic represented by `report`, with `at` being the compiler
+  ## source location to which the report should point to. This is a fire-and-
+  ## forget operation; the callsite has to assume `emit` returns normally.
+  assert config.diagHandler != nil
+  report.reportFrom = toReportLineInfo(at)
+  config.diagHandler(config, report)
+
+proc report(config: ConfigRef, report: sink Report,
+            at: InstantiationInfo) {.inline.} =
+  ## Sends `report` to the report hook and discards the result.
```
saem, line 670:
```suggestion
  ## Sends `report` to the report hook and discards the result.
  ## xxx `report` is a legacy name and needs to change
```

I want to ensure we don't codify the vocabulary, I still firmly believe the current "report" plague needs to be excised from the compiler.

compiler/tools/docgen.nim
```diff
@@ -253,27 +253,35 @@ template declareClosures =
...
+      of meFootnoteMismatch:        rbackRstFootnoteMismatch
+      of mwRedefinitionOfLabel:     rbackRstRedefinitionOfLabel
+      of mwUnknownSubstitution:     rbackRstUnknownSubstitution
+      of mwBrokenLink:              rbackRstBrokenLink
+      of mwUnsupportedLanguage:     rbackRstUnsupportedLanguage
+      of mwUnsupportedField:        rbackRstUnsupportedField
+      of mwRstStyle:                rbackRstRstStyle
+    if kind in rbackErrorKinds:
+      # TODO: errors/events that are locally fatal have to handle aborting
+      #       where the event is emitted, not here via `globalReport`
```
saem, line 275: Testing my understanding: this means that various parts of docgen should be updated to use locally fatal error reporting, rather than relying on this catchall, correct?

saem (approved):
That's a huge readability improvement, it gets rid of a lot of knots in the compiler (compiles, sigmatch, error kind handling, writability, etc).

compiler/tools/docgen.nim
zerbina: Yea, that's what I meant. Either there needs to be a dedicated "fatal error" noreturn handler, or the respective callsites need to raise an exception themselves.

The handler implementation here should just send the diagnostic to the reporting/display management, without also doing control flow. If it does do control flow, that's totally fine, but it should not be *required* to.

compiler/front/msgs.nim
zerbina: What's the terminology you want to replace "report" with (in this specific case, not generally)? I'm currently using "report" to mean "render/show/display". Essentially, `report` is called when all internal handling (e.g., recording, discarding, suppression, etc.) for a diagnostic is done and it now needs to be sent to the display engine in a fire-and-forget manner. Maybe `display` or `show` would fit better here?

To be be clear, I have no problem with adding the comment.

saem: I was thinking we'd call it `log`, although maybe I'll have less of a strong reaction to `report` once the actual legacy code is removed.

saem:
I've taken a peek at the failure (leak detection), but I can't figure out why it's happening

log: https://github.com/nim-works/nimskull/actions/runs/17535039994/job/49796993395

zerbina:
> I've taken a peek at the failure (leak detection), but I can't figure out why it's happening

There's an issue with `sempass2` that leads to the lambda expressions' type not having its lifetime hooks synthesized properly. I've worked around the issue by using a dedicated setter routine, which has shown to fix the leak when I tested it locally (though I might have missed some occurrences).

PR #1595, opened 2025-08-25
remove the sink inference feature

zerbina:
## Summary

Remove the compiler's unsound sink inference feature. This includes the
`sinkinference` switch and pragma, as well as the `.nosinks` pragma.

## Details

Modifying procedure signatures after they already participated in type
inference and checking (which is what sink inference did) is unsound,
as it invalidates the results of previous analysis.

Even if sink inference were implemented in a way such that it doesn't
make unsound modifications to routine signatures, it would still have
the problems that:
1. it silently changes the operational semantics at a routine's
   callsite
2. it changes concrete (i.e., non generic) routine signatures to
   something incompatible (type wise), allowing for confusing type
   mismatch errors

Everything related to the feature is removed from the compiler:
* the sink inference implementation
* the `sinkinference` compiler switch and option
* the `.nosinks` pragma
* the feature's documentation

The compiler also no longer advertises the sink inference capability
(`nimHasSinkInference`). All mentions and usages of sink inference
are removed from both the compiler and system/standard library.

### Impact

Sink inference was previously disabled by default, with it being
selectively enabled for `system` and a few standard library modules
(i.e., the ones including `inclrtl`).

The feature being gone means that `sink` annotations aren't inferred
anymore for the aforementioned modules.

### Additional Adjustments

* remove `.nosinks` annotations from tests
* adjust tests relying on sink inference
* explicitly mark unjoinable tests as such. They were previously
  prevented from being joined into by the local `nim.cfg` file
* add explicit `sink` annotations to some parameters in `system` where
  needed for proper cleanup


saem (approved):
I didn't realize how this worked, yup, it's deeply unsound.

saem (changes requested):
I think I found a couple of missed items:

- inclrtl.nim
- assertions.nim

This is the search I used, if you're curious: https://github.com/search?q=repo%3Anim-works%2Fnimskull%20nimHasSinkInference&type=code

zerbina:
> I think I found a couple of missed items:

Yep, I've deliberately left those two occurrences in, as they're guarded by a conditional symbol and I wanted to focus on the removal of the feature and less on the cleanup.

The usages in the compiler and tests I *had* to remove, as they mostly weren't guarded by a `nimHasSinkInference` check, meaning that the new compiler wouldn't able to compile itself anymore (due to unknown pragmas being used) if they stayed.

That said, removing the usages in the system or system-adjacent modules wouldn't cause trouble w.r.t. the csources compiler, so I could also just remove them if desired.

saem:
Thanks for the explainer, I recommend removing them as they're awkward to figure out if they should be removed in the future for any newcomers or people looking through the source without historical knowledge.

zerbina:
I've removed the two remaining occurrences. In addition, I've also updated the PR description accordingly and added a mention regarding the impact.

Going over the affected standard library modules and using sink where it makes sense would be a good idea now -- this could also be done before merging this PR, so that there's no performance regression (though I haven't actually measured whether there is).

saem (approved):
Thank you

saem:
> Going over the affected standard library modules and using sink where it makes sense would be a good idea now -- this could also be done before merging this PR, so that there's no performance regression (though I haven't actually measured whether there is).

I'm not seeing CI behave too differently, so maybe we just tackle it incrementally?

zerbina:
> I'm not seeing CI behave too differently, so maybe we just tackle it incrementally?

Sure, that would work and be alright with me. I'll create a GitHub issue for it.

**Edit**: [here's](https://github.com/nim-works/nimskull/issues/1597) the issue.

PR #1606, opened 2025-08-31
transf: fix the compiler crashing on valid conversions

zerbina:
## Summary

Fix the compiler crashing for certain distinct type conversions
involving `object` or `ref object` types.

## Details

`mirgen` expects conversions to be "single-step" conversion, where the
conversion is either coercion from or to a distinct type, a numeric
conversion, or an object up- or down-conversion, triggering an
assertion for conversions involving object types if this is not the
case.

`transf` folded away distinct conversions involving object types,
modifying the operand's type, which led to "multi-step" conversions
when the folded conversion's operand was a conversion itself.

Conversion transformation is changed to only elide same-type object
conversions, keeping the distinct coercion intact. This breaks a test
case relying on distinct conversions that should be disallowed (which
now changed behaviour) -- it's removed.

### DIFF
--- a/compiler/backend/cgirgen.nim
+++ b/compiler/backend/cgirgen.nim
@@ -193,9 +193,10 @@ proc handleSpecialConv(c: ConfigRef, n: CgNode, info: TLineInfo,
   ## between the source type (i.e. that of `n`) and the destination type.
   ## If it is, generates the conversion operation IR and returns it -- nil
   ## otherwise
-  if dest.skipTypes(skipPtrs - {tyDistinct}).kind == tyObject:
-    # if the destination is an object (or ptr/ref object), it must be an
-    # object conversion
+  if dest.skipTypes(skipPtrs - {tyDistinct}).kind == tyObject and
+     n.typ.skipTypes(skipPtrs - {tyDistinct}).kind == tyObject:
+    # if the destination and source are an object (or ptr/ref object), it must
+    # be an object conversion
     genObjConv(n, dest, info)
   else:
     nil
--- a/compiler/sem/transf.nim
+++ b/compiler/sem/transf.nim
@@ -627,8 +627,10 @@ proc transformConv(c: PTransf, n: PNode): PNode =
     of tyObject:
       let diff = inheritanceDiff(dest, source)
       if diff == 0 or diff == high(int):
-        result = transform(c, n[1])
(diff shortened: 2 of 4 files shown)


zerbina:
@alaviss: A heads-up that `nim-sys` might break, given that the regression test for https://github.com/nim-lang/Nim/issues/16722 now fails and is removed.

**Edit:** behaviour changes in that for
```nim
type
  Obj      = object of RootObj
  Distinct = distinct Obj
  ObjRef   = ref Obj
  DistRef  = ref Distinct

proc `=destroy`(x: var Obj) =
  echo "original"

proc `=destroy`(x: var DistRef) =
  echo "distinct"

var x = DistRef(ObjRef())
```
"original" is now echoed instead of "distinct". Behaviour doesn't change for objects that are final, even though the conversion is equally nonsensical.

saem:
I added the `language-design` label as this changes how the user should think about conversions going forward.

compiler/backend/cgirgen.nim
```diff
@@ -193,9 +193,10 @@ proc handleSpecialConv(c: ConfigRef, n: CgNode, info: TLineInfo,
   ## between the source type (i.e. that of `n`) and the destination type.
   ## If it is, generates the conversion operation IR and returns it -- nil
   ## otherwise
-  if dest.skipTypes(skipPtrs - {tyDistinct}).kind == tyObject:
-    # if the destination is an object (or ptr/ref object), it must be an
-    # object conversion
+  if dest.skipTypes(skipPtrs - {tyDistinct}).kind == tyObject and
```
saem, line 196: I don't recall but are `tyTuple` non-existent at the `cgir` level? If not then they should be considered as well.

saem (approved):
LGTM

Possibly need to consider tuples, but I can't remember if they're already gone by the time we get to cgir.

compiler/backend/cgirgen.nim
zerbina: > I don't recall but are `tyTuple` non-existent at the `cgir` level?

They still exist (in the sense that may appears as types on CGIR nodes), but this procedure is only for handling object up- and down-conversions, which use dedicated nodes in the CGIR (`cnkObjUpConv` and `cnkObjDownConv`), hence their special casing.

PR #1611, opened 2025-09-14
diff: greatly speed up `myersDiff`

zerbina:
## Summary

Significantly improve performance of `diff.myersDiff`, especially for
sequences with large differences.

## Details

For input sequences with a large number of differences, the procedure
was very slow, due to the explicit `seq` copy it created for every
search point.

The implementation is changed to use a single sequence for representing
the graph, with the search points then pointing into the sequence.
This:
* gets rid of the `seq` copies
* reduces the in-memory size of the table entries
* reduces the number of allocations

### DIFF
--- a/lib/experimental/diff.nim
+++ b/lib/experimental/diff.nim
@@ -446,11 +446,29 @@ proc myersDiff*[T](
   ## Generate series of sequence edit operations necessary to trasnform
   ## `aSeq` into `bSeq`. For item equality comparison use `itemCmp`
   ##
+  ## Ported from
   ## https://gist.github.com/adamnew123456/37923cf53f51d6b9af32a539cdfa7cc4
-  var front: Table[int, tuple[x: int, history: seq[SeqEdit]]]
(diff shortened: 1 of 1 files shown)


saem (approved):
Looks good.

zerbina:
Using the following following setup:
```nim
import experimental/diff
import std/[monotimes, times, strutils]

let input1 = repeat("a", 256)
let input2 = repeat("b", 256)

let a = getAllocStats()
var took: Duration

# ---- the actual test
block test:
  let start = getMonoTime()
  discard myersDiff(input1, input2)
  took = getMonoTime() - start
# ----

let b = getAllocStats()
echo "took: ", (inMicroseconds(took).int / 1_000_000), " alloc: ", b - a, " peak: ", formatSize(getTotalMem())
```
I get the following results:

|Version|Took|Allocations|Peak|
|-------|----|-------------|---|
|Old|~0.25s|131590|48.188MiB|
|New|~0.01s|8|21.801MiB|

PR #1612, opened 2025-09-15
colordiff: improve `formatDiffed` performance

zerbina:
## Summary

Improve the performance of `colordiff.formatDiffed` where the inputs
are strings and no custom line comparison operator is used.

## Details

The string-taking overload dispatched to the generic sequence
`formatDiffed` overload that converts the sequence to a sequence of
strings first.

While a no-op for sequence-of-string inputs, it still incurs extra
allocations, so the string-taking `formatDiffed` overload is changed to
call the internal diff formatting directly, getting rid of the
unnecessary allocations.

### DIFF
--- a/lib/experimental/colordiff.nim
+++ b/lib/experimental/colordiff.nim
@@ -704,6 +704,10 @@ proc formatDiffed*(
     text1, text2: string,
     conf: DiffFormatConf = diffFormatter()
   ): ColText =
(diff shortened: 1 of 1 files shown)


alaviss (approved):
I'm not a big fan of redefining the names with different types, something like `lines1` and `lines2` might be better, but the function is short enough that it's not a blocker for me.

saem (approved):

PR #1618, opened 2025-09-17
mirtypes: rename `tkRecord` to `tkStruct`

zerbina:
## Summary

Changed the name of the `tkRecord` MIR type kind to the better fitting
`tkStruct` and updated the name of related routines.

## Details

In the context of type theory, 'record' usually refers to an unordered
set of fields, which is what was previously called record types in the
MIR is not -- the fields in an MIR record type are ordered.

'struct' (short for structure) better describes what the type
represents.


saem (approved):
Nice, the new naming is better.

PR #1621, opened 2025-09-26
cmdline/parsers: fixed float edge cases and enums

Luyten-Orion:
## Summary
* fixed `parseCli` raising an error on NaN inputs
* fixed `parseCli` for enum values not compiling

## Details
* Added tests for cmdline/parsers.nim
* The float parser test was split into a separate file since  `isNaN` 
doesn't work on the VM.

### DIFF
--- a/lib/experimental/cmdline/parsers.nim
+++ b/lib/experimental/cmdline/parsers.nim
@@ -78,7 +78,7 @@ proc parseCli*(T: typedesc[SomeInteger], value: string): T =
 
 proc parseCli*(T: typedesc[SomeFloat], value: string): T =
   ## Implements `parseCli` for all floats.
-  parseFloat(value)
+  T(parseFloat(value))
 
 proc parseCli*(T: typedesc[string], value: string): T =
   ## Implements `parseCli` for `string`.
--- a/lib/experimental/cmdline/parsers.nim
+++ b/lib/experimental/cmdline/parsers.nim
@@ -61,6 +61,7 @@ runnableExamples:
 
 import std/strutils
 import std/typetraits
+import std/math
 
 proc parseCli*(T: typedesc[SomeInteger], value: string): T =
   ## Implements `parseCli` for all integers.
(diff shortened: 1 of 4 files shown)


lib/experimental/cmdline/parsers.nim
```diff
@@ -80,7 +81,7 @@ proc parseCli*(T: typedesc[SomeFloat], value: string): T =
   ## Implements `parseCli` for all floats.
   let parsed = parseFloat(value)
 
-  if parsed notin low(T)..high(T):
+  if not (parsed in low(T)..high(T) or parsed.isNaN):
```
alaviss:
```suggestion
  if parsed notin low(T)..high(T) and not parsed.isNaN:
```

Use De Morgan's to remove unnecessary nesting.

Though for this case it's better to just remove the range test, since it's always between two `inf`s, making this a no-op.

zerbina (changes requested):
Thanks for the PR!

Since it's a bugfix PR affecting user-facing behaviour, please change the summary such that it focuses on the *what* (the *why* and *how* go into the details section). For example:
```
* fixed `parseCli` for float values reporting an error for `nan` inputs
* fixed `parseCli` for enum values not compiling
```

The idea is that someone interested in whether the changes/fixes affect them can quickly spot this by only looking at the summary.

zerbina (approved):

lib/experimental/cmdline/parsers.nim
```diff
@@ -78,13 +78,7 @@ proc parseCli*(T: typedesc[SomeInteger], value: string): T =
...
 proc parseCli*(T: typedesc[SomeFloat], value: string): T =
   ## Implements `parseCli` for all floats.
-  let parsed = parseFloat(value)
-
-  if parsed notin low(T)..high(T):
-    raise newException(ValueError):
-      $parsed & " is not in the range of " & $low(T) & ".." & $high(T)
-
-  result = T(parsed)
+  parseFloat(value)
```
alaviss: Needs conversion to target type or this will fail for float32.

alaviss (changes requested):

lib/experimental/cmdline/parsers.nim
zerbina: It presently won't (`float` is implicitly-convertible to `float32`), but having an explicit conversion to be future proof is better.

alaviss (approved):

saem (approved):

PR #1622, opened 2025-09-26
math: add full support for the VM target.

Luyten-Orion:
## Summary
Added pure implementations for `isNaN`, `signbit`, `copySign`, and
`frexp`.

## Details
Besides adding pure implementations for `isNaN` , `signbit` , `copySign`, and
`frexp`, removed the VM overrides for `signbit` and `copySign`
as these are no longer necessary.

The tests for the `math` module weren't run for the VM target
previously, which has now been changed to avoid future regressions.


zerbina (commented):
Thanks again :D

A test for the `math` module already exists, but it's currently misplaced at `tests/stdlib/algorithm/tmath.nim`.

I think it'd better to make the existing `math` test work with the VM target, which would only involve adding a few VM overrides (for `signbit`, `copySign`, and `frexp`).

saem:
@Luyten-Orion I updated the details, see how it now describes some more of the why.

Luyten-Orion:
Noted, I'll move my test to the appropriate place tomorrow then, as well as
looking into implementing the rest of the ops potentially. Would moving
algorithm/tmath.nim to stdlib/tmath.nim be acceptable, then?

zerbina:
> Would moving algorithm/tmath.nim to stdlib/tmath.nim be acceptable, then?

Yep, that's where the test should be located.

Luyten-Orion:
Not too sure why the tooling runs fail...

lib/pure/math.nim
```diff
@@ -212,20 +225,27 @@ func copySign*[T: SomeFloat](x, y: T): T {.inline, since: (1, 5, 1).} =
...
 
+  template copySignImpl: T =
+    when T is float32:
+      const signMask = 1'u32 shl 31
+      type U = uint32
+    else:
+      const signMask = 1'u64 shl 63
+      type U = uint64
+
+    cast[T]((cast[U](x) and not (signMask) )or (cast[U](y) and (signMask)))
```
zerbina:
```suggestion
    cast[T]((cast[U](x) and not signMask) or (cast[U](y) and signMask))
```
The extra parenthesis are not needed and make reading the code harder, in my opinion.

zerbina (changes requested):
Nice, thank you!

Please remove the overrides for `copySign` and `signbit` (in `compiler/vm/vmops.nim`) -- the new `copySign` and `signbit` implementations are not actually used/tested with the VM right now.

lib/pure/math.nim
```diff
@@ -987,7 +1005,39 @@ func ceilDiv*[T: SomeInteger](x, y: T): T {.inline, since: (1, 5, 1).} =
...
-func frexp*[T: float32|float64](x: T): tuple[frac: T, exp: int] {.inline.} =
+func frexp*[T: SomeFloat](x: T): tuple[frac: T, exp: int] {.inline.}
+
+template pureLog2Impl[T: SomeFloat](x: T): T =
+  # ln(2)
+  const Ln2 = 0.693147180559945309417232121458176568075500134360255254120680009
+  if x == 0.0 or x == -0.0: -Inf
+  elif x < 0: NaN
+  elif x >= Inf or x.isNaN: x
+  else: ln(x) / Ln2
```
zerbina: A problem with this implementation is that you don't get exact results for power-of-two inputs anymore. For example, `pureLog2Impl(536870912.0)` does not yield exactly `29.0`, whereas the previous `log2Impl` did.

Given that float values in almost all cases use base 2, having `log2` return exact values for power-of-two values would be good.

zerbina (changes requested):
I think it'd be better to keep the old `log2` emulation for now. Beyond that, the changes look good to me.

To fix the `tcmdline_parsers_float.nim` test failure, you can simply remove the `knownIssue` declaration from the test specification.

Also, please update the details section of the PR message to match the current set of changes.

lib/pure/math.nim
Luyten-Orion: noted, I'll fix this later

zerbina (approved):
Looks good to me.

For your info, I've updated the PR message details section to reflect the current state.

Luyten-Orion:
Understood, sorry I didn't realise you meant to be more explicit about what I did.

lib/pure/math.nim
```diff
@@ -1002,19 +1053,15 @@ func frexp*[T: float32|float64](x: T): tuple[frac: T, exp: int] {.inline.} =
...
-    result.frac = c_frexp(x, exp)
-    result.exp = exp
-  else:
+  template frexpImpl: untyped =
     if x == 0.0:
       # reuse signbit implementation
-      let uintBuffer = toBitsImpl(x)
-      if (uintBuffer[1] shr 31) != 0:
+      let sign = x.signbit
+      if sign:
```
saem:
```suggestion
      if x.signbit:
```

The temporary is unnecessary (I didn't see any other usage) and it's probably to keep these routines lighter of extraneous operations.

lib/pure/math.nim
```diff
@@ -212,20 +223,27 @@ func copySign*[T: SomeFloat](x, y: T): T {.inline, since: (1, 5, 1).} =
...
+  template copySignImpl: T =
+    when T is float32:
+      const signMask = 1'u32 shl 31
+      type U = uint32
+    else:
+      const signMask = 1'u64 shl 63
+      type U = uint64
+
+    cast[T]((cast[U](x) and not (signMask) )or (cast[U](y) and (signMask)))
   # TODO: use signbit for examples
```
saem:
```suggestion
    cast[T]((cast[U](x) and not signMask) or (cast[U](y) and signMask))

# TODO: use signbit for examples
```

The extra parentheses aren't helping, the spacing between `or` was inconsistent, and the shouldn't be butted up against what it doesn't relate to. I'd probably also get rid of the new line between the `cast` and the `when` stmt above, as that's a collective unit and it's a part of the embed routine.

Also, that todo, should probably be moved up towards the doc as that's where the examples themselves are.

lib/pure/math.nim
```diff
@@ -212,20 +223,27 @@ func copySign*[T: SomeFloat](x, y: T): T {.inline, since: (1, 5, 1).} =
...
-      if y > 0.0 or (y == 0.0 and 1.0 / y > 0.0):
-        result = abs(x)
-      elif y <= 0.0:
-        result = -abs(x)
-      else: # must be NaN
-        result = abs(x)
-    else: result = c_copysign(x, y)
+    when defined(js):
+      let uintBuffer = toBitsImpl(y)
+      let sgn = (uintBuffer[1] shr 31) != 0
```
saem: Rather than a `let` this is a great use case for a named parameter directly in `jsSetSign`.

saem (changes requested):
Overall this looks pretty good.

The cast formatting issue needs fixing for sure, some of the other ones are softer suggestions, let me know if you have any questions.

lib/pure/math.nim
Luyten-Orion: I don't understand what you mean by this, so I will await clarification

lib/pure/math.nim
```diff
@@ -212,20 +223,25 @@ func copySign*[T: SomeFloat](x, y: T): T {.inline, since: (1, 5, 1).} =
     doAssert copySign(NaN, 1.0).isNaN
     doAssert copySign(1.0, copySign(NaN, -1.0)) == -1.0
 
+  template copySignImpl: T =
```
saem, line 227: This template definition should be _after_ the todo down below, as that todo relates to the examples above.

lib/pure/math.nim
```diff
@@ -211,21 +222,26 @@ func copySign*[T: SomeFloat](x, y: T): T {.inline, since: (1, 5, 1).} =
...
+
+  template copySignImpl: T =
+    when T is float32:
+      const signMask = 1'u32 shl 31
+      type U = uint32
+    else:
+      const signMask = 1'u64 shl 63
+      type U = uint64
+
+    cast[T]((cast[U](x) and not signMask) or (cast[U](y) and signMask))
```
saem:
```suggestion
    cast[T]((cast[U](x) and not signMask) or (cast[U](y) and signMask))

```

keeps the template as a logical unit, and the rest of the body of the outer routine with the `when` follows.

saem (approved):

saem:
Looks good to me, nicely done!

saem:
@Luyten-Orion just noticed this but the PR description had extraneous spaces around things that were quoted with backticks, and the details were missing items in the summary.

Luyten-Orion:
Checking the edit history, seems like GitHub's bot borked the formatting slightly, and the changes you made to the description have been noted. o7

PR #1623, opened 2025-09-27
sem: codegendecl pragma works for generic routines

saem:
## Summary

`codegendecl` now works on generic routines.

## Details

`codegendecl`  pragma processing no longer corrupts the pragma AST. The
processing error meant asyncjs' use of  `codegendecl`  with generic
routines mutated the pragma AST position of the annotated routine
definition on the first pass, upon generic instantiation the pragma was
processed again but this time it was malformed, resulting in misleading
errors.

Besides fixing this behaviour, by not unnecessarily mutating the AST in
the first place, a test was added to prevent future regressions.


tests/pragmas/tcodgendecl.nim
```diff
@@ -0,0 +1,36 @@
+discard """
+description: "Tests for the codegendecl pragma"
```
saem: I opted out of the `ccodeCheck` directive as that only works on the c target.

The proper approach for that is we introduce `jscodeCheck` and a `codeCheck` where the target specific ones will only run for those targets, while the general one will run for all supported targets (c and js).

compiler/sem/pragmas.nim
```diff
@@ -296,11 +296,11 @@ proc getOptionalStrLit(c: PContext, n: PNode, defaultStr: string): PNode =
   if n.kind in nkPragmaCallKinds: result = getStrLitNode(c, n)
   else: result = newStrNode(defaultStr, n.info)
 
-proc processCodegenDecl(c: PContext, n: PNode, sym: PSym): PNode =
+proc processCodegenDecl(c: PContext, n: PNode, sym: PSym) =
```
saem: The lack of return value sorta breaks the "symmetry" with its sibling routines, but it seemed goofy to maintain it for that reason alone.

tests/pragmas/tcodgendecl.nim
zerbina: Not that it matters much right now, but using `codeCheck.c` and `codeCheck.js` (like how `knownIssue` can be enabled for all targets, or just for some) would also work.

compiler/sem/pragmas.nim
zerbina: The procedure is only used once (and I don't see this changing), so you could just inline it at the callsite.

compiler/sem/pragmas.nim
```diff
@@ -1382,7 +1382,8 @@ proc applySymbolPragma(c: PContext, sym: PSym, it: PNode): PNode =
         result = noVal(c, it)
         incl(sym.flags, sfNoInit)
       of wCodegenDecl:
-        result = processCodegenDecl(c, it, sym)
+        processCodegenDecl(c, it, sym)
```
zerbina: While not strictly necessary now that `newError` emits the error, for good form, it could still be good to place the error node back into the AST. The behaviour of `getStrLitNode` is silly, so this might be somewhat tricky to do right (when continuing to use `getStrLitNode`).

Small aside: it's not exactly obvious, but input AST is still mutated (by `getStrLitNode`).

tests/pragmas/tcodgendecl.nim
```diff
@@ -0,0 +1,36 @@
...
+  block generic:
+    # there was a regression with generics as the pragma was corrupted during
+    # the initial pass
+    proc foo[T]() {.codegendecl: "/* test */ function $2($3)".} =
+      discard
+
+    foo[int]()
+
+else:
+  # c backend
```
zerbina: Not having the C tests would be better, I think, as while the `.codegendecl` is still allowed for the C target (which might have been an oversight, I don't remember the details), it doesn't actually do anything.

zerbina (approved):
I think it'd be better to remove the tests for the C target, but otherwise, looks good.

compiler/sem/pragmas.nim
saem: I inlined it.

compiler/sem/pragmas.nim
saem: I did the re-add, the hoops weren't too bad.

tests/pragmas/tcodgendecl.nim
saem: Good call, I removed them and made a note.

compiler/sem/pragmas.nim
```diff
@@ -1382,8 +1374,16 @@ proc applySymbolPragma(c: PContext, sym: PSym, it: PNode): PNode =
...
-        processCodegenDecl(c, it, sym)
-        result = it
+        ## produces (mutates) sym using the `TSym.constraint` field (xxx) to store
+        ## the string literal from `n`
+        sym.constraint = getStrLitNode(c, it)
+        # issue a deprecation warning:
+        c.config.localReport(it.info, reportSem(rsemCodegenDeclDeprecated))
+        result =
+          case sym.constraint.kind
+          of {nkStrLit, nkRStrLit, nkTripleStrLit}: it
```
zerbina:
```suggestion
          of nkStrLiterals: it
```
Soft suggestion. It's a little more concise and possibly conveys intent better.

saem: I figured we had a constant for that, but forgot to check, thanks for the reminder.

PR #1625, opened 2025-10-02
implement a new CGIR and C code generator

zerbina:
## Summary

Add an all new CGIR, together with a new code generation architecture
using it and a new C code generator.

### New CGIR

Core design decisions for the language:
* structured control-flow: makes translation to Wasm/`asm.js` easier
* syntax that resembles NimSkull/C: familiar to the someone that
  already knows NimSkull and its AST 
* statically and strongly typed: makes debugging and analysis easier 
* narrow(er) operations: keep decision making out of the code
generators
* built-in simple exception handling

Core design decisions for the IR:
* packed AST for storage; terminals are stored out-of-band and
  are interned
* every node carries source location information, to make high
  precision debug information possible
* one IR for everything: types, statements, expressions

The IR also comes with a grammar and type checker, to help with
debugging, troubleshooting, and codifying the static semantics. It's
always built into the compiler, but due to its overhead, has to be
enabled at run-time by passing `-d:validateCgir` to the compiler.

The old CGIR is still used for the JS and VM code generators, and thus
has to be kept for now.

### New Architecture

The intertwined MIR -> CGIR -> C translation is replaced with
separate MIR -> CGIR and CGIR -> C translation steps. This allows
reusing the MIR -> CGIR parts for other code generators.

Code generator may only support a subset/dialect of the CGIR, which the
MIR -> CGIR facilitates by accepting a set of code generator
capabilities.

As a preparation for incremental compilation, the CGIR has support for
being split multiple units (i.e., modules), though this feature is not
actually used right now.

### Breaking Changes

* features: TLS emulation is not supported anymore. The switch still
  exists, but enabling emulation now causes an error
* features: header generation is not supported anymore
* C FFI: field symbols cannot be used in `emit` statements anymore

### Changes To The Produced C Code

* C scopes reflect the NimSkull-level scopes
* immutable pass-by-reference parameters are marked with `NIM_NOALIAS`
* error handling in `.compilerprocs` is not omitted anymore
* except for returning array values, out parameters are not
  used anymore
* large set operations are implemented as runtime procedures
* updating the `TFrame` instances doesn't use C macros anymore
* non-inline routines are emitted in the order they're processed
  by sem (which is roughly the order they appear in the source code)
* RTTI for nominal types is defined in the nominal type's home module
* RTTI for structural types is defined in the project's entry module
* RTTI is initialized using static initializer expressions

### Tests

* split up `thard_alignment.nim` so that the parts not requiring AVX
  can be run on non-amd64 systems
* change  `thard_alignment.nim`  to use  `-mavx`  instead of 
`-march=native` ,
  so that the test is reproducible


zerbina:
Third time's the charm. While far from perfect, I do think the result is a lot better than both the previous attempts *and* the present-day C code generator.

I've spent a lot of time polishing it, but `mir2cg` will have to be rewritten from scratch at some point - it just does too many things.

While I did have to use some macros, I tried to keep them as simple as possible. One major macro is the `build` macro for building CGIR trees, another the `genGrammar` macro, for describing the grammar and generating syntax checks based on it. In both cases, writing the resulting code by hand would be unmaintainable.

Wasm, `asm.js`, and LLVM being usable as target languages was an active consideration while designing the CGIR, and while I haven't tested, the translation from CGIR to the respective langauges *should* be relatively straightforward (except for the FFI bits).

zerbina:
In C, arrays are not first class types: they're implicitly converted to pointers to their first element, array declarations decay to pointers in parameter positions, and returning arrays from a procedure (without a pointer indirection) is not possible.

To keep `mirgen` simpler, I had previously opted to implement them as C arrays wrapped in structs in `cgen` (with `mir2cg` always emitting identified array types, never inlining them). This worked very well, but has one major flaw that I didn't consider: imported types.

```nim
{.emit: "/*TYPESECTION*/ struct Foreign { int x[2]; };".}

type Foreign {.importc: "struct Foreign", nodecl.} =
  x: array[2, cint]
```

In the context of the declaration `x`, `array[2, cint]` actually refers to a *C array*. There's no distinction between NimSkull vs. C arrays in neither the source language nor the MIR, so C arrays can be used where NimSkull arrays are expected, and vice versa.

There are multiple ways to address this problem, but for now, I've simply opted for using the same approach as the previous code generator, namely to translate NimSkull/MIR arrays to C arrays directly and accommodate for the array limitation in the code generator (mostly `mir2cg`). Having pass-by-value arrays would have allowed for some `sink` improvements and the struct wrapping also fixed:

```nim
type Obj = object
  x: ptr array[2, array[2, Obj]]
# the C code generated for `Obj` doesn't compile
```
so this is quite unfortunate.

Still, the changes also allowed for simplifying the CGIR type system a bit, by removing the dedicated pointer-to-array type and flexible struct fields (both are subsumed by zero-length arrays, inspired by LLVM).

zerbina:
I've split out all commits tagged with `[upstream]`, except for https://github.com/nim-works/nimskull/pull/1625/commits/4ff9ec3794c8b330aa147c628ffe113da2ead96d, https://github.com/nim-works/nimskull/pull/1625/commits/e9e50c3d636bacd4d31087433e546f34c088fcae, https://github.com/nim-works/nimskull/pull/1625/commits/4a185e334df79daf9f39b47b6ad5770864126d2c. These changes are either too small, require too many changes of code made obsolete by this PR, or cannot be easily tested/explained.

zerbina:
Some time measurements performed on Windows using

```cmd
hyperfine --warmup 1 "<exe> --compileOnly --verbosity:0 --hints:off --warnings:off compiler/nim.nim"
```

on the compiler sources at [devel](https://github.com/nim-works/nimskull/commit/71315c071027d622415b68669211f12c97972571). The C compiler used is `MinGW GCC 11.1.0`.

| Compiler | Built By | Time Taken |
| - | - | - |
| https://github.com/nim-works/nimskull/pull/1625/commits/3550429d1521189df31c27591dd40173db7f2fdf | [devel](https://github.com/nim-works/nimskull/commit/71315c071027d622415b68669211f12c97972571) |  14.574 s ±  0.143 s |
| [devel](https://github.com/nim-works/nimskull/commit/71315c071027d622415b68669211f12c97972571) | [devel](https://github.com/nim-works/nimskull/commit/71315c071027d622415b68669211f12c97972571) | 15.337 s ±  0.144 s |
| https://github.com/nim-works/nimskull/pull/1625/commits/3550429d1521189df31c27591dd40173db7f2fdf | https://github.com/nim-works/nimskull/pull/1625/commits/3550429d1521189df31c27591dd40173db7f2fdf | 15.240 s ±  0.167 s |

These results show that the new backend is significantly faster at producing C code, but with the produced C code being a lot slower. A surface-level profiling at some earlier point during development only yielded that everything became a little slower, and I haven't taken a deeper look at it yet.

My guess is that that the `.compilerproc` error handling or the out parameter change are the culprit.

compiler/backend/backends.nim
```diff
@@ -311,6 +295,58 @@ proc generateMainProcedure*(graph: ModuleGraph, idgen: IdGenerator,
...
+proc generateThreadTeardownProc*(graph: ModuleGraph, idgen: IdGenerator,
+                                 modules: ModuleList): PSym =
+  ## Generates the 'nimTeardownThreadVars' procedure, which runs the
+  ## destructors for all live threadvars part of the program.
+  let owner = mainModule(modules).sym
+
+  # setup the symbol:
+  result = newSym(skProc, getIdent(graph.cache, "nimTeardownThreadVars"),
+                  nextSymId idgen, owner, unknownLineInfo, {})
+  result.flags.incl sfExportc
```
saem, line 334: I'm probably not remembering the semantics correctly, apologies if that's the case.

I thought destructors couldn't raise so can this be marked `sfNeverRaises` or is there something else in there that potentially raises?

compiler/backend/cgbackend.nim
```diff
@@ -0,0 +1,353 @@
...
+    ## Accumulator for single module artifacts.
+    sym: PSym
+    idgen: IdGenerator
+
+    globals: seq[(StringId, int32)]
+    procs: seq[(StringId, int32)]
+    emits: Emit
+
+  BModuleList = object
+    ## Bundles all global processor state.
```
saem:
```suggestion
    ## Bundles all global processor state, acting as an accumulator for
    ## the jumbo module definition.
```

Mostly to borrow the language of the module documentation at the top so there is continuity of language, and hopefully also understanding.

compiler/backend/cgbackend.nim
```diff
@@ -0,0 +1,353 @@
...
+    graph: ModuleGraph
+    config: ConfigRef
+    env: MirEnv
+    modules: Table[ModuleId, BModule]
+    total: Context
+      ## CGIR translation context for the single output module
+
+  OutModule* = object
+    ## A view into the total CCGIR module, derived from a source module. The
```
saem, line 95:
```suggestion
    ## A view into the jumbo CCGIR module, derived from a source module. The
```

The alternative to this and the change above would be perhaps to update the module docs at the top to define "total".

compiler/backend/cgbackend.nim
```diff
@@ -0,0 +1,353 @@
...
+    env: initMirEnv(graph),
+    total: initContext(graph, caps)
+  )
+
+  for key, val in mlist.modules.pairs:
+    g.modules[key] = BModule(sym: val.sym, idgen: val.idgen)
+
+  # give the init procedure a pre-defined external name
+  for key, m in mlist.modules.pairs:
+    m.init.extname = getInitName(m.sym)
```
saem, line 244: You could combine this with the loop above.

compiler/backend/cgir.nim
```diff
@@ -1,3 +1,6 @@
+## **Important:** this module provides the definitions for the *lgeacy* CGIR.
+## It's usage is meant to be phased out, and new code generators must not use.
+##
```
saem, line 4:
```suggestion
## For the newer version see the `cgir2` module
##
```

compiler/backend/cgir2.nim
```diff
@@ -0,0 +1,349 @@
...
+  CgNodeKind* = enum
+    cnkInvalid       ## the node is uninitialized
+    # terminals:
+    cnkBool   ## bool value
+    cnkInt    ## signless packed integer value
+    cnkFloat  ## packed float value
+    cnkString ## packed interned string value
+    cnkLabel  ## label name (32-bit uint)
+    cnkType   ## reference to a type (by name)
+    cnkDatum  ## reference to an anonymous (by ID)
```
saem:
```suggestion
    cnkDatum  ## reference to an anonymous constant (by ID)
```

compiler/backend/cgir2.nim
```diff
@@ -0,0 +1,349 @@
...
+
+  CgModule* = object
+    ## A CGIR module using an in-memory representation meant for
+    ## easy processing.
+    tast*: Ast
+      ## all AST for type bodies
+    ast*: Ast
+      ## all AST for everything that's not types
+    data*: Store[Datum, NodeIndex]
+      ## anonymous constants represented as construction expressions
```
saem:
```suggestion
      ## anonymous constants represented as construction expressions
      ## constant datum (id) -> index of corresponding declaration in `ast`
```

I haven't read that far yet, but this is my guess of where it's looking things up.

doc/debug.rst
```diff
@@ -432,7 +432,8 @@ This is done by using one of the following options with the
   applied
 * `mir_out`:option: : shows the MIR after all passes were applied, right before
   translation to the CGIR
-* `cgir`:option: : shows the CGIR as produced by `cgirgen`
+* `cgir`:option: : shows the CGIR as produced by `mir2cg` (`cgirgen` for the
+  legacy CGIR)
```
```suggestion
* `cgir`:option: : shows the CGIR as produced by `mir2cg` (use `cgirgen` for
  the legacy CGIR)
```

compiler/mir/mirtypes.nim
```diff
@@ -1249,9 +1267,46 @@ func usizeType*(env: TypeEnv): TypeId {.inline.} =
   ## unsigned integer type of target-dependent bit-width.
   env.usizeType
 
-# type creation routines
-# ----------------------
+# ---- convenience type constructors
+
+func newArray*(env: var TypeEnv, count: Positive, typ: TypeId): TypeId =
+  ## Generates a array type with `count` elements of type `typ`.
```
saem:
```suggestion
  ## Generates an array type with `count` elements of type `typ`.
```

compiler/backend/cgirbuilder.nim
```diff
@@ -0,0 +1,260 @@
...
+      `b`.nodes.setLen(`start` + (`staticLen` + `dynLen`))
+    result.add quote do:
+      NodeRef(node(`RefTag`, uint32(`start`)))
+
+macro build*(b: var Builder, e: untyped): NodeRef =
+  ## Constructs the sub-tree represented by `e` and returns a reference to it.
+  ##
+  ## The following syntax is used for `e`:
+  ##
+  ## * a call expressions describes a tree; they can be arbitrarily nested, and
```
saem:
```suggestion
  ## * a call expression describes a tree; it can be arbitrarily nested, and
```

compiler/backend/validation.nim
```diff
@@ -0,0 +1,1501 @@
+## Implements the grammer and type checker for the CGIR's in-memory
```
saem:
```suggestion
## Implements the grammar and type checker for the CGIR's in-memory
```

compiler/backend/validation.nim
```diff
@@ -0,0 +1,1501 @@
...
+    blocks: seq[tuple[isTry: bool, label: int]]
+    ret: Type
+    scopes: seq[Table[StringId, Type]]
+
+  MsgContext* = object
+    ## Contextual information about an error.
+    top*: Option[NodeIndex]
+      ## the index of the top-level tree node the error appears in
+    where*: NodeIndex
+      ## the node where the
```
saem:
```suggestion
      ## the node where the error occurred
```

compiler/backend/rtti.nim
```diff
@@ -0,0 +1,496 @@
...
+#       2. emitting the type header initialization for objects in mirgen
+
+import compiler/sem/sighashes
+import compiler/mir/rtti_helper
+from compiler/ast/typesrenderer import typeToString, addTypeHeader
+from std/strutils import toLowerAscii
+
+# field positions are hardcoded because looking up the fields to retrieve
+# the position is too cumbersome
+const
```
saem, line 16: Not that I would necessarily change it now, but any reason to favour `const` over `enum`?

compiler/mir/rtti_queries.nim
```diff
@@ -0,0 +1,55 @@
+## Some RTTI-related query routines.
```
saem, line 1: Maybe rename this module `rtti_queries`, it's more descriptive.

saem (approved):
I left a few minor comments, but it looks good to me, nicely done!

compiler/backend/backends.nim
zerbina: Yep, this procedure can be marked as `sfNeverRaises`, as it only calls destructors, which cannot raise (not even defects).

```suggestion
  result.flags.incl sfExportc
  result.flags.incl sfNeverRaises
```

compiler/backend/cgbackend.nim
zerbina: I think "total" is the clearer/better terminology here.

compiler/backend/cgir2.nim
zerbina: It's the index of the construction expression itself; there's no proper declaration for constants.

Having gained some distance from the code, I think there being no proper declaration syntax (like there is for procedures and globals) is a mistake, as it means that `ast` is not self-contained (ignoring out-of-band literal values), which makes it not possible to write it directly to a file.

doc/debug.rst
zerbina: What I meant to express here is that when the legacy CGIR is used (which happens when compiling using the VM or JS targets), `--showir:cgir` will show the CGIR produced by `cgirgen`, not the one produced by `mir2cg`.

compiler/backend/rtti.nim
zerbina: I chose `const` over `enum` because using an `enum` gives the impression that the values having a dedicated type is of significance (at least to me), which is not the case here -- the integer value is what's important here.

Using an `enum` would allow restricting the `addField` templates to just accept these values (and not arbitrary integers), but given that usage of `addField` is confined to just this module, I don't think this would buy much.

compiler/backend/cgir2.nim
zerbina: I've only improved the comment for now.

Just adding a dedicated constant declaration syntax wouldn't make the AST self-contained at this point, as there's still the "loose" storage types of, and I'd rather combine these changes with adding a (somewhat) proper transportation format (plus the associated reader/writer and integration into the compiler thereof).

Also, to maybe clarify, with self-contained I mean that the AST contains all information necessary to process it / work with it (without having to store any of the lookup tables - except the out-of-band values - currently part of `CgModule` alongside it).

doc/debug.rst
zerbina: I've reworded the documentation to make it easier to understand (the previous wording was indeed quite confusing).

zerbina:
Surprisingly, the `thard_alignment.nim` [test failure](https://github.com/nim-works/nimskull/actions/runs/18514336582/job/52762376761) in the Windows CI is gone now. I've only changed comments and renamed various entities in the commits afterwards, meaning that the changes *shouldn't* have caused the problem to go away.

When I last tried to reproduce the failure, I wasn't able to, with both ASAN and UBSAN builds also not reporting any issues. I even tried using the exact same GCC downloaded by the "Install MinGW" step, but that also didn't yield any failures. The line reported by the signal handler, indicates that the error was *during* calling the `q1` closure, meaning that the procedure pointer itself must have been invalid somehow, but even after looking through the generated C code, I don't see how this would have been possible.

What did change between the previous CI run and the most [recent one](https://github.com/nim-works/nimskull/actions/runs/19054419864/job/54422400584) is a new Windows runner image being used ([20251021.76](https://github.com/actions/runner-images/blob/win22/20251021.76/images/windows/Windows2022-Readme.md) compared to [20250929.55](https://github.com/actions/runner-images/blob/win22/20250929.55/images/windows/Windows2022-Readme.md)), with a potentially relevant difference between the two being that the shipped GCC changed (14.2.0 vs. 12.2.0), which could hint at a GCC bug and the workflow accidentally using the wrong GCC.

zerbina:
The memory leak is caused by the VM's allocator not freeing the cells it creates upon being destroyed itself. I've split out https://github.com/nim-works/nimskull/pull/1625/commits/fe9115925434708198d9e343fbb909d13124c555 into a separate PR that also implements the missing allocator cleanup (#1653).

alaviss (commented):
I don't have the time to look at this PR in detail, but I really liked the shape of the new backend and specifics like how entry points are now written in Skull.

Just wanted to chime in to say that you did a wonderful job :smile:.

zerbina:
The `thard_alignment.nim` failure is due to a known bug with the MinGW gcc compiler, where the alignment requirement for over-aligned temporaries isn't met, causing the aligned AVX instruction gcc generates for the `return` of the anonymous procedure returned by `lambdaTest2` to trap.

I've split up the test now, with only the part really requiring the AVX instruction set extension to be enabled to use `-mavx`, which fixes the test failure (the crash was in the non-AVX-requiring part).

PR #1627, opened 2025-10-02
sigmatch: fix missing conversion for varargs containers

zerbina:
## Summary

Fix an internal-only problem with `sigmatch` handling of named
parameters in conjunction with `varargs`.

## Details

Unlike for unnamed parameters, the implicit array construction created
when using named parameters for passing a single value to a `varargs`
parameter was not wrapped in a to-varargs conversion, nor was the type
marked as being a varargs container (via the `tfVarargs` flag).

This caused no downstream issues so far, as all three code generators
handle the necessary run-time conversion themselves, but it's
still incorrect.


saem (approved):
nice, I ran into this bug when reworking varargs/dispatch.

PR #1633, opened 2025-10-04
tssl: ignore SslError in "close during write" test

alaviss:
## Summary
Fix intermittent Windows CI failures from an unhandled  `SslError` 
being raised due to abrupt peer disconnection in  `tssl` .

## Details
The goal of the test is to verify that closing an SSL socket after a
fatal error happens will not cause the action to raise an error.
Sometimes this fatal error manifests as  `SslError` , which was not
handled, causing intermittent failures (on Windows).

This commit modifies the test to also ignore  `SslError`  should that
happen, since we are only interested in the  `close`  statement not
raising an error.

### DIFF
--- a/tests/stdlib/web/tssl.nim
+++ b/tests/stdlib/web/tssl.nim
@@ -69,7 +69,7 @@ proc main() =
       while true:
         # Send data until we get EPIPE.
         peer.send(DummyData, {})
-    except OSError:
+    except OSError, SslError:
       discard
     finally:
       peer.close()


saem:
@alaviss so why the sudden increase in frequency of occurrence, is that because of underlying version changes in the CI system or something?

alaviss:
Seems to me that GH is slowly rolling out OpenSSL 3.x: https://github.com/actions/runner-images/issues/12676

This might be hitting us on upgraded runners.

saem (approved):
I'll be honest, I don't trust the actual module or the test, and don't know the ssl api well enough to know if this is right... but a less chirpy CI and the rest of the tests passing seem like a win.

PR #1635, opened 2025-10-06
change how top-level emit/asm statements work

zerbina:
## Summary

Only treat `emit` and `asm` statements at the module-scope *and*
outside of any complex expression as top-level emit/asm statements.

## Details

The main goal is to move top-level emit/asm handling largely out of the
code generator and into the orchestrator; the user-facing behaviour
changing is a side effect.

When creating the module lifecycle routines from module AST, emit/asm
statements that are at the top-level are now extracted and put into a
separate section. They're then translated from AST to MIR and
emitted as the new `bekEmit` event by the `backends.process` iterator.

For the C backend, they're passed onto the code generator; they're not
possible for the other two backends. Section determination is removed
from `cgen`.

While the section modifiers are translated to the same C file section
as before, all top-level emit/asm statements are now processed *before*
other top-level code. Tests for top-level emit statements and those
using them are updated (where needed).

### DIFF
--- a/compiler/backend/backends.nim
+++ b/compiler/backend/backends.nim
@@ -4,6 +4,7 @@ import
   std/[
     deques,
     dynlib, # for computing possible candidate names
+    strutils,
     strtabs,
     tables
   ],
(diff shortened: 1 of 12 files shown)


zerbina:
I do think that emit sections are a misfeature (not only in their current incarnation, but in general), but given that they're still used by some tests without and there being a direct replacement, it was easier to just decouple the feature from `cgen` for now.

saem (approved):

PR #1639, opened 2025-10-07
mirtypes: remove the `tkTaggedUnion` type kind

zerbina:
## Summary

Make processing of MIR types simpler by replacing tagged union types
with just union types, using a different mechanism for associating
unions with their tag.

## Details

The `tkTaggedUnion` type was a "logical type", meant to associate
unions with their tag, but without implying any layout or layout
changes. This slightly complicates translating the MIR type
representation to the target language.

To make translation simpler:
* the `tkTaggedUnion` type kind is removed
* tagged unions are now translated to a separate tag and union field
  (mirroring how they're represented in the final C code)
  * fields representing the union part of a tagged union are marked as
    such via a new field flag
  * associating the union with its tag is done via the field name
* fields representing inline structs/unions are now marked as such via
  a flag (instead of using a 'nil'-valued `inst` field)

Translation to C types in `ccgtypes` is updated to work with the new
changes.

### DIFF
--- a/compiler/mir/mirtypes.nim
+++ b/compiler/mir/mirtypes.nim
@@ -410,6 +410,13 @@ func isNoMangle*(f: StructField): bool =
 func isNoAlias*(f: StructField): bool =
   (f.extra and NoAliasFlag) != 0
 
+func isEmbedded*(f: StructField): bool =
+  (f.extra and EmbeddedFlag) != 0
+
+func isTagged*(f: StructField): bool =
+  ## Whether the field stores the union part of a tagged union.
+  (f.extra and TaggedFlag) != 0
+
 func bitsize*(f: StructField): int =
   int(f.extra and 0x00FF)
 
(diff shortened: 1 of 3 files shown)


compiler/mir/mirtypes.nim
```diff
@@ -414,6 +410,13 @@ func isNoMangle*(f: StructField): bool =
 func isNoAlias*(f: StructField): bool =
   (f.extra and NoAliasFlag) != 0
 
+func isEmbedded*(f: StructField): bool =
```
saem, line 413:
```suggestion
func isEmbedded*(f: StructField): bool =
  ## whether the tag is embedded within the containing structure
```

Not sure if that's quite the right description, so checking my understanding + adding docs.

saem (approved):
Minor suggestion, feel free to merge without.

compiler/mir/mirtypes.nim
zerbina: Embeddedness is currently an attribute of fields whose type is either struct or union; it means that the struct/union is inlined into the containing struct/union. Strictly speaking, being embedded is an attribute of the struct/union (which is how things were prior to this PR), but there are some problems with representing this in the MIR type IR, which is why I've moved away from it.

I could rename the attribute to "embedding" and the function to `isEmbedding`.

Ideally, there would be no such things as embedded structs/unions, but they're currently needed to implement variants, as:
* both the inner union and its sub-structs lack a corresponding `PType` to compute their mangled name from
* there has to be a one-to-one correspondence between `PType` field positions and MIR field positions at the moment

saem: Thanks for clarifying

PR #1644, opened 2025-10-11
vm: fix `macros.add` crashing with `openArray` argument

zerbina:
## Summary

Fix the compiler crashing when passing `openArray` value to
`macros.add`.

## Details

* handle `akOpenArray` in `vmtypes.elemType` (used by the
  `mNAddMultiple` implementation)
* handle `akString` in `vmtypes.elemType` (fixes nothing, but it
  should still be handled)


Clyybber (approved):

saem (approved):
It _might_ make sense to handle `akString` but that's more for uniformity/not tripping up assert unnecessarily -- not a blocker, can merge without.

PR #1647, opened 2025-10-12
sem: make `isCyclical` work with resolved concepts

zerbina:
## Summary

Make the `isCyclical` type-trait return whether the bound type for
resolved concept types is cyclical, instead of always `false`.

## Details

* change the "is concrete" test in `evalTypeTrait` to "skip" resolved
  concept types
* consider all resolved concept types in `types.productReachable` (used
  in implementing the `isCyclical` type trait), not just generic
  concept types
* extend the `tcyclic.nim` test with a test case for resolved concept 
  operands


zerbina:
This change fixes the bug causing a CI failure #1646. The relevant part is the change to `productReachable` -- the fix for `isCyclical` is just included so that the `productReachable` fix can be tested (`productReachable` is transitively called by the `isCyclical` implementation).

Clyybber (approved):

PR #1653, opened 2025-11-04
compiler: release VM instance prior to code generation

zerbina:
## Summary

Release the compiler's VM instance before code generation in order to
reduce peak memory usage.

## Details

* the `ModuleGraph` instance is the only location holding a reference
  to the `vmdef.TCtx` instance, setting it to `nil` leads to the VM
  instance being cleaned up and its memory released
* implement a destructor for  `VmAllocator` , to free the
manually-managed
  heap locations pointed to by the stored `VmCell`s
* prevent implicit copies of `VmAllocator` (and thus `vmdef.TCtx`)

### DIFF
--- a/compiler/front/main.nim
+++ b/compiler/front/main.nim
@@ -183,7 +183,7 @@ template prepareForCodegen(g: ModuleGraph) =
   # the compiler, so make sure that they do
   g.config.errorMax = 1
   # the VM instance is not needed anymore. Free it to save memory
-  g.vm = nil
+  reset g.vm
 
 proc commandCompileToC(graph: ModuleGraph) =
   let conf = graph.config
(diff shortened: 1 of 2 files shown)


zerbina:
`VmCell` is still used like a non-resource type in a few places, and I didn't want to refactor them at this time, hence me not going with the more proper approach of using a unique-pointer-like type for the pointer stored in a `VmCell`.

compiler/front/main.nim
```diff
@@ -182,6 +182,8 @@ template prepareForCodegen(g: ModuleGraph) =
   # the backend / code generation phase generally expects errors to terminate
   # the compiler, so make sure that they do
   g.config.errorMax = 1
+  # the VM instance is not needed anymore. Free it to save memory
+  g.vm = nil
```
alaviss:
```suggestion
  reset g.vm
```

I prefer `reset` here since the field being a pointer or whatever doesn't matter, while being a little more self-explanatory

alaviss (approved):
LGTM, only one non-blocking suggestion

saem (approved):

PR #1662, opened 2025-11-29
mir2cg: handle casts between imported types

### LINKED ISSUE #1661: Casting between unknown object types resulted in invalid C code
### Example

* In this example, `SockaddrStorage` is bigger than `SockaddrIn`, but both are `incompleteStruct` to Skull. `sizeof()` within static context will fail.

```nim
import posix

proc main() =
  var x: SockaddrStorage
  let y = cast[SockaddrIn](x)
  echo y

main()
```

### Actual Output

This is the generated C code

```c
      struct sockaddr_storage x;
      FR_.line = (NI64)4;
      nimZeroMem(((void*)(&x)), (NI64)sizeof(struct sockaddr_storage));
      struct sockaddr_in _3;
      FR_.line = (NI64)5;
      nimCopyMem(((void*)(&_3)), ((void*)(&x)), (NI64)-3);
```

### Expected Output

```c
      struct sockaddr_storage x;
      FR_.line = (NI64)4;
      nimZeroMem(((void*)(&x)), (NI64)sizeof(struct sockaddr_storage));
      struct sockaddr_in _3;
      FR_.line = (NI64)5;
      nimCopyMem(((void*)(&_3)), ((void*)(&x)), (NI64)sizeof(struct sockaddr_in));
```

### Additional Information

* Found while debugging nim-sys failures with new codegen.

alaviss:
## Summary
Fix stack corruption due to oversized  `copyMem`  when casting from/to
imported types.

## Details
Type sizes encoded in AST and MIR both use special negative values to
denote either an error or unknown size. This detail is then picked up by
CGIR and rendered verbatim into C code, causing  `nimCopyMem`  to be
called with an excessively large value, corrupting the stack.

This change renders the required computation directly into CGIR when the
size could not be computed by the compiler for casts, solving the
problem.

Fixes https://github.com/nim-works/nimskull/issues/1661

### DIFF
--- a/tests/lang/s05_pragmas/s01_interop/t08_imported_struct_cast.nim
+++ b/tests/lang/s05_pragmas/s01_interop/t08_imported_struct_cast.nim
@@ -1,0 +1,30 @@
+discard """
+  description: '''
+    Imported C struct types can be casted to other types.
+  '''
+  targets: "c"
+  joinable: false
+"""
+
+type
+  ExternBig {.importc: "struct Struct1", header: "t08_imported_struct_cast.nim.h".} = object
+    field1: cint
+    field2: cint
+    # not all fields need to exposed
+
+  ExternSmall {.importc: "struct Struct2", header: "t08_imported_struct_cast.nim.h"} = object
+    field1: cint
+
+  Intern = object
+    field1: cint
+
+let v1 = ExternBig(field1: 1, field2: 2)
+let v2 = cast[ExternSmall](v1)
+let v3 = cast[Intern](v1)
+
+doAssert v1.field1 == 1
+doAssert v1.field2 == 2
+
+doAssert v2.field1 == 1
+
+doAssert v3.field1 == 1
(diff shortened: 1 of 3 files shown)


tests/lang/s05_pragmas/s01_interop/t08_imported_struct_cast.nim
```diff
@@ -0,0 +1,30 @@
+discard """
+  description: '''
+    Imported C struct types can be casted to other types.
```
zerbina:
```suggestion
    Imported C struct types can be cast to other types.
```

tests/lang/s05_pragmas/s01_interop/t08_imported_struct_cast.nim
```diff
@@ -0,0 +1,30 @@
...
+  '''
+  targets: "c"
+  joinable: false
+"""
+
+type
+  ExternBig {.importc: "struct Struct1", header: "t08_imported_struct_cast.nim.h".} = object
+    field1: cint
+    field2: cint
+    # not all fields need to exposed
```
zerbina:
```suggestion
    # not all fields need to be exposed
```
Sorry for missing this oversight in my own code :D

zerbina (approved):
This is well done, thank you for the fix!

Modifying the translation like you've done is exactly the way to handle this edge case (i.e., cast between types with unknown sizes).

With how the test suite is currently structured, I'd consider the test to be in the most fitting place (and I also put the one I added there).

PR #1669, opened 2025-12-06
sem: fix wrong procedure type relation w.r.t. effects

zerbina:
## Summary

Fix `proc` types derived from routines with inferred effects being
treated as if they had no effects, in the context of `proc` type
relations.

## Details

* always initialize the effect slot list for valid proc types, even
  when there's no pragma 
  * this removes the "missing effect list" state, simplifying further
    processing, but also increasing overall memory usage a little
  * generated enum-to-string procedure have their effect lists set to
    empty (they never raise) and are marked as side-effect free
  * the same goes for generated hooks
* always store effect lists as `nkBracket` trees in the respective
  effect slots in `PType`s
  * removes the wrong distinguishing between inferred and declared
    effects from types
  * also makes the encoding more uniform
* rewrite effect compatibility test (`types.compatibleEffects`) to be
  clearer and better documented, taking the previously described
  changes into account
* add a basic test for the effects of effects on proc type relations

### DIFF
--- a/compiler/ast/types.nim
+++ b/compiler/ast/types.nim
@@ -1030,7 +1030,8 @@ proc safeInheritanceDiff*(a, b: PType): int =
 proc compatibleEffectsAux(se, re: PNode, unknown, differ: EffectsCompat
                          ): EffectsCompat =
   ## Computes whether effect lists `se` and `re` are compatible; they are when
-  ## `se` is a superset of `re`.
+  ## `se` is a superset of `re`. If compatible returns `efCompat`, otherwise `differ`
+  ## or `unknown` if incompatible or cannot be determined, respectively.
   if se.isNil:
     # assume that it means "any effect"
     # FIXME: ^^ this is not always true! 'nil' can also mean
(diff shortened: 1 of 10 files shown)


compiler/ast/types.nim
```diff
@@ -1027,15 +1027,30 @@ proc safeInheritanceDiff*(a, b: PType): int =
...
-    block search:
-      for s in items(se):
-        if safeInheritanceDiff(s.typ, r.typ) <= 0:
-          break search
-      return false
-  result = true
+proc compatibleEffectsAux(se, re: PNode, unknown, differ: EffectsCompat
+                         ): EffectsCompat =
+  ## Computes whether effect lists `se` and `re` are compatible; they are when
+  ## `se` is a superset of `re`.
```
```suggestion
  ## `se` is a superset of `re`. If compatible returns `efCompat`, otherwise `differ`
  ## or `unknown` if incompatible or cannot be determined, respectively.
```

Just to clarify the parameter to return relationship if someone is just seeing the call + docs.

saem (approved):
One small soft suggestion for a doc comment, otherwise looks good to me.

PR #1671, opened 2025-12-09
sem: only allow single `ref`/`ptr` indirection for effects

zerbina:
## Summary

For types in `.raises` and `.tags` list effect, only allow a single
`ref` or `ptr` indirection.

## Details

* change `.raises` and `.tags` list processing to not modify input AST
* preserve the original type for effects (the type was overwritten with
  the skipped type, previously)
* disallow more than one `ref` or `ptr` indirection for types in effect
  lists
* add a test for ensuring the current rules regarding what types are
  allowed in effect list work


zerbina:
For both the exceptions and tags lists, the only types allowed should be those that are a subtype of `Exception` and `RootEffect`, respectively. Right now, all object types are allowed, which is nonsense (especially for exceptions).

alaviss (approved):

tests/effects/tallowed_effect_types.nim
```diff
@@ -0,0 +1,57 @@
...
+# multiple ref or ptr indirections are not allowed
+doAssert not compiles(proc() {.raises: [ref ref ProperException].})
+doAssert not compiles(proc() {.raises: [ptr ptr ProperException].})
+doAssert not compiles(proc() {.raises: [ptr ref ProperException].})
+
+# arbitrary object types are allowed in .raises effect lists too
+# TODO: disallow these
+doAssert compiles(proc() {.raises: [ImproperException].})
+doAssert compiles(proc() {.raises: [ptr ImproperException].})
+doAssert compiles(proc() {.raises: [ref ImproperException].})
```
saem: Should these be moved to a known issue test? Or at least an assert message set, maybe instead of a comment, so it might pop up in CI logs?

saem (approved):
One soft question/suggestion, looks good otherwise

tests/effects/tallowed_effect_types.nim
zerbina: I've split the test file into multiple ones and removed the usage of `compiles`. The "shouldn't work" parts are now in a `knownIssue` test.

PR #1675, opened 2025-12-25
mirflow: make optimization algorithm non-recursive

zerbina:
## Summary

Change the space optimization applied to the internal, structured
control-flow representation used in the code generator to use a non-
recursive approach.

## Details

* reorder the fields in `mirflow.Stmt` in order to reduce the type's
  required space (24 byte vs. 32 byte, on a 64-bit architecture)
* apply the block folding right when emitting a block. This gets rid of
  the recursive post-processing
* implement a simple optimization pass for inlining some CGIR `block`
  continuations into dispatcher targets
* add a dedicated statement kind corresponding to the MIR's `continue`,
  so that the latter is visible as a terminator to statement analysis

The optimization pass eliminates deep nesting when there are `case`
statements with many branches, though it's unable to fully optimize the
hash-table lookup generated for `case` statements over `string` values.

Processing of the structure control-flow representation in `mir2cg`
still uses recursion, meaning that its still susceptible to stack
overflow when the structures are very nested.

---

## Notes for Reviewers
* the first half of fixing `thtmlparser.nim` not compiling with a debug compiler

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/compiler/backend/mir2cg.nim
+++ b/compiler/backend/mir2cg.nim
@@ -3274,7 +3274,7 @@ proc procToCgir(c; env; sym: PSym): StringId =
 
   # create the structured control-flow view and use it to guide translation
   var list = toStructured(c.prc.body.code)
-  optimize(list)
+  optimize(c.prc.body.code, list)
   if hasExit(c.prc.body.code):
     # the body is wrapped in a block, which is used as the target for both
     # `Return` and `Unwind`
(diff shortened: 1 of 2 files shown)


zerbina:
Changing `mir2cg` to not use recursion for traversing the control-flow representation isn't as easy, given that there are multiple, mutually recursive routines. It's likely that the test (or other code with large string `case` statements/expression) won't compile for a long while (until `mir2cg` is replaced with something better).

As for the CI failure, the C code generator currently emits code that relies on C compiler extensions or C23 (i.e., declarations immediately following a label), and the clang used by the macOS CI seems to be rather strict. I'll make a separate PR for fixing this.

PR #1678, opened 2026-01-18
mir2cg: use a better implementation for `of`

### LINKED ISSUE #1677: `of` operator incorrectly returns true in some inheritance trees.
### Example

```nim
type
  TraitBase[T] = ref object of RootObj
    table: T
  Trait[T, Y] = ref object of TraitBase[T]
    obj: Y


type 
  A = object
    x: int
  B = object
    y: float32

var a = Trait[int, A]().TraitBase[: int]()
assert a of Trait[int, B]
```

### Actual Output

Compiles and runs fine

### Expected Output

An assertion raised as `a` does not inherit from `Trait[int, B]`

zerbina:
## Summary

Use a pointer-chasing approach for implementing `of` for CGIR-based
backends, fixing a bug with generic base type and improving the
performance of `of` and `method` dispatching.

## Details

* add the `base` field to the version 2 RTTI, for storing a reference
  to the parent type's RTTI
* add the `isObjV2` compilerproc, which uses the new `base` field for
  implementing the `of` type relation test
  * `isObj` cannot be reused, as it's still needed by the csources
    compiler
* change `mir2cg` to emit the necessary RTTI initialization and to
  use `isObjV2`

In theory, the new approach also allows for RTTI to work across dynamic
library borders (by using known names for the RTTI global and then
importing them), but this isn't implemented nor tested yet.

Judging by some limited testing, the pointer-chasing approach to
implementing the `of` relation test is quite a bit faster than the
previous string comparison.

Fixes https://github.com/nim-works/nimskull/issues/1677 .

### DIFF
--- a/compiler/backend/mir2cg.nim
+++ b/compiler/backend/mir2cg.nim
@@ -1372,9 +1372,9 @@ proc emitDefault(c; env; dest: Expr, stmts; bu) =
 
 proc genOf(c; env; tree; e: Expr, typ: TypeId; bu): NodeRef =
   bu.build Call(
-    ^bu.useCompilerProc(c, env, "isObj"),
+    ^bu.useCompilerProc(c, env, "isObjV2"),
     ^c.fieldAccess(env, e, -1, bu),
-    Value(CstringType, ^genTypeInfo2Name(env[typ])))
+    *use(^c.getTypeInfoV2(env, env[typ], bu)))
 
 proc emitLength(c; env; dest, val: Expr, stmts, bu) =
   ## Emits a statement for storing the length of sequence-like `val` in `dest`.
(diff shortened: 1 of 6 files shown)


zerbina:
I've contemplated using the same approach Nim uses nowadays. That is, to compute unique, 4 byte tokens for object types and store a sequence of those representing the parent type hierarchy in the RTTI.

It would allow for `of` to be `O(1)`, but computing the 4 byte tokens requires either having to know all existing object types, or to derive the token from the types' signature hash (like how Nim does it). The former is problematic for isolated module compilation (needed for IC) and RTTI that's stable across dynamic library boundaries, while the latter doesn't guarantee unique tokens, hence me ultimately deciding against the "token sequence" approach.

saem (approved):
That's a really clean change.

PR #1689, opened 2026-03-15
dust: use a better reduction strategy

zerbina:
Replace the reduction strategy preferring the largest removal at each step with one preferring progress (i.e., a successful mutation to be found quickly, even if only a small reduction).

Whereas the old strategy was to start with the mutation representing the largest removal at each step, successively trying the next smaller mutation until a reproducer is found, discarding context (i.e., where the previous mutation took place) between successive steps, the new strategy is to go over the tree with a cursor (from end to start) and remove/simplify nodes/subtrees at every step until a fixpoint is reached.

Especially for larger programs, the new strategy results in finding a minimal reproducer with significantly less semchecks.

The new strategy is derived from an attempt at simulating how one would reduce programs by hand, which is based on the observation that removing from the end of an AST is, in most cases, more likely to be yield a reproduction than removing from the start. This is because NimSkull programs, much like programs written in most other programming languages, are structured in a top-to-bottom left-to-right manner.

---

## To-Do

- [ ] implement dedicated iteration for `nkCommand` and `nkPragmaExpr` again
- [ ] (maybe) improve the documentation in some places

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:

### DIFF
--- a/tools/dust/dust.nim
+++ b/tools/dust/dust.nim
@@ -5,9 +5,7 @@
 
 import
   std/[
-    algorithm,
-    os,
-    sets
+    os
   ],
   compiler/ast/[
     ast,
@@ -21,7 +19,6 @@ import
   compiler/sem/[
     passes,
     sem,
-    sighashes,
   ],
   compiler/utils/[astrepr,],
   std/options as std_options, # due to legacy reports stupidity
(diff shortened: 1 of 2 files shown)


zerbina:
This is only conjecture (as I don't have a formal proof), but the new strategy should make the number of needed semchecks correlate much stronger to the overall size of the input program (as in, number of nodes) than was the case previously.

To give an example of a practical improvement, I tried the following: adding `echo missing` (i.e., an erroneous term) to the end of the innermost `for` loop in `mleak` in `tests/lang_objects/destructor/tarc.nim` and have `dust` create a minimal reproducer. With the previous strategy, this took 1275 semchecks and more than 10 minutes on my machine (making `dust` practically useless), whereas with the new strategy, finding a minimal reproducer takes 65 semchecks and is done in under a minute.

PR #1692, opened 2026-03-21
sigmatch: fix missing conversion for number constants

zerbina:
## Summary

Fix integer and float constants having an incorrect type when passed to
generic parameters. This is presently an internal-only problem, with no
observable consequences.

## Details

The type relation between an integer/float type and a literal is
`isFromIntLit`, but when the substitution of a type variable is
necessary for reaching the integer/float type, the relation regresses
to `isGeneric`, which led to no hidden conversion being injected by
`paramTypesMatchAux`.

Apart from an inconsequential type error at the MIR level, these
wrongly
typed number constants caused no issue, but they would once final
translation differs between the different integer/floats.

`paramTypesMatchAux` is changed to insert a hidden conversion when the
argument is a literal number and the type relation is  `isGeneric` .
Since
this makes it possible for `concept`-typed integers (conversions of
constant integer values are folded) to reach `mirgen`, `intLiteral` has
to skip `tyUserTypeClasses`.

### DIFF
--- a/compiler/mir/mirgen.nim
+++ b/compiler/mir/mirgen.nim
@@ -334,7 +334,7 @@ proc toIntLiteral(env: var MirEnv, n: PNode): Value =
   assert n.kind in nkIntLiterals
   let typ = env.types.add(n.typ)
   # use the type for deciding what whether it's a signed or unsigned value
-  case n.typ.skipTypes(abstractRange + {tyEnum}).kind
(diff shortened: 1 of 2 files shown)


zerbina:
There is a countless amount of issues a MIR-level type checker would have helped in the past, and not implementing one early on really was a mistake.

In general, when introducing a new IR that has its own type system, it makes sense to also implement a type checker for it.

saem (approved):

PR #1694, opened 2026-03-21
jsgen: implement proper 64-bit integer support

zerbina:
## Summary

For the JavaScript backend, make `int64`, `uint64` and other integer
types work like they do when using the other backends.

## Details

The changes to run-time behaviour are:
* `int64` and `uint64` are full 64-bit precision integers
* casting from unsigned to signed integers performs a proper bitcast
* casting float to int and vice versa performs a bitcast instead of
  being a no-op
* converting `bool` to numeric types yields either 1 or 0 instead of
  being a no-op
* converting from float to unsigned int performs a truncation +
  clamp-to-range instead of being a no-op
* converting from float to signed int clamps to range instead of
  truncating and bitcasting
* `uint32` multiplication returns the correct result for large operand
  values now
* left-shifting `uint16` and `uint8` doesn't yield invalid values
  anymore
* inverting the bits of `uint16` and `uint8` via `not` doesn't yield
  invalid values anymore

The main intention was to implement support for 64-bit integers for the
JavaScript backend, with some of the other fixes being required by
the 64-bit integer implementation.

Internally, `jsgen` is changed such that it uses different paths for
`int64` and `uint64` compared to the other integer types, both for
constants and when used as operands. The implementation for the various
64-bit integer operations is provided as procedure in the new `jsint64`
system module -- they're pure implementation as much as possible in
order for them to still work without change when switching to `asm.js`.

When used as an index, 64-bit integer are internally converted to a
`float64` (a JS `Number`) first, which keeps the behaviour the same as
it was previously.

Similarly, for range checks, 64-bit integers are first converted to
`float64`, meaning that range checks with values outside the
32-bit integer range still don't work. The reason for this is to not
having to duplicate the logic in `rtchecks` into `jsgen` for now
(`rtchecks` should be enabled for the JavaScript target eventually).

Finally:
* the 64-bit integer conversion to `JsBigInt` in the standard library
  are adjusted such that they keep the full precision
* internal adjustments are made to the `hashes` and `times` modules
  where the previous `cast` and `int` behaviour was relied on
* various tests related to casting, conversions, and bit operations are
  enabled for the JavaScript backend

### DIFF
--- a/lib/system/jsint64.nim
+++ b/lib/system/jsint64.nim
@@ -317,8 +317,7 @@ proc castFloatToInt64(val: float32): Int64 {.compilerproc, asmNoStackFrame.} =
     var buf = new ArrayBuffer(4);
     (new Float32Array(buf))[0] = `val`;
     var view = new Uint32Array(buf);
-    `result`.lo = view[0];
-    `result`.hi = 0;
+    return {lo: view[0], hi: 0};
   """
 
 proc castDoubleToInt64(val: float64): Int64 {.compilerproc, asmNoStackFrame.} =
@@ -326,8 +325,7 @@ proc castDoubleToInt64(val: float64): Int64 {.compilerproc, asmNoStackFrame.} =
     var buf = new ArrayBuffer(8);
     (new Float64Array(buf))[0] = `val`;
     var view = new Uint32Array(buf);
-    `result`.lo = view[0];
-    `result`.hi = view[1];
+    return {lo: view[0], hi: view[1]};
   """
 
 proc castInt64ToDouble(val: Int64): float64 {.compilerproc, asmNoStackFrame.} =
@@ -336,7 +334,7 @@ proc castInt64ToDouble(val: Int64): float64 {.compilerproc, asmNoStackFrame.} =
     var view = new Uint32Array(buf);
     view[0] = `val`.lo;
     view[1] = `val`.hi;
-    `result` = (new Float64Array(buf))[0];
+    return (new Float64Array(buf))[0];
   """
 
 proc castInt64ToFloat(val: Int64): float32 {.compilerproc, asmNoStackFrame.} =
@@ -344,7 +342,7 @@ proc castInt64ToFloat(val: Int64): float32 {.compilerproc, asmNoStackFrame.} =
     var buf = new ArrayBuffer(4);
     var view = new Uint32Array(buf);
     view[0] = `val`.lo;
-    `result` = (new Float32Array(buf))[0];
+    return (new Float32Array(buf))[0];
   """
 
 {.pop.}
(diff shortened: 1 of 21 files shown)


lib/system/jsint64.nim
```diff
@@ -0,0 +1,350 @@
...
+  else:
+    divInt64(a, b)
+
+proc checkedModInt64(a, b: Int64): Int64 {.compilerproc.} =
+  if (b.lo == 0 and b.hi == 0):
+    raiseDivByZero()
+  else:
+    modInt64(a, b)
+
+# MARK: cast operations
```
saem, line 313: I see you've discovered the joys of `# MARK: ...` as well. 😆

zerbina: I started using them in place of `# ------ <section description goes here> -----` after you mentioned on Matrix that `# MARK: ...` their highlighted in the sidebar with VSCode.

zerbina:
Aside from making 64-bit integer conversion to `JsBigNum` work and adjusting a few places (`hashes` and `times`) to make tests pass, this PR deliberately doesn't include any standard library cleanup/adjustment in order to stay focused on just the feature and related fixes.

zerbina:
From my testing, it looks like there's a bug with the Windows distribution of `node` (version < 24) where this minimized snippet runs into an internal error:

```js
switch (0) {
case 2147483642:
case 2147483643:
case 2147483644:
case 2147483645:
case 2147483646:
case 2147483647:
}
```

For the `tcase_with_uint_values.nim` test, the `switch`/`case` statement generated for the `uint` instance of `inRange` includes these (seemingly) problematic `case` values.

While `jsgen` could be made to accommodate `node` in this case, I think the issue is too fringe to warrant the effort, so I've disabled the test on Windows for now, with a comment detailing the reason.

metagn:
Kudos for the work here, I am curious how much faster this is than bigints in case you did any benchmarks. From what I remember VMs used to not be mature enough to optimize for smaller bigint values but maybe the situation has changed. Also if there is a difference between wrapping in an object or using 2 inlined int values.

saem:
> Kudos for the work here

@zerbina really hit it out of the park.

> I am curious how much faster this is than bigints in case you did any benchmarks. From what I remember VMs used to not be mature enough to optimize for smaller bigint values but maybe the situation has changed.

At least with V8 `BigInt` and friends will use SIMD instructions for accelerating operations, at least on x86. On other architectures (e.g.: RISC-V, you can end up hitting unaligned access issues). I stopped reading up on JS/V8 engine optimizations by the time this work was really kicking into high gear so I'm not sure about more recent work here.

> Also if there is a difference between wrapping in an object or using 2 inlined int values

IIRC, both V8, and spidermonkey (mozilla's engine), at the very least could detect same objects with unchanging keys and create the equivalent of a structure for them. The only issue that I can recall is that both will likely run into issues if you use full 32bits (as in the value actually hits the limit), because then it goes from SMI (small machine integers?, which are 31bits) and need some extra indirection. I would be shocked if Safari's engine wasn't in roughly the same boat. Again, this is all if memory serves.

PR #1696, opened 2026-03-22
sempass2: fix crash with tuple-in-array edge case

### LINKED ISSUE #1695: t01_array fails with SIGSEGV with release compiler
### Example

```
$ ./koch.py temp r tests/lang/s01_basics/s08_primitive_compound_types/t01_array.nim
```

### Actual Output

```
/nimskull/bin/nim_temp  r tests/lang/s01_basics/s08_primitive_compound_types/t01_array.nim
Traceback (most recent call last)
/nimskull/compiler/nim.nim(151) nim
/nimskull/compiler/nim.nim(88) handleCmdLine
/nimskull/compiler/front/main.nim(522) mainCommand
/nimskull/compiler/front/main.nim(477) compileToBackend
/nimskull/compiler/front/main.nim(207) commandCompileToC
/nimskull/compiler/backend/cbackend.nim(93) generateCode
/nimskull/compiler/backend/backends.nim(884) generateCode
/nimskull/compiler/backend/backends.nim(469) translate
/nimskull/compiler/backend/backends.nim(434) process
/nimskull/compiler/mir/injecthooks.nim(311) injectHooks
/nimskull/compiler/sem/injectdestructors.nim(372) injectHooks
/nimskull/compiler/ast/ast_query.nim(195) add
FAILURE
```

### Expected Output

Test passes

zerbina:
## Summary

Fix the compiler crashing in an edge case where a tuple type used as
the element type of an array is not used anywhere.

## Details

Synthesizing the hooks for a type embedding a tuple type doesn't, by
itself, synthesize the hooks for the tuple type. For array construction
expressions, `sempass2` didn't separately pass the the array's element
type to `createTypeBoundOps`, leaving the tuple type without hooks
(unless synthesized elsewhere), causing the compiler to crash when
the hooks were needed (happens when the result of a sub-expression is
materialized).

Passing the array's element type to `createTypeBoundOps` fixes the
issue.

Fixes https://github.com/nim-works/nimskull/issues/1695 .

<!--
Pull Request(PR) Help

Before Merge Ensure:

See contributor
(guide)[https://nim-works.github.io/nimskull/contributing.html]

Tips that make PRs easier:
* for big/impactful changes, start with chat/discussions to refine
ideas
* refine the pull request message over time; don't have to nail it in
one go

### DIFF
--- a/tests/lang_objects/destructor/tmissing_array_element_hooks_bug.nim
+++ b/tests/lang_objects/destructor/tmissing_array_element_hooks_bug.nim
@@ -1,0 +1,19 @@
+discard """
+  description: '''
+    Regression test for a compiler crash in the edge case where the tuple type
+    used as an error is not used anywhere and a temporary has to be
+    materialized.
+  '''
+  joinable: false
+"""
+
+type Object = object
+
+proc `=destroy`(x: var Object) =
+  discard
+
+proc test(): Object = discard
+
+# how the second element expression looks like is irrelevant as long as
+# it conditionally raises
+var arr = [(Object(),), (test(),)]
--- a/tests/lang_objects/destructor/tmissing_array_element_hooks_bug.nim
+++ b/tests/lang_objects/destructor/tmissing_array_element_hooks_bug.nim
@@ -1,8 +1,8 @@
 discard """
   description: '''
     Regression test for a compiler crash in the edge case where the tuple type
-    used as an error is not used anywhere and a temporary has to be
-    materialized.
+    used as an array element type is not used anywhere else and a temporary of
+    said tuple type has to be materialized.
   '''
   joinable: false
 """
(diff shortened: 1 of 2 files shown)


tests/lang_objects/destructor/tmissing_array_element_hooks_bug.nim
```diff
@@ -0,0 +1,19 @@
+discard """
+  description: '''
+    Regression test for a compiler crash in the edge case where the tuple type
+    used as an error is not used anywhere and a temporary has to be
```
saem:
```suggestion
    used as an element is not used anywhere and a temporary has to be
```

Had trouble parsing this, which could be me misunderstanding, but I think this is what was meant as the anonymous tuple `tuple [0: Object]` is never otherwise instantiated so when the temporary must be instantiated for the first element it would have crashed.

saem (approved):
One minor correction, otherwise it looks good.

tests/lang_objects/destructor/tmissing_array_element_hooks_bug.nim
zerbina: I've adjusted the description a bit with the goal of making it clearer. Let me know if it's better now.

saem: Much better, thanks.

PR #1697, opened 2026-03-25
remove obsolete JS accommodations from the standard library

zerbina:
## Summary

Remove various JavaScript backend accommodations made obsolete by
64-bit integer support and related fixes. Except for `float` hashing
now working the same regardless of  `int`  size, behaviour doesn't
change.

## Details

* `math`: remove usage of embedded JavaScript code for fetching and
  modifying the bit patterns of `float` values
* `hashes`: replace `imul` usage with built-in multiplication
* `hashes`: replace `hiXorLoJs` (based on `BigInt`) usage with the pure
  `hiXorLo` 
* `hashes`: use bit casting instead of JS-specific workarounds for
  accessing bit pattern of `float` values
* remove workarounds for 64-bit integer overflow checks not working in
  `parseutils`, `times`, and `monotimes` 

For hashing of `float` values, always cast to `int64` instead of `Hash`
(which is an alias of  `int`  and thus has a target-dependent size)
first,
making `hash(float)` target-independent.

### DIFF
--- a/lib/pure/hashes.nim
+++ b/lib/pure/hashes.nim
@@ -127,37 +127,6 @@ proc hiXorLo(a, b: uint64): uint64 {.inline.} =
     else:
       result = hiXorLoFallback64(a, b)
 
-when defined(js):
-  import std/jsbigints
-  import std/private/jsutils
-
-  proc hiXorLoJs(a, b: JsBigInt): JsBigInt =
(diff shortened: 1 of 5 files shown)


saem (approved):
Gotta love that add/delete ratio :D

PR #1700, opened 2026-04-10
remove the `canonical` filename rendering option

Luyten-Orion:
## Summary
Removed the  `--filenames:canonical`  option and related logic from the
compiler.

## Details
* Removed `foCanonical` and canonical filename output in the compiler.
* Patched testament so it no longer uses canonical filenames.
* Amended the CLI and relevant tests to no longer use the canonical
option.
* This change addresses the fact that  `--filenames:canonical`  often
produced paths that weren't real paths on the filesystem, which made it
difficult to track down errors to their actual modules.

### DIFF
--- a/nimsuggest/nim.cfg
+++ b/nimsuggest/nim.cfg
@@ -1,1 +1,0 @@
-# xxx not sure why this flag isn't needed: processing:filenames
--- a/nimsuggest/nim.cfg
+++ b/nimsuggest/nim.cfg
@@ -1,2 +1,1 @@
 # xxx not sure why this flag isn't needed: processing:filenames
-filenames:canonical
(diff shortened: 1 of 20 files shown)


nimsuggest/nim.cfg
zerbina, line 1: Please remove this file.

zerbina (changes requested):
The changes themselves are good, but there's two additional things that need to be changed:
* please also update `advopt.txt` (remove the mention of `canonical`)
* please inline `canonicalImportAux` into `canonicalImport`, as the former only has a single caller now (and it shouldn't have existed in the first place)

For the PR message, what's currently listed in the summary belongs into the details section; the summary should only mention that `--filenames:canonical` is no more (I've adjusted the title already), so that the high-level impact on the compiler is easy to parse. References to other PRs should also not be part of what later becomes the commit message.

One thing that the PR message should mention as a reason for this change (in addition to unblocking the package manager integration rework) is that with the `canonical` option, file paths that - by design - aren't real paths were output (meaning it's not possible to resolve errors or other message containing those paths to physical modules), making it a dubious option.

zerbina (approved):
Looks good, thank you!

PR #1706, opened 2026-05-02
require `std` prefix for standard library module imports

zerbina:
## Summary

Make the `std` prefix a requirement for standard library modules and
remove the search path fallback for it; an import path starting with
`std` will never be treated as relative path now. This is a step
towards requiring qualified import paths for all foreign-package
module imports.

## Details

* fully qualify all standard library module imports in the compiler,
  tools, tests, documentation, and samples
* remove the relative path fallback treatment for import paths starting
  with `std/`, making it behave like a package name
* add `std` as a standard-library source directory. Needed now that
  `std/xxx` is never treated as a relative path
* update the documentation of `options.findModule` to match reality
* remove all `lib/` subdirectory additions to the search path from the
  global `nim.cfg`
* remove the recommendation to use the `std` prefix for standard
  library modules from the manual
* normalize the trailing newlines in all test files where imports are
  adjusted

The `lib` path is still added as a search path, so that the existing
imports of system modules continue to work as they are.


tests/gc/bintrees.nim
```diff
@@ -1,6 +1,6 @@
 # -*- nim -*-
 
-import os, strutils
+import std/os, strutils
```
saem:
```suggestion
import std/[os, strutils]
```

I might be misunderstanding the point of this module, but I believe both need to be prefixed with `std`

tests/stdlib/formats/tjsontestsuite.nim
saem, line 1: The unicode seems to have been altered in this file, probably some issue during a save/autotransformation of the encoding?

saem (commented):
This is a really comprehensive change, nicely done.

I found a couple of minor items, at least one of them might be me misunderstanding the module usage.

tests/gc/bintrees.nim
zerbina: The import does need to be prefixed, yea.

The `bintrees` module is a standalone program that isn't built and run as part of the test suite, hence no error being reported.

saem (approved):
LGTM

PR #1711, opened 2026-05-26
std: enumutils no longer warns iterating over holey enums

saem:
## Summary

`enumutils`  no longer results in a warning of conversion when iterating
over a holey enum.

## Details

The internal macro  `enumFullRange`  used by  `enumutils.items` 
expanded to a constant  `set` , and iteration of that set via  `items` 
performs a conversion from int to the holey enum type internally,
causing the warning. Returning an array instead of a set solves this.

### DIFF
--- a/lib/std/enumutils.nim
+++ b/lib/std/enumutils.nim
@@ -12,7 +12,7 @@ from std/typetraits import OrdinalEnum, HoleyEnum
 
 macro enumFullRange(a: typed): untyped =
   let typ = getTypeImpl(getType(a)[1]) # the ``nnkEnumTy`` AST
-  nnkPrefix.newTree(ident"@", newNimNode(nnkBracket).add(typ[1..^1]))
+  newNimNode(nnkBracket).add(typ[1..^1])
 
 # xxx `genEnumCaseStmt` needs tests and runnableExamples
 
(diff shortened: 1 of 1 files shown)


lib/std/enumutils.nim
```diff
@@ -12,7 +12,7 @@ from std/typetraits import OrdinalEnum, HoleyEnum
 
 macro enumFullRange(a: typed): untyped =
   let typ = getTypeImpl(getType(a)[1]) # the ``nnkEnumTy`` AST
-  newNimNode(nnkCurly).add(typ[1..^1])
+  nnkPrefix.newTree(ident"@", newNimNode(nnkBracket).add(typ[1..^1]))
```
zerbina:
```suggestion
  newNimNode(nnkBracket).add(typ[1..^1])
```

zerbina (changes requested):
Having `enumFullRange` return a `seq`, even if constant, is unnecessary -- an array suffices and is easier to construct (and doesn't depend on the correct symbol being bound to `@` where the macro is expanded).

For the PR title and description summary, could you clarify that the warning only happened when iterating over holey enum fields with `items`, and not when using `enumutils`-provided routines with holey enums in general?

saem:
thanks for the review, I've updated the title and summary

zerbina (approved):
Looks good to me, thanks. I've modified the details section a bit to remove what read a bit redundant to me (and also changed "sequence" to "array").

If you're okay with these changes, feel free to kick-off the merge.

PR #1712, opened 2026-05-26
std: `random` unify c and js implementations

saem:
## Summary

The `random` module now behaves consistently across C and JS backends.

## Details

The std  `random`  module now uses the same implementation for both C
and JS backends. Mirroring state, seeding approach, bit width, etc.

The only difference is that the C backend supports threading, and the JS
backend initializes random state slightly differently, but the behaviour
remains consistent.

### DIFF
--- a/lib/pure/random.nim
+++ b/lib/pure/random.nim
@@ -78,14 +78,7 @@ import std/private/since
 include system/inclrtl
 {.push debugger: off.}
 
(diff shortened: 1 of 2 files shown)


zerbina (approved):
Nice, thanks. Another step away from unnecessary dialects and discriminating between backends.

PR #1719, opened 2026-06-29
sem: add and use new error detection strategy

zerbina:
## Summary

Prepare for the incremental removal of `nkError` wrapping by adding a
separate mechanism for disabling macro evaluation, based on deep-
scanning the invocation AST for errors.

### Rationale

Errors, be it `nkError` nodes or `tyError` types, must not reach into
either macros or compile-time code generation. This is currently
enforced through `nkError` wrapping, with macro invocations (and other
compile-time evaluation) not being evaluated if either the callee or
any of the arguments is an `nkError` node.

Since the plan is to remove `nkError` wrapping, there needs to be
another way to reliably disable compile-time evaluation.

### Implementation

1. add the `tfHasError` flag for marking compound types as containing
   errors somewhere and propagate it through types via the
   `propagateToOwner` mechanism
2. scan AST used in or connected with compile-time evaluation for
   reachable error AST or types via the new `areErrorsReachable`
   procedure, and disable evaluation of the code in question when
   errors are found

The after-the-fact scanning for AST is chosen because it's transparent
to the rest of semantic analysis, making it a small impact change (in
terms of added code) that's trivial to revert, should a better approach
to disabling compile-time evaluation emerge.

Traversing the AST is not free, so the result of scanning the AST of
routines is cached in the module graph.

### DIFF
--- a/compiler/ast/ast.nim
+++ b/compiler/ast/ast.nim
@@ -392,13 +392,16 @@ proc isGCedMem*(t: PType): bool {.inline.} =
            t.kind == tyProc and t.callConv == ccClosure
 
 proc propagateToOwner*(owner, elem: PType; propagateHasAsgn = true) =
-  owner.flags.incl elem.flags * {tfHasMeta, tfTriggersCompileTime}
+  owner.flags.incl elem.flags * {tfHasMeta, tfHasError, tfTriggersCompileTime}
   if tfNotNil in elem.flags:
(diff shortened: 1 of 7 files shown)


zerbina:
This is not great, but it's the best solution that I could come up with that doesn't require redoing all of `sem` in one go (or significantly altering the language). Once error propagation via `nkError` wrapping is removed, maybe some better approach surfaces.

saem (approved):

PR #1720, opened 2026-07-05
sem: replace diagnostic-based error recovery for symbols

zerbina:
## Summary

Replace the diagnostic-based error recovery for symbol production with
creating a recovery symbol right away, which is a lot simpler.

## Details

1. turn `newSymGNode` from a procedure returning a node into one
   returning a symbol, and rename the procedure to `produceSymbol`.
   Errors are emitted right away, with the recovery symbol returned
   directly
2. update all callsites of the former `newSymGNode`; they don't need to
   handle error recovery anymore 
3. remove the now-obsolete routines for working with potentially
   erroneous symbol nodes (`defNameErrorNodeAllowsSymUpdate` and
   `getDefNameSymOrRecover`)
4. remove the diagnostics meant for carrying recovery symbols
5. add an `emit` procedure to the `errorhandling` module, for emitting
   diagnostics directly and without going through `newError`. This
   means that it's now possible for diagnostics that have no
   `wrongNode` attached to reach diagnostic handling
6. add a dedicated variant for `adSemDotOperatorsNotEnabled` and
   `adSemCallOperatorsNotEnabled`, so that they don't rely on the
   diagnostic having an attached `wrongNode`

For the AST produced by type-checking, this means that there's always a
valid `nkSym` node in symbol definition positions, even if the
untyped AST at that position is bogus (e.g., not an identifier, some
illegal symbol, etc.).

### DIFF
--- a/compiler/ast/ast_types.nim
+++ b/compiler/ast/ast_types.nim
@@ -1365,8 +1365,6 @@ type
         adSemSelectorMustBeOfCertainTypes,
         adSemInvalidPragmaBlock,
         adSemConceptPredicateFailed,
-        adSemDotOperatorsNotEnabled,
-        adSemCallOperatorsNotEnabled,
         adSemUnexpectedPattern,
         adSemCannotBeRaised,
         adSemCannotRaiseNonException,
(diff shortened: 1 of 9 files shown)


saem (approved):
This is much better, this was about the time where I lost faith in nkError, but couldn't see a way out beyond transitioning to diagnostics and reworking the complected reports

PR #1724, opened 2026-09-23
support `system` directory being separate from `lib`

zerbina:
## Summary

Support `nimbase.h` and `system.nim` being located within the `system`
directory, and allow overriding the latter's location via the new
`--system` switch. 

## Details

* add the `systemPath` configuration field, configurable on the CLI via
 the new  `--system` switch and defaulting to the default `libpath`
* base the test on whether a directory is a valid `lib` directory based
  on the presence of `stdlib.nimble`, as `system.nim` may now be located
  within the `system` directory
* look for `system.nim` in the `system` directory first, before
  falling back to looking in the `lib` directory
* look for `nimbase.h` in the `system` directory first, before falling
  back to the `lib` directory 
  * `nimbase.h` is tied to a specific compiler version and thus needs
    to be located in `system` too
* treat `system/` as a package name when resolving module paths, with
  the remainder treated as a mode path relative to `systemPath`
* replaced all double-tick quoted `system` module import paths, which are
  turned into OS-specific relative paths, with regular import paths

`nimbase.h` and `system.nim` cannot be moved into `system` right now,
nor can the `system` directory moved elsewhere, as the csources compiler
still expects them to be in the current location.

---

## Notes for Reviewers
* the first step towards decoupling the compiler interface and runtime from the standard library
* with these changes, the `system` directory for the csources compiler can be shipped alongside it, which will remove the need to keep the code in the system directory backwards compatible with the csources compiler 
* before actually shipping the system directory with the csources compiler, some changes to system and the standard library should be made first, however, so as to keep the number of csource updates smaller
* once the csources compiler has a configurable system directory location, the lookup fallbacks can be removed from the compiler, and the system directory can be moved elsewhere (if desired)

<!--
Pull Request(PR) Help

Before Merge Ensure:

Tips that make PRs easier:


saem:
quick question, what does `mode path` mean in the following fragment from the PR message?

> the remainder treated as a mode path relative to `systemPath`

zerbina:
> quick question, what does `mode path` mean in the following fragment from the PR message?

Ah, my bad, "mode" was supposed to be "module" -- I've corrected the mistake now.

